PR #1075, opened 2020-11-12
Use OS PRNG

joyceerhl:


DavidKutu (approved):

src/test/datascience/interactive-common/randomBytes.functional.test.ts
```diff
@@ -0,0 +1,29 @@
+import { ISystemPseudoRandomNumberGenerator } from '../../../client/datascience/types';
+import { assert } from 'chai';
+import { DataScienceIocContainer } from '../dataScienceIocContainer';
+
+suite('DataScience - RandomBytes', () => {
```
rchiodo, line 5: This seems like it could be a unit test? It doesn't need to run as part of other functional tests.

rchiodo: Instead of using the container, you'd just instantiate the Pseudo random number generator.

rchiodo (approved):

src/test/datascience/interactive-common/randomBytes.functional.test.ts
joyceerhl: I wanted to test with a real process, not a mocked one, and I was under the impression that we mock dependencies (especially resource-intensive dependencies) in unit tests. If it's OK to use real processes I can do that!

joyceerhl: Blech just noticed the functional test is running with a mocked process anyway. Will redo the test

.github/workflows/insiders.yml
```diff
@@ -31,7 +31,7 @@ env:
 jobs:
   build-vsix:
     name: Build VSIX
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
```
joyceerhl, line 34: Build on Windows, since the napi libraries are only needed on Windows. So we don't need a build step to generate the binaries for all platforms.

joyceerhl:
ref-napi is 2+MB, ffi-napi is 7+MB :(

IanMatthewHuff (approved):

.github/workflows/insiders.yml
rchiodo: Does this mean we have to ship the windows binaries? So this HAS to build on windows? If so, we need this in the release.yml too.

rchiodo (changes requested):
⏲️

src/test/datascience/interactive-common/randomBytes.vscode.test.ts
```diff
@@ -0,0 +1,50 @@
+import { ISystemPseudoRandomNumberGenerator } from '../../../client/datascience/types';
```
rchiodo, line 1: Wait do we need to run this inside of VS code? Can't this just be a unit test?

joyceerhl: @DonJayamanne recommended running this in a VSCode test because we want to make sure this actually works, and across all platforms

rchiodo: It's not actually using it inside of a window (which I'd argue one of the other tests will cover). It's just using the functional  test DataScienceIOC container.

A unit test would essentially be the same but would not need the container or the vscode bring up and runs on all platforms already.

DonJayamanne: > A unit test would essentially be the same but would not need the container or the vscode bring up and runs on all platforms already.

Agreed, if we dont need the DI container, then we can just use a unit test.

src/client/datascience/interactive-ipynb/randomBytes.ts
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    // Calls into BCryptGenRandom in bcrypt.dll using node-ffi
+    // See https://github.com/node-ffi/node-ffi/wiki/Node-FFI-Tutorial for details on usage
+    private randomBytesForWindows(numBytes: number) {
+        try {
+            // tslint:disable: no-require-imports
+            // Lazy-load modules required for calling BCryptGenRandom
+            const ffi = require('ffi-napi');
```
DonJayamanne, line 36:
```suggestion
            const ffi = require('ffi-napi') as typeof import('ffi-napi);
```

This way you get strong typing.
If you don't need then then we should uninstall the @types/ffi-napi & other type definitinos.

DonJayamanne (approved):

src/test/datascience/interactive-common/randomBytes.vscode.test.ts
joyceerhl: DI makes it way easier to resolve the dependencies of this class 🤣 But yup will rewrite as unit test

rchiodo (approved):

src/test/smoke/datascience.smoke.test.ts
```diff
@@ -3,30 +3,49 @@
...
+import { ISystemPseudoRandomNumberGenerator } from '../../client/datascience/types';
+import { IExtensionTestApi, openFile, setAutoSaveDelayInWorkspaceRoot, waitForCondition } from '../common';
 import { EXTENSION_ROOT_DIR_FOR_TESTS, IS_SMOKE_TEST } from '../constants';
 import { noop, sleep } from '../core';
 import { closeActiveWindows, initialize, initializeTest } from '../initialize';
 
 const timeoutForCellToRun = 3 * 60 * 1_000;
-
-suite('Smoke Test: Interactive Window', () => {
+suite('Smoke Tests', () => {
```
DonJayamanne, line 19: Thanks for renaming this.

DonJayamanne:
@joyceerhl please could you ensure we have smoke tests running on all three platformas. In the insiders (optionally in PRs - not necessary, but definitely in insiders.)
We need to ensure the code work in all three platforms & when using the bundled code (VSIX)

Issue #1074, opened 2020-11-11
Insiders build requires python


rchiodo:
Install just the jupyter extension
Open a notebook
You'll be told you have to install the python extension

![image](https://user-images.githubusercontent.com/19672699/98877778-1db27280-2436-11eb-9d67-1f48eb14d0e2.png)

First dialog is not expected. Second is.

First dialog is put up by something forcing interpreter search.

DonJayamanne:
See here for similar suggestions no merging these two prompts https://github.com/microsoft/vscode-jupyter/issues/1092

rchiodo:
Validated

PR #1100, opened 2020-11-12
Ensure native notebooks can be used without Python or Python extension

DonJayamanne:
For #1074
Separate test where Python extension is not installed.


.github/workflows/pr.yml
```diff
@@ -560,7 +564,27 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           report_paths: ${{ env.TEST_RESULTS_GLOB }}
           check_name: Notebook Test Report
-        if: (steps.test_notebook_vscode.outcome == 'failure' || steps.test_notebook_vscode.outcome == 'failure') && failure()
+        if: steps.test_notebook_vscode.outcome == 'failure' && failure()
+
+      - name: Run Native Notebook with VSCode & Jupyter (without Python)
```
DonJayamanne, line 569: Specific tests without python extension & python.
This way we never break & gets tested on CI.

src/client/datascience/notebook/kernelProvider.ts
```diff
@@ -135,7 +137,7 @@ export class VSCodeKernelPickerProvider implements NotebookKernelProvider {
                 undefined,
                 token
             ),
-            this.interpreterService.getActiveInterpreter(document.uri)
+            isPythonNb ? this.interpreterService.getActiveInterpreter(document.uri) : Promise.resolve(undefined)
```
rchiodo, line 140: Won't this still ask if the notebook is a python notebook? What we really want is to do nothing with the interpreters unless the python extension is installed. See IPythonExtensionChecker

rchiodo (changes requested):
⏲️

src/client/datascience/notebook/kernelProvider.ts
DonJayamanne: Oh ok, thanks didn't expect that.
Makes sense

DavidKutu (approved):

src/client/datascience/notebook/kernelProvider.ts
rchiodo: This doesn't seem to be resolved yet.

DonJayamanne: Damn it, yes

DonJayamanne: Done

DonJayamanne:
Need to fix the failing test.

rchiodo (approved):

Issue #250, opened 2020-10-19
Implement Variable Explorer in VS Code sidebar


greazer:

rchiodo:
Validated. I can see the variable view.

PR #3914, opened 2020-11-14
Variable View Panel (part deux)

IanMatthewHuff:
For #250


IanMatthewHuff:
So I mentioned at standup that I'd taken a wrong tact earlier in the week on this and was walking it back. For context, initially I was approaching this as creating more of a new variable view react + redux control. However when implementing that I started to run into more and more things that I needed to change or alter about either how we configured all the redux mappings or the code of the variable view. Part way into that I felt that I was losing the point of being able to quickly reuse the existing control with how much I was changing and if I really needed to change so much. The variable panel had too much dependency on the existing main state and high level controls to be able to just pop it out and use it. 

What I went with instead (as seen in this PR) is that I started with basically the interactivePanel from the interactive window and created the new VariableViewPanel based very closely off of that. Instead of a new variable view we have a VariableViewPanel that is basically the following:
1. Start with most of the code from interactivePanel
2. For rendering just render the VariablePanel
3. Rip out all the actions / mapping / ect... that don't apply to the VariablePanel
Now that the existing VariablePanel is hosted in something that feels just like the InteractivePanel I was able to use it 1:1 without any customization and with basically the same redux mapping methods and sharing actions. I feel this was a much better approach given the goal of getting this working with the least amount of code change from the existing variable view.

At this point all the UI is hooked up minus any layout tweaking that we might need at the end. Last step is to hook up variable fetching from the active native notebook. Right now I just have test data in to check that the control works.

package.json
```diff
@@ -1708,16 +1708,16 @@
             }
         ],
         "viewsContainers": {
-            "activitybar": [
+            "panel": [
                 {
-                    "id": "jupyter",
+                    "id": "jupyter-variables",
```
IanMatthewHuff, line 1713: New ID here. We want this placed in the Panel by default, not in the activityBar. But for Don's new jupyter remote stuff (and other activity bar items) we do want them in the activity bar. So created a new ID and changed the placement to panel. Now it shows up by default where we want it.

src/client/datascience/variablesView/variableView.ts
```diff
@@ -51,4 +53,54 @@ export class VariableView extends WebviewViewHost<IVariableViewMapping> implemen
...
+        payload: any,
+        handler: (args: M[T]) => void
+    ) {
+        const args = payload as M[T];
+        handler.bind(this)(args);
+    }
+
+    // This is called when the UI side requests new variable data
+    private async requestVariables(args: IJupyterVariablesRequest): Promise<void> {
+        // For now, this just returns a fake variable that we can display in the UI
```
IanMatthewHuff, line 83: This is where part 3 will hook up. Probably with a new NativeNotebookActiveVariables service or something to that effect. Just put in a fake variable for now to test the UI and messages working.

src/datascience-ui/variable-view/variableViewPanel.tsx
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        // Return our variable panel, we wrap this in one more top level element "variable-view-main-panel" so that
+        // we can size and host it differently from the variable panel in the interactive window or native editor
+        return (
+            <div id="variable-view-main-panel" role="Main" style={dynamicFont}>
+                <button onClick={this.props.toggleVariableExplorer}>OPEN</button>
+                {this.renderVariablePanel(this.props.baseTheme)}
+            </div>
+        ); // NOTE: Currently the OPEN, button just exists to mimic the toggling of the variable view, make it easier to test when working
```
IanMatthewHuff, line 55: The OPEN button is a slight hack that I want for now, and will remove later. Just an easy way for me to toggle on and off the variable explorer (in final version it will just always be open) as that triggers the variable fetching code to run. This is all behind an experiment still so customers won't see the view or anything in it.

package.json
rchiodo: I'm confused. Is this going to be in the same place as the remote stuff or its own panel?

src/client/datascience/variablesView/variableView.ts
joyceerhl: Does this work not also apply to the interactive window?

package.json
rchiodo: Somewhat related, is this panel only going to show up for native notebooks? Or will it show up for old webviews too?

src/client/datascience/variablesView/variableView.ts
rchiodo: I believe you should be able to use IJupyterVariables to handle these requests. It should work with native notebooks too as they have an INotebook under the covers.

rchiodo (approved):

package.json
IanMatthewHuff: @rchiodo Sorry if my comment didn't make this clear. The change here was that I wanted "Panel" positioning for my view, not "activityBar" positioning. Panel positioning puts it down next to the output and debug console windows. However I think that we want Don's remote stuff (and other general contributions) to live in the activityBar region. So I created a new viewContainer with a new ID here that explicitly docks in the Panel area. 

I see the "jupyter" viewContainer in the activityBar as our general tab / extensibility area, while this ID is specific to the variables. I'd assume that it's explicitly for just variables for now unless we run across more stuff that we want docked in the panel by default.

IanMatthewHuff: And current plan was just for NativeNotebooks. I don't see a reason that we maybe couldn't change the old views to use this as well. But wasn't specced out in the current work. Not fully sure that I would want to for the old notebooks, since they work better with an explorer per document.

rchiodo: Is there something that disables this view for the old stuff? (Hides it if you open an old webview or switch to the interactive window?)

src/client/datascience/variablesView/variableView.ts
IanMatthewHuff: Yeah, I'd not dug in deep yet, but my assumption would be that I'd be able to use IJupyterVariables pretty directly. Figuring most of the work here is just juggling the refresh when your active NativeNotebook document changes in the IDE.

package.json
IanMatthewHuff: Hmm, currently it's just all hidden by the experiment switch. I was planning on using the existing editor context switches that we have for controlling visibility to show for native and hide if native is not active. Honestly though, with you and Joyce mentioning it. Maybe it's even easier to just have it always reflecting the active document, regardless of if native or not. Could give interactive / old webview customers a chance to use this in a panel if they prefer that.

src/client/datascience/variablesView/variableView.ts
IanMatthewHuff: @joyceerhl Not as currently specced. Though I'm starting to think that maybe it's worth considering. I think I mentioned this in another comment, but even if we don't remove the current IW variable view right away we could still have this view added in as an option. Customers might prefer it as they can move / dock it around themselves.

IanMatthewHuff: Is specced even a word? I feel guilty looking at it, will research...

package.json
DonJayamanne: I think this needs to be experimented.
Personally I would prefer to have this displayed side by side, instead of displaying this at the bottom.
Ie. when user clicks on `show variables` button, I think we can display this view (so that it will be visible to the user).

@claudiaregio /cc
I think we should experiment on different placements.

src/client/datascience/variablesView/types.ts
```diff
@@ -1,17 +1,31 @@
...
+    public [InteractiveWindowMessages.ShowDataViewer]: IShowDataViewer;
+    public [InteractiveWindowMessages.GetVariablesRequest]: IJupyterVariablesRequest;
+    public [InteractiveWindowMessages.GetVariablesResponse]: IJupyterVariablesResponse;
+    public [InteractiveWindowMessages.VariableExplorerToggle]: boolean;
+    public [InteractiveWindowMessages.SetVariableExplorerHeight]: IVariableExplorerHeight;
+    public [InteractiveWindowMessages.VariableExplorerHeightResponse]: IVariableExplorerHeight;
     public [CssMessages.GetCssRequest]: IGetCssRequest;
     public [CssMessages.GetCssResponse]: IGetCssResponse;
+    public [CssMessages.GetMonacoThemeRequest]: IGetMonacoThemeRequest; // Needed for started message
+    public [CssMessages.GetMonacoThemeResponse]: IGetMonacoThemeResponse;
```
DonJayamanne, line 24: Why do we need monaco theme in variable view?
We dont' have a monaco editor there, hence this doesnb't make sense.

DonJayamanne: If not required, i think we should remove this

DonJayamanne (approved):

package.json
IanMatthewHuff: @DonJayamanne I'm not fully sure what you mean by side-by-side. Do you mean the activityBar? Right now I'm pretty sure that panel or activitybar are our only two default placement options. 

But on the plus side pretty easy to move this to a new location, and experimenting with that seems smart.

src/client/datascience/variablesView/types.ts
IanMatthewHuff: I think that it was in as it was part of the interactive startup commands and I wanted to have as little custom code as possible between the variable view and interactive / old notebook views. But that seems lazy, pretty sure I can cut this out and it will still work fine, might have to make a new action, but that's easy enough. Will try.

IanMatthewHuff: Chopped out the editor specific bits of startup.

Issue #426, opened 2020-11-09
Native Notebooks - Remove Duplicate Command


claudiaregio:
VS Code surfacing "Notebook: Select Notebook Kernel" command
We are surfacing "Jupyter: Select a Kernel" command

Unless there is a use case for having both that I am not seeing, hide our select a kernel command. 
![image](https://user-images.githubusercontent.com/19276747/98604282-3c7b0280-2298-11eb-9994-5aa28ce1a4d2.png)

rchiodo:
Validated

PR #3929, opened 2020-11-16
Hide webview based command for native notebooks

DonJayamanne:
For #426

### DIFF
--- a/package.json
+++ b/package.json
@@ -1172,7 +1172,7 @@
                     "command": "jupyter.switchKernel",
                     "title": "%DataScience.selectKernel%",
                     "category": "Jupyter",
-                    "when": "jupyter.isnativeactive"
+                    "when": "jupyter.isnativeactive && !notebookEditorFocused"
                 },
                 {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1172,7 +1172,7 @@
                     "command": "jupyter.switchKernel",
                     "title": "%DataScience.selectKernel%",
                     "category": "Jupyter",
-                    "when": "jupyter.isnativeactive"
+                    "when": "jupyter.isnativeactive && !notebookEditorFocused"
```
IanMatthewHuff, line 1175: If this is the webview based command, then why does it need isnativeactive?

IanMatthewHuff: Ohh, wait, is the name just off? isnativeactive covers any old / custom / native notebooks? If so that's kinda funky naming.

IanMatthewHuff (approved):
Approved, though I don't fully grok the naming on that context key.

rchiodo (approved):

Issue #1089, opened 2020-11-12
`shift + enter` broken again ... and again


clayms:
Issue Type: <b>Bug</b>

Every couple of VS Code updates, this keeps breaking.

https://github.com/microsoft/vscode-python/issues/6578

Yet again... pressing <kbd>shift</kbd> + <kbd>enter</kbd> after an update gives:
`command 'python.datascience.execSelectionInteractive' not found`
and does not run the selected code block in the Python Interactive Pane.

Extension version: 2020.11.358366026
VS Code version: Code 1.51.1 (e5a624b788d92b8d34d1392e4c4d9789406efe8f, 2020-11-10T23:34:32.027Z)
OS version: Windows_NT x64 10.0.18363

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-6300U CPU @ 2.40GHz (4 x 2496)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: disabled_off<br>opengl: enabled_on<br>protected_video_decode: unavailable_off<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|15.88GB (4.61GB free)|
|Process Argv|--crash-reporter-id f652528b-5678-4b12-9ca0-3de6b1e1135f|
|Screen Reader|no|
|VM|0%|
</details>

clayms:
looks like it was replaced with `jupyter.execSelectionInteractive` 

are all of these changes to settings and commands documented somewhere?  A lot of other things have broken in the last update.

DonJayamanne:
I'm sorry the migration (from Python to Jupyter extension) hasn't been very smooth for you.
With regards to the commands, all `python.datascience.xyz` commands have been renamed to `jupyter.xyz`.

With regards to documentation, we will look into this, currently our documentation sits in https://code.visualstudio.com/docs/python/jupyter-support

StevenLi-DS:
[...]

Issue #1088, opened 2020-11-12
Some python settings not migrated to jupyter extension correctly (Was: Glitches due to vscode-jupyter transition)


janosh:
It appears the [VS Code Jupyter](https://github.com/microsoft/vscode-jupyter) extension was recently auto-installed on my machine, presumably by the Python extension.

It seems to have done a decent job of porting my existing data science settings over to the new extension, e.g.

```diff
- "python.datascience.magicCommandsAsComments": true,
+ "jupyter.magicCommandsAsComments": true,
```

However, after the transition, the interactive window was no longer showing any code lenses. Took me a minute to figure out that's because the settings had been renamed and they now needed to be called

```diff
- "jupyter.codeLenses": "python.datascience.runcell, python.datascience.runallcellsabove, python.datascience.runcellandallbelow, python.datascience.debugcell",
+ "jupyter.codeLenses": "jupyter.runcell, jupyter.runallcellsabove, jupyter.runcellandallbelow, jupyter.debugcell",
```

Only the JSON key `"python.datascience.codeLenses"` had been automatically renamed, not the string itself. I think it's worth triggering another automatic rewrite of user settings that handles this change so others don't encounter the same issue and perhaps spend longer on figuring out why their code lenses disappeared.

Second, some of my custom keybindings stopped working as well. E.g. I had
```json
{
  "key": "ctrl+shift+enter",
  "command": "python.datascience.runallcells"
}
```

which now just triggers a warning

> command 'python.datascience.runallcells' not found

Perhaps those can be auto-migrated as well.

DonJayamanne:
Thanks for filing this issue & I'm sorry you had to migrate these settings manually.
Looks like we missed some settings when migrating the settings to the new extension.

PR #3931, opened 2020-11-16
Migrate settings.json values and keybindings

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/1088, https://github.com/microsoft/vscode-jupyter/issues/1089

Also rework unit test to test class public interface, not implementation details


src/client/activation/migrateDataScienceSettingsService.ts
```diff
@@ -64,6 +65,12 @@ export class MigrateDataScienceSettingsService implements IExtensionActivationSe
                 k = 'xxxxxx.dataScience.jupyterServerType';
             }
 
+            // If the value contains references to python.dataScience.* commands, migrate those too
+            if (typeof val === 'string') {
+                // There may be multiple occurrences of commands in the object value
+                val = val.replace(/python\.dataScience\./gi, 'jupyter.');
```
DonJayamanne, line 71: What happens if users already created the correct version of the settings.
Now we end up with duplicates.
E.g. if a user sees someone doesn't work, they have two choices:
* Rename existing settings
* Create new settings.

I'm concerned some may have created the new settings. & this could would not create duplicates.

src/client/activation/migrateDataScienceSettingsService.ts
```diff
@@ -72,13 +79,38 @@ export class MigrateDataScienceSettingsService implements IExtensionActivationSe
...
+        if (errors.length > 0) {
+            traceError('JSONC parser returned ParseError codes', errors);
+            return;
+        }
+        if (!Array.isArray(keybindings)) {
+            return;
+        }
+        keybindings.forEach((keybinding) => {
+            if (typeof keybinding.command === 'string') {
+                keybinding.command = keybinding.command.replace('python.dataScience.', 'jupyter.');
```
DonJayamanne, line 99: Same as preivous comment. 
We might end up creating duplicates.
Users may have created the new keyboard mappings & left the old ones.

src/client/activation/migrateDataScienceSettingsService.ts
```diff
@@ -72,13 +79,38 @@ export class MigrateDataScienceSettingsService implements IExtensionActivationSe
...
     }
 
     @traceDecorators.error('Failed to update test settings')
     private async updateSettings(resource: Resource): Promise<void> {
-        const filesToBeFixed = await this.getFilesToBeFixed(resource);
-        await Promise.all(filesToBeFixed.map((file) => this.fixSettingInFile(file)));
+        const filesToBeFixed = (await this.getFilesToBeFixed(resource)).map((file) => this.fixSettingsFile(file));
+        const userCustomKeybindingsFile = this.application.userCustomKeybindingsFile;
+        if (userCustomKeybindingsFile && (await this.fs.localFileExists(userCustomKeybindingsFile))) {
+            filesToBeFixed.push(this.fixKeybindingsFile(userCustomKeybindingsFile));
```
DonJayamanne, line 111: Will this fix the files for users who have already been upgraded?
Or is this code going to work only for users who have not ugraded to Jupyter extension?

src/client/activation/migrateDataScienceSettingsService.ts
```diff
@@ -26,16 +26,17 @@ export class MigrateDataScienceSettingsService implements IExtensionActivationSe
     ) {}
 
     public async activate(resource: Resource): Promise<void> {
-        this.updateSettings(resource).ignoreErrors();
+        await this.updateSettings(resource);
```
IanMatthewHuff, line 29: I'm guessing we need this. But I'm pretty sure this does slow down activation. I take it that this was an issue as other bits were activating before settings were migrated? If it's only slow once that's probably ok.

DonJayamanne (changes requested):
Got some questions & concerns

src/client/activation/migrateDataScienceSettingsService.ts
DonJayamanne: I agree. We might want to re-view the code, as we're looking for files in the method `updateSettings`.
Instead we should check if we have performed the check previously & short circuit (check some flag in global memento) if not alraedy doing that.

DonJayamanne: IO on windows makes a big difference.

src/test/activation/migrateDataScienceSettingsService.unit.test.ts
```diff
@@ -15,130 +15,157 @@ suite('Migrate data science settings', () => {
...
+        "foo",
+        "bar"
+    ],
+    "jupyter.alwaysScrollOnNewCell": true,
+}`;
+    const KEYBINDINGS_FILEPATH = '/path/to/keybindings.json';
+    const originalKeybindings = `[
+    {
+        "key": "ctrl+shift+enter",
+        "command": "python.dataScience.runallcells"
```
rchiodo, line 153: Since you're editing settings in this change, perhaps we can fix the code lens commands for values in this change too?

When we ported over the ```python.dataScience.codeLenses```

joyceerhl: Not sure what this means @rchiodo, would you mind elaborating?

rchiodo: Sorry should have linked a bug. That might make it clearer:
https://github.com/microsoft/vscode-jupyter/issues/1088

Okay I'm an idiot. You're fixing that bug.

src/client/activation/migrateDataScienceSettingsService.ts
joyceerhl: Today this runs on every extension activation, so yes it will fix for users who have already upgraded. But I think we need to add some global state to ensure the migration happens once from here onward.

DonJayamanne: > But I think we need to add some global state to ensure the migration happens once from here onward.

Yes we should.

src/client/common/application/applicationEnvironment.ts
```diff
@@ -41,6 +41,28 @@ export class ApplicationEnvironment implements IApplicationEnvironment {
                 return;
         }
     }
+    public get userCustomKeybindingsFile(): string | undefined {
+        const vscodeFolderName = this.channel === 'insiders' ? 'Code - Insiders' : 'Code';
```
rchiodo, line 45: Could the path for this be acquired from the global storage path (so you wouldn't need special code for each OS)

rchiodo: Meaning go up from the global storage path?

path.resolve will eliminate '..' in paths I believe.

rchiodo (approved):

src/client/common/application/applicationEnvironment.ts
joyceerhl: This is modeled on how we find the user settings file (defined right above this), I didn't want to deviate from that in case there was a particular reason it was implemented that way.

rchiodo: Ah. Maybe whomever wrote that didn't realize they could just go up. 

This won't work when using say VS code OSS bits (the bits you build).

The path if you used the global storage path would be something like:

```
path.resolve(path.join(extensionContext.globalStoragePath, '..', '..', 'keybindings.json'))
```

rchiodo (approved):

src/client/activation/migrateDataScienceSettingsService.ts
DonJayamanne: Couldn't find code that addresses this concern.
Please could you point me in that direction, I may have missed it

src/client/activation/migrateDataScienceSettingsService.ts
```diff
@@ -48,37 +74,88 @@ export class MigrateDataScienceSettingsService implements IExtensionActivationSe
...
+            if (typeof command === 'string' && command.startsWith('jupyter.')) {
+                return command;
+            }
+        });
+        const migratedKeybindings: IKeyBinding[] = [];
+        keybindings.forEach((keybinding) => {
+            const command = keybinding.command;
+            if (typeof command === 'string' && command.startsWith('python.dataScience.')) {
+                const targetCommand = `jupyter.${command.substr(19)}`;
+                if (!jupyterCommands.includes(targetCommand)) {
```
joyceerhl, line 136: @DonJayamanne here's the code for checking for dupes. First collect all the jupyter commands, then when migrating the python.dataScience commands, only include them if the equivalent jupyter command isn't mapped to something in the keybindings.json

DonJayamanne (approved):

Issue #398, opened 2020-11-04
Native Notebooks - "Getting Started" Kernel Behavior


claudiaregio:
When a user is working with a new empty notebook (any number of cells but all empty) and changes the kernel to a different language, have language picker for empty cell match that of the kernel.

IanMatthewHuff:
@DonJayamanne Looks like there is a bug here when validating. 

Two empty python cells, kernel set to python:
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/3af5944d-f649-4c75-8082-97fd58f369eb)

Change kernel to java kernel:
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/15bd6943-0e30-4501-a9b7-2a1f6bece928)

I'd expect two empty java cells. Instead I have two empty java cells and one empty python cell.

DonJayamanne:
Thats odd, I can see that as well.

IanMatthewHuff:
Maybe a vscode side issue with committing the change? Wasn't sure, so pulled it back to look at.

DonJayamanne:
Looks like a VS Code issue.
If there is one cell, then it works, if we have 2 ceels we end up with 1 extra.
If there are 3 cells, then we end up with another extra 2 cells..

claudiaregio:
Does any extra work need to be done to enable the .NET team's experience to work here? Their .NET interactive kernel supports multiple languages, and comes with a default language. In their user's case they would need to have the cell language change to the default language of the kernel.

DonJayamanne:
> enable the .NET team's experience to work here

I don't think so, of things don't work i would expect then to file an issue

DonJayamanne:
Upstream issue https://github.com/microsoft/vscode/issues/111534

joyceerhl:
Validated with Bash kernel: 
![image](https://user-images.githubusercontent.com/30305945/101232344-131a8000-3666-11eb-96dc-ec3896d9b0cd.png)

PR #3934, opened 2020-11-16
Capture the language of the notebook/kernel used

DonJayamanne:
For #398

### DIFF
--- a/src/client/datascience/constants.ts
+++ b/src/client/datascience/constants.ts
@@ -18,6 +18,14 @@ export const KernelLauncherDaemonModule = 'vscode_datascience_helpers.kernel_lau
 export const PythonExtension = 'ms-python.python';
 
 // List of 'language' names that we know about. All should be lower case as that's how we compare.
+export const KnownKernelLanguageAliases = new Map<string, string>([
+    ['qsharp', 'q#'],
+    ['csharp', 'c#'],
+    ['fsharp', 'f#'],
+    ['c++11', 'c++'],
+    ['c++12', 'c++'],
+    ['c++14', 'c++']
+]);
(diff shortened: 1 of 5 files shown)


src/client/datascience/constants.ts
```diff
@@ -18,18 +18,31 @@ export const KernelLauncherDaemonModule = 'vscode_datascience_helpers.kernel_lau
 export const PythonExtension = 'ms-python.python';
 
 // List of 'language' names that we know about. All should be lower case as that's how we compare.
+export const KnownKernelLanguageAliases = new Map<string, string>([
+    ['qsharp', 'q#'],
+    ['csharp', 'c#'],
```
DonJayamanne, line 23: On my mac, `qsharp` is the langauge of the kernel.
Ie `q#` is not using the same pattern as `c$` & `f#`.
Added support for both, if there are such kernels.
With C++, we have xeus c++ using the version (added support for 11, 12 & 14).
Can't remember if there's a 12. Eitehr way doesn't matter.

src/client/datascience/constants.ts
```diff
@@ -18,18 +18,31 @@ export const KernelLauncherDaemonModule = 'vscode_datascience_helpers.kernel_lau
...
 export const KnownNotebookLanguages: string[] = [
     'python',
     'r',
     'julia',
     'c++',
     'c#',
     'f#',
+    'q#',
+    'powershell',
+    'java',
```
DonJayamanne, line 38: Added missing `rust, java & powershell`.
Recently we assumed that no one uses java, but we didn't track it.

rchiodo (approved):

Issue #3927, opened 2020-11-16
Kernel interrupt not working when interrupt mode is message (Julia on windows)


DonJayamanne:
Our assumption is that kernel interrupts work with signals.
However kernels have the ability to get interrupted via messages.
This information is sent in the kernel info message.

**Based on this we should use signals or messages.**

Julia kernel on Windows (not mac) uses messages for interrupting (I believe Rust is the same on Mac & Windows)

DonJayamanne:
Fixed part of https://github.com/microsoft/vscode-jupyter/pull/8125

PR #3951, opened 2020-11-17
Add interrupt_mode & getSpec

DonJayamanne:
Part of #3927 
This is something that came out of the long running kernels work.
It does'nt fix anything, but adds stuff we will need eventually.

### DIFF
--- a/src/client/datascience/jupyter/kernels/jupyterKernelSpec.ts
+++ b/src/client/datascience/jupyter/kernels/jupyterKernelSpec.ts
@@ -20,6 +20,7 @@ export class JupyterKernelSpec implements IJupyterKernelSpec {
     public readonly env: NodeJS.ProcessEnv | undefined;
     public display_name: string;
     public argv: string[];
+    public interrupt_mode?: 'message' | 'signal';
 
     // tslint:disable-next-line: no-any
     public metadata?: Record<string, any> & { interpreter?: Partial<PythonEnvironment> };
@@ -36,6 +37,8 @@ export class JupyterKernelSpec implements IJupyterKernelSpec {
         this.metadata = specModel.metadata;
         // tslint:disable-next-line: no-any
         this.env = specModel.env as any; // JSONObject, but should match
+        // tslint:disable-next-line: no-any
+        this.interrupt_mode = specModel.interrupt_mode as any;
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/client/datascience/jupyter/kernels/jupyterKernelSpec.ts
```diff
@@ -34,6 +35,8 @@ export class JupyterKernelSpec implements IJupyterKernelSpec {
         this.metadata = specModel.metadata;
         // tslint:disable-next-line: no-any
         this.env = specModel.env as any; // JSONObject, but should match
+        // tslint:disable-next-line: no-any
+        this.interrupt_mode = specModel.interrupt_mode as any;
```
DonJayamanne, line 39: This is part of the spec. If interrupt_mode = 'message' we should send a message to the kernel to interrupt. Rust is one such kernel.
**Not fixing this, just exposing this information so we can use it when fixing other realted issues**

src/client/datascience/raw-kernel/rawKernel.ts
```diff
@@ -111,7 +113,16 @@ export class RawKernel implements Kernel.IKernel {
...
+        if (this.kernelProcess.kernelConnectionMetadata.kind === 'startUsingKernelSpec') {
+            const kernelSpec = cloneDeep(this.kernelProcess.kernelConnectionMetadata.kernelSpec) as any;
+            const resources = 'resources' in kernelSpec ? kernelSpec.resources : {};
+            return {
+                ...kernelSpec,
+                resources
+            };
+        }
+        traceError('Fetching kernel spec from raw kernel using JLab API');
         return this.realKernel.getSpec();
```
DonJayamanne, line 126: This API doesn't work, `realKernel.getSpec` goes to Jupyter Lab API and that attempts to make a REST call to the base (dummy url).
This should not get executed hence added error logging.

src/client/datascience/raw-kernel/rawKernel.ts
```diff
@@ -111,7 +113,16 @@ export class RawKernel implements Kernel.IKernel {
         await this.kernelProcess.dispose();
         this.socket.dispose();
     }
-    public getSpec(): Promise<Kernel.ISpecModel> {
+    public async getSpec(): Promise<Kernel.ISpecModel> {
+        if (this.kernelProcess.kernelConnectionMetadata.kind === 'startUsingKernelSpec') {
+            const kernelSpec = cloneDeep(this.kernelProcess.kernelConnectionMetadata.kernelSpec) as any;
+            const resources = 'resources' in kernelSpec ? kernelSpec.resources : {};
```
DonJayamanne, line 119: When using raw kernels, we have the kernel spec, hence return that.
Not used today, but will be used in long running kernels, noop change.

rchiodo (approved):

DavidKutu (approved):

Issue #1373, opened 2020-10-30
"__file__" variable not available in IPython after kernel restart


alexbogun:
## Environment data

-   VS Code version: 1.50.1
-   Extension version (available under the Extensions sidebar): v2020.10.332292344
-   OS and version: Win 10
-   Python version (& distribution if applicable, e.g. Anaconda): 3.8
-   Type of virtual environment used conda
-   Value of the `python.languageServer` setting: Jedi

## Expected behaviour

`__file__` variable is always defined in IPython

## Actual behaviour

`__file__` variable is available when IPython first starts, but is not available after kernel restart.

Reverting to version v2020.9.112786 fixes the problem

## Steps to reproduce:

1. open VSCode
2. create new Python script (.py)
3. type `print(__file__)`
4. Run Selection/Current line in Python Interactive
5. Restart Jupyter kernel
6. Run Selection/Current line in Python Interactive again
</details>

IanMatthewHuff:
Validated.
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/293ac1fe-76f9-436d-b063-aa70a69e8c97)

alexbogun:
Unfortunately the issue is back again, could we please reopen this?

rchiodo:
I'm reproing too.

IanMatthewHuff:
@rchiodo I don't believe this was in the point fix. Yeah, I don't see it in the release branch. I think this is fixed, but just not live for customers yet.

rchiodo:
Ah okay. Thanks. @alexbogun this will be fixed in our next stable release.

IanMatthewHuff:
Validated via the previous validation. Will be in the next release as mentioned.

Issue #446, opened 2020-11-09
Input Problems with Interactive Terminal


tsoud:
Issue Type: <b>Bug</b>

Typing directly into the Python interactive terminal is buggy. This seems to be an issue in the latest (October 2020) release since it began immediately after the update.

Expected result: You type something into the interactive terminal input cell and characters appear...

Actual result: You type into the interactive terminal and nothing happens. Usually, you can start typing after clicking the cursor a few times but many times the terminal input cell does not respond at all requiring a restart.

Extension version: 2020.10.332292344
VS Code version: Code 1.51.0 (fcac248b077b55bae4ba5bab613fd6e9156c2f0c, 2020-11-05T18:16:10.374Z)
OS version: Linux x64 5.4.0-52-generic

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz (12 x 4000)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: disabled_off<br>opengl: enabled_on<br>protected_video_decode: unavailable_off<br>rasterization: disabled_software<br>skia_renderer: enabled_on<br>video_decode: unavailable_off<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|0, 0, 0|
|Memory (System)|31.20GB (7.46GB free)|
|Process Argv|. --no-sandbox|
|Screen Reader|no|
|VM|0%|
|DESKTOP_SESSION|cinnamon|
|XDG_CURRENT_DESKTOP|X-Cinnamon|
|XDG_SESSION_DESKTOP|cinnamon|
|XDG_SESSION_TYPE|x11|
</details>

ihnorton:
Click-dragging the "Python Interactive" tab and moving it to a new split position (eg side by side or the reverse) seems to fix the problem temporarily :/

rboyes:
Same issue here

```
Version: 1.51.1 (user setup)
Commit: e5a624b788d92b8d34d1392e4c4d9789406efe8f
Date: 2020-11-10T23:34:32.027Z
Electron: 9.3.3
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
[...]

PR #3954, opened 2020-11-18
Fix focus tracking for IW

rchiodo:
For #446, #1373 

Change the way we track focus from VS code activating a window. Our old 'activate' message was interfering with the focus message that VS code sends to the iframe around a webview. 

This new way actually tracks focus changes from VS code by watching focus events on the window. When the focus changes to the body element (the default) then we set the focus to the last known TEXTAREA (monaco editor). This also works for the old webview notebook editor for remembering what cell you were editing.

It should also be noted that it doesn't interfere with tabbing because tabbing doesn't put focus into the body.

I don't see an easy way to write a test for this though. David's old test relied on being able to set focus to a random div on the page and then causing an activate. Since activate isn't handled and a random div won't give focus back to the input box, the test won't work. Additionally Enzyme doesn't support adding event handlers so it wouldn't even know about the TextAreaFocusTracker. 

### DIFF
--- a/src/datascience-ui/interactive-common/textAreaFocusTracker.tsx
+++ b/src/datascience-ui/interactive-common/textAreaFocusTracker.tsx
@@ -19,7 +19,7 @@ export class TextAreaFocusTracker extends React.Component {
 
     private onWindowGotFocus() {
         // When the window receives focus (from outside), on a delay attempt to reset our focus
-        setTimeout(this.setToLastKnown.bind(this), 100);
+        setTimeout(this.setToLastKnown.bind(this), 0);
     }
 
     private onWindowLostFocus() {
--- a/src/datascience-ui/interactive-common/textAreaFocusTracker.tsx
+++ b/src/datascience-ui/interactive-common/textAreaFocusTracker.tsx
@@ -31,6 +31,9 @@ export class TextAreaFocusTracker extends React.Component {
     }
 
     private setToLastKnown() {
+        // Try to set the focus to the last TEXTAREA if we are currently
+        // pointing at the BODY. Only do this for the BODY because the
+        // only thing that sets the focus to the BODY is the VS code focus code.
         if (
             document.activeElement &&
             document.activeElement.nodeName === document.body.nodeName &&
--- a/src/datascience-ui/interactive-common/textAreaFocusTracker.tsx
+++ b/src/datascience-ui/interactive-common/textAreaFocusTracker.tsx
@@ -1,0 +1,42 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+'use strict';
+import * as React from 'react';
+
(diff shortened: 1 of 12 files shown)


src/datascience-ui/interactive-common/textAreaFocusTracker.tsx
```diff
@@ -0,0 +1,42 @@
...
+        // When the window loses focus, remember the lost focus as long as it
+        // is a text area
+        if (document.activeElement && document.activeElement.nodeName === 'TEXTAREA') {
+            this.lastFocusedTextArea = document.activeElement as HTMLElement;
+        }
+    }
+
+    private setToLastKnown() {
+        if (
+            document.activeElement &&
```
rchiodo, line 35: Maybe I should add a comment here, but this is key. The window got focus sets a timer that will fire and will only change the focus if the current focus is the "BODY" element. Otherwise it can steal focus from other objects. The only thing that sets the focus to the body element is the generic focus call that VS code does.

rchiodo: Added a comment.

src/datascience-ui/interactive-common/textAreaFocusTracker.tsx
```diff
@@ -0,0 +1,45 @@
...
+        const hiddenStyle: React.CSSProperties = {
+            display: 'none'
+        };
+
+        return <div id="focus_tracker" style={hiddenStyle} />;
+    }
+
+    private onWindowGotFocus() {
+        // When the window receives focus (from outside), on a delay attempt to reset our focus
+        setTimeout(this.setToLastKnown.bind(this), 100);
```
IanMatthewHuff, line 22: So this mostly seems ok. But the timeout still seems a bit dangerous to me. Would this possibly rip focus back if you set focus elsewhere during the timeout? 100ms is unlikely, but could be possible. Would it make sense to cancel the timeout in onWindowLostFocus? I think that we would not want to run it if focus had been removed in that time.

rchiodo: Cancel sounds good. I don't think 100ms is necessary either. I can make it zero.

src/datascience-ui/interactive-common/textAreaFocusTracker.tsx
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    private onWindowLostFocus() {
+        // When the window loses focus, remember the lost focus as long as it
+        // is a text area
+        if (document.activeElement && document.activeElement.nodeName === 'TEXTAREA') {
```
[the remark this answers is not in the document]
rchiodo: Yes. That's the behavior. It's what we did before.

IanMatthewHuff (approved):
With the change to cancel the timeout (or make it zero) I'm down. I'd like to poke at it some after it's in though.

PR #3964, opened 2020-11-18
Change cell lang when changing kernel for new nbs

DonJayamanne:
For #398

### DIFF
--- a/src/client/datascience/common.ts
+++ b/src/client/datascience/common.ts
@@ -110,17 +110,19 @@ export function traceCellResults(prefix: string, results: ICell[]) {
     }
 }
 
+const jupyterLanguageToMonacoLanguageMapping = new Map([
+    ['c#', 'csharp'],
+    ['f#', 'fsharp'],
+    ['q#', 'qsharp'],
+    ['c++11', 'c++'],
+    ['c++12', 'c++'],
+    ['c++14', 'c++']
+]);
 export function translateKernelLanguageToMonaco(kernelLanguage: string): string {
     // At the moment these are the only translations.
     // python, julia, r, javascript, powershell, etc can be left as is.
-    switch (kernelLanguage.toLowerCase()) {
-        case 'c#':
-            return 'csharp';
-        case 'f#':
-            return 'fsharp';
-        default:
-            return kernelLanguage.toLowerCase();
-    }
+    kernelLanguage = kernelLanguage.toLowerCase();
+    return jupyterLanguageToMonacoLanguageMapping.get(kernelLanguage) || kernelLanguage;
 }
 
 export function generateNewNotebookUri(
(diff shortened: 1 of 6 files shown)


src/client/datascience/notebook/helpers/helpers.ts
```diff
@@ -57,6 +59,13 @@ export function isJupyterNotebook(option: NotebookDocument | string) {
     }
 }
 
+export function isJupyterKernel(kernel?: VSCNotebookKernel): kernel is VSCodeNotebookKernelMetadata {
```
DonJayamanne, line 62: I think this is cleaner than writing this code everywhere.
@IanMatthewHuff (made me realize in our conversation that this could be improved by moving into a simple function).

src/client/datascience/notebook/kernelProvider.ts
```diff
@@ -313,7 +314,7 @@ export class VSCodeKernelPickerProvider implements INotebookKernelProvider {
         kernel: VSCNotebookKernel | undefined;
     }) {
         // We're only interested in our Jupyter Notebooks & our kernels.
-        if (!kernel || !(kernel instanceof VSCodeNotebookKernelMetadata) || !isJupyterNotebook(document)) {
+        if (!isJupyterKernel(kernel) || !isJupyterNotebook(document)) {
```
DonJayamanne, line 317: Much better

src/client/datascience/notebook/helpers/helpers.ts
DonJayamanne: This is an unrelated change in the PR

rchiodo (approved):

src/client/datascience/common.ts
```diff
@@ -110,17 +110,19 @@ export function traceCellResults(prefix: string, results: ICell[]) {
     }
 }
 
+const jupyterLanguageToMonacoLanguageMapping = new Map([
+    ['c#', 'csharp'],
+    ['f#', 'fsharp'],
+    ['q#', 'qsharp'],
+    ['c++11', 'c++'],
+    ['c++12', 'c++'],
+    ['c++14', 'c++']
```
DavidKutu, line 119:
```suggestion
    ['c++14', 'c++'],
    ['R', 'r'],
    ['Rust', 'rust'],
    ['julia', 'julia'],
    ['powershell', 'powershell'],
    ['python', 'python']
```

I was already doing this in a namespace. But now that you have it here, please add these languages so we can always call this.
Also, I'm not sure 'julia' is correct for Monaco, can you check please?

DavidKutu (approved):

IanMatthewHuff (approved):

src/client/datascience/common.ts
DonJayamanne: Not sure we need `, rust, julia, powershell & python`. They are identitcal.
If you look at the code, we just lowercase them. Hence not necessary to add to the dict, as we'll get the same behavior.

DavidKutu: We didn't need them for the code to work, but its good to explicitly mention the languages we support in code.
A case might happen where the monaco name is capitalized and the kernel name is not, we'd need to add that case, in contrast to cases like r, or rust. I'd rather have them all listed.

Issue #1363, opened 2020-09-28
Export uses the wrong interpreter to perform the export


rchiodo:
It defaults to using the 'jupyter interpreter' instead of using the user's interpreter. This makes it impossible for the user to diagnose errors as the errors don't match what their current interpreter might report.

See this issue:
microsoft/vscode-jupyter#710

greazer:
Keeping this as must-fix since #710 is affecting many users. We closed that bug in the python repo prior to opening the jupyter repo.

IanMatthewHuff:
@greazer Hang on a second. I think that I already fixed this. Did this maybe get moved over from a completed pipeline?

IanMatthewHuff:
Yeah, this was fixed.

It was this PR here in python: https://github.com/microsoft/vscode-python/pull/14308
And this one in Jupyter: https://github.com/microsoft/vscode-jupyter/pull/181

rchiodo:
This doesn't seem to be fixed?

Repro steps:

1) Open notebook
2) Switch kernel to new env
3) Install nbconvert in new env
4) Run export

It uses the selected interpreter instead?

Looking at the callstack, I'm not sure how it would ever get passed a candidate interpreter.

IanMatthewHuff:
@rchiodo  are you using the command from the palette Jupyter: Export to HTML? Realized that I fixed this working with the ExportAs button, which does pass and use a candidate interpreter. Pretty sure that I didn't look at the palette commands. Will probably have to make an additional fix for them.

rchiodo:
I used the export button at the top of a native notebook.

joyceerhl:
Validated.

PR #3965, opened 2020-11-18
Fix export commands and set notebook property on native editors

IanMatthewHuff:
For #1363

### DIFF
--- a/package.json
+++ b/package.json
@@ -1187,6 +1187,12 @@
                     "when": "false"
                 },
                 {
+                    "command": "jupyter.export",
+                    "title": "%DataScience.notebookExportAs%",
+                    "category": "Jupyter",
+                    "when": "false"
+                },
+                {
                     "command": "jupyter.notebookeditor.export",
                     "title": "%DataScience.notebookExportAs%",
                     "category": "Jupyter",
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
         "onCommand:jupyter.exportAsPythonScript",
         "onCommand:jupyter.exportToHTML",
         "onCommand:jupyter.exportToPDF",
+        "onCommand:jupyter.notebookeditor.export",
         "onCommand:jupyter.switchOffInsidersChannel",
         "onCommand:jupyter.switchToDailyChannel",
         "onCommand:jupyter.switchToWeeklyChannel",
@@ -273,7 +274,7 @@
                 "category": "Jupyter"
             },
             {
-                "command": "jupyter.export",
+                "command": "jupyter.notebookeditor.export",
                 "title": "%DataScience.notebookExportAs%",
                 "category": "Jupyter",
                 "icon": {
@@ -281,6 +282,11 @@
                     "dark": "resources/dark/export_to_python.svg"
                 },
                 "enablement": "notebookViewType == jupyter-notebook && jupyter.isnotebooktrusted"
+            },
+            {
+                "command": "jupyter.export",
+                "title": "%DataScience.notebookExportAs%",
+                "category": "Jupyter"
             },
             {
                 "command": "jupyter.exportAsPythonScript",
@@ -744,7 +750,7 @@
                     "when": "resourceLangId == jupyter && notebookViewType == 'jupyter-notebook' && !jupyter.isnotebooktrusted && jupyter.trustfeatureenabled"
                 },
                 {
-                    "command": "jupyter.export",
+                    "command": "jupyter.notebookeditor.export",
                     "title": "%DataScience.notebookExportAs%",
                     "group": "navigation",
                     "when": "resourceLangId == jupyter && notebookViewType == 'jupyter-notebook' && jupyter.isnotebooktrusted"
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -53,6 +53,7 @@
         "onCommand:jupyter.exportAsPythonScript",
         "onCommand:jupyter.exportToHTML",
         "onCommand:jupyter.exportToPDF",
+        "onCommand:jupyter.notebookeditor.export",
```
IanMatthewHuff, line 56: Note on why I added the new command in the notebookCommands code file.

src/client/common/application/commands.ts
```diff
@@ -125,10 +125,11 @@ export interface ICommandNameArgumentTypeMapping extends ICommandNameWithoutArgu
     [DSCommands.GotoPrevCellInFile]: [];
     [DSCommands.ScrollToCell]: [Uri, string];
     [DSCommands.ViewJupyterOutput]: [];
-    [DSCommands.ExportAsPythonScript]: [INotebookModel, PythonEnvironment | undefined];
-    [DSCommands.ExportToHTML]: [INotebookModel, string | undefined, PythonEnvironment | undefined];
-    [DSCommands.ExportToPDF]: [INotebookModel, string | undefined, PythonEnvironment | undefined];
-    [DSCommands.Export]: [Uri | INotebookModel, string | undefined, PythonEnvironment | undefined];
+    [DSCommands.ExportAsPythonScript]: [INotebookModel | undefined, PythonEnvironment | undefined];
```
IanMatthewHuff, line 128: Correcting types. undefined is allowed for any of these when calling from the command palette.

src/client/datascience/commands/exportCommands.ts
```diff
@@ -63,23 +63,24 @@ export class ExportCommands implements IDisposable {
...
+    }
+
     private async export(
-        modelOrUri: Uri | INotebookModel,
+        model?: INotebookModel,
         exportMethod?: ExportFormat,
         defaultFileName?: string,
         interpreter?: PythonEnvironment
     ) {
-        defaultFileName = typeof defaultFileName === 'string' ? defaultFileName : undefined;
```
IanMatthewHuff, line 72: This export command was hooked up to the editor/title contribution point for native notebooks and shares a signature with the existing export command that the interactive window / old editor still use. However there were some type issues here. The editor/title command populates both a Uri and an object with editor placement info in it. So it was passing something like this:
(URI, { groupId: 1 })

so that groupId object was getting passed in as the defaultFileName optional parameter. Which I think is why this line was here to remove it.

This seemed like an subversion of the type system and the URI | INotebookModel didn't make the most sense as an overload (the two are not related URI is just used to convert to a model. So I added the new nativenotebook export command to avoid the incorrectly typed parameter and the modelOrUri checking code. That code does the correct conversion and calls into export.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -223,6 +225,13 @@ export class Kernel implements IKernel {
         if (!this.notebook) {
             return;
         }
+
+        // Set the notebook property on the matching editor
+        const editor = this.editorProvider.editors.find((item) => this.fs.arePathsSame(item.file, this.uri));
```
IanMatthewHuff, line 230: I needed this property as I was looking at editor.notebook?.getMatchingInterpreter to get the candidate interpreter to use. However editor.notebook was never getting set for native notebooks. It was always undefined. 

As a side effect this code here wasn't actually doing anything before (as notebook was always undefined):
https://github.com/microsoft/vscode-jupyter/blob/cff6406f7d942b13fdc8dd16688ff9c376aa43da/src/client/datascience/notebook/notebookEditorProvider.ts#L133

So this is the side effect that I mentioned at the meeting that native notebooks will actually dispose their notebooks now when closed. Before they would only dispose on vscode close.

src/client/datascience/commands/exportCommands.ts
```diff
@@ -89,6 +90,11 @@ export class ExportCommands implements IDisposable {
             }
             model = activeEditor.model;
 
+            // At this point also see if the active editor has a candidate interpreter to use
+            if (!interpreter) {
+                interpreter = activeEditor.notebook?.getMatchingInterpreter();
```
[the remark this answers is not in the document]
rchiodo: Actually that seems like reversed. What about for native notebooks?

rchiodo (approved):

src/client/datascience/commands/exportCommands.ts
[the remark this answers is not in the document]
IanMatthewHuff: .notebook is a property which is shared by old notebook editors and new native notebook editors. In the fix I made previously I already called getMatchingInterpreter in the webview editor case when triggered from the button like so:

https://github.com/microsoft/vscode-jupyter/blob/cff6406f7d942b13fdc8dd16688ff9c376aa43da/src/client/datascience/interactive-ipynb/nativeEditor.ts#L736

This code here is just if nothing is passed into export, which would imply execution from the command palette. In this case both new native and old webview can use this same path to find the active editor and get an associated interpreter if available.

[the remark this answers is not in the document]
IanMatthewHuff: I went and triple checked quick and both native and old worked via the "no model" code path. However I did see an odd error doing export to HTML via native notebook. Going to look into that quick before I finish up this PR.

DonJayamanne (approved):

Issue #365, opened 2020-10-30
Review kernel search & startup environment variables (support for non-python kernels with Conda environments)


DonJayamanne:
**Issues when starting non-python kernels**
### Starting Java kernel registered in a conda environment (like BeakerX) will never work as the paths are incorrect. We need to activate conda environment and use those env vars...
See other comments.

**Context:**
https://github.com/microsoft/vscode-jupyter/issues/361
https://github.com/microsoft/vscode-jupyter/pull/362

@IanMatthewHuff 

This doesn't seem right as there is a specific search order for jupyter kernel specs. You can have the same name without Conda involved at all. And if you didn't have conda active you wouldn't want a conda path to be searched at all.

https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs
https://github.com/jupyter/jupyter_client/blob/540c996396ad60c312016fecc48200b5452ef1f2/jupyter_client/kernelspec.py

I think we'd want the following:
1. If conda wasn't active you'd use the Jupyter search priority order.
2. If conda was active (only if python extension is installed) you'd want to prioritize the active conda env (and no other ones) then the jupyter ones in priority order.

I get that's a bigger change, but that seems right to me. Could file it as a work item for after parity release?

@DonJayamanne 

> This doesn't seem right as there is a specific search order for jupyter kernel specs. You can have the same name without Conda involved at all. And if you didn't have conda active you wouldn't want a conda path to be searched at all.

Agreed. This PR ensures that if we have duplicates we at least get that working.
Would you agree this PR is sufficient or does this break anything. 

> I think we'd want the following:
> If conda wasn't active you'd use the Jupyter search priority order.

Also, we need to run the kernels with the conda environment variables.
E.g. when running a JavaKernel installed by a conda environment, then we'd need to activate the conda environment to run the Java kernel. (as java executable & the like is in Conda environment) & all of these paths are setup when activating the conda environment.

IanMatthewHuff:
Validated via our testing on this.

PR #3968, opened 2020-11-19
Track the interpreter that owns kernelspec in kernelspec info

DonJayamanne:
Part of #365
With Jupyter kernels in Conda environments, we need to inherit the environment variabels of the conda environment.
This means, if we start a kernel, we need to first chekc if it belongs to a conda or python environment & then inherit the variables from that (for now, plan is to do this just for conda environments).

Steps:
* User select a kernel
* Before we start a kernel, check if it belongs to conda env
* Get corresponding conda env that owns this kernel
* Get env variabels of that conda env
* When starting the kernel, use those env variables instead of the current `process.env`.

This PR merely allows us to identify easily the python environment that a kernel belongs to.
* Not a fan of adding `interpreterPath` to kernel Spec, but this makes it easier than querying this information all over again.

Optionally:
* I could add a method to kernel Finder that returns the corresopnding interpreter
* Behind the scenes (private variable) i'd hold this mapping in some dict
* I'm ok with either change or alternatives
I.e `KernelFinder.getOwningInterpreter(kernelSpecName: string): Promoise<PythonEnvironment | undefined>`

### DIFF
--- a/src/client/datascience/kernel-launcher/kernelFinder.ts
+++ b/src/client/datascience/kernel-launcher/kernelFinder.ts
@@ -438,7 +438,7 @@ export class KernelFinder implements IKernelFinder {
         if (
             !this.cache.find(
                 (item) =>
-                    item.interpreterPath === newPath.interpreterPath && item.kernelSpecFile === item.kernelSpecFile
+                    item.interpreterPath === newPath.interpreterPath && item.kernelSpecFile === newPath.kernelSpecFile
             )
         ) {
             this.cache.push(newPath);
--- a/src/client/datascience/kernel-launcher/kernelFinder.ts
+++ b/src/client/datascience/kernel-launcher/kernelFinder.ts
@@ -21,8 +21,6 @@ import { Telemetry } from '../constants';
 import { JupyterKernelSpec } from '../jupyter/kernels/jupyterKernelSpec';
 import { IJupyterKernelSpec } from '../types';
 import { IKernelFinder } from './types';
-// tslint:disable-next-line:no-require-imports no-var-requires
-const flatten = require('lodash/flatten') as typeof import('lodash/flatten');
 
 const winJupyterPath = path.join('AppData', 'Roaming', 'jupyter', 'kernels');
 const linuxJupyterPath = path.join('.local', 'share', 'jupyter', 'kernels');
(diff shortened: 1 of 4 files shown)


src/client/datascience/types.ts
```diff
@@ -420,6 +420,18 @@ export interface IJupyterKernelSpec {
...
+     */
+    specFile?: string;
+    /**
+     * Optionally the Interpreter this kernel spec belongs to.
+     * You can have kernel specs that are scoped to an interpreter.
+     * E.g. if you have Python in `c:\Python\Python3.8`
+     * Then you could have kernels in `<sys.prefix folder for this interpreter>\share\jupyter\kernels`
+     * Plenty of conda packages ship kernels in this manner (beakerx, etc).
+     */
+    interpreterPath?: string;
```
DonJayamanne, line 434: This is what feels iffy...

src/client/datascience/types.ts
```diff
@@ -420,6 +420,18 @@ export interface IJupyterKernelSpec {
     // tslint:disable-next-line: no-any
     readonly metadata?: Record<string, any> & { interpreter?: Partial<PythonEnvironment> };
     readonly argv: string[];
+    /**
+     * Optionally where this kernel spec json is located on the local FS.
+     */
+    specFile?: string;
```
DonJayamanne, line 426: we already have this info, and using this in launcher.

DonJayamanne:
TODO: Need to fix tests, but would like a review first, whether this design/api is good enough or we need to change it to have an explicit method in kernelFinder to get the owning interpreter.

rchiodo (approved):

src/client/datascience/kernel-launcher/kernelFinder.ts
```diff
@@ -394,16 +426,21 @@ export class KernelFinder implements IKernelFinder {
                 return;
             }
             this.cache = JSON.parse(
-                await this.fs.readLocalFile(path.join(this.context.globalStoragePath, cacheFile))
-            ) as string[];
+                await this.fs.readLocalFile(path.join(this.context.globalStorageUri.fsPath, cacheFile))
```
IanMatthewHuff, line 429: When you change the cache type, will this break existing customers here?

IanMatthewHuff: Actually looks like it might just generate a new one, that's probably fine.

src/client/datascience/types.ts
IanMatthewHuff: It's a bit funky that the interpreter concept is creeping into more areas, especially now that we are factoring out more python stuff. But as you noted, we're doing all this searching now, so it is the time to log and save this information. We'd have to save this mapping somehow, not sure that getOwningInterpreter would really be any cleaner or easier to deal with.

IanMatthewHuff (approved):

PR #4000, opened 2020-11-22
Replace ffi-napi and ref-napi with BCryptGenRandom executable

joyceerhl:
~This should probably only be merged after I've gotten codesigning set up for the executable, although I've uploaded an unsigned copy to the blob storage for now.~
I've gotten codesigning set up for the executable now so this is ready for review.


joyceerhl:
Sorry, forgot that the exe needs to be compiled for ARM, so this PR is not ready for review yet. (Although you're free to comment on other aspects while I address the compilation issue).

build/ci/postInstall.js
```diff
@@ -76,5 +77,23 @@ function createJupyterKernelWithoutSerialization() {
...
+/**
+ * In order to generate random bytes on Windows without taking a dependency on native node modules
+ * which we then need to build xplat and bundle with the extension, download a prebuilt executable
+ * which directly consumes BCryptGenRandom in bcrypt.dll and outputs random bytes as hex. This
+ * executable is required for trusted notebooks key generation and is included with the built extension.
+ */
+function downloadBCryptGenRandomExecutable() {
+    console.log('Downloading BCryptGenRandom.exe...');
+    const executableName = 'BCryptGenRandom.exe';
+    const uri = `https://pvsc.blob.core.windows.net/extension-builds-juypter/${executableName}`;
```
rchiodo, line 89: Hmm, this would likely make me nervous if I was outside of Microsoft. I'm trusting this executable to be downloaded and run on my machine. How do I know this blob storage site is secure? What if somebody hacks our blob storage?

Trying to think of a better solution but not having much luck at the moment.

rchiodo: I wonder if we could publish the exe to the vscode marketplace?

rchiodo: As a separate extension?

rchiodo: Oh wait never mind. This isn't happening at run time. It's happening during build time. That seems fine to me.

rchiodo (approved):

build/ci/postInstall.js
```diff
@@ -76,5 +77,23 @@ function createJupyterKernelWithoutSerialization() {
...
+    const dest = path.resolve(path.dirname(__dirname), '..', 'src', 'BCryptGenRandom');
+    fs.ensureDirSync(dest);
+    const destination = path.join(dest, executableName);
+    cp.execSync(`curl --output ${destination} ${uri}`);
```
DonJayamanne, line 93: Not everyone has `curl`, why not just use node API to download a file or use an npm package like `download` already used today.

package.json
```diff
@@ -1745,7 +1745,7 @@
     },
     "scripts": {
         "package": "gulp clean && gulp prePublishBundle && vsce package -o ms-toolsai-jupyter-insiders.vsix",
-        "compile": "tsc -watch -p ./",
+        "compile": "gulp includeBCryptGenRandomExe && tsc -watch -p ./",
```
DonJayamanne, line 1748: Instead of this, it would be better to do this as part of a `postInstall` step.
Thats where we should be pulling down required dependencies.

DonJayamanne: See existing `postInstall.js`

DonJayamanne (changes requested):
Please downoading using `download` or node API instead of `curl`
Please move gulp step into `postInstall`.

package.json
joyceerhl: Yeah I have a postinstall step to download the executable into the src dir. THis is copying the executable to the outdir. I guess I could just copy the executable to the outdir as part of the postinstall?

src/client/datascience/interactive-ipynb/randomBytes.ts
```diff
@@ -24,60 +28,49 @@ export class SystemPseudoRandomNumberGenerator implements ISystemPseudoRandomNum
...
-            traceInfo('Initializing FFI bindings for BCryptGenRandom...');
-            const bcryptlib = ffi.Library('BCrypt', {
-                // Name of DLL function: [ return type, [ arg1 type, arg2 type, ... ] ]
-                // https://docs.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom
-                BCryptGenRandom: [NTSTATUS, [BCRYPT_ALG_HANDLE, PUCHAR, ULONG, ULONG]]
+    // Run a small bundled executable which directly calls BCryptGenRandom and
+    // outputs 1024 random bytes to stdout as a hex string.
+    private async randomBytesForWindows(_numBytes: number): Promise<string> {
+        // Ensure the exe is present. If it's not we can't generate bytes for Windows
+        const executable = path.resolve(EXTENSION_ROOT_DIR, 'out', 'BCryptGenRandom', 'BCryptGenRandom.exe');
```
rchiodo, line 35: Is there any way to check signing? Or does this happen automatically on windows?

rchiodo (approved):

src/client/datascience/interactive-ipynb/randomBytes.ts
joyceerhl: Are you asking if we can verify that the executable is signed before running it?

build/ci/postInstall.js
```diff
@@ -76,5 +77,38 @@ function createJupyterKernelWithoutSerialization() {
...
+async function downloadBCryptGenRandomExecutable() {
+    console.log('Downloading BCryptGenRandom.exe...');
+    const executableName = 'BCryptGenRandom.exe';
+    const uri = `https://pvsc.blob.core.windows.net/jupyter-dev-builds/${executableName}`;
+    const srcDestination = path.resolve(path.dirname(__dirname), '..', 'src', 'BCryptGenRandom');
+    const srcDestinationFilename = path.join(srcDestination, executableName);
+    if (fs.existsSync(srcDestinationFilename)) {
+        console.log('BCryptGenRandom.exe is already downloaded.');
+    } else {
+        fs.ensureDirSync(srcDestination);
```
DonJayamanne, line 95: No need to download if `outDestinationFileName` already exists.
I think we should check if that file exists first.

build/ci/postInstall.js
```diff
@@ -76,5 +77,38 @@ function createJupyterKernelWithoutSerialization() {
...
+        await download(uri, srcDestination, { filename: executableName });
```
DonJayamanne, line 96: Why not download the file into the out destination instead of downloading & then copying.
Seems unnecessary to have two steps.

src/client/datascience/interactive-ipynb/digestStorage.ts
```diff
@@ -102,15 +102,16 @@ export class DigestStorage implements IDigestStorage {
...
             if (key) {
-                traceInfo(`Successfully generated keyfile`);
-                await this.fs.writeLocalFile(defaultKeyFileLocation, key);
-                return key;
+                traceInfo(`Successfully generated keyfile with OS PRNG`);
             } else {
-                traceError('Failed to initialize secret for notebook trust');
+                // Fallback to crypto.randomBytes
+                traceError('Failed to initialize secret for notebook trust with OS PRNG');
+                key = randomBytes(1024).toString('hex');
```
DonJayamanne, line 111: I think we should capture the telemetry here to see how often this fails.
If we don't care, then why not always use `randomBytes`

src/client/datascience/interactive-ipynb/randomBytes.ts
DonJayamanne: Do we need to verify that, we're shipping the binary our selves, hence we trust what we are shipping.
If the files have been tampered with, then end user has bigger problems (machine has already been compromised)

DonJayamanne (approved):
Don't see the need to download & copy, when we can just download into required folder, but not a mandatory change, else all good.

src/client/datascience/interactive-ipynb/digestStorage.ts
joyceerhl: I put the telemetry capture in randomBytes itself. `Telemetry.NativeRandomBytesGenerationFailed`

build/ci/postInstall.js
joyceerhl: This is because `npm run clean` wipes the outdir, so we still need to keep a copy in src in case that happens.

PR #4008, opened 2020-11-23
fix wording for the send selection setting

IanMatthewHuff:
For #

### DIFF
--- a/package.json
+++ b/package.json
@@ -1432,7 +1432,7 @@
                 "jupyter.sendSelectionToInteractiveWindow": {
                     "type": "boolean",
                     "default": false,
-                    "description": "Determines if selected code in a python file will go to the terminal or the Interactive window when hitting shift+enter",
+                    "description": "When pressing shift+enter, send selected code in a Python file to the Jupyter interactive window as opposed to the Python terminal.",
                     "scope": "resource"
                 },
                 "jupyter.showJupyterVariableExplorer": {


package.json
```diff
@@ -1432,7 +1432,7 @@
                 "jupyter.sendSelectionToInteractiveWindow": {
                     "type": "boolean",
                     "default": false,
-                    "description": "Determines if selected code in a python file will go to the terminal or the Interactive window when hitting shift+enter",
+                    "description": "When pressing shift+enter, send selected code in a Python file to the Jupyter interactive window as opposed to the Python terminal.",
```
IanMatthewHuff, line 1435: Just a quick proposed fix here that I ran into when investigating the shift-enter issue. This is a boolean setting, and as such it needs to be clear about what checking the box does versus unchecking it. The old sentence here doesn't indicate what checking the box will do as "Determines" is not indicative of the direction of the boolean. 

New proposed sentence clearly indicates that "checking the setting => interactive window" while "unchecking the setting => terminal" by making the description indicate what the setting will change.

joyceerhl: Technically users can remap the shift+enter keybinding to other commands, but this is still better than what we had before.

joyceerhl (approved):

package.json
IanMatthewHuff: @joyceerhl Yeah, that actually did bug me a bit as well. I wanted to omit the shift-enter part since it's configurable. But at the same time, it's most likely that someone searching the settings might search based on that. With some wavering I just left it in for now. The whole reason we had this setting was specifically because we'd overloaded shift-enter, which has always been a bit icky.

DavidKutu (approved):

Issue #3988, opened 2020-11-20
Unable to open Jupyter Notebooks: Using the Jupyter notebook editor requires the stable version of VS code and the CustomEditor experiment to be enabled..


xcojonny:
## Environment data

-   VS Code version: 1.51.1
-   Extension version (available under the Extensions sidebar): XXX
-   Jupyter Extension version (available under the Extensions sidebar): v2020.11.372831992
-   OS (Windows | Mac | Linux distro) and version: Win10
-   Python and/or Anaconda version: Python 3.8.1
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A 
-   Jupyter server running: Local and Remote 

## Expected behaviour

Clicking on a Jupyter-Notebook opens the notebook.

## Actual behaviour

Clicking on a Jupyter-Notebook opens the Notebook for a second. Then the file is being closed and the following error is displayed:
`Unable to open '[*].ipynb': Using the Jupyter notebook editor requires the stable version of VS code and the CustomEditor experiment to be enabled..`

This behaviour occurred suddenly, without me changing knowingly anything in VSC. The Notebooks can still be opened on different machines. Furthermore creating a new Notebook works. 

## Steps to reproduce:

1. Try to open any jupyter notebook within VSC over the side bar -> Leads to error
2. Opening via terminal (code file.ipynb) -> Leads to error

## Logs

<p>

```
User belongs to experiment group 'jupyterTest'
```

</p>
</details>

rchiodo:
@claudiaregio  this might be a side effect of the custom editor experiment being turned on.

rchiodo:
This shouldn't be happening but might be caused by this setting being in your user settings.json:

```json
    "workbench.editorAssociations": [
        {
            "viewType": "jupyter.notebook.ipynb",
            "filenamePattern": "*.ipynb"
        }
    ],
```

Can you check if that editor association is there?

This would likely cause the bug above.

We think you can workaround this bug by:
1. Deleting this section from your user settings.json 
or
2. Add this additional setting to your settings.json:
```json
    "jupyter.experiments.optInto": [
        "CustomEditor"
    ],
```

greazer:
One other question. @xcojonny, @takotab, do either of you have "Settings Sync" turned on (keep it off, if you've not ever turned it on). 

![image](https://user-images.githubusercontent.com/11685408/99863692-3ae0f280-2b54-11eb-9166-94417a551cad.png)

[...]

PR #4018, opened 2020-11-24
always remove our editor association if disabled

IanMatthewHuff:
For #3988 

### DIFF
--- a/src/client/common/application/customEditorService.ts
+++ b/src/client/common/application/customEditorService.ts
@@ -6,11 +6,9 @@ import * as vscode from 'vscode';
 
 import { UseCustomEditorApi } from '../constants';
 import { traceError } from '../logger';
-import { IExtensionContext } from '../types';
 import { InvalidCustomEditor } from './invalidCustomEditor';
 import { CustomEditorProvider, ICommandManager, ICustomEditorService, IWorkspaceService } from './types';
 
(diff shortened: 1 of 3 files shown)


src/client/common/application/customEditorService.ts
```diff
@@ -65,24 +62,17 @@ export class CustomEditorService implements ICustomEditorService {
                 viewType: ViewType,
                 filenamePattern: '*.ipynb'
             });
-            await Promise.all([
-                this.extensionContext.globalState.update(EditorAssociationUpdatedKey, true),
-                settings.update('editorAssociations', editorAssociations, vscode.ConfigurationTarget.Global)
-            ]);
+            await settings.update('editorAssociations', editorAssociations, vscode.ConfigurationTarget.Global);
```
IanMatthewHuff, line 65: This is a bit tricky, as we don't have an exact lockdown from the associated customer issue. However this change should prevent getting incorrect editor associations if you copy around settings files.

This seems like the right fix regardless as the settings file isn't scoped the same as the extension global state so one can't be used to guard the other.

DavidKutu (approved):

src/client/common/application/customEditorService.ts
DonJayamanne: Makes perfect sense.

DonJayamanne (approved):

Issue #4002, opened 2020-11-22
Endless product feedback request pop-ups!


r11-moto:
Issue Type: <b>Bug</b>

I keep getting asked to give feedback on jupyter notebooks! I've already given it. But every time I run a cell, I restart my environment, or randomly -- I get prompted to give my feedback.

Extension version: 2020.11.372831992
VS Code version: Code 1.51.1 (e5a624b788d92b8d34d1392e4c4d9789406efe8f, 2020-11-11T01:11:34.018Z)
OS version: Darwin x64 20.1.0

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|VirtualApple @ 2.50GHz processor (8 x 2400)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>metal: disabled_off<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>protected_video_decode: unavailable_off<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|2, 2, 2|
|Memory (System)|16.00GB (0.15GB free)|
|Process Argv|--crash-reporter-id e987354c-7428-47d8-b68a-ac5a78bcde9c|
|Screen Reader|no|
|VM|0%|
</details>

joyceerhl:
Hey @r11-moto sorry about the trouble here. Would you mind providing the following details:
- Are you on the insider or stable build of VS Code?
- What does the text on the popup say? Is it the same popup each time?
- If you click 'no' on the popup, does it continue to appear?

searri:
I am also having this issue!

- I am on the stable build
- The popup comes up **anytime** I interact with the notebook, as @r11-moto said. It always says "Can you please take 2 minutes to tell us how the Python Data Science features are working for you?"
- I click "No, thanks" every time and it still appears, regardless of VSCode or system restart

The first time it came up, I *did* fill out the survey, and I tried clicking "Yes, take survey now" (to see if that would suppress the prompt), but SurveyMonkey says I've already completed it. There appears to be no way to suppress this in the Extension Settings.

DonJayamanne:
@r11-moto @searri 

Apologies for this issue, I think we have identified the cause of this issue.

[...]

PR #4019, opened 2020-11-24
Ensure survey prompt is not displayed multiple times

DonJayamanne:
For #4002

### DIFF
--- a/src/client/datascience/dataScienceSurveyBanner.ts
+++ b/src/client/datascience/dataScienceSurveyBanner.ts
@@ -83,9 +83,9 @@ export type ShowBannerWithExpiryTime = { data: boolean; expiry?: number };
 @injectable()
 export class DataScienceSurveyBanner implements IJupyterExtensionBanner {
     public get enabled(): boolean {
-        // if (this.applicationEnvironment.uiKind !== UIKind.Desktop || this.applicationEnvironment.channel !== 'stable') {
-        //     return false;
-        // }
+        if (this.applicationEnvironment.uiKind !== UIKind.Desktop || this.applicationEnvironment.channel !== 'stable') {
+            return false;
+        }
         if (!this.showBannerState.value.expiry) {
             return true;
         }
--- a/src/client/datascience/dataScienceSurveyBanner.ts
+++ b/src/client/datascience/dataScienceSurveyBanner.ts
@@ -4,7 +4,7 @@
 'use strict';
 
 import { inject, injectable, named } from 'inversify';
-import { env, Event, EventEmitter, UIKind } from 'vscode';
+import { Event, EventEmitter, UIKind } from 'vscode';
 import { IApplicationEnvironment, IApplicationShell } from '../common/application/types';
 import '../common/extensions';
 import {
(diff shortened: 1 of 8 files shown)


src/client/datascience/dataScienceSurveyBanner.ts
```diff
@@ -77,21 +80,31 @@ export class DataScienceSurveyBannerLogger implements IInteractiveWindowListener
...
+        if (this.applicationEnvironment.uiKind !== UIKind.Desktop || this.applicationEnvironment.channel !== 'stable') {
+            return false;
+        }
+        if (!this.showBannerState.value.enabled) {
+            return false;
+        }
+        if (!this.showBannerState.value.expiryTimeInMs) {
+            return this.showBannerState.value.enabled;
+        }
+        return this.showBannerState.value.expiryTimeInMs! < Date.now();
```
DonJayamanne, line 99: Explicit check for expiry

src/client/datascience/dataScienceSurveyBanner.ts
```diff
@@ -77,21 +79,25 @@ export class DataScienceSurveyBannerLogger implements IInteractiveWindowListener
...
-            env.uiKind !== UIKind?.Web &&
-            this.applicationEnvironment.channel === 'stable'
-        );
+        // if (this.applicationEnvironment.uiKind !== UIKind.Desktop || this.applicationEnvironment.channel !== 'stable') {
+        //     return false;
+        // }
+        if (!this.showBannerState.value.expiry) {
+            return true;
+        }
+        return this.showBannerState.value.expiry! < Date.now();
```
DonJayamanne, line 92: Explicit check for expiry.
This is the structure of the data stored in persistent storage if it contains any expiry information (hence ensuring we keep using the old data).

src/client/datascience/dataScienceSurveyBanner.ts
```diff
@@ -149,12 +159,11 @@ export class DataScienceSurveyBanner implements IJupyterExtensionBanner {
...
     }
-
     private async disable(monthsTillNextPrompt: number) {
-        const expiration = monthsTillNextPrompt * 31 * 24 * 60 * 60 * 1000;
-        await this.persistentState
-            .createGlobalPersistentState<boolean>(DSSurveyStateKeys.ShowBanner, false, expiration)
-            .updateValue(false);
+        await this.showBannerState.updateValue({
+            expiry: monthsTillNextPrompt * 31 * MillisecondsInADay,
+            data: true
```
DonJayamanne, line 165: Take over expiry as it can change over time.

IanMatthewHuff, line 165: I think we want false and not true here in disable.

DonJayamanne: Doesn't get used, I'm looking at the `expiry` value instead.

DonJayamanne: Basically if there is no `expiry` then its enabled, else its disabled based on expiry value.

IanMatthewHuff: Ahh got it now, I see that in the enabled function implementation. Yeah, that works. I might toss a comment here at this point? Without looking at the enabled implementation it's a bit non-intuitive.

IanMatthewHuff: @DonJayamanne Wait hang on one more second. Don't you need to add a Date.now() onto the value here? disable(1) will be the same value today and tomorrow. Persistent state expiry did that under the surface.

DonJayamanne: Thanks

DonJayamanne: Phew, big bug, thanks. Tests pass as its a relative time.

src/client/datascience/dataScienceSurveyBanner.ts
```diff
@@ -77,21 +79,38 @@ export class DataScienceSurveyBannerLogger implements IInteractiveWindowListener
...
+    /**
+     * This value is not used.
+     * We are only interested in the value for `expiry`.
+     * This structure is based on the old data for older customers when we used PersistentState class.
+     */
+    data: boolean;
+    /**
+     * If this is value `undefined`, then prompt can be displayed.
+     * If this value is `a number`, then a prompt was displayed at one point in time &
+     * we need to wait for Date.now() to be greater than that number to display it again.
```
DonJayamanne, line 92: @IanMatthewHuff added comments here.

src/client/datascience/dataScienceSurveyBanner.ts
IanMatthewHuff: Yeah, that's the tricky bit with tests like that. You don't want tests to wait a month so you gotta fake it up somehow :).

IanMatthewHuff (approved):

joyceerhl:
I think we'd need this fix in the insidersNativeNotebookSurvey.ts too.

DonJayamanne:
Dmn it, just merged the PR.
Will create another PR for that.

PR #4021, opened 2020-11-25
Fix survey prompt for Native notebooks

DonJayamanne:
For #4002


DavidKutu (approved):

joyceerhl (approved):
Thanks for fixing these 😊 I wrote a test for this banner that is likely completely useless, I can update it in a separate PR

PR #4060, opened 2020-12-01
Fix generation of env variables

DonJayamanne:
I broke the generation of env variables.
Previously it was doing the right thing, found that starting raw kernels was failing because the path was completely wrong (this is because we were not copying all of the env variables).

## TODO
Will add a separate PR to run (C#) raw kernels to ensure we do not regress again.

### DIFF
--- a/src/client/common/variables/environmentVariablesProvider.ts
+++ b/src/client/common/variables/environmentVariablesProvider.ts
@@ -14,6 +14,9 @@ import { EnvironmentVariables, IEnvironmentVariablesProvider, IEnvironmentVariab
 const CACHE_DURATION = 60 * 60 * 1000;
 @injectable()
 export class EnvironmentVariablesProvider implements IEnvironmentVariablesProvider, Disposable {
+    public get onDidEnvironmentVariablesChange(): Event<Uri | undefined> {
+        return this.changeEventEmitter.event;
+    }
     public trackedWorkspaceFolders = new Set<string>();
     private fileWatchers = new Map<string, FileSystemWatcher>();
     private disposables: Disposable[] = [];
@@ -29,10 +32,6 @@ export class EnvironmentVariablesProvider implements IEnvironmentVariablesProvid
         this.changeEventEmitter = new EventEmitter();
         const disposable = this.workspaceService.onDidChangeConfiguration(this.configurationChanged, this);
         this.disposables.push(disposable);
-    }
-
-    public get onDidEnvironmentVariablesChange(): Event<Uri | undefined> {
-        return this.changeEventEmitter.event;
     }
 
     public dispose() {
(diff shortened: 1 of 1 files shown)


src/client/common/variables/environmentVariablesProvider.ts
```diff
@@ -57,22 +56,6 @@ export class EnvironmentVariablesProvider implements IEnvironmentVariablesProvid
         promise.then((result) => (cacheStoreIndexedByWorkspaceFolder.data = result)).ignoreErrors();
         return promise;
     }
-    public async _getEnvironmentVariables(resource?: Uri): Promise<EnvironmentVariables> {
-        let mergedVars = await this.getCustomEnvironmentVariables(resource);
-        if (!mergedVars) {
-            mergedVars = {};
-        }
-        this.envVarsService.mergeVariables(mergedVars!, process.env);
```
DonJayamanne, line 65: The problem was in a previous PR i flipped these two arguments, that was wrong.

src/client/common/variables/environmentVariablesProvider.ts
```diff
@@ -57,22 +56,6 @@ export class EnvironmentVariablesProvider implements IEnvironmentVariablesProvid
         promise.then((result) => (cacheStoreIndexedByWorkspaceFolder.data = result)).ignoreErrors();
         return promise;
     }
-    public async _getEnvironmentVariables(resource?: Uri): Promise<EnvironmentVariables> {
```
DonJayamanne, line 60: Changed from public to private.

src/client/common/variables/environmentVariablesProvider.ts
```diff
@@ -104,6 +87,23 @@ export class EnvironmentVariablesProvider implements IEnvironmentVariablesProvid
             this.disposables.push(envFileWatcher.onDidDelete(() => this.onEnvironmentFileChanged(workspaceFolderUri)));
         }
     }
+    private async _getEnvironmentVariables(resource?: Uri): Promise<EnvironmentVariables> {
+        let customEnvVars = await this.getCustomEnvironmentVariables(resource);
+        if (customEnvVars) {
```
IanMatthewHuff, line 92: This looks wrong. I think that you need a ! here?

IanMatthewHuff: Is that why you also have customEnvVars! below here in the if check? Those shouldn't be needed if this if a ! here.

src/client/common/variables/environmentVariablesProvider.ts
IanMatthewHuff: Yeah, I remember this function from a previous review. Target is the second arg, so this would have merged it into process.env. 

I don't think that we have to change this now, but given that this function has caused some issues maybe it's a problem? Might make more sense as a side effect free version.

mergeVariables(readonly source: EnvVars, readonly target: EnvVars): EnvVars.

Then:

mergedVars = mergeVariables(process.env, mergedVars);

Don't need to change now I think. But the current version does require digging into the function to see what is actually happening.

IanMatthewHuff (changes requested):
Think there might be a ! issue here.

src/client/common/variables/environmentVariablesProvider.ts
rchiodo: Yeah this looks backwards to me too.

DonJayamanne: Yup, explains why all tests are failing.

src/client/common/variables/environmentVariablesProvider.ts
```diff
@@ -104,6 +87,23 @@ export class EnvironmentVariablesProvider implements IEnvironmentVariablesProvid
...
+            customEnvVars = {};
+        }
+        const mergedVars: EnvironmentVariables = {};
+        this.envVarsService.mergeVariables(process.env, mergedVars); // Copy current proc vars into new obj.
+        this.envVarsService.mergeVariables(customEnvVars!, mergedVars); // Copy custom vars over into obj.
+        const pathVariable = this.platformService.pathVariableName;
+        if (customEnvVars![pathVariable]) {
+            this.envVarsService.appendPath(mergedVars!, customEnvVars![pathVariable]!);
```
rchiodo, line 100: Does this function work the opposiste of mergeVariables? The destination is the first argument?

DonJayamanne: Yes
The source is a string.

DonJayamanne: Do you want me to flip it?

rchiodo: Maybe. Or do like Ian suggested for the other function (it returns the appended vars instead of modifying the input param)

rchiodo (approved):

src/client/common/variables/environmentVariablesProvider.ts
rchiodo: But after your last fix, I'm okay with how it is too.

IanMatthewHuff: Looks good. I think that you could remove the !s now that are on the customEnvVars uses. Looks like it should never be undefined now.

IanMatthewHuff (approved):

PR #4064, opened 2020-12-01
Fix variable fetching on remote machines

IanMatthewHuff:
For #4006 


pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -0,0 +1,96 @@
+import pandas as _VSCODE_pd
```
IanMatthewHuff, line 1: The two .py files have not really changed in contents. The helper and two data frame related files have just been squished into one single file here that doesn't need to import anything other than python libraries.

pythonFiles/vscode_datascience_helpers/getVariableInfo/vscodeGetVariableInfo.py
```diff
@@ -1,10 +1,6 @@
 # Query Jupyter server for the info about a dataframe
 import json as _VSCODE_json
-import pandas as _VSCODE_pd
```
IanMatthewHuff, line 3: Imports that were removed here were all unused in variable fetching code.

src/client/datascience/constants.ts
```diff
@@ -488,17 +488,26 @@ export namespace Settings {
...
-    export const DataFrameInfoImportName = '_VSCODE_InfoImport';
-    export const DataFrameInfoImport = `import vscodeGetDataFrameInfo as ${DataFrameInfoImportName}`;
-    export const DataFrameInfoFunc = `${DataFrameInfoImportName}._VSCODE_getDataFrameInfo`;
-    export const DataFrameRowImportName = '_VSCODE_RowImport';
-    export const DataFrameRowImport = `import vscodeGetDataFrameRows as ${DataFrameRowImportName}`;
-    export const DataFrameRowFunc = `${DataFrameRowImportName}._VSCODE_getDataFrameRows`;
-    export const VariableInfoImportName = '_VSCODE_VariableImport';
-    export const VariableInfoImport = `import vscodeGetVariableInfo as ${VariableInfoImportName}`;
-    export const VariableInfoFunc = `${VariableInfoImportName}._VSCODE_getVariableInfo`;
+    export const ScriptPath = path.join(
```
IanMatthewHuff, line 491: Separate GetVariableInfo from DataFrameLoading since the two are independent from each other. We will now directly input the code into the kernel instead of adding to sys.path and importing.

IanMatthewHuff:
Thinking a bit on testing here. We already have tests that cover local testing. However this issue only shows up in remote cases. And in particular you can't test it with "fake" remote by starting a server on your own machine, as the sys paths will be there in that case.

Can we run Docker in tests? That's how I was reproing this locally without needing a second machine.

rchiodo:
> Thinking a bit on testing here. We already have tests that cover local testing. However this issue only shows up in remote cases. And in particular you can't test it with "fake" remote by starting a server on your own machine, as the sys paths will be there in that case.
> 
> Can we run Docker in tests? That's how I was reproing this locally without needing a second machine.

I believe so, yes.

See the answers to this issue:
https://github.community/t/running-automated-tests-inside-the-docker-container-after-building-it/18047

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -243,25 +248,45 @@ export class DebuggerVariables extends DebugLocationTracker
         try {
             // Run our dataframe scripts only once per session because they're slow
             const key = this.debugService.activeDebugSession?.id;
-            if (key && !this.importedIntoKernel.has(key)) {
-                await this.evaluate(DataFrameLoading.DataFrameSysImport);
-                await this.evaluate(DataFrameLoading.DataFrameInfoImport);
-                await this.evaluate(DataFrameLoading.DataFrameRowImport);
-                await this.evaluate(DataFrameLoading.VariableInfoImport);
-                this.importedIntoKernel.add(key);
+            if (key && !this.importedDataFrameScriptsIntoKernel.has(key)) {
```
rchiodo, line 251: Just wanted to make sure you tested this with interactive debugging and run by line?

rchiodo (approved):

src/client/datascience/jupyter/debuggerVariables.ts
IanMatthewHuff: @rchiodo I tested it with interactive cell debugging, but not RBL yet. I'll check that out quick.

IanMatthewHuff: Yeah, looks good here (I'd already run the cell once, so the vars were already there):
image

rchiodo: And dataframes open while debugging too?

IanMatthewHuff: @rchiodo Bleh, I'm glad you asked this. I think it was working in the above example as the variables had already been seen before in a non-debug context.

I don't think this is all working right for the debugger right now. I think that evaluate just takes an expression, so it's only getting one function if the script file contents are passed in.

I think that I can still use this approach for kernelVariables (as those work remote) then convert debugger back to using an import statement. Debugger is only local, so that should be acceptable. The import is a single expression which is probably why it works with evaluate.

DonJayamanne (approved):

rchiodo (approved):

src/client/datascience/jupyter/debuggerVariables.ts
IanMatthewHuff: @rchiodo Not sure if you want to take another look, but I reverted debugger back to using imports. Pretty close to the old code only we just have one file that we import now instead of two for the data frame functions. Checked the following scenarios:
Local IW Vars + DataFrame opening
Local Notebook Vars + DataFrame opening
Local IW Cell Debugging Vars + DataFrame opening
Local Notebook Run By Line Vars + DataFrame opening
Remote IW and Notebook Vars + DataFrame opening

rchiodo: Did you see Joyce's Team's comment about the view in variable menu? This might not work in remote situations for that too.

IanMatthewHuff: @rchiodo Did you mean on GitHub, not Teams? I commented on a thread on GitHub.

If I was reading that right it's more an issue that the sys path might not work for the situation of opening a data frame viewer from a remote debugging session. Since the debuggerVariables path hasn't really changed in this PR I was figuring that I'd still push this PR to fix the non-debugging cases, then pick that up next if it's considered a must-fix.

Was that right? Or were you concerned about this PR regressing something there?

Might have to do something more interesting if that's the case. I think that I could push each individual function over in an evaluate, though that sounds kinda ugly.

rchiodo: Yeah sorry it was this issue here:
https://github.com/microsoft/vscode-jupyter/issues/4065

I agree this isn't necessary now. (especially if this bug is must fix)

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -0,0 +1,98 @@
...
+            row = df.iloc[0:1]
+            json_row = _VSCODE_pd_json.to_json(None, row, date_format="iso")
+            columnNames = list(_VSCODE_json.loads(json_row))
+        except:
+            columnNames = list(df)
+    else:
+        columnNames = list(df)
+
+    # Compute the index column. It may have been renamed
+    indexColumn = df.index.name if df.index.name else "index"
```
joyceerhl, line 72: Not a problem with this PR, but this line seems to be causing opening the data viewer from the debug window to fail with the following error on some machines (no root cause yet).

> line 27, in _VSCODE_getDataFrameInfo
    indexColumn = df.index.name if df.index.name else "index"
AttributeError: 'builtin_function_or_method' object has no attribute 'name'

While we're here:

```suggestion
    try:
        indexColumn = df.index.name
    except AttributeError:
        indexColumn = "index"
```

IanMatthewHuff:
Widget test only failures.

Issue #4067, opened 2020-12-01
Smoke tests for mac is broken


DonJayamanne:
I think i broke it as part of using conda environments for testing purposes.
Moving to must fix as this breaks CI.

joyceerhl:
Validated based on latest insiders run: https://github.com/microsoft/vscode-jupyter/actions/runs/401194617

PR #4068, opened 2020-12-01
Fix smoke tests on Mac

DonJayamanne:
For #4067

### DIFF
--- a/.github/actions/set-python-conda/action.yml
+++ b/.github/actions/set-python-conda/action.yml
@@ -1,5 +1,8 @@
 name: 'Set Python'
-description: "Sets python to a specific version"
+description: "Sets python to a specific version when using Conda Environments"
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yml
```diff
@@ -678,10 +679,16 @@ jobs:
       - name: Install dependencies (npm ci)
         run: npm ci --prefer-offline
 
-      - name: Install test requirements
+      - name: pip install system test requirements
```
DonJayamanne, line 682: Changed to be consistent with what's installed in `insiders.yml`

.github/actions/set-python-conda/action.yml
```diff
@@ -0,0 +1,28 @@
+name: 'Set Python'
```
rchiodo, line 1: Maybe the name should be different?

PR #4069, opened 2020-12-01
Revert change until VSC API is fixed

DonJayamanne:
Disables the change introduced in #398 until the VSC API is fixed (or we know how to use it).
Else we end up making things worse by adding new cells when switching kernels.


src/client/datascience/notebook/serviceRegistry.ts
```diff
@@ -34,10 +33,10 @@ export function registerTypes(serviceManager: IServiceManager) {
         IExtensionSingleActivationService,
         RendererExtension
     );
-    serviceManager.addSingleton<IExtensionSingleActivationService>(
-        IExtensionSingleActivationService,
-        EmptyNotebookCellLanguageService
-    );
+    // serviceManager.addSingleton<IExtensionSingleActivationService>(
```
rchiodo, line 36: Maybe add a comment here explaining when we should turn this back on?

rchiodo (approved):

IanMatthewHuff (approved):

Issue #414, opened 2020-11-06
Gather: Use notebookIdentity instead of notebook.identity when handling external commands


DavidKutu:

rchiodo:
Not sure this is working? I tried this on an untitled notebook and I get this error (after running cells)

```
Please run cells before gathering
```

DavidKutu:
That's expected. Before nothing showed up.

rchiodo:
Maybe that's a separate bug. Works on a notebook with a file.

rchiodo:
> It does work if you run the cell, right?

Not in an untitled notebook.

DavidKutu:
I couldn't repro that, what did you do? I did:

1. Create new blank NB
1. press gather
`Please run cells before gathering` comes out
1. write whatever code on the first cell
1. run the cell
1. press gather
`works as expexted`

rchiodo:
I opened a notebook that was already untitled (reloaded VS code) and ran the first cell. Tried to gather and got the error.

DavidKutu:
Weird, I can't repro in that case either.

rchiodo:
Now the gather button isn't showing up on reload.

rchiodo:
Reloaded again and gather button shows up again. Something is fishy.

PR #4070, opened 2020-12-01
Use notebookIdentity instead of the notebook when handling external buttons

DavidKutu:
For #414

It handles the case when the user doesn't autostart the kernel.

### DIFF
--- a/pythonFiles/pyvsc-run-isolated.py
+++ b/pythonFiles/pyvsc-run-isolated.py
@@ -4,20 +4,14 @@
 if __name__ != "__main__":
     raise Exception("{} cannot be imported".format(__name__))
 
-import os
 import os.path
 import runpy
 import sys
 
-
-def normalize(path):
-    return os.path.normcase(os.path.normpath(path))
-
-
-# We "isolate" the script/module (sys.argv[1]) by removing current working
-# directory or '' in sys.path and then sending the target on to runpy.
-cwd = normalize(os.getcwd())
-sys.path[:] = (p for p in sys.path if p != "" and normalize(p) != cwd)
+# We "isolate" the script/module (sys.argv[1]) by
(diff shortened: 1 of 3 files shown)


rchiodo (approved):

pythonFiles/pyvsc-run-isolated.py
```diff
@@ -4,14 +4,20 @@
...
-# on to runpy.
-sys.path[0] = os.path.join(os.path.dirname(__file__), ".does-not-exist")
+
+def normalize(path):
+    return os.path.normcase(os.path.normpath(path))
+
+
+# We "isolate" the script/module (sys.argv[1]) by removing current working
+# directory or '' in sys.path and then sending the target on to runpy.
+cwd = normalize(os.getcwd())
```
IanMatthewHuff, line 19: I feel like I'm missing context on the run-isolated change. What is this for?

DavidKutu: Nothing related. I was hitting an error related to executing 'import ipykernel'. I updated this file while trying to fix that. And while that wasn't it, I don't see why not keep this change, it just normalizes paths.

DonJayamanne: Personally I would prefer to have a separate PR for such changes.
Helps with simpler reviews, track changes & revert commits as well.

DonJayamanne (approved):

pythonFiles/pyvsc-run-isolated.py
IanMatthewHuff: Yup, this run-isolated file can get pretty funky in some cases, so I think we should consider this as a separate fix. But the interactiveBase part all looks good to me.

IanMatthewHuff (approved):

Issue #3991, opened 2020-11-20
Don't default to exporting HTML and PDF files to the root (/) directory of the user's machine


SiddhanthUnnithan:
Currently when you export to PDF or HTML, the location of the exported file is the root of the user's machine (e.g. /filename.pdf)

Users working on remote hosts (e.g. Azure ML compute, Azure VM) are sometimes unable to write to the root of their host. They have to manually change the export path to a directory that they can write to.

Ideally the exported path defaults to either the root of the user's workspace, current working directory, or the directory in which their Notebook is in.

DonJayamanne:
Bug - as this breaks simple HTML in AML (the export doesn't work with default directory & it needs to be changed to workspace) - could be permissions or the like.
Either way, not the best default experience

greazer:
This more seriously affects remote scenarios.

IanMatthewHuff:
Validated. Didn't default to Mac home directory anymore and tested from Native Notebooks, old webviews, and custom editors.

PR #4072, opened 2020-12-02
Fix export to use the source file for the directory it wants to open

rchiodo:
For #3991 

Rework how export is done to not have to use INotebookModel anymore. This allows the decoupling of the model from the file it has (necessary for the interactive window to export correctly).

Then change all exporting paths to go through the same 'IExportDialog' class which will use the source file to pick the directory to export to.


src/client/datascience/export/exportDialog.ts
```diff
@@ -0,0 +1,70 @@
...
+import { computeWorkingDirectory } from '../jupyter/jupyterUtils';
+import { ExportFormat, IExportDialog } from './types';
+
+// File extensions for each export method
+export const PDFExtensions = { PDF: ['pdf'] };
+export const HTMLExtensions = { HTML: ['html', 'htm'] };
+export const PythonExtensions = { Python: ['py'] };
+
+@injectable()
+export class ExportDialog implements IExportDialog {
```
rchiodo, line 15: This file used to be called ExportFileManager. It really just puts up the dialog now.

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -535,27 +528,32 @@ export class InteractiveWindow extends InteractiveBase implements IInteractiveWi
...
         }
 
-        let model: INotebookModel;
-        this.startProgress();
-        try {
-            model = await this.exportUtil.getModelFromCells(cells);
-        } finally {
-            this.stopProgress();
+        // Pull out the metadata from our active notebook
+        const metadata: nbformat.INotebookMetadata = { orig_nbformat: 3 };
```
rchiodo, line 532: This wasn't working before. Now the exported notebook will have the actual kernel inside of it.

src/client/datascience/jupyter/jupyterUtils.ts
```diff
@@ -96,5 +96,5 @@ export async function computeWorkingDirectory(resource: Resource, workspace: IWo
     }
 
     // Otherwise a file without an extension or directory doesn't exist. Just use the workspace root
-    return workspace.getWorkspaceFolder(resource)?.uri.fsPath || process.cwd();
+    return workspace.getWorkspaceFolder(resource)?.uri.fsPath || workspace.rootPath || process.cwd();
```
rchiodo, line 99: Seems getWorkspaceFolder only works if you're in a workspace.

src/client/datascience/commands/exportCommands.ts
```diff
@@ -69,26 +70,28 @@ export class ExportCommands implements IDisposable {
...
-        model?: INotebookModel,
+        contents?: string,
+        source?: Uri,
         exportMethod?: ExportFormat,
         defaultFileName?: string,
         interpreter?: PythonEnvironment
     ) {
-        if (!model) {
-            // if no model was passed then this was called from the command palette,
+        if (!contents || !source) {
```
IanMatthewHuff, line 86: Can you check calling this Export command from the command palette with all our types of editor? I thought when I was looking at this recently there was a case where a Model was getting passed in. I think that it might have been Old editor active editor = Command is called with all undefined. CustomEditor active editor = Command is called with an INotebookModel. Not 100% sure I'm recalling correctly.

src/client/datascience/jupyter/jupyterUtils.ts
DonJayamanne: FYI - Under the hood `workspace.rootPath` also uses `workspace.worksapceFolders`, hence if you don't have a workspace folder opened, then `workspace.root` will still return `undefined`.

DonJayamanne (approved):

src/client/datascience/jupyter/jupyterUtils.ts
DonJayamanne: No change required.

DonJayamanne:
Like the new changes.

src/client/datascience/export/exportDialog.ts
IanMatthewHuff: Bleh, github didn't pick up the rename though? Can't easily tell what changed...

IanMatthewHuff (approved):
Seems good with just that one thing to check.

DavidKutu (approved):

src/client/datascience/commands/exportCommands.ts
rchiodo: Good thing I double checked. Found another issue. Interactive window would show the export commands if a notebook had ever been opened. Since the active editor here is from the notebook, the export command would essentially do nothing.

The bug was in our tracking of active editor. We weren't firing the change of the active editor on close.

Issue #1254, opened 2020-10-05
Timeout waiting for kernel should have a better error message


rchiodo:
Whenever we fail to launch a kernel (whether raw or jupyter) we should make the kernel stderr (or jupyter stderr) more prevalent.

We consistently get bugs from customers about this sort of thing and if they could find the error in our output they might be able to fix the problem themselves.

greazer:
Let's at least make the message more actionable for the user.

joyceerhl:
Validated with a busted kernel: 

![image](https://user-images.githubusercontent.com/30305945/104361567-77cbc480-54c7-11eb-8d09-27aa87f7d7a6.png)

PR #4092, opened 2020-12-02
Show error after kernel timed out message

DavidKutu:
For #1254


src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -143,8 +143,9 @@ export class KernelProcess implements IKernelProcess {
         } catch (error) {
             // Make sure to dispose if we never get a heartbeat
             this.dispose().ignoreErrors();
-            traceError('Timed out waiting to get a heartbeat from kernel process.');
-            throw new Error('Timed out waiting to get a heartbeat from kernel process.');
+            const message = 'Timed out waiting to get a heartbeat from kernel process.\n';
+            traceError(message + (error as Error).toString());
+            throw new Error(message + (error as Error).toString());
```
DavidKutu, line 148: There are more errors on this file that just show a hardcoded message. Should we show the error to the user everywhere? The original bug only refers to this error which is the most common, but it should be helpful to show it everywhere.

Also, shouldn't all of those messages be localized?

IanMatthewHuff: Was this tried with a broken kernel? I think that the issue here is that the error might not have the full info from the output or the usable output might be way too much to show. My Julia kernel was broken recently and the output in the log looks like this: 
image

I though that we were instead considering pointing users at the output window. Like this error message did:

image

IanMatthewHuff (changes requested):

IanMatthewHuff:
Yeah, I just checked a breakpoint and in a timeout situation here the error is not useful. 

image

src/client/datascience/kernel-launcher/kernelProcess.ts
DonJayamanne: If wrapping an error, please use `WrappedError`, as that preserves the stack trace instead of `new Error(...error.toString())`

Also, I agree with @IanMatthewHuff 
We should check if the kernel has died. In `launchAsObservable` as have the code `exeObs.proc.on('exit', (exitCode) => {` & `exeObs.proc.stderr.on('data', (data: Buffer | string) => {`
We should use that information to determine if the kernel has died (same output is available a few lines earlier).

**My suggestion is:**
* If the kerrnel dies while waiting for heart beat, there's no point waiting for 30s (e.g. if the kernel dies in the first 1s). 
This is very easy to replicate. Just create a bad kernel.
(I came across this recently when I broken starting of raw kernels, the only message i got was a timeout error, when the kernel died immediately).
In such a case we should stop waiting for the heart beat and let the user know the kernel died & display a messages such as the one @IanMatthewHuff suggested `View Jupyter log for further details`
* If the kernel heart beat times out, again we can display the stderror message to the user (even if kernel has not exited).

DonJayamanne (changes requested):

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -291,6 +295,12 @@ export class KernelProcess implements IKernelProcess {
             });
             exeObs.proc.stderr.on('data', (data: Buffer | string) => {
                 traceInfo(`KernelProcess error: ${(data || '').toString()}`);
+                throw new Error(
```
DonJayamanne, line 298: This will not work.

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -103,18 +105,20 @@ export class KernelProcess implements IKernelProcess {
...
+                        return;
                     } else {
                         traceError('KernelProcess Exit', `Exit - ${error.exitCode}, ${error.reason}`, error);
                     }
-                    if (this.disposed) {
-                        return;
-                    }
                     this.exitEvent.fire({ exitCode: error.exitCode, reason: error.reason || error.message });
                 }
+                throw new WrappedError(
```
DonJayamanne, line 114: I dont think this will work.

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -144,7 +148,7 @@ export class KernelProcess implements IKernelProcess {
             // Make sure to dispose if we never get a heartbeat
             this.dispose().ignoreErrors();
             traceError('Timed out waiting to get a heartbeat from kernel process.');
-            throw new Error('Timed out waiting to get a heartbeat from kernel process.');
+            throw new Error(localize.DataScience.kernelTimeout().format(stderr, Commands.ViewJupyterOutput));
```
DonJayamanne, line 151: I don't think we shoudl throw an error like this with instructions in the error mssage.
This means were just going to do a `error.toString()` and display it in the messagebox.

We should instead handle the error and display a meaningful message with details of the error.

DavidKutu:
Here's how the error looks:
image

DavidKutu:
And here's the log when the user clicks on log:
image

Hopefully they attach all of this in future issues.

rchiodo (approved):

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -85,6 +85,33 @@ export class KernelProcess implements IKernelProcess {
...
+            if (!exitEventFired) {
+                this.exitEvent.fire({ exitCode: exitCode || undefined, reason: stderrProc });
+                exitEventFired = true;
+            }
+            cancelWaiting.cancel();
+        });
+
+        exeObs.proc!.stdout.on('data', (data: Buffer | string) => {
+            stdoutProc += data.toString();
+            traceInfo(`Kernel Output: ${stdoutProc}`);
```
IanMatthewHuff, line 107: Looks like on the stdout data here it's doing an append then a print. Unless I'm reading this wrong. So for each trace error are you going to print the current state as well as all previous stdout?

DavidKutu: Right, those should just be printing. We only need to append the errors to show them later

DonJayamanne (approved):

IanMatthewHuff (approved):

Issue #4079, opened 2020-12-02
Functional tests group 2 always failing with dependency issue.


IanMatthewHuff:
![image](https://user-images.githubusercontent.com/812783/100911380-fe7e9200-3483-11eb-8027-0bd809790ab2.png)

Tests all pass, but the stage shows up as an failure due to this.

joyceerhl:
Validated based on latest main branch run.

PR #4098, opened 2020-12-03
Attempt to fix test by not including two classes in same file

rchiodo:
For #4079 

Inversify is failing when loading the helpers.ts file. Seems weird. Trying by removing the class helpers is using into a separate file.


rchiodo:
Seems to have worked.

IanMatthewHuff:
Not fully sure that I understand the why...but if it works, it works.

IanMatthewHuff (approved):

rchiodo:
> Not fully sure that I understand the why...but if it works, it works.

Yeah I'm not 100% sure either. If you look at the location in helpers.js where the failure occurs, it looks like it's trying to load an inversified class when it shouldn't be. The helpers file only needs the non inversified class. So maybe it's a limitation of inversify. Maybe you can't load the file until after the container is ready? I couldn't repro locally.

PR #4103, opened 2020-12-03
Fix smoke test on Mac

DonJayamanne:
For #4067

### DIFF
--- a/.github/actions/set-python-conda/action.yml
+++ b/.github/actions/set-python-conda/action.yml
@@ -1,4 +1,4 @@
-name: 'Set Python'
+name: 'Set Python for Conda'
 description: "Sets python to a specific version when using Conda Environments"
 # For some reason we cannot use `if` conditions in the steps of an action. Github Action throws errors.
 # Hence created a separate action and used the action conditionally.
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yml
```diff
@@ -678,10 +692,16 @@ jobs:
       - name: Install dependencies (npm ci)
         run: npm ci --prefer-offline
 
-      - name: Install test requirements
+      - name: pip install system test requirements
         run: |
-          python -m pip install numpy
-          python -m pip install --upgrade jupyter
+          python -m pip install --upgrade -r build/test-requirements.txt
```
DonJayamanne, line 697: Changed to be consistent with what's installed in insiders.yml

IanMatthewHuff (approved):

.github/actions/set-python-conda/action.yml
```diff
@@ -0,0 +1,28 @@
+name: 'Set Python'
```
rchiodo, line 1: Maybe change the name? This one is specific to conda

rchiodo (approved):

Issue #4105, opened 2020-12-03
Error notification: CustomEditor and NativeNotebook experiments cannot be turned on together


aeschli:
Testing codespaces scenario https://github.com/microsoft/vssaas-planning/issues/1593
- open a codespace with a Python file
- among other notification dialogs (see microsoft/vscode-python#14833) I get 

![image](https://user-images.githubusercontent.com/6461412/100997910-268cf480-355b-11eb-8674-8fa82a3cc237.png)

Unclear to me as user what has caused this and what I need to do differently.

DonJayamanne:
I think i know what's going on here.
* In insiders NativeNotebook is always on in experiment.
* Now assume user happens to fall into the CustomEditor experiment in VS Code Insiders.
At this point, user is in both experiment groups.

Solution
* Modify the `inExperiment` method to check if `experiment argument == CustomEditor` & if user already belongs to native notebook experiment, in such a case, just return `false`

PaoloTCS:
I have the identical message.  Latest insiders edition of VSC.

joyceerhl:
Validated. I can't simulate control tower opting me into custom editor, but if I manually opt into native notebooks and custom editor experiments in VSCode Insiders, creating a new blank Jupyter notebook still opens into the native notebooks UI and I don't get a prompt. 

@PaoloTCS we have a fix in for this. If the fix works as expected, it will be available in our next release, which will hopefully be in about 2 weeks' time. Thanks for letting us know 😊

PR #4115, opened 2020-12-03
Check if already in native nb experiment

DonJayamanne:
For #4105

### DIFF
--- a/src/client/common/experiments/service.ts
+++ b/src/client/common/experiments/service.ts
@@ -87,6 +87,14 @@ export class ExperimentService implements IExperimentService {
             return false;
         }
 
+        // If user is already in Native Notebook experiment, then they cannot be in Custom Editor experiment.
+        if (
+            experiment === ExperimentGroups.CustomEditor &&
+            this.getOptInOptOutStatus(ExperimentGroups.NativeNotebook) === 'optIn'
+        ) {
+            return false;
+        }
+
         // Currently the service doesn't support opting in and out of experiments,
         // so we need to perform these checks and send the corresponding telemetry manually.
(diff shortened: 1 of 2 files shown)


src/client/common/experiments/service.ts
```diff
@@ -87,6 +87,14 @@ export class ExperimentService implements IExperimentService {
             return false;
         }
 
+        // If user is already in Native Notebook experiment, then they cannot be in Custom Editor experiment.
```
rchiodo, line 90: I swear we had this somewhere else already? Maybe it was in the wrong spot before

rchiodo (approved):

src/client/common/experiments/service.ts
IanMatthewHuff: extensionActivation.ts?

IanMatthewHuff: That location just gives a warning if you are in both and asks you to disable one. I think the intention was that we only though that could happen if you opted in manually.

IanMatthewHuff: Do we know that this issue was due to the user opting in? This change doesn't look like it guards against control tower issues.

DonJayamanne: > This change doesn't look like it guards against control tower issues.

This will ensure we don't even bother checking.
Hence telemetry will not be sent for this user for this experiment.

IanMatthewHuff: Ahh, I didn't see where you had added the Insiders => NativeNotebook forcing function. Yeah, this works since that force is in the optinoptout function.

IanMatthewHuff (approved):

Issue #1701, opened 2020-10-02
Code getting shuffled -- causing errors 


claudiaregio:
# Bug: Notebook Editor, Interactive Window, Editor cells

## Issue

Not sure how to get this to reproduce but got the same complaint from another customer two days ago. 
Was using the notebook and running and editing code and all of a sudden I got an error for code that is not in my cell. 
The cell appears to have somehow laced code together that led to an error. 

The only solution is to copy and paste the code into a new cell and delete the old. 

## Actual behavior

![image](https://user-images.githubusercontent.com/19276747/94889196-2e7fc980-0430-11eb-965b-f47a6171b1f2.png)

### Your Jupyter and/or Python environment

_Please provide as much info as you readily know_

- Jupyter server running: Local
- Extension version: v2020.10.111333
- VS Code version: 1.49.2
- Setting python.languageServer: Pylance
- Python and/or Anaconda version: 3.7.6
- OS: Windows: x64 10.0.19042
- Virtual environment: conda

## Python Output

Info 2020-10-01 21:53:08: Executing cell NotebookImport#0
Info 2020-10-01 21:53:08: Submitting code for 3d9bab26-629b-4fb7-bf33-f50ce597a9e7
Info 2020-10-01 21:53:08: Kernel switching to busy
Info 2020-10-01 21:53:08: Kernel switching to idle
Info 2020-10-01 21:53:08: Kernel switching to busy
Info 2020-10-01 21:53:09: Kernel switching to idle
Info 2020-10-01 21:53:09: Finished execution for NotebookImport#0
Info 2020-10-01 21:53:09: Finished executing cell NotebookImport#0
Info 2020-10-01 21:53:12: Executing cell NotebookImport#1
Info 2020-10-01 21:53:12: Submitting code for 3d9bab26-629b-4fb7-bf33-f50ce597a9e7
Info 2020-10-01 21:53:12: Kernel switching to busy
Info 2020-10-01 21:53:13: Kernel switching to idle
Info 2020-10-01 21:53:13: Finished execution for NotebookImport#1
Info 2020-10-01 21:53:13: Finished executing cell NotebookImport#1
Info 2020-10-01 21:53:14: Executing cell NotebookImport#2
Info 2020-10-01 21:53:14: Submitting code for 3d9bab26-629b-4fb7-bf33-f50ce597a9e7
Info 2020-10-01 21:53:14: Kernel switching to busy
[...]

PR #4134, opened 2020-12-05
Reenforce code synchronization

rchiodo:
For #1701 

Force syncing the model in memory whenever 
- Saving (native and old webview)
- Executing a cell (old webview only as VS code sends this value).

Comments in 1701 indicate saving seems to fail for VS code notebooks too.

Implemented this for old webviews by:
- New messages to get code from cell(s) (not entire cell, took too long for large notebooks)
- New messages to return code from cell(s)
- Send these messages and wait for response before executing a cell or saving

Implemented this for native editor by:
- Reasking for the source from the VS code cell documents when saving


src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -95,6 +97,15 @@ export class NotebookEditor implements INotebookEditor {
         );
         disposables.push(model.onDidDispose(this._closed.fire.bind(this._closed, this)));
     }
+    @captureTelemetry(Telemetry.SyncAllCells)
+    public async syncAllCells(): Promise<void> {
+        // Need to read content out of the UI and save in our model
+        const modifiedCells = (this.model as VSCodeNotebookModel).getCells().map((c, i) => {
```
rchiodo, line 103: Need to double check this. This is making an assumption that the cells in the document have the same relative position as they do in our notebook model

rchiodo: Yeah they seem to.

src/client/datascience/notebookStorage/notebookModelSynchronization.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { inject, injectable } from 'inversify';
+import { traceInfo } from '../../common/logger';
+import { IFileSystem } from '../../common/platform/types';
+import { INotebookEditor, INotebookEditorProvider, INotebookModel, INotebookModelSynchronization } from '../types';
+
+@injectable()
+export class NotebookModelSynchronization implements INotebookModelSynchronization {
```
rchiodo, line 10: Created this extra class so that NotebookStorageProvider wouldn't have to know about the UI

src/client/datascience/interactive-ipynb/nativeEditor.ts
```diff
@@ -233,6 +236,58 @@ export class NativeEditor extends InteractiveBase implements INotebookEditor {
...
+    @captureTelemetry(Telemetry.SyncAllCells)
+    public async syncAllCells(): Promise<void> {
+        const timer = new StopWatch();
+        // Ask the UI for all of our all cells
+        this.waitingForMessageResponse = {
+            message: InteractiveWindowMessages.ReturnAllCellCode,
+            result: createDeferred<string[]>()
+        };
+        await this.postMessage(InteractiveWindowMessages.GetAllCellCode);
+        const code = await this.waitingForMessageResponse.result.promise;
```
DonJayamanne, line 248: This method will never complete if we have another request for `syncAllCell`.
For the second execution of `syncAllCells`, the `result` points to another deferred.
Hence the first request will never complete.

Thus when saving, only the last save wins, I guess its ok. Last one wins.
If this is by design then fine, else concerned that if we have auto save, then it will never save until the user stops typing or doing something (because as the user types, we get more of these getting into the pipeline and the subsequent saves result in new promises, meaning only last one wins).

src/client/datascience/interactive-ipynb/nativeEditor.ts
```diff
@@ -233,6 +236,58 @@ export class NativeEditor extends InteractiveBase implements INotebookEditor {
...
+
+    @captureTelemetry(Telemetry.SyncSingleCell)
+    public async syncCell(cellId: string): Promise<void> {
+        // Ask the UI for the code for this cell
+        this.waitingForMessageResponse = {
+            message: InteractiveWindowMessages.ReturnCellCode,
+            result: createDeferred<ICell | undefined>()
+        };
+        await this.postMessage(InteractiveWindowMessages.GetCellCode, cellId);
+        const code = await this.waitingForMessageResponse.result.promise;
```
DonJayamanne, line 274: This will prevent the `syncAllCells` from working, if this method is call after `syncAlLCells`
Meaning we'd end up only saving the specific cell.

src/client/datascience/interactive-ipynb/nativeEditor.ts
rchiodo: Yes good point. I'll make it a list and pass in an id.

src/client/datascience/notebookStorage/notebookModelSynchronization.ts
```diff
@@ -0,0 +1,28 @@
...
+        @inject(INotebookEditorProvider) private readonly notebookEditorProvider: INotebookEditorProvider,
+        @inject(IFileSystem) private readonly fs: IFileSystem
+    ) {}
+    public async syncAllCells(model: INotebookModel): Promise<void> {
+        // Find the owner of this model and ask it to do the sync (this is inherently a UI operation as we need to pull the data from the UI)
+        const owner = await this.getOwner(model);
+        if (owner) {
+            return owner.syncAllCells();
+        } else {
+            traceInfo(`Sync all not possible for ${model.file.toString()}`);
```
DonJayamanne, line 21: Shouldn't this be a `traceError`, I can't see why or why this would occur.

src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -95,6 +97,15 @@ export class NotebookEditor implements INotebookEditor {
...
+            const dc = this.document.cells[i];
```
DonJayamanne, line 104: Curious, why do we need this for `native notebooks`?
Was that failing as well?

src/client/datascience/notebookStorage/vscNotebookModel.ts
```diff
@@ -89,6 +89,14 @@ export class VSCodeNotebookModel extends BaseNotebookModel {
         this._cells = this.notebookJson.cells || [];
         this._preferredLanguage = cellLanguageService.getPreferredLanguage(this.metadata);
     }
+    public getCells() {
+        return this._cells;
```
DonJayamanne, line 93: Please change to `cells` property & add in the base class, the other class used for WebView notebooks has a `cells` property.

DonJayamanne (changes requested):
Minor tweeks

src/client/datascience/notebook/notebookEditor.ts
rchiodo: Yes, but only for save, not for execute. See this comment here:
https://github.com/microsoft/vscode-jupyter/issues/1701#issuecomment-738929286

src/client/datascience/notebookStorage/vscNotebookModel.ts
rchiodo: cells property on other class returns a different type though (ICell []) instead of nbformat.IBaseCell[]. I can change it in this sub class, but changing the return type will ripple elsewhere.

DonJayamanne:
With regards to the bug in native notebooks.
It is not necessary
Today when ever we generate the JSON, we get the raw cell code from the cell text document. I.e. the ICell is not used. We dropped usage of `ICell` in native editor. It is only required for initial load (couldn't remove ICell completely as we use that interface for reading & parsing ipynb files).

If it isn't working for a user I'd rather try to get to the bottom of that bug, and fix that. In this case it looks like the VSC API isn't returning the correct text, hence I'd say we should report this to VS Code.

src/client/datascience/notebookStorage/vscNotebookModel.ts
DonJayamanne: Not required.
See comments regarding not requiring this change for native notebooks.
When saving, we call `model.getContent`
& the `getContent` method calls the `getnerateNotebookContent` -> `generateNotebookJson` -> `getJupyterCells`.

In VSC Native notebooks we use `createJupyterCellFromVSCNotebookCell` which gets the raw text from the Text Document.
I.e. ICell is not used in native notebooks when editing/saving (only required for loading - would love to get rid of that as well, but thats a bigger change).

rchiodo:
> With regards to the bug in native notebooks.
> It is not necessary
> Today when ever we generate the JSON, we get the raw cell code from the cell text document. I.e. the ICell is not used. We dropped usage of `ICell` in native editor. It is only required for initial load (couldn't remove ICell completely as we use that interface for reading & parsing ipynb files).
> 
> If it isn't working for a user I'd rather try to get to the bottom of that bug, and fix that. In this case it looks like the VSC API isn't returning the correct text, hence I'd say we should report this to VS Code.

Well I couldn't repro it myself. I thought since it was using the model it would be reading the cells from the code we loaded but you're right it eventually calls getJupyterCells. That should be using the notebook document unless the document isn't there? Doesn't seem likely though.

rchiodo:
I think for now I'll leave the sync all cells method. It doesn't do much of anything in the native notebook case.

rchiodo:
> I think for now I'll leave the sync all cells method. It doesn't do much of anything in the native notebook case.

Actually it could be unnecessarily recreating the cell array. You delete it in the trust functions.

src/client/datascience/notebookStorage/notebookModelSynchronization.ts
```diff
@@ -0,0 +1,28 @@
...
+        const owner = await this.getOwner(model);
+        if (owner) {
+            return owner.syncAllCells();
+        } else {
+            traceError(`Sync all not possible for ${model.file.toString()}`);
+        }
+    }
+
+    private async getOwner(model: INotebookModel): Promise<INotebookEditor | undefined> {
+        return this.notebookEditorProvider.editors.find((e) => this.fs.arePathsSame(e.model.file, model.file));
```
DonJayamanne, line 26: This need not be async, can be a sync method.

DonJayamanne (approved):
Last change, no need to make a method async (slows things down unnecessarily by yieling to some other code that could be time consuming).
Else all good.

DonJayamanne (approved):

Issue #4166, opened 2020-12-08
Julia kernel working in previous release is now broken in current release


IanMatthewHuff:

rchiodo:
Is this fixed now?

IanMatthewHuff:
Sure and I supposed it's validated by me testing the VSIX. Closing.

PR #4168, opened 2020-12-08
Fixes merging PATH and PYTHONPATH in env variables

DonJayamanne:
For #4166 
Was working for me as i have a custom .env file with some stuff in there.

### DIFF
--- a/src/client/common/variables/environment.ts
+++ b/src/client/common/variables/environment.ts
@@ -32,7 +32,11 @@ export class EnvironmentVariablesService implements IEnvironmentVariablesService
         if (!target) {
             return;
         }
+        const settingsNotToMerge = ['PYTHONPATH', this.pathVariable];
         Object.keys(source).forEach((setting) => {
+            if (settingsNotToMerge.indexOf(setting) >= 0) {
+                return;
+            }
             target[setting] = source[setting];
         });
     }
--- a/src/client/common/variables/environment.ts
+++ b/src/client/common/variables/environment.ts
@@ -32,11 +32,7 @@ export class EnvironmentVariablesService implements IEnvironmentVariablesService
         if (!target) {
             return;
         }
-        const settingsNotToMerge = ['PYTHONPATH', this.pathVariable];
         Object.keys(source).forEach((setting) => {
-            if (settingsNotToMerge.indexOf(setting) >= 0) {
-                return;
-            }
             target[setting] = source[setting];
         });
     }
--- a/src/client/common/variables/environmentVariablesProvider.ts
+++ b/src/client/common/variables/environmentVariablesProvider.ts
@@ -96,6 +96,12 @@ export class EnvironmentVariablesProvider implements IEnvironmentVariablesProvid
         this.envVarsService.mergeVariables(process.env, mergedVars); // Copy current proc vars into new obj.
         this.envVarsService.mergeVariables(customEnvVars!, mergedVars); // Copy custom vars over into obj.
         const pathVariable = this.platformService.pathVariableName;
+        if (process.env[pathVariable]) {
+            mergedVars[pathVariable] = process.env[pathVariable];
+        }
+        if (process.env.PYTHONPATH) {
+            mergedVars.PYTHONPATH = process.env.PYTHONPATH;
+        }
         if (customEnvVars![pathVariable]) {
             this.envVarsService.appendPath(mergedVars!, customEnvVars![pathVariable]!);
         }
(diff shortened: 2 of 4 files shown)


DavidKutu (approved):

src/client/common/variables/environment.ts
```diff
@@ -32,11 +32,7 @@ export class EnvironmentVariablesService implements IEnvironmentVariablesService
         if (!target) {
             return;
         }
-        const settingsNotToMerge = ['PYTHONPATH', this.pathVariable];
         Object.keys(source).forEach((setting) => {
-            if (settingsNotToMerge.indexOf(setting) >= 0) {
```
joyceerhl, line 37: I'm not sure I understand the fix here. So the problem is that we were previously skipping merging the PYTHONPATH and PATH settings?

rchiodo (approved):

src/client/common/variables/environment.ts
DonJayamanne: Yes, and then we added code to explicity merge it later,

IanMatthewHuff: Doesn't this conflict with the kernelEnv and custeEditVars merging of path and python path? It's now replacing on merge and appending later.

IanMatthewHuff (changes requested):

DonJayamanne:
Closed accidentally

DonJayamanne:
> Doesn't this conflict with the kernelEnv and custeEditVars merging of path and python path? It's now replacing on merge and appending later.

It will merge and append yes thats correct.
Previously it was doing the same thing, but didn't append as the value was empty.

Thus:
* Assume user has PATH = ABC
* & Custom variable with PATH=HELLO
* Now we have PATH=ABC,HELLO

IanMatthewHuff:
@DonJayamanne Take these two lines as an example.
```
        this.envVarsService.mergeVariables(customEditVars, mergedVars); // custom vars win over all.
        if (customEditVars[this.platformService.pathVariableName]) {
            this.envVarsService.appendPath(mergedVars, customEditVars[this.platformService.pathVariableName]!);
        }
```

IanMatthewHuff (changes requested):
Still a small issue with the merging order it looks like to me.

DonJayamanne:
> Still a small issue with the merging order it looks like to me.

Agreed, fixed

src/test/common/variables/envVarsService.unit.test.ts
```diff
@@ -206,7 +205,7 @@ PYTHON=${BINDIR}/python3\n\
                 variablesService.mergeVariables(vars1, vars2);
 
                 expect(Object.keys(vars1)).lengthOf(2, 'Source variables modified');
-                expect(Object.keys(vars2)).lengthOf(5, 'Variables not merged');
+                expect(Object.keys(vars2)).lengthOf(3, 'Variables not merged');
```
DonJayamanne, line 208: PATH & PYTHONPATH should not be merged (we validate this by checking the expected 3 properties)

src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
```diff
@@ -69,24 +69,25 @@ export class KernelEnvironmentVariablesService {
         this.envVarsService.mergeVariables(interpreterEnv, mergedVars); // interpreter vars win over proc.
         this.envVarsService.mergeVariables(kernelEnv, mergedVars); // kernels vars win over interpreter.
         this.envVarsService.mergeVariables(customEditVars, mergedVars); // custom vars win over all.
-        if (customEditVars[this.platformService.pathVariableName]) {
-            this.envVarsService.appendPath(mergedVars, customEditVars[this.platformService.pathVariableName]!);
+        const pathVariable = this.platformService.pathVariableName;
+        if (process.env[pathVariable]) {
```
IanMatthewHuff, line 73: mergedVars is initialized from process.env and mergeVariables is now back to ignoring PATH and PYTHONPATH so these two ifs don't look needed. If process.env has a PATH or PYTHONPATH it's already in mregedVars and none of the merges above would overwrite them.

src/client/common/variables/environmentVariablesProvider.ts
```diff
@@ -96,6 +96,12 @@ export class EnvironmentVariablesProvider implements IEnvironmentVariablesProvid
         this.envVarsService.mergeVariables(process.env, mergedVars); // Copy current proc vars into new obj.
         this.envVarsService.mergeVariables(customEnvVars!, mergedVars); // Copy custom vars over into obj.
         const pathVariable = this.platformService.pathVariableName;
+        if (process.env[pathVariable]) {
```
IanMatthewHuff, line 99: Anything that uses this now has a possible PATH or PYTHONPATH value from the process.env that it would not have had before. Seems like a bit of a big change, but I guess we'll just have to rely on tests catching any fallout on this.

DonJayamanne: > Anything that uses this now has a possible PATH or PYTHONPATH value from the process.env that it would not have had before. Seems like a bit of a big change, but I guess we'll just have to rely on tests catching any fallout on this.

It should be there if it wasn't.

IanMatthewHuff (approved):

Issue #4047, opened 2020-11-30
Playwright error in ipywidgets ui test


joyceerhl:
![image](https://user-images.githubusercontent.com/30305945/100644626-6c925000-32f0-11eb-846b-db30f2c6e113.png)

First time I've seen this, logging as an issue here so we can track it. Might be related to https://github.com/microsoft/playwright/issues/2974

DonJayamanne:
Looks like the original issue reported by others aren't reproducing, and I haven't seen this either, hence suggest we close this.
After all, not sure what we're supposed to do here, its a flaky issue with playwright.
@joyceerhl thoughts?

DonJayamanne:
Closing for now, as we haven't seen these issues.

PR #4189, opened 2020-12-09
Add logging to determine why variable tests were timing out

DonJayamanne:
For #4047

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -338,20 +338,20 @@ jobs:
 
       - name: Install dependencies (npm ci)
         run: npm ci --prefer-offline
-      
-      - name: Install Dot.net	
-        uses: actions/setup-dotnet@v1	
-        with:	
-          dotnet-version: ${{env.DOTNET_VERSION}}	
-        if: matrix.test-suite != 'notebookWithoutPythonExt'	
 
-      - name: Install .NET Interactive	
-        run: dotnet tool install -g --add-source "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" Microsoft.dotnet-interactive	
-        if: matrix.test-suite != 'notebookWithoutPythonExt'	
+      - name: Install Dot.net
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: ${{env.DOTNET_VERSION}}
+        if: matrix.test-suite != 'notebookWithoutPythonExt'
 
-      - name: Install .NET Kernel	
-        run: dotnet interactive jupyter install	
-        if: matrix.test-suite != 'notebookWithoutPythonExt'	
+      - name: Install .NET Interactive
+        run: dotnet tool install -g --add-source "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" Microsoft.dotnet-interactive
+        if: matrix.test-suite != 'notebookWithoutPythonExt'
+
+      - name: Install .NET Kernel
+        run: dotnet interactive jupyter install
+        if: matrix.test-suite != 'notebookWithoutPythonExt'
 
         # This step is slow.
       - name: Install Julia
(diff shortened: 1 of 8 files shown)


src/client/common/process/baseDaemon.ts
```diff
@@ -177,7 +178,7 @@ export abstract class BasePythonDaemon {
         return Object.keys(options).every((item) => daemonSupportedSpawnOptions.indexOf(item as any) >= 0);
     }
     protected sendRequestWithoutArgs<R, E, RO>(type: RequestType0<R, E, RO>): Thenable<R> {
-        if (this.proc && typeof this.proc.exitCode !== 'number') {
+        if (this.isAlive && this.proc && typeof this.proc.exitCode !== 'number') {
```
DonJayamanne, line 181: While running the tests locally found that there were some other unhandled exceptions.
Basically after the tests we kill the daemon, however some async code is still attempting to write to the daemon.

This ensures we handle that situation & thus prevent logging or unhandled errors (i.e. spurious errors being logged)

src/test/datascience/debugger.functional.test.tsx
```diff
@@ -155,8 +158,9 @@ suite('DataScience Debugger tests', () => {
         }
 
         // Start the jupyter server
+        traceInfoIf(!!process.env.VSC_CI_ENABLE_TOO_MUCH_LOGGING, 'Getting interactive window');
         const history = await getOrCreateInteractiveWindow(ioc);
-
+        traceInfoIf(!!process.env.VSC_CI_ENABLE_TOO_MUCH_LOGGING, 'Got interactive window');
```
DonJayamanne, line 163: Added some logging to find out what part is hanging when the variable tests timeout.

src/test/datascience/testHelpers.tsx
```diff
@@ -59,29 +59,41 @@ async function testInnerLoop(
 
 export function runDoubleTest(
     name: string,
-    testFunc: (type: 'native' | 'interactive') => Promise<void>,
+    testFunc: (this: Mocha.Context, type: 'native' | 'interactive') => Promise<void>,
```
DonJayamanne, line 62: Fixes to properly pass the context/scope, so we can use `this.skip`.
We had `c.skip` but that didn't work as the scope wasn't setup properly.

src/test/initialize.ts
```diff
@@ -76,23 +76,43 @@ export async function closeActiveNotebooks(): Promise<void> {
...
-                reject(ex);
-            }
-        );
-    });
+    class CloseEditorsTimeoutError extends Error {
+        constructor() {
+            super("Command 'workbench.action.closeAllEditors' timed out");
+        }
+    }
+    const closeWindowsImplementation = (timeout = 15_000) => {
```
DonJayamanne, line 84: Found in the PR that VS Code was timing out here.
We've had this issue in the past hence the timeout.
Updated to ensure we retry again after sleeping for 1/2s.
Possible VSC is stuck/busy, hence giving it time & retrying.

I.e. a work around for something flaky in VS Code.

rchiodo (approved):

src/client/common/process/baseDaemon.ts
IanMatthewHuff: Nice. Think that I'd seen those errors cluttering up the logs before.

IanMatthewHuff (approved):

.github/workflows/pr.yml
```diff
@@ -600,6 +600,20 @@ jobs:
 
       # Used by tests for non-python kernels.
       # Test are enabled via env variable `VSC_JUPYTER_CI_RUN_NON_PYTHON_NB_TEST`
+      - name: Install Dot.net
```
DonJayamanne, line 603: This seems to have been removed from main, I added it back.

.github/workflows/pr.yml
```diff
@@ -338,20 +338,20 @@ jobs:
...
-      - name: Install Dot.net	
-        uses: actions/setup-dotnet@v1	
-        with:	
-          dotnet-version: ${{env.DOTNET_VERSION}}	
-        if: matrix.test-suite != 'notebookWithoutPythonExt'	
 
-      - name: Install .NET Interactive	
-        run: dotnet tool install -g --add-source "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" Microsoft.dotnet-interactive	
-        if: matrix.test-suite != 'notebookWithoutPythonExt'	
+      - name: Install Dot.net
```
DonJayamanne, line 342: Not sure why this is here, but leave it (no harm cause).
Difficult with all of the copy paste going on in github actions - problem is with github actions.

Issue #4109, opened 2020-12-03
Native Notebooks are creating a new kernel on any execution after moving to a different document


IanMatthewHuff:
Pretty bad issue right now, especially as I don't think that the old kernel is getting shutdown. It will do this again and again for me when moving to another document. 

![NewKernel.gif](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/0adba581-9d05-44ca-a668-bd621f8874f6)

Have one suspicion that I'm looking into right now.

greazer:
Not able to repro this on shipping extension. May be a regression that was recently caused (but thankfully not shipped)

DonJayamanne:
I tried, & cannot repro this issue.

DavidKutu:
I'm also unable to repro this.

IanMatthewHuff:
@DavidKutu @DonJayamanne I'll take over investigation on this.

IanMatthewHuff:
Did some more work to pin down at least one repro. Scenario is as follows. Python extension required for this scenario.

1. Create a new jupyter notebook
2. Run this notebook with a selected installed kernelspec, not just an interpreter (we need 'startUsingKernelSpec' as our connection type)
3. Change the active interpreter
4. Do something to retrigger vscode kernel provider fetching swap to a different document and back
5. Execute a cell again (it will reconnect to a new kernel instance)

rchiodo:
Validated. Does not create a new kernel after following the steps above.

PR #4192, opened 2020-12-09
only check kernelspec on kernel provider comparison for kernelspec connections

IanMatthewHuff:
For #4109

### DIFF
--- a/src/client/datascience/jupyter/kernels/kernelProvider.ts
+++ b/src/client/datascience/jupyter/kernels/kernelProvider.ts
@@ -42,8 +42,21 @@ export class KernelProvider implements IKernelProvider {
     }
     public getOrCreate(uri: Uri, options: KernelOptions): IKernel | undefined {
         const existingKernelInfo = this.kernelsByUri.get(uri.toString());
-        if (existingKernelInfo && fastDeepEqual(existingKernelInfo.options.metadata, options.metadata)) {
-            return existingKernelInfo.kernel;
+        if (existingKernelInfo) {
+            if (
+                existingKernelInfo.options.metadata.kind === 'startUsingKernelSpec' &&
+                options.metadata.kind === 'startUsingKernelSpec'
+            ) {
+                // When using a specific kernelspec, just compare the actual kernel specs
+                if (fastDeepEqual(existingKernelInfo.options.metadata.kernelSpec, options.metadata.kernelSpec)) {
+                    return existingKernelInfo.kernel;
+                }
+            } else {
+                // If not launching via kernelspec, compare the entire metadata
+                if (fastDeepEqual(existingKernelInfo.options.metadata, options.metadata)) {
+                    return existingKernelInfo.kernel;
+                }
+            }
         }
 
         this.disposeOldKernel(uri);
--- a/news/2 Fixes/4109.md
+++ b/news/2 Fixes/4109.md
@@ -1,0 +1,1 @@
+When comparing to existing running kernel only consider the kernelspec when launched via kernelspec.


src/client/datascience/jupyter/kernels/kernelProvider.ts
```diff
@@ -42,8 +42,21 @@ export class KernelProvider implements IKernelProvider {
...
         const existingKernelInfo = this.kernelsByUri.get(uri.toString());
-        if (existingKernelInfo && fastDeepEqual(existingKernelInfo.options.metadata, options.metadata)) {
-            return existingKernelInfo.kernel;
+        if (existingKernelInfo) {
+            if (
+                existingKernelInfo.options.metadata.kind === 'startUsingKernelSpec' &&
+                options.metadata.kind === 'startUsingKernelSpec'
+            ) {
+                // When using a specific kernelspec, just compare the actual kernel specs
+                if (fastDeepEqual(existingKernelInfo.options.metadata.kernelSpec, options.metadata.kernelSpec)) {
```
DonJayamanne, line 51: So what was causing the issue here.

I.e. under what condition was the old code failing?
I'm trying to understand what was different in other properties of the `options.metadata` object that was different and cause the bug.

DonJayamanne: Oh noticed instructions here https://github.com/microsoft/vscode-jupyter/issues/4109#issuecomment-742043965

src/client/datascience/jupyter/kernels/kernelProvider.ts
```diff
@@ -42,8 +42,21 @@ export class KernelProvider implements IKernelProvider {
     }
     public getOrCreate(uri: Uri, options: KernelOptions): IKernel | undefined {
         const existingKernelInfo = this.kernelsByUri.get(uri.toString());
-        if (existingKernelInfo && fastDeepEqual(existingKernelInfo.options.metadata, options.metadata)) {
-            return existingKernelInfo.kernel;
+        if (existingKernelInfo) {
```
IanMatthewHuff, line 45: This getOrCreate runs on every cell execution. It's checking to see if the new KernelOptions specified still match the current kernel running the notebook. If it matches it just keeps using that kernel, if it doesn't match it starts a new one.

When you launch via a kernelspec that doesn't have an associated interpreter it just fills in the interpreter in the metadata with the current active interpreter. So if you change the active interpreter you can now create a mismatch here and the kernel restarts.

rchiodo (approved):

src/client/datascience/jupyter/kernels/kernelProvider.ts
IanMatthewHuff: Essentially, if you are doing kernelspec launches, then to compare if you are already running the specified kernelspec just look at the kernelspec, not the interpreter that you launched or would launch it with.

src/client/datascience/jupyter/kernels/kernelProvider.ts
DonJayamanne: Cant we use `getKernelConnectionId` when comparing the kernel connection metadata instead of a deep compare

DonJayamanne: i.e. `getKernelConnectionId(existingKernelInfo.options.metadata) === getKernelConnectionId(options.metadata`

This way it will handle comparing of connections that deal with live kernels (we only compare the `metata.kernelModel` property

DonJayamanne: This PR might also be required/handy https://github.com/microsoft/vscode-jupyter/pull/4192/files

DonJayamanne (changes requested):

src/client/datascience/jupyter/kernels/kernelProvider.ts
DonJayamanne: > When you launch via a kernelspec that doesn't have an associated interpreter it just fills in the interpreter in the metadata with the current active interpreter. So if you change the active interpreter you can now create a mismatch here and the kernel restarts.

Got it, thanks
I don' think my suggestion works in this case.

DonJayamanne (approved):

src/client/datascience/jupyter/kernels/kernelProvider.ts
DonJayamanne: Added a bug for https://github.com/microsoft/vscode-jupyter/issues/4194 where it doesn't work in this case. Ideally we should be able to use that, if not then the function is not doing what its meant to do.

PR #4193, opened 2020-12-09
Fixes to comparison check of kernel connections

DonJayamanne:
While working on the duplicate kernel spec issue found an issue.
Might also be useful for this PR https://github.com/microsoft/vscode-jupyter/pull/4192/files

### DIFF
--- a/src/client/datascience/ipywidgets/cdnWidgetScriptSourceProvider.ts
+++ b/src/client/datascience/ipywidgets/cdnWidgetScriptSourceProvider.ts
@@ -105,6 +105,12 @@ export class CDNWidgetScriptSourceProvider implements IWidgetScriptSourceProvide
                 // Then get the first one that returns.
                 tempFile = await this.downloadFastestCDN(moduleName, moduleVersion);
                 if (tempFile) {
+                    traceInfo(
+                        !!process.env.VSC_JUPYTER_FORCE_LOGGING,
+                        `Widget Script downloaded for ${moduleName}:${moduleVersion}, already downloaded ${await this.fs.localFileExists(
+                            diskPath
+                        )}`
+                    );
                     // Need to copy from the temporary file to our real file (note: VSC filesystem fails to copy so just use straight file system)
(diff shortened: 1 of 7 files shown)


rchiodo (approved):

IanMatthewHuff (approved):

DavidKutu (approved):

src/test/datascience/uiTests/helpers.ts
```diff
@@ -141,6 +141,14 @@ export class BaseWebUI implements IAsyncDisposable {
         await this.browser.newContext();
         this.page = await this.browser.newPage();
         await this.page.goto(url);
+        this.page.on('console', (output) => {
+            try {
+                // Log output for diagnostic purposes (see what errors & the like are printed in console).
+                console.info(`IPyWidgets Browser: ${output.type()} ${output.text()}`);
```
DonJayamanne, line 147: @joyceerhl added some logging so we can see what's going on in the browser side (i.e look at errors & the like logged to the browser console).
Hopefully useful for diagnosing issues (can remove later when we have addressed flakyness of ipywidget tests)

src/client/datascience/ipywidgets/cdnWidgetScriptSourceProvider.ts
```diff
@@ -81,53 +81,59 @@ export class CDNWidgetScriptSourceProvider implements IWidgetScriptSourceProvide
         this.cache.clear();
     }
     public async getWidgetScriptSource(moduleName: string, moduleVersion: string): Promise<WidgetScriptSource> {
+        // First see if we already have it downloaded.
```
DonJayamanne, line 84: I believe this solves the issue with ipywidgets (or at least one of them).
The problem was we were download the widget file twice.
First time we download its all good.
Then IPyWidgets starts to fetch the file from the webview.
While webview is fetching the file from disc (that was already downloaded), 
- the extension is downloading the file and overwriting the local disc file again

Thus, we end up overwriting a file that is being read by another process.
& the ipywidget webview then fails to download the file & kaboom...

PR #4197, opened 2020-12-10
Fix race condition in download & copy of scripts

DonJayamanne:
Fixes part of the flakyness of ipywidget tests

### DIFF
--- a/src/client/datascience/ipywidgets/cdnWidgetScriptSourceProvider.ts
+++ b/src/client/datascience/ipywidgets/cdnWidgetScriptSourceProvider.ts
@@ -111,12 +111,7 @@ export class CDNWidgetScriptSourceProvider implements IWidgetScriptSourceProvide
             // Then get the first one that returns.
             tempFile = await this.downloadFastestCDN(moduleName, moduleVersion);
             if (tempFile) {
-                traceInfo(
-                    !!process.env.VSC_JUPYTER_FORCE_LOGGING,
-                    `Widget Script downloaded for ${moduleName}:${moduleVersion}, already downloaded ${await this.fs.localFileExists(
-                        diskPath
-                    )}`
-                );
+                traceInfo(`Widget Script downloaded for ${moduleName}:${moduleVersion}`);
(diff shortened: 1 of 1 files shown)


src/client/datascience/ipywidgets/cdnWidgetScriptSourceProvider.ts
```diff
@@ -81,53 +81,54 @@ export class CDNWidgetScriptSourceProvider implements IWidgetScriptSourceProvide
...
+        try {
+            // Make sure the disk path directory exists. We'll be downloading it to there.
+            await this.fs.createLocalDirectory(path.dirname(diskPath));
+
+            // Then get the first one that returns.
+            tempFile = await this.downloadFastestCDN(moduleName, moduleVersion);
+            if (tempFile) {
+                traceInfo(`Widget Script downloaded for ${moduleName}:${moduleVersion}`);
+                // Need to copy from the temporary file to our real file (note: VSC filesystem fails to copy so just use straight file system)
+                await this.fs.copyLocal(tempFile.filePath, diskPath);
```
DonJayamanne, line 116: Previously we'd have two downloads & the second download overwriting the existing file.
While overwriting the existing file, the webview is busy fetching the file from disc (which was downloaded the first time).

Thus we end up corrupting the file (ie writing to a file while its being read by another process)

Solution - ensure we download & copy once (ensure we cache the promise, old cache implementation was not right).

DonJayamanne: FYI - This was a genuine bug (overwriting existing file & corrupting it), not a flaky test.

DonJayamanne: FYI - This was a genuine bug (overwriting existing file & corrupting it), not a flaky test.

rchiodo: I'm not seeing the change? Were is the cache of the promise?

PR #4204, opened 2020-12-10
Review & fix ipywidget script download issues

DonJayamanne:
Fixes part of the flakyness of ipywidget tests

I was testing the fix in another PR & then accidentally ended up merging those changes without getting it reviewed (please see this comment for the bug that was fixed https://github.com/microsoft/vscode-jupyter/pull/4204#discussion_r540441023)

Realised my mistake when i created a PR to fix this (https://github.com/microsoft/vscode-jupyter/pull/4197)
Its now in main & would like to get it reviewed.

Please review the method `getWidgetScriptSource` & `getWidgetScriptSourceImplementation`.
Old code can be found here https://github.com/microsoft/vscode-jupyter/blob/b003bc5805bd0cefb7d27b84718aa180ff6e753f/src/client/datascience/ipywidgets/cdnWidgetScriptSourceProvider.ts#L83-L132

### DIFF
--- a/src/client/datascience/ipywidgets/cdnWidgetScriptSourceProvider.ts
+++ b/src/client/datascience/ipywidgets/cdnWidgetScriptSourceProvider.ts
@@ -88,7 +88,7 @@ export class CDNWidgetScriptSourceProvider implements IWidgetScriptSourceProvide
         }
         return this.cache.get(key)!;
     }
-    public async getWidgetScriptSourceImplementation(
+    private async getWidgetScriptSourceImplementation(
         moduleName: string,
         moduleVersion: string
     ): Promise<WidgetScriptSource> {


src/client/datascience/ipywidgets/cdnWidgetScriptSourceProvider.ts
```diff
@@ -88,7 +88,7 @@ export class CDNWidgetScriptSourceProvider implements IWidgetScriptSourceProvide
         }
         return this.cache.get(key)!;
     }
-    public async getWidgetScriptSourceImplementation(
+    private async getWidgetScriptSourceImplementation(
```
DonJayamanne, line 91: Previously we'd have two downloads & the second download overwriting the existing file.
While overwriting the existing file, the webview is busy fetching the file from disc (which was downloaded the first time).

Thus we end up corrupting the file (ie writing to a file while its being read by another process)

Solution - ensure we download & copy once (ensure we cache the promise, old cache implementation was not right).

DonJayamanne: FYI - This was a genuine bug (overwriting existing file & corrupting it), not a flaky test.

rchiodo: Did you make too many PRs or something? How is making this private fixing the issue you described?

DonJayamanne: > Did you make too many PRs or something?

I added some logging in the original PR https://github.com/microsoft/vscode-jupyter/pull/4193
Then the PR failed & looked at the logs.
Submitted a fix in the same PR to check the hypothesis..
The tests passed & I thought i reverted that commit (
Merged that PR

Created a new PR to fix the ipywidgets, & found that the code was already in master.
Thats when I learnt that I had not reverted the change (& ended up sending that commit in aother PR).

That code never got reviewed & i'd like to ensure that code is reviewed.
The change was to the method `getWidgetScriptSoruce` & adding new method `getWidgetScriptSourceImplemtnation`

Here's that commit that introduced the change (was only meant for testing) https://github.com/microsoft/vscode-jupyter/pull/4193/files#diff-d00c71b9e8b082a4eb1152a9163bf7840cb9b9053b60d1e0f6622fd26d8e06d4

Hope this makes sense.

DonJayamanne: I can revert that change & submit a new PR if that makes it simpler.

IanMatthewHuff: Change in the other merge looked good to me.

IanMatthewHuff (approved):

Issue #4203, opened 2020-12-10
%matplotlib widget infinite loop and running


eskemojoe007:
## Environment data

-   VS Code version: 1.51.1
-   Jupyter Extension version (available under the Extensions sidebar): 2020.12.411183155
-   Python Extension version (available under the Extensions sidebar): 2020.11.371526539
-   OS (Windows | Mac | Linux distro) and version: Windows 10 Enterprise
-   Python and/or Anaconda version: Python 3.8.6
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A
-   Jupyter server running: Local
- ipympl version: 0.5.8 (with restart already applied)

## Expected behaviour

Using `ipympl` and `%matplotlib widget` to plot within a widget for zoom and other functionality from a standar python file using sections such as `# %%`

## Actual behaviour

Infinite executuion of the cell
![image](https://user-images.githubusercontent.com/22135005/101817845-04621c00-3af1-11eb-8180-f8a93f152782.png)

## Steps to reproduce:

main.py
```py
# %%
import matplotlib.pyplot as plt
%matplotlib widget
plt.plot([1,2,3,4],[1,2,3,4])
```
Running with `jupyter: run current cell`

_Note: This works when in a native notebook, just not in a python standard python file_

## Logs

<details>

<summary>Output for <code>Jupyter</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Jupyter</code>)
</summary>

<p>

```
> ~\AppData\Local\Programs\Python\Python38\python.exe c:\Users\USERNAME\.vscode\extensions\ms-toolsai.jupyter-2020.12.411183155\pythonFiles\pyvsc-run-isolated.py -c "import jupyter"
> ~\AppData\Local\Programs\Python\Python38\python.exe c:\Users\USERNAME\.vscode\extensions\ms-toolsai.jupyter-2020.12.411183155\pythonFiles\pyvsc-run-isolated.py -c "import jupyter"
> ~\AppData\Local\Programs\Python\Python38\python.exe c:\Users\USERNAME\.vscode\extensions\ms-toolsai.jupyter-2020.12.411183155\pythonFiles\pyvsc-run-isolated.py -c "import notebook"
[...]

PR #4208, opened 2020-12-10
Fix support for IPyWidgets in Interactive Window

DonJayamanne:
For #4203

### DIFF
--- a/src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
+++ b/src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
@@ -39,7 +39,7 @@ export class NotebookIPyWidgetCoordinator implements INotebookKernelResolver {
         let promise = this.messageCoordinators.get(document.uri.toString());
         if (!promise) {
             const coordinator = CommonMessageCoordinator.create(document.uri, this.serviceContainer);
-            const promise = coordinator.initialize().then(() => coordinator);
+            promise = coordinator.initialize().then(() => coordinator);
             this.messageCoordinators.set(document.uri.toString(), promise);
         }
         return Cancellation.race(() => promise!.then(this.attachCoordinator.bind(this, document, webview)), token);
--- a/src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
+++ b/src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
@@ -38,7 +38,8 @@ export class NotebookIPyWidgetCoordinator implements INotebookKernelResolver {
         // entire VS code session, we have a map of notebook document to message coordinator
         let promise = this.messageCoordinators.get(document.uri.toString());
         if (!promise) {
-            promise = CommonMessageCoordinator.create(document.uri, this.serviceContainer);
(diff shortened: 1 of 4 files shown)


src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
```diff
@@ -38,7 +38,8 @@ export class NotebookIPyWidgetCoordinator implements INotebookKernelResolver {
         // entire VS code session, we have a map of notebook document to message coordinator
         let promise = this.messageCoordinators.get(document.uri.toString());
         if (!promise) {
-            promise = CommonMessageCoordinator.create(document.uri, this.serviceContainer);
+            const coordinator = CommonMessageCoordinator.create(document.uri, this.serviceContainer);
+            const promise = coordinator.initialize().then(() => coordinator);
```
rchiodo, line 42: We should test native and old webviews just to make sure this still works.

src/client/datascience/ipywidgets/webviewIPyWidgetCoordinator.ts
```diff
@@ -71,5 +71,6 @@ export class WebviewIPyWidgetCoordinator implements IInteractiveWindowListener {
                 this.postEmitter.fire(e);
             }
         });
+        this.messageCoordinator.initialize();
```
rchiodo, line 74: I believe this needs to be a return
```suggestion
        return this.messageCoordinator.initialize();
```

rchiodo (changes requested):
🌮

src/client/datascience/ipywidgets/webviewIPyWidgetCoordinator.ts
DonJayamanne: Oops, didn't see as a suggestion, fixed

rchiodo (approved):

src/client/datascience/ipywidgets/webviewIPyWidgetCoordinator.ts
```diff
@@ -62,7 +62,7 @@ export class WebviewIPyWidgetCoordinator implements IInteractiveWindowListener {
         // There should be an instance of the WebviewMessageCoordinator per notebook webview or interactive window. Create
         // the message coordinator as soon as we're sure what notebook we're in.
         this.notebookIdentity = args.resource;
-        this.messageCoordinator = await CommonMessageCoordinator.create(this.notebookIdentity, this.serviceContainer);
+        this.messageCoordinator = CommonMessageCoordinator.create(this.notebookIdentity, this.serviceContainer);
```
Victor-Savu, line 65: It seems that the reason for this change was to allow for a specialization of the `postEmitter.event` of the `CommonMessageCoordinator` before the initialization takes place.

To avoid making `initialize` public and relying on end-user cooperation, I would recommend modifying the factory method `create` to take an optional parameter that allows the caller to inject the `postEmitter`.

Victor-Savu (commented):
Proposing a way to avoid the split between the `create` and `initialize`.

Victor-Savu:
For convenience, I opened #4273 addressing these comments.

PR #4221, opened 2020-12-12
Fixes to coverage and generate HTML report

DonJayamanne:
* Generate HTML coverage report (optional)
* Fixes
	* Ensure coverage code runs first (we need to instrument & setup hooks before loading any of the other modules. I.e. src/client/**/*.ts files cannot be loaded until we have setup nyc.
	* Hence moved nyc to the very top so its obvious.
	* Also ensured we wait for report to finish (its async)

### DIFF
--- a/src/test/index.ts
+++ b/src/test/index.ts
@@ -10,6 +10,10 @@ require('./common/exitCIAfterTestReporter');
 if ((Reflect as any).metadata === undefined) {
     require('reflect-metadata');
 }
+// Always place at top, must be done before we import any of the files from src/client folder.
+// We need to ensure nyc gets a change to setup necessary hooks before files are loaded.
+const { setupCoverage } = require('./coverage');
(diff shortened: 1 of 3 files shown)


src/test/index.ts
```diff
@@ -10,6 +10,10 @@ require('./common/exitCIAfterTestReporter');
 if ((Reflect as any).metadata === undefined) {
     require('reflect-metadata');
 }
+// Always place at top, must be done before we import any of the files from src/client folder.
+// We need to ensure nyc gets a change to setup necessary hooks before files are loaded.
+const { setupCoverage } = require('./coverage');
+const nyc = setupCoverage();
```
DonJayamanne, line 16: Using `raw` require instead of `import` to ensure this is on top just as we do with `reflect-metadata`.

DavidKutu (approved):

rchiodo (approved):

Issue #4010, opened 2020-11-23
Native Notebooks - Add Run All Above/Run All Below Icons


claudiaregio:
Add run all above and run all below to global notebook toolbar. 
No changes in functionality, functions same as stable implementation.

joyceerhl:
Validated

claudiaregio:
Can we have the run all above/below icons to the right of the run all icon in the notebook toolbar? 

![image](https://user-images.githubusercontent.com/19276747/105642724-3d0b4a00-5e40-11eb-887c-98a0562be53f.png)

Just like the old implementation
![image](https://user-images.githubusercontent.com/19276747/105642732-4bf1fc80-5e40-11eb-9c78-ed6a5d8cbfe5.png)

PR #4240, opened 2020-12-16
add run above and run below to native notebooks

DavidKutu:
For #4010

### DIFF
--- a/package.json
+++ b/package.json
@@ -280,7 +280,7 @@
                 "enablement": "notebookViewType == jupyter-notebook && jupyter.isnotebooktrusted"
             },
             {
-                "command": "jupyter.notebookEditor.runallcellsabove",
+                "command": "jupyter.notebookeditor.runallcellsabove",
                 "title": "%DataScience.runAbove%",
                 "category": "Jupyter",
                 "icon": {
@@ -290,7 +290,7 @@
                 "enablement": "notebookViewType == jupyter-notebook && jupyter.isnotebooktrusted"
             },
             {
-                "command": "jupyter.notebookEditor.runcellandallbelow",
+                "command": "jupyter.notebookeditor.runcellandallbelow",
                 "title": "%DataScience.runBelow%",
                 "category": "Jupyter",
                 "icon": {
@@ -754,13 +754,13 @@
             ],
             "editor/title": [
                 {
-                    "command": "jupyter.notebookEditor.runallcellsabove",
+                    "command": "jupyter.notebookeditor.runallcellsabove",
                     "title": "%DataScience.runAbove%",
                     "group": "navigation",
                     "when": "resourceLangId == jupyter && notebookViewType == 'jupyter-notebook' && jupyter.isnotebooktrusted && jupyter.havenativecells"
                 },
                 {
-                    "command": "jupyter.notebookEditor.runcellandallbelow",
+                    "command": "jupyter.notebookeditor.runcellandallbelow",
                     "title": "%DataScience.runBelow%",
                     "group": "navigation",
                     "when": "resourceLangId == jupyter && notebookViewType == 'jupyter-notebook' && jupyter.isnotebooktrusted && jupyter.havenativecells"
--- a/package.json
+++ b/package.json
@@ -280,6 +280,26 @@
                 "enablement": "notebookViewType == jupyter-notebook && jupyter.isnotebooktrusted"
             },
             {
+                "command": "jupyter.notebookEditor.runallcellsabove",
+                "title": "%DataScience.runAbove%",
+                "category": "Jupyter",
+                "icon": {
+                    "light": "resources/light/runabove.svg",
+                    "dark": "resources/dark/runabove-inverse.svg"
+                },
+                "enablement": "notebookViewType == jupyter-notebook && jupyter.isnotebooktrusted"
+            },
+            {
+                "command": "jupyter.notebookEditor.runcellandallbelow",
+                "title": "%DataScience.runBelow%",
+                "category": "Jupyter",
+                "icon": {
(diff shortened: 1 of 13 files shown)


src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -94,6 +94,14 @@ export class NotebookEditor implements INotebookEditor {
...
         );
         disposables.push(model.onDidDispose(this._closed.fire.bind(this._closed, this)));
+        disposables.push(
+            commandManager.registerCommand(Commands.NativeNotebookRunAllCellsAbove, (uri) => this.runAbove(uri))
+        );
+        disposables.push(
+            commandManager.registerCommand(Commands.NativeNotebookRunCellAndAllBelow, (uri) =>
+                this.runCellAndBelow(uri)
+            )
+        );
```
DavidKutu, line 104: In old notebooks, the buttons are enabled only when the user selects a cell. In this case I don't know if we can do that.
We can create the context for it, but there's no event to keep it updated.

package.json
```diff
@@ -279,6 +279,26 @@
                 },
                 "enablement": "notebookViewType == jupyter-notebook && jupyter.isnotebooktrusted"
             },
+            {
+                "command": "jupyter.notebookEditor.runallcellsabove",
```
IanMatthewHuff, line 283: From what I can see we don't camelCase any other command names. So should keep "notebookeditor" here for consistancy.

package.nls.json
```diff
@@ -394,6 +394,8 @@
     "DataScience.insertAbove": "Insert cell above",
     "DataScience.addCell": "Add cell",
     "DataScience.runAll": "Run all cells",
+    "DataScience.runAbove": "Run cells above",
```
IanMatthewHuff, line 397: We already have some command titles from the old code. Do any of the old strings work here so that we don't need to add new ones?

src/client/datascience/constants.ts
```diff
@@ -49,6 +49,8 @@ export namespace Commands {
     export const RunAllCells = 'jupyter.runallcells';
     export const RunAllCellsAbove = 'jupyter.runallcellsabove';
     export const RunCellAndAllBelow = 'jupyter.runcellandallbelow';
+    export const NativeNotebookRunAllCellsAbove = 'jupyter.notebookEditor.runallcellsabove';
```
IanMatthewHuff, line 52: notebookEditor change to notebookeditor here as well.

src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -256,6 +264,52 @@ export class NotebookEditor implements INotebookEditor {
...
+    }
+
+    private getSelectedCellId(uri: Uri): string {
+        const editor = this.vscodeNotebook.notebookEditors.find((nb) => nb.document.uri.toString() === uri.toString());
+
+        if (editor && editor.selection) {
+            return editor.selection.uri.toString();
+        }
+
+        return '';
```
IanMatthewHuff, line 295: I'd make this function return string or undefined. While '' can work as a "not found value" the undefined make it more clear that you have something valid any string is returned and something not valid if undefined. Even if both empty string and undefined are falsy.

src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -256,6 +264,52 @@ export class NotebookEditor implements INotebookEditor {
...
+    private runCellRange(cells: NotebookCell[]) {
+        const kernel = this.kernelProvider.get(this.file);
+
+        if (!kernel || this.restartingKernel) {
+            return;
+        }
+
+        cells.forEach(async (cell) => {
+            // 2 means code cell
+            if (cell.cellKind === 2) {
```
IanMatthewHuff, line 307: You can use the enum here I think:
vscodeNotebookEnums.CellKind.Code

IanMatthewHuff (changes requested):
David. Code looks solid here to me with a few small suggestions. I do think that we should see about adding in a .vscode. test here for this functionality. Hopefully isn't too hard to add, but the functionality here seems to need a test to go along with it.

package.nls.json
DavidKutu: We never had, the ones on toolbar.tsx for example were returning a default.

There are messages for the command palette:
"jupyter.command.jupyter.runallcellsabove.palette.title": "Run Cells Above Current Cell",
"jupyter.command.jupyter.runcurrentcellandallbelow.palette.title": "Run Current Cell and Below",

Should I use those instead? They're a bit different

IanMatthewHuff: Hmmm, maybe just leave as is then. Keeps the native notebook stuff separate.

joyceerhl (approved):
Looks good. A vscode test could be helpful here.

src/test/datascience/notebook/notebookEditor.vscode.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        await waitForExecutionCompletedSuccessfully(thirdCell);
+
+        // The third cell should have a runState of Success
+        assert.strictEqual(thirdCell?.metadata.runState, vscodeNotebookEnums.NotebookCellRunState.Success);
+    });
+
+    test('Run cells above', async function () {
+        // This test is skipped because there is no way of selecting a cell in this context
+        // since by default the first cell is selected nothing happens when running all cells above
+        this.skip();
```
IanMatthewHuff, line 76: Looks good. I would just file the issue and then put the issue number here in the comment. That way when the test issue is resolved we know where to go to fix it. It's what we usually do for things like this.

IanMatthewHuff (approved):

Issue #238, opened 2020-10-16
IPyWidget tests randomly fail


rchiodo:
## Environment data

-   VS Code version: XXX
-   Extension version (available under the Extensions sidebar): XXX
-   Python Extension version (available under the Extensions sidebar): XXX
-   OS (Windows | Mac | Linux distro) and version: XXX
-   Python and/or Anaconda version: XXX
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): XXX
-   Jupyter server running: Local | Remote | N/A

## Expected behaviour

XXX

## Actual behaviour

XXX

## Steps to reproduce:

[**NOTE**: Self-contained, minimal reproducing code samples are **extremely** helpful and will expedite addressing your issue]

1. XXX

## Logs

<details>

<summary>Output for <code>Python</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Python</code>)
</summary>

<p>

```
XXX
```

</p>
</details>

rchiodo:
This is still happening. My current theory is that the execution fails because the correct cell is not selected.

DonJayamanne:
> This is still happening. My current theory is that the execution fails because the correct cell is not selected.

How about splitting the cells into separate notebooks & using the commands run all, might be less flaky, as we tests separte notebooks instead of output of individual files.

joyceerhl:
Was just looking at this as part of investigating https://github.com/microsoft/vscode-jupyter/issues/3955. If I manually run the ipyvolume notebook, the first cell actually does take a fair bit of time to run, so I'm not surprised that we sometimes timeout waiting for the cell to execute. I noticed that at least for the ipyvolume test, we have retry logic for the part where we verify the outputs, but we don't retry execution if that times out. I tried moving the cell execution into the retry for the ipyvolume test since that seems to be the one that times out the most often, if that helps and sounds like a good idea we can maybe consider doing this for the other tests.

DonJayamanne:
> g the cell execution into the retry for the ipyvolume test since that seems to be the one th

Helps to have a new pair of eyes looking into a problem..

joyceerhl:
Ah, I see now that the retry logic was there because we don't always know whether the widget has finished rendering. I still think moving the cell execution into the retry is a good idea since the output sometimes fails to show up, but might contribute to more tests timing out after the 3 min limit.

DonJayamanne:
[...]

PR #4273, opened 2020-12-30
Ensure exactly once initialization of CommonMessageCoordinator

Victor-Savu:
The `CommonMessageCoordinator` requires an asynchronous initialization
step. Since constructors cannot return promises, the initialization is
delegated to the asynchronous method `initialize`.
`WebviewIPyWidgetCoordinator` takes advantage of the two-step creation
(`constructor` + `initialize`) to use the
`CommonMessageCoordinator::postEmitter` member between construction and
initialization.

This pattern requires a few safeguards to ensure correct use of
`CommonMessageCoordinator` instances by preventing prevent accidental
errors such as:
- forgetting to call `initialize`
- calling `initialize` more than once
- using the object before initialization

Use the factory method pattern to create initialized instances of
the CommonMessageCoordinator to ensure the initialization always takes
place. Make the `initialize`method private to avoid double
initialization.

Allow the caller to customize the internal `postEmitter` (an
instance of `EventEmitter<{message: string; payload: any}>`) through
dependency injection in the  constructor and factory method. This way,
the `WebviewIPyWidgetCoordinator` can interact with the emitter without
waiting for the initialization to finish.

For #

### DIFF
--- a/src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
+++ b/src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
@@ -36,11 +36,12 @@ export class NotebookIPyWidgetCoordinator implements INotebookKernelResolver {
     ): Promise<void> {
         // Create a handler for this notebook if we don't already have one. Since there's one of the notebookMessageCoordinator's for the
         // entire VS code session, we have a map of notebook document to message coordinator
-        let promise = this.messageCoordinators.get(document.uri.toString());
+        const promise = this.messageCoordinators.get(document.uri.toString());
         if (!promise) {
-            const coordinator = CommonMessageCoordinator.create(document.uri, this.serviceContainer);
-            promise = coordinator.initialize().then(() => coordinator);
-            this.messageCoordinators.set(document.uri.toString(), promise);
+            this.messageCoordinators.set(
+                document.uri.toString(),
+                CommonMessageCoordinator.create(document.uri, this.serviceContainer)
+            );
         }
         return Cancellation.race(() => promise!.then(this.attachCoordinator.bind(this, document, webview)), token);
     }
--- a/src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
+++ b/src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
@@ -36,12 +36,10 @@ export class NotebookIPyWidgetCoordinator implements INotebookKernelResolver {
     ): Promise<void> {
         // Create a handler for this notebook if we don't already have one. Since there's one of the notebookMessageCoordinator's for the
         // entire VS code session, we have a map of notebook document to message coordinator
-        const promise = this.messageCoordinators.get(document.uri.toString());
-        if (!promise) {
-            this.messageCoordinators.set(
-                document.uri.toString(),
-                CommonMessageCoordinator.create(document.uri, this.serviceContainer)
-            );
+        let promise = this.messageCoordinators.get(document.uri.toString());
+        if (promise === undefined) {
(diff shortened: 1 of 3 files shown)


Victor-Savu:
I am not sure whether injecting the entire `postEmitter` (`EventEmitter<...>`) is a good idea. This design tries to solve the need for the `WebviewIPyWidgetCoordinator` to specialize how a certain type of message is dispatched, but it might offer too much freedom.
As with any other design decision, I would be happy to hear your feedback.

Victor-Savu:
Since this is only a refactoring, I did not add/change any tests and did not add a news entry. I am not sure how to add a `skip news` label to this PR, so I would appreciate your advice and help with that.

Victor-Savu:
I am not sure how to reproduce the failing test. When I try to run the UI functional tests, it skips most of them.

src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
```diff
@@ -36,11 +36,12 @@ export class NotebookIPyWidgetCoordinator implements INotebookKernelResolver {
     ): Promise<void> {
         // Create a handler for this notebook if we don't already have one. Since there's one of the notebookMessageCoordinator's for the
         // entire VS code session, we have a map of notebook document to message coordinator
-        let promise = this.messageCoordinators.get(document.uri.toString());
+        const promise = this.messageCoordinators.get(document.uri.toString());
```
rchiodo, line 39: You can end up in a situation here where the promise is not set on the return for line 46. I suggest you set it to the CommonMessageCoordinator.create value (instead of just sticking it into the map)

rchiodo:
Thanks for the refactor. It was like this before, but we had a problem with the order of the event emitters. Your refactor seems to handle that case though.

Just to be sure, can you try creating a widget in the interactive window?

This code would work:

```python
import ipywidgets as widgets
widgets.IntSlider()
```

It should show a slider in the interactive window.

rchiodo:
> I am not sure how to reproduce the failing test. When I try to run the UI functional tests, it skips most of them.
I ran the steps from the failing test by hand in debug mode and everything seemed to work fine.
I think I need some help to figure it out.

The bqplot failure can be ignored. Our ipywidget tests are flakey:
https://github.com/microsoft/vscode-jupyter/issues/238

src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
Victor-Savu: Thank you for spotting this! I completely overlooked the "create if missing" pattern here.

Victor-Savu:
> Just to be sure, can you try creating a widget in the interactive window?

This is what I got:
image

src/client/datascience/ipywidgets/notebookIPyWidgetCoordinator.ts
Victor-Savu: This might also explain the failing test, since this promise would never have been awaited.

rchiodo:
> > Just to be sure, can you try creating a widget in the interactive window?
> 
> This is what I got:
> image

That's actually in a notebook. The interactive window can be tested by running a cell from a python file.

Add the 'cell' comment at the top of the same code in a python file like so:
image

If you click Run cell, you should get something like so:
image

Victor-Savu:
Thank you for clarifying :) I am unfortunately not familiar with the nomenclature. 

Here is what I got
image

rchiodo (approved):

rchiodo:
Thanks looks good to me 👍

Victor-Savu:
Thank you so much for your review and for your patience :slightly_smiling_face:

IanMatthewHuff:
@Victor-Savu Thanks for the PR, much appreciated.

PR #4284, opened 2021-01-04
Variable Panel part 3

IanMatthewHuff:
For #

### DIFF
--- a/src/client/datascience/jupyter/kernels/cellExecution.ts
+++ b/src/client/datascience/jupyter/kernels/cellExecution.ts
@@ -500,7 +500,7 @@ export class CellExecution {
                 traceInfoIf(!!process.env.VSC_JUPYTER_LOG_KERNEL_OUTPUT, 'KernelMessage = ExecuteResult');
                 await this.handleExecuteResult(msg as KernelMessage.IExecuteResultMsg, clearState);
             } else if (jupyterLab.KernelMessage.isExecuteInputMsg(msg)) {
-                await this.handleExecuteInput(msg as KernelMessage.IExecuteInputMsg, clearState, loggers);
+                await this.handleExecuteInput(msg as KernelMessage.IExecuteInputMsg, clearState);
             } else if (jupyterLab.KernelMessage.isStatusMsg(msg)) {
                 traceInfoIf(!!process.env.VSC_JUPYTER_LOG_KERNEL_OUTPUT, 'KernelMessage = StatusMessage');
                 // Status is handled by the result promise. While it is running we are active. Otherwise we're stopped.
@@ -626,17 +626,9 @@ export class CellExecution {
         }
     }
 
-    private async handleExecuteInput(
-        msg: KernelMessage.IExecuteInputMsg,
-        _clearState: RefBool,
-        loggers: INotebookExecutionLogger[]
-    ) {
+    private async handleExecuteInput(msg: KernelMessage.IExecuteInputMsg, _clearState: RefBool) {
         if (msg.content.execution_count) {
             await updateCellExecutionCount(this.editor, this.cell, msg.content.execution_count);
-            // IANHU: Move this to the finalize section. In the initial request msg correctly reflect silent instead of just passing true
-            //loggers.forEach((l) =>
-            //l.postExecute(translateCellFromNative(this.cell), true, this.cell.language, this.cell.notebook.uri)
-            //);
         }
     }
 
--- a/src/client/datascience/jupyter/kernels/cellExecution.ts
+++ b/src/client/datascience/jupyter/kernels/cellExecution.ts
@@ -413,16 +413,13 @@ export class CellExecution {
 
         // For Jupyter requests, silent === don't output, while store_history === don't update execution count
         // https://jupyter-client.readthedocs.io/en/stable/api/client.html#jupyter_client.KernelClient.execute
-        // for our usage (such as variable requests) we do want output, but we don't want to update execution count
-        const executeSilently = false;
-
         const request = session.requestExecute(
             {
                 code,
                 silent: false,
                 stop_on_error: false,
                 allow_stdin: true,
-                store_history: !executeSilently
+                store_history: true
             },
             false,
             metadata
@@ -482,15 +479,13 @@ export class CellExecution {
             }
         } finally {
             // After execution log our post execute, regardless of success or failure
-            // Use our definition of silent here, not the silent on the request message as the
-            // meaning is different
+
+            // For our post execution logging we consider silent either silent execution or
+            // non-silent execution with store_history set to false
+            // Explicit false check as undefined store_history defaults to true if silent is false
+            const wasSilent = request.msg.content.silent || request.msg.content.store_history === false;
             loggers.forEach((l) =>
-                l.postExecute(
-                    translateCellFromNative(this.cell),
-                    executeSilently,
-                    this.cell.language,
-                    this.cell.notebook.uri
-                )
+                l.postExecute(translateCellFromNative(this.cell), wasSilent, this.cell.language, this.cell.notebook.uri)
             );
             cancelDisposable.dispose();
         }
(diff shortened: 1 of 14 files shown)


IanMatthewHuff:
This is the next stage of the variable panel work, and now has most of the functionality hooked up. Variables are now hooked up, variables update on execute and change when swapping between active documents. Also the data frame viewer is hooked up. 

There are two remaining planned chunks of work:
- [ ] Adding Tests
- [ ] UI clean up (there is still an open button for easy refresh testing, it's still trying to use existing height / width mechanisms instead of filling the view space)

I plan on tackling them in that order. Plus any other issue that come up. Already have one question on what we should do when changing active document to a editor that has not yet executed to one that has.

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -411,13 +411,18 @@ export class CellExecution {
             return this.completedSuccessfully().then(noop, noop);
         }
 
+        // For Jupyter requests, silent === don't output, while store_history === don't update execution count
+        // https://jupyter-client.readthedocs.io/en/stable/api/client.html#jupyter_client.KernelClient.execute
+        // for our usage (such as variable requests) we do want output, but we don't want to update execution count
+        const executeSilently = false;
```
IanMatthewHuff, line 417: I added this here as a constant. Basically I added this to keep separate our concept of execute silent (silent=false, store_history=false) from the jupyter concept which is silent = don't show output.

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -622,16 +638,9 @@ export class CellExecution {
         }
     }
 
-    private async handleExecuteInput(
-        msg: KernelMessage.IExecuteInputMsg,
-        _clearState: RefBool,
-        loggers: INotebookExecutionLogger[]
-    ) {
+    private async handleExecuteInput(msg: KernelMessage.IExecuteInputMsg, _clearState: RefBool) {
```
IanMatthewHuff, line 641: ExecuteInput reply is logged when Jupyter starts executing some code. It's the appropriate place to change execution count, but not the appropriate place to log a postExecute so I moved the logging up to the correct location.
https://jupyter-client.readthedocs.io/en/stable/messaging.html#code-inputs

IanMatthewHuff: Also removed the hardcoded slient here. Both of these issues might have been causing problems for anything looking at the loggers for use with native notebooks.

src/client/datascience/variablesView/variableView.ts
```diff
@@ -78,29 +113,62 @@ export class VariableView extends WebviewViewHost<IVariableViewPanelMapping> imp
...
+        if (
+            kernelStateEvent.state === KernelState.executed &&
+            kernelStateEvent.cell &&
+            kernelStateEvent.cell.metadata.executionOrder &&
+            kernelStateEvent.silent !== true
+        ) {
+            // We only want to update the variable view execution count when it's the active document executing
+            if (
+                this.notebookEditorProvider.activeEditor &&
+                this.notebookEditorProvider.activeEditor.file.toString() === kernelStateEvent.resource.toString()
```
rchiodo, line 161: Should you maybe do a arePathsSame here? File casing might not always be consistent.

rchiodo (approved):

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -411,13 +411,18 @@ export class CellExecution {
...
 
+        // For Jupyter requests, silent === don't output, while store_history === don't update execution count
+        // https://jupyter-client.readthedocs.io/en/stable/api/client.html#jupyter_client.KernelClient.execute
+        // for our usage (such as variable requests) we do want output, but we don't want to update execution count
+        const executeSilently = false;
+
         const request = session.requestExecute(
             {
                 code,
                 silent: false,
```
DonJayamanne, line 422: Shouldn't we then use `executeSilently` here.

src/client/datascience/jupyter/kernels/cellExecution.ts
DonJayamanne: Not a fan of negating another variable.
silent & store_history are two seprate flags.
However to we're not usign a single flag and negating it to use in the place of `store_history`. Feels confusing when we do that.
I.e. Why not have `storeHistory` as a variable, instead of creating `silent`. after all, Jupyter has two separate variables for that, why are we try to merge them into one & assume that store_history is the opposite of `silent` (that's the impression i get from this code).

DonJayamanne: I'd prefer to have two separate constants.

src/client/datascience/variablesView/variableView.ts
```diff
@@ -78,29 +113,62 @@ export class VariableView extends WebviewViewHost<IVariableViewPanelMapping> imp
...
+                const jupyterVariableDataProvider = await this.jupyterVariableDataProviderFactory.create(
+                    request.variable,
+                    this.notebookEditorProvider.activeEditor.notebook
+                );
+                const title: string = `${localize.DataScience.dataExplorerTitle()} - ${request.variable.name}`;
+                await this.dataViewerFactory.create(jupyterVariableDataProvider, title);
+            }
+        } catch (e) {
+            traceError(e);
+            this.appShell.showErrorMessage(e.toString());
```
DonJayamanne, line 134: I'd prefer to have a better error message than just a stringified error.
E.g. `Failed to display data viewer, please check logs for further info...`.

Should we optionally capture telemetry here.
Else we're expecting users to file bugs if somethign goes wrong.
With the telemtry we'd know how oftent this is failing for users.

src/client/datascience/variablesView/variableView.ts
DonJayamanne: Doesn't this mean this new variable viewer is hardcoded to only work with Native notebooks?
I'm ok with that, just feels like we're writing code that will be difficult to change later due to this tight coupling. I.e. I'd prefer to such coupling be outside this class, ie. reduce the coupling (**no need to change, just a thought**)

DonJayamanne (changes requested):
Blocking on error message & adding telemetry.
Others are optional

src/client/datascience/jupyter/kernels/cellExecution.ts
IanMatthewHuff: @DonJayamanne Hmmm, taking another look. I don't think that we need this here at all. I was adding it as I considered it a bit confusing as we have our own concept of silent in our code above the JMP level. Basically what we call silent corresponds to store_history, in all our old code we never actually set silent, we just translate silent to store_history at the execution level since we always wanted output. But since it's hard coded here I can just ditch the constant and have the logger base silent off of the store_history value of the request. That might work better if the constant was just introducing more confusion.

src/client/datascience/jupyter/kernels/cellExecution.ts
IanMatthewHuff: No, not here, because it's the mismatch between our concept of silent and the JMC message concept. But shouldn't matter as I'm going to change this per my above comment.

src/client/datascience/jupyter/kernels/cellExecution.ts
DonJayamanne: Got it, thanks for clarification.

src/client/datascience/variablesView/variableView.ts
IanMatthewHuff: Yeah, I'd just pulled this from the old code, but no time like now to fix it up. I'll change the message here and in the other location and add a telemetry point.

DonJayamanne (approved):

src/client/datascience/variablesView/variableView.ts
IanMatthewHuff: Good call I'll change it. Interestingly enough it looks like vscode has code to directly compare URIs but it doesn't look publicly accessible to me. Seems like a URI should be something that you would be able to compare without worrying about the concept of paths.

IanMatthewHuff: @DonJayamanne it is currently specced for native only. Lemme make a note on this for the next PR, might be good to remove that knowledge as I think that it would be nice to support the old webviews with the view based variable explorer as well.

IanMatthewHuff:
Only test failure was widgets test failure so merging.

Issue #4058, opened 2020-12-01
Cells recovered with the "z" shortcut are not saved


alanhett:
## Environment data

-   VS Code version: 1.51.1
-   Jupyter Extension version (available under the Extensions sidebar): v2020.11.392013122
-   Python Extension version (available under the Extensions sidebar): v2020.11.371526539
-   OS (Windows | Mac | Linux distro) and version: Windows 10
-   Python and/or Anaconda version: 3.8.6
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A
-   Jupyter server running: Local

## Expected behaviour

When I delete a cell, then bring it back using the "z" shortcut key, it should save this change and keep the recovered cell.

## Actual behaviour

When I delete a cell, then bring it back using the "z" shortcut key, then if I save the file and close and reopen it, the cell is permanently lost. It doesn't save the recovered cell.

## Steps to reproduce:

[**NOTE**: Self-contained, minimal reproducing code samples are **extremely** helpful and will expedite addressing your issue]

1. Create a new jupyter notebook and create a couple of cells, enter some arbitrary code
2. Delete a cell using the trash can icon at the right side of the cell
3. Recover the cell by pressing the "z" key (while no cells are selected)
4. Save the notebook and close it
5. Reopen the notebook, the cell is missing

IanMatthewHuff:
@alanhett Thanks for the report on this. I just gave it a try and I repro the same issue. When the cell comes back with 'z' the cell reappears, but the file doesn't correctly get marked as dirty and the change is not saved if you close. We'll triage this issue.

alanhett:
Thanks @IanMatthewHuff , by the way I also attempted to make the file "dirty"  by making some arbitrary changes w/in the cell and w/in other cells (like between steps 3 and 4 above), and it still would lose the cell.

IanMatthewHuff:
@alanhett Yeah it's reflecting that chance visually, but not in the backing data. We'll have to clean this up.

IanMatthewHuff:
Validated. Removed the cell and did an undo command. After that, the cell was back, file was not dirty and the cell was still there after opening and closing.

PR #4298, opened 2021-01-05
Fix issue with 'undo' not updating model

rchiodo:
For #4058 

Update the model when the user hits the undo or redo key sequence in the old editor
Also fix a problem in the custom editor where delete undos are not sent to the correct location


src/datascience-ui/native-editor/redux/reducers/creation.ts
```diff
@@ -100,8 +100,8 @@ export namespace Creation {
         if (position >= 0) {
             newList.splice(position, 0, newVM);
         } else {
-            newList.splice(0, 0, newVM);
-            position = 0;
+            newList.splice(newList.length, 0, newVM);
```
rchiodo, line 103: This fixes the problem in the custom editor with undo. When the 'above' cell isn't found it was inserting at the top instead of the bottom.

DonJayamanne (approved):

src/datascience-ui/native-editor/redux/reducers/creation.ts
IanMatthewHuff: @rchiodo Sorry wasn't fully getting this. insertAbove requires the vm value, right? Wouldn't the case where we can't find that vm be an error? Or if it's not an error I'm not sure why the default for insertAbove would be the end and not the start.

rchiodo: It doesn't require the VM value, but it did assume if it wasn't found it should insert at the top. The way the function is called though is it passes an empty value when it expects it to be inserted at the end.

I agree it's confusing but it was that or special case the delete to have a new insert. Maybe that would be better. It'd certainly make more sense given the name and the common assumption that it would insert above everything if it can't find the item to insert above.

rchiodo: But I can also see the argument that if I can't find what to insert above, why would I assume above everything? If there's an item in the list I should definitely be able to find it.

IanMatthewHuff: Ok, I get it a bit more now. The API is just a bit wonky for that function. I think that special casing the delete might be better if we are getting super strict. But at this point, especially with us moving away from the webviews, I'm perfectly fine with trusting your judgement on if that's worth the time.

IanMatthewHuff: So I'll approve.

IanMatthewHuff (approved):

src/datascience-ui/native-editor/redux/reducers/creation.ts
rchiodo: Thanks. I added a separate code path for delete outside the index.

Issue #4218, opened 2020-12-11
Hover Documentation Pop up does not work after VSCode 1.52 update


canersnli:
## Environment data

-   VS Code version: 1.52.0
-   Jupyter Extension version (available under the Extensions sidebar): v2020.12.414227025
-   Python Extension version (available under the Extensions sidebar): v2020.11.371526539
-   OS (Windows | Mac | Linux distro) and version: Windows10
-   Python and/or Anaconda version: 3.8.5
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): conda
-   Jupyter server running: Local | Remote | N/A  Local
-  Microsoft Language Server: Pylance v2020.12.2

## Expected behaviour
In ipynb files, on hover, showing documentation pop-up 

## Actual behaviour
Right after VSCode 1.52 update , on hover documentation pop-up stopped working, nothing pops up.  I have downgraded from 1.52 to 1.51.1 and bug has been solved. 

## Steps to reproduce:

[**NOTE**: Self-contained, minimal reproducing code samples are **extremely** helpful and will expedite addressing your issue]

## Logs

<details>

<summary>Output for <code>Jupyter</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Jupyter</code>)
</summary>

<p>

```
XXX
```

</p>
</details>

claudiaregio:
Thanks @canersnli , I am able to repro this in stable.

greazer:
Sounds like a regression.

luiztauffer:
so, it looks like #4299 was merged and possibly solved this? When is a patch release planned?

rchiodo:
Sometime this month I believe.

luiztauffer:
I found that I was having autocompletion error in Ipython and Jupyter and found this Issue: https://github.com/ipython/ipython/issues/12740

`pip install jedi==0.17.2` solved for me, also within VSCode

canersnli:
> I found that I was having autocompletion error in Ipython and Jupyter and found this Issue: [ipython/ipython#12740](https://github.com/ipython/ipython/issues/12740)
> 
> `pip install jedi==0.17.2` solved for me, also within VSCode

Thanks for the advice but it didn't work for me.

joyceerhl:
Validated with numpy documentation popup in debug bits.

canersnli:
[...]

PR #4299, opened 2021-01-05
Fix hover regression in webview editor

rchiodo:
For #4218 

Root cause was the structure returned by the language server did not show 'hasOwnProperty' as true even though the field was present.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -339,20 +339,6 @@ jobs:
       - name: Install dependencies (npm ci)
         run: npm ci --prefer-offline
 
-      - name: Install Dot.net
-        uses: actions/setup-dotnet@v1
-        with:
-          dotnet-version: ${{env.DOTNET_VERSION}}
-        if: matrix.test-suite != 'notebookWithoutPythonExt'
-
-      - name: Install .NET Interactive
-        run: dotnet tool install -g --add-source "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" Microsoft.dotnet-interactive
-        if: matrix.test-suite != 'notebookWithoutPythonExt'
-
-      - name: Install .NET Kernel
-        run: dotnet interactive jupyter install
-        if: matrix.test-suite != 'notebookWithoutPythonExt'
-
         # This step is slow.
       - name: Install Julia
         uses: julia-actions/setup-julia@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yml
```diff
@@ -339,20 +339,6 @@ jobs:
       - name: Install dependencies (npm ci)
         run: npm ci --prefer-offline
 
-      - name: Install Dot.net
```
rchiodo, line 342: This isn't necessary for functional tests and was intermittently failing.

DonJayamanne: I added them with the intention of keeping all actions consistently the same...
Else it was getting too difficult to figure out what was reqquired 
hate this copy paste nonense...

src/client/datascience/interactive-common/intellisense/conversion.ts
```diff
@@ -262,14 +262,16 @@ function convertToMonacoMarkdown(
         | vscode.MarkedString
         | vscode.MarkedString[]
 ): monacoEditor.IMarkdownString[] {
-    if (strings.hasOwnProperty('kind')) {
+    // tslint:disable-next-line: no-any
+    if (strings.hasOwnProperty('kind') || (strings as any).kind) {
```
DonJayamanne, line 266: The new condition is sufficient `(strings as any).kind`, won't need `hasOwnProperty`

DonJayamanne (approved):
Might as well remove the check for `hasOwnProperty`

src/client/datascience/interactive-common/intellisense/conversion.ts
rchiodo: That's not true as (strings as any).kind will resolve as false if the value is 0 or false. I want it to continue to work if the property happens to be there.

IanMatthewHuff: Doesn't 'in' keyword work here? 'kind' in strings.

rchiodo: Let me double check

rchiodo: Yes that seems to work. Not sure what the difference is. I assumed 'in' used hasOwnProperty under the covers.

IanMatthewHuff: @rchiodo Heh, I only know because I'd been going through a JS basics book (felt like there was stuff that I didn't know jumping directly into TS) and just read the chapter covering JS object. hasOwnProperty is only checking the top level object prototype for the property, so if it's inherited it won't show up. 'in' keywork check the entire property chain. So for an object hasOwnProperty('toString') return false, since that method is in the base object prototype.

IanMatthewHuff (approved):

Issue #3936, opened 2020-11-17
Renderer extension not being installed -- causes some packages not to work, like plotly.


greazer:

DonJayamanne:
Fixed by #3938

DonJayamanne:
**Testing:**
* Install Jupyter
* Open a notebook
* Run a cell with plotly or vega or the like
* Ensure the renderer displays the output
	* Confirm the renderer is used (in the drop down against mim types you'll see the name of the renderer extension)

hellmrf:
@DonJayamanne, I'm still having problems in the latest insider build.

![image](https://user-images.githubusercontent.com/25724069/99757515-4984c680-2ace-11eb-9fab-40cc37b8737b.png)

The LaTeX renderer seems to not be present, so I always need to select `text/plain` manually to see the ugly output.

thakkarparth007:
I'm unable to view Plotly plots in the new notebooks. This screenshot shows the problem and the version being used:

![image](https://user-images.githubusercontent.com/747423/100549317-49d44d00-3298-11eb-832f-c4adc56eb0cd.png)

No new updates were available when I tried checking for updates.

MichaelSteurer:
Same here. I run this as a quick workaround:
```python
import plotly.io as pio
pio.renderers.default = 'iframe_connected'
```

rchiodo:
This doesn't seem to have been fixed. I sometimes get the renderer extension and sometimes don't. If I uninstall the jupyter extension, reinstall, and then run a notebook with plotly in it, the renderer extension is not downloaded.

giammi56:
> Same here. I run this as a quick workaround:
> 
> ```python
> import plotly.io as pio
> pio.renderers.default = 'iframe_connected'
> ```

works good in Jupiter-notebook preview

rchiodo:
https://github.com/microsoft/vscode-jupyter/issues/4149 is an item tracking the work to fix this.

rchiodo:
Actually just removing #4149

andreamoro:
> ```python
> renderers.default = 'iframe_connected'
> ```

Do you have any corresponding for plotly express. I can't see the corresponding entry point in the library.

rchiodo:
Validated. Next release should have this fixed.

Issue #1423, opened 2020-06-12
Renderer extension for new Notebooks


DonJayamanne:
If and when we decide to publish a separate renderer extension.

https://github.com/microsoft/vscode-notebook-renderers

- [ ] Review Extension information
	- name, id, description, license
	- readme
	- contribution doc
- [ ] We need to update the renderer API in renderer extension as we did here: 
https://github.com/microsoft/vscode-python/pull/13665/files
Changes related to the following need to be applied to renderer extension:
	* [index.tsx](https://github.com/microsoft/vscode-python/pull/13665/files#diff-e9d196ad86eab236c4a11ac079541dc7)https://github.com/microsoft/vscode-python/pull/13665/files#diff-e9d196ad86eab236c4a11ac079541dc7
	* [package.json](https://github.com/microsoft/vscode-python/pull/13665/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2)https://github.com/microsoft/vscode-python/pull/13665/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

- [ ] After publishing extension, remove related renderer code from DS code (from Python extension)

* [ ] Component governance
* [ ] Version, publisher id, links to repo, issue link, (package.json)
* [ ] Third party notices file
* [ ] Usual MD files (lic, sec, contibuting, code of conduct, etc)
* [ ] Ship with VS Code core (as dependent extension?)
* [ ] Azo Project

PR #4303, opened 2021-01-06
Include renderer extension output during build steps

DonJayamanne:
For #1423 , #3936
Corresponding PR is here https://github.com/microsoft/vscode-notebook-renderers/pull/31

**Workflow:**
* Download the Renderer extension VSIX from blob store
* Extract into jupyter extension (during install & build)

### DIFF
--- a/build/ci/downloadRenderer.js
+++ b/build/ci/downloadRenderer.js
@@ -55,13 +55,16 @@ async function downloadRendererExtension() {
 
         console.log('Extracting Renderer extension');
         const extensionDir = path.join(downloadDir, 'extensionDir');
-        await unzip(downloadedFile, extensionDir)
+        await unzip(downloadedFile, extensionDir);
         console.log('Extracted Renderer extension');
 
-        await fs.copy(path.join(extensionDir, 'out', 'client_renderer'), path.join(ExtensionRootDir, 'out', 'client_renderer'));
+        await fs.copy(
+            path.join(extensionDir, 'out', 'client_renderer'),
+            path.join(ExtensionRootDir, 'out', 'client_renderer')
+        );
         console.log('Copied Renderer extension output.');
     } finally {
-        if (typeof cleanup == 'function'){
+        if (typeof cleanup == 'function') {
             try {
                 cleanup();
             } catch {
--- a/build/ci/downloadRenderer.js
+++ b/build/ci/downloadRenderer.js
@@ -1,0 +1,74 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+'use strict';
+
+const fs = require('fs-extra');
+const path = require('path');
+const tmp = require('tmp');
+const { ExtensionRootDir } = require('../constants');
+const download = require('download');
+const StreamZip = require('node-stream-zip');
+
+async function unzip(zipFile, targetFolder) {
+    await fs.ensureDir(targetFolder);
+    return new Promise((resolve, reject) => {
+        const zip = new StreamZip({
+            file: zipFile,
+            storeEntries: true
+        });
+        zip.on('ready', async () => {
+            zip.extract('extension', targetFolder, (err) => {
+                if (err) {
+                    reject(err);
+                } else {
+                    resolve();
+                }
+                zip.close();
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1737,6 +1738,31 @@
                 "mimeTypes": [
                     "application/vnd.jupyter.widget-view+json"
                 ]
+            },
+            {
+                "id": "jupyter-notebook-renderer",
+                "entrypoint": "./out/client_renderer/renderers.js",
```
DonJayamanne, line 1744: Copy output from renderer extension into jupyter extension & things just work.
During `npm install` & builds on CI we will download the renderer extension output & use the latest. 
This way it works for dev & in builds.

build/ci/postInstall.js
```diff
@@ -111,4 +114,5 @@ async function downloadBCryptGenRandomExecutable() {
     fixJupyterLabDTSFiles();
     createJupyterKernelWithoutSerialization();
     await downloadBCryptGenRandomExecutable();
+    await downloadRendererExtension();
```
DonJayamanne, line 117: This way it will work when using `npm install`

gulpfile.js
```diff
@@ -244,16 +245,20 @@ gulp.task('includeBCryptGenRandomExe', async () => {
     await fs.copyFile(src, dest);
 });
 
-gulp.task('prePublishBundle', gulp.series('includeBCryptGenRandomExe', 'webpack'));
+gulp.task('downloadRendererExtension', async () => {
+    await downloadRendererExtension();
+});
+
+gulp.task('prePublishBundle', gulp.series('includeBCryptGenRandomExe', 'downloadRendererExtension', 'webpack'));
```
DonJayamanne, line 252: this way it will work on CI or anyone using gulp to build locally 
(I don't build webviews locally, hence just use the vsc task to compile extesnion).

build/ci/downloadRenderer.js
```diff
@@ -0,0 +1,74 @@
...
+        }
+        await download(uri, downloadDir, { filename: extensionName });
+        console.log('Downloaded Renderer extension');
+
+        console.log('Extracting Renderer extension');
+        const extensionDir = path.join(downloadDir, 'extensionDir');
+        await unzip(downloadedFile, extensionDir)
+        console.log('Extracted Renderer extension');
+
+        await fs.copy(path.join(extensionDir, 'out', 'client_renderer'), path.join(ExtensionRootDir, 'out', 'client_renderer'));
```
rchiodo: The problem with this is that npm run clean will destroy the renderer. So if I clean my output the extension goes away.  Perhaps we can put this and the bcrypt thing (which I have the same problem with) in a different directory?

rchiodo: Or fix npm run clean to not delete it.

rchiodo (approved):

build/ci/downloadRenderer.js
joyceerhl: The bcrypt exe is downloaded to src and then copied to out on each compile for this reason. (It's not supposed to go missing. If it is that's a bug)

rchiodo: @joyceerhl I usually do this:

- git checkout main
- git pull
- git checkout -b rchiodo/<new branch>
- npm ci (this copies bcrypt to the out folder)
- npm run clean (this deletes it)

Building the source doesn't seem to recopy it as far as I can tell.

DonJayamanne: I would prefer to keep all output in the `out` folder.
I'll create a separate issue to address this for both renderer & bycrypt.

One option is to setup a VSC task (thats a depednency of the compile task) that'll check if this goes missing & then download them auto magically... (using nodemon, pretty standard in node to use nodemon for such things).

DonJayamanne: > Building the source doesn't seem to recopy it as far as I can tell.

@rchiodo How do you build the source, what command do you use? Gulp, VSC Task, `npm run compile` or other.

rchiodo: I use the build tasks in VS code.

IanMatthewHuff (approved):

Issue #4295, opened 2021-01-05
cant create jupyter notebook


tingmwu:
Issue Type: <b>Bug</b>

when I use ctrl+shift+p create a jupyter notebook, it shows "Create New Blank Jupyter Notebook' resulted in an error (command 'jupyter.createnewnotebook' not found)". 

Extension version: 2020.12.414227025
VS Code version: Code 1.49.3 (2af051012b66169dde0c4dfae3f5ef48f787ff69, 2020-10-02T17:57:01.595Z)
OS version: Windows_NT x64 10.0.19041

IanMatthewHuff:
@tingmwu Looks like you are on an older version of VS Code there. Could you please update VS Code to the newest version and see if you are still having the issue?

tingmwu:
As you suggested, I updated the VS code and all the plug-ins, and still showed “Command 'Jupyter: Create New Blank Jupyter Notebook' resulted in an error (command 'jupyter.createnewnotebook' not found)”. The reinstallation is still like this. my jupyter worked well without VS Code.



At 2021-01-06 01:21:05, "Ian Huff" <notifications@github.com> wrote:

@tingmwu Looks like you are on an older version of VS Code there. Could you please update VS Code to the newest version and see if you are still having the issue?

IanMatthewHuff:
@tingmwu Looks like we had an activation ordering issue here. The PR in progress above #4313 should resolve this issue and be in shortly.

rchiodo:
Validated. THis should be fixed in our next release.

PR #4313, opened 2021-01-06
Wait for DS code to activate as part of extension activation

DonJayamanne:
For #4295
@IanMatthewHuff  I ran into this issue this morning with a new workspace.
I think when DS extension activates first without any Python extension then you might be able to replicate this issue.
I had accidentally deleted the python extension (trying to remove an older version) & then ran into this this morning.

### DIFF
--- a/src/client/extensionActivation.ts
+++ b/src/client/extensionActivation.ts
@@ -129,13 +129,13 @@ async function activateLegacy(
     context.subscriptions.push(manager);
     const activationPromise = manager.activate();
 
-    // Activate data science features
+    // Activate data science features after base features.
     const dataScience = serviceManager.get<IDataScience>(IDataScience);
-    dataScience.activate().ignoreErrors();
+    const dsActivationPromise = dataScience.activate();
 
     const deprecationMgr = serviceContainer.get<IFeatureDeprecationManager>(IFeatureDeprecationManager);
     deprecationMgr.initialize();
     context.subscriptions.push(deprecationMgr);
 
-    return { activationPromise };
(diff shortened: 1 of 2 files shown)


src/client/extensionActivation.ts
```diff
@@ -129,13 +129,13 @@ async function activateLegacy(
     context.subscriptions.push(manager);
     const activationPromise = manager.activate();
 
-    // Activate data science features
+    // Activate data science features after base features.
     const dataScience = serviceManager.get<IDataScience>(IDataScience);
-    dataScience.activate().ignoreErrors();
```
DonJayamanne, line 134: This is the bug

src/client/extensionActivation.ts
```diff
@@ -129,13 +129,13 @@ async function activateLegacy(
...
+    const dsActivationPromise = dataScience.activate();
 
     const deprecationMgr = serviceContainer.get<IFeatureDeprecationManager>(IFeatureDeprecationManager);
     deprecationMgr.initialize();
     context.subscriptions.push(deprecationMgr);
 
-    return { activationPromise };
+    return { activationPromise: activationPromise.then(() => dsActivationPromise) };
```
DonJayamanne, line 140: chaining the promise as its used later to determine the total time take to complete all of this.

src/client/extensionActivation.ts
IanMatthewHuff: Interesting, so this has been broken since the split it looks like?

IanMatthewHuff (approved):

rchiodo (approved):

joyceerhl (approved):

PR #4319, opened 2021-01-07
Open variables view command and remove the test OPEN button from it

IanMatthewHuff:
Stalled a little bit on the test work, so checking in a couple of small tweaks that I made separate from the test work to keep things flowing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
         "onCommand:jupyter.exportfileandoutputasnotebook",
         "onCommand:jupyter.selectJupyterInterpreter",
         "onCommand:jupyter.selectjupytercommandline",
+        "onCommand:jupyter.openVariableView",
         "onNotebook:jupyter-notebook",
         "onCustomEditor:jupyter.notebook.ipynb"
     ],
@@ -697,6 +698,11 @@
             {
                 "command": "jupyter.clearSavedJupyterUris",
                 "title": "%jupyter.command.jupyter.clearSavedJupyterUris.title%",
+                "category": "Jupyter"
+            },
+            {
+                "command": "jupyter.openVariableView",
+                "title": "%jupyter.command.jupyter.openVariableView.title%",
                 "category": "Jupyter"
             }
         ],
--- a/src/client/common/application/commands.ts
+++ b/src/client/common/application/commands.ts
@@ -33,6 +33,7 @@ interface ICommandNameWithoutArgumentTypeMapping {
     ['editor.action.formatDocument']: [];
     ['editor.action.rename']: [];
     ['jupyter.selectJupyterInterpreter']: [];
+    ['jupyterViewVariables.focus']: [];
     [DSCommands.RunCurrentCell]: [];
     [DSCommands.RunCurrentCellAdvance]: [];
     [DSCommands.ExecSelectionInInteractiveWindow]: [];
@@ -58,6 +59,7 @@ interface ICommandNameWithoutArgumentTypeMapping {
     [DSCommands.CreateNewNotebook]: [];
     [DSCommands.EnableDebugLogging]: [];
     [DSCommands.ResetLoggingLevel]: [];
+    [DSCommands.OpenVariableView]: [];
 }
 
 /**
--- a/src/client/datascience/commands/commandRegistry.ts
+++ b/src/client/datascience/commands/commandRegistry.ts
@@ -116,6 +116,7 @@ export class CommandRegistry implements IDisposable {
             this.enableLoadingWidgetScriptsFromThirdParty
         );
         this.registerCommand(Commands.ClearSavedJupyterUris, this.clearJupyterUris);
+        this.registerCommand(Commands.OpenVariableView, this.openVariableView);
         if (this.commandListeners) {
             this.commandListeners.forEach((listener: IDataScienceCommandListener) => {
                 listener.register(this.commandManager);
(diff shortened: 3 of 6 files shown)


src/client/common/application/commands.ts
```diff
@@ -58,6 +59,7 @@ interface ICommandNameWithoutArgumentTypeMapping {
     [DSCommands.CreateNewNotebook]: [];
     [DSCommands.EnableDebugLogging]: [];
     [DSCommands.ResetLoggingLevel]: [];
+    [DSCommands.OpenVariableView]: [];
```
IanMatthewHuff, line 62: @rchiodo I like the idea of hooking this up to a button on the native notebook toolbar. @claudiaregio thoughts on this? I think I just need an icon to do that.

src/client/datascience/commands/commandRegistry.ts
```diff
@@ -487,6 +488,10 @@ export class CommandRegistry implements IDisposable {
         env.openExternal(Uri.parse(`https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter`));
     }
 
+    // Open up our variable viewer using the command that VS Code provides for this
+    private async openVariableView(): Promise<void> {
+        this.commandManager.executeCommand('jupyterViewVariables.focus');
```
IanMatthewHuff, line 493: I actually took a bit of time working on this looking at how VS Code handles view / opening commands. I was struggling a bit as the View: Open View command didn't take a parameter, it just always pops up a selector. Was expecting to use that command with my view name. Digging into the code I found that VS code actually creates focus commands for all views that are contributed, which I didn't realize. I'm using that command to do the focus, but I still want our own command so that we can find it in the menus under the Jupyter titles instead of in the views section.

src/datascience-ui/variable-view/variableViewPanel.tsx
```diff
@@ -48,7 +51,6 @@ export class VariableViewPanel extends React.Component<IVariableViewPanelProps>
         // we can size and host it differently from the variable panel in the interactive window or native editor
         return (
             <div id="variable-view-main-panel" role="Main" style={dynamicFont}>
-                <button onClick={this.props.toggleVariableExplorer}>OPEN</button>
```
IanMatthewHuff, line 51: Was just using this for testing and hadn't gotten around to removing it until now.

src/client/datascience/commands/commandRegistry.ts
```diff
@@ -487,6 +488,10 @@ export class CommandRegistry implements IDisposable {
         env.openExternal(Uri.parse(`https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter`));
     }
 
+    // Open up our variable viewer using the command that VS Code provides for this
```
rchiodo, line 491: The connection between 'jupyterViewVariables.focus' and the actual webview opening is hard to find. Could you add some more comments here explaining it?

IanMatthewHuff: Yeah, I actually found that command hard to find. I'll add a comment into the code here.

src/datascience-ui/variable-view/variableViewPanel.tsx
```diff
@@ -22,6 +22,9 @@ export class VariableViewPanel extends React.Component<IVariableViewPanelProps>
 
     constructor(props: IVariableViewPanelProps) {
         super(props);
+
+        // For the variable view we want to start toggled open
```
rchiodo, line 26: When would the variable view ever want to be closed? Or does closed mean it's not visible? (I'm wondering if it still supports the 'hide/collapse' we have in the other view)

rchiodo (approved):

DavidKutu (approved):

src/datascience-ui/variable-view/variableViewPanel.tsx
IanMatthewHuff: When I was implementing this I found the simplest way to do it was to just lift the state and control as directly as possible from its usage in the Interactive Window and old webviews. So it does still have the toggle command available on it. However that toggle command was only triggered by a control on the IW or notebook webview, so it's not accessible for this version of the control.

package.json
```diff
@@ -698,6 +699,11 @@
                 "command": "jupyter.clearSavedJupyterUris",
                 "title": "%jupyter.command.jupyter.clearSavedJupyterUris.title%",
                 "category": "Jupyter"
+            },
+            {
+                "command": "jupyter.openVariableView",
```
DonJayamanne, line 704: Can we show (enable) this command only when a native notebook is open.

DonJayamanne: Ignore that

src/client/datascience/commands/commandRegistry.ts
DonJayamanne: I think we need an away for `executeCommand`, wait for it to complete.

src/datascience-ui/variable-view/variableViewPanel.tsx
DonJayamanne: Don't we need to localize the text `OPEN`? Can't remember what we do with such text in our webviews.

DonJayamanne (approved):

IanMatthewHuff:
@DonJayamanne For the `OPEN` text if you see my comment in the PR that was just a button that I was using for testing, it's just being removed at this point. 

Per your point on the executeCommand (honestly because I'm not 100% sure myself) when we just have a single awaitable function call returning from an async function, do we need to await on it? Or do we just return the result of the call? In this file I actually see us doing three different things.

debugCell returns the promise. debugStepOver uses .executeCommand and doesn't return or await the result. And createNewNotebook awaits the command. 

I think that opening the view would be fire and forgot, so maybe return the thenable result instead of awaiting?

DonJayamanne:
> I think that opening the view would be fire and forgot, so maybe return the thenable result instead of awaiting?

Yes agreed with that, `fire & forget`.

src/client/datascience/commands/commandRegistry.ts
DonJayamanne: If we want to fire & forget, then we should make `openVariablesView` a sync function.
Else someone calling this in the future might make the wrong assumption.

src/datascience-ui/variable-view/variableViewPanel.tsx
joyceerhl: @DonJayamanne I don't think we actually want to keep that button so no need to localize here.

joyceerhl (approved):

Issue #4191, opened 2020-12-09
Ensure we can start kernels even when we have kernels with the same name


DonJayamanne:
This needs to be done as we have reverted thi s change in the release and change is in main

DonJayamanne:
* Ensure duplicates are not displayed in kernel picker
* If we have an interpreter XYZ and we have a python kernel inside that, then we should only display the interpreter or the kernel (not both) as both are identical
	* use command `python -m ipykernel install --sys-prefix --name duplicate
	* Now edit the kernelspec.json to give it some custom variables in the json.
* After we add env in the kernelspec.json then confirm we have both kernels

rchiodo:
I still have dupes in the kernel picker:

![image](https://user-images.githubusercontent.com/19672699/105779817-0149b080-5f24-11eb-843c-6c951ea3d922.png)

However they're not dupes? I can select them and they will show the correct sys paths. So I think this is working as designed.

Validated.

PR #4322, opened 2021-01-07
Use KernelConnectionMetadata instead of INotebookMetadata where possible

DonJayamanne:
For #4191


src/client/datascience/notebookStorage/notebookModel.ts
```diff
@@ -23,6 +24,9 @@ export class NativeEditorNotebookModel extends BaseNotebookModel {
     public get cells(): ICell[] {
         return this._cells;
     }
+    public get kernelConnection(): KernelConnectionMetadata | undefined {
+        return this._kernelConnection;
```
DonJayamanne, line 28: In NativeNotebooks, VSC keeps track of the current selected kernel (which contains the KernelConnectionMetadata).
I didn't want to create a property in the base INotebookModel to keep track of the KernelConnection as this is only required for interactive window & WebView notebooks. 

Thus for non-native notebooks, the KernelConnectionMetadta is stored against the model.
& for NativeNotebooks we let VSC store that information.

I could create a separate class that stores this information, but then its yet another class that needs to be added to interacgive window, native notebooks & webview notebooks, I think adding it into the NotebookModel (non-native notebook model) is simpler.

Also, anyone writing code for non-native notebooks will not be able to access this property (hence guarding against accidental usage in wrong places).

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
```diff
@@ -146,11 +145,12 @@ export class HostRawNotebookProvider
         traceInfo(`Getting preferred kernel for ${identity.toString()}`);
         try {
             // We need to locate kernelspec and possible interpreter for this launch based on resource and notebook metadata
-            const kernelConnectionMetadata = await this.kernelSelector.getPreferredKernelForLocalConnection(
+            // TODO: Confirm this logic is valid.
```
DonJayamanne, line 148: Some personal TODOs: I need to check if we even need `getPreferredKernelForLocalConnection`
Its possible we need it, but i would like to understand it better (possible its only required for live share, if thats the case, i'll update comments).
FYI - I don't see when the `getPreferredKernelForLocalConnection` will get called for non-live share scenearios - hence the note

src/client/datascience/jupyter/liveshare/hostJupyterServer.ts
```diff
@@ -310,26 +310,36 @@ export class HostJupyterServer extends LiveShareParticipantHost(JupyterServerBas
...
-            notebookMetadata?.id ||
+            kernelConnection ||
             resourceInterpreter?.displayName !== launchInfo.kernelConnectionMetadata?.interpreter?.displayName
         ) {
-            const kernelInfo = await (launchInfo.connectionInfo.localLaunch
+            // TODO: Need to read and ensure this logic is correct.
+            // Basically, if we have the connection information, then there's no need to perform any search.
+            // If we don't have the kernel connection information, then we haven't performed the search
+            // Is this assumption correct, do we not search for kernel information when starting a notebook without selecting a kernel.
+            // Doesn't extension today rely on this code to select a default kernel? (mayby not)..
```
DonJayamanne, line 320: Some personal TODOs: I need to check if we even need `getPreferredKernelForLocalConnection`
Its possible we need it, but i would like to understand it better (possible its only required for live share, if thats the case, i'll update comments).
FYI - I don't see when the `getPreferredKernelForLocalConnection` will get called for non-live share scenearios - hence the note

src/client/datascience/notebookStorage/baseModel.ts
```diff
@@ -28,6 +29,47 @@ type KernelIdListEntry = {
...
+            list.splice(index, 1);
+        }
+
+        // If adding a new one, push
+        if (id) {
+            list.push({ fileHash, kernelId: id });
+        }
+
+        // Prune list if too big
+        while (list.length > MaximumKernelIdListSize) {
```
rchiodo, line 66: 40 may not be enough. We might have telemetry on the number notebooks in a workspace though.

DonJayamanne: I'll bump it to 100, 100 in a list is super small

DonJayamanne: Updated to 100 & added telemetry to capture the size of this list

src/client/datascience/interactive-common/interactiveBase.ts
```diff
@@ -127,7 +126,8 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
     }
 
     public abstract isInteractive: boolean;
-    protected abstract get notebookMetadata(): INotebookMetadataLive | undefined;
+    protected abstract get notebookMetadata(): Readonly<nbformat.INotebookMetadata> | undefined;
+    protected abstract get kernelConnection(): Readonly<KernelConnectionMetadata> | undefined;
```
DonJayamanne, line 130: Converted these to readonly so that we can ensure we never attempt to update any properties.
Not that we have, e.g. in the case of notebook metadata there's a very specific way to update the metadata (you cannot edito this directly).
Same way with connections, added comments in the code why they are readonly (future proofing & ensure they are treated as immutable types).

src/client/datascience/jupyter/jupyterExecution.ts
```diff
@@ -192,13 +192,16 @@ export class JupyterExecutionBase implements IJupyterExecution {
...
-                        await sessionManager.dispose();
+                        try {
+                            kernelConnectionMetadata = await this.kernelSelector.getPreferredKernelForRemoteConnection(
+                                undefined,
+                                sessionManager,
+                                options?.metadata,
+                                cancelToken
+                            );
+                        } finally {
+                            await sessionManager.dispose();
```
DonJayamanne, line 203: An unrelated clean up, not required, but harmless & slightly better (again unrelated, couldn't help my self when fixing the typo). 😄

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -133,16 +131,12 @@ export class Kernel implements IKernel {
             return;
         } else {
             await this.validate(this.uri);
-            const metadata = ((getDefaultNotebookContent().metadata || {}) as unknown) as nbformat.INotebookMetadata;
-            // Create a dummy notebook metadata & update the metadata before starting the notebook (required to ensure we fetch & start the right kernel).
-            // Lower layers of code below getOrCreateNotebook searches for kernels again using the metadata.
-            updateNotebookMetadata(metadata, this.metadata);
```
DonJayamanne, line 139: Got rid of the hack, yay...

src/client/datascience/jupyter/kernels/types.ts
```diff
@@ -65,23 +65,28 @@ export type DefaultKernelConnectionMetadata = {
...
     kernelSpec?: IJupyterKernelSpec;
     interpreter: PythonEnvironment;
     kind: 'startUsingPythonInterpreter';
-};
+}>;
+/**
+* Readonly to ensure these are immutable, if we need to make changes then create a new one.
+* This ensure we don't update is somewhere unnecessarily (such updates would be unexpected).
+* Unexpected as connections are defined once & not changed, if we need to change then user needs to create a new connection.
+ */
```
DonJayamanne, line 84: Added comments to explain why i marked them readonly.

src/client/datascience/jupyter/liveshare/hostJupyterServer.ts
DonJayamanne: This is a WIP

src/test/datascience/nativeEditor.functional.test.tsx
```diff
@@ -482,77 +484,79 @@ suite('DataScience Native Editor', () => {
...
-                        when(ioc.applicationShell.showQuickPick(anything(), anything(), anything())).thenCall(
-                            async (o: IKernelSpecQuickPickItem[]) => {
-                                const existing = o.filter(
-                                    (s) =>
-                                        s.selection.kind === 'connectToLiveKernel' &&
-                                        s.selection.kernelModel.numberOfConnections
-                                );
+                    // Skip test for older python and raw kernel
+                    if (!pythonService || ioc.mockJupyter){
+                        return this.skip();
```
DonJayamanne, line 495: Used `skip` instead of conditionally not running a test.
I was running the tests locally and spent a lot of time trying to figure out why it was workign locally & not on the server, as locally mocha results indicated it ran and completed successfully.

Found that we had if conditions that didn't use `skip`, which produces different outputs.

src/test/datascience/jupyter/kernels/installationPrompts.vscode.test.ts
```diff
@@ -61,7 +61,8 @@ suite('DataScience Install IPyKernel (slow) (install)', () => {
         vscodeNotebook = api.serviceContainer.get<IVSCodeNotebook>(IVSCodeNotebook);
     });
 
-    setup(async () => {
+    setup(async function () {
+        console.log(`Start test ${this.currentTest?.title}`);
```
DonJayamanne, line 65: Added this as its a little difficult to figure out with all of the logging when tests start & when they end, as we have a lot of other stuff in setup & teardown (& sometimes we have additoinal logging that leaks from other tests - due to stuff not getting disposed properly)

src/client/datascience/interactive-common/notebookProvider.ts
```diff
@@ -154,6 +154,7 @@ export class NotebookProvider implements INotebookProvider {
                   resource,
                   options.disableUI,
                   options.metadata,
+                  options.kernelConnection,
```
DonJayamanne, line 157: Keeping both, if we have kernel connection, that trumps the metadata.

src/client/datascience/jupyter/kernels/kernelSelector.ts
```diff
@@ -227,15 +229,23 @@ export class KernelSelector implements IKernelSelectionUsage {
...
+            selection.kind === 'startUsingKernelSpec'
+        ) {
+            const itemToReturn = cloneDeep(selection) as ReadWrite<
+                KernelSpecConnectionMetadata | PythonKernelConnectionMetadata | DefaultKernelConnectionMetadata
+            >;
             itemToReturn.interpreter =
                 itemToReturn.interpreter ||
                 (this.extensionChecker.isPythonExtensionInstalled
-                    ? await this.interpreterService.getActiveInterpreter(resource)
+                    ? await this.kernelService.findMatchingInterpreter(selection.kernelSpec, cancelToken)
```
DonJayamanne, line 244: Use a better API to detect the interpreter associated with a kernel.

src/client/datascience/jupyter/kernels/kernelSelector.ts
```diff
@@ -227,15 +229,23 @@ export class KernelSelector implements IKernelSelectionUsage {
         telemetryProps.kernelSpecFound = !!selection?.kernelSpec;
         telemetryProps.interpreterFound = !!selection?.interpreter;
         sendTelemetryEvent(Telemetry.FindKernelForLocalConnection, stopWatch.elapsedTime, telemetryProps);
-        const itemToReturn = cloneDeep(selection);
-        if (itemToReturn) {
+        if (
+            selection &&
+            !selection.interpreter &&
+            isPythonKernelConnection(selection) &&
```
DonJayamanne, line 235: Bug fix, ensure we only do this for python kernels.

src/client/datascience/jupyter/kernels/kernelSelector.ts
```diff
@@ -259,8 +269,11 @@ export class KernelSelector implements IKernelSelectionUsage {
         ]);
 
         // First check for a live active session.
-        if (notebookMetadata && notebookMetadata.id) {
-            const session = sessions?.find((s) => s.kernel.id === notebookMetadata?.id);
+        const preferredKernelId = resource
+            ? this.preferredRemoteKernelIdProvider.getPreferredRemoteKernelId(resource)
```
DonJayamanne, line 273: Use separate class to find the remote session (the kernelId is tracked against the Uri in that class, thi way we don't need to pass & store id as part of metadata, removing)

src/client/datascience/jupyter/kernels/kernelService.ts
```diff
@@ -305,7 +309,7 @@ export class KernelService {
         });
         // Swallow errors if we get out of here and not resolve this.
         execServicePromise.ignoreErrors();
-        const name = this.generateKernelNameForIntepreter(interpreter);
+        const name = this.generateKernelNameForInterpreter(interpreter);
```
DonJayamanne, line 312: Fixed typo

src/client/datascience/jupyter/kernels/types.ts
```diff
@@ -25,20 +25,20 @@ export type LiveKernelModel = IJupyterKernel & Partial<IJupyterKernelSpec> & { s
  * Connection metadata for Live Kernels.
  * With this we are able connect to an existing kernel (instead of starting a new session).
  */
-export type LiveKernelConnectionMetadata = {
+export type LiveKernelConnectionMetadata = Readonly<{
```
DonJayamanne, line 28: Added comments further below why these are readonly (basically these are immutable types & connections cannot change)

src/client/datascience/kernel-launcher/kernelFinder.ts
```diff
@@ -107,7 +108,16 @@ export class KernelFinder implements IKernelFinder {
         this.writeCache().ignoreErrors();
 
         // ! as the has and set above verify that we have a return here
-        return this.workspaceToKernels.get(workspaceFolderId)!;
+        const promise = this.workspaceToKernels.get(workspaceFolderId)!;
+        promise
+            .then((items) =>
+                traceInfoIf(
```
DonJayamanne, line 114: Logging on CI for diagnostics, found that we don't log kernels that we have discovered.

src/client/datascience/kernel-launcher/kernelFinder.ts
```diff
@@ -480,6 +489,10 @@ export class KernelFinder implements IKernelFinder {
                 path.join(this.context.globalStorageUri.fsPath, cacheFile),
                 JSON.stringify(this.cache)
             );
+            traceInfoIf(
+                !!process.env.VSC_JUPYTER_LOG_KERNEL_OUTPUT,
```
DonJayamanne, line 493: Only for CI (logging for diagnostics)

.github/actions/create-venv-for-tests/action.yml
```diff
@@ -14,7 +14,7 @@ runs:
     # These tests are slow hence will only run on linux.
     # This env will be used to install ipykernel & test for prompts if ipykernel is missing & imilar tests.
     # Ensure this is registered as a kernel.
-    - name: Create virtual environment for Tests
+    - name: Create virtual environment without ipykernel
```
rchiodo, line 17: Doesn't this install ipykernel?

src/client/datascience/jupyter/kernels/kernelSelections.ts
```diff
@@ -200,24 +199,61 @@ export class InstalledJupyterKernelSelectionListProvider
...
+                    // Ensure we have the associated interpreter information.
+                    const selection = item.selection as ReadWrite<KernelSpecConnectionMetadata>;
+                    if (selection.interpreter || !isPythonKernelConnection(selection)) {
+                        return item;
+                    }
+                    selection.interpreter = await this.kernelService.findMatchingInterpreter(
+                        selection.kernelSpec,
+                        cancelToken
+                    );
+                    item.selection = selection;
```
rchiodo, line 253: Is this line necessary?

src/client/datascience/notebookStorage/baseModel.ts
```diff
@@ -210,7 +203,7 @@ export abstract class BaseNotebookModel implements INotebookModel {
     public get onDidEdit(): Event<NotebookModelChange> {
         return this._editEventEmitter.event;
     }
-    public get metadata(): INotebookMetadataLive | undefined {
+    public get metadata(): Readonly<nbformat.INotebookMetadata> | undefined {
         return this.kernelId && this.notebookJson.metadata
             ? {
                   ...this.notebookJson.metadata,
```
rchiodo, line 209: Wouldn't the id not be necessary here anymore?

src/client/datascience/notebookStorage/baseModel.ts
```diff
@@ -230,12 +223,14 @@ export abstract class BaseNotebookModel implements INotebookModel {
     protected _isDisposed?: boolean;
     protected _changedEmitter = new EventEmitter<NotebookModelChange>();
     protected _editEventEmitter = new EventEmitter<NotebookModelChange>();
+    protected _kernelConnection?: KernelConnectionMetadata;
     private kernelId: string | undefined;
```
rchiodo, line 227: I think you can get rid of this now? Is'nt the storage for this in that other location now?

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
```diff
@@ -232,6 +270,27 @@ export class HostRawNotebookProvider
         return notebookPromise.promise;
     }
 
+    // If we need to install our dependencies now (for non-native scenarios)
+    // then install ipykernel into the interpreter or throw error
+    private async installDependenciesIntoInterpreter(
+        interpreter: PythonEnvironment,
+        ignoreDependencyCheck?: boolean,
```
rchiodo, line 277: Is this set to true anywhere?

.github/actions/create-venv-for-tests/action.yml
DonJayamanne: & then it uninstalls it (few lines below in the script).
We install it so that we can register the kernel.

src/client/datascience/jupyter/kernels/kernelSelections.ts
DonJayamanne: Yes, to update it with a new item.

src/client/datascience/notebookStorage/baseModel.ts
DonJayamanne: Nope, its not. We don't need the id at all.
When connecting to remote, all we need is the Uri of the notebook and then the class will get the id based on that Uri. This way we don't need to pass this id around.

src/client/datascience/notebookStorage/baseModel.ts
DonJayamanne: Yes, but the other location inherits this class. Hence is required.

src/client/datascience/jupyter/kernels/kernelSelections.ts
rchiodo: Maybe I need to read up on javascript then. Isn't this a reference to the item.selection object?

```typescript
const selection = item.selection
```

So if you change selection.foo, it also changes item.selection.foo?

rchiodo: No it does behave as I expected. Object assignment is by reference unless you explicitly clone/copy in the assignment.

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
DonJayamanne: Yes, in kernelProvider.ts (when getting the preferred kernel for VSCode notebooks, we don't display prompts or errors about dependencies missing). 
We do that only when running a cell.

rchiodo: But it's a private function? Am I missing where it's called outside of this with 'ignoreDependencyCheck' set to true?

src/client/datascience/notebookStorage/baseModel.ts
rchiodo: So shouldn't you be able to remove the 'this.kernelId' from the class?

src/client/datascience/notebookStorage/baseModel.ts
rchiodo: Not the _kernelConnection, the kernelId.

DonJayamanne: Yup, removed, thanks

src/client/datascience/notebookStorage/baseModel.ts
DonJayamanne: Yup removed, i thought i removed it, removed for certain now.
Also our existing functional tests will ensure this is tested correctly.

src/client/datascience/jupyter/kernels/kernelSelections.ts
DonJayamanne: Oh shoot, you are right, i don't need to reset the variable, I've done that in `getKernelSelections` (few lines above, line 181).
Agreed, this is not necessary.

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
DonJayamanne: The argument `ignoreDependencyCheck` is passed from another public function into this private function.

DonJayamanne: The public function `getPreferredKernelForLocalConnection` has the same argument that gets passed into this private function.

rchiodo: Sorry I'm still not seeing it. The function ```installDependenciesIntoInterpreter``` you added in this PR. It's only called once as far as I can tell.

rchiodo: And it's called with 'false' as the value.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -133,16 +131,13 @@ export class Kernel implements IKernel {
...
             this._notebookPromise = this.notebookProvider.getOrCreateNotebook({
                 identity: this.uri,
                 resource: this.uri,
                 disableUI: options?.disableUI,
                 getOnly: false,
-                metadata,
+                metadata: undefined, // No need to pass this, as we have a kernel connection (metadata is required in lower layers to determine the kernel connection).
```
IanMatthewHuff, line 139: Hmm, not sure on this. But due to the name changes and the difference between INotebookMetadata and KernelConnectionMetadata we have some stuff like this:
```
metadata: undefined,
kernelConnection: this.metadata
```
Which just scans odd to me. Looks funny to not have metadata be metadata (though I get why it's not). If we have more instances like this we could rename this.metadata in this class to kernelConnection to line the names up.

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
DonJayamanne: In most cases its call as `false` or `undefined`. But When called from Kernel Provider in native notebooks, (which invokes `getpreferredKernelForLocalConnection`), then we pass `true`.

`false` is the default.

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: Agreed, will fix.

DonJayamanne: Fixed

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
rchiodo: Still don't see where 'installDependenciesIntoInterpreter ' is called more than once? What line is called on?

rchiodo: I believe your talking about 'installDependenciesIntoInterpreter' in kernelSelector.ts. I'm talking about the copy of that function you added to hostRawNotebookProvider. It looks to me like it's only called once. And ignoreDependencies is always false.

IanMatthewHuff: Yeah, I'm a bit confused on this as well. This seems like a really odd thing to be happening here in the code. Doesn't seem like a responsibility of this class.

DonJayamanne: > I believe your talking about 'installDependenciesIntoInterpreter' in kernelSelector.ts. I'm talking about the copy of that function you added to hostRawNotebookProvider. It looks to me like it's only called once. And ignoreDependencies is always false.

Oops. Let me check the comments again.

DonJayamanne: Sorry for the confusion, fixed

rchiodo (approved):

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
IanMatthewHuff: @DonJayamanne Sorry last comment here before approving. Why are we installing the dependencies here at this point? I remember from you and I talking about this before that we wanted dependency installation at one point. But now it's happening both in the selector and here, seems confusing and not the responsibility of this class.

DonJayamanne: I agree. However at this point there's no need to invoke the selector again as we have already selected the kernel. Happy to move this elsewhere, not sure where though?

I think we will have it at two points.
In webview & interactive window we display install instructions when selecting.
In new notebooks, we display install instructions only when starting.
As these are two separate UX  workflows I don't see how we can keep it in one place. Either way. 

Also, the code to install missting dependencies is already used in like 6 places. 
I think 1 for export, 1 for data viewer, 1 for native notebooks & the rest are just duplicates due to complex work flows.

IanMatthewHuff: You mentioned invoking the selector again. I was pretty sure that your previous changes meant that when we select we should get an interpreter with dependencies installed. What's the code path where we get here with an interpreter without dependencies installed?

IanMatthewHuff: Sorry I promise I'll stop fussing soon. Just seems like this is adding a very odd responsibility into this class. Does the kernelSelector code not run in the native notebook launch path now?

IanMatthewHuff: Bleh I don't want to hold up overnight on this as I'm not that concerned, basically just saying that I agree with your comments way back here: https://github.com/microsoft/vscode-python/pull/11341#discussion_r413392156 when you and I were debating the same issue.

IanMatthewHuff (approved):

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
DonJayamanne: Thats ok happy to discuss.
The the past, we didn't explictly call this method in native notebooks.
In native notebooks, we call the selector & pass a flag to ignore checking dependencies.
When starting a notebook (in native notebooks), we used to call the selector once again (that's what i'm trying to remove) & at that point the flag is false (i.e. it WILL check dependnecies).
Thus in old code, when starting a native notebook, the dependencies were checked as part of the selector.

Now that we dont have the selector anymore, we still need to ensure the dependencies are still checked.

Hope that makes sense.

package.json
```diff
@@ -1832,7 +1832,7 @@
         "preTestJediLSP": "node ./out/test/languageServers/jedi/lspSetup.js",
         "testJediLSP": "node ./out/test/languageServers/jedi/lspSetup.js && cross-env CODE_TESTS_WORKSPACE=src/test VSC_JUPYTER_CI_TEST_GREP='Language Server:' node ./out/test/testBootstrap.js ./out/test/standardTest.js && node ./out/test/languageServers/jedi/lspTeardown.js",
         "pretestVSCode": "cross-env VSC_JUPYTER_CI_TEST_VSC_CHANNEL=stable node ./out/test/datascience/dsTestSetup.js",
-        "testVSCode": "cross-env CODE_TESTS_WORKSPACE=src/test/datascience VSC_JUPYTER_CI_TEST_VSC_CHANNEL=stable TEST_FILES_SUFFIX=vscode.test VSC_JUPYTER_FORCE_LOGGING=1 VSC_JUPYTER_LOAD_EXPERIMENTS_FROM_FILE=true node ./out/test/testBootstrap.js ./out/test/standardTest.js",
+        "testVSCode": "cross-env CODE_TESTS_WORKSPACE=src/test/datascience VSC_JUPYTER_CI_TEST_VSC_CHANNEL=stable VSC_JUPYTER_CI_TEST_GREP=xxx TEST_FILES_SUFFIX=vscode.test VSC_JUPYTER_FORCE_LOGGING=1 VSC_JUPYTER_LOAD_EXPERIMENTS_FROM_FILE=true node ./out/test/testBootstrap.js ./out/test/standardTest.js",
```
DonJayamanne, line 1835: Will fix before merging, trying to fix failing tests

src/test/datascience/notebook/notebookEditor.vscode.test.ts
```diff
@@ -22,7 +22,7 @@ import {
 } from './helper';
 const vscodeNotebookEnums = require('vscode') as typeof import('vscode-proposed');
 
-suite('Notebook Editor tests', () => {
+suite('Notebook Editor testsxxx', () => {
```
DonJayamanne, line 25: Will fix before merging, trying to fix failing tests

Issue #4164, opened 2020-12-08
Switching kernels in the interactive window loses the ```__file__``` value.


rchiodo:
```python
print(__file__)
```

IanMatthewHuff:
Validated.
![image](https://user-images.githubusercontent.com/812783/104361772-c4170480-54c7-11eb-9cce-a344dad802a8.png)

PR #4324, opened 2021-01-07
set file in kernel when we change kernel in the interactive window

DavidKutu:
For #4164

### DIFF
--- a/src/client/datascience/interactive-common/interactiveBase.ts
+++ b/src/client/datascience/interactive-common/interactiveBase.ts
@@ -1165,14 +1165,23 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
 
     private selectNewKernel() {
         // This is handled by a command.
-        this.commandManager.executeCommand(Commands.SwitchJupyterKernel, {
-            identity: this.notebookIdentity.resource,
-            resource: this.owningResource,
-            currentKernelDisplayName:
-                this.notebookMetadata?.kernelspec?.display_name ||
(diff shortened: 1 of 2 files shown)


src/client/datascience/interactive-common/interactiveBase.ts
```diff
@@ -1165,14 +1165,23 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
...
+                identity: this.notebookIdentity.resource,
+                resource: this.owningResource,
+                currentKernelDisplayName:
+                    this.notebookMetadata?.kernelspec?.display_name ||
+                    this.notebookMetadata?.kernelspec?.name ||
+                    getDisplayNameOrNameOfKernelConnection(this._notebook?.getKernelConnection())
+            })
+            .then(async () => {
+                // Reset our file in the kernel.
+                const fileInKernel = this.fileInKernel;
```
rchiodo, line 1179: I think it would actually be better to do this in the 'kernelChangedHandler' function. That would handle the case where this function isn't called (like if the command is executed from the palette).

DonJayamanne: Not a fan of this, we have a lot of other code where we initialize the kernel once started/restarted (setting cwd, matplot lib options, jedi options, & more )
I'd prefer this was done in the same place where we do the other stuff.

DonJayamanne (changes requested):

rchiodo (approved):

DonJayamanne (approved):

Issue #304, opened 2020-10-27
Make tensors viewable in data viewer


joyceerhl:
- Add tensors to the list of types that are supported in the dataviewer
- Update vscodeGetDataFrameRows to massage tensors into the right format (currently all the other supported types get converted to pandas dataframes)

joyceerhl:
To validate this:

Run the following code in the IW or a notebook:

```python
# PyTorch Tensors
import torch 
i = torch.LongTensor([[2, 4]])
v = torch.FloatTensor([[1, 3], [5, 7]])
w = torch.sparse.FloatTensor(i, v)  # sparse tensor

# TensorFlow EagerTensors
import tensorflow as tf
a = tf.constant([[1.0, 2.0], [3.0, 4.0]])
```

Verify that you can open all these tensors in the data viewer and that their shapes show up correctly in the variable explorer.

rchiodo:
I can't get this code to run. It just hangs when I run it.

joyceerhl:
You might have better luck splitting them into two cells (one for torch and one for tensorflow). Importing each module takes a really long time.

rchiodo:
They already are. The first cell has been running for 5 minutes.

joyceerhl:
Posting for posterity: if you're on Linux or Windows you need to check pytorch.org for the specific install command (do not just run `pip install torch`).

rchiodo:
Validated.

PR #4358, opened 2021-01-12
Support 2D tensors in data viewer

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/304

Note that this doesn't yet handle n-dimensional tensors. That'll be a separate PR. So for now if there's a 3D tensor it's still going to crash the data viewer.

tensors


rchiodo:
Looks good except it would be nice to have some tests for it? Either in the functional tests or in the ipython command line tests (pythonFiles\tests\ipython\test_variables.py)

rchiodo (approved):

src/client/datascience/jupyter/kernelVariables.ts
```diff
@@ -35,7 +35,15 @@ const DocStringRegex = /.*?\[.*?;31mDocstring:.*?\[0m\s+(.*)/;
...
 
-const DataViewableTypes: Set<string> = new Set<string>(['DataFrame', 'list', 'dict', 'ndarray', 'Series', 'Tensor']);
+const DataViewableTypes: Set<string> = new Set<string>([
+    'DataFrame',
+    'list',
+    'dict',
+    'ndarray',
+    'Series',
+    'Tensor',
+    'EagerTensor'
```
joyceerhl, line 45: Not sure what the best way to feature gate this is. The problem is we can only query experiment status from TS code. I have 2 ideas:
1. Add 'Tensor' and 'EagerTensor' to the set of data viewable types only if the user is in the experiment. This will stop the user from viewing tensors in the data viewer
2. Create a completely separate Python script codepath that is invoked from TS based on whether we're in the experiment (I don't like this one much, adds lots of complexity)

Thoughts?

rchiodo: I don't think we should bother putting this in an experiment. I think it should just work.

rchiodo: Otherwise I just pick option 1. Have the kernelVariables class not allow viewing them.

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -3,8 +3,34 @@
 import json as _VSCODE_json
 import pandas.io.json as _VSCODE_pd_json
 
+# PyTorch and TensorFlow tensors which can be converted to numpy arrays
+_VSCODE_allowedTensorTypes = ["Tensor", "EagerTensor"]
+
+# Function that converts tensors to DataFrames
+def convertTensorToDataFrame(ten):
```
DonJayamanne, line 10: Can we rename from `ten` to `tensor` or the like.
I thought this was a `number ten` and was confused.

DonJayamanne (approved):

DavidKutu (approved):

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
IanMatthewHuff, line 10: Unless this changed these functions do go into the users python global namespace. So I think that we need the _VSCODE_ prefix here on the function like the other functions. Otherwise it might mess with an existing convertTensorToDataFrame function.

IanMatthewHuff (changes requested):
Sorry, code looks good, but I think that we need the prefix on the function.

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
joyceerhl: Whoops yes thank you! 

I also wanted to ask: I noticed we explicitly `del` some Python variables that we declare in our scripts. Should I be doing that with the variables I declare as well?

IanMatthewHuff: Yeah, we were doing that. We should do that for this function as well. Wonder if we are missing any? I don't think it's a big deal since most tools hide stuff that starts with _.

Seems like it might be good to automate this somehow? Feels easy to forget / miss. In any case no need to worry about it for this PR, just add the prefix and `del`.

IanMatthewHuff (approved):

src/client/datascience/jupyter/kernelVariables.ts
joyceerhl: jmew says we can skip the experiment here 🎉

Issue #4163, opened 2020-12-08
Test switching conda kernels on CI


DonJayamanne:
To ensure we do not regress

rchiodo:
Adding a native test to verify kernel switching

rchiodo:
Already have one for interactive window.

IanMatthewHuff:
@rchiodo Does this test exist? In the PR it looks like it was disabled on vscode upstream. Not seeing it in the code currently.

rchiodo:
Sorry I removed it on a recent submission.

It was in here:
https://github.com/microsoft/vscode-jupyter/commit/233454316f3dbed6d336a4bfb5e81a6c8f541f48#diff-407846ad077861133760b3073c6dffee75297a69f158c447eb3e9a39fb9ee22f

rchiodo:
You should be able to add that back once Peng's change goes through.

IanMatthewHuff:
Sounds good. I'll just move to upstream for now.

joyceerhl:
Validated: https://github.com/microsoft/vscode-jupyter/runs/1780254622?check_suite_focus=true#step:29:8917

PR #4370, opened 2021-01-13
Add/enable tests for switching kernels

rchiodo:
For #4163 (must fix bug)

Turn on tests for switching kernels. We already had a functional test for this. It runs if there's more than one kernel (so I added a second interpreter in the conda case).

Did not add tests for native though. This is pending https://github.com/microsoft/vscode/issues/114289

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,6 @@ on:
       - 'release'
       - 'release/*'
       - 'release-*'
-      - rchiodo/switching_kernels
   workflow_dispatch:
 
 env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,6 +16,7 @@ on:
       - 'release'
       - 'release/*'
       - 'release-*'
+      - rchiodo/switching_kernels
   workflow_dispatch:
 
 env:
@@ -267,8 +268,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Use conda for group1 and group2
-        if: matrix.test-suite == 'group1' || matrix.test-suite == 'group2'
+      - name: Use conda for group1 and group3
+        if: matrix.test-suite == 'group1' || matrix.test-suite == 'group3'
         run: |
           echo "conda_python=true" >> $GITHUB_ENV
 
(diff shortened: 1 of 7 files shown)


src/test/datascience/group1.nonConda.functional.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    function verifyHtmlOnInteractiveCell(html: string | undefined | RegExp, cellIndex: number | CellPosition) {
+        const iw = ioc.getInteractiveWebPanel(undefined).wrapper;
+        iw.update();
+        verifyHtmlOnCell(iw, 'InteractiveCell', html, cellIndex);
+    }
+
+    runTest(
+        'Simple text with no python extension',
```
rchiodo, line 58: Seems jupyter extension will not run a 'conda' kernel if there's no python extension there to get it to work. So these tests will only run if there's a plain python. I moved them to a separate file to indicate this.

File name also forces the group number for the functional test split.

.github/workflows/main.yml
```diff
@@ -267,8 +267,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Use conda for group1 and group2
-        if: matrix.test-suite == 'group1' || matrix.test-suite == 'group2'
+      - name: Use conda for group1 and group3
```
IanMatthewHuff, line 270: Looking at the code below, as these groups dynamic?

rchiodo: Not sure what you mean by dynamic. The name? No. The number of them? No. There's always 4. Although it's not decided here. runFunctionalTests.js decides the number.

rchiodo: It would probably be better to actually set the groups in the yml file than in some javascript file. Although writing code in bash scripts is not my forte. But then groups would be decided in the same spot as we're using them.

rchiodo: I'll enter an issue for that.

IanMatthewHuff: Sorry, not exactly what I meant. I looks like the grouping code is making the bucket of files for each group based on file size. So if the size of files changes then something that was in group X might now be in group Y, which would mess up which tests were running with conda. Is that not the case?

rchiodo: Yes but that was kinda on purpose. I wanted it to be random which tests used conda as they should all work. 

However for those that need to be a certain group, I added the logic for forcing it. 

Not super happy with it all because it's so separated. I think the issue I entered would make it better.

rchiodo: Do you think we should 
- Force certain tests to run with conda?
or
- Run all tests with conda and without?

I was thinking we could do that too but didn't think it mattered that much. Conda failures generally happen for everything the same way when they do happen, so just running some of the tests with conda would verify conda works.

DonJayamanne (approved):

.github/workflows/main.yml
IanMatthewHuff: Yeah, I see the forcing now. The tricky bit is that if random tests are assigned to conda then we get run to run variability which always seems like a bad thing in general. Getting test randomly assigned to conda would throw off David's perf tests as well I would think. Lemme look at the new issue.

IanMatthewHuff (approved):

Issue #1197, opened 2020-08-20
Native Notebooks - Kernel Restart (Progress Indicator) Indicator


claudiaregio:
# Native Notebooks

## Current Behavior: 
When restarting the kernel the run all button and cell toolbar icons flash for cells that have outputs (for either selected or editor state). No cell toolbar icons appear to flash for cells that do not have outputs. 

![NNb Kernel Restart](https://user-images.githubusercontent.com/19276747/90815897-41a36380-e2e0-11ea-8f4c-396692b22e8e.gif)

## Expected Behavior: 
1. Disable green restart kernel button while in progress. 
2. Disable run all button while in progress.
3. Have blue bar scan across top of notebook canvas while in progress. 
4. Stop blue bar and resurface restart kernel and run buttons when done. 
5. No other icons flashing

Microsoft Data Science for VS Code Engineering Team: @rchiodo, @IanMatthewHuff, @DavidKutu, @DonJayamanne, @greazer, @joyceerhl

DavidKutu:
I talked with VS code and extensions aren't supposed to have access to the blue bar. Will do this with a notification.

DonJayamanne:
Validated

PR #4386, opened 2021-01-13
Indicate with a notification that the kernel is restarting

DavidKutu:
-disable restart command (and button) when the notebook is busy

For #1197


rchiodo (approved):

src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -233,6 +233,8 @@ export class NotebookEditor implements INotebookEditor {
         const kernel = this.kernelProvider.get(this.file);
 
         if (kernel && !this.restartingKernel) {
+            const promise = this.restartKernelInternal(kernel);
```
DonJayamanne, line 236: We have already started restarting the kernel here even before displaying the message
Also, if the user clicks `No` or cancels the prmopt, its too late, as we have restarted it.

DonJayamanne (changes requested):

DonJayamanne (commented):
Did we address the enablement of the button?

DavidKutu:
> Did we address the enablement of the button?

No, it either doesn't work right, or it happens so fast that it doesn't make a difference.

src/client/datascience/jupyter/jupyterNotebook.ts
```diff
@@ -168,6 +169,9 @@ export class JupyterNotebookBase implements INotebook {
     public get disposed() {
         return this._disposed;
     }
+    public get isRestarting() {
+        return this._isRestarting;
```
DonJayamanne, line 173: I think we might want to discuss this issue as a team.
Not a fan of this. We have 2 way to determine the status of a kernel now.
This property and the status.

DonJayamanne: I would prefer to have a single way of determining this, i.e. using the existing status.
E.g. when restarting a kernel, manually emit `Restarting` and supress all status emissions until it has restarted or restart has failed at which point we can send the current status of the kernel (which would most likely be in a busted state).

DonJayamanne: @rchiodo @IanMatthewHuff what do you think?
Book a meeting to discuss how we address this?

DonJayamanne (changes requested):
Blocking on the new `isRestarting` API.

src/client/datascience/jupyter/jupyterNotebook.ts
IanMatthewHuff: Catching up from my interview day. Is the crux of the issue here the way that we handle restart (by swapping to an existing session) doesn't line up with the actual status of the kernel?

DonJayamanne: Yes, thus inn such cases the status isn't doing what we expect it to.

DavidKutu:
Claudia's ruling: disabling is a nice to have. Just have the notification appear, and the process not happening a bunch of times if the user spams the button.

rchiodo (approved):

IanMatthewHuff (approved):

DonJayamanne (approved):

Issue #3979, opened 2020-11-19
When using native notebooks, if kernel selector is shown, interpreter selector should be hidden


joyceerhl:
![image](https://user-images.githubusercontent.com/30305945/99731205-69869c80-2a72-11eb-9bb9-357c4dd83e39.png)

DonJayamanne:
TODO:
* Restore old code in python extension
* When a notebook with a (viewtype in jupyter) is displayed, then hide interprter

greazer:
Looks like this should be in validate.

DonJayamanne:
Validated in development version of Python extension.

PR #4391, opened 2021-01-14
Use python API to register status bar filter

rchiodo:
For #3979 

Corresponding PR in python is here:
https://github.com/microsoft/vscode-python/pull/15149

Register an interpreter status bar filter to allow us to disable the interpreter when showing notebook files.

### DIFF
--- a/src/client/api/types.ts
+++ b/src/client/api/types.ts
@@ -42,6 +42,14 @@ export enum JupyterProductToInstall {
     pandas = 'pandas'
 }
 
+/**
+ * Implement this interface to control the visibility of the interpreter statusbar.
+ */
+export interface IInterpreterStatusbarVisibilityFilter {
+    readonly changed?: Event<void>;
+    readonly hidden: boolean;
+}
+
 export type PythonApi = {
     /**
      * IInterpreterService
@@ -94,6 +102,10 @@ export type PythonApi = {
      * @param resource file that determines which connection to return
      */
     getLanguageServer(resource?: InterpreterUri): Promise<ILanguageServer | undefined>;
+    /**
+     * Registers a visibility filter for the interpreter status bar.
+     */
+    registerInterpreterStatusFilter(filter: IInterpreterStatusbarVisibilityFilter): void;
 };
 
 export const IPythonInstaller = Symbol('IPythonInstaller');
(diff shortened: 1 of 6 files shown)


src/test/interpreters/visibilityFilter.ts
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
IanMatthewHuff, line 1: Why is this file in the test directory?

rchiodo: Whoops. Not supposed to be there

src/client/api/types.ts
```diff
@@ -94,6 +102,10 @@ export type PythonApi = {
      * @param resource file that determines which connection to return
      */
     getLanguageServer(resource?: InterpreterUri): Promise<ILanguageServer | undefined>;
+    /**
+     * Registers a visibility filter for the interpreter status bar.
+     */
+    registerInterpreterStatusFilter(filter: IInterpreterStatusbarVisibilityFilter): void;
```
DonJayamanne, line 108: Hmm, if we make this API public, then its possible for any extension to hide this statusbar item. 
Most if not all of the API exposed by the Python extension is idempotent, however this one is not, it could be abused, not that any one will, however considering how VS Code works, they generally disallow such stuff. I.e. we won't want another extension hiding UI elements that it didn't contribute, however here' we're allowing that.

Just a thought.
I wouldn't ever want this to be made public, and considering that I'd prefer to just use VSC API in Python extension to hide the status bar when a notebook is active (this doesn't require any jupyter dependency).

DonJayamanne (approved):

src/client/api/types.ts
rchiodo: I thought the 'jupyter' api is private? And if our extension can hide the python extension's status bar, can't another extension then do the same thing?

IanMatthewHuff, line 108: It's a bit funny to me that the name of the function doesn't line up with the name of the thing being registered. ie. InterpreterStatusFilter doesn't sound like the same thing as InterpreterStatusbarVisibilityFilter. But I must admit that registerInterpreterStatusbarVisibilityFilter is a bit of a mouthful, and it's clear from the parameter types. So ok with leaving as is.

IanMatthewHuff (approved):

Issue #4355, opened 2021-01-12
Add .vscode tests for variable view to ensure we don't break functionality.


IanMatthewHuff:
Need to make sure we have some tests here, note that we have functional tests for the old variable viewer which do much of the testing around specific variable types. That will all be the same. What we need here is variable view + native notebook + vscode specific stuff.

1 - Open a notebook, execute, verify variables
2 - 1 + execute a second cell, verify updated variables
3 - 1 + open a second editor, execute a cell, verify variables are from second editor
4 - 3 + switch back to first editor, verify variables are from first editor

IanMatthewHuff:
I'd like to reopen this. We have the single sniff test in, but I had at least one more that I wanted to add. Already had it in progress.

DavidKutu:
Validated, tests are in

PR #4400, opened 2021-01-15
Main VariableView test code

IanMatthewHuff:
For #4355 


IanMatthewHuff:
Note: I'll have to merge up with the eslint change, and fix the xvfb action. But this is a full example of the variable view tests. Have a question or two on this, so would like to get the PR out while fixing the github action.

.vscode/launch.json
```diff
@@ -207,6 +207,7 @@
                 "VSC_FORCE_REAL_JUPYTER": "true", // Enalbe tests that require Jupyter.
                 "VSC_JUPYTER_CI_RUN_NON_PYTHON_NB_TEST": "", // Initialize this to run tests again Julia & other kernels.
                 "VSC_JUPYTER_RUN_NB_TEST": "true", // Initialize this to run notebook tests (must be using VSC Insiders).
+                "VSC_JUPYTER_WEBVIEW_TEST_MIDDLEWARE": "true", // Initialize to create the webview test middleware
```
IanMatthewHuff, line 210: All an env var to force on the test middle ware, needed for the middleware VariablesComplete message, and handy to be able to turn on quick just when running VS Code from debugger.

src/client/datascience/variablesView/variableViewProvider.ts
```diff
@@ -17,6 +19,24 @@ import { VariableView } from './variableView';
 export class VariableViewProvider implements IVariableViewProvider {
     public readonly viewType = 'jupyterViewVariables';
 
+    // Either return the active variable view or wait until it's created and return it
+    public get activeVariableView(): Promise<VariableView> {
+        // For test execution only
```
IanMatthewHuff, line 24: This is one of the main areas that I'm still mulling over some. Both this provider and the webview host now exposed some functions that should only be used by test code. I'm a bit leery of this, but had the following reasoning:

1. Wrapping / patching options would be adding more complexity
2. Keeping the code that we test as much 'exact product code' as possible felt good for the .vscode test
3. Only a few functions were needed, and they could be marked with isTestExecution to prevent product usage

On the downside, it's exposing a few things that don't need to be exposed in non-test scenarios. Opinions on this? And on the HTML fetching code in the webview? If there are ideas for a cleaner way to hide this without adding a bunch of test -> product code complexity I'd be down for it.

IanMatthewHuff: It's just typescript check so I guess I could keep them private and then cast to a test specific interface in the test code, not sure that's any better.

src/test/datascience/vscodeTestHelpers.ts
```diff
@@ -0,0 +1,104 @@
...
+
+// This class is for usage in .vscode test to hook up to webviews which expose the addMessageListener and removeMessageListener functions
+export class OnMessageListener {
+    private target: IOnMessageListener;
+    constructor(target: IOnMessageListener) {
+        this.target = target;
+    }
+
+    // For our target object wait for a specific message to come in from onMessage function
+    public async waitForMessage(message: string, options?: WaitForMessageOptions): Promise<void> {
```
IanMatthewHuff, line 42: The waitForMessage implementation here is very close to the existing one in mountedWebView.ts. Figured that interface already existed and was doing the same basic thing so just reused it here with minor changes for not being part of a mountedWebView.

src/test/datascience/variableView/variableView.vscode.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    teardown(async function () {
+        await closeNotebooks(disposables);
+        await closeNotebooksAndCleanUpAfterTests(disposables);
+    });
+
+    // Cleanup after suite is finished
+    suiteTeardown(() => closeNotebooksAndCleanUpAfterTests(disposables));
+
+    // Test showing the basic variable view with a value or two
+    test('Can show VariableView', async function () {
```
IanMatthewHuff, line 70: More tests in the queue. But wanted to get approval here and get one basic test running on CI before adding more.

src/client/datascience/variablesView/variableViewProvider.ts
IanMatthewHuff: Making the isTestExecution bit a decorator might also work.

joyceerhl: I've done the equivalent of the following (involves the least typing, in both senses of the word). Maybe bad practice but nobody mentioned it on the code reviews:
```typescript
   (variableViewProvider as any).activeVariableView
```

But the more consistent thing to do would probably be the test interface idea. We do this for IExtensionTestApi as well:

https://github.com/microsoft/vscode-jupyter/blob/e8f311aa27269c2433d00a04f90cfd18c803f463/src/test/initialize.ts#L31

rchiodo: I like the extension test api that exposes privates only in tests.

So I believe here this would be IVariableViewTestProvider or something like that and it would have:

```
interface IVariableViewTestProvider {
   readonly activeVariableView: Promise<IVariableView>
}
```

IanMatthewHuff: @joyceerhl Yeah, leaving them dangling out in public doesn't seem good. Maybe the following?
1. Decorator for the isTestExecution check (I like this as it also annotates the function)
2. Interface for access to the test specific functions, Interfaces live in the test code. Keeping them in the test code prevents them from being seen as "oh, maybe I can use this" when looking at product code.
3. Us the (X as any) as Y cast in test code to get access to the private test methods.

rchiodo: The real class wouldn't actually implement this but would just leave the activeVariableView as a private Although that makes it a tad breakable.

You could also do like we did for the InteractiveWindowProvider where in tests, a different provider is added to the container. The test provider implements the IVariableViewTestProvider so then at least there's compile protection for the activeVariableView.

rchiodo (approved):

DonJayamanne (approved):

Issue #305, opened 2020-10-27
Slice data in data viewer


joyceerhl:
- Show Slice Data panel for >1D tensors and ndarrays
- Select first dimension by default
- Show shape of tensor and currently-selected axis
- Axis dropdown
- Index dropdown
- Refetch dataframe rows on axis/index selection change

joyceerhl:
Messaging is implemented, blocked on design of actual UI controls for slicing.

IanMatthewHuff:
There is probably a lot to fully test here. But I pulled up a 3x3 and tried to test all the basics. All looked good to me.

1. Slicing using the expression worked as expected
2. Slicing using Axis / Index boxes also worked as expected
3. Axis / Index and the slice expression correctly updated based on the other changing
4. When slice expression could not be reflected in Axis / Index it was disabled
5. Could stop slicing data then re-enable and it saved the slice that I had.

Validated.

PR #4402, opened 2021-01-15
Data slicing part 1: messaging and UI updates with a temporary slice control

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/305


joyceerhl:
This isn't ready for check in but I would appreciate some feedback on the approach I've taken here. I'm wondering if there's a way to reuse existing code / not special case slicing / add an API that is more general purpose than what I have now.

Two immediate limitations of the current PR:
1. Doesn't support fetching rows in a slice in chunks. I think we need this in the case of large data and because the debugger only supports fetching 100 rows at a time (if I understand the code comments correctly)
2. I haven't implemented slicing for debuggerVariables yet
3. Adds a whole special cased set of messages and methods for requesting a slice and updating the data viewer

I'm trying to decide how I want to approach adding the ability to request a slice to the data viewer because the approach I currently have feels off. Here are some initial ideas:
1. Add a new optional method to `IDataViewerDataProvider` called `getSlice` like so:
```ts
public getSlice?(sliceExpression: string, start: number, end: number)
 ```

The UI would then ask for a particular slice, and in the extension we would handle that request by getting the rows in chunks and sending them back to the UI, basically just like the way we eval whole variables and send them to the UI today.
(This is another problem with my current approach: because the slice indicator lives in the reactSlickGrid.tsx component, the grid component itself is aware of the notion of a slice and handles slice requests and responses. On second thought that state should be hoisted into the mainPanel and we should create a new grid for each slice, since the columns and row data all change so from a UI standpoint it's effectively it's a new grid with completely different data items.)

2. Treat each new slice as a new IJupyterVariable, i.e. pretend the user requested to open the slice in the data viewer to begin with. I don't think this really works because when the user is looking at a slice we do need to remember what the original variable was if they request a different slice on the same variable. But including it for completeness anyway.

3. Extend the functionality of the existing IJupyterVariables `getDataFrameRows` function. I currently have a `_VSCODE_getSlice` function to request the slice, but in fact we could achieve the same thing by passing the name + slice expression directly into  the existing `_VSCODE_getDataFrameRows` function, so that already eliminates one special purpose function. Wondering if I can also make `getDataFrameRows` more general purpose. One idea that avoids changing the signature of the existing method is to add a property to IJupyterVariable called 'sliceExpression'. `getDataFrameRows` would look for that property and figure out how to apply that operation to the variable's name. There's currently no need to support other kinds of variable manipulations but there could be in future especially with the codegen ideas we've talked about today.

4. Extend the functionality of the existing IJupyterVariables `getDataFrameRows` function by adding a fifth, optional, 'slice' parameter to the `getDataFrameRows` function. This is fine but will result in ugly method signatures if we continue down this path for other codegen in future.

Other ideas? Have I missed anything important that would preclude some of these approaches?

File: src/client/datascience/data-viewing/types.ts [hunk 1]
```diff
@@ -38,29 +40,39 @@ export interface IGetRowsResponse {
...
     [DataViewerMessages.CompletedData]: never | undefined;
+    [DataViewerMessages.GetSliceRequest]: IGetSliceRequest;
 };
 
 export interface IDataFrameInfo {
     columns?: { key: string; type: ColumnType }[];
     indexColumn?: string;
     rowCount?: number;
+    shape?: number[];
+    dataDimensionality?: number;
```
joyceerhl, line 66: This property isn't being used yet, leaving it in here because it'll be useful when building UI (for controlling whether to show slice controls for some data)

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -278,7 +293,7 @@ export class DebuggerVariables extends DebugLocationTracker
         }
     }
 
-    private async getFullVariable(variable: IJupyterVariable): Promise<IJupyterVariable> {
+    public async getFullVariable(variable: IJupyterVariable): Promise<IJupyterVariable> {
```
joyceerhl, line 296: Made this public because `getMatchingVariable` requires only a name, which it uses to lookup a partial `IJupyterVariable` amongst its known variables, and passes that to `getFullVariable`. This doesn't work at all when opening the data viewer from the debug variable context menu, because at that point DebuggerVariables doesn't know about any of our variables yet.

src/client/datascience/commands/commandRegistry.ts
```diff
@@ -496,11 +498,12 @@ export class CommandRegistry implements IDisposable {
...
-        if (this.debugService.activeDebugSession) {
-            const jupyterVariable = convertDebugProtocolVariableToIJupyterVariable(
-                request.variable as DebugProtocol.Variable
-            );
+        if (this.debugService.activeDebugSession && this.variableProvider.getFullVariable) {
             try {
+                const variable = convertDebugProtocolVariableToIJupyterVariable(
+                    request.variable as DebugProtocol.Variable
+                );
+                const jupyterVariable = await this.variableProvider.getFullVariable(variable);
```
joyceerhl, line 506: What we were doing previously (superficial conversion from `DebugProtocol.Variable` to `IJupyterVariable`) was sufficient for what we shipped in Jan, but isn't enough for slicing, because for slicing we need to know the `shape` and `type`. Those weren't being filled out for DebuggerVariables before because it wasn't necessary until now.

src/client/datascience/data-viewing/dataViewer.ts
```diff
@@ -140,23 +163,35 @@ export class DataViewer extends WebviewPanelHost<IDataViewerMapping> implements
         return output;
     }
 
-    private async getAllRows() {
+    // Deprecate this
+    private async getAllRows(sliceExpression?: string) {
```
joyceerhl, line 167: As part of moving all the debugger row chunking logic to the UI, I want to suggest that we deprecate usage of `getAllRows`, because we can express the same thing with `getRowChunk` (just ask for all the rows in the chunk request). But I wanted to leave this for the review in case I'm missing a reason to have this here.

src/client/datascience/types.ts
```diff
@@ -863,22 +863,25 @@ export interface IJupyterExtraSettings extends IJupyterSettings {
...
     type: string;
     size: number;
     shape: string;
+    dataDimensionality?: number;
     count: number;
     truncated: boolean;
     columns?: { key: string; type: string }[];
     rowCount?: number;
     indexColumn?: string;
+    maximumRowChunkSize?: number;
```
joyceerhl, line 884: This is set in IJupyterVariables from DebuggerVariables to tell the UI the maximum chunk size it can ask for at once.

src/datascience-ui/data-explorer/mainPanel.tsx
```diff
@@ -218,37 +232,42 @@ export class MainPanel extends React.Component<IMainPanelProps, IMainPanelState>
...
+                    originalVariableShape: variable.originalVariableShape ?? variable.shape,
+                    dataDimensionality: variable.dataDimensionality ?? 2,
+                    isSliceDataSupported: variable.isSliceDataSupported,
+                    // Maximum number of rows is 100 if evaluating in debugger, undefined otherwise
+                    maximumRowChunkSize: variable.maximumRowChunkSize ?? this.state.maximumRowChunkSize
                 });
 
+                // New data coming in, so reset everything and clear our cache of columns
+                this.columnsContainingInfOrNaN.clear();
+                this.resetGridEvent.notify();
```
joyceerhl, line 245: There's something very weird happening here where the filter box does not render on a grid reset unless it was already set to visible before. Still working on tracking this down.

src/client/datascience/data-viewing/dataViewer.ts
```diff
@@ -61,6 +65,12 @@ export class DataViewer extends WebviewPanelHost<IDataViewerMapping> implements
         );
     }
 
+    private async isSliceDataSupported(dataFrameInfo: IDataFrameInfo) {
+        return (
+            !!dataFrameInfo?.supportsSlicing && (await this.experimentService.inExperiment(Experiments.SliceDataViewer))
```
rchiodo, line 70: Why do we need an experiment for a backend enhancement? Wouldn't this only affect the UI?

joyceerhl: Not sure what you mean? The data slicer is a new feature, hence putting it behind a feature flag makes sense. There is a test UI control there now so the feature flag ensures that it's not visible to everyone even if we ship these bits.

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
```diff
@@ -62,41 +78,58 @@ export class JupyterVariableDataProvider implements IJupyterVariableDataProvider
...
 
-    public async getDataFrameInfo(): Promise<IDataFrameInfo> {
+    public async getDataFrameInfo(sliceExpression?: string): Promise<IDataFrameInfo> {
         let dataFrameInfo: IDataFrameInfo = {};
         await this.ensureInitialized();
-        if (this.variable) {
+        let variable = this.variable;
+        if (variable) {
+            if (sliceExpression) {
+                variable = await this.variableManager.getDataFrameInfo(variable, this.notebook, sliceExpression);
```
rchiodo, line 87: This seems weird? You're caching the variable. At the very least shouldn't it be called something else, like the 'slicedVariable' or something?

Is there the potential for some other code elsewhere not to understand which slice it is looking at?

rchiodo: Yeah later on the variable name is used to make the row request. If the variable changes here, wouldn't that affect the stuff later? 

What is the reason to refetch here? Does it get a new shape? Perhaps you should just be caching the returned shape with something like 'currentSliceShape' or something?

rchiodo (changes requested):
⏲️

src/client/datascience/data-viewing/dataViewer.ts
rchiodo: I don't think we want feature flags in back end code unless necessary. It further complicates the code paths with extra ifs every where. The way you have this coded, the sliceExpression won't be used unless the UI sends it. So I think the feature flag can be only checked in the UI.

That way we test the new code paths with or without the feature flag.

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
rchiodo: So really you're just refetching the shape? Should that be the newly cached value?

src/client/datascience/commands/commandRegistry.ts
```diff
@@ -322,23 +324,23 @@ export class CommandRegistry implements IDisposable {
     private async debugStepOver(): Promise<void> {
         // Make sure that we are in debug mode
         if (this.debugService.activeDebugSession) {
-            this.commandManager.executeCommand('workbench.action.debug.stepOver');
+            void this.commandManager.executeCommand('workbench.action.debug.stepOver');
```
DonJayamanne, line 327: I think we should await on these, as this method is async & the calling method is also async.

src/client/datascience/data-viewing/dataViewer.ts
```diff
@@ -99,23 +114,31 @@ export class DataViewer extends WebviewPanelHost<IDataViewerMapping> implements
...
 
         super.onMessage(message, payload);
     }
 
-    private getDataFrameInfo(): Promise<IDataFrameInfo> {
-        if (!this.dataFrameInfoPromise) {
-            this.dataFrameInfoPromise = this.dataProvider ? this.dataProvider.getDataFrameInfo() : Promise.resolve({});
+    private getDataFrameInfo(sliceExpression?: string): Promise<IDataFrameInfo> {
+        // If requesting a new slice, refresh our cached info promise
+        if (!this.dataFrameInfoPromise || sliceExpression !== this.currentSlice) {
```
IanMatthewHuff, line 137: Might be reading this wrong. But this is the case where the slice expression has been updated. In that case would you want to cancel the existing dataFrameInfoPromise and start a new one? Looks like if there is already an existing promise here then it will just return the existing one. Doesn't seem to match up with how I read the comment.

src/client/datascience/commands/commandRegistry.ts
```diff
@@ -496,11 +498,12 @@ export class CommandRegistry implements IDisposable {
     }
     private async onVariablePanelShowDataViewerRequest(request: IShowDataViewerFromVariablePanel) {
         sendTelemetryEvent(EventName.OPEN_DATAVIEWER_FROM_VARIABLE_WINDOW_REQUEST);
-        if (this.debugService.activeDebugSession) {
-            const jupyterVariable = convertDebugProtocolVariableToIJupyterVariable(
-                request.variable as DebugProtocol.Variable
-            );
+        if (this.debugService.activeDebugSession && this.variableProvider.getFullVariable) {
```
DonJayamanne, line 501: Doesn't this mean, that every one implementing the variable provider MUST also implement `getFullVariable`. Isn't this a breaking change?
Is this ok for AML & others we consume our API?

src/client/datascience/data-viewing/dataViewer.ts
joyceerhl: `this.currentSlice` is the cache of the last slice expression we evaluated (that might not have been clear from the name). If we're asking for a new slice expression that doesn't match the one we already have cached, then we ask for a new one. And of course if there isn't a cached promise already then we need to ask for one anyway. Did I understand your question correctly?

src/client/datascience/commands/commandRegistry.ts
joyceerhl: I added it as an optional function in the interface definition for `IDataViewerDataProvider`, and this function is only used for debugger integration, so it should not be a problem for AML.

src/client/datascience/commands/commandRegistry.ts
IanMatthewHuff: Maybe? It's a functionality change though. Looks like this was just an eslint update. Previously our code would call debugStepOver then fire and forget on the workbench action. void keeps that existing functionality with a thenable. Changing to await changes how debugStepOver returns, might be ok? But might not be what we want.

src/client/datascience/data-viewing/dataViewer.ts
DonJayamanne: I agree.
The UI should just block data slicing features if its not enabled, independent of whether the variable supports slicing.
Also the `dataFrameInfo`  is sent to the UI, hence we can check on the UI whether the variable supports slicing on the UI based on whether:
* Slicing is enabled
* Slicing is supported for the variable.

src/client/datascience/data-viewing/dataViewer.ts
IanMatthewHuff: Hmmm, maybe I'm reading it wrong. It was the ! on the dataFrameInfoPromise that I was not sure about. Is that ever cleared? Looks to me like currentSlice would never update here after that promise is set for the first time.

DonJayamanne, line 137: This implies that previous slicing has not yet completed and the user chose a new slice expression.
Does this mean we need to support the concept of cancellations?
Is it possible that slicing can take a while (in which case a cancellation might prove useful, & possibly complicate the entire messaging as well)?

src/client/datascience/data-viewing/dataViewer.ts
```diff
@@ -99,23 +114,31 @@ export class DataViewer extends WebviewPanelHost<IDataViewerMapping> implements
...
+            this.dataFrameInfoPromise = this.dataProvider
+                ? this.dataProvider.getDataFrameInfo(sliceExpression)
+                : Promise.resolve({});
+            this.currentSlice = sliceExpression;
```
DonJayamanne, line 141: Need to check the code, but what happens if the slice expression is invalid?

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
```diff
@@ -53,6 +54,21 @@ export class JupyterVariableDataProvider implements IJupyterVariableDataProvider
         });
     }
 
+    // Parse a string of the form (1, 2, 3)
+    private static parseShape(shape: string) {
```
IanMatthewHuff, line 58: Looks like ShapeSplitterRegEx might be doing something similar?

src/client/datascience/data-viewing/dataViewer.ts
```diff
@@ -174,16 +209,19 @@ export class DataViewer extends WebviewPanelHost<IDataViewerMapping> implements
             if (e instanceof JupyterDataRateLimitError) {
                 traceError(e);
                 const actionTitle = localize.DataScience.pythonInteractiveHelpLink();
-                this.applicationShell.showErrorMessage(e.toString(), actionTitle).then((v) => {
-                    // User clicked on the link, open it.
-                    if (v === actionTitle) {
-                        this.applicationShell.openUrl(HelpLinks.JupyterDataRateHelpLink);
-                    }
-                });
+                this.applicationShell.showErrorMessage(e.toString(), actionTitle).then(
```
DonJayamanne, line 212: Feels like a goofy way to display errors `e.toString()`, existing code, hence no need to change.
I feel the UX is pretty poor in such cases, its basically dumping a stack trace to the user & they'd go, `what is this?`

Again, no need to change as its existing code.

src/client/datascience/data-viewing/dataViewer.ts
joyceerhl: > the sliceExpression won't be used unless the UI sends it.

Yes but the slice control will show up with the variable shape? I'm not sure we'll get the redesigned UI implemented before we cut the February release. I guess I could literally hardcode `supportsSlicing` to false or something in the UI, but I don't think that's what you're suggesting?

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
joyceerhl: No I'm refetching everything. New shape, new rowCount, etc

src/client/datascience/data-viewing/dataViewer.ts
```diff
@@ -174,16 +209,19 @@ export class DataViewer extends WebviewPanelHost<IDataViewerMapping> implements
...
+                            this.applicationShell.openUrl(HelpLinks.JupyterDataRateHelpLink);
+                        }
+                    },
+                    (e) => traceError(e)
+                );
                 this.dispose();
             }
             traceError(e);
-            this.applicationShell.showErrorMessage(e);
+            this.applicationShell.showErrorMessage(e).then(noop, (e) => traceError(e));
```
DonJayamanne, line 224: I'd just use `noop` for `showErrorMessage`, its not like we can do anything with these errors.
Its unlike it would fail, & unlikely we'll ever see this & do anything with it..
Nohting to change, just saying that I think its unnecessary.

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
```diff
@@ -53,6 +54,21 @@ export class JupyterVariableDataProvider implements IJupyterVariableDataProvider
...
+        try {
+            if (shape.startsWith('(') && shape.endsWith(')')) {
+                return shape
+                    .substring(1, shape.length - 1)
+                    .split(',')
+                    .map((shapeEl) => parseInt(shapeEl));
+            }
+        } catch (e) {
+            traceError(`Could not parse IJupyterVariable with malformed shape: ${shape}`);
```
DonJayamanne, line 67: Would be a good telemetry point to see how often this fails for users.
This way we'd know its not always working as expected & we need to look into this.

src/client/datascience/data-viewing/dataViewer.ts
joyceerhl: Not a problem yet, as we're not shipping the UI control I have implemented as is. Depending on the UI (e.g. dropdowns vs a freeform input box) I'll solve this in part 2.

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -127,7 +137,8 @@ export class DebuggerVariables extends DebugLocationTracker
         return results
             ? {
                   ...targetVariable,
-                  ...JSON.parse(results.result)
+                  ...JSON.parse(results.result),
+                  maximumRowChunkSize: 100
```
DonJayamanne, line 141: Can we make this a constant somewhere `100`

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -136,8 +147,14 @@ export class DebuggerVariables extends DebugLocationTracker
         targetVariable: IJupyterVariable,
         start: number,
         end: number,
-        notebook?: INotebook
+        notebook?: INotebook,
+        sliceExpression?: string
     ): Promise<{}> {
+        // Developer error. The debugger cannot eval more than 100 rows at once.
+        if (end - start > 100) {
```
DonJayamanne, line 154: I believe this is the same constant value as earlier.

src/client/datascience/data-viewing/dataViewer.ts
joyceerhl: Not necessarily, this is here because we used to cache `dataFrameInfoPromise` (to avoid running our Python code every time we needed to get the info). We need to force a refetch when the slice being requested isn't already cached.

It is possible that slicing can take a while. One solution for that would be disabling the UI when a slice is in progress. But this PR doesn't handle UI, as that hasn't been designed yet.

src/client/datascience/jupyter/debuggerVariables.ts
joyceerhl: Touche, will fix

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -290,10 +289,13 @@ export class DebuggerVariables extends DebugLocationTracker
...
         if (results && results.result) {
             // Results should be the updated variable.
+            const info = JSON.parse(results.result);
+
             return {
                 ...variable,
+                ...info,
                 truncated: false,
-                ...JSON.parse(results.result)
+                supportsSlicing: SliceableTypes.has(info.type || '')
```
DonJayamanne, line 298: Not sure whether we should be passing such an attribute all around.
If we had a JSON viewer or another type of viewer for variables we'd be adding more attributes to indicate what can & cannot be done with the variable, when its a decision to be made by a higher order.

I.e. why not just call a common function `isSliceableType` rather than passing this all around.

src/client/datascience/data-viewing/dataViewer.ts
joyceerhl: Right now if the slice expression is invalid the Python scripts don't return anything. But as I mentioned the intent is to ensure that our slice expression is always valid, e.g. through dropdowns or validating the provided slice expression against the variable's shape

src/client/datascience/data-viewing/dataViewer.ts
IanMatthewHuff: Cancellations would be nice but just making slicing a blocking procedure from the UI would work for a start. I think that I'm following now. I guess I would argue that (until we support cancellations) this might be more of a traceError case? The state where there is an outstanding dataFrameInfoPromise but the slice expression has changed seems like an invalid state to me. It would be unclear if getDataFrameInfo would be returning the results from the previous dataFrameInfoPromise or from the current slice expression.

src/client/datascience/jupyter/kernelVariables.ts
```diff
@@ -481,6 +497,10 @@ export class KernelVariables implements IJupyterVariables {
             if (DataViewableTypes.has(result.type)) {
                 result.supportsDataExplorer = true;
             }
+
+            if (SliceableTypes.has(result.type)) {
+                result.supportsSlicing = true;
```
DonJayamanne, line 502: I don't think this needs to be an attribute of the variable, this is something that can be derived.
Also if this had to be an attirbute, then it should come all the way from the backend (Python) not something we just tack onto the result (else feels hackish).

src/client/datascience/data-viewing/dataViewer.ts
IanMatthewHuff: Makes sense to me. It's a full subset of the other command.

src/client/datascience/data-viewing/dataViewer.ts
joyceerhl: I'm afraid I'm not quite following why this is invalid. If there is a cached promise and the slice expression has changed, that just means the user has asked for a new slice, and we should refetch, not reuse the cached promise.

src/datascience-ui/data-explorer/sliceControl.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <div className="slice-data-control-container" style={{ display: 'flex', justifyContent: 'space-around' }}>
+                <form onSubmit={this.handleSubmit} style={{ alignSelf: 'center' }}>
+                    <label>
+                        {'Slice data:  '}
+                        <input
+                            type="text"
+                            className="slice-data"
+                            value={this.state.value}
+                            onChange={this.handleChange}
+                            style={{ width: '80px' }}
```
DonJayamanne, line 46: I'd prefer to use CSS instead of hardcoding pixes in the tsx files.
Optionally I think we should try to use `ems` instead of pixels for better scaling.

DonJayamanne: Nothing to do, but move style into CSS.

src/datascience-ui/data-explorer/sliceControl.tsx
```diff
@@ -0,0 +1,54 @@
...
+        event.preventDefault();
+        this.props.handleSliceRequest({
+            slice: this.state.value,
+            originalVariableShape: this.props.originalVariableShape
+        });
+    }
+
+    render() {
+        return (
+            <div className="slice-data-control-container" style={{ display: 'flex', justifyContent: 'space-around' }}>
```
DonJayamanne, line 37: I think we can create one class for this & do the rest with CSS
```
div.slice-data-control-cotainer form { align.... }
div.slice-data-control-cotainer form input { width: 80px }
```

DonJayamanne (changes requested):
Everything is ok, just not too happy about the additional attribute that we are sticking onto the variable. Its an inferred attribute of the data, hence I don't think we should add a property to it.

Whether a variable supports XYZ (a UI feature) doesn't need to be an attribute if it can be derived from a simple function.

Happy to hear your thoughts on this.

src/datascience-ui/data-explorer/sliceControl.tsx
joyceerhl: This slice control is temporary and will never be shown to users, i.e. will literally be deleting this file in Part 2 of implementing slice data. So I don't think there's a need to fix this up here, but that's good feedback for when I implement the real UI 😁

src/client/datascience/jupyter/kernelVariables.ts
joyceerhl: This was modeled on `supportsDataExplorer` but in retrospect can be done in the UI. But I don't really agree with the idea of this attribute coming from the backend. The UI can just look at the `type` property for this to figure out whether to show the slice controls.

src/client/datascience/data-viewing/dataViewer.ts
joyceerhl: @rchiodo do you mean query control tower in the UI code?

src/client/datascience/data-viewing/dataViewer.ts
joyceerhl: The `!this.dataFrameInfoPromise` check is prexisting, i.e. if we don't have anything cached, we need to ask the data provider for the info. It's never cleared, but `this.currentSlice` and `this.dataFrameInfoPromise` are updated when we ask for a slice expression that is different than the one we have cached because the second condition in the if will evaluate to false even if the promise is set. 

Does this make any sense 😂

src/client/datascience/commands/commandRegistry.ts
DonJayamanne: If that's the case, then why do we need to make it optional?

src/client/datascience/jupyter/kernelVariables.ts
DonJayamanne: > he UI can just look at the type property for this to figure out whether to show the slice controls.

Agred. I only suggested sending the stuff from the backend, if you still wanted to keep the property on the object. 
So, my understanding is that this property will go away and we'll have a simple function call to determien whether a variable supports slicing or not.
Is that correct?

src/client/datascience/commands/commandRegistry.ts
joyceerhl: Just for additional type safety to communicate that that function is not standard and not guaranteed to be implemented by all IDataViewerDataProviders. I should probably also add a comment to that effect. Otherwise future users of this interface may not realize that this function should not be used in showDataViewer codepaths.

src/client/datascience/jupyter/kernelVariables.ts
joyceerhl: Yes that's right.

joyceerhl: Oh sorry, I just noticed. Currently we do not send the `type` property on `IDataFrameInfo` because it's just not necessary, so we must add either a `supportsSlicing` or `type` property. Assuming you're fine with that (we'll still be adding a property), I think that's still better because slicing is a UI operation and the backend shouldn't need to know about that if the UI can figure it out.

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
joyceerhl: Oops, missed this one. ShapeSplitterRegEx is inadequate for this use case--it only works for 2D data, and here we need all the components of the shape in a list, but the regex only generates a capture group for the last number: https://regex101.com/r/0a1nQz/1 

We also can't replace ShapeSplitterRegEx with this function, because ShapeSplitterRegEx exists to match a shape string within a Jupyter inspect result.

joyceerhl:
@DonJayamanne I think I've resolved your feedback. I had to add a `type` property to `IDataFrameInfo`, but now we are doing the check on whether to show the slice control in the UI and not in debuggerVariables/kernelVariables.

src/client/datascience/data-viewing/dataViewer.ts
joyceerhl: @rchiodo I think I understand your question now. You were asking why we have an experiment for a backend enhancement. The answer is that this PR introduces a temporary UI for testing, and this slice UI (as well as the real one in Part 2) needs to be hidden behind a feature flag for controlled rollout. So here we're checking if the user is in the experiment, and forwarding that information to the UI. I believe this is consistent with other places where we retrieve the experiment info in the backend code and pass it through to the UI via the redux store. Does this help?

FWIW I have dropped the `supportsSlicing` check from this function here and moved it to the UI per Don's feedback that we don't need to pass a derived attribute as a property from the backend to the UI.

src/client/datascience/commands/commandRegistry.ts
DonJayamanne: agreed, it could change existing code, 
Though i feel its the right thing to do, as `debugStepOver` is already waiting & the command also completes when its done.
Feels incorrect not to do it... but not something we should address in this PR...
Lets leave it as is

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
joyceerhl: I think the point of confusion might be over the fact that in `IJupyterVariable`, the `name` property never changes, but the `sliceExpression` property does. All we need to resolve a new variable is a `name` and, if provided, a `sliceExpression`. So when asking for variable info or rows based on the name property we are always guaranteed to recover the original variable.

src/client/datascience/data-viewing/dataViewer.ts
DonJayamanne: I agree with Ian, 
Its possible i add a slice expression & then immediately change it to something else.
Its possibel I'd get the first result (even though the cached promise was cleared) & then later I'd get the next restul.

This is generally handled at the UI that makes the request (i.e. ignore the last request sent, kind of how debouncing works in UI when making network calls, throw away previous requests), where we cancel the output from the previous results.

One way of addressing this is to pass a request Id to the extension side.
Then instead of returning the promise, we await on the results in the function. & if the latest request id is the same as the request id for which we have a response, then we send that, else await on the latest promise in the cached property.

Note: But this all feels overly complicated, again, this is generally handled at the UI layer (Redux, ReduxThunk or vanilla JS with debounce solves this)..

I personally don't think we need to implement anything to address this issue.
I.e. I'm happy with what you have today (if we did want to fix it, i'd fix it at the UI end & introduce cancellations to cancel previous requests in backend to save CPU - but not required now, just complicate it).

DonJayamanne: Once again, I don't think we need to fix anything here.

src/client/datascience/jupyter/kernelVariables.ts
DonJayamanne: I think we should send the `type` property, and let other information/behaviours be derived from that information. I.e. `supporsSlicing` can be derived from that information.

DonJayamanne: This way its more consistent, to send type from backend, then to extension, then to UI. instead of augmenting with custom information from extension.

src/client/datascience/commands/commandRegistry.ts
joyceerhl: Wow, sorry, I was thinking of the wrong interface. `getFullVariable` is implemented on `IJupyterVariable`, not `IDataViewerDataProvider`. You're totally right, this function doesn't need to be optional at all because AML will never see it (it's specific to JupyterVariableDataProvider which is our implementation of IDataViewerDataProvider).

src/client/datascience/data-viewing/dataViewer.ts
joyceerhl: I understand the scenario now, thanks for explaining. Whichever UI we settle on, I'll make sure I address pending slice requests in part 2. The request ID approach feels pretty intuitive to me.

src/client/datascience/jupyter/kernelVariables.ts
joyceerhl: One disadvantage of this approach is that the UI now knows about which specific types we need to enable slicing for. That's language-specific. Not a problem now, just one more thing to disentangle if/when we start looking at supporting other languages in the data viewer...

DonJayamanne (approved):

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
rchiodo: I think the problem I have is that there's a two step process and we're caching the second step in the same place as the first. 

First step - get data about the entire data frame (although I guess this would not technically be a dataframe anymore? Cause it might have more than 2 dimensions). Save this information in the this.variable field. I'm assuming this has all of the dimensions and what the UI will use to build the slicing control.

Second step - get data about a specific slice. This will be 2D data. It's being stored in this.variable field as well.

These seem like they're not really related? That maybe there should be two different messages for them? And they should be cached individually?

And for the second step, it should be cached based on the original variable name and the slice expression?

joyceerhl: The reason I'm caching everything in one spot is that I don't see a reason to keep the original variable around. The user asks to see a variable, we fetch the whole variable. The jupyterVariableDataProvider will never be asked for the original variable again, unless the user asks to see the original variable, but the original variable is just one type of slice. We will only ever show the user one slice at a time, so I didn't want to start caching maps of slice expressions to data (although I can also do that if you prefer).

What I have here looks weird because it will result in us always refetching when any sliceExpression is passed, even if the currently cached variable represents the same slice as the caller requested. However this isn't actually an issue because I'm also caching the sliceExpression and the variable together in dataViewer.ts (previously we were already caching the variable), so by the time we actually ask the jupyterVariableDataProvider for info, the slice must have changed.

In any case, would you prefer that I mirror the caching behavior in dataViewer.ts here?

```ts
if (!this.dataFrameInfoPromise || sliceExpression !== this.currentSlice) {
```

rchiodo: > The jupyterVariableDataProvider will never be asked for the original variable again, unless the user asks to see the original variable, but the original variable is just one type of slice. We will only ever show the user one slice at a time, so I didn't want to start caching maps of slice expressions to data (although I can also do that if you prefer).

This is implicit, right? There's nothing in the code you're changing here that guarantees this?

I'm not saying your code won't work, I'm worried about the implicit assumptions about behavior that are in the code.

rchiodo: If we can either prevent a second call for the whole enchilada (in this code isntead of because of how the UI works) or make a separate function that's called to make it super explicit, I'd rather have that.

joyceerhl: Yeah that's fair, nothing guarantees this. 

So you'd prefer a `this.variable`, `this.variableSlice` and a `this.variableSliceExpression`?

rchiodo: I guess I'd prefer two API calls. They both essentially do the same thing but they're obvious they're for different use cases. Under the covers they call the same python code we wrote but in this class there's a two step process. Then we could also throw an exception if the first API is called more than once.

But if you don't agree, feel free to not do it that way. It's really just a way to make the code seem more specific to me. If you don't think the implicit thing is a big deal, that's fine with me.

joyceerhl: I think this was exactly the question I had earlier this week--whether to add a set of new methods or to extend the functionality of existing methods with an optional parameter. I actually [originally implemented](https://github.com/microsoft/vscode-jupyter/pull/4402/commits/d6a250450395f690cad0af018a21ad3fa2e41121) it the way you describe, but thought better of it because I was hesitant to introduce a new set of methods down the call hierarchy when the method implementations would be virtually identical to something that already exists. 

Just to be clear, it sounds like the preference is for the former. Something like 

```ts
getSlice?(variable: IJupyterVariable, sliceExpression: string): Promise<IDataFrameInfo> // Optional because dataViewer.ts consumes IDataViewerDataProvider, so we need to define this method on IDataViewerDataProvider and not on IJupyterVariableDataProvider 
```

And this would then call into `this.variableManager.getDataFrameInfo`? Or would you prefer a separate `this.variableManager.getSlice` as well?

rchiodo: All the way down to the spot where we call the python code. If we weren't caching it wouldn't be as big a deal. But there's logic that's dependent upon the outside world not calling something in the variableManager too.

joyceerhl: So debuggerVariables.ts, kernelVariables.ts, jupyterVariables.ts, jupyterVariableDataProvider.ts/IDataViewerDataProvider are each going to have the following methods added:

```ts
getSliceInfo(
        targetVariable: IJupyterVariable,
        notebook?: INotebook,
        sliceExpression?: string
): Promise<IJupyterVariable>;
getSliceRows(
        targetVariable: IJupyterVariable,
        start: number,
        end: number,
        notebook?: INotebook,
        sliceExpression?: string
): Promise<JSONObject>;
```

And their definitions are going to basically be the same as getDataFrameInfo and getDataFrameRows? That feels kinda wasteful to me, but I'll make that change if you think it's more readable.

rchiodo: Maybe we should just chat.

rchiodo: I think if you got rid of the caching or cached based on slice expression, there would be no need for extra functions

joyceerhl: I guess they could share private `getExpressionInfo` and `getExpressionRows` impl methods, which would be less duplicative... But isn't your original concern about caching the slice in a spot that looks like the variable? Wouldn't another solution be to simply never cache an IJupyterVariable representing a slice, or, as we discussed, cache it as a separate private property? I guess adding a new set of methods feels like overkill for the cache problem.

joyceerhl: Ugh Github. Just saw your reply. Yeah I'll change it to cache based on slice expression.

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
joyceerhl: This code in particular should never fail, because the `shape` attribute comes from the Python code and in fact we validate `shape` in the Python scripts as well. Not convinced this is worth sending new telemetry for.

pythonFiles/vscode_datascience_helpers/getVariableInfo/vscodeGetVariableInfo.py
```diff
@@ -8,16 +8,21 @@ def _VSCODE_getVariableInfo(var):
     result = {}
     result["shape"] = ""
     result["count"] = 0
+    result["type"] = ""
+
+    vartype = type(var)
```
rchiodo, line 13: Can this crash? Or does type always return something?

rchiodo: We used to check for TypeError

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -290,10 +289,12 @@ export class DebuggerVariables extends DebugLocationTracker
         );
         if (results && results.result) {
             // Results should be the updated variable.
+            const info = JSON.parse(results.result);
```
rchiodo, line 292: Was this just for debugging purposes? This seems equivalent to what we had before.

src/datascience-ui/data-explorer/reactSlickGrid.tsx
```diff
@@ -265,10 +282,7 @@ export class ReactSlickGrid extends React.Component<ISlickGridProps, ISlickGridS
             // Save in our state
             this.setState({ grid, fontSize });
         }
-
-        // Act like a resize happened to refresh the layout.
```
rchiodo, line 269: This change seems unnecessary? It's functionally equivalent.

rchiodo (changes requested):
Still waiting on the possible TypeError being thrown in vscodeGetVariableInfo.py

src/client/datascience/jupyter/debuggerVariables.ts
joyceerhl: Yeah just for debugging purposes, reverting

src/datascience-ui/data-explorer/reactSlickGrid.tsx
joyceerhl: I had extracted some grid setup functionality into a `createSlickGrid` method to setup the grid on receiving a slice, but found this didn't work (caused problems with header row rerendering when setting columns). Submitting a new approach

joyceerhl:
@rchiodo I think I've gotten to your comments, LMK if otherwise

rchiodo (approved):

Issue #4369, opened 2021-01-13
Native Notebooks - Stop execution (cell and kernel level) don't work


claudiaregio:
1. Run any long running cell
2. Try to stop the execution (either at the cell level or notebook kernel level) 

Nothing happens, have to close and reopen the notebook

DonJayamanne:
@claudiaregio please could you provide a sample code to replicate this issue.
If possible, try to provide a simple sample so we can test/setup easily at our end.

Also, please confirm this is not an issue when you try the same in classic Jupyter (browser).

rchiodo:
This is expected (as evidenced here:)
https://github.com/ipython/ipython/issues/3400

However we can make native notebooks automatically restart if they take too long like we do for the old editor.

I'll also add more logging to make sure we're interrupting if this comes up again.

greazer:
Looks like this belongs in validate.

DonJayamanne:
I couldn't replicate the original issue, hence leaving for someone else to validate.

rchiodo:
To repro you have to run with debug bits and uninstall the json rpc stuff. Then the extra layer for interrupt doesn't work.

claudiaregio:
I have had all diary study participants mention this same behavior so far and had it happen again during a wrap up interview. Can pull the footage to share on Monday.

joyceerhl:
This is still a problem at least on Windows 

![image](https://user-images.githubusercontent.com/30305945/106099066-eecfa280-60ee-11eb-9ed8-b420414344d9.png)

![image](https://user-images.githubusercontent.com/30305945/106099092-fd1dbe80-60ee-11eb-9b29-6af252762990.png)

rchiodo:
Joyce, it looks you just validated it. That's the intended behavior when the interrrupt fails. Jupyter is not guaranteed to interrupt so we ask for a restart if it doesn't work.

rchiodo:
For me this always happens with a sleep if you don't have the json rpc library installed (dev bits). But it can happen for other code as well. See the ipython bug:
ipython/ipython#3400

joyceerhl:
Oh, good. Closing as validated then :)

PR #4414, opened 2021-01-15
Make a timeout for interrupt ask for a restart

rchiodo:
For #4369 

The jupyterNotebook class was being used for interrupting. The problem is the jupyterNotebook assumes it was the one that started an execution and waits for the last pending execution to finish before saying in interrupt succeeded. Since there wasn't any, interrupt returns super fast and doesn't wait for the keyboard exception.

Changed interrupt to happen in the cellExecution class instead. 

Also added an interruptTimedOut event because interrupt doesn't actually go through the notebook editor (where the code lives to show a restart if a timeout occurs). This allows the UI for this to remain in the notebook editor.

Interrupt is working for me in native notebooks now but only if I make sure the jsonrpc stuff is setup. 


DavidKutu (approved):

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -118,6 +120,7 @@ export class CellExecution {
     private disposables: IDisposable[] = [];
     private cancelHandled = false;
     private requestHandlerChain = Promise.resolve();
+    private activeExecutions: { execution: Promise<void>; session: IJupyterSession }[] = [];
```
IanMatthewHuff, line 123: I've not played around with new native execution much, so this might just be a misunderstanding on my part (looking over it more now). But I thought that a new CellExecution was created each time we executed a notebook cell. So why would we be getting multiple active executions on one CellExecution?

rchiodo: The way the code is structured it's possible? There's nothing that prevents (from the CellExecution class) there being multiple executions for the same cell. But the calling class does prevent that.

DonJayamanne (changes requested):
Will look into this, please note we have tests today to ensure users are presented with an option to restart the kernel when interrupt fails.

src/test/datascience/notebook/interrupRestart.vscode.test.ts

Don'nt understand the issue, will read that soon & review the code..

DonJayamanne:
> However we can make native notebooks automatically restart if they take too long like we do for the old editor.

We have tests that do this today. So not sure what this is fixing.

rchiodo:
> > However we can make native notebooks automatically restart if they take too long like we do for the old editor.
> 
> We have tests that do this today. So not sure what this is fixing.
> Tests are here src/test/datascience/notebook/interrupRestart.vscode.test.ts

The test passes if the interrupt succeeds. The old code would ALWAYS think it succeeded while the cell UI would actually still be running.

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -165,6 +168,23 @@ export class CellExecution {
         );
     }
 
+    public async interrupt(timeoutMs: number): Promise<InterruptResult> {
+        // Go through our active executions and interrupt each one.
+        const results = await Promise.all(
```
DonJayamanne, line 173: We might want to ensure this method cannot be called more than once & if already completed.
I.e. `if this._interrupted || this._completed) { return false}`
Where ``this._interrupted` is a promise that contains the result for the previous `interrupt` call for this cell execution (i.e. all interrupts will be doing the same thing - pointing to the same promise)

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -394,7 +470,9 @@ export class CellExecution {
     private async execute(session: IJupyterSession, loggers: INotebookExecutionLogger[]) {
         const code = this.cell.document.getText();
         traceCellMessage(this.cell, 'Send code for execution');
-        return this.executeCodeCell(code, session, loggers);
+        const execution = this.executeCodeCell(code, session, loggers);
+        this.activeExecutions.push({ execution, session });
```
DonJayamanne, line 474: Might want to clear this property in the dispose method.

src/client/datascience/jupyter/kernels/cellExecution.ts
DonJayamanne: & once the promise resolves we might want to clear that propery.
This way if user clicks multiple times, they will get the samae interrupt promise.
After user gets the interrupt error message if they try again, it should start a new interrupt sequence (without using the same - previously resolved promise).

DonJayamanne:
> The old code would ALWAYS think it succeeded while the cell UI would actually still be running.

Not sure about this, as it works for me.
Anyways, like the new code. Approving it.

src/client/datascience/jupyter/kernels/cellExecution.ts
DonJayamanne: I don't see how we can have more than one item in the array, if we do its incorrect or i'm mistaken.
Can we not have an array instead. As we create new instances of this class for each execution (even if we run the same cell multiple times).

DonJayamanne (changes requested):
Minor issues, 
* Single item in array
* Cache promise for interrupt (if user clicks interrupt multiple times)

src/client/datascience/jupyter/kernels/cellExecution.ts
rchiodo: Good idea.

rchiodo:
@DonJayamanne can you have another look?

DonJayamanne (approved):

Issue #298, opened 2020-10-26
Add support for multidimensional data in data viewer


joyceerhl:
- Modify vscodeGetDataFrameRows to convert extra dimensions to strings
- Truncate long values UI side
- When user clicks into a cell, display the full flattened element

joyceerhl:
To validate this one please open manualTestFile.py and verify that you can open `myTensor` in the data viewer via the following aveneues:
1. Variable view
2. Debugging cell in interactive window
2. Debug current Python file --> debug variables panel --> View Variable in Data Viewer

rchiodo:
Validated

PR #4416, opened 2021-01-16
Support multidimensional tensors in data viewer

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/298

Behavior changes:
1. All values will now be truncated past 10 characters per @jmew
2. Higher dimensional data will be represented as flattened strings with the ability to view the full value
3. Full view is readonly

Note, I wasn't able to bind esc/enter to exit and enter edit mode with `autoEdit: false`--there's a weird bug where after exiting edit mode, focus gets lost. If we allow auto edits as soon as a cell receives focus (current behavior, see GIF below for illustration), this preserves the custom cell navigation behavior implemented in our keydown handlers. I'll log that as a separate bug since I think investigating that is going to be somewhat involved.

tabbing

Multidimensional tensors:
image

Ragged ndarrays:
image

Tooltip contains the full value:
image


pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -1,11 +1,57 @@
...
+        pass
+
+    if y_len:
+        # Figure out what kind of object the rows are
+        flattened = _VSCODE_np.full((x_len, y_len), "", dtype="object")
+        for i in range(x_len):
+            for j in range(len(temp[i])):
+                element = temp[i][j]
+                if isinstance(element, _VSCODE_np.ndarray):
+                    stringified = _VSCODE_np.array2string(element, separator=", ")
```
rchiodo, line 37: I thought you found this was super slow and you were going to do this in the UI?

rchiodo (approved):

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
joyceerhl: No, I found that lazy fetching details of individual cells was too slow (2 seconds for long values). Here I'm requesting the full string and then doing all the truncation in the UI.

src/datascience-ui/data-explorer/cellFormatter.tsx
```diff
@@ -74,3 +72,12 @@ export function cellFormatterFunc(
 ): string {
     return ReactDOMServer.renderToString(<CellFormatter value={value} columnDef={columnDef} />);
 }
+
+function truncate(value: string) {
+    // This should be based on font size and type
+    if (value.length <= 10) {
+        return value;
+    } else {
+        return value.slice(0, 7).concat('...');
```
joyceerhl, line 81: This part here is where we do the truncation now @rchiodo

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
rchiodo: Oh so you're concatenating the value here (instead of sending the full json).

rchiodo: Sorry I though the slowdown in the UI meant you were going to do the concat there as well.

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -1,11 +1,57 @@
...
+                else:
+                    stringified = element
+                flattened[i][j] = stringified
+        temp = flattened
+    temp = _VSCODE_pd.DataFrame(temp)
+    ndarray = temp
+    del temp
+
+    # Restore the user's printoptions
+    _VSCODE_np.set_printoptions(threshold=current_options["threshold"])
```
IanMatthewHuff, line 51: Should we possibly have everything in this function in a try and then restore this in a finally? Just worried about not restoring users print options.

IanMatthewHuff (approved):
Looks legit to me. Though I have to confess that I'm pretty weak with Python :D. Just one comment regarding making sure that we restore user's print options.

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
joyceerhl: Bit confused. By 'concatenating' do you mean transforming the data from an array to a string?

Just to be sure we're on the same page, the intent in this bit of code is to generate the full string representation of all data past 2 dimensions, without any truncation, over to the UI. Then in the cellFormatter I'm displaying all values with a trailing ellipsis, so we only render the first 10 characters of any value. We display the full untruncated value later.

In a separate PR I'll also implement pagination like we talked about, which should minimize the performance impact of sending untruncated strings to the UI.

rchiodo: Yes I meant transforming the data from an array to a string. Theoretically the json for the array isn't much more than the json for the string so I thought you were just going to send the array itself over. But maybe that is significantly slower.

And I believe you'll be throwing this code away when you implement slicing?

joyceerhl: > I thought you were just going to send the array itself over

I think this doesn't really work for >3D data. Python objects know how to represent themselves as strings in Python code, but we can't get the string representation out of them in TypeScript. At minimum we'd have to recursively convert Python objects like Tensors or numpy ndarrays to arrays, and then I'm not sure how we'd reconstitute nested arrays.

> you'll be throwing this code away when you implement slicing

Unfortunately not, because the slice data view as currently spec-ed only applies to 3D data. We've talked to Jill about reworking the design, but I believe the current plan is to maybe extend it to work for 2D or 3D data. 

If we made the slice data view work for arbitrary dimensions and selected default dimensions for higher dimensional arrays, then yes we'd be able to deprecate this code. Jeffrey and I have discussed whether the latter would be confusing for users (e.g. they might think the data is 2D because we default them to a flat view with a preselected slice). So the current plan is to show the flattened view (implemented in this PR) by default, and then also provide a way to turn on slicing mode.

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
joyceerhl: Ah you're right. I was trying to minimize my use of expansive try/catch blocks (got feedback from Eric on that in my TypeScript code 😅)

FWIW this Python code all needs to move to the Python extension. But given the hard deadline on this work for March I'm planning to get to that refactor afterwards, since that involves designing an API for providing debug info for variable views as well as the data viewer.

IanMatthewHuff: Interesting, was Eric's feedback python specific or just in general? I would assume that an exception here would be a true exceptional case (i.e. not something that we know will happen from time to time) so in the non-exception thrown case I don't think this would be a perf issue.

joyceerhl: His feedback was general and in the context of regular TypeScript code, so I don't know that it applies here. But I wrote this particular catch block accounts with something that will happen from time to time in mind. I think there's a way to handle that error without incurring the exception, but we would still want to avoid bombing out on an unexpected error here.

joyceerhl:
Just updated this PR to allow users to view and scroll the expanded view of the cell using left/right arrows. I added a GIF to the issue description if y'all want to take a look @IanMatthewHuff and @rchiodo.

Issue #254, opened 2020-10-19
Remote jupyter server connections for Native Notebook


greazer:

DonJayamanne:
* For GA at a minimum get feature parity
	* Get estimate
* Nice to have for experiment (based on estimate)

DonJayamanne:
Notes from https://github.com/microsoft/vscode-jupyter/issues/4328

Here's a temporary solution outlined by @ayim (temporary = Old notebook experience in Native notebooks without the full blow remote experience)


The Jupyter Server entry point that used to be located here: 
![image](https://user-images.githubusercontent.com/10951869/103951208-1dabb780-50f3-11eb-997e-6a39646dc06a.png)

Will be moved to the global status bar in Native Notebooks since the original entry point is no longer available.
![image](https://user-images.githubusercontent.com/10951869/103958685-79ca0800-5102-11eb-8e9a-784bfc7e73c8.png)

All functionality should remain the same - this will help achieve experience parity in native notebooks.

claudiaregio:
https://github.com/microsoft/vscode-jupyter/issues/1206

Use as inspo

ayim:
### Parity UI/UX experience

Prototype Link (all users with MS SSO should be able to sign in): https://www.figma.com/file/OGTDR9c9U5sFemy6xMlaE2/Visual-Studio-Code-Toolkit-Community?node-id=91%3A627

**Screen: Local**
![Local](https://user-images.githubusercontent.com/10951869/105392940-a2590400-5bd0-11eb-91b5-0b4425a3f13e.png)
**Screen: Local / Connections**
![Local _ Connections](https://user-images.githubusercontent.com/10951869/105392947-a38a3100-5bd0-11eb-9786-6a41abfe4f5d.png)
**Screen: Local / Connections / URI**
![Local _ Connections _ URI](https://user-images.githubusercontent.com/10951869/105392951-a4bb5e00-5bd0-11eb-8b63-5d54f897596c.png)
**Screen: Remote**
![Remote](https://user-images.githubusercontent.com/10951869/105392965-a84ee500-5bd0-11eb-96e0-c4940bc18849.png)
**Screen: Remote/Connections**
![Remote _ Connections](https://user-images.githubusercontent.com/10951869/105392969-a9801200-5bd0-11eb-8c3c-200b7fd4c8f7.png)

**Prototype in Image Form**
**Local Flow:**
[...]

PR #4424, opened 2021-01-20
Add statusbar + toolbar item for selecting jupyter

DonJayamanne:
For #254
* Adding an item in the statusbar and toolbar for native notebooks
* Changing existing command so we can track how the UI is displayed (whether from command, toolbar, statusbar).
*


package.nls.json
```diff
@@ -78,6 +78,8 @@
     "jupyter.command.jupyter.showhistorypane.title": "Show Interactive Window",
     "jupyter.command.jupyter.createnewinteractive.title": "Create Interactive Window",
     "jupyter.command.jupyter.selectjupyteruri.title": "Specify local or remote Jupyter server for connections",
+    "DataScience.jupyterNativeNotebookUriStatusLableForLocal": "Jupyter Server: local",
```
IanMatthewHuff, line 81: StatusLabel

src/client/common/application/commands.ts
```diff
@@ -149,4 +148,7 @@ export interface ICommandNameArgumentTypeMapping extends ICommandNameWithoutArgu
     [DSCommands.SubmitGitHubIssue]: [];
     [DSCommands.ShowDataViewer]: [IShowDataViewerFromVariablePanel];
     [DSCommands.ClearSavedJupyterUris]: [];
+    [DSCommands.SelectJupyterURI]: [undefined, 'toolbar' | undefined];
+    [DSCommands.SelectNativeJupyterUriFromStatusBar]: [];
+    [DSCommands.SelectNativeJupyterUriFromToolBar]: [];
```
IanMatthewHuff, line 153: We seem to have a set of commands with arguments and a set without. Seems like these two new ones don't have arguments. But...does this matter? Not sure if I see a reason for the distinction.

IanMatthewHuff: Also, why | undefined? Isn't 'commandPalette' the other option?

IanMatthewHuff (approved):
There are a couple of small cleanup issues (spelling and command arguments) but code looks legit so approving with cleanup.

src/client/datascience/notebook/remoteSwitcher.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+    private async updateStatusBar() {
+        if (!this.notebook.activeNotebookEditor) {
+            this.statusBarItem.hide();
+            return;
+        }
+        this.statusBarItem.show();
+        const uri = await this.serverUriStorage.getUri();
+        const label =
+            uri === Settings.JupyterServerLocalLaunch
```
rchiodo, line 66: For remote we're not showing the URI? It's just going to say 'jupyter server: remote'?

DonJayamanne: Yes, else we end up taking a whole lot of space in the status bar.
Ooops, I'm meant to display the Uri in the tooltip.

src/test/datascience/jupyter/serverSelector.unit.test.ts
```diff
@@ -91,7 +91,7 @@ suite('DataScience - Jupyter Server URI Selector', () => {
 
     test('Local pick server uri', async () => {
         const { selector, storage } = createDataScienceObject('$(zap) Default', '', true);
-        await selector.selectJupyterURI(true);
+        await selector.selectJupyterURI(true,);
```
joyceerhl, line 94: Typo or required by linters?

src/client/common/application/commands.ts
DonJayamanne: This is because `SelectJupyterUri` command can only be invoked from the command palette in which case the argument is `undefiend`.
The only other place its used is Interactive & WebView notebooks where we pass the argument `toolbar` to indicate its been invoked from the interactive/webview notebook toolbar.

Other than that there are no places where this command is invoked from.

We have similar commands that do the exact same thing, launch the UI. However we have separate commands so we can uniquely identify how and where those commands are launched. Pity we cannot use the same command with arguments (in VSC toolbar and statusbars).
I know in treeviews we can use commands with args.

DonJayamanne: Hope that makes sense. The complication comes from the fact that we want to identify the differnt places the same command is launched, hence duplicate commands.

DonJayamanne: Oh wait, just found we can use args in the statusbar.

DonJayamanne: Fixed by removing one of the duplicate commands.

src/test/datascience/jupyter/serverSelector.unit.test.ts
DonJayamanne: Thanks, fixed.

Issue #1215, opened 2020-08-04
Support non-raw (start Jupyter Server) kernels for VSC Native Notebooks


DonJayamanne:
Today it only supports raw kernels, not tested with Jupyter (ie. spin up jupyter and connect to it using the jupyter API)

DonJayamanne:
Closing in favour of #254

DonJayamanne:
Oops, closed the wrong issue here

DonJayamanne:
Most of it has been supported, just added some tests

rchiodo:
Validated. Used 
```json
"jupyter.disableZMQSupport": true
``` 
to test

PR #4431, opened 2021-01-20
Fixes & tests for non-raw kernels

DonJayamanne:
For #1215

Most of it has been supported, just added some tests


src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -222,19 +223,28 @@ export class Kernel implements IKernel {
         }
         const key = uri.toString();
         if (!this.kernelValidated.get(key)) {
+            this.isRawNotebookSupported =
+                this.isRawNotebookSupported || this.rawNotebookSupported.isSupportedForLocalLaunch();
+
             const promise = new Promise<void>((resolve) =>
-                this.kernelSelectionUsage
-                    .useSelectedKernel(kernel?.kernelConnectionMetadata, uri, 'raw')
```
DonJayamanne, line 227: Only place where we hardcoded `raw`

src/test/datascience/notebook/executionService.vscode.test.ts
```diff
@@ -265,7 +265,7 @@ suite('DataScience - VSCode Notebook - (Execution) (slow)', function () {
         );
 
         // Interrupt the kernel).
-        await commands.executeCommand('notebook.cancelExecution');
+        await commands.executeCommand('jupyter.notebookeditor.interruptkernel');
```
DonJayamanne, line 268: The VSC command doesn't work, i guess its an internal command and not meant to be invokved.
Even while debugging it doesn't always work (used our command that is invoked from the Command Palette.

.github/workflows/main.yml
```diff
@@ -698,10 +698,11 @@ jobs:
           VSC_PYTHON_FORCE_LOGGING: 1
           VSC_JUPYTER_CI_RUN_NON_PYTHON_NB_TEST: 1
           VSC_JUPYTER_REMOTE_NATIVE_TEST: ${{ contains(matrix.test-suite, 'remote') }}
+          VSC_JUPYTER_NON_RAW_NATIVE_TEST: ${{ contains(matrix.test-suite, 'nonRawNotebook') }}
```
DonJayamanne, line 701: Just like remote, run native notebook tests with real jupyter instead of Raw kernels.

rchiodo (approved):

Issue #4153, opened 2020-12-08
Failing to prompt to install ipykernel when starting kernel with conda env


joyceerhl:
1. Install miniconda
2. Create a new env without anything installed in it
3. Create a new blank native notebook
4. Try to run any code cell
5. We simply timeout waiting for a heartbeat without ever asking the user to install data science dependencies
6. If you try to run a code cell again you get an uninformative 'Canceled' error prompt

For the old webview notebooks I do get the prompt when first creating a notebook (maybe because we're prewarming kernels there and can't do that for native notebooks), but if I use the kernel selector to switch to an env that doesn't have ipykernel installed and then try to run code, I get the same timeout problem.

DonJayamanne:
Validate if this has been fixed

rchiodo:
This has not been fixed. I still repro it.

rchiodo:
@DonJayamanne I'm guessing your latest kernel change should fix this? Or did you already submit it?

DonJayamanne:
When I get a prompt to install IPykernel I clicked no & I get two messages, one error & one warning.
Ideally I should get just one message, not multiple:

See below screen

![Screen Shot 2021-01-22 at 15.19.35.png](https://images.zenhubusercontent.com/59df9ec4b0222d5de47b6b0f/32527bb6-e126-4a66-8ccc-aecdbe6f114b)shot

DonJayamanne:
Wait , let me test if this was the original behaviro.

PR #4433, opened 2021-01-20
Kernel launcher dependency check

rchiodo:
For #4153 

First step in the refactor of how dependencies are installed. This makes the kernel launcher check and install dependencies.

Also updated the test to check non jupyter kernels will also install dependencies.


src/client/datascience/kernel-launcher/kernelLauncher.ts
```diff
@@ -140,4 +184,32 @@ export class KernelLauncher implements IKernelLauncher {
...
+                this.kernelDependencyService.installMissingDependencies(interpreter, cancelToken)
+            );
+            this.dependencyPromises.set(interpreter.path, deferred);
+        }
+
+        // Get the result of the question
+        try {
+            const result = await deferred.promise;
+            if (result !== KernelInterpreterDependencyResponse.ok) {
+                throw new Error(
```
DonJayamanne, line 204: Can we have a custom error, this way we can deal with this in a different way (higher up) if required.
E.g. `IPyKernelNotInstalledError`
Also feel having custom errors is better than just `Error`, feels too generic.

Only suggesting so that we can have special code tomorrow if required, and feels like a good practice

DonJayamanne: Not a mandatory change.

DonJayamanne: On second thought I thinmk we'll need it, so that we can ignore this error higher up.

rchiodo: No it's not necessary for ignoring

src/client/datascience/kernel-launcher/kernelLauncher.ts
```diff
@@ -84,6 +92,42 @@ export class KernelLauncher implements IKernelLauncher {
...
+            ]),
+            timeout
+        );
+
+        if (result instanceof CancellationError) {
+            throw result;
+        }
+
+        if (!result) {
+            throw new TimedOutError(localize.DataScience.rawKernelProcessNotStarted());
```
DonJayamanne, line 124: This is a timeout error or `KernelLaunchError`. There's no timeout component here, its just that the kernel didn't launch & we don't know why. Isn't it.

DonJayamanne: Please ignore the above.

rchiodo: No if the result is undefined it's a timeout error (waitForPromise returns 'null' on timeout). Kernel launch failure will throw an exception. this.launchProcess always returns an object.

rchiodo: I'll add a comment to that effect

src/client/datascience/kernel-launcher/types.ts
```diff
@@ -15,8 +15,10 @@ export const IKernelLauncher = Symbol('IKernelLauncher');
 export interface IKernelLauncher {
     launch(
         kernelConnectionMetadata: KernelSpecConnectionMetadata | PythonKernelConnectionMetadata,
+        timeout: number,
         resource: Resource,
-        workingDirectory: string
+        workingDirectory: string,
+        cancelToken?: CancellationToken
```
DonJayamanne, line 21: Nice...

src/client/datascience/notebook/kernelWithMetadata.ts
```diff
@@ -99,9 +99,16 @@ export class VSCodeNotebookKernelMetadata implements VSCNotebookKernel {
...
-    private chainExecution(next: () => Promise<void>): Promise<void> {
-        const prev = this.pendingExecution ?? Promise.resolve();
-        this.pendingExecution = prev.then(next);
+    private async chainExecution(next: () => Promise<void>): Promise<void> {
+        if (this.pendingExecution) {
+            try {
+                await this.pendingExecution;
+            } catch (e) {
+                // Errors are handled elsewhere
+                traceInfo(`Kernel execution previous failure: ${e}`);
```
DonJayamanne, line 108: Is this correct? Doesn't this mean, if you have 5 cells & first cell crashes, then the rest of the 4 cells will execute? that seems to be a change to how things should behave.
I thought in classic jupyter, it stops execution when a cell fails.

rchiodo: That's not what is happening here. The pendingExecution will handle the errors itself. This prevents errors from previous runs from not running anything which is what the old code would do.

src/client/datascience/raw-kernel/rawJupyterSession.ts
```diff
@@ -86,50 +85,41 @@ export class RawJupyterSession extends BaseJupyterSession {
...
-                this.kernelConnectionMetadata = newSession.kernelProcess?.kernelConnectionMetadata;
-
-                this.outputChannel.appendLine(
-                    localize.DataScience.kernelStarted().format(
-                        getDisplayNameOrNameOfKernelConnection(this.kernelConnectionMetadata)
-                    )
-                );
+                // Send our telemetry event with the error included
+                sendTelemetryEvent(Telemetry.RawKernelSessionStartException, undefined, undefined, error);
+                traceError(`Failed to connect raw kernel session: ${error}`);
```
DonJayamanne, line 121: I think having `IPYkernelNotInstalled` error or the like is useful here.
Basically user cancelled it (see lines 109 & 110)
We know user cancelled it. & not installing ipykernel is the same, meaning user has cancelled the start of the kernel.

src/client/datascience/raw-kernel/rawJupyterSession.ts
```diff
@@ -86,50 +86,41 @@ export class RawJupyterSession extends BaseJupyterSession {
...
+
+            // Update kernelspec and interpreter
+            this.kernelConnectionMetadata = newSession.kernelProcess?.kernelConnectionMetadata;
+
+            this.outputChannel.appendLine(
+                localize.DataScience.kernelStarted().format(
+                    getDisplayNameOrNameOfKernelConnection(this.kernelConnectionMetadata)
+                )
+            );
+        } catch (error) {
```
IanMatthewHuff, line 107: Seems good, we can keep all the same telemetry and stuff at this level

IanMatthewHuff: At least for now

src/client/datascience/raw-kernel/rawJupyterSession.ts
DonJayamanne: Else we log the wrong telemetry & wrong message.
We could optionally log telemetry indicating user opted not to install hence startup failed.

src/client/datascience/raw-kernel/rawJupyterSession.ts
```diff
@@ -212,7 +198,7 @@ export class RawJupyterSession extends BaseJupyterSession {
             // Need to have connected before restarting and can't use a LiveKernelModel
             throw new Error(localize.DataScience.sessionDisposed());
         }
-        const startPromise = this.startRawSession(kernelConnection, cancelToken);
+        const startPromise = this.startRawSession(kernelConnection, -1, cancelToken);
```
DonJayamanne: Can we use `undefined`, this way its more clear. Using -1 or 0 or the like feels odd.
E.g. in .net i know we use TimeSpan.infinity of the like for such things.

src/client/datascience/raw-kernel/rawJupyterSession.ts
```diff
@@ -230,17 +217,13 @@ export class RawJupyterSession extends BaseJupyterSession {
         ) {
             throw new Error(`Unable to start Raw Kernels for Kernel Connection of type ${kernelConnection.kind}`);
         }
-        const cancellationPromise = createPromiseFromCancellation({
```
DonJayamanne, line 233: Nice to move this lower down

src/client/datascience/kernel-launcher/kernelLauncher.ts
```diff
@@ -84,6 +92,42 @@ export class KernelLauncher implements IKernelLauncher {
...
+    ): Promise<IKernelProcess> {
+        // If this is a python interpreter, make sure it has ipykernel
+        if (kernelConnectionMetadata.interpreter) {
+            await this.installDependenciesIntoInterpreter(kernelConnectionMetadata.interpreter, cancelToken);
+        }
+
+        // Should be available now, wait with a timeout
+        const result = await waitForPromise(
+            Promise.race([
+                this.launchProcess(kernelConnectionMetadata, resource, workingDirectory),
```
DonJayamanne, line 109: Shouldn't we also pass the `token` into the `launchProcess`.
If the user cancels, then we should kill the process as well.
Same applies to `timeout`.
If the time span has exceeded, then we should grafefully dispose the kernelProcess that we were attempting to start.

More of a cleanup (since you area cleaning this part).

src/client/datascience/raw-kernel/rawJupyterSession.ts
rchiodo: That code isn't there. You're commenting on the older version of the file.

src/client/datascience/kernel-launcher/kernelLauncher.ts
rchiodo: Good idea. I'll investigate. We should dispose the kernelProcess if the cancel happened or the timeout occurred.

src/test/datascience/jupyter/kernels/installationPrompts.vscode.test.ts
```diff
@@ -91,54 +96,68 @@ suite('DataScience Install IPyKernel (slow) (install)', function () {
     });
 
     test('Ensure prompt is displayed when ipykernel module is not found and it gets installed', async () => {
-        const installed = createDeferred();
+        // Do this for both kernels
+        await Promise.all(
+            ['.venvnokernel', '.venvnoreg'].map(async (kName) => {
```
DonJayamanne, line 101: Doesn't this mean we're running both together? I.e. opening two notebooks side by side & running them for both kernels?
Did you instead mean to put this for loop outside the test & run each test seprately.

DonJayamanne (dismissed):
Few minor changes

src/client/datascience/notebook/kernelWithMetadata.ts
DonJayamanne: Got it thanks.

src/client/datascience/kernel-launcher/kernelLauncher.ts
DonJayamanne: I never did like the waitForPromise. ideally i'd like that to throw an error instead of always checking the return value. This way its easier, else everyhere this is used we have to do the same thing... I think ther'es an issue open to fix how waitForPromise works.

DonJayamanne: Anyways, not nothing that needs to be fixed now.

src/client/datascience/kernel-launcher/kernelLauncher.ts
DonJayamanne: Again, not necessary, but a good cleanup (cleaning up resources instead of leaving these dangling processes).

src/client/common/cancellation.ts
```diff
@@ -14,6 +14,12 @@ export class CancellationError extends Error {
         super(localize.Common.canceled());
     }
 }
+
+/**
+ * Error type thrown when a timeout occurs
+ */
+export class TimedOutError extends Error {}
```
IanMatthewHuff, line 21: This is a bit of an odd spot for this class, it's not cancellation related. If it's specific to the kernel launcher I'd put it as KernelTimedOutError and put it in the class that throws it. Or if it's general, maybe errorUtils or a types.ts file?

src/test/datascience/jupyter/kernels/installationPrompts.vscode.test.ts
rchiodo: Yeah that's what I intended. Not in parallel

src/client/datascience/kernel-launcher/kernelLauncher.ts
IanMatthewHuff: Could make it explicit with === null, not just a falsy check. I see that most of the other locations that we use waitForPromise (jupyterNotebook, rawJupyterSession) we do an explicit null check for the timeout case. It works now with the falsy check, but if something was refactored to possibly return undefined then it would go down the wrong path here.

IanMatthewHuff (approved):
Had a comment or two, but seems legit.

src/client/datascience/kernel-launcher/kernelLauncher.ts
rchiodo: This isn't being used anymore so don't have to worry about it. I agree we should change waitForPromise to reject on timeout instead of returning null

src/client/datascience/raw-kernel/rawJupyterSession.ts
rchiodo: Should be fixed now.

src/client/datascience/kernel-launcher/kernelLauncher.ts
rchiodo: Should be lower down now. Timeout will happen waiting for the heartbeat. No longer need the waitForPromise stuff.

src/test/datascience/jupyter/kernels/installationPrompts.vscode.test.ts
rchiodo: Test isn't passing. Can't get Peng's command to work. Going to disable for now.

src/client/common/cancellation.ts
```diff
@@ -14,6 +14,7 @@ export class CancellationError extends Error {
         super(localize.Common.canceled());
     }
 }
+
```
IanMatthewHuff, line 17: Revert just to not have unneeded history touch.

src/client/datascience/jupyter/kernels/kernelDependencyService.ts
```diff
@@ -71,6 +71,8 @@ export class KernelDependencyService implements IKernelDependencyService {
             ]);
             if (response === InstallerResponse.Installed) {
                 return KernelInterpreterDependencyResponse.ok;
+            } else if (response === InstallerResponse.Ignore) {
```
IanMatthewHuff, line 74: From the code it looks like ignoring or closing the prompt also logs this. In that case I would expect a KernelInterpreterDependencyResponse.cancel not .failed. Can we distinguish the two?

IanMatthewHuff: I guess we can't any non-ok response the same, so maybe ok as is.

IanMatthewHuff (approved):

src/client/common/cancellation.ts
rchiodo: Linter won't let me ;(

rchiodo: Two lines aren't allowed

src/client/datascience/jupyter/kernels/kernelDependencyService.ts
rchiodo: I don't believe it returns that value when canceling. It returns InstallerResponse.Disabled. Yeah verified that with the debugger. It only returns that when it fails.

IanMatthewHuff: Ok, that's odd. I believe the debugger. Just in line 164 in productInstaller.ts it looks like a 'No' response from the prompt would lead to disable.

rchiodo: It doesn't go thorugh the productInstaller's prompt. The kernel dependency stuff uses its own.

src/client/datascience/baseJupyterSession.ts
```diff
@@ -165,18 +169,18 @@ export abstract class BaseJupyterSession implements IJupyterSession {
         this.session?.statusChanged.connect(this.statusHandler); // NOSONAR
 
         // Start the restart session promise too.
-        this.restartSessionPromise = this.createRestartSession(kernelConnection, newSession);
+        this.restartSessionPromise = this.createRestartSession(kernelConnection, newSession, timeoutMS);
     }
 
-    public async restart(_timeout: number): Promise<void> {
+    public async restart(timeout: number): Promise<void> {
```
DonJayamanne, line 175: Yay, nice to see these getting used. Code is looking cleaner.
Thanks.

src/client/datascience/kernel-launcher/kernelLauncher.ts
```diff
@@ -99,17 +126,34 @@ export class KernelLauncher implements IKernelLauncher {
             this.extensionChecker,
             this.kernelEnvVarsService
         );
-        await kernelProcess.launch(workingDirectory);
+        await kernelProcess.launch(workingDirectory, timeout, cancelToken);
+
+        // Double check for cancel
+        if (cancelToken?.isCancellationRequested) {
```
DonJayamanne, line 132: Is this required here.
I would expect cancellation to be checked before we do somethign that CPU intensive of the like.
I.e. if something has already happened and user requested a caancellation, i don't see the need to throw an error.
I'd expect a higher level to handle that scenario.

DonJayamanne (approved):

src/client/datascience/kernel-launcher/kernelLauncher.ts
DonJayamanne: Nothing to fix here.

Issue #1191, opened 2020-09-10
Jupyter autocomplete needs to be wired into native notebooks.


rchiodo:
Basically registerCompletionProvider for notebook cell schema and use the kernel to get autocomplete data.

This is independent of getting the LS to understand notebooks.

rchiodo:
Validated. Loaded a dataframe and the columns showed up in the native editor.

PR #4434, opened 2021-01-20
Jupyter completions in native notebooks

DonJayamanne:
For #1191

Test is passing locally. On CI Jedi seems to be crapping out in Jupyter. need to check install scripts or other.


src/test/datascience/notebook/intellisense/completionProvider.vscode.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        await insertCodeCell('a.', { index: 1 });
+        const cell2 = vscodeNotebook.activeNotebookEditor!.document.cells[1];
+
+        const token = new CancellationTokenSource().token;
+        const position = new Position(0, 2);
+        const context: CompletionContext = {
+            triggerKind: CompletionTriggerKind.TriggerCharacter,
+            triggerCharacter: '.'
+        };
+        const completions = await completionProvider.provideCompletionItems(cell2.document, position, token, context);
```
DonJayamanne, line 86: This is the same completion provider we hook up with VS Code, i.e. VSCode will use this API.
Hence testing this is sufficient (also this is how VS Code recommends testing such code).
Unless we do full blown UI test where we launch an editor & click it etc, which feels unnecessary as that's VSC codes job and we're provide the API they require.

src/client/datascience/notebook/intellisense/completionProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+            getOnly: true
+        });
+        if (token.isCancellationRequested) {
+            return [];
+        }
+        if (!notebook) {
+            traceError(`Live Notebook not available for ${notebookDocument.uri.toString()}`);
+            return [];
+        }
+        const result = await notebook.getCompletion(document.getText(), document.offsetAt(position), token);
```
rchiodo, line 55: What happens if this takes too long? Does VS code just skip this one?

rchiodo (approved):

src/client/datascience/notebook/helpers/helpers.ts
```diff
@@ -814,3 +814,9 @@ export async function updateVSCNotebookAfterTrustingNotebook(
         });
     });
 }
+
+export function findAssociatedNotebookDocument(cellUri: Uri, vscodeNotebook: IVSCodeNotebook) {
+    return vscodeNotebook.notebookDocuments.find((item) =>
+        item.cells.some((cell) => cell.uri.toString() === cellUri.toString())
```
IanMatthewHuff, line 820: Should be the file path comparison?

IanMatthewHuff:
That was way less code than I expected.

IanMatthewHuff (approved):

rchiodo (changes requested):
Just waiting on the answer of what VS code does if a completion provider takes too long. I thought we had special code in the old notebooks to handle this.

src/client/datascience/notebook/intellisense/completionProvider.ts
DonJayamanne: Yes

src/client/datascience/notebook/helpers/helpers.ts
DonJayamanne: Doesn't matter here as these are VSCode generated Uris (not from us). hence didn't bother with that.
But will add so that we're consistent.

src/client/datascience/notebook/intellisense/completionProvider.ts
DonJayamanne: No wait, i need to check this.

DonJayamanne: Nope, it doesn't.

DonJayamanne: Fixed, @rchiodo added the timeout, please check.

src/client/datascience/notebook/integration.ts
```diff
@@ -97,6 +100,15 @@ export class NotebookIntegration implements IExtensionSingleActivationService {
             }
         }
     }
+
+    private registerCompletionItemProvider() {
+        const disposable = languages.registerCompletionItemProvider(
+            { language: PYTHON_LANGUAGE, scheme: NotebookCellScheme },
```
joyceerhl, line 106: In the Python extension we have two different document selectors for Python notebooks:
https://github.com/microsoft/vscode-jupyter/blob/e8af5b35ebba334e5e1a2e521895d3ffee0e4131/src/client/common/constants.ts#L11-L12

Do we need to use both here, or do they serve different purposes?

DonJayamanne: Removed. this was added in the early days when VSC was still working on the schemes for the cells.

.github/workflows/main.yml
```diff
@@ -406,6 +410,8 @@ jobs:
           python ./pythonFiles/install_debugpy.py
           conda env update --file ./build/conda-functional-requirements.yml
           python -m pip install --upgrade -r ./build/conda-functional-requirements.txt
+          python -m pip uninstall jedi --yes
+          python -m pip install jedi==0.17.2
```
DonJayamanne, line 414: See notes above.

Issue #1185, opened 2020-08-06
Stop generating svg for matplotlib in native notebooks


DonJayamanne:
Based on jupyter the SVG plots should have a higher priority over png and others when rendering output.
However as a side effect we are no longer able to open the plot viewer as a result of this:
Solutions:
* Change priority (works for our case with matplot lib, but not in other cases where user has SVG + PNG)
* Create a custom renderer for SVG and add the icon to this (if this is an SVG that was generated by our code)
* Detect when SVGs are generated as a result of our changes to matplot lib rendering and only add PNG into output
* etc...
* NOte, we have also looked at removing this approach and using the ipywidget for matplot lib as that results in a smaller ipynb file as there will be no SVG generated by us.

DonJayamanne:
Not required, instead ensure users can use matplot lib widget with ability to:
* zoom in/out of plots
* Save a plot (ipyml supports saving plots, this needs to be tested)

rchiodo:
How do we point users to using the widget though? It's isn't as simple as just adding ```%matplotlib widget``` for them. That ends up making all plots use the same widget (they have to change their code to use a widget).

DonJayamanne:
I'd argue if users want a plot viewer (zoom in out & the like), then they are better of using the widget as that far more superior than our plot viewer (which has perf issues, blowing notebooks to a few GB).
I'm arguing for pointing users towards better (existing) tools..

> How do we point users to using the widget though? 

Yes, agree. I did ping @claudiaregio . 
My suggest was to probably add a prompt (code lens or the like) prompting users to install the widget like we do with tenshorboards...
Else users will never know about this widget and asking them on GitHub would not be the best way to educate the users

claudiaregio:
After reviewing we are going to move forward to stop generating svg for matplotlib. 
Instead we are going to recommend users use the matplotlib widget that gives them better interactive capabilities. 

We are going to generate the same icon for these plots, and when users familiar with the plot viewer click on it, it will bring up a popup that recommends the use of this widget instead.

rchiodo:
@claudiaregio What's our recommendation going to be? It isn't a drop in replacement. The user has to change their code to use it.

claudiaregio:
We are going to generate the same icon that users familiar with plot viewer would expect to see. When they click it it will send them to a wiki page on our github with instructions on how to get started using the matplotlib widget instead

kieferrm:
The upstream issue is closed.

[...]

PR #4436, opened 2021-01-20
Stop generating svg for matplotlib in native notebooks

DonJayamanne:
For #1185
The experiment is to get users to try to use Maplotlib Widget (which supports more features & doesn't require us to build something).
Eventually we might consider shipping or recommending this widget like we do with pytorch.


rchiodo:
This is just disabling SVG support. Is there a separate issue to 'recommend' matplotlib widget? Not sure the matplotlib widget is a replacement for the plot viewer either.

rchiodo (approved):

IanMatthewHuff (approved):

DonJayamanne:
>  Is there a separate issue to 'recommend' matplotlib widget? Not sure the matplotlib widget is a replacement for the plot viewer either.

The feedback from PMs was we'll address this based on user feedback. the matplot lib viewer supports zoom in/out, saving without having to maintain our own library & any perf issues of generating & saving SVGs in notebooks.

joyceerhl (approved):

Issue #4409, opened 2021-01-15
Clicking on run button (native notebooks) when no kernel running looks like nothing happened.


rchiodo:
Every time I do this I think it didn't work. Can we make the run button switch to busy (but somehow not include the execution time). I believe we do this so the execution time is accurate.

rchiodo:
Validated

PR #4448, opened 2021-01-21
Add cell indicator for pre-execution

DonJayamanne:
For #4409
The solution is to wait for notebook to start after a cell execution has been queued (this happens by creating an instance of CellExecution).
Previously it was the other way around.

### DIFF
--- a/src/client/datascience/jupyter/kernels/kernel.ts
+++ b/src/client/datascience/jupyter/kernels/kernel.ts
@@ -67,7 +67,7 @@ export class Kernel implements IKernel {
     private readonly _onRestarted = new EventEmitter<void>();
     private readonly _onDisposed = new EventEmitter<void>();
     private readonly _onInterruptTimedOut = new EventEmitter<void>();
-    private _notebookPromise?: Promise<INotebook | undefined>;
+    private _notebookPromise?: Promise<INotebook>;
     private readonly hookedNotebookForEvents = new WeakSet<INotebook>();
     private restarting?: Deferred<void>;
     private readonly kernelValidated = new Map<string, { kernel: IKernel; promise: Promise<void> }>();
@@ -105,12 +105,12 @@ export class Kernel implements IKernel {
         );
     }
     public async executeCell(cell: NotebookCell): Promise<void> {
-        await this.start({ disableUI: false });
-        await this.kernelExecution.executeCell(cell);
+        const notebookPromise = this.startInternal({ disableUI: false });
+        await this.kernelExecution.executeCell(notebookPromise, cell);
     }
     public async executeAllCells(document: NotebookDocument): Promise<void> {
-        await this.start({ disableUI: false });
-        await this.kernelExecution.executeAllCells(document);
+        const notebookPromise = this.startInternal({ disableUI: false });
+        await this.kernelExecution.executeAllCells(notebookPromise, document);
     }
     public async cancelCell(cell: NotebookCell) {
         this.startCancellation.cancel();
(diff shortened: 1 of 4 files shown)


src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -168,12 +166,12 @@ export class KernelExecution implements IDisposable {
...
     }
 
-    public cancelAllCells(document: NotebookDocument): void {
+    public async cancelAllCells(document: NotebookDocument): Promise<void> {
         if (this.documentExecutions.get(document)) {
             this.documentExecutions.get(document)!.cancel();
         }
         traceInfo('Cancel document execution');
-        document.cells.forEach((cell) => this.cancelCell(cell));
+        await Promise.all(document.cells.map((cell) => this.cancelCell(cell)));
```
DonJayamanne, line 174: Floating promises fixed, we are waiting when cancelling a single cell, now when cancelling all cell, we were not waiting.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -105,54 +106,24 @@ export class Kernel implements IKernel {
...
         this.startCancellation.cancel();
         await this.kernelExecution.cancelCell(cell);
     }
     public async cancelAllCells(document: NotebookDocument) {
         this.startCancellation.cancel();
-        this.kernelExecution.cancelAllCells(document);
+        await this.kernelExecution.cancelAllCells(document);
     }
     public async start(options?: { disableUI?: boolean }): Promise<void> {
-        if (this.restarting) {
```
DonJayamanne, line 124: Moved to a separate method.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -105,54 +106,24 @@ export class Kernel implements IKernel {
...
-                resource: this.uri,
-                disableUI: options?.disableUI,
-                getOnly: false,
-                metadata: undefined, // No need to pass this, as we have a kernel connection (metadata is required in lower layers to determine the kernel connection).
-                kernelConnection: this.kernelConnectionMetadata,
-                token: this.startCancellation.token
-            });
-
-            this._notebookPromise
-                .then((nb) => (this.kernelExecution.notebook = this.notebook = nb))
```
DonJayamanne, line 144: Removed the hack of intiailizing `kernelExecution.notebook`.
Now we pass this aa an argument to `executeCell` which is a better way than initialing a property & then calling a method (& then in the method checking if the property was initialized - yuck)

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -209,6 +179,48 @@ export class Kernel implements IKernel {
...
+                            identity: this.uri,
+                            resource: this.uri,
+                            disableUI: options?.disableUI,
+                            getOnly: false,
+                            metadata: undefined, // No need to pass this, as we have a kernel connection (metadata is required in lower layers to determine the kernel connection).
+                            kernelConnection: this.kernelConnectionMetadata,
+                            token: this.startCancellation.token
+                        });
+                        if (!this.notebook) {
+                            // This is an unlikely case.
```
DonJayamanne, line 202: Unlikely, but necessary due to the fact that `getOrCreateNotebook` returns a type of `undefined` (not possible, too messy to fix this typing issue)

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -30,8 +30,6 @@ const vscodeNotebookEnums = require('vscode') as typeof import('vscode-proposed'
  * Else the `Kernel` class gets very big.
  */
 export class KernelExecution implements IDisposable {
-    public notebook?: INotebook;
```
DonJayamanne, line 33: Removed this hacky property.

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -57,36 +55,35 @@ export class KernelExecution implements IDisposable {
...
+        if (!editor) {
+            return;
+        }
         const cellExecution = this.executionFactory.create(cell, isPythonKernelConnection(this.metadata));
         this.cellExecutions.set(cell, cellExecution);
 
-        const kernel = this.getKernel(cell.notebook);
-
         try {
+            const notebook = await notebookPromise;
```
DonJayamanne, line 73: At this point, the status of the cell has been updated & we wait for the notebook.
In the `finally` we call `cellExecution.cancel` which reverts the cell state.

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -220,12 +218,9 @@ export class KernelExecution implements IDisposable {
 
     private async executeIndividualCell(
         kernelPromise: Promise<IKernel>,
-        cellExecution: CellExecution
+        cellExecution: CellExecution,
+        notebook: INotebook
     ): Promise<NotebookCellRunState | undefined> {
-        if (!this.notebook) {
```
DonJayamanne, line 225: We no longer need this hack, we now pass `notebook` as an argument. Cleaner, else calling code needs to ensure property is set before invoking this class methods (not inutitive & poor design)

src/client/datascience/jupyter/kernels/kernelExecution.ts
DonJayamanne: `cellExecution` sets the state when its created & udpates again when it is disposed/cancelled/completed.

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -57,36 +55,35 @@ export class KernelExecution implements IDisposable {
...
             // clear cell output
             cell.outputs = [];
             return;
         }
+        const editor = this.vscNotebook.notebookEditors.find((item) => item.document === cell.notebook);
+        if (!editor) {
+            return;
+        }
         const cellExecution = this.executionFactory.create(cell, isPythonKernelConnection(this.metadata));
         this.cellExecutions.set(cell, cellExecution);
```
rchiodo, line 70: So this line here needs to run in order to indicate the cell is executing. We used to wait for the notebook promise before we did this, but now it gets passed the notebook promise instead?

rchiodo (approved):

src/client/datascience/jupyter/kernels/kernelExecution.ts
DonJayamanne: Yes, the line is `executionFactory.create`. We need to create the `CellExecution`.
I.e. we   need to start queuing the cell execution (which happens when we create that class).

Issue #4356, opened 2021-01-12
UI Cleanup for the VariableView


IanMatthewHuff:
UI cleanup needed to get the VariableView working well for release

1 - Remove the 'x' that lets you close the view, I believe this is a leftover from being hosted in the IW / old webview
2 - Size correctly to fill the webview area, right now it's not filling and it's still looking for height updates from the splitter

Optional
1 - Add the name of the active document on top
2 - Add a button into the editor window to open up the view

IanMatthewHuff:
Adding, color syntax highlighting doesn't look correct in the view right now.

claudiaregio:
For 1/27: 
P0: Remove 'X', Icon in notebook toolbar 
P1: name of active document on top, color syntax highlighting

IanMatthewHuff:
@claudiaregio Sounds good. X removal and sizing went in on Friday. Will add icon next.

IanMatthewHuff:
@claudiaregio Open view button just went in. At this point I might resolve this work item and create new items for syntax color and adding the name of the active document, just to make tracking easier.

IanMatthewHuff:
For Validation. This bug is removing the x close button + sizing to fill. Will open new bugs for the other issues.

rchiodo:
Validated

PR #4449, opened 2021-01-21
Variable View UI Cleanup part 1

IanMatthewHuff:
For #4356 


IanMatthewHuff:
A bit of context here. This change does two things. When hosted in a view (instead of a document) the variable explorer now no longer has a close button and sizes it self appropriate to the given view area. 

Change is a bit funkier then might be expected due to how the variable view works. The grid that we used needed a hard code vertical size. So we had lots of old code (Tim worked on this) to handle resizing vertically with a dragger. It's all a bit non-clean / non reacty. Given that we're probably going to deprecate this view the goal was to fit this into the old context as cleanly as possible with the least disturbance to the existing variable code paths. Basics is that the variable view panel watches for resizes, then sets a viewHeight value in the state that is only used then in view mode, which is when the variable explorer is hosted in a view as opposed to a document. 

Old Code (close button, doesn't fill space, has a draggable divider at the bottom):
image

New Code:
image

src/datascience-ui/interactive-common/variableExplorer.tsx
```diff
@@ -178,15 +180,57 @@ export class VariableExplorer extends React.Component<IVariableExplorerProps, IV
...
+        ){
+            return true;
+        }
+
         return false;
     }
 
     public render() {
+        // This control renders differently when hosted in a versus a document
+        if (this.props.viewMode) {
```
IanMatthewHuff, line 201: This is where we split the rendering. There were a number of differences, so I just used two different functions for the view mode versus document mode.

src/datascience-ui/interactive-common/variableExplorer.tsx
```diff
@@ -234,6 +278,13 @@ export class VariableExplorer extends React.Component<IVariableExplorerProps, IV
     }
 
     private renderGrid() {
+        let newGridHeight: number | undefined;
+
+        // In in view mode, just use the viewHeight prop for calculating size
+        if(this.props.viewMode) {
```
IanMatthewHuff, line 284: Previously variableExplorer kept an internal state outside of redux for setting vertical height. For the new view mode we don't need this, so we can just look at the viewHeight prop for sizing.

src/datascience-ui/variable-view/variableViewPanel.tsx
```diff
@@ -50,12 +59,27 @@ export class VariableViewPanel extends React.Component<IVariableViewPanelProps>
...
         return (
-            <div id="variable-view-main-panel" role="Main" style={dynamicFont}>
+            <div id="variable-view-main-panel" role="Main" ref={this.panelRef} style={dynamicFont}>
                 {this.renderVariablePanel(this.props.baseTheme)}
             </div>
-        ); // NOTE: Currently the OPEN, button just exists to mimic the toggling of the variable view, make it easier to test when working
+        );
     }
 
+    private windowResized = () => {
```
IanMatthewHuff, line 68: Same resize / timer code that we used in other spots reactSlickGrid / monacoEditor.

src/datascience-ui/history-react/interactivePanel.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
IanMatthewHuff, line 1: this seems to spuriously trigger all over in .tsx files. Should I look into it more? I just disabled in .tsx files that I used. Might be able to turn it off for all of them, but I don't know if that would hide legit errors.

rchiodo: unused vars should start with _. They do in the .ts files. So I think it's legit?

src/datascience-ui/interactive-common/variableExplorer.tsx
rchiodo: I'm not sure which one is which? Is view mode inside of a webview?

rchiodo (approved):

src/datascience-ui/interactive-common/variableExplorer.tsx
IanMatthewHuff: @rchiodo Yeah, in retrospect the name isn't the best, view is too general. I was thinking view versus document...maybe standaloneMode? That's really the crux of the issue here as the variable view is hosted all by itself outside of the document context.

IanMatthewHuff: variableWebviewViewMode seems descriptive, but too wordy.

rchiodo: standaloneMode gets my vote. That's clear to me at least.

DavidKutu (approved):

Issue #4095, opened 2020-12-02
Widget Icons for play controls are all broken in native notebooks


IanMatthewHuff:
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/cf9d9dbc-22ee-4efd-85ae-fcae579d4729)

They all show up as boxes. Same controls are used for stuff like animation in py3js.

rchiodo:
This is font awesome. It needs to be included into the list of stuff we load into the output webview.

DonJayamanne:
Preload the font-awesome stuff.

joyceerhl:
Validated. 
![image](https://user-images.githubusercontent.com/30305945/106095161-22f39500-60e8-11eb-9ad7-aff39a46ae68.png)

PR #4467, opened 2021-01-23
Preload font awesome for ipywidgets

DavidKutu:
For #4095

It fixes the issue:
image

BUT, it requires an internet connection. That shouldn't be a big deal, since you need it for ipywidgets anyway, and we could take this approach to font awesome to remove it entirely from the repo and just get it by http.


rchiodo (changes requested):
I think we should remove it from the package.json and from our other spots where we load it and replace it with this.

joyceerhl (approved):
This approach works. I'm wondering whether we could download fontawesome at build time instead?

rchiodo:
> This approach works. I'm wondering whether we could download fontawesome at build time instead?

Unfortunately that's the same as loading it from our node_modules which was turning out to be a pain in the butt. This is much simpler and then we can just remove our shipping of it altogether.

build/webpack/font-awesome.js
```diff
@@ -0,0 +1,12 @@
+addCSS('https://use.fontawesome.com/releases/v5.8.1/css/all.css');
```
DonJayamanne, line 1: Doesn't this mean this will never work for users without internet connection?
Not sure this is what we want. Is this what we do in webviews?

DonJayamanne (changes requested):

build/webpack/font-awesome.js
DonJayamanne: > This is much simpler and then we can just remove our shipping of it altogether.

Shouldn't we identify the problem & fix it.
E.g. i WSL users don't have network connection, hence this will not work even for basic widgets thats built in.
I would prefer we identify the cause and fix it.

DonJayamanne: Also we're shipping 	4.7.0 and using 5.8.1 here, not sure this is correct. Also would be incorrect in the licensing docs..

rchiodo: Old approach fails because we can't get a webviewURI at the time of the preload.

rchiodo: preloads are on the kernel, not the webview? Maybe there's a way to do a preload on the webview?

rchiodo: And I'm not sure about WSL. We can test it out. I can do a curl of say http://www.google.com from a WSL bash script.

src/client/datascience/notebook/types.ts
```diff
@@ -12,7 +12,9 @@ import {
 export const INotebookContentProvider = Symbol('INotebookContentProvider');
 
 export const INotebookKernelProvider = Symbol('INotebookKernelProvider');
-export interface INotebookKernelProvider extends NotebookKernelProvider {}
+export interface INotebookKernelProvider extends NotebookKernelProvider {
+    onDidGetFontAwesomeMessage: vscode.Event<{ webview: NotebookCommunication; message: any }>;
```
rchiodo, line 16: The name here seems a bit odd. Maybe 'onDidGetMessage'? The notebook kernel provider doesn't need to know which message, right?

src/client/datascience/notebook/kernelProvider.ts
```diff
@@ -94,6 +98,7 @@ export class VSCodeKernelPickerProvider implements INotebookKernelProvider {
         webview: NotebookCommunication,
         token: CancellationToken
     ): Promise<void> {
+        webview.onDidReceiveMessage((m) => this._onDidGetFontAwesomeMessage.fire({ webview, message: m }));
```
rchiodo, line 101: I think we need to dispose this (or use the 3rd parameter here that is a list of disposables)

rchiodo:
```suggestion
webview.onDidReceiveMessage((m) => this._onDidReceiveMessage.fire({ webview, message: m }), undefined, this.disposables);
```

src/client/datascience/notebook/kernelWithMetadata.ts
```diff
@@ -37,8 +42,31 @@ export class VSCodeNotebookKernelMetadata implements VSCNotebookKernel {
         private readonly kernelProvider: IKernelProvider,
         private readonly notebook: IVSCodeNotebook,
         private readonly context: IExtensionContext,
-        private readonly preferredRemoteKernelIdProvider: PreferredRemoteKernelIdProvider
-    ) {}
+        private readonly preferredRemoteKernelIdProvider: PreferredRemoteKernelIdProvider,
+        private readonly notebookKernelProvider: INotebookKernelProvider
+    ) {
+        this.notebookKernelProvider.onDidGetFontAwesomeMessage((e) => {
```
DonJayamanne, line 48: I'd prefer we move this code into the `kernelProvider.ts` file.
This way we do'nt need to add `onDidGetFontAwesomeMessage` (fewer API changes).

src/client/datascience/notebook/kernelProvider.ts
DonJayamanne: I would prefer we moved the code from `kernelWithMetadata` into this method here directly. This way we do'nt need to add a new event & an event handler, less changes as well & wee know exactly what's going on.

src/client/datascience/notebook/fontAwesomeLoader.js
```diff
@@ -0,0 +1,31 @@
+var vscodeApi = acquireApi();
```
DonJayamanne, line 1: `const` is better.

src/client/datascience/notebook/fontAwesomeLoader.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+function baseHandler(e) {
+    if (e.data.type === 'GetFontAwesomeUriResponse' && e.data.payload) {
+        addCSS(e.data.payload.path);
+    }
+}
+
+function addCSS(filename) {
+    var head = document.head;
```
DonJayamanne, line 23: Would prefer to use `const`.

DonJayamanne (changes requested):
Suggest moving the code to to `kernelProvider`, I think its simpler & minimizes the changes as well

src/client/datascience/notebook/kernelProvider.ts
rchiodo: I suggested we added an event because the handler for the message is the thing that knows about preloads. Otherwise the idea of fontAwesome propagates out to another class.

rchiodo (approved):

src/client/datascience/notebook/fontAwesomeLoader.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+const getPublicPath = () => {
+    return new URL(scriptUrl.replace(/[^/]+$/, '')).toString();
+};
+const uri = getPublicPath() + '../notebook/node_modules/font-awesome/css/font-awesome.min.css';
```
IanMatthewHuff, line 6: Anytime I see paths, just make sure that it works on Windows + Mac / Linux.

IanMatthewHuff (approved):

src/client/datascience/notebook/fontAwesomeLoader.js
DonJayamanne: Curious why do we need `../notebook/node_modules`, isn't `node_modules/font...` sufficient?

src/client/datascience/notebook/kernelWithMetadata.ts
```diff
@@ -22,6 +22,9 @@ export class VSCodeNotebookKernelMetadata implements VSCNotebookKernel {
             Uri.file(join(this.context.extensionPath, 'out', 'ipywidgets', 'dist', 'ipywidgets.js')),
             Uri.file(
                 join(this.context.extensionPath, 'out', 'datascience-ui', 'ipywidgetsKernel', 'ipywidgetsKernel.js')
+            ),
+            Uri.file(
+                join(this.context.extensionPath, 'out', 'datascience-ui', 'ipywidgetsKernel', 'fontAwesomeLoader.js')
```
DonJayamanne, line 27: Not sure why `fontAwesomeLoader.js`1 needs to be in the `ipywidgetKernel` folder.
This sends the wrong message, i'd move it else where like `notebook`

DonJayamanne (changes requested):
Please can we move the js into `notebook` folder out from `ipywidgetsKernel` else it sends the message that this is realted to ipywidgets (if its in that folder).

src/client/datascience/notebook/fontAwesomeLoader.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+const getPublicPath = () => {
+    return new URL(scriptUrl.replace(/[^/]+$/, '')).toString();
+};
+const uri = getPublicPath() + 'node_modules/font-awesome/css/font-awesome.min.css';
```
joyceerhl, line 6: Does this need to be
```suggestion
const path = require('path');
const uri = getPublicPath() + path.join('node_modules', 'font-awesome', 'css', 'font-awesome.min.css');
```

to ensure the correct path separator is used?

DavidKutu: normally yes, but in that context the path module can't be found.

DonJayamanne: I don't think we can use `path.join` this is browser code, also `/` (forward slases) works across all OS (yes even windows). No idea why windows supports `\`

DonJayamanne: Super consistent, even with browser urls, its `/`.

DonJayamanne (approved):

Issue #4486, opened 2021-01-25
Add a native notebook toolbar button to show the variable view


IanMatthewHuff:
Per convo with Claudia. When the view is visible already, keep the icon, but make it disabled.

rchiodo:
Validated

PR #4487, opened 2021-01-25
Variable View Editor Button

IanMatthewHuff:
For #4486 


IanMatthewHuff:
Per Claudia convo button still there, but disabled when view is currently visible.

src/client/common/application/types.ts
```diff
@@ -1093,7 +1093,10 @@ export interface IWebview {
 
 // Wraps the VS Code webview view
 export const IWebviewView = Symbol('IWebviewView');
-export interface IWebviewView extends IWebview {}
+export interface IWebviewView extends IWebview {
+    readonly onDidChangeVisiblity: Event<void>;
```
IanMatthewHuff, line 1097: vscode webviewview doesn't handle visibility the same as vscode webviewpanel, so this is handled specific to the IWebviewView interface.

src/client/common/application/webviewViews/webviewView.ts
```diff
@@ -38,5 +49,11 @@ export class WebviewView extends Webview implements IWebviewView {
                 this.panelOptions.listener.onMessage(message.type, message.payload);
             })
         );
+
+        this.disposableRegistry.push(
+            webviewHost.onDidChangeVisibility(() => {
```
rchiodo, line 54: Does this fire when the host is launched? Seems like it should if it doesn't.

rchiodo (approved):

src/client/common/application/webviewViews/webviewView.ts
IanMatthewHuff: @rchiodo Yeah, I think you are right. I don't believe that I noticed as my variable view does an initial direct check for visible state before it hooks up and watches for the event. But we should also fire an initial fire of this.

PR #4491, opened 2021-01-25
Use VSC Notebook API in stable

DonJayamanne:
For #4488


src/client/extensionActivation.ts
```diff
@@ -83,25 +83,12 @@ async function activateLegacy(
...
-    setLoggingLevel(configuration.getSettings().logging.level);
-
     // Load the two data science experiments that we need to register types
     // Await here to keep the register method sync
     const experimentService = serviceContainer.get<IExperimentService>(IExperimentService);
     experimentService.logExperiments();
-    let useVSCodeNotebookAPI = await experimentService.inExperiment(Experiments.NativeNotebook);
+    let useVSCodeNotebookAPI =
+        env.appName.includes('Insider') || (await experimentService.inExperiment(Experiments.NativeNotebook));
     let inCustomEditorApiExperiment = await experimentService.inExperiment(Experiments.CustomEditor);
```
rchiodo, line 92: I think inCustomEditorApiExperiment should be false if nativeEditor is true.

rchiodo: Maybe in the experiment service level? So nothing can have both of them turned on at the same time.

DavidKutu (changes requested):
We should make sure a user can't be in both experiments at once.
Or if they're in both, useVSCodeNotebookAPI  should stay true and inCustomEditorApiExperiment  should go to false.

src/client/extensionActivation.ts
DonJayamanne: This is done further below in that code.

src/client/extensionActivation.ts
```diff
@@ -112,6 +99,20 @@ async function activateLegacy(
         inCustomEditorApiExperiment = false;
```
DonJayamanne, line 99: Please chekc the line above this @rchiodo , if both are enabled, then native notebook is false with a message.

DonJayamanne: Please let me know if this addresses your concern, Check line `111`.  @rchiodo

DonJayamanne (commented):
Accidental comments

src/client/extensionActivation.ts
rchiodo: Sorry but that seems incorrect too. It's possible to be in both now without any work on the user. I don't think they'd know what to do.

joyceerhl: > I don't think they'd know what to do.
Can we give users an option on the error message to 'Use Custom Editor' or 'Use Native Notebooks', and then opt them into just the one that they select?

Alternatively we could give precedence to whatever the user has in their settings.json, i.e. if the user has opted into custom editor in their settings.json, and the experiment service has opted them into native notebooks, we should opt them out of native notebooks to prevent the conflict since they've explicitly asked to be in custom editor. i.e. manually configured settings should have priority over our experiment.

DonJayamanne: Sure, but if the user is opting into something manually then they should also know how to opt out. Not sure its worth the effort, will check with PMs.

DonJayamanne: Upon further review, I do'nt agree wit the sugestion:
>  'Use Custom Editor' or 'Use Native Notebooks', and then opt them into just the one that they select?

The user already added them selves manually, hence they can opt out.
This should be a rare case as we haven't actively asked users to opt into the experiment as we had a few issues in custom editor that made is unusable.

> we should opt them out of native notebooks to prevent the conflict since they've explicitly asked to be in custom editor. i.e. manually configured settings should have priority over our experiment.

Will bring on teams, i think we will have to do this.

DonJayamanne: Will add new custom editor, see discussion here https://teams.microsoft.com/l/message/19:9b163ad1e1ed4c49938d9692be1ce751@thread.skype/1611862176803?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=6182cd1a-acf2-40e6-8b08-07bd25990df8&parentMessageId=1611784871932&teamName=Python%20for%20VS%20Code&channelName=Data%20Science&createdTime=1611862176803

DonJayamanne: @rchiodo @DavidKutu pleasee re-review

src/client/extensionActivation.ts
```diff
@@ -83,25 +83,12 @@ async function activateLegacy(
     const fs = serviceManager.get<IFileSystem>(IFileSystem);
     await setExtensionInstallTelemetryProperties(fs);
 
-    const applicationEnv = serviceManager.get<IApplicationEnvironment>(IApplicationEnvironment);
```
rchiodo, line 86: Why did this stuff go away? Seems unrelated?

rchiodo: Nevermind. It just moved.

src/client/extensionActivation.ts
rchiodo: You didn't change this? Not sure what the difference is. It's still possible to be in both experiments at the same time. Or are you saying it's not possible unless the user explicitly does it? Because we're changing the custom editor experiment in control tower?

However when we change the 'control tower' value for customEditor, it should use the same optIn/optOutFrom value in our package.json, right? So that users that already using don't have to change their setting? 

Regardless of that, this code still seems wrong. 

User gets put in native experiment. 
They're already opting into custom editor 

This error comes up. 

They don't know why. What are they supposed to do?

rchiodo (changes requested):
⏲️

src/client/extensionActivation.ts
DonJayamanne: > Or are you saying it's not possible unless the user explicitly does it?

Yes this was my assumption. The experiments should be mutually exclusive else the whole experiment framework seems busted. 
@claudiaregio Please can you confirm that users in CustomEditor will not end up in NativeNotebooks as well

DonJayamanne: > They don't know why. What are they supposed to do?

This is pre-existing code and we were always displaying errors to the user, I've only changed this PR to ensure we can use NativeNotebook in stable.

rchiodo: Even if control tower makes it impossible, it's still possible for a user to be in one manually and then end up in the other one by 'accident'.

DonJayamanne: > However when we change the 'control tower' value for customEditor, it should use the same optIn/optOutFrom value in our package.json, right? So that users that already using don't have to change their setting?

@greazer suggested we create a new experiment for Custom Editor.
As this PR focuses on just enabling Native Notebooks in stable, I"m not going to fix custom editror & the like here. Theres a separate issue to address custom editor (experiment) issues  https://github.com/microsoft/vscode-jupyter/issues/4561.

DonJayamanne: Here's the existing code that prevents users from having both experiments turned on:
```typescript
    if (useVSCodeNotebookAPI && inCustomEditorApiExperiment) {
        const appShell = serviceContainer.get<IApplicationShell>(IApplicationShell);
        appShell.showErrorMessage(localize.DataScience.illegalEditorConfig()).then(noop, noop);
        useVSCodeNotebookAPI = false;
        inCustomEditorApiExperiment = false;
```

DavidKutu (approved):

src/client/extensionActivation.ts
rchiodo: Okay so that error message may come up but you'll fix it elsewhere? That's good for me.

rchiodo (approved):

PR #4500, opened 2021-01-25
Readd back the add cells below test

rchiodo:
It looked to me like it was timing out when closing windows. I changed it to not keep trying over and over if it failed and the test is passing.

Here's the passing run with these changes:
https://github.com/microsoft/vscode-jupyter/runs/1764948285?check_suite_focus=true


src/test/initialize.ts
```diff
@@ -124,9 +124,7 @@ async function closeWindowsInternal() {
         await closeWindowsImplementation();
     } catch (ex) {
         if (ex instanceof CloseEditorsTimeoutError) {
-            // Try again with a smaller timeout (no idea why VSCode is timeout out here).
-            await sleep(500); // Possible VSC is busy & wasn't able to handle previous command.
-            await closeWindowsImplementation(5_000);
+            // Do nothing. Just stop waiting.
```
IanMatthewHuff, line 127: This seems ok as far as I know, but I'm assuming that this retry was here for a reason. @DonJayamanne what was this needed for?

IanMatthewHuff (approved):

src/test/initialize.ts
DonJayamanne: Yeah sometimes, VSC would timeout waiting to close editors. I have seen similar code in the VSCode repo in some test.
Hence the retry, if this works, i don't see any issues with removing it.
In the past this would time out very frequently and windows would not close

DonJayamanne (approved):

src/test/initialize.ts
```diff
@@ -99,7 +99,7 @@ async function closeWindowsInternal() {
             super("Command 'workbench.action.closeAllEditors' timed out");
         }
     }
-    const closeWindowsImplementation = (timeout = 15_000) => {
+    const closeWindowsImplementation = (timeout = 2_000) => {
```
DonJayamanne, line 102: Didn't you already make this change in another PR @rchiodo 
I left this issue as I saw a related change in another PR.

rchiodo: I thought I did. Maybe I didn't submit it.

Issue #4423, opened 2021-01-19
Provide a mechanism for an extension to determine or set a preferred kernel.


brettfo:
As a corollary to #4363, there should be a way for a VS Code concept of a kernel to map to an existing `.ipynb` file without requiring an installed kernelspec file.  E.g., consider a situation where a user installs this extension as well as the .NET Interactive extension.  From there they open an existing `.ipynb` file.  At that point, there is no mechanism for the VS Code version of the .NET Interactive kernel to be auto-selected.

This issue is hopefully more of a discussion about how to add this functionality, and the end result may require changes from VS Code itself.

As a single data point, if a new `.ipynb` notebook is created, the kernel set to .NET Interactive and it's saved, the notebook's metadata looks like this:

``` json
...
"metadata": {
    "language_info": {
        "name": "csharp",
        "nbconvert_exporter": "python"
    },
    "orig_nbformat": 2
}
...
```

DonJayamanne:
Thanks @brettfo 
We have discussed this internally & our plan was to not pre-select the non-python kernels unless we had an exact match for the kernel.
For instance in this example, we might not have a matching kernel & we shoudlnt' ideally pre-select any kernel.
Then the .net extension can nominate its kernel as a preferred kernel using the `NotebookKernel.isPreferred` property.

DonJayamanne:
At our end the fix woudl be in `getKernelForLocalRawConnection`.
```typescript
	const kernelSpec = await this.kernelFinder.findKernelSpec(resource, notebookMetadata, cancelToken);
....
....
        if (!kernelSpec && activeInterpreter) {
            await this.installDependenciesIntoInterpreter(activeInterpreter, ignoreDependencyCheck, cancelToken);

            // Return current interpreter.
            return {
                kind: 'startUsingPythonInterpreter',
                interpreter: activeInterpreter
            };
        } else if (kernelSpec) {
```

**To discuss**
* Should we ever preselect a kernel & just add an exception for .NET interactive?
	* I think yes, majority of our (Python/Jupyter) users will have our extension.
	* We can add exceptions for .NET kernels & look for a more generic solution later (if other extensions do not like this pre-selection of kernels by jupyter extension).
	* If the preselected kernel is a .NET kernel & .NET interactive is installed, then never preselect a .NET kernelspec

**Note:**
@brettfo VS Code has added the ability for users to define the default kernel provider.
[...]

PR #4504, opened 2021-01-26
Do not store kernel info in nb metadata for other VSC extension kernels

DonJayamanne:
For https://github.com/microsoft/vscode-jupyter/issues/4423#issuecomment-763903125

This way we don't set the kernel metadata when a non-jupyter kernel is selected.
Allowing others to update the notebook metadata without us blowing it away.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -991,7 +991,6 @@ module.exports = {
         'src/client/datascience/interactive-common/intellisense/conversion.ts',
         'src/client/datascience/activation.ts',
         'src/client/datascience/jupyterUriProviderWrapper.ts',
-        'src/client/datascience/errorHandler/errorHandler.ts',
         'src/client/datascience/cellMatcher.ts',
         'src/client/datascience/notebookStorage/notebookModel.ts',
         'src/client/datascience/notebookStorage/notebookModelEditEvent.ts',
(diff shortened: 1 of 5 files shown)


rchiodo:
I'm confused. This looks like it's clearing the metadata? Don't we want to just not set it?

DonJayamanne:
> I'm confused. This looks like it's clearing the metadata? Don't we want to just not set it?
If  you select a Jupyter kernel then we set it.
If you then select a non-jupyter kenrel we need to unset it, i.e. clear it.
This way, if users change kernels to a non-jupyter kernel (not ours), & save, then there's no custom metadata that will be set, **apart from what ever is already in there**.

If you check the method, we use call the same function to save the kernel information (but only if its our kernel).

rchiodo:
> If you then select a non-jupyter kenrel we need to unset it, i.e. clear it.
This way, if users change kernels to a non-jupyter kernel (not ours), & save, then there's no custom metadata that will be set, **apart from what ever is already in there**.

Why wouldn't this clear the metadata set by the .NET kernel provider? Sure it won't overwrite something that's already on disk, but won't it overwrite a value in memory?

IanMatthewHuff:
As much as I'm following this looks right to me, though I'm not 100% sure on the scenario that the C# folks are after. Is it ok that it's setting metadata to undefined as opposed to {}?

IanMatthewHuff (approved):

DavidKutu (approved):

rchiodo (changes requested):
Sorry still not clear on how tis works. Maybe we should chat?

rchiodo (approved):

Issue #4411, opened 2021-01-15
Auto start doesn't work for native notebooks


rchiodo:
Makes startup take all that much longer.

rchiodo:
This is actually not working correctly. It puts up errors on auto start. The old code would ignore errors.

rchiodo:
Meaning if you have an error starting a kernel, it shouldn't just show up randomly. It should only show if you click run a cell.

DonJayamanne:
> puts up errors on auto start. The old code would ignore errors.

Can you let me know what you did to get errors? Is it an invalid kernel or what?

rchiodo:
I have a kernel without ipykernel installed. Well just an interpreter. It's set a the default for a notebook. On opening the notebook an error pops up.

rchiodo:
This still isn't working? I believe I merged your latest changes. I still get the error to come up without clicking on anything.

DonJayamanne:
Unable to replicate, please could you re-test.

rchiodo:
I believe I fixed it with the subsequent submission. We should probably have somebody else validate.

IanMatthewHuff:
@DonJayamanne @rchiodo This didn't validate for me. When I start up VS Code with a native notebook I see the message that it's looking to start a server due to preload conditions. But I don't see an actual server start. When I execute the first cell, I actually see the message for the server starting hit:
[I 10:31:22.120 NotebookApp] The Jupyter Notebook is running at:
[I 10:31:22.121 NotebookApp] http://localhost:8888/?token=5578ca9da346e3451deb88eefcb727316bbef761f33788ce

rchiodo:
@IanMatthewHuff  does it work for raw?

DonJayamanne:
Validated and seems to work for me:
* Raw
    * Open a notebook and kernel will get selected by VS Code
    * A kernel is auto started
    * Test: Wait for a few seconds, then run a cell it should run immediately
* Jupyter (non-Raw)
    * Open a notebook and kernel will get selected by VS Code
    * Jupyter server is auto started auto started
    * Test: Wait for a few seconds, then run a cell it should not start jupyter, but Kernel will get started (this cannot be avoided, unless we start Jupyter server with the exact same default kernel).
    * We should not have to start the server, but use existing server

I managed to validate both while debugging, i.e. confirming at a lower level that it works as expected.

joyceerhl:
Validated

PR #4511, opened 2021-01-26
Next step in the dependency moving

rchiodo:
More of what we discussed about moving the dependency installation deeper. I also fixed disableUI to be applied in both cases.

The next step will be to move registering of kernels into the kernel launch code. Otherwise when using jupyter, the native notebook kernel provider will register interpreters as kernels just by opening a notebook.

Should also address #4411 comments I had. 


src/client/datascience/jupyter/jupyterSession.ts
```diff
@@ -196,11 +210,17 @@ export class JupyterSession extends BaseJupyterSession {
...
         kernelConnection: KernelConnectionMetadata | undefined,
-        cancelToken?: CancellationToken
+        cancelToken?: CancellationToken,
+        disableUI?: boolean
     ): Promise<ISessionWithSocket> {
         // Create our backing file for the notebook
         const backingFile = await this.createBackingFile();
 
+        // Make sure the kernel has ipykernel installed if on a local machine.
+        if (kernelConnection?.interpreter && this.connInfo.localLaunch) {
```
rchiodo, line 220: This spot here is where we'll install ipykernel if necessary just before starting a kernel.

I believe in the future this will also be where we'll register a kernel (or maybe not as I don't think it's actually necessary)

src/client/datascience/jupyter/jupyterSession.ts
```diff
@@ -246,4 +266,37 @@ export class JupyterSession extends BaseJupyterSession {
             this.outputChannel.appendLine(output);
         }
     }
+
+    private async installDependenciesIntoInterpreter(
+        interpreter: PythonEnvironment,
+        cancelToken?: CancellationToken,
+        disableUI?: boolean
+    ) {
+        // TODO: On next submission move this code into a common location.
```
rchiodo, line 275: This code is common between the old dependency checker, the kernel launcher, and the jupyter session. When I remove it from the kernel provider, I'll move it to a common location (probably the kernelDependencyService)

rchiodo: That will happen in the next submit.

IanMatthewHuff (approved):

PR #4513, opened 2021-01-26
Fixes cooridnation of cell executions/cancellation

DonJayamanne:
For #4514
**Summary of changes**
* Moved interrupt from CellExecution to KernelExecution (as we need a single interrupt for kernel, not for each cell)
* Cells cannot cancel them selves, we cancel a kernel & kernel stops execution of cells.
	* Previously we'd cancel a kernel via the cancelMethod or cancellation Token and then the cellExecution would propagate that message upto KernelExecution asking that to interrupt the kernel... 
	* **I.e was backwards, cell should not have to ask the Kernel to interrupt itself.**
* Ensure we order the executions (moved from kernel layer to KernelExecution, closer to where we queue executions)
	* Else sometimes cells that are queued don't look queued at all (until previous cells finish execution)
* Remove interruptEvent

**Fix a number of bugs:**
* Things running out of order
	* Run multiple cells one after the other
	* Run whole notebook & then run another cell
* Run multiple cells, and the timers seem to indicate things are running out of order or longer than necessary
* Run all cells, then stop & it doesn't stop properly (document seems to be running)
* Run cells with empty cells and cell status isn't updated
* Issues with cancellation/interrupts (not able to re-run after).

**Will require a lot of tests**
Writing test is still a WIP

### DIFF
--- a/src/client/common/utils/async.ts
+++ b/src/client/common/utils/async.ts
@@ -240,9 +240,6 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
 
 export class ChainedExecutions<T> {
     private pendingCellExecution?: Promise<T> | undefined;
-    public clear() {
-        this.pendingCellExecution = undefined;
-    }
     public async chainExecution(next: () => Promise<T>): Promise<T> {
         if (this.pendingCellExecution) {
             const previous = this.pendingCellExecution;
--- a/src/client/common/utils/async.ts
+++ b/src/client/common/utils/async.ts
@@ -242,7 +242,9 @@ export class ChainedExecutions<T> {
     private pendingCellExecution?: Promise<T> | undefined;
     public async chainExecution(next: () => Promise<T>): Promise<T> {
         if (this.pendingCellExecution) {
-            await this.pendingCellExecution;
+            const previous = this.pendingCellExecution;
+            this.pendingCellExecution = undefined;
+            await previous;
         }
         const promise = next();
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/src/client/common/utils/async.ts
+++ b/src/client/common/utils/async.ts
@@ -3,6 +3,8 @@
 
 'use strict';
 
+import { traceError } from '../logger';
+
 /**
  * Error type thrown when a timeout occurs
  */
@@ -236,4 +238,23 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
         result = await iterator.next();
     }
     return results;
+}
+
+export class ChainedExecutions<T> {
+    private pendingCellExecution?: Promise<T> | undefined;
+    constructor(private readonly messageForErrorTracing: string) {}
+    public async chainExecution(next: () => Promise<T>): Promise<T> {
+        if (this.pendingCellExecution) {
+            try {
+                await this.pendingCellExecution;
+            } catch (ex) {
+                traceError(this.messageForErrorTracing, ex);
+                throw ex;
+            }
+        }
+        const promise = next();
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        this.pendingCellExecution = promise;
+        return promise;
+    }
 }
--- a/src/client/common/utils/async.ts
+++ b/src/client/common/utils/async.ts
@@ -3,8 +3,6 @@
 
 'use strict';
 
-import { traceError } from '../logger';
-
 /**
  * Error type thrown when a timeout occurs
  */
@@ -242,15 +240,9 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
 
 export class ChainedExecutions<T> {
     private pendingCellExecution?: Promise<T> | undefined;
-    constructor(private readonly messageForErrorTracing: string) {}
     public async chainExecution(next: () => Promise<T>): Promise<T> {
         if (this.pendingCellExecution) {
-            try {
-                await this.pendingCellExecution;
-            } catch (ex) {
-                traceError(this.messageForErrorTracing, ex);
-                throw ex;
-            }
+            await this.pendingCellExecution;
         }
         const promise = next();
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/src/client/common/utils/async.ts
+++ b/src/client/common/utils/async.ts
@@ -240,21 +240,21 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
 
 export class ChainedExecutions<T> {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    private pendingCellExecution: Promise<any> = Promise.resolve();
+    private pendingExecution: Promise<any> = Promise.resolve();
     /**
      * Clears the chained promises.
      */
     public clear() {
-        this.pendingCellExecution = Promise.resolve();
+        this.pendingExecution = Promise.resolve();
     }
     public async chainExecution(next: () => Promise<T>): Promise<T> {
         const deferred = createDeferred<T>();
-        const chainedPromise = this.pendingCellExecution.then(() => {
+        const chainedPromise = this.pendingExecution.then(() => {
             const nextPromise = next();
             nextPromise.then((result) => deferred.resolve(result)).catch((ex) => deferred.reject(ex));
             return nextPromise;
         });
-        this.pendingCellExecution = chainedPromise;
+        this.pendingExecution = chainedPromise;
 
         return Promise.race([deferred.promise, chainedPromise]);
     }
--- a/src/client/common/utils/async.ts
+++ b/src/client/common/utils/async.ts
@@ -240,15 +240,29 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
 
 export class ChainedExecutions<T> {
     private pendingCellExecution?: Promise<T> | undefined;
+    public clear() {
+        this.pendingCellExecution = undefined;
+    }
     public async chainExecution(next: () => Promise<T>): Promise<T> {
         if (this.pendingCellExecution) {
             const previous = this.pendingCellExecution;
+            // Clear in case this promise fails.
             this.pendingCellExecution = undefined;
             await previous;
         }
         const promise = next();
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         this.pendingCellExecution = promise;
+        promise
+            .finally(() => {
+                // Clear in case this next promise fails.
+                if (this.pendingCellExecution === promise) {
+                    this.pendingCellExecution = undefined;
+                }
+            })
+            .catch(() => {
+                //
+            });
         return promise;
     }
 }
--- a/src/client/common/utils/async.ts
+++ b/src/client/common/utils/async.ts
@@ -239,27 +239,23 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
 }
 
 export class ChainedExecutions<T> {
-    private pendingCellExecution?: Promise<T> | undefined;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private pendingCellExecution: Promise<any> = Promise.resolve();
+    /**
+     * Clears the chained promises.
+     */
+    public clear() {
+        this.pendingCellExecution = Promise.resolve();
+    }
     public async chainExecution(next: () => Promise<T>): Promise<T> {
-        if (this.pendingCellExecution) {
-            const previous = this.pendingCellExecution;
-            // Clear in case this promise fails.
-            this.pendingCellExecution = undefined;
-            await previous;
-        }
-        const promise = next();
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        this.pendingCellExecution = promise;
-        promise
-            .finally(() => {
-                // Clear in case this next promise fails.
-                if (this.pendingCellExecution === promise) {
-                    this.pendingCellExecution = undefined;
-                }
-            })
-            .catch(() => {
-                //
-            });
-        return promise;
+        const deferred = createDeferred<T>();
+        const chainedPromise = this.pendingCellExecution.then(() => {
+            const nextPromise = next();
+            nextPromise.then((result) => deferred.resolve(result)).catch((ex) => deferred.reject(ex));
+            return nextPromise;
+        });
+        this.pendingCellExecution = chainedPromise;
+
+        return Promise.race([deferred.promise, chainedPromise]);
     }
 }
--- a/src/client/datascience/jupyter/kernels/cellExecution.ts
+++ b/src/client/datascience/jupyter/kernels/cellExecution.ts
@@ -155,23 +155,6 @@ export class CellExecution {
         );
     }
 
-    /**
-     * If a cell has not yet started, then de-queue the cell from execution.
-     * If it has started, then wait for it to complete.
-     */
-    public async interrupt(): Promise<void> {
-        if (this.started) {
-            // At this point the cell execution can only be stopped from kernel & we should not
-            // stop handling execution results & the like from the kernel.
-            // The result will resolve when execution completes or kernel is restarted.
-            await this.result;
-            return;
-        }
-        // If we haven't even started this cell, then cancel it.
-        traceCellMessage(this.cell, 'Cell cancelled as a result of interrupt');
-        await this.cancel();
-    }
-
     public async start(kernelPromise: Promise<IKernel>, notebook: INotebook) {
         if (this.cancelHandled) {
             traceCellMessage(this.cell, 'Not starting as it was cancelled');
@@ -210,9 +193,20 @@ export class CellExecution {
     }
     /**
      * Cancel execution.
-     * If execution has commenced, then interrupt (via cancellation token) else dequeue from execution.
+     * If execution has commenced, then wait for execution to complete or kernel to start.
      */
     public async cancel() {
+        await this.cancelInternal(false);
+    }
+    public async cancelInternal(forced = false) {
+        if (this.started && !forced) {
+            // At this point the cell execution can only be stopped from kernel & we should not
+            // stop handling execution results & the like from the kernel.
+            // The result will resolve when execution completes or kernel is restarted.
+            traceCellMessage(this.cell, 'Cell is already running, waiting for it to finish or kernel to start');
+            await this.result;
+            return;
+        }
         if (this.cancelHandled || this._completed) {
             return;
         }
@@ -240,7 +234,7 @@ export class CellExecution {
         kernel.onRestarted(
             () => {
                 traceCellMessage(this.cell, 'Kernel restart handled in CellExecution, cancelling cell');
-                this.cancel().catch(noop);
+                this.cancelInternal(true).catch(noop);
             },
             this,
             this.disposables
--- a/src/client/datascience/jupyter/kernels/cellExecution.ts
+++ b/src/client/datascience/jupyter/kernels/cellExecution.ts
@@ -195,6 +195,13 @@ export class CellExecution {
     public async cancel() {
         await this.cancelInternal(false);
     }
+    /**
+     * @param {boolean} [forced=false]
+     * If `true`, then dequeue this & do not wait for execution to complete (basically kill it).
+     * This is used when we restart the kernel (either as a result of kernel interrupt or user initiated).
+     * When restarted, the execution needs to stop as jupyter will not send more messages.
+     * Thus `forced=true` is more like a hard kill.
+     */
     private async cancelInternal(forced = false) {
         if (this.started && !forced) {
             // At this point the cell execution can only be stopped from kernel & we should not
--- a/src/client/datascience/jupyter/kernels/cellExecution.ts
+++ b/src/client/datascience/jupyter/kernels/cellExecution.ts
@@ -47,7 +47,6 @@ import {
     INotebookExecutionLogger
 } from '../../types';
 import { translateCellFromNative } from '../../utils';
-import { IKernel } from './types';
 // eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
 const vscodeNotebookEnums = require('vscode') as typeof import('vscode-proposed');
 
@@ -150,8 +149,7 @@ export class CellExecution {
             context
         );
     }
-
-    public async start(kernelPromise: Promise<IKernel>, notebook: INotebook) {
+    public async start(notebook: INotebook) {
         if (this.cancelHandled) {
             traceCellMessage(this.cell, 'Not starting as it was cancelled');
             return;
@@ -164,8 +162,15 @@ export class CellExecution {
         if (!this.canExecuteCell()) {
             return;
         }
-        await this.initPromise;
+        if (this.started) {
+            traceCellMessage(this.cell, 'Cell has already been started yet CellExecution.Start invoked again');
+            traceError(`Cell has already been started yet CellExecution.Start invoked again ${this.cell.index}`);
+            // TODO: Send telemetry this should never happen, if it does we have problems.
+            return this.result;
+        }
         this.started = true;
+
+        await this.initPromise;
         // Ensure we clear the cell state and trigger a change.
         await clearCellForExecution(this.editor, this.cell);
         if (!this.isEmptyCodeCell) {
@@ -180,9 +185,7 @@ export class CellExecution {
         this.notifyCellExecution();
 
         // Begin the request that will modify our cell.
-        kernelPromise
-            .then((kernel) => this.handleKernelRestart(kernel))
-            .then(() => this.execute(notebook.session, notebook.getLoggers()))
+        this.execute(notebook.session, notebook.getLoggers())
             .catch((e) => this.completedWithErrors(e))
             .finally(() => this.dispose())
             .catch(noop);
@@ -191,18 +194,13 @@ export class CellExecution {
      * Cancel execution.
      * If execution has commenced, then wait for execution to complete or kernel to start.
      * If execution has not commenced, then ensure dequeue it & revert the status to not-queued (remove spinner, etc).
-     */
-    public async cancel() {
-        await this.cancelInternal(false);
-    }
-    /**
      * @param {boolean} [forced=false]
-     * If `true`, then dequeue this & do not wait for execution to complete (basically kill it).
+     * If `true`, then do not wait for cell execution to complete gracefully (just kill it).
      * This is used when we restart the kernel (either as a result of kernel interrupt or user initiated).
      * When restarted, the execution needs to stop as jupyter will not send more messages.
-     * Thus `forced=true` is more like a hard kill.
+     * Hence `forced=true` is more like a hard kill.
      */
-    private async cancelInternal(forced = false) {
+    public async cancel(forced = false) {
         if (this.started && !forced) {
             // At this point the cell execution can only be stopped from kernel & we should not
             // stop handling execution results & the like from the kernel.
@@ -232,17 +230,6 @@ export class CellExecution {
         if (deferred) {
             deferred.resolve();
         }
-    }
-
-    private handleKernelRestart(kernel: IKernel) {
-        kernel.onRestarted(
-            () => {
-                traceCellMessage(this.cell, 'Kernel restart handled in CellExecution, cancelling cell');
-                this.cancelInternal(true).catch(noop);
-            },
-            this,
-            this.disposables
-        );
     }
 
     private async completedWithErrors(error: Partial<Error>) {
--- a/src/client/datascience/jupyter/kernels/cellExecution.ts
+++ b/src/client/datascience/jupyter/kernels/cellExecution.ts
@@ -198,7 +198,7 @@ export class CellExecution {
     public async cancel() {
         await this.cancelInternal(false);
     }
-    public async cancelInternal(forced = false) {
+    private async cancelInternal(forced = false) {
         if (this.started && !forced) {
             // At this point the cell execution can only be stopped from kernel & we should not
             // stop handling execution results & the like from the kernel.
--- a/src/client/datascience/jupyter/kernels/cellExecution.ts
+++ b/src/client/datascience/jupyter/kernels/cellExecution.ts
@@ -87,10 +87,6 @@ export class CellExecution {
     public get result(): Promise<NotebookCellRunState | undefined> {
         return this._result.promise;
     }
-
-    public get completed() {
-        return this._completed;
-    }
     /**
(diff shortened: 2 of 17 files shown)


src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -68,7 +65,6 @@ export class Kernel implements IKernel {
     private readonly _onStatusChanged = new EventEmitter<ServerStatus>();
     private readonly _onRestarted = new EventEmitter<void>();
     private readonly _onDisposed = new EventEmitter<void>();
-    private readonly _onInterruptTimedOut = new EventEmitter<void>();
```
DonJayamanne, line 71: This wasn't required, as the place this was required is the same code that starts the interrupt & it has information whether interrupt timed out or not. 
Hence we can just use the return value. Else we have two ways to determine if a kernel interrupt timeodout - check return value or wait for event.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -114,42 +109,21 @@ export class Kernel implements IKernel {
...
-        await this.kernelExecution.cancelCell(cell);
-    }
-    public async cancelAllCells(document: NotebookDocument) {
-        this.startCancellation.cancel();
-        await this.kernelExecution.cancelAllCells(document);
-    }
     public async start(options?: { disableUI?: boolean }): Promise<void> {
         await this.startNotebook(options);
     }
-    public async interruptCell(cell: NotebookCell): Promise<InterruptResult> {
```
DonJayamanne, line 128: Removed cancelCell, cancelAllCells, interruptCell & interruptAllCells
In Jupyter, its all the same - `interrupt`.
Hence collapsed all into a single method

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -32,48 +34,50 @@ const vscodeNotebookEnums = require('vscode') as typeof import('vscode-proposed'
...
     private readonly kernelValidated = new WeakMap<NotebookDocument, { kernel: IKernel; promise: Promise<void> }>();
 
     private readonly executionFactory: CellExecutionFactory;
     private readonly disposables: IDisposable[] = [];
+    private readonly pendingExecution = new ChainedExecutions<void>('Kernel execution failure');
+    private readonly pendingCellExecution = new ChainedExecutions<NotebookCellRunState | undefined>(
+        'Kernel execution previous cell failure'
+    );
     private isRawNotebookSupported?: Promise<boolean>;
+    private _interruptPromise?: Promise<InterruptResult>;
```
DonJayamanne, line 49: Moved the interrupt code from cellExzecution into KernelExecution.

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -32,48 +34,50 @@ const vscodeNotebookEnums = require('vscode') as typeof import('vscode-proposed'
...
-        this.cellExecutions.set(cell, cellExecution);
-
+        const cellExecution = this.createCellExecution(cell);
         try {
-            const notebook = await notebookPromise;
-            const kernel = this.getKernel(cell.notebook);
-            traceCellMessage(cell, 'executeCell started in KernelExecution');
-            await this.executeIndividualCell(kernel, cellExecution, notebook);
+            await this.pendingExecution.chainExecution(async () =>
+                this.executeQueuedCells(notebookPromise, cell.notebook)
```
DonJayamanne, line 79: Fixes to how we queued cells.
Whether you run the whole notebook or individual notebook, they all go into the same queue.
problem - run all cells & then run an individual cell was causing problems, this ensures they are queued & things don't go out of order.

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -91,117 +95,189 @@ export class KernelExecution implements IDisposable {
...
-            () => codeCellsToExecute.forEach((cell) => cell.cancel()),
-            this,
-            this.disposables
-        );
+            .forEach((cell) => this.createCellExecution(cell));
 
         try {
-            const notebook = await notebookPromise;
-            const kernel = this.getKernel(document);
-            codeCellsToExecute.forEach((exec) => traceCellMessage(exec.cell, 'Ready to execute'));
```
DonJayamanne, line 118: Moved all of this code into a new method, so that running a cell or all cells & queueing other cells end up in the same place - chained promises & in the right order.

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -91,117 +95,189 @@ export class KernelExecution implements IDisposable {
...
-
-    public async cancelCell(cell: NotebookCell) {
-        const execution = this.cellExecutions.get(cell);
-        if (execution) {
-            this.cellExecutions.delete(cell);
-            traceCellMessage(cell, 'Cancel cell from Kernel Execution');
-            await execution.cancel();
-        } else {
-            traceCellMessage(cell, 'Cannot cancel cell execution from Kernel Execution');
+    public async interrupt(notebookPromise: Promise<INotebook>, document: NotebookDocument): Promise<InterruptResult> {
```
DonJayamanne, line 118: Moved from CellExecution, else we ended up with multiple interrupts & things got into a bad state.

src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -198,12 +204,22 @@ export class NotebookEditor implements INotebookEditor {
         const status = this.statusProvider.set(DataScience.interruptKernelStatus(), true, undefined, undefined);
 
         try {
-            await kernel.interruptAllCells(this.document);
-            status.dispose();
+            const result = await kernel.interrupt(this.document);
+            if (result === InterruptResult.TimedOut) {
```
DonJayamanne, line 208: Removed event as we didn't need it anymore, now that the interrupt takes place in KernelExecution and that can communicate (return) the result directly.

src/client/datascience/jupyter/kernels/types.ts
```diff
@@ -169,8 +168,7 @@ export interface IKernel extends IAsyncDisposable {
     readonly info?: KernelMessage.IInfoReplyMsg['content'];
     readonly kernelSocket: Observable<KernelSocketInformation | undefined>;
     start(options?: { disableUI?: boolean }): Promise<void>;
-    interruptCell(cell: NotebookCell): Promise<InterruptResult>;
-    interruptAllCells(document: NotebookDocument): Promise<InterruptResult>;
+    interrupt(document: NotebookDocument): Promise<InterruptResult>;
```
DonJayamanne, line 171: No more cancelCell, cancelAllCells, interruptCell, interruptAllCells - they all mean interrupt in Jupyter world.

src/client/datascience/notebook/kernelWithMetadata.ts
```diff
@@ -37,28 +37,31 @@ export class VSCodeNotebookKernelMetadata implements VSCNotebookKernel {
...
     }
     public cancelCellExecution(_: NotebookDocument, cell: NotebookCell) {
-        this.kernelProvider.get(cell.notebook.uri)?.interruptCell(cell).ignoreErrors(); // NOSONAR
+        traceCellMessage(cell, 'Cell cancellation requested');
+        this.commandManager.executeCommand(Commands.NotebookEditorInterruptKernel).then(noop, noop);
     }
     public cancelAllCellsExecution(document: NotebookDocument) {
-        this.kernelProvider.get(document.uri)?.interruptAllCells(document).ignoreErrors(); // NOSONAR
+        traceInfo(`Document cancellation requested ${document.uri}`);
+        this.commandManager.executeCommand(Commands.NotebookEditorInterruptKernel).then(noop, noop);
```
DonJayamanne, line 64: We must use the command, so that we restart if the interrupt fails.

DonJayamanne: Using `Kernel.interrupt` was too low level an API to use

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -91,10 +88,6 @@ export class CellExecution {
         return this._result.promise;
     }
 
-    public get token(): CancellationToken {
```
DonJayamanne, line 94: Cleanup, cells cannot cancel them selves.

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -252,72 +229,16 @@ export class CellExecution {
...
         traceCellMessage(this.cell, 'Execution disposed');
-        this.activeExecution = undefined;
         this.disposables.forEach((d) => d.dispose());
         const deferred = CellExecution.cellsCompletedForTesting.get(this.cell);
         if (deferred) {
             deferred.resolve();
         }
     }
 
-    private async interruptExecution(
```
DonJayamanne, line 263: Moved to KenrelExecution, else we ended up with multiple interrupts, one for each cell.

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -234,15 +219,7 @@ export class CellExecution {
...
         this.cancelHandled = true;
         await this.initPromise;
-        // We need to notify cancellation only if execution is in progress,
-        // coz if not, we can safely reset the states.
-        if (this.started) {
-            this.source.cancel();
-        }
 
-        if (!this.started) {
-            await this.dequeue();
```
DonJayamanne, line 244: Same as `completedDueToCancellation`

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -32,48 +34,50 @@ const vscodeNotebookEnums = require('vscode') as typeof import('vscode-proposed'
...
     ) {
         this.executionFactory = new CellExecutionFactory(errorHandler, editorProvider, appShell, vscNotebook, context);
     }
 
     @captureTelemetry(Telemetry.ExecuteNativeCell, undefined, true)
     public async executeCell(notebookPromise: Promise<INotebook>, cell: NotebookCell): Promise<void> {
-        // Cannot execute empty cells.
-        if (this.cellExecutions.has(cell) || cell.document.getText().trim().length === 0) {
-            // clear cell output
-            cell.outputs = [];
```
DonJayamanne, line 62: This was incorrect, we must use the VSC Edit API to update the cell output.
Moved updating this into cellExecution. So that everything goes through cell execution (consistent way to update output).

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -91,117 +95,190 @@ export class KernelExecution implements IDisposable {
...
+        if (!editor || !stackOfCellsToExecute) {
+            return;
+        }
+        try {
+            const notebook = await notebookPromise;
+            const kernel = this.getKernel(document);
+            stackOfCellsToExecute.forEach((exec) => traceCellMessage(exec.cell, 'Ready to execute'));
+            while (stackOfCellsToExecute.length) {
+                // Stack of cells to be executed, this way we maintain order of cell executions.
+                const cellToExecute = stackOfCellsToExecute[0];
```
DonJayamanne, line 246: Maintain order of execution of cells.

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -221,21 +298,13 @@ export class KernelExecution implements IDisposable {
...
-                traceCellMessage(cellExecution.cell, 'Cell cancellation requested');
-                this.commandManager.executeCommand(Commands.NotebookEditorInterruptKernel).then(noop, noop);
-            },
-            this,
-            this.disposables
-        );
-
-        // Start execution
-        await cellExecution.start(kernelPromise, notebook);
+        return this.pendingCellExecution.chainExecution(async () => {
```
DonJayamanne, line 301: Ensure all executions are chained.

src/client/common/utils/async.ts
```diff
@@ -237,3 +239,22 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
     }
     return results;
 }
+
+export class ChainedExecutions<T> {
```
DonJayamanne, line 243: Could be deemed overkill to create a class, but i had to copy this code into two methods, hence decided to create a class...

src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -268,16 +284,6 @@ export class NotebookEditor implements INotebookEditor {
         }
     }
 
-    private async onInterruptTimedOut(_: IKernel) {
```
DonJayamanne, line 271: This would restart the wrong kernel. If we had two notebooks, this would result both kernels being restarted.

DonJayamanne:
Need to try and add some more tests to ensure we guarantee the order of executions.
Thought i'd get the code reviewed initially.

src/client/common/utils/async.ts
```diff
@@ -237,3 +237,16 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
...
+    private pendingCellExecution?: Promise<T> | undefined;
+    public async chainExecution(next: () => Promise<T>): Promise<T> {
+        if (this.pendingCellExecution) {
+            await this.pendingCellExecution;
```
rchiodo, line 245: I believe this may screw stuff up if there's an error. The error will continually come out.

I think you should clear pendingCellExecution before awaiting it. I think we can await it only once.

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -91,117 +93,190 @@ export class KernelExecution implements IDisposable {
...
+    private async cancelAllCells(notebookPromise: Promise<INotebook>, document: NotebookDocument): Promise<void> {
+        await this.interrupt(notebookPromise, document);
+        await this.waitForAllQueuedCells(document);
     }
 
-    public async cancelAllCells(document: NotebookDocument): Promise<void> {
-        if (this.documentExecutions.get(document)) {
-            this.documentExecutions.get(document)!.cancel();
+    private async waitForAllQueuedCells(document: NotebookDocument, interrupt = false) {
+        traceInfo('Interrupt kernel execution');
```
rchiodo, line 145: This traceInfo is confusing. I'm not entirely sure what this method does? REturns the list of promises for executing cells?

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -91,117 +93,190 @@ export class KernelExecution implements IDisposable {
...
+        const cellExecutions = Array.from(cellsToCancel)
+            .map((cell) => this.cellExecutions.get(cell))
+            .filter((item) => item !== undefined)
+            .map((item) => item!);
+
+        // Wait for all to complete.
+        await Promise.all(
+            cellExecutions.map(async (item) => {
+                if (interrupt) {
+                    await item.interrupt();
```
rchiodo, line 165: Now I'm really confused. I thought cellExecution didn't have interrupt anymore?

src/client/datascience/notebook/notebookEditor.ts
rchiodo: Does the command route here? That would be how this could work.

src/client/common/utils/async.ts
rchiodo: You can test this by removing ipykernel, run a cell, then attempt to run again.

src/client/datascience/notebook/notebookEditor.ts
DonJayamanne: Yes, the command ends up calling this code (`INotebookEditor.interruptKernel`).

src/client/common/utils/async.ts
DonJayamanne: Yup missed that, thanks

src/client/datascience/jupyter/kernels/kernelExecution.ts
DonJayamanne: Already on this, I too realized this & used existing `cancel`

src/client/common/utils/async.ts
```diff
@@ -237,3 +237,32 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
     }
     return results;
 }
+
+export class ChainedExecutions<T> {
+    private pendingCellExecution?: Promise<T> | undefined;
+    public clear() {
```
DonJayamanne, line 243: Not used, this is a WIP (trying to figure out if i need this or ensure `chainExecution` never caches & ends up in a weird state

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -225,24 +193,27 @@ export class CellExecution {
     }
     /**
      * Cancel execution.
-     * If execution has commenced, then interrupt (via cancellation token) else dequeue from execution.
+     * If execution has commenced, then wait for execution to complete or kernel to start.
      */
     public async cancel() {
+        await this.cancelInternal(false);
+    }
+    public async cancelInternal(forced = false) {
```
DonJayamanne, line 201: Oops shouldn't be public

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -91,117 +99,199 @@ export class KernelExecution implements IDisposable {
...
     }
+    /**
+     * Interrupts the execution of cells.
+     * If we don't have a kernel (Jupyter Session) available, then just abort all of the cell executions.
+     */
+    public async interrupt(document: NotebookDocument, notebookPromise?: Promise<INotebook>): Promise<InterruptResult> {
+        // Possible we don't have a notebook.
+        const notebook = notebookPromise ? await notebookPromise.catch(() => undefined) : undefined;
+        traceInfo('Interrupt kernel execution');
+        // Interrupt all the cells, & wait for all to complete.
```
rchiodo, line 135: THis is still a little confusing. Maybe the comment should be:

```
// Get the execution promises for all of the pending cells and cancel all non executing cells.
```

At least I think that's what it _should_ be doing? However it looks like it cancels cells that are running too? Whereas a cancel doesn't cancel? 

Okay now I'm totally confused again.

rchiodo: Shouldn't on interrupt waitForAllQueuedCells have a value of false? It's not cancelling everyting, it's interrrupting it all. Although it should cancel pending cells?

rchiodo: Maybe it would be better to split it into more parts?

Part 1 - Get pending executions
Part 2 - Interrupt session and wait for pending executions
Part 3 - Cancel all non pending executions

src/client/datascience/jupyter/kernels/kernelExecution.ts
DonJayamanne: Fixed, renamed method.

src/client/datascience/jupyter/kernels/kernelExecution.ts
DonJayamanne: Let me try something else, I don't want to differentiate make a differentiation between cancelling a cell that has started vs cell that hasn't.
`CellExecution.cancel` should be agnostic of the state of the cell, its just a command for `cancel what ever you are doing`.

Trying to split it means we cancel pending cells & cancel non-pending cells separately when I don't think we should.
Will look to clean up this code.

DonJayamanne: But agree, the code is a little confusing, too much going on

rchiodo: What if you don't cancel the pending cells in the first step. Just get their list. Then cancel after the session interrupt.

DonJayamanne: Let me know if the new code is better (thanks, i think its better - but that doesn't mean it can't get better), the feedback is most certainly useful & helping me make it more readable..

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -91,117 +99,211 @@ export class KernelExecution implements IDisposable {
...
+        // Possible we don't have a notebook.
+        const notebook = notebookPromise ? await notebookPromise.catch(() => undefined) : undefined;
+        traceInfo('Interrupt kernel execution');
+        // First cancel all the cells & then wait for them to complete.
+        // Both must happen together, we cannot just wait for cells to complete, as its possible
+        // that cell1 has started & cell2 has been queued. If Cell1 completes, then Cell2 will start.
+        // What we want is, if Cell1 completes then Cell2 should not start (it must be cancelled before hand).
+        const pendingCells = this.cancelAllPendingCells(document).then(() =>
+            this.waitForPendingCellsToComplete(document)
+        );
```
rchiodo, line 141: Much easier to reason about now 👍

rchiodo (approved):

src/client/datascience/jupyter/kernels/kernelExecution.ts
DonJayamanne: Will try to move all of the code that does the following:
* Tracking pending cells
* Get next pending cell
* Cancel all pending cells

Basically thats more like a `PendingCellsStack` class that:
* Has a method to push into the stack
* Cancel all items in the stack
* Wait for all to complete
* Get next item from the stack

Not sure if it will be overkill, but will make the code in `KernelExecution` simpler, let me see ...

DonJayamanne: Not going with that, the code is small.

src/client/datascience/notebook/kernelWithMetadata.ts
```diff
@@ -98,17 +101,4 @@ export class VSCodeNotebookKernelMetadata implements VSCNotebookKernel {
...
-    private async chainExecution(next: () => Promise<void>): Promise<void> {
-        if (this.pendingExecution) {
-            try {
-                await this.pendingExecution;
-            } catch (e) {
-                // Errors are handled elsewhere
-                traceInfo(`Kernel execution previous failure: ${e}`);
-            }
-        }
-        this.pendingExecution = next();
```
DonJayamanne, line 111: This wasn't chaining promises correctly, possible that this method got called twice.
Both would wait for previous pending execution to complete.
Then both would end up with the `next` getting executed & only one of them will have `pendingExecution` updated with the promise of `next`. Meaning both get executed in parallel.

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -252,112 +225,51 @@ export class CellExecution {
...
-                return InterruptResult.Restarted;
-            }
-
-            // Otherwise a real error occurred.
-            throw exc;
-        } finally {
-            restartHandlerToken.dispose();
-        }
-    }
-    private handleKernelRestart(kernel: IKernel) {
```
DonJayamanne, line 318: Moved to `KernelExecution`, this was `CellExecution` never has to cancel itself.
Moved a lot of these out to kernel execution, cell execution should never cancel itself, it should be told to cancel.
Much clear, removes need to pass in `IKernel` as well.

src/client/common/utils/async.ts
```diff
@@ -237,3 +237,25 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
...
+    private pendingCellExecution: Promise<any> = Promise.resolve();
+    /**
+     * Clears the chained promises.
+     */
+    public clear() {
+        this.pendingCellExecution = Promise.resolve();
+    }
+    public async chainExecution(next: () => Promise<T>): Promise<T> {
+        const deferred = createDeferred<T>();
+        const chainedPromise = this.pendingCellExecution.then(() => {
```
DonJayamanne, line 252: Must use `then` to properly chain the promises.
Added `clear` to clear any previous results when starting a new batch of promises.
Calling `clear` is crucial for cases where we start a new batch and wish to ignore previous failures & results.

src/client/common/utils/async.ts
```diff
@@ -237,3 +237,25 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
...
+            const nextPromise = next();
+            nextPromise.then((result) => deferred.resolve(result)).catch((ex) => deferred.reject(ex));
+            return nextPromise;
+        });
+        this.pendingCellExecution = chainedPromise;
+
+        return Promise.race([deferred.promise, chainedPromise]);
```
DonJayamanne, line 259: Either return when the new succeeds or the chain get rejected.

src/client/datascience/jupyter/kernels/kernelExecution.ts
rchiodo: Yes definitely better.

src/client/common/utils/async.ts
rchiodo: This seems rather complicated. Could you do this instead?

```
let pendingError = false;
const pending = this.pending ?? Promise.resolve():
this.pending = pending.then(next).catch((e) => { if pendingError { throw e; } return next(); });
return this.pending;
```

That should:
1) Chain the promises
1) Only throw a pending exception once

I believe that's the behavior we want? 

I think your new version does not work if pending fails.

rchiodo: Ah I see you require the 'clear' to be called. That is kinda sucky. It woudl be better if we could eliminate that interdependency.

src/client/common/utils/async.ts
IanMatthewHuff: Class seems good to me.

src/client/common/utils/async.ts
```diff
@@ -237,3 +237,25 @@ export async function flattenIterator<T>(iterator: AsyncIterator<T, void>): Prom
     }
     return results;
 }
+
+export class ChainedExecutions<T> {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private pendingCellExecution: Promise<any> = Promise.resolve();
```
IanMatthewHuff, line 243: Would change the name here, as this class is now general so pendingCellExecution is misleading. Might be something else pending.

IanMatthewHuff: Also is this a Promise<T> here?

IanMatthewHuff: NVM, I guess it can't be since it's empy Promse.resolve() at the start.

src/client/datascience/jupyter/kernels/cellExecution.ts
IanMatthewHuff: I like these moves. Makes it more clear. I remember being confused some on previous reviews by the cancellation / interrupt flow.

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -32,48 +35,53 @@ const vscodeNotebookEnums = require('vscode') as typeof import('vscode-proposed'
...
 
     @captureTelemetry(Telemetry.ExecuteNativeCell, undefined, true)
     public async executeCell(notebookPromise: Promise<INotebook>, cell: NotebookCell): Promise<void> {
-        // Cannot execute empty cells.
-        if (this.cellExecutions.has(cell) || cell.document.getText().trim().length === 0) {
-            // clear cell output
-            cell.outputs = [];
+        // Return current execution.
+        if (this.cellExecutions.get(cell)) {
+            traceError(`Cell already executing/queued for execution, requested re-execution of ${cell.index}`);
```
IanMatthewHuff, line 69: Is this an error? Just wondering if it should be traceInfo. Seems like if it's requested again we just return the existing result, which seems ok to me. Something that we don't think should ever happen?

src/client/datascience/jupyter/kernels/kernelExecution.ts
IanMatthewHuff: Agreed

IanMatthewHuff (approved):

src/client/common/utils/async.ts
DonJayamanne: Lold yes

src/client/datascience/jupyter/kernels/kernelExecution.ts
DonJayamanne: > Is this an error? 

Kinda yes, error in the work flow. This should never happen. You cannot execute a cell that's already being executed.

> Just wondering if it should be traceInfo. 

I want a traceError so that this stands out, as this should never happen, if it does then we know immediately that something is wrong.
I plan on adding telemetry for this as well. Found that we were re-executing cells that were already executed & that resulted in weird UI state..

> Something that we don't think should ever happen?

Yes, add comments

src/client/common/utils/async.ts
DonJayamanne: Sure, will try that.

DonJayamanne: I see a problem;
> this.pending = pending.then(next).catch((e) => { if pendingError { throw e; } pendingError = false; return next(); });

Assume previous pending failed.
Now it will go into the catch block and the exception is swallowed, because of the `catch` clause.
So when we chain a new proimise on top of this, the previous promise failed, & then due to the chaining we ignore that & continue to the next promise.

Ideally the whole thing should just fail. Else its not a chained. Its more like throw once & continue.

DonJayamanne: Basically we want everything that gets chained to that to stop running

DonJayamanne: Feel like we need a separate class that manages the batch of executions. If one thing fails, then everything in that batch shoulld stop (single thing should shortcircuit the whole operation).

This way if we have a batch of executions already running, just add the new cell to the existing batch. & if one cell fails (errors for kernel or what ever reason), then stop all.

& if a batch doesn't exist, then create a new batch.

& once that batch completes, then tear down that class & dispose everything in it.

Personally I prefer this option, not sure if its over kill.
Today we kinda do that by using the `stack` in `this.stackOfCellsToExecuteByDocument`. If we have an array with items, this means we have an execution going on & we just append the new execution to that job.

@IanMatthewHuff  @rchiodo thoughts?
Or should we just meet up and discuss this.

DonJayamanne: At the same time i do'nt want to over complicate it, but would also like to simplify it..
question is, is the current code clear enough...

rchiodo: I believe this is what we want. 

>  Else its not a chained. Its more like throw once & continue

Well the reason for that was the caller would keep getting the same error over and over. It would never retry. Maybe your clear idea would be better then?

rchiodo: So then we could just have this:

```
const pending = this.pending ?? Promise.resolve();
this.pending = pending.then(next);
return this.pending;
```

and then do the clear like you have now?

rchiodo (approved):

Issue #4474, opened 2021-01-24
Native Notebooks - Variable explorer doesn't reset with kernel restart


claudiaregio:
### Steps to reproduce: 
1. Open notebook
2. Run cells
3. Open variable explorer
4. Restart kernel 

### Expected behavior
Variable explorer should be empty 

### Actual behavior 
Variables from last run remain 

I have to restart the kernel, clear all outputs, close and reopen the notebook

rchiodo:
Dupe of #4492

Issue #4492, opened 2021-01-25
Handle kernel restarts correctly with the variableview


IanMatthewHuff:
Correctly handle kernel restarts with the variable view

joyceerhl:
This is kind of weird. If I open the data viewer from the variable view, switch back to the notebook, and request a cell restart, the variable view fails to clear even after the restart:

![restart](https://user-images.githubusercontent.com/30305945/106097587-6b14b680-60ec-11eb-85c3-abf2b8ee0b7e.gif)

Works fine as long as you don't open the data viewer though.

IanMatthewHuff:
@joyceerhl Good catch. I'll pull this back to in progress and investigate.

DonJayamanne:
Validated

PR #4517, opened 2021-01-27
Fix variable view on kernel restart

rchiodo:
For #4474 

Add restart handling for new variable view. Make sure variable provider is updating on refresh.

### DIFF
--- a/news/2 Fixes/4474.md
+++ b/news/2 Fixes/4474.md
@@ -1,0 +1,1 @@
+Refresh variable view when restarting a kernel.
--- a/src/client/datascience/jupyter/kernelVariables.ts
+++ b/src/client/datascience/jupyter/kernelVariables.ts
@@ -342,7 +342,12 @@ export class KernelVariables implements IJupyterVariables {
     ): Promise<IJupyterVariablesResponse> {
         // See if we already have the name list
         let list = this.notebookState.get(notebook.identity);
-        if (!list || list.currentExecutionCount !== request.executionCount) {
+        if (
+            !list ||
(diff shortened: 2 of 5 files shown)


IanMatthewHuff (changes requested):
I'm really sorry. We ended up with two issues on this, my fault I didn't see that the claudia had opened the new issue. I have #4492 in progress. When looking at restart I'm seeing some bigger issues with switching between two files (at times) so I think that I have a more complete fix that should cover this coming tomorrow. Can you hold off until that?

rchiodo:
Oh no worries. What's the problem with multiple notebooks? Seems to work at least after my change.

IanMatthewHuff:
> Oh no worries. What's the problem with multiple notebooks? Seems to work at least after my change.

So initially I just added a refresh tracker like you did (I will steal your event name changes, they are much better) and hooked it up to the refresh event. But was still seeing issues both with paging and with the refresh event not clearing the variables sometimes (after moving from one file to another).

Honestly I'm a bit embarrassed, my initial work does pretty fine for a single file, but it's not tracking the execution order well. There is no update for execution order when changing a file, so it would refresh on file change but it could refresh with either the execution order from the old file (if you had not executed anything in the new file after switching to it) or with 0 if you had a reset on one file or another. My change now tracks executions on all files and restores the execution count when you change to that file as the active file. Also had to change the ordering a bit on the updateExecutionCount action. This way kernelVariables won't have to change, and it will correctly use the cached list in kernelVariables if the execution count doesn't change.

PR #4518, opened 2021-01-27
Add more tests for switching kernels

rchiodo:
For #4163 

Turn on more tests and use the new api provided by VS code to switch kernels. Somehow it all works? I can't believe the test run passed.

See test run here:
https://github.com/microsoft/vscode-jupyter/runs/1773425848?check_suite_focus=true

### DIFF
--- a/src/test/datascience/notebook/helper.ts
+++ b/src/test/datascience/notebook/helper.ts
@@ -19,7 +19,7 @@ import {
     NotebookDocument
 } from '../../../../typings/vscode-proposed';
 import { IApplicationEnvironment, IApplicationShell, IVSCodeNotebook } from '../../../client/common/application/types';
-import { MARKDOWN_LANGUAGE, PYTHON_LANGUAGE } from '../../../client/common/constants';
+import { JVSC_EXTENSION_ID, MARKDOWN_LANGUAGE, PYTHON_LANGUAGE } from '../../../client/common/constants';
 import { disposeAllDisposables } from '../../../client/common/helpers';
 import { traceInfo } from '../../../client/common/logger';
 import {
@@ -245,7 +245,7 @@ export async function waitForKernelToGetSelected(kernelNameSearch: string) {
     const id = kernels?.find((k) => k.label.includes(kernelNameSearch))?.id;
 
     // Send a select kernel on the active notebook editor
-    void commands.executeCommand('notebook.selectKernel', { id });
+    void commands.executeCommand('notebook.selectKernel', { id, extension: JVSC_EXTENSION_ID });
     const isRightKernel = () => {
         if (!vscodeNotebook.activeNotebookEditor) {
             return false;
--- a/src/test/datascience/notebook/helper.ts
+++ b/src/test/datascience/notebook/helper.ts
@@ -254,9 +254,10 @@ export async function waitForKernelToChange(labelOrId: string | undefined) {
             return false;
         }
         if (vscodeNotebook.activeNotebookEditor.kernel.id === id) {
+            traceInfo(`Found selected kernel ${vscodeNotebook.activeNotebookEditor.kernel.id}`);
             return true;
         }
-        traceInfo(`Active kernel is ${vscodeNotebook.activeNotebookEditor.kernel.label}`);
+        traceInfo(`Active kernel is ${vscodeNotebook.activeNotebookEditor.kernel.id}`);
         return false;
     };
     await waitForCondition(async () => isRightKernel(), 15_000, `Kernel with label/id ${labelOrId} not selected`);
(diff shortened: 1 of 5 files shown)


IanMatthewHuff (approved):

joyceerhl (approved):

src/test/datascience/notebook/helper.ts
```diff
@@ -245,7 +245,7 @@ export async function waitForKernelToGetSelected(kernelNameSearch: string) {
     const id = kernels?.find((k) => k.label.includes(kernelNameSearch))?.id;
 
     // Send a select kernel on the active notebook editor
-    void commands.executeCommand('notebook.selectKernel', { id });
+    void commands.executeCommand('notebook.selectKernel', { id, extension: JVSC_EXTENSION_ID });
```
DonJayamanne, line 248: I don't think this is the right place to force the right kernel to get selected.
This forces the kernel to get selected.
We have tests that wait for VS Code to auto select the right kernel.
E.g. when we open a notebook with a java kernel, we would like the java kernel to get selected by VS Code.
This will always ensure that we forcefully select the kernel & if the code in the extension does't pre-select the right kernel, it doesn't matter as we force it in here.

DonJayamanne: I think we should remove this line & move it close to where ever we need this.
All of our existings test pass without this today.

DonJayamanne: Nope ignore my comments, the one I'm referring to is `waitForKernelToGetAutoSelected` missed `auto`. this is a new function.

DonJayamanne: Shouldn't we rename this function `changeKernel..../switchKernel...`
As we aren't waiting for it to get selected, but we're forcing a new selection

src/test/datascience/notebook/notebookEditor.vscode.test.ts
```diff
@@ -115,4 +121,32 @@ suite('Notebook Editor tests', () => {
...
+            CancellationToken.None
+        );
+        traceInfo(`Kernels found for switch kernel: ${kernels?.map((k) => k.label).join('\n')}`);
+        if (kernels?.length && kernels?.length > 1) {
+            // We have multiple kernels. Try switching
+            await waitForKernelToGetSelected(kernels[1].label);
+        }
+
+        // Execute cell and verify output
+        await executeCell(cell);
```
DonJayamanne, line 148: This doesn't verify the cell was selected.
Shouldnt we check `sys.executable`.
In the past we have had issues where kernel selection says one thing & the python environment under which things are running is different.
I'd prefer a more robust test to check `sys.exeutable`

DonJayamanne: Also guards against conda kernel selection. Where users aren't able to select conda kernels.
This way we also test changing a kernel & also executing against that kernel.

DonJayamanne (changes requested):

src/test/datascience/notebook/notebookEditor.vscode.test.ts
rchiodo: THe problem would be verifying what path was selected. Hard to predict what it should be.

src/test/datascience/notebook/helper.ts
rchiodo: Sure

src/test/datascience/notebook/notebookEditor.vscode.test.ts
rchiodo: Sorry I don't think we should change it to check the sys path. Too hard to get it to come out correct. I can't predict which kernel it's going to install and then there's a dependency between the kernels we have in main.yml and this code.

However we can determine if it's different though. I'll add that.

src/test/datascience/notebook/notebookEditor.vscode.test.ts
```diff
@@ -115,4 +122,49 @@ suite('Notebook Editor tests', () => {
...
+        cell = vscodeNotebook.activeNotebookEditor?.document.cells![0]!;
+
+        assert.strictEqual(cell?.outputs.length, 1);
+        assert.strictEqual(cell?.metadata.runState, vscodeNotebookEnums.NotebookCellRunState.Success);
+
+        if (anotherKernel && preferredKernel) {
+            const newSysPath = (cell.outputs[0] as CellDisplayOutput).data['text/plain'].toString();
+            assert.notEqual(
+                newSysPath,
+                originalSysPath,
```
DonJayamanne, line 165: Yes this is better,

src/test/datascience/notebook/notebookEditor.vscode.test.ts
DonJayamanne: Yea, checking if its different is better if we cannot determine the kernel.
At least in the case of remote Jupyter it will not work with the tests.

DonJayamanne (approved):

Issue #4485, opened 2021-01-25
Variable explorer doesn't update list of hidden variables when docked vertically.


greazer:
VS Code Insiders with Jupyter extension from main.

1. Open or create a new notebook (native notebook)
2. Add this code to a cell but don't execute yet.
```python
import time
t0 = time.perf_counter()
t1 = time.perf_counter()
t2 = time.perf_counter()
t3 = time.perf_counter()
t4 = time.perf_counter()
t5 = time.perf_counter()
t6 = time.perf_counter()
t7 = time.perf_counter()
t8 = time.perf_counter()
```
3. Open the variable viewer -> Jupyter: Open Jupyter Variable View
4. Drag and dock the viewer to the left panel.
5. Size the Variable Viewer as short as possible.
6. Execute the above cell
7. Drag the bottom of the Variable View down to reveal all the tx variable. You'll only see the first 4. The rest will say "Loading..."

joyceerhl:
Is there supposed to be an icon? It's not showing for me, I spent a couple minutes hunting for it after docking it to the left panel.

Apart from that, seems to work now.

IanMatthewHuff:
Nope, no icon. Mostly because I've always been working with it in the bottom view panel where it doesn't really need one. I'll file a new issue and loop with Claudia about getting one.

PR #4523, opened 2021-01-27
Handle variable restart, variable redocking, paging in data

IanMatthewHuff:
For #4492, #4485

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1224,6 +1224,7 @@ module.exports = {
             }
         ],
         'react/jsx-uses-vars': 'error',
+        'react/jsx-uses-react': 'error',
         strict: 'off'
     },
     settings: {
--- a/src/client/datascience/variablesView/notebookWatcher.ts
+++ b/src/client/datascience/variablesView/notebookWatcher.ts
@@ -15,7 +15,6 @@ import { INotebookWatcher } from './types';
 // NOTE: Currently this class is only looking at native notebook documents
 @injectable()
 export class NotebookWatcher implements INotebookWatcher {
-    //public get onDidChangeActiveVariableViewNotebook(): Event<INotebook | undefined> {
     public get onDidChangeActiveVariableViewNotebook(): Event<{ notebook?: INotebook; executionCount?: number }> {
         return this._onDidChangeActiveVariableViewNotebook.event;
     }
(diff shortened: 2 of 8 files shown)


IanMatthewHuff:
This fixes a number of issues with the variable view, ran into them while fixing the kernel restart issue. 

Not happy with my original work. That work used execution count basically just as an increasing number on the variable view. Which works reasonably well for a single file, but not for multiple files or handling restart. Now I'm tracking the current execution count of open documents, and passing that to the variable view when we switch documents. This way if you have document A on execution count 3, then you swap to document B on execution count 5 it will correctly fetch variables with execution count 5. Previously it would use execution count 3 (unless you actually executed a cell on document B).

IanMatthewHuff:
A note on tests. This change here is (in my opinion) the last current experiment blocking change on the variable view. I'm rushing this out a bit since tests were blocked on the CI xvfb issue. After this PR I'm going to go back and prioritize that test issue and no new features until the basic tests are working CI. I don't like working like this, but given that I'm not sure what the CI issue is this at least gets us in better state for the experiment / demos.

.eslintrc.js
```diff
@@ -1224,6 +1224,7 @@ module.exports = {
             }
         ],
         'react/jsx-uses-vars': 'error',
+        'react/jsx-uses-react': 'error',
```
IanMatthewHuff, line 1227: Just another JSX error that raw eslint incorrectly flags.

src/client/datascience/variablesView/notebookWatcher.ts
```diff
@@ -2,31 +2,46 @@
...
+}
 
 // For any class that is monitoring the active notebook document, this class will update you
 // when the active notebook changes or if the execution count is updated on the active notebook
 // NOTE: Currently this class is only looking at native notebook documents
 @injectable()
 export class NotebookWatcher implements INotebookWatcher {
-    public get onDidChangeActiveVariableViewNotebook(): Event<INotebook | undefined> {
-        return this._onDidChangeActiveVariableViewNotebook.event;
+    public get onDidChangeActiveNotebook(): Event<IActiveNotebookChangedEvent> {
```
IanMatthewHuff, line 23: The names that Rich had in his PR were better, so I stole 'em.

src/client/datascience/variablesView/variableView.ts
```diff
@@ -141,19 +137,22 @@ export class VariableView extends WebviewViewHost<IVariableViewPanelMapping> imp
             visible = this.webviewView.visible;
         }
         context.set(visible).ignoreErrors();
+
+        // Also perform a refresh on a visibily change
+        this.postMessage(InteractiveWindowMessages.ForceVariableRefresh).ignoreErrors();
```
IanMatthewHuff, line 142: Also from Rich's PR this fixes the re-docking issue that Jim was seeing.

src/datascience-ui/variable-view/variableViewPanel.tsx
```diff
@@ -118,7 +118,7 @@ export class VariableViewPanel extends React.Component<IVariableViewPanelProps>
 
     private pageInVariableData = (startIndex: number, pageSize: number) => {
         this.props.getVariableData(
-            this.props.currentExecutionCount,
+            this.props.variableState.currentExecutionCount,
```
IanMatthewHuff, line 121: This code I had actually copied over from the interactivePanel code. Was causing all the page in issues. With my code now I'm only updating and using the variableState from the main state. This actually looks to me like it might be "wrong" in the existing old editor code, but I didn't want to change that for no reason.

IanMatthewHuff: Think that it worked fine in the old code as both main state currentExecutionCount and variableState.currentExecutionCount were in sync. But since I'm just using variable state main state currentExecutionCount was not getting updated.

IanMatthewHuff: So all pageIn data was initially getting fetched with 0.

rchiodo (approved):

Issue #4453, opened 2021-01-22
Can run a Julia notebook test is failing, disabled


IanMatthewHuff:

IanMatthewHuff:
Failing > 50% of the time on CI runs.

joyceerhl:
Validated. Passed on latest successful main branch run: https://github.com/microsoft/vscode-jupyter/actions/runs/516679030

PR #4530, opened 2021-01-28
Fix julia test to pass again

rchiodo:
For #4453

I believe the root cause was not waiting long enough for the cell to execute. Default ```waitForExecutionCompletedSuccessfully``` is only 15 seconds.

Should be passing now. At least it did on this run here:
https://github.com/microsoft/vscode-jupyter/actions/runs/516574919

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,6 @@ on:
       - 'release'
       - 'release/*'
       - 'release-*'
-      - rchiodo/fix_julia_test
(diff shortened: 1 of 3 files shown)


DavidKutu (approved):

IanMatthewHuff:
I had thought that only the initial start of Julia was slow, not cell execution. But if you get good results on this, then works for me.

IanMatthewHuff (approved):

PR #4533, opened 2021-01-28
Reuse existing cell execution

DonJayamanne:
Found a bug, run a cell, then run whole document.
We end up creating a new cell execution, when we shouldn't (potential for cell the run again)

### DIFF
--- a/src/client/datascience/jupyter/kernels/kernelExecution.ts
+++ b/src/client/datascience/jupyter/kernels/kernelExecution.ts
@@ -332,6 +332,10 @@ export class KernelExecution implements IDisposable {
         );
     }
     private createQueuedCellExecution(cell: NotebookCell) {
+        const existingCellExecution = this.cellExecutions.get(cell);
+        if (existingCellExecution) {
+            return existingCellExecution;
+        }
         if (!this.stackOfCellsToExecuteByDocument.has(cell.notebook)) {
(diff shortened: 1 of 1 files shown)


src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -332,6 +332,10 @@ export class KernelExecution implements IDisposable {
         );
     }
     private createQueuedCellExecution(cell: NotebookCell) {
+        const existingCellExecution = this.cellExecutions.get(cell);
```
IanMatthewHuff, line 335: Good catch. 

Seems like a fine fix for now, but I did have a bit of an architectural question here. Should either the execution factory, or something that contains the execution factory be wrapping both the existing check and the create function?

Bit of a tricky thing at times with these small services. Someone else working in this class could call executionFactory.create without the check that you've added here, and again mess things up. If it's the case that we should never, ever create a cell execution when one already exists, then it might make sense to have another layer here or a structural change that would prevent that from happening.

rchiodo: Yeah maybe this could be refactored when you make the other refactor change. Make it impossible to have more than one execution for the same cell.

rchiodo (approved):

IanMatthewHuff (approved):

PR #4534, opened 2021-01-28
Refactor KernelExecution & CellExecution

DonJayamanne:
Will add new tests in another PR


src/client/datascience/jupyter/kernels/cellExecutionJob.ts
```diff
@@ -0,0 +1,187 @@
...
+import { CellExecution, CellExecutionFactory } from './cellExecution';
+// eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
+const vscodeNotebookEnums = require('vscode') as typeof import('vscode-proposed');
+
+/**
+ * A job responsible for execution of cells.
+ * If this has not completed execution of the cells queued, we can continue to add more cells to this job.
+ * All cells queued using `runCells` are added to the queue and processed in order they were added/queued.
+ */
+export class CellExecutionJob {
```
DonJayamanne, line 21: No idea why i kept calling it a stack, its not a stack, its a queue.
Renamed class to remove references to implementation detail.

rchiodo: Maybe the name should be CellExecutionQueue then?

src/client/datascience/jupyter/kernels/cellExecutionJob.ts
```diff
@@ -0,0 +1,189 @@
...
+        private readonly isPythonKernelConnection: boolean
+    ) {
+        // If the editor is closed, then just stop handling the UI updates.
+        this.editor.onDidDispose(() => this.cancel(true), this, this.disposables);
+        this.completion.promise.finally(() => this.dispose()).catch(noop);
+    }
+    /**
+     * Run cells and wait for completion.
+     */
+    public async runCells(cells: NotebookCell[]) {
```
rchiodo, line 49: Then this would be 'queue' cells?

src/client/datascience/jupyter/kernels/cellExecutionJob.ts
```diff
@@ -0,0 +1,189 @@
...
+        const chainedExecution = this.chainedCellExecutionPromise.then(async () => {
+            traceCellMessage(cellExecution.cell, 'Get cell from queue for execution');
+            // Start execution
+            await cellExecution.start(notebook);
+
+            // The result promise will resolve when complete.
+            const promise = cellExecution.result;
+            promise
+                .finally(() => traceCellMessage(cellExecution.cell, 'Cell from queue completed execution'))
+                .catch(noop);
```
rchiodo, line 159: Why would exceptions be swallowed?

rchiodo:
DOes feel simpler but might it be broken down further? The whole cellExecutions map and queue of cells and completion promise is very confusing. Can we further separate it where the queue is separate from the other two parts?

src/client/datascience/jupyter/kernels/cellExecutionJob.ts
DonJayamanne: We return the original promise and don't swallow those.
I'm adding a finally for logging (so the original promise is still left untouched).
Will simplify this, i was just guarding against traceCellMesage failing & unnecessarily causing other errors, hut `traceCellMessage` cannot fail.

rchiodo: Oh yeah duh. You're not setting the promise to the catch result.

src/client/datascience/jupyter/kernels/cellExecutionJob.ts
DonJayamanne: Done

src/client/datascience/jupyter/kernels/cellExecutionJob.ts
```diff
@@ -0,0 +1,189 @@
...
+    private readonly disposables: IDisposable[] = [];
+    private readonly completion = createDeferred<void>();
+    private cancelledOrCompletedWithErrors = false;
+    private startedRunningCells = false;
+    private chainedCellExecutionPromise: Promise<NotebookCellRunState | undefined> = Promise.resolve(undefined);
+    /**
+     * Whether all cells have completed processing or cancelled, or some completed & others cancelled.
+     * Even if this property is true, its possible there is still some cleanup remaining (or some promises waiting to be completed).
+     */
+    public get completed(): boolean {
```
DonJayamanne: Should this be renamed to `isEmpty`, as this class represents a `Queue`, the name `isEmpty` feels more appropriate than `completed`

@rchiodo @IanMatthewHuff /cc

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -0,0 +1,183 @@
...
+    private readonly disposables: IDisposable[] = [];
+    private readonly completion = createDeferred<void>();
+    private cancelledOrCompletedWithErrors = false;
+    private startedRunningCells = false;
+    private chainedCellExecutionPromise: Promise<NotebookCellRunState | undefined> = Promise.resolve(undefined);
+    /**
+     * Whether all cells have completed processing or cancelled, or some completed & others cancelled.
+     * Even if this property is true, its possible there is still some cleanup remaining (or some promises waiting to be completed).
+     */
+    public get completed(): boolean {
```
DonJayamanne, line 33: As this is class represents a queue, should this property be renamed to `isEmpty`, that feels more appropriate for the class, but would that convey the same meaning?

@rchiodo @IanMatthewHuff /cc

rchiodo: IsEmpty does sound better but the comment above scares me. ```Even if this property is true, it's possible there is still some cleanup remaining```. Uh why is that is the first thing I'd ask. Can we fix it?

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -0,0 +1,183 @@
...
+    private queueCellForExecution(cell: NotebookCell) {
+        const existingCellExecution = this.cellExecutions.get(cell);
+        if (existingCellExecution) {
+            traceCellMessage(cell, 'Use existing cell execution');
+            return existingCellExecution;
+        }
+        const cellExecution = this.executionFactory.create(cell, this.isPythonKernelConnection);
+        this.cellExecutions.set(cell, cellExecution);
+        this.queueOfCellsToExecute.push(cellExecution);
+        cellExecution.result.finally(() => this.onCellExecutionCompleted(cellExecution));
```
IanMatthewHuff, line 169: Trying to trace it, looks like this onCellExecutionCompleted might be happening twice? It's on the finally for the result here, and happening when it's executed in executeQueuedCells.

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -33,26 +34,20 @@ const vscodeNotebookEnums = require('vscode') as typeof import('vscode-proposed'
  * Else the `Kernel` class gets very big.
  */
 export class KernelExecution implements IDisposable {
-    private readonly cellExecutions = new WeakMap<NotebookCell, CellExecution>();
-
-    private readonly documentExecutions = new WeakMap<NotebookDocument, CancellationTokenSource>();
-    private readonly stackOfCellsToExecuteByDocument = new WeakMap<NotebookDocument, CellExecution[]>();
-
+    private readonly documentExecutions = new WeakMap<NotebookDocument, CellExecutionQueue>();
```
IanMatthewHuff, line 37: This is where the refactor helps. Some are in the new cellExecutionQueue class, but previously there were just too many tracking queues and maps being juggled here. cellExecutions, documentExecutions, stackOfCellsToExecuteByDocument, kernelValidated, pendingExecution, pendingCellExecution. This class is much improved now, even if we still have to deal with some complexity in the cellExecutionQueue.

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
IanMatthewHuff: Yeah, I noticed that it signals this to true at the start of the error or cancel, not when the cancel is completed.

DonJayamanne:
> DOes feel simpler but might it be broken down further? The whole cellExecutions map and queue of cells and completion promise is very confusing. Can we further separate it where the queue is separate from the other two parts?

Removed the map, & simplified the promise.
@rchiodo /cc

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -0,0 +1,165 @@
...
+     * If cells are cancelled, they are not processed, & that too counts as completion.
+     */
+    public async waitForCompletion(cells: NotebookCell[] = []): Promise<void> {
+        if (cells.length === 0) {
+            return this.completion.promise;
+        }
+        const executions = Promise.all(
+            this.queueOfCellsToExecute.filter((item) => cells.includes(item.cell)).map((cell) => cell.result)
+        );
+        await Promise.race([executions, this.completion.promise]);
```
rchiodo, line 86: Not sure why there's a race here? Is there a scenario where the executions don't finish but somehow we're done?

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -0,0 +1,165 @@
...
+            }
+            this.onCellExecutionCompleted(cellToExecute);
+        }
+    }
+    private async executeIndividualCell(
+        cellExecution: CellExecution,
+        notebook: INotebook
+    ): Promise<NotebookCellRunState | undefined> {
+        traceCellMessage(cellExecution.cell, 'Push cell into queue for execution');
+        const chainedExecution = this.chainedCellExecutionPromise.then(async () => {
```
rchiodo, line 142: Wondering why the chainedCellExecutionPromise is necessary anymore? Doesn't the queue prevent any executions from happening at the same time?

IanMatthewHuff (approved):

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
DonJayamanne: > Uh why is that is the first thing I'd ask. Can we fix it?

Unfortunately i don't see an easy way.
Lets assume we queue 10 cells, and something fails.
When there's a failure we have to do two things:
1. Notify users of this class that this class CANNOT be used anymore (has it has failed)
2. We have to cancel other cells (this is an async operation - updating cell status).

So, step 1 is crucial. If for instance a user runs another cell in the time being, we need to ensure it does not get added to this queue, as this queue is in an error state & we need to create a whole new queue.

I could instead add two properties `isEmpty` and `failed`.
If `failed=true`, we still have pending updates that require updates to cancel other cells. But the caller knows that this cannot be used.
The caller then uses the two properties instead of using the single property `completed`.

Hope this makes sense.

rchiodo: Two properties sounds better to me :)

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
DonJayamanne: E.g. here's the queue of cells `A B C D E`
Assume user runs cell A, then B, then C, then D, then E in order (one by one).
The KernelExecution calls `queue` & needs to wait until each cell is complete.
So, when cell A is queued KernelExecution waits until thats executed. It will call `waitForCOmpletion(A)`.

But we should not have to wait for ALL cells to complete just to let the caller know that cell `B` has completed.
I.e. cell B can complete even while C, D, E are still in the queue & getting processed.

DonJayamanne: Here's the code form KernelExecution
```typescript
    public async executeCell(.....) {
		...

        const executionQueue = this.getOrCreateCellExecutionQueue(editor, notebookPromise);
        executionQueue.queueCell(cell);
        await executionQueue.waitForCompletion([cell]);
	}
```

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
DonJayamanne: Oh yes, you're right.

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
rchiodo: So shouldn't there be a specific completion promise tied to the input passed in then? Why is it waiting for all executions at all?

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
DonJayamanne: Managed to clean up even more, thanks

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
DonJayamanne: Yes, i can turn `queueCell` into a promise then.
Initially that's what i did, when it was called `runCell`.
I can leave it as `queueCell(): Promise<void>` if that works. Then one doesn't need to call `waitForCompletion`. 
But its weird that the `queueCell` method is async, one does't expect a queueing operation to be async.

DonJayamanne: Heres the suggestion
```typescript
    public async executeCell(.....) {
		...

        const executionQueue = this.getOrCreateCellExecutionQueue(editor, notebookPromise);
		await executionQueue.queueCell(cell);
	}
```

rchiodo: Maybe the name shouldn't be queue then? Just execute? 

Why does kernel execution need to know it's even using a queue? It's just a service it uses to dispatch execution.

rchiodo (approved):

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
DonJayamanne: Its called by IKernel that is async.

DonJayamanne: However I would still have the race condition in a private method, I.e. the Race doesn't go away.
Is that ok?

rchiodo: Not sure why we have the race still? Seems like the executions complete or they don't? Sorry maybe you explained this above. Let me look some more.

rchiodo: So the executions is the cell executions that were asked to run. That makes sense. The caller expects to wait for those. I still don't get why completionPromise is needed? When will that resolve before the execution promises?

DonJayamanne: Only one place, when executing a cell fails.
I.e. if `executeQueueCells` fails for any reason (kernel error, notebook doesn't start, etc...). In such cases the cells complete without errors, but the error needs to be propagated.

DonJayamanne: On second thought, i don't think we do anything with those errors, let me check.
We just need to know whether the cells completed or not (with or without errors).
Give me a moment to double check

DonJayamanne: Yes, removed the promise.
Let me know what you think about the wait now...

DonJayamanne: Thanks, getting a lot cleaner now

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -0,0 +1,157 @@
...
+    public async cancel(forced?: boolean): Promise<void> {
+        this.cancelledOrCompletedWithErrors = true;
+        traceInfo('Cancel pending cells');
+        await Promise.all(this.queueOfCellsToExecute.map((item) => item.cancel(forced)));
+    }
+    /**
+     * Wait for cells to complete (for for the queue of cells to be processed)
+     * If cells are cancelled, they are not processed, & that too counts as completion.
+     */
+    public async waitForCompletion(cells: NotebookCell[] = []): Promise<void> {
```
rchiodo, line 88: Default here of empty seems weird as it's defined to do nothing? Would it be better to force at least one item?

DonJayamanne: Oh shoot, thats wrong.

.vscode/launch.json
```diff
@@ -23,8 +23,8 @@
                 "<node_internals>/**"
             ],
             "env": {
-                // Enable this to turn on redux logging during debugging
-                "XVSC_JUPYTER_FORCE_LOGGING": "1",
+                // Disable this to turoff on redux & console logging during debugging
+                "VSC_JUPYTER_FORCE_LOGGING": "1",
```
DonJayamanne, line 27: I use the console log a lot, as it helps me see errors & warnings (colors in console window).
I don't see why turning this on is a problem for debugging, hence leaving as deafult. else I end up turning this on always & not checking this in..

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -0,0 +1,158 @@
...
+     * If cells are cancelled, they are not processed, & that too counts as completion.
+     * If no cells are provided, then wait on all cells in the current queue.
+     */
+    public async waitForCompletion(cells?: NotebookCell[]): Promise<void> {
+        const cellsToCheck =
+            Array.isArray(cells) && cells.length > 0
+                ? this.queueOfCellsToExecute.filter((item) => cells.includes(item.cell))
+                : this.queueOfCellsToExecute;
+
+        await Promise.all(cellsToCheck.map((cell) => cell.result));
```
DonJayamanne, line 95: Fixed @rchiodo let me know if this is better.
Or should i have two methods?
`waitForCompletionOfCell(cell)` & `waitForCompletionOfAllCells()`

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
DonJayamanne: Fixed

rchiodo (approved):

Issue #4541, opened 2021-01-28
Variables tab does not update when running a new cell if it's not visible


rchiodo:
1. Run some cells in a notebook with variables
2. Open variable viewer
3. Switch to another output (hide variable viewer)
4. Add new cell with new variable
5. Switch back to variable viewer

Expected result:
Variable shows up

Actual result:
Variable is not there. Variable list seems to have previous execution list.

IanMatthewHuff:
Hmm, maybe this is the same as you were seeing, but this is actually a bit odder than I thought. I checked the messages coming in, and they are getting fetched with the correct execution order, and coming back with the correct variables. 

To get them to show up, I don't actually even have to execute anything more, just moving my mouse over the control makes them show up. Not quite what I expected.

DonJayamanne:
Validated

Issue #4379, opened 2021-01-13
Investigate use of retainContext for variableView


IanMatthewHuff:

IanMatthewHuff:
We're using retainContext now, so I'm just validating myself.

PR #4564, opened 2021-01-29
Don't retain variable view context, update to new execution count on visibility change

IanMatthewHuff:
For #4541 

Also #4492 and #4379

### DIFF
--- a/src/client/datascience/variablesView/notebookWatcher.ts
+++ b/src/client/datascience/variablesView/notebookWatcher.ts
@@ -32,6 +32,15 @@ export class NotebookWatcher implements INotebookWatcher {
     public get activeNotebook(): INotebook | undefined {
         return this.notebookEditorProvider.activeEditor?.notebook;
     }
+    public get activeNotebookExecutionCount(): number | undefined {
+        if (this.notebookEditorProvider.activeEditor) {
+            const executionCount = this.getExecutionCount(this.notebookEditorProvider.activeEditor.file);
+
+            return executionCount?.executionCount;
+        }
+
+        return undefined;
+    }
 
     private readonly _onDidExecuteActiveNotebook = new EventEmitter<{ executionCount: number }>();
     private readonly _onDidChangeActiveNotebook = new EventEmitter<{
--- a/src/client/datascience/variablesView/variableView.ts
+++ b/src/client/datascience/variablesView/variableView.ts
@@ -139,7 +139,16 @@ export class VariableView extends WebviewViewHost<IVariableViewPanelMapping> imp
         context.set(visible).ignoreErrors();
 
         // Also perform a refresh on a visibily change
-        this.postMessage(InteractiveWindowMessages.ForceVariableRefresh).ignoreErrors();
+        if (visible) {
+            //this.postMessage(InteractiveWindowMessages.ForceVariableRefresh).ignoreErrors();
+            if (this.notebookWatcher.activeNotebookExecutionCount !== undefined) {
+                this.postMessage(InteractiveWindowMessages.UpdateVariableViewExecutionCount, {
+                    executionCount: this.notebookWatcher.activeNotebookExecutionCount
+                }).ignoreErrors();
+            } else {
+                this.postMessage(InteractiveWindowMessages.ForceVariableRefresh).ignoreErrors();
+            }
+        }
     }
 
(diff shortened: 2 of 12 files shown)


src/client/datascience/variablesView/notebookWatcher.ts
```diff
@@ -32,6 +32,15 @@ export class NotebookWatcher implements INotebookWatcher {
     public get activeNotebook(): INotebook | undefined {
         return this.notebookEditorProvider.activeEditor?.notebook;
     }
+    public get activeNotebookExecutionCount(): number | undefined {
```
IanMatthewHuff, line 35: This was a change that we needed along with not retaining context anymore. When we become visible, check if there is a currently active notebook and update the execution count in the variable view. Basically the same operation as if you switch active documents.

src/client/datascience/variablesView/variableView.ts
```diff
@@ -138,8 +138,19 @@ export class VariableView extends WebviewViewHost<IVariableViewPanelMapping> imp
         }
         context.set(visible).ignoreErrors();
 
-        // Also perform a refresh on a visibily change
-        this.postMessage(InteractiveWindowMessages.ForceVariableRefresh).ignoreErrors();
+        // I've we've been made visible, make sure that we are updated
+        if (visible) {
```
IanMatthewHuff, line 142: This was a change that we needed along with not retaining context anymore. When we become visible, check if there is a currently active notebook and update the execution count in the variable view. Basically the same operation as if you switch active documents.

src/client/datascience/variablesView/variableViewActivationService.ts
```diff
@@ -18,9 +18,8 @@ export class VariableViewActivationService implements IExtensionSingleActivation
         // Only activate this when in the NativeNotebook experiment
         if (await this.experimentService.inExperiment(Experiments.NativeNotebook)) {
             this.extensionContext.subscriptions.push(
-                // Consider not using retainContext here? This will save the context of our view, but take more memory in VS Code.
                 window.registerWebviewViewProvider(this.variableViewProvider.viewType, this.variableViewProvider, {
-                    webviewOptions: { retainContextWhenHidden: true }
+                    webviewOptions: { retainContextWhenHidden: false }
```
IanMatthewHuff, line 22: So the bug was a bit more interesting than I thought at first. Turns out I was actually keeping the view and execution count updated. But with retaincontext on the view was updating state when hidden. 

When the view was made visible after executing it refreshed, but after the refresh (since the state was already updating while the view was not visible) it didn't trigger a re-render since there wasn't a state change. You could actually see this if you moused over the view in this state it would re-render and have the correct values.

I already had a bug #4379 to consider not using retain context. This does mean that it might be a tiny bit slowing opening this view for a second / third / forth time, but it's not very noticeable as the variable fetch is the slow part, not the UI initialization.

src/datascience-ui/variable-view/variableViewPanel.tsx
```diff
@@ -28,8 +28,6 @@ export class VariableViewPanel extends React.Component<IVariableViewPanelProps>
         super(props);
 
         this.panelRef = React.createRef<HTMLDivElement>();
-        // For the variable view we want to start toggled open
```
IanMatthewHuff, line 31: Noticed that on view show an executionCount 0 message was getting sent first. Didn't really cause a bug (since it was immediately followed by the correct execution count fetch) but was unneeded and was due to this little hack. The toggle action triggered a fetch. Instead just added a way to specify on initial state if the view should be visible.

src/client/datascience/variablesView/variableViewActivationService.ts
rchiodo: I think you should add a comment here saying why it has to be false. Somebody might change this later and not realize it messes up the execution count.

rchiodo (approved):

src/client/datascience/variablesView/variableViewActivationService.ts
IanMatthewHuff: @rchiodo Will do. Also I think that it's just the default (since vscode doesn't want views to retain context) so maybe I can comment and just remove the options object.

DonJayamanne (approved):

Issue #1147, opened 2020-09-30
Inf, -Inf, and NaNs don't show up in the Data Viewer


jmew:
If i have a data frame with infinity, -infinity, or NaN values, they show up as blank in the data viewer

IanMatthewHuff:
Closing as a dupe of #299 which I'll be validating.

Issue #299, opened 2020-10-26
Support NaN, inf, -inf in data viewer


joyceerhl:
- Today the pandas DataFrame `to_json` doesn't support serializing NaN, inf, -inf. `json.dumps` does support NaN, inf, -inf, but can't be used on DataFrames:
![image](https://user-images.githubusercontent.com/30305945/97235234-b0e57a00-179f-11eb-973d-a2c70619dd64.png)
- Possible solution: preprocess NaN/inf/-inf in vscodeGetDataFrameRows with `df.replace(np.inf, ...)` before calling `df.to_json()`
- Need to map values over to JavaScript equivalents in order for the dataviewer column sort to work

IanMatthewHuff:
Validated. Looks good to me. Added the values to a few lists and checked sorting as well.

PR #4577, opened 2021-01-30
Support NaN, Inf, -Inf in data viewer

joyceerhl:
For #299, https://github.com/microsoft/vscode-jupyter/issues/1147

This requires some hackery because the JSON spec doesn't actually specify how to represent Infinity, -Infinity, and NaN.

Theoretically all that is required is the following: 
1. pd.replace NaN, inf and -inf to unique strings in our Python scripts before dumping DFs to JSON
2. Pass a reviver to JSON which turns the inf, -inf and NaN back into JavaScript equivalent numerical values like so:

```ts
    private deserializeJupyterResult<T>(cells: ICell[]): T {
        const text = this.extractJupyterResultText(cells);
        const reviver = (_k: any, v: any) => {
            if (v === '_VSCODE_infinity') {
                return Infinity;
            } else if (v === '_VSCODE_neg_infinity') {
                return -Infinity;
            } else if (v === '_VSCODE_nan') {
                return NaN;
            }
            return v;
        }
        return JSON.parse(text, reviver) as T;
    }
``` 

Unfortunately it looks like VS Code's webview API's postMessage also converts Infinity, -Infinity and NaN to null, because the message JSON also gets serialized and deserialized without preserving Infinity, -Infinity and NaN. 

Without an upstream fix in VSCode the best we can do is to perform the reviving in the UI itself.


src/datascience-ui/data-explorer/mainPanel.tsx
```diff
@@ -305,16 +310,59 @@ export class MainPanel extends React.Component<IMainPanelProps, IMainPanelState>
...
         // Make sure we have an index field and all rows have an item
-        return rows.map((r: any | undefined) => {
+        const normalizedRows = rows.map((r: any | undefined) => {
             if (!r) {
                 r = {};
             }
             if (!r.hasOwnProperty(this.state.indexColumn)) {
                 r[this.state.indexColumn] = uuid();
             }
+            for (let [key, value] of Object.entries(r)) {
```
joyceerhl, line 326: In practice it should be unusual for inf, nan and -inf to appear in user data. But when it happens it usually indicates something is wrong in the training process, so we need to perform this check for every cell in order to accurately flag their presence.

One heuristic we could use is to only do this check if the first column type is Number, or turn on this check only for Tensors. Not sure if strictly necessary though.

src/datascience-ui/data-explorer/mainPanel.tsx
```diff
@@ -304,17 +309,60 @@ export class MainPanel extends React.Component<IMainPanelProps, IMainPanelState>
         return [];
     }
 
-    private normalizeRows(rows: JSONArray): ISlickRow[] {
+    private normalizeData(rows: JSONArray): ISlickRow[] {
+        // While processing rows we may encounter Inf, -Inf or NaN.
+        // These rows' column types will initially be 'string' or 'object' so
+        // make sure we update the column types
+        // Set of columns to update based on this batch of rows
+        const columnsToUpdate = new Set<string>();
```
joyceerhl, line 317: `normalizeRows` already loops through the row data, so I implemented checking cell data for inf/-inf/nan inside the loop. I thought about whether to pull the column checks out into a separate function but didn't really want to because we'd either need to maintain some global state and check it after this function returns, or return the set of columns to update from this function. So ended up renaming this function to `normalizeData` as that's what it's doing. But I'm open to feedback on whether this is confusing.

IanMatthewHuff: Outside of strict statistics 'normalize' is pretty overloaded. To me at least this type of operation doesn't seem too out of place to be happening here..

src/datascience-ui/data-explorer/mainPanel.tsx
```diff
@@ -304,17 +309,60 @@ export class MainPanel extends React.Component<IMainPanelProps, IMainPanelState>
...
+                            this.columnsContainingInfOrNaN.add(key);
+                        }
+                        break;
+                    default:
+                }
+            }
             return r;
         });
+        // Need to update the column types so that that column gets number treatment.
+        // This should be unusual in practice.
```
IanMatthewHuff, line 355: Considering if this is actually something that would happen or not, but in a rare case could this be converting something like a string column to a number column? If it happened to have a string value of 'nan' or 'inf' in it? Not sure if that possibility is worth guarding against. You know the data best here Joyce, possible issue? Or not something worth worrying about.

src/datascience-ui/data-explorer/reactSlickGrid.tsx
```diff
@@ -458,6 +474,11 @@ export class ReactSlickGrid extends React.Component<ISlickGridProps, ISlickGridS
         return null;
     }
 
+    private updateColumns = (_e: Slick.EventData, newColumns: Slick.Column<Slick.SlickData>[]) => {
+        this.state.grid?.setColumns(newColumns);
+        this.state.grid?.render(); // We might be able to skip this rerender?
```
IanMatthewHuff, line 479: Looks like an open comment here. Have you tried it without?

rchiodo (approved):

src/datascience-ui/data-explorer/reactSlickGrid.tsx
joyceerhl: Yes and it worked, but the API documentation says

> Column headers will be recreated and all rendered rows will be removed. To re-render the grid (if necessary), call render().

So not really sure what to make of that.

src/datascience-ui/data-explorer/mainPanel.tsx
joyceerhl: It's possible because data can be heterogenous. Our code today assumes ColumnType applies to all the data in the column. I don't really know what the best way to handle this is. I originally replaced `np.inf` with `_VSCODE_inf` and so on in our Python scripts for that exact reason, and that worked in some cases (e.g. ndarray, tensor) but not others (e.g. DataFrame).

One option is to enable this check only for tensors, since it's especially critical to detect inf/-inf/nan gradients. In other scenarios it doesn't matter as much. But I know jmew wants this fixed globally so that wouldn't work.

DonJayamanne (approved):

IanMatthewHuff (approved):

Issue #4540, opened 2021-01-28
Viewing data from a conda kernel when non conda is the active interpreter does not work


rchiodo:
1. Create (or install) python venv
1. Set this as your python interpreter (open a python file if you have to)
1. Create conda env
1. Open notebook
1. Set its kernel to the conda env
1. Add something that you can view with the data viewer
1. Try to view it
1. It will ask to install pandas
1. Pandas install will fail

rchiodo:
You can work around the problem if you set the active interpreter to the conda environment.

DonJayamanne:
IPyKernel install into conda doesn't work.

DonJayamanne:
Found the problem is due to a bug upstream in Python extension - https://github.com/microsoft/vscode-python/pull/15286

DonJayamanne:
Install latest dev version of Python extension.

rchiodo:
Validated.

PR #4583, opened 2021-02-01
Improved handling of  failures in Kernell Daemon

DonJayamanne:
Related to #4540

The kernel dies pretty much immediately but we need to wait for 3 minutes for a timeout.
The problem is we're erroring out in the Rx Subject, & the assumption was that Rx Subjects can be errored multiple times.

Here's what it looks like now (no timeout error & better error message)
Screen Shot 2021-02-01 at 10 42 48

### DIFF
--- a/src/client/datascience/kernel-launcher/kernelProcess.ts
+++ b/src/client/datascience/kernel-launcher/kernelProcess.ts
@@ -9,6 +9,7 @@ import * as tmp from 'tmp';
 import { CancellationToken, CancellationTokenSource, Event, EventEmitter } from 'vscode';
 import { IPythonExtensionChecker } from '../../api/types';
 import { createPromiseFromCancellation, wrapCancellationTokens } from '../../common/cancellation';
+import { getErrorMessageFromPythonTraceback } from '../../common/errors/errorUtils';
 import { traceDecorators, traceError, traceInfo, traceWarning } from '../../common/logger';
 import { IFileSystem } from '../../common/platform/types';
 import { IProcessServiceFactory, ObservableExecutionResult } from '../../common/process/types';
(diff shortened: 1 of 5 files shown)


src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -130,18 +135,30 @@ export class KernelProcess implements IKernelProcess {
                 }
                 traceError('Kernel died', error, stderr);
                 if (error instanceof PythonKernelDiedError) {
+                    providedExitCode = error.exitCode;
```
DonJayamanne, line 138: This `error` was getting hit, but we didn't get `PythonKernelDiedError` as this callback is only invoked once.
The PR (change) now ensures this will get called only once with `PythonKernelDiedError`

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -162,12 +179,11 @@ export class KernelProcess implements IKernelProcess {
...
-                    localize.DataScience.kernelDied().format(
-                        Commands.ViewJupyterOutput,
-                        (stderrProc || stderr).substring(0, 100)
-                    )
-                );
+                // If we have the python error message, display that.
+                const errorMessage =
+                    getErrorMessageFromPythonTraceback(stderrProc || stderr) ||
+                    (stderrProc || stderr).substring(0, 100);
+                throw new Error(localize.DataScience.kernelDied().format(Commands.ViewJupyterOutput, errorMessage));
```
DonJayamanne, line 186: Better display of error messages, since we have logged the Python error in console, we can display the meaningful Python error message from the traceback.
(previously this used to be empty as error message is sent as part of other messages in the rx subject).

rchiodo (approved):

Issue #4412, opened 2021-01-15
Fix xvfb Github action to support the parameters we need and only run for linux testing


IanMatthewHuff:
Reenable variable view test so that it can run in CI when I fix this.

PR #4584, opened 2021-02-01
Final Variable Test work

IanMatthewHuff:
For #4412 


.github/workflows/main.yml
```diff
@@ -634,6 +634,11 @@ jobs:
       - name: Install dependencies (npm ci)
         run: npm ci --prefer-offline
 
+        # Check to see if we should build webviews or not
+      - name: Check if we need to build webviews
```
IanMatthewHuff, line 638: I think this makes what is happening a bit more obvious. I went with a not if check as the default for most of the notebook checks is to skip webview build. Also previously only the 'notebook' step was skipping this. Now the other steps that don't need this (remoteNotebook, nonRawNotebook...) will also skip it for the perf gain.

.github/workflows/main.yml
```diff
@@ -672,12 +677,20 @@ jobs:
...
       - name: Create Virtual Env for Tests
         uses: ./.github/actions/create-venv-for-tests
-        if: matrix.test-suite == 'integration' || matrix.test-suite == 'notebook' || matrix.test-suite == 'nonRawNotebook' || matrix.test-suite == 'remoteNotebook'
+        if: matrix.test-suite == 'integration' || matrix.test-suite == 'notebook' || matrix.test-suite == 'nonRawNotebook' || matrix.test-suite == 'remoteNotebook' || matrix.test-suite == 'notebookAndWebview'
+
+      # Set the correct xvfb commands to run vscode tests
+      # https://code.visualstudio.com/api/working-with-extensions/continuous-integration
+      # Note that xvfb github action only runs through xvfb on linux, so only set the commands there
+      # as on windows / mac they would get directly passed to run, not xvfb-run
+      - name: Set xvfb parameters linux
```
IanMatthewHuff, line 686: xvfb doesn't support passing in server parameters (there is actually an open PR on the github to support this). To get around this we now just pass in the xvfb parameters to the run command, but only on linux, because on windows / mac xvfb just directly forwards on the run command. .xvfbCommand evaluates to empty string in those cases. Credit to @rchiodo for doing this instead of having to create a new action.

.github/workflows/main.yml
```diff
@@ -745,7 +785,6 @@ jobs:
           name: ${{runner.os}}-${{env.COVERAGE_REPORTS}}
           path: .nyc_output
           retention-days: 1
-          if: always()
```
IanMatthewHuff, line 748: if is not valid on the with statement, it's already on this job. This is creating warnings in the test runs for invalid syntax.

IanMatthewHuff:
Per the results of the current run, the new notebookAndWebview tests are 6 min to run. So as long as we don't put too many tests in here, having a step with the webview build included won't be slowing down our PRs.

joyceerhl (approved):

rchiodo (approved):

DavidKutu (approved):

PR #4603, opened 2021-02-02
WIP - Controlled rollout of native for new users

DonJayamanne:


src/client/common/experiments/service.ts
```diff
@@ -125,34 +130,65 @@ export class ExperimentService implements IExperimentService {
...
+    private async isNewUserInNativeNotebookExperiment() {
+        // Only new users from stable can be in experiment.
+        if (this.appEnvironment.channel === 'insiders') {
+            return false;
+        }
+        if (!this.extensionUsage.isFirstTimeUser) {
+            return false;
+        }
+        // If this user was already set to use native notebook, then keep using native notebooks.
+        if (this.globalState.get('IS_IN_NATIVE_NOTEBOOK_NEW_USER_EXP', false)) {
```
rchiodo, line 163: These strings should probably be constants.

rchiodo:
Can you add a description as to why you're doing this? I can kinda surmise that from the code itself, but it would be better if you updated the description of the PR.

Issue #4550, opened 2021-01-28
Ensure that all Azure ML compute users are opted into native notebooks


greazer:

rchiodo:
How do we validate this? Do we need to remote into an azure compute instance?

rchiodo:
@SiddhanthUnnithan do you have directions on how to use the remote support?

DonJayamanne:
> How do we validate this? Do we need to remote into an azure compute instance?

Yes, sorry i missed this.
Setup AML Compute.
Open in VS Code stable, upload the VSIX to the AML Compute & install that VSIX using the command palette & reload VS Code.

rchiodo:
I'm hitting this, but it seems like it at least loads the native editor

```
Error: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.22' not found
```

And for some reason jupyter notebook doesn't work.

PR #4608, opened 2021-02-02
AML Compute users to get Native Notebooks

DonJayamanne:
For #4550

### DIFF
--- a/src/client/common/amlContext.ts
+++ b/src/client/common/amlContext.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import { inject, named } from 'inversify';
+import { inject, injectable, named } from 'inversify';
 import { Memento } from 'vscode';
 import { IExtensionSingleActivationService } from '../activation/types';
 import { setSharedProperty } from '../telemetry';
@@ -11,6 +11,7 @@ import { noop } from './utils/misc';
 
 const amlComputeMementoKey = 'JVSC_IS_AML_COMPUTE_INSTANCE';
 
+@injectable()
 export class AmlComputeContext implements IExtensionSingleActivationService {
     constructor(
         @inject(IMemento) @named(GLOBAL_MEMENTO) private readonly memento: Memento,
--- a/src/client/common/amlContext.ts
+++ b/src/client/common/amlContext.ts
@@ -1,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { inject, named } from 'inversify';
+import { Memento } from 'vscode';
+import { IExtensionSingleActivationService } from '../activation/types';
+import { setSharedProperty } from '../telemetry';
+import { IWorkspaceService } from './application/types';
+import { GLOBAL_MEMENTO, IMemento } from './types';
+import { noop } from './utils/misc';
+
+const amlComputeMementoKey = 'JVSC_IS_AML_COMPUTE_INSTANCE';
+
+export class AmlComputeContext implements IExtensionSingleActivationService {
+    constructor(
+        @inject(IMemento) @named(GLOBAL_MEMENTO) private readonly memento: Memento,
+        @inject(IWorkspaceService) private readonly workspace: IWorkspaceService
+    ) {}
+
+    public get isAmlCompute() {
+        return this.memento.get<boolean>(amlComputeMementoKey, false) || this.isAmlComputeWorkspace();
+    }
+    private isAmlComputeWorkspace() {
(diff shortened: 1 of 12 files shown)


src/client/common/amlContext.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    public get isAmlCompute() {
+        return this.memento.get<boolean>(amlComputeMementoKey, false) || this.isAmlComputeWorkspace();
+    }
+    private isAmlComputeWorkspace() {
+        let isRemoteConnection: boolean = false;
+        // If there is at least 1 folder, we can check if the scheme and authority match a remote connection
+        if (Array.isArray(this.workspace.workspaceFolders) && this.workspace.workspaceFolders.length > 0) {
+            // We are in a remote connection to a CI if the scheme is vscode-remote and the authority contains amlext
+            isRemoteConnection = this.workspace.workspaceFolders.some(
```
DonJayamanne, line 28: Logic courtesy of Shantnu (from AML), @SiddhanthUnnithan

src/client/telemetry/index.ts
```diff
@@ -330,6 +330,11 @@ export interface ISharedPropertyMapping {
      * For every DS telemetry we would like to know the type of Notebook Editor used when doing something.
      */
     ['ds_notebookeditor']: undefined | 'old' | 'custom' | 'native';
+    /**
+     * For every DS telemetry we would like to know whether the this is from AML compute or not.
+     * If not in AML compute, then do not send this telemetry.
+     */
+    ['isamlcompute']: 'yes';
```
DonJayamanne, line 337: So we can identify this telemetry & exclude from others (as these forced optins into NativeNotebooks could mess up with experiments)

src/client/datascience/variablesView/variableViewActivationService.ts
```diff
@@ -11,12 +11,12 @@ export class VariableViewActivationService implements IExtensionSingleActivation
     constructor(
         @inject(IExtensionContext) private extensionContext: IExtensionContext,
         @inject(IVariableViewProvider) private variableViewProvider: IVariableViewProvider,
-        @inject(IExperimentService) private experimentService: IExperimentService
+        @inject(UseVSCodeNotebookEditorApi) private useVSCodeNotebookEditorApi: boolean
```
DonJayamanne, line 14: Changes to always use `UseVSCodeNotebookEditorAPI`, to be consistent.

src/client/extensionActivation.ts
```diff
@@ -86,9 +87,11 @@ async function activateLegacy(
     // Load the two data science experiments that we need to register types
     // Await here to keep the register method sync
     const experimentService = serviceContainer.get<IExperimentService>(IExperimentService);
+    const amlCompute = serviceContainer.get<AmlComputeContext>(AmlComputeContext);
     experimentService.logExperiments();
+
     let useVSCodeNotebookAPI =
-        env.appName.includes('Insider') || (await experimentService.inExperiment(Experiments.NativeNotebook));
+        amlCompute.isAmlCompute || (await experimentService.inExperiment(Experiments.NativeNotebook));
```
rchiodo, line 94: I think we still need the 'in insiders' check, right? Or is that handled elsewhere now?

DonJayamanne: That's handled in experiment class (was always there, hence removed that check from this place)

src/client/common/amlContext.ts
rchiodo: Not super happy with this one off. Feels like we should have the AML extension do this but I don't have a good idea for how we might let them make this decision. 

Opening up our experiment API seems kinda like overkill.

IanMatthewHuff: It does seem a little bit fragile, but aren't we hoping that this all goes away if we can push native notebooks to full usage in a few months?

src/client/telemetry/index.ts
IanMatthewHuff: So this part sounds like an unclean mess to me to have to deal with in telemetry. But as long as PMs have signed off on this I guess ok?

IanMatthewHuff: Like, can't someone be an AML compute user and then also get put in the experiment? I guess we can code around in queries, but feels nasty.

IanMatthewHuff (approved):
I'll confess, I don't love it. The code seems fine, just the design of forcing some in and out outside of the experiment seems messy. But if that's the choice then we do it.

src/client/telemetry/index.ts
DonJayamanne: Agreed, and this is why they initially didn't want anyone to be opted into the experiment, but this is required & PMs are ok/happy with this.

rchiodo (approved):

PR #4610, opened 2021-02-02
Remove usage of jupyter to enumerate local kernels

rchiodo:
Part of the refactor work for removing kernel installation dependencies, we don't need two ways to find kernelspecs.

So I changed everything to use the kernelFinder (raw version of finding) and removed a bunch of code.


IanMatthewHuff:
Wasn't this waiting until after we cut the release branch?

rchiodo:
> Wasn't this waiting until after we cut the release branch?

Maybe? I didn't think this as risky. It's just reusing something we already have in a different spot. No change otherwise

DavidKutu (approved):

DonJayamanne (approved):
Always nice to see code getting deleted....
Woo hoo..

DonJayamanne:
> Maybe? I didn't think this as risky. It's just reusing something we already have in a different spot. No change otherwise

Provided we test this thoroughly, I'm ok with this going in, we have the bug bash.
Also I think we do have a lot of tests now & feel fairly confident that if tests pass this should be ok (Would like to ensure we test Conda scnearios well dueing bug bash).

IanMatthewHuff (approved):

PR #4623, opened 2021-02-03
Remove checking of vscode channel for Notebooks

DonJayamanne:


src/client/datascience/notebook/integration.ts
```diff
@@ -110,11 +107,6 @@ export class NotebookIntegration implements IExtensionSingleActivationService {
         this.disposables.push(disposable);
     }
     private async enableNotebooks() {
-        if (this.env.channel === 'stable') {
```
DonJayamanne, line 113: In correct check, removed.

src/client/datascience/insidersNativeNotebookSurveyBanner.ts
```diff
@@ -127,8 +127,8 @@ export class InsidersNativeNotebooksSurveyBanner implements IExtensionSingleActi
         this.browserService.launch('https://aka.ms/vscjupyternb');
     }
 
-    private async isInsidersNativeNotebooksUser() {
-        return this.useVSCodeNotebookEditorApi;
+    private isInsidersNativeNotebooksUser() {
+        return this.useVSCodeNotebookEditorApi && this.applicationEnvironment.channel === 'insiders';
```
DonJayamanne, line 131: Fixed the assumpgion, the prompt is required only for Insider users. Previously the experiment was enabled only in insiders, hence that condition worked, but not anymore (not if native is available for Stable users as well).

src/client/datascience/insidersNativeNotebookSurveyBanner.ts
```diff
@@ -127,8 +127,8 @@ export class InsidersNativeNotebooksSurveyBanner implements IExtensionSingleActi
         this.browserService.launch('https://aka.ms/vscjupyternb');
     }
 
-    private async isInsidersNativeNotebooksUser() {
-        return this.useVSCodeNotebookEditorApi;
+    private isInsidersNativeNotebooksUser() {
```
joyceerhl, line 130: Can we rename this file and method to `nativeNotebooksSurveyBanner` and `nativeNotebooksUser` respectively?

DonJayamanne: Nope, this banner is still applicable only for VS Code Insiders, & nothing else.

DonJayamanne: I.e file name & class name is still correct.

src/client/datascience/insidersNativeNotebookSurveyBanner.ts
joyceerhl: Do we need to update dataScienceSurveyBanner as well to ensure we don't show prompts if the user has hit the threshold for usage of both old and native notebooks? This could happen if the user was on old notebooks before, hit the threshold, but ignored our prompts (in which case the agreed behavior was to keep showing them until they click either 'Yes' or 'No'), then got opted into native notebooks and also hit the threshold there.

I think you mentioned that the plan is to only show native notebooks to new users, but if I misremembered, this scenario could occur.

DonJayamanne: > plan is to only show native notebooks to new users

Nope, native notebooks will be for both existing & new users.

IanMatthewHuff (approved):

joyceerhl (approved):

rchiodo (approved):

DavidKutu (approved):

Issue #4632, opened 2021-02-03
Order of Icons in Native Notebook toolbar is odd


DonJayamanne:
![Screen Shot 2021-02-03 at 13 07 20](https://user-images.githubusercontent.com/1948812/106809742-e44a6700-6620-11eb-8304-cea0f0d733fc.png)

**What is the expected order of icons**
1. Run All (added by vs code, when running all, this changes to `Stop`)
![Screen Shot 2021-02-03 at 13 13 28](https://user-images.githubusercontent.com/1948812/106810397-b9acde00-6621-11eb-9d84-97c1d5007977.png)

2. Clear output (added by VS Code)
![Screen Shot 2021-02-03 at 13 13 31](https://user-images.githubusercontent.com/1948812/106810396-b9144780-6621-11eb-8dd7-bae9b61e5c00.png)

3. Interrupt Kernel (added by us)
	Enabled only when cells are running (kernel is busy)
![Screen Shot 2021-02-03 at 13 13 33](https://user-images.githubusercontent.com/1948812/106810393-b9144780-6621-11eb-96d2-cb7fac315ea0.png)
4. Open Changes (added by VS Code - no idea why this is after our icon?)
![Screen Shot 2021-02-03 at 13 13 36](https://user-images.githubusercontent.com/1948812/106810391-b9144780-6621-11eb-85b8-d3ae7e2b4eaf.png)
5. Restart Kernel
	Enabled only when cells are running (kernel is busy)
![Screen Shot 2021-02-03 at 13 13 39](https://user-images.githubusercontent.com/1948812/106810390-b9144780-6621-11eb-9975-993844e5392d.png)
6. Run Cell Above
	Enabled only when any cell other than 1st is selected
![Screen Shot 2021-02-03 at 13 13 42](https://user-images.githubusercontent.com/1948812/106810389-b87bb100-6621-11eb-96f6-5581f256c5d5.png)
7. Run cell & below
![Screen Shot 2021-02-03 at 13 13 44](https://user-images.githubusercontent.com/1948812/106810388-b87bb100-6621-11eb-89d0-1d7ea099e381.png)
8. Variable viewer
![Screen Shot 2021-02-03 at 13 13 47](https://user-images.githubusercontent.com/1948812/106810387-b87bb100-6621-11eb-98a8-8897a95fd72c.png)
9. Export Notebook
[...]

PR #4655, opened 2021-02-04
Correctly hide old interpreters registered as kernelspecs from kernel selection

IanMatthewHuff:
For #4632

### DIFF
--- a/src/client/datascience/jupyter/kernels/providers/installedLocalKernelProvider.ts
+++ b/src/client/datascience/jupyter/kernels/providers/installedLocalKernelProvider.ts
@@ -39,11 +39,8 @@ export class InstalledLocalKernelSelectionListProvider
                         // If it is, then no need to display that (selecting kernels registered is done by selecting the corresponding interpreter).
                         // Hence we can hide such kernels.
                         // Kernels we create will end with a uuid (with - stripped), & will have interpreter info in the metadata.
-                        if (
-                            item.metadata?.interpreter &&
-                            item.name.length > 32 &&
-                            item.name.slice(-32).toLowerCase() === item.name
-                        ) {
+                        const guidRegEx = /[a-f0-9]{32}$/;
+                        if (item.metadata?.interpreter && item.name.toLowerCase().search(guidRegEx)) {
                             return false;
                         }
 
--- a/src/client/datascience/jupyter/kernels/providers/installedLocalKernelProvider.ts
+++ b/src/client/datascience/jupyter/kernels/providers/installedLocalKernelProvider.ts
@@ -28,7 +28,7 @@ export class InstalledLocalKernelSelectionListProvider
         resource: Resource,
         cancelToken?: CancellationToken
     ): Promise<IKernelSpecQuickPickItem<KernelSpecConnectionMetadata>[]> {
-        const rawNotebookSupported = this.rawNotebookSupportedService.isSupportedForLocalLaunch();
+        const rawNotebookSupported = await this.rawNotebookSupportedService.supported();
         const items = await this.kernelFinder.listKernelSpecs(resource);
         const selections = await Promise.all(
             items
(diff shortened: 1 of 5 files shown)


IanMatthewHuff:
This is to fix the issue that I showed at the meeting with my covid env not showing up. It does require that you have a kernelspec on your system that we registered from an interpreter in the old days (which would be true of many customers). Was bit hard to track down as the actually issue doesn't show up until much later in the UI.

Basically:
1. Old kernelspec registered from an interpreter is on the system
2. Kernelspec finder finds it
3. Existing check to see if it was our old registered interpreter does not correctly detect that it's one that we registered
4. interpreterPath is undefined on the kernelspec, interpreterPath is filled out with current interpreter in the kernelSelection code
5. When creating the final kernel selector list, the current environment path matches the same python path that got stuck into the interpreter path of the existing old kernelspec (in my case my old conda base kernelspec ended up with interpreterPath = covidEnv/python) thus the current environment path is flagged as a duplicate of the old kernelspec and is removed from the selection list.

src/client/datascience/jupyter/kernels/providers/installedLocalKernelProvider.ts
```diff
@@ -39,11 +39,8 @@ export class InstalledLocalKernelSelectionListProvider
                         // If it is, then no need to display that (selecting kernels registered is done by selecting the corresponding interpreter).
                         // Hence we can hide such kernels.
                         // Kernels we create will end with a uuid (with - stripped), & will have interpreter info in the metadata.
-                        if (
-                            item.metadata?.interpreter &&
-                            item.name.length > 32 &&
-                            item.name.slice(-32).toLowerCase() === item.name
-                        ) {
+                        const guidRegEx = /[a-f0-9]{32}$/;
+                        if (item.metadata?.interpreter && item.name.toLowerCase().search(guidRegEx)) {
```
IanMatthewHuff, line 43: I'm not fully sure what the old check was trying here, but I don't think that it would have ever worked as a substring slice would never be equal to the full string. My one slight question here is if we still fill out the interpreter metadata in kernelspecs or if that was just in the old code. If it's not just in the old code this does mean that if you create a kernelspec with a name that ends in a guid we would exclude it, but I believe that we only fill out the interpreter metadata in the old code paths.

DavidKutu (approved):

DonJayamanne (approved):

src/client/datascience/jupyter/kernels/providers/installedLocalKernelProvider.ts
rchiodo: With ZMQ off I found that these kernels are still created. They're created for every interpreter that you try to use. So it would be weird if they didn't show up?

rchiodo: The old code is obviously wrong but not sure what the side effects are.

IanMatthewHuff: @rchiodo Good call, it's making the code work the right way. Lemme verify quick.

IanMatthewHuff: @rchiodo I was trying to trace all the paths here, then I realized that just doing this might be cleaner. Only hide the old interpreters in the raw scenarios. Keeps the old paths the same for when they register interpreters. Does introduce a new concept into the kernel provider code, but that doesn't seem too bad as that's the whole point of having these small services.

IanMatthewHuff: @rchiodo were you ok with the change here? Didn't want to commit otherwise.

rchiodo: Yeah just wanted to make sure you tested it.

rchiodo (approved):

src/client/datascience/jupyter/kernels/providers/installedLocalKernelProvider.ts
IanMatthewHuff: Gave it a whirl. Looks like I see the expected specs / interpreters on there with ZMQ off. However I am seeing some issues there with kernel switching. If I repro those same issues without my change here then I'll just check in and file that as a new issue.

joyceerhl (approved):

Issue #4598, opened 2021-02-02
Horizontal scroll bar is not visible on variables view when docked in the side bar


rchiodo:
![image](https://user-images.githubusercontent.com/19672699/106538055-6ebf8900-64b0-11eb-8db9-9bad9a05aac7.png)

You can kind of see the bar at the bottom but I can't click on it.

IanMatthewHuff:
Validate. Scrollbar looks good now.

PR #4659, opened 2021-02-04
Fix the Variable Explorer height so the horizontal scroll bar is shown.

DavidKutu:
For #4598

scroll

### DIFF
--- a/news/2 Fixes/4598.md
+++ b/news/2 Fixes/4598.md
@@ -1,0 +1,1 @@
+Fix the Variable Explorer height so the horizontal scroll bar is shown.
--- a/src/datascience-ui/interactive-common/variableExplorer.tsx
+++ b/src/datascience-ui/interactive-common/variableExplorer.tsx
@@ -357,7 +357,9 @@ export class VariableExplorer extends React.Component<IVariableExplorerProps, IV
             return baseHeight;
         }
 
-        return baseHeight - variableExplorerMenuBar.clientHeight;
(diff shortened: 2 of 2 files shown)


rchiodo (approved):
Did you try in the interactive window too? Just want to make sure there isn't some extra space there when there wasn't before.

IanMatthewHuff:
> Did you try in the interactive window too? Just want to make sure there isn't some extra space there when there wasn't before.

Yeah, the variableExplorer is the part that is shared between both view styles, so would need to check that as well. If the variableExplorer is now looking worse in the interactiveWindow you can consider taking a peek at setVariableViewHeight in the VariableViewPanel. That panel is the variable view specific host, if it's passing in a bit too big of a value to setVariableViewHeight, then padding that would fix sizing for just the variable view case and leave the interactive window spacing as is.

IanMatthewHuff (changes requested):
I think this works, but could you try it quick with just reducing the value sent in via the height change in variableViewPanel? That would make the change specific to the new variable view.

DavidKutu:
@IanMatthewHuff it works, but weirdly, only half of the scroll bar appears
image

DavidKutu:
The interactive window looks the same in both solutions:

- -10 in variableViewPanel
variableViewPanel

- -10 in VariableExplorer
variableExplorer

IanMatthewHuff (approved):

IanMatthewHuff:
@DavidKutu Thanks for checking. Looks good to me.

PR #4661, opened 2021-02-04
Fix handling of 1D/2D ndarrays and tensors

joyceerhl:
Was testing and noticed #298 regressed ndarray handling--all elements were automatically stringified, so filtering stopped working. Same goes for tensors.

### DIFF
--- a/pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
+++ b/pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
@@ -18,16 +18,24 @@ def _VSCODE_convertNumpyArrayToDataFrame(ndarray):
     try:
         x_len = temp.shape[0]
         # Figure out if we're dealing with ragged data
-        # Handle ragged arrays by making a container where the number of
-        # columns is the max length of all rows. Missing elements are
-        # represented as empty strings.
         row_lengths = [
             len(temp[i])
             if isinstance(temp[i], (list, tuple, _VSCODE_np.ndarray))
             else 1
             for i in range(x_len)
         ]
-        flattened = _VSCODE_np.full((x_len, max(row_lengths)), "", dtype="object")
+        max_y_len = max(row_lengths)
+        is_ragged_or_ndimensional = hasattr(temp, "ndim") and (
+            (temp.ndim == 1 and max_y_len > 1) or temp.ndim > 2
+        )
+
+        if not is_ragged_or_ndimensional:
+            return _VSCODE_pd.DataFrame(temp)
+
+        # Handle ragged arrays by making a container where the number of
+        # columns is the max length of all rows. Missing elements are
+        # represented as empty strings.
+        flattened = _VSCODE_np.full((x_len, max_y_len), "", dtype="object")
 
         def _VSCODE_stringifyElement(element):
             if isinstance(element, _VSCODE_np.ndarray):
@@ -50,15 +58,17 @@ def _VSCODE_convertNumpyArrayToDataFrame(ndarray):
                 cell = row if row_length == 1 else row[j]
                 flattened[i][j] = _VSCODE_stringifyElement(cell)
 
-        temp = flattened
+        ndarray = flattened
         del flattened
+        del is_ragged_or_ndimensional
+        del x_len
+        del max_y_len
+        del row_lengths
+        del temp
+        return _VSCODE_pd.DataFrame(ndarray)
     finally:
         # Restore the user's printoptions
         _VSCODE_np.set_printoptions(threshold=current_options["threshold"])
-        temp = _VSCODE_pd.DataFrame(temp)
-        ndarray = temp
-        del temp
-        return ndarray
 
 
 # Function that converts tensors to DataFrames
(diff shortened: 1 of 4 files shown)


pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -18,16 +18,24 @@ def _VSCODE_convertNumpyArrayToDataFrame(ndarray):
...
             for i in range(x_len)
         ]
-        flattened = _VSCODE_np.full((x_len, max(row_lengths)), "", dtype="object")
+        max_y_len = max(row_lengths)
+        is_ragged_or_ndimensional = hasattr(temp, "ndim") and (
+            (temp.ndim == 1 and max_y_len > 1) or temp.ndim > 2
+        )
+
+        if not is_ragged_or_ndimensional:
+            return _VSCODE_pd.DataFrame(temp)
```
joyceerhl, line 33: This is the key change--do a direct conversion instead of filling out a new ndarray of type object if the data is not ragged or ndimensional.

DavidKutu (approved):

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -18,16 +18,24 @@ def _VSCODE_convertNumpyArrayToDataFrame(ndarray):
...
-        # represented as empty strings.
         row_lengths = [
             len(temp[i])
             if isinstance(temp[i], (list, tuple, _VSCODE_np.ndarray))
             else 1
             for i in range(x_len)
         ]
-        flattened = _VSCODE_np.full((x_len, max(row_lengths)), "", dtype="object")
+        max_y_len = max(row_lengths)
+        is_ragged_or_ndimensional = hasattr(temp, "ndim") and (
```
IanMatthewHuff, line 28: Ok, this is probably just explaining python to a python noob. But is there a missing case here in the case where ndim == 2, but it also has a ragged type?

IanMatthewHuff: Looking on the ndarray docs now

joyceerhl: I was originally thinking that that case doesn't matter because the elements along the 3rd dimension get stringified, which is what we would've done anyway in the special handling. But FWIW the stringification for those nested array elements looks like this:

image

Maybe I shouldn't be handling 1D ragged ndarrays specially, i.e. for this data:
```python
bar = np.array([[1, 2, 3, 4], [5, 6, 7]])
```

instead of presenting them nicely like this (current special-cased behavior):
image

we would instead present them like this: 
image

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -18,16 +18,24 @@ def _VSCODE_convertNumpyArrayToDataFrame(ndarray):
...
+            (temp.ndim == 1 and max_y_len > 1) or temp.ndim > 2
+        )
+
+        if not is_ragged_or_ndimensional:
+            return _VSCODE_pd.DataFrame(temp)
+
+        # Handle ragged arrays by making a container where the number of
+        # columns is the max length of all rows. Missing elements are
+        # represented as empty strings.
+        flattened = _VSCODE_np.full((x_len, max_y_len), "", dtype="object")
```
rchiodo, line 38: I think we prefix all variables with _VSCODE_ so we can make sure we don't delete existing variables?

rchiodo: For example if in the kernel I had a variable named flattened, doesn't this delete it/change it?

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
IanMatthewHuff: This is honestly beyond my user knowledge on what a data scientist would want. Personally I think your current non-special case scenario looks like what I would want. I see that as a 1D nparray which contains two things that happen to be lists. But I do not know if the general community would agree.

IanMatthewHuff: Is there prior art here from spyder / pycharm / r studio?

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
joyceerhl: Based on https://docs.python.org/3/faq/programming.html#what-are-the-rules-for-local-and-global-variables-in-python, 

> If a variable is assigned a value anywhere within the function’s body, it’s assumed to be a local unless explicitly declared as global.

For example,

image

That being said, I'm no Python expert...

rchiodo: TIL I learned there locally scoped variables in python. I thought they were all global

IanMatthewHuff (approved):

DonJayamanne (approved):

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
joyceerhl: Serves me right for trying to be too clever with handling ragged arrays 😁 thanks for bringing this up Ian. Much happier with the code as well, and this way we're more consistent with other DS tooling as well.

PR #4662, opened 2021-02-04
Fixes to validation of remote kernels

DonJayamanne:
Found an issue while testing, didn't cause any issues but found it was incorrect.
If we're using Remote Connections, then the value should ALWAYS be `jupyter`.

### DIFF
--- a/src/client/datascience/jupyter/kernels/helpers.ts
+++ b/src/client/datascience/jupyter/kernels/helpers.ts
@@ -12,7 +12,7 @@ import { nbformat } from '@jupyterlab/coreutils';
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 import cloneDeep = require('lodash/cloneDeep');
 import { PYTHON_LANGUAGE } from '../../../common/constants';
-import { ReadWrite } from '../../../common/types';
+import { IConfigurationService, ReadWrite } from '../../../common/types';
 import { PythonEnvironment } from '../../../pythonEnvironments/info';
 import {
     DefaultKernelConnectionMetadata,
@@ -22,6 +22,7 @@ import {
     LiveKernelModel,
     PythonKernelConnectionMetadata
 } from './types';
+import { Settings } from '../../constants';
 
 // Helper functions for dealing with kernels and kernelspecs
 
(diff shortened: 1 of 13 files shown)


src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -266,3 +267,14 @@ export function cleanEnvironment<T>(spec: T): T {
 
     return copy as T;
 }
+
+export function isLocalLaunch(configuration: IConfigurationService) {
```
DonJayamanne, line 271: Ideally this should be a value should be a constant, something that needs to be injected into our DI container like the following
```typescript
serviceManager.addSingletonInstance<boolean>(IsLocalLaunch, true or false);
```

As this cannot change in the same session of VS Code.& if the user changes it they must reload VS Code to get things working.
This way, we don't need to keep checking values all over the place for `isLocalLaunch` etc..
Its not a mandatory change & its a cleanup that we can do later.

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -266,3 +267,14 @@ export function cleanEnvironment<T>(spec: T): T {
...
+    const settings = configuration.getSettings(undefined);
+    const serverType: string | undefined = settings.jupyterServerType;
+
+    if (!serverType || serverType.toLowerCase() === Settings.JupyterServerLocalLaunch) {
```
DonJayamanne, line 275: Created a single function rather than copying pasting everywhere.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -219,8 +220,9 @@ export class Kernel implements IKernel {
         }
         const key = uri.toString();
         if (!this.kernelValidated.get(key)) {
-            this.isRawNotebookSupported =
-                this.isRawNotebookSupported || this.rawNotebookSupported.isSupportedForLocalLaunch();
```
DonJayamanne, line 223: This was wrong, if we're using remote connection, we cannot use `raw`

DonJayamanne: This is the bug that is being fixed in this PR.

rchiodo (approved):

src/client/datascience/jupyter/kernels/kernel.ts
IanMatthewHuff: @DonJayamanne Isn't this just the .supported method on the rawNotebookSupported Service?

IanMatthewHuff (approved):

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: Yes, youre right, we don't need to cache this.

PR #4676, opened 2021-02-05
Capture information related to kernel issues

DonJayamanne:
* Added a shared telemetry property to indicate whether the user executed a cell or not.
This way, if we get errors, we know whether this was due to auto start or not.
I.e. if jupyter/kernel auto starts & we send telemetry but user doen't run a notebook, then the errors can be ignored. Hopefully this property will help.
* Also modified some teleemtry like Jupyter Start failed, etc to send details of the error.
I look at the exception & try to determine the error type & send that information.

@IanMatthewHuff @rchiodo @DavidKutu @joyceerhl


DonJayamanne:
@IanMatthewHuff @DavidKutu @joyceerhl @rchiodo This is still a WIP.
Please have a look at the general solution.
Basically we call some method that is used to track some information against a Uri.
Later when we send the kernel information, we get that from the Uri.

Still more to do, but hopefully the current work outlines the general solution.

src/client/datascience/context/telemetry.ts
```diff
@@ -0,0 +1,103 @@
...
+            } else {
+                currentData.pythonEnvironmentVersion = undefined;
+            }
+        }
+        currentData.kernelConnectionType = currentData.kernelConnectionType || information.kernelConnection?.kind;
+    }
+
+    trackedInfo.set(key, currentData);
+}
+export function deleteTrackedInformation(resource: Uri) {
```
DonJayamanne, line 97: Plan is to invoke this when we close notebooks/interacive window..

src/client/datascience/context/telemetry.ts
```diff
@@ -0,0 +1,103 @@
...
+import { getOSType } from '../../common/utils/platform';
+import { EnvironmentType } from '../../pythonEnvironments/info';
+import { KernelConnectionMetadata } from '../jupyter/kernels/types';
+import * as hashjs from 'hash.js';
+import { Resource } from '../../common/types';
+import { IEventNamePropertyMapping, sendTelemetryEvent, sendTelemetryWhenDone } from '../../telemetry';
+import { StopWatch } from '../../common/utils/stopWatch';
+
+let connection: KernelConnectionMetadata;
+type ResourceSpecificTelemetryProperties = {
```
DonJayamanne, line 14: This is sent with the telemetry (will move this into the telemetry file so we track what properties are sent & the like).

src/client/datascience/raw-kernel/rawJupyterSession.ts
```diff
@@ -109,24 +114,51 @@ export class RawJupyterSession extends BaseJupyterSession {
         } catch (error) {
             this.connected = false;
             if (error instanceof CancellationError) {
-                sendTelemetryEvent(Telemetry.RawKernelSessionStartUserCancel);
+                sendKernelTelemetryEvent(resource, Telemetry.RawKernelSessionStart, stopWatch.elapsedTime, {
```
DonJayamanne, line 117: Use a new telemetry event.
This way we can use that single event & check the property for:
* success or failed
* reason if failed
With the queries we can easily group these, etc (i.e single event with props is more powerful for analysis).

Else if things fail, we have to use multiple events today to figure out why it failed & the like.

src/client/telemetry/index.ts
```diff
@@ -110,14 +113,21 @@ export function sendTelemetryEvent<P extends IEventNamePropertyMapping, E extend
...
+            // As we have errors for each event, those properties are treated as new data items.
+            // Hence they need to be classified as part of the GDPR process, and thats unnecessary and onerous.
+            eventNameSent = 'ERROR';
+            customProperties = { originalEventName: eventName as string, stackTrace: serializeStackTrace(ex) };
+            reporter.sendTelemetryErrorEvent(eventNameSent, customProperties, measures, []);
+        } else {
+            // Include a property failed, to indicate there are errors.
+            // Lets pay the price for better data.
+            customProperties = { failed: 'true', stackTrace: serializeStackTrace(ex) };
+            reporter.sendTelemetryEvent(eventNameSent, customProperties, measures);
```
DonJayamanne, line 129: This way we can keep using the same envet.
`NotebookStart` & look for properties whether it failed or not.
Easier than looking for other events `ERROR` & then filtering the properties to check when a particular telemetry point failed.

Hopefully @greazer  will be able to just accept the new telemetry proeprties as they arrive. Worst case we have 2 new properties for each event. & the names are generic enough to be accepted easily.

src/client/telemetry/index.ts
```diff
@@ -1032,14 +1046,29 @@ export interface IEventNamePropertyMapping {
     [Telemetry.RawKernelStartRawSession]: never | undefined;
     [Telemetry.RawKernelProcessLaunch]: never | undefined;
 
+    // Applies to everything (interactive+Notebooks & local+remote)
+    [Telemetry.NotebookStart]:
+        | ResourceSpecificTelemetryProperties // If successful.
+        | (ResourceSpecificTelemetryProperties & TelemetryErrorProperties); // If there any any unhandled exceptions.
```
DonJayamanne, line 1052: The benefit of sending `TelemetryErrorProperties` with this same event is that we know whether this event failed or not. 
Previously we'd send a seprate event `ERROR` and that would contain a property `originalEvent` & then we'd need to find out whether an event failed. (a very convoluted approach).
Now we can just query based on property `failed = true`

DonJayamanne:
@rchiodo @IanMatthewHuff @DavidKutu @joyceerhl This is WIP
Just wanted to get some early feedback

src/client/datascience/raw-kernel/rawJupyterSession.ts
DonJayamanne: Eg. if we wanted to take a count of success vs failure, then we have to do a join, this way we can just take a count grouped by `reason`.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -176,6 +179,10 @@ export class Kernel implements IKernel {
                         }
                     } catch (ex) {
                         traceError('failed to create INotebook in kernel', ex);
+                        sendKernelTelemetryEvent(options.document.uri, Telemetry.NotebookStart, stopWatch.elapsedTime, {
+                            failed: 'true',
+                            reason: getKernelFailureReason(ex)
```
DonJayamanne, line 184: Regardless of Remote, local, zmq/jupyter, if somethign fails, we know why it failed & we capture the reason here (look at `getkernelFailureReason`).

I.e. for Native Notebooks, this is the choke point for error handling/logging.

Hopefully I can find a similar point in WebViews & Interactive Window to handle all errors in one place & capure the logs in one place.

src/client/datascience/context/telemetry.ts
```diff
@@ -0,0 +1,139 @@
...
+    kernelConnection: KernelConnectionMetadata;
+    wasJupyterAutoStarted: boolean;
+    kernelDied: boolean;
+};
+
+const trackedInfo = new Map<string, ResourceSpecificTelemetryProperties>();
+const currentOSType = getOSType();
+
+export function getKernelFailureReason(error: Error) {
+    if (isErrorType(error, JupyterWaitForIdleError) || isErrorType(error, TimedOutError)) {
```
DonJayamanne, line 34: Depending on the error thrown we have a reason for the failure.

DonJayamanne: We have different errors for remote/local/zmq/jupyter, etc,

src/client/datascience/context/telemetry.ts
```diff
@@ -0,0 +1,139 @@
...
+type ContextualTelemetryProps = {
+    kernelConnection: KernelConnectionMetadata;
+    wasJupyterAutoStarted: boolean;
+    kernelDied: boolean;
+};
+
+const trackedInfo = new Map<string, ResourceSpecificTelemetryProperties>();
+const currentOSType = getOSType();
+
+export function getKernelFailureReason(error: Error) {
```
rchiodo, line 33: What about parsing the error returned from the kernel (the python exception)? Maybe you did that elsewhere.

rchiodo: Didn't find it elsewhere. That was one of the properties I thought we were going to try sending. Like if it was pyzmq, or pywin32 or DLL load or zmqconstants not found, etc.

rchiodo: Oh maybe you haven't gotten to it yet.

src/client/datascience/baseJupyterSession.ts
```diff
@@ -355,6 +355,9 @@ export abstract class BaseJupyterSession implements IJupyterSession {
 
     protected async waitForIdleOnSession(session: ISessionWithSocket | undefined, timeout: number): Promise<void> {
         if (session && session.kernel) {
+            if (!process.env.ASDYSDSD) {
```
DonJayamanne, line 358: Ooops, testing code, will remove

src/client/datascience/constants.ts
```diff
@@ -410,6 +410,7 @@ export enum Telemetry {
     RawKernelSessionConnect = 'DS_INTERNAL.RAWKERNEL_SESSION_CONNECT',
     RawKernelStartRawSession = 'DS_INTERNAL.RAWKERNEL_START_RAW_SESSION',
     RawKernelSessionStartSuccess = 'DS_INTERNAL.RAWKERNEL_SESSION_START_SUCCESS',
+    RawKernelSessionStart = 'DS_INTERNAL.RAWKERNEL_SESSION_START',
```
DonJayamanne, line 413: More info here https://github.com/microsoft/vscode-jupyter/pull/4676/files#r571327802

rchiodo:
I like the URI to properties matching. What resets it though? You said delete on close. What if I switch kernels 5 times on the same notebook? Are we going to send all 5 or one? I think we'd send all 5?

src/client/datascience/context/telemetry.ts
```diff
@@ -0,0 +1,139 @@
...
+        return;
+    }
+    const key = getUriKey(resource);
+    const currentData: ResourceSpecificTelemetryProperties = trackedInfo.get(key) || {
+        resourceType: getResourceType(resource)
+    };
+    if (information.kernelConnection) {
+        const interpreter = information.kernelConnection.interpreter;
+        if (interpreter) {
+            currentData.pythonEnvironmentType = interpreter.envType;
```
rchiodo, line 109: Yeah this looks like we only track interpreter once.

src/client/datascience/context/telemetry.ts
DonJayamanne: Yes, still more to be done

DonJayamanne: Lots more.

src/client/datascience/context/telemetry.ts
DonJayamanne: Yes, basically we track/add the information & when we send the data, we send what we have.

src/client/datascience/context/telemetry.ts
DonJayamanne: > Didn't find it elsewhere. That was one of the properties I thought we were going to try sending. Like if it was pyzmq, or pywin32 or DLL load or zmqconstants not found, etc.

 
@rchiodo looking at the past issues, users managed to re-install pyzmq or ipykernel to resolve these. Hence I believe just getting the versions of these related packages might be sufficient & the type of python they are using (conda, etc).
E.g. when using conda and it isn't activated, we get DLL load errors.

DonJayamanne: I.e. i'm suggesting we not check the stderr for now.
& add that later if required (if we still find many users failing there & the version information isnt' sufficient).

src/client/activation/types.ts
```diff
@@ -14,6 +14,10 @@ export const IExtensionActivationManager = Symbol('IExtensionActivationManager')
  * @extends {IDisposable}
  */
 export interface IExtensionActivationManager extends IDisposable {
+    /**
+     * Method invoked when extension activates (invoked once).
+     */
+    activateSync(): void;
```
DonJayamanne, line 20: Found having a sync version is better, sometimes we need things to happen before other things.

src/client/datascience/baseJupyterSession.ts
```diff
@@ -30,10 +31,9 @@ import { IJupyterSession, ISessionWithSocket, KernelSocketInformation } from './
  * @class JupyterSessionStartError
  * @extends {Error}
  */
-export class JupyterSessionStartError extends Error {
+export class JupyterSessionStartError extends WrappedError {
     constructor(originalException: Error) {
-        super(originalException.message);
-        this.stack = originalException.stack;
+        super(originalException.message, originalException);
```
DonJayamanne, line 36: So we dont lose original exception type.

src/client/activation/activationManager.ts
```diff
@@ -46,6 +53,9 @@ export class ExtensionActivationManager implements IExtensionActivationManager {
             this.docOpenedHandler = undefined;
         }
     }
+    public activateSync(): void {
```
joyceerhl, line 56: I'm not seeing where we call this method, do we still need it?

src/client/datascience/context/telemetry.ts
rchiodo: I believe we need the stderr to even tell if people are hitting these common errors.  Version information isn't sufficient for these things (for example, the pyzmq problem. Requires uninstall and reinstall - same version on both, something else messed it up) and there may be other ones we don't even know about that are common. 

What's the downside to adding the stderr?

src/client/activation/activationManager.ts
DonJayamanne: Fixed

DonJayamanne:
Note: Still need to test every single item, that will take a while.

src/client/datascience/context/telemetry.ts
rchiodo: Could we add a counter to the number of interpreters? Actually no, why not add a list of interpreters?

src/client/datascience/context/telemetry.ts
IanMatthewHuff: Stderr can contain PII was the consideration. It's not a structured error or exception, just raw output. Could contain stuff like paths or folder names, env vars, other stuff.

rchiodo: I think we can workaround that by just sending the filename and line number of the topmost frame. That should be non PII. We could even double check if it was a site packages thing or not.

rchiodo: filename and linenumber should be unique enough to tell what error it is.

DonJayamanne: Sure, will work on the stderr serpately then.
Challenge will be trying to tie it to the same telemetry, will be challenging, worst case we can have a seprate telemetry event for the stderr contents

rchiodo:
> Note: Still need to test every single item, that will take a while.
> @rchiodo @IanMatthewHuff @DavidKutu @joyceerhl
> 
> Personally I would like to have tests for this stuff, however it could take a lot of time & work.
> Trying to decide whether I should write the tests or not. I want to, but feel useful tests would take a lot of time (at least another 2 days, as there's a lot to be tested).

We could add the tests later so that we get this in before we ship.

src/client/datascience/context/telemetry.ts
DonJayamanne: > Could we add a counter to the number of interpreters? Actually no, why not add a list of interpreters?


Will add counter, not sure the list will be of any use.

DonJayamanne: Added a counter for number of interpreters as well.

src/client/datascience/context/telemetry.ts
```diff
@@ -0,0 +1,304 @@
...
+        let kv: Pick<IEventNamePropertyMapping, Telemetry.NotebookInterrupt>;
+        const data: typeof kv[Telemetry.NotebookInterrupt] = properties;
+        // Check result to determine if success.
+        if ('result' in data && data.result === InterruptResult.Success) {
+            clearInterruptCounter(resource);
+        }
+    }
+    // Once we have successfully restarted, clear the interrupt counter.
+    if (eventName === Telemetry.NotebookRestart) {
+        let kv: Pick<IEventNamePropertyMapping, Telemetry.NotebookInterrupt>;
```
IanMatthewHuff, line 264: This all looks incorrect in this block. Looks like it should be using restart methods and events? Copy paste error?

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -150,6 +162,29 @@ export class Kernel implements IKernel {
...
+        const notebook = await notebookPromise;
+        if (!notebook) {
+            return;
+        }
+        // Setup telemetry
+        if (!this.perceivedJupyterStartupTelemetryCaptured) {
+            this.perceivedJupyterStartupTelemetryCaptured = true;
+            sendTelemetryEvent(Telemetry.PerceivedJupyterStartupNotebook, stopWatch?.elapsedTime);
+            const disposable = notebook.onSessionStatusChanged((e) => {
+                if (e === ServerStatus.Busy) {
```
IanMatthewHuff, line 181: Name of this function is throwing me a bit. Is it perceived cold time to execute a cell? It looks like it's firing on changing of status to busy, which would just be pre-execute startup time at least as I see it. Not sure if it's a code issue or if I'm not getting what it's trying to measure.

IanMatthewHuff (changes requested):
There was one big block of code that looked like a copy paste issue. Rest of it seems pretty reasonable.

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: > s it perceived cold time to execute a cell?

Yes

> oks like it's firing on changing of status to busy, which would just be pre-execute startup time at least as I see i

yes, borrowed existing logic from the `ineractiveBase.ts` file for the same funciotnality, see here 
https://github.com/microsoft/vscode-jupyter/blob/88f11c2efc7db79b8faf10f10440667db2adc8bd/src/client/datascience/interactive-common/interactiveBase.ts#L685-L685

IanMatthewHuff: Sure that works. The name just implied to me that value would be from user opens a notebook and presses some type of run command to the time that the first cell completes, not when the first cell starts executing, but that may just be my understanding.

IanMatthewHuff (approved):

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: I have fixed this to just want for the promise to complete, that will be a more accurate measure of when execution completes.

src/client/datascience/context/interpreterCountTracker.ts
```diff
@@ -0,0 +1,29 @@
...
+    private static interpreterCount = 0;
+    public static get totalNumberOfInterpreters() {
+        return InterpreterCountTracker.interpreterCount;
+    }
+    constructor(
+        @inject(IPythonExtensionChecker) private readonly pythonExtensionChecker: IPythonExtensionChecker,
+        @inject(IInterpreterService) private readonly interpreterService: IInterpreterService
+    ) {}
+    public async activate() {
+        if (!this.pythonExtensionChecker.isPythonExtensionInstalled) {
```
rchiodo, line 21: Should this refresh when the python extension does get installed?

src/client/datascience/context/telemetry.ts
```diff
@@ -0,0 +1,320 @@
...
+    } else if (isErrorType(error, JupyterInstallError)) {
+        return 'jupyterinstall';
+    } else if (isErrorType(error, KernelDiedError)) {
+        return 'kerneldied';
+    } else if (isErrorType(error, FetchError)) {
+        return 'fetcherror';
+    }
+    return 'unknown';
+}
+export function sendKernelTelemetryEvent<P extends IEventNamePropertyMapping, E extends keyof P>(
```
rchiodo, line 85: I'm worried nobody will know when to sendKernelTelemetryEvent versus a sendTelemetryEvent

Could we send the different event based on the eventName?

rchiodo: Meaning sendTelemetryEvent would know whether or not to send all the extra kernel data based on the event name instead of having the caller determine whether or not to call sendKernelTelemetryEvent or sendTelemetryEvent.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -150,6 +164,27 @@ export class Kernel implements IKernel {
...
+        const notebook = await notebookPromise;
+        if (!notebook) {
+            return;
+        }
+        // Setup telemetry
+        if (!this.perceivedJupyterStartupTelemetryCaptured) {
+            this.perceivedJupyterStartupTelemetryCaptured = true;
+            sendTelemetryEvent(Telemetry.PerceivedJupyterStartupNotebook, stopWatch.elapsedTime);
+            executionPromise.finally(() =>
+                sendTelemetryEvent(Telemetry.StartExecuteNotebookCellPerceivedCold, stopWatch.elapsedTime)
```
rchiodo, line 184: Here's an example of where I'm not sure if this should be a 'kernel' event or just a normal telemetry event.

src/client/datascience/context/telemetry.ts
DonJayamanne: We can, but then I'd have to modify the `sendTelemetryEvent` method to overload it with the `resource` arguments.
& you do raise a very good point.
How about we re-visit that in a subseparate PR.
I do agree it will not get used much & it needs to be fixed. But feels a bigger change, hence would prefer to defer it.

src/client/datascience/context/interpreterCountTracker.ts
DonJayamanne: Fixed in another place, good pick.
Done

src/client/datascience/notebook/kernelProvider.ts
```diff
@@ -220,18 +228,32 @@ export class VSCodeKernelPickerProvider implements INotebookKernelProvider {
             );
         }
     }
-    private async getJupyterSessionManager() {
+    private async getJupyterSessionManager(resource: Uri) {
```
rchiodo, line 231: I thought you did this in another change? Maybe my diffing isn't working. Seems unrelated to telemetry.

rchiodo: Oh never mind. Saw the new telemetry at the bottom.

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: This is existing telemetry, hence didn't want to mess around with that one.
I created a new telemetry event that will get used in all remote, local, jupyter, raw, notebooks & interactive:

* Starting notebook, interrupting, restarting & switching kernels (capture success & failures) along with all of the information we are interested in (error type, etc)


That should give us what we need to identify issues.
In a serparate PR i will look into adding this to all other events as well.

src/client/datascience/raw-kernel/rawJupyterSession.ts
rchiodo: Is this same sort of detail sent for jupyter kernel failures? I didn't think it was but maybe I'm missing something.

src/client/datascience/notebook/kernelProvider.ts
DonJayamanne: That was different (separate PR to address kernel selection).
Here when things fail, i need the URI so we can log the errors (hence passing it into this).

src/client/extensionActivation.ts
```diff
@@ -131,6 +131,7 @@ async function activateLegacy(
 
     const manager = serviceContainer.get<IExtensionActivationManager>(IExtensionActivationManager);
     context.subscriptions.push(manager);
+    manager.activateSync();
     const activationPromise = manager.activate();
```
rchiodo, line 135: Is this necessary if you're activating sync?

rchiodo (approved):

src/client/extensionActivation.ts
DonJayamanne: Yes, this guarantees the fact that it will happen before we execute the rest of the async code.
We need to ensure some of the code has been setup before we can start async code.
E.g. setting of shared properties, initiallizing some stuff before we start async, async stuff could be things like `starting juypyter` in the background based on auto start.

PR #4687, opened 2021-02-05
Use Environment variable to detect if AML Compute

DonJayamanne:
For #4550

### DIFF
--- a/src/client/common/amlContext.ts
+++ b/src/client/common/amlContext.ts
@@ -1,36 +1,22 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import { inject, injectable, named } from 'inversify';
-import { Memento } from 'vscode';
+import { injectable } from 'inversify';
+import { env } from 'vscode';
 import { IExtensionSingleActivationService } from '../activation/types';
 import { setSharedProperty } from '../telemetry';
-import { GLOBAL_MEMENTO, IMemento } from './types';
-import { noop } from './utils/misc';
 
-const amlComputeMementoKey = 'JVSC_IS_AML_COMPUTE_INSTANCE';
+const amlComputeRemoteName = 'amlext';
 
 @injectable()
 export class AmlComputeContext implements IExtensionSingleActivationService {
-    constructor(@inject(IMemento) @named(GLOBAL_MEMENTO) private readonly memento: Memento) {}
-
+    constructor() {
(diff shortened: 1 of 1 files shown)


src/client/common/amlContext.ts
```diff
@@ -5,32 +5,20 @@ import { inject, injectable, named } from 'inversify';
...
 
     public get isAmlCompute() {
         return this.memento.get<boolean>(amlComputeMementoKey, false) || this.isAmlComputeWorkspace();
     }
     private isAmlComputeWorkspace() {
-        let isRemoteConnection: boolean = false;
-        // If there is at least 1 folder, we can check if the scheme and authority match a remote connection
-        if (Array.isArray(this.workspace.workspaceFolders) && this.workspace.workspaceFolders.length > 0) {
-            // We are in a remote connection to a CI if the scheme is vscode-remote and the authority contains amlext
-            isRemoteConnection = this.workspace.workspaceFolders.some(
```
DonJayamanne, line 29: This code doesn't work as we're on the remote machine.

DavidKutu (approved):

src/client/common/amlContext.ts
```diff
@@ -5,32 +5,20 @@ import { inject, injectable, named } from 'inversify';
...
-                (item) => item.uri.scheme === 'vscode-remote' && item.uri.authority.indexOf('amlext') >= 0
-            );
-        }
-        return isRemoteConnection;
+        return typeof process.env.AZURE_EXTENSION_DIR === 'string';
```
IanMatthewHuff, line 21: Did the AML team point this out? I guess this works as far as I know, but it seems much broader versus the other check. This seems like it would always be set for all of VS Code if you AML installed?

DonJayamanne: Nope, their solution didn't work. I'm using this for now.
Worst case scenario users on a VM with this variable will end up with Native Notebook UI instead of webview UI.

Hence done see much of an issue at this state, i.e. this is a hack.
FYI - @greazer & AML team have been informed of this, Worst case we revert this, at least this gets it working (again no harm).

IanMatthewHuff (approved):
If the PMs are fine with it, then I'm good.

src/client/common/amlContext.ts
DonJayamanne: Found a better fix @IanMatthewHuff

rchiodo (approved):

joyceerhl (approved):

IanMatthewHuff (approved):

Issue #302, opened 2020-10-27
Tensor tooltips


joyceerhl:
To implement this, modify this code here:
https://github.com/microsoft/vscode-jupyter/blob/f29073ed195d134827f581251a53c481d88584ef/src/client/datascience/editor-integration/hoverProvider.ts#L111-L118

@greazer mentioned wanting to make the tooltips we display configurable for users. One way to do this is retrieve more custom properties in `IJupyterVariable`, then create a setting (sorry PMs), then display those properties in our tooltips if available for each type. Something like

```
"jupyter.variableTooltipFields": {
    "Tensor": [ "dtype", "shape", "device" ],
    "DataFrame": [ "somePropertyOfInterest" ]
}
```

Then we can read this setting in the hoverProvider.ts and filter out the properties that the user has opted into when constructing the `vscode.Hover` contents for types that are explicitly configured in settings.

Not a requirement yet, we can just hardcode these for the MVP, but just an idea, and this mechanism would work for any other custom variable tooltips we want to support.

rchiodo:
So I think you're suggesting:
- A configurable setting based on type? that uses a regex to parse the value for a tooltip?
- We implement this parsing in the code mentioned above?

joyceerhl:
To validate, you'll need to opt into the experiment:

> "jupyter.experiments.optInto": [ "jupyterEnhancedTensorTooltips" ]

Then run an interactive Python file cell containing this code in the interactive window, place your cursor above the variable name, and verify that a tooltip appears containing shape, dtype and device info:
```python
import torch
foo = torch.tensor([[0, 1, 2], [3, 4, 5]])
```

IanMatthewHuff:
Validated. At least I'm pretty sure this is what we are looking for. I can also try adding a new property.

![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/861656c2-60cc-47ef-a434-9ba0e79e38e4)

PR #4693, opened 2021-02-06
Add improved Tensor tooltips for Python files which have been run in the interactive window

joyceerhl:
For #302

Extends https://github.com/microsoft/vscode-python/pull/11756 by introducing a `jupyter.variableTooltipFields` setting allowing configuration of tooltips based on type.

image


package.json
```diff
@@ -1782,6 +1782,16 @@
                         "daily"
                     ],
                     "scope": "application"
+                },
+                "jupyter.variableTooltipFields": {
```
joyceerhl, line 1786: Ran this setting past jmew and greazer, they thought having a setting for this was fine. We may be changing which attributes we show users based on a QP anyway.

src/test/datascience/editor-integration/hoverProvider.vscode.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        await sleep(1_000);
+
+        // Execute its contents in the interactive window
+        await vscode.commands.executeCommand<void>('jupyter.runallcells', textDocument.uri);
+
+        // Request a hover on the line containing a tensor variable declaration
+        const hoverProvider = api.serviceContainer.get<IHoverProvider>(IHoverProvider);
+        const position = new vscode.Position(11, 1);
+        const cancelTokenSource = new vscode.CancellationTokenSource();
+        const result = await hoverProvider.provideHover(textDocument, position, cancelTokenSource.token);
```
joyceerhl, line 39: Test is still failing because kernelVariables doesn't appear to know anything about our variables, will fix before merge.

src/client/datascience/jupyter/kernelVariables.ts
```diff
@@ -421,6 +422,37 @@ export class KernelVariables implements IJupyterVariables {
         return result;
     }
 
+    public async getVariableProperties(
```
joyceerhl, line 425: The alternative is to fetch these attributes on the original IJupyterVariable, which might help with returning a hover more quickly.

package.json
IanMatthewHuff: It is a bit of a complex setting. Not sure how many customers would understand it based on the description. But doesn't hurt to have I suppose. Even if no one external uses it, it keeps us thinking in a more extensible way.

IanMatthewHuff: Having keyed by language when the description say that it's only for python is a bit funny.

src/client/datascience/editor-integration/hoverProvider.ts
```diff
@@ -67,7 +74,7 @@ export class HoverProvider implements INotebookExecutionLogger, vscode.HoverProv
         token: vscode.CancellationToken
     ): vscode.ProviderResult<vscode.Hover> {
         const timeoutHandler = async () => {
-            await sleep(100);
+            await sleep(300);
```
IanMatthewHuff, line 77: Just for info, why was the change needed here? Is it slower now and was getting called too often?

package.json
joyceerhl: Yeah, that's true. I do think it should be keyed by language but I don't believe we support executing interactive cells in other languages in the IW today, so I didn't want the description to imply that other languages should work now.

DonJayamanne: I personally think we shouldn't be adding this. Based on what I've learnt over the years & what other PMs have said, this isn't the best way. To quote a famous man `You don't hand out a cookie to someone & ask if they want a cookie`. 
But, not going to block this PR for this.

DonJayamanne: Would we remove this property if its not used?
If 10 out of 10,000 people use this, would we still keep this property or remote it.

The draw back of adding such properties is the properties window gets overloaded with 1000s of items (as reported by plenty of PMs as well).
@jmew /cc

src/client/common/configSettings.ts
```diff
@@ -95,6 +96,11 @@ export class JupyterSettings implements IWatchableJupyterSettings {
     // Hidden settings not surfaced in package.json
     public disableZMQSupport: boolean = false;
     public verboseLogging: boolean = false;
+    public variableTooltipFields: JSONObject = {
```
IanMatthewHuff, line 99: I don't know that I would sink a ton of time into this. But could we examine if we could make this a specific type, and not JSONObject? JSONObject captures like...everything I think. I would think that we could get a specific interface for the format that we expect here.

src/client/datascience/editor-integration/hoverProvider.ts
joyceerhl: The original timeout was too aggressive, so even without my changes the tooltips would show up maybe 30% of the time because of this timeout.

package.json
DonJayamanne: Why not remove this from `package.json` & we support it.
Just as we support other settings but not advertised in `package.json`
Blocking based on the request that we remove this from here & just add support for it.

We do this today for `"jupyter.disableZMQSupport"`, its a property that is supported in settings.json but not advertised, so users don't see it.

@joyceerhl @jmew (this way we can customize it for QPs & never have to add it unless users need it).

.github/test_plan.md
```diff
@@ -68,6 +68,9 @@
     1. Verify the output from what you added
 -   [ ] Verify dark and light main themes
     1. Repeat the `Start and connect to local server` and `Verify basic outputs` steps using `Default Dark+` and `Default Light+` themes
+-   [ ] Verify tooltips in Python file that has been run in the interactive window
+    1. After manualTestFile.py has been run in the interactive window, place your cursor over the `myTensor` variable.
+    1. Verify that the tooltip that appears contains four lines where the first line says `myTensor: Tensor`, the second line has `shape` information, the third line has `dtype` information, and the fourth line has `device` information.
```
DonJayamanne, line 73: Do we need this if we have an end to end test for this.
I think we should only add stuff that's difficult to test via code

package.json
joyceerhl: Good idea, I'm fine with that.

DonJayamanne (changes requested):
Can we remove the changes from package.json & leave the rest (this way you can still configure if required)

.github/test_plan.md
joyceerhl: Aren't we supposed to update the test plan with new functionality? I'm happy to remove this, as the test I have simulates this functionality pretty well. Just not sure what our policy on test plan updates is.

src/client/datascience/editor-integration/hoverProvider.ts
```diff
@@ -94,13 +101,19 @@ export class HoverProvider implements INotebookExecutionLogger, vscode.HoverProv
...
-                            notebooks.map((n) => this.variableProvider.getMatchingVariable(word, n, t))
+                        const attributes = await Promise.race(
+                            notebooks.map((n) => this.variableProvider.getVariableProperties!(word, n, t))
                         );
-                        if (match) {
-                            return {
-                                contents: [`${word} = ${match.value}`]
+                        if (attributes) {
+                            const asMarkdown =
+                                Object.entries(attributes).reduce(
```
IanMatthewHuff, line 109: I always forget about reduce. That's a nice way to put the string together. If it doesn't have any matching attributes does it create an empty line? Just not sure if the \n goes on the end of the entry or on the start. From the docs it looks like maybe the initial value should just be ``` and the \n would be on the start of each entry, not the end.

src/client/datascience/editor-integration/hoverProvider.ts
```diff
@@ -94,13 +101,19 @@ export class HoverProvider implements INotebookExecutionLogger, vscode.HoverProv
                     const notebooks = this.getMatchingNotebooks(document);
                     if (notebooks && notebooks.length) {
                         // Just use the first one to reply if more than one.
-                        const match = await Promise.race(
-                            notebooks.map((n) => this.variableProvider.getMatchingVariable(word, n, t))
+                        const attributes = await Promise.race(
+                            notebooks.map((n) => this.variableProvider.getVariableProperties!(word, n, t))
```
IanMatthewHuff, line 105: I might toss on a comment here about the use of !. Had to actually look it up myself to see if this was safe. It's a bit funny since it's optional on IJupyterVariables, but this class now demands that it exists. It works since we are specifically injecting the kernelVariables, but a comment explaining that here might be nice.

src/client/datascience/editor-integration/hoverProvider.ts
joyceerhl: If it doesn't have any matching attributes we probably shouldn't return a hover result at all.

IanMatthewHuff (approved):

.github/test_plan.md
rchiodo: I don't think it hurts anything here other than it taking longer for CTI to test.

src/client/common/configSettings.ts
rchiodo: I believe it can just be an object.

src/client/datascience/jupyter/kernelVariables.ts
rchiodo: I have another idea for this. Improve IPython such that it returns these items for tensors in the inspect request. Then we use that to get the values (this way jupyter gets improved tensor support too).

We could do that in parallel to this though.

src/client/datascience/jupyter/kernelVariables.ts
```diff
@@ -421,6 +422,37 @@ export class KernelVariables implements IJupyterVariables {
...
+        word: string,
+        notebook: INotebook,
+        cancelToken: CancellationToken | undefined
+    ): Promise<JSONObject> {
+        const matchingVariable = await this.getMatchingVariable(word, notebook, cancelToken);
+        const settings = this.configService.getSettings(notebook.resource).variableTooltipFields;
+        const languageSettings = settings['python'] as JSONObject;
```
rchiodo, line 432: Hardcoding python here shouldn't be necessary. Other code in this file tries to be language agnostic.

rchiodo (approved):

src/client/datascience/jupyter/kernelVariables.ts
joyceerhl: Good idea. Would we file an issue for them and offer to submit a PR? Or just submit the PR?

rchiodo: File an issue, then submit a PR I'd say.

joyceerhl: Opened an issue here: https://github.com/ipython/ipython/issues/12816

joyceerhl:
Finally got test passing 🙄 https://github.com/microsoft/vscode-jupyter/pull/4693/checks?check_run_id=1866750633#step:31:942

rchiodo (approved):

joyceerhl:
@DonJayamanne I think I've addressed your feedback (package.json should not list setting).

src/client/datascience/jupyter/kernelVariables.ts
joyceerhl: Oh wait, didn't see this. GitHub shows outdated code snippets for stuff that's been reviewed, so I think you were looking at the old version. I initially passed the language through from the hover provider request, but I much prefer this idea and it's more consistent with what the rest of kernelVariables does. Thanks for the suggestion :)

DonJayamanne (approved):

Issue #4694, opened 2021-02-07
Native Notebooks - Notebook to display on open for experiment


claudiaregio:
Notebook to be displayed to those in Native Notebooks open first opening. Notebook only to be displayed once. 
[Welcome to Native Notebooks!.zip](https://github.com/microsoft/vscode-jupyter/files/5938002/Welcome.to.Native.Notebooks.zip)

DonJayamanne:
For #4600

rchiodo:
Validated.

PR #4714, opened 2021-02-08
Create session manager when required

DonJayamanne:
For #4694
Previously we were trying to use a session manager that had already been disposed elsewhere.

### DIFF
--- a/src/client/datascience/jupyter/kernels/kernelSelections.ts
+++ b/src/client/datascience/jupyter/kernels/kernelSelections.ts
@@ -95,27 +95,35 @@ export class KernelSelectionProvider {
      */
     public async getKernelSelectionsForRemoteSession(
         resource: Resource,
-        sessionManager: IJupyterSessionManager,
+        sessionManagerCreator: () => Promise<IJupyterSessionManager>,
         cancelToken?: CancellationToken
     ): Promise<IKernelSpecQuickPickItem<LiveKernelConnectionMetadata | KernelSpecConnectionMetadata>[]> {
         const getSelections = async () => {
-            const installedKernelsPromise = new InstalledJupyterKernelSelectionListProvider(
-                this.kernelService,
-                this.pathUtils,
-                this.extensionChecker,
-                this.interpreterService,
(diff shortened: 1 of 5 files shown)


src/client/datascience/notebook/kernelProvider.ts
```diff
@@ -113,7 +113,7 @@ export class VSCodeKernelPickerProvider implements INotebookKernelProvider {
         }
         const [preferredKernel, kernels] = await Promise.all([
             this.getPreferredKernel(document, token, sessionManager),
-            this.getKernelSelections(document, token, sessionManager)
+            this.getKernelSelections(document, token)
         ]).finally(() => (sessionManager ? sessionManager.dispose() : undefined));
```
DonJayamanne, line 117: The problem was, the list of kernels was still being fetched & sessionManager was disposed.
This is possible as we were lazy loading the kernels.
We return cached items immediately  & fetch in the background using the session manager.
But this code sees that the result has returned hence we dispose the session manager, & the background code falls over.

src/client/datascience/jupyter/kernels/kernelSelections.ts
```diff
@@ -95,27 +95,35 @@ export class KernelSelectionProvider {
      */
     public async getKernelSelectionsForRemoteSession(
         resource: Resource,
-        sessionManager: IJupyterSessionManager,
+        sessionManagerCreator: () => Promise<IJupyterSessionManager>,
```
DonJayamanne, line 98: We cannot give it a session manager, we need to pass a factory so that this function can manage the lifetime of the session manager object.

DonJayamanne: Previously we'd return the cached item while in the background we'd retrieve the kernels from session manager, but the calling code would see that the promise was resolved hence the session manager would be disposed.
Hence calling code cannot own the lifetime of the session manager.

rchiodo (approved):

IanMatthewHuff (approved):

PR #4732, opened 2021-02-10
Update outputs API usage

rebornix:
This PR adopts the new output API (some test files are not updated properly yet), which includes following changes

* adopt the new `NotebookCellOutput`, which replaces `IErrorOutput`, `IDisplayOutput`, etc.
  * now instead of providing a object literal, we always create a `NotebookCellOutput` object and give this back to VS Code
* change how output types are validated
  * `hasErrorOutput = output.outputs.find(opit => opit.mime === 'application/x.notebook.error-traceback')`
  * `hasStreamOutput = output.outputs.find(opit => opit.mime === 'application/x.notebook.stream')`
  * others are all rich outputs

The build is not green as there are quite a few places where we check `cellKind === CellKind.Text/Error/Rich` which is no longer necessary anymore.

https://github.com/microsoft/vscode-jupyter/pull/4732/files#diff-afc015d140e9fa1b823f6e2faa3e8153bf58510b747d05e776054ba2330afce9 represents how outputs should be created with the new API. The change in this file is simple as it has good abstraction. While other files would check by

```
        assert.equal(cells[0].outputs.length, 2, 'Incorrect number of output');
        const output1 = cells[0].outputs[0] as CellDisplayOutput;
        const output2 = cells[0].outputs[1] as CellDisplayOutput;
        assert.equal(output1.metadata?.custom?.vscode?.outputType, 'stream', 'Incorrect output type');
        assert.equal(output2.metadata?.custom?.vscode?.outputType, 'stream', 'Incorrect output type');
        assert.equal(output1.metadata?.custom?.vscode?.name, 'stdout', 'Incorrect stream name');
        assert.equal(output2.metadata?.custom?.vscode?.name, 'stderr', 'Incorrect stream name');
        assert.equal(output1.outputKind, vscodeNotebookEnums.CellOutputKind.Rich, 'Incorrect output type');
        assert.equal(output2.outputKind, vscodeNotebookEnums.CellOutputKind.Rich, 'Incorrect output type');
```

I would suggest introducing helper functions similar to `validateCellOutputTranslation` https://github.com/microsoft/vscode-jupyter/pull/4732/files#diff-afc015d140e9fa1b823f6e2faa3e8153bf58510b747d05e776054ba2330afce9R75.

### DIFF
--- a/src/client/datascience/jupyter/kernels/cellExecution.ts
+++ b/src/client/datascience/jupyter/kernels/cellExecution.ts
@@ -6,9 +6,10 @@
 import { nbformat } from '@jupyterlab/coreutils';
 import type { KernelMessage } from '@jupyterlab/services/lib/kernel/messages';
 import { ExtensionMode } from 'vscode';
-import type {
-    CellDisplayOutput,
+import {
     NotebookCell,
+    NotebookCellOutput,
+    NotebookCellOutputItem,
     NotebookCellRunState,
     NotebookEditor as VSCNotebookEditor
 } from '../../../../../types/vscode-proposed';
@@ -35,6 +36,7 @@ import {
     cellOutputToVSCCellOutput,
     clearCellForExecution,
     getCellStatusMessageBasedOnFirstCellErrorOutput,
+    hasErrorOutputs,
     isStreamOutput,
     traceCellMessage,
     updateCellExecutionTimes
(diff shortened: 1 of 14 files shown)


rebornix:
File opening, outputs rendering and execution works with latest code. Tests are not working but leave it to @DonJayamanne as there are some infra changes need to be made, which I'm not familiar with.

PR #4742, opened 2021-02-10
Add more logging to capture versions of ipykenrel & other packages

DonJayamanne:
Logging of StdErr will be done in a separate PR.


src/client/api/pythonApi.ts
```diff
@@ -201,14 +201,26 @@ const ProductMapping: { [key in Product]: JupyterProductToInstall } = {
...
         product: Product,
         resource?: InterpreterUri,
         cancel?: CancellationToken
     ): Promise<InstallerResponse> {
-        return this.apiProvider.getApi().then((api) => api.install(ProductMapping[product], resource, cancel));
+        return this.apiProvider
+            .getApi()
+            .then((api) => api.install(ProductMapping[product], resource, cancel))
+            .then((result) => {
+                if (result === InstallerResponse.Installed) {
```
DonJayamanne, line 219: Any time a module is installed we want to fetch this once again.
Else we keep the same items in the case.
Should we blow away the cache evey `n` hours?
Feels like we should (incase user keeps this open).
But if the user installs/uninstalls from CLI, there's not much we can do.
Don't want to fech very time, that will be very slow.

src/client/datascience/telemetry/interpreterPackageTracker.ts
```diff
@@ -0,0 +1,81 @@
...
+    ) {}
+    public async activate(): Promise<void> {
+        if (!isLocalLaunch(this.configurationService)) {
+            return;
+        }
+        this.notebook.onDidChangeActiveNotebookKernel(this.onDidChangeActiveNotebookKernel, this, this.disposables);
+        this.interpreterService.onDidChangeInterpreter(this.trackPackagesOfActiveInterpreter, this, this.disposables);
+        this.installer.onInstalled(this.onDidInstallPackage, this, this.disposables);
+        this.extensions.onDidChange(this.trackUponActivation, this, this.disposables);
+        this.trackUponActivation().catch(noop);
```
DonJayamanne, line 39: Fetch the list:
* When a kernel is selected in a VSC Notebook
* When packages are installed
* When interpreter changes
* When extension loads
* When Python extension is installed

src/client/datascience/jupyter/kernels/kernelSelector.ts
```diff
@@ -504,13 +544,18 @@ export class KernelSelector implements IKernelSelectionUsage {
     }
     /**
      * Get our kernelspec and interpreter for a local raw connection
+     *
+     * @param {boolean} [ignoreTrackingKernelInformation]
+     * As a side effect these method tracks the kernel information for telemetry, we should ensure that tracking is disabled for Native Notebooks. Native Notebooks knows exactly what the current kernel information is, webviews/interactive is not the same.
+     * I.e. whenever these methods are called by webviews/interactive assume the return value is the active kernel.
```
rchiodo, line 550: This seems like the wrong place to track this then? There's an implicit assumption about another piece of code understanding that this is where kernels are tracked whereas the other piece of code is deciding to select them. Does this assumption make this code more fragile?

src/client/datascience/telemetry/interpreterPackageTracker.ts
rchiodo: Perhaps you should put this in a comment?

src/client/datascience/telemetry/interpreterPackages.ts
```diff
@@ -0,0 +1,109 @@
...
+    @captureTelemetry(Telemetry.AttemptedToLaunchRawKernelWithoutInterpreter)
+    private async getPackageInformation(interpreter: PythonEnvironment) {
+        const service = await this.executionFactory.createActivatedEnvironment({
+            allowEnvironmentFetchExceptions: true,
+            bypassCondaExecution: true,
+            interpreter
+        });
+
+        // Ignore errors, and merge the two (in case some versions of python write to stderr).
+        const output = await service.execModule('pip', ['list'], { throwOnStdErr: false, mergeStdOutErr: true });
```
rchiodo, line 89: Does pip always work in a conda situation?

DonJayamanne: Yes, also we have no way of invoking conda commands, Python extension has that API & we don't.
But i checked this & it works well enough for us.

src/client/datascience/jupyter/kernels/kernelSelector.ts
DonJayamanne: Not sure we can do anything about it, unless we refactor a lot of the existing code. this is how WebViews & Interactive window works today.
When we try to run a cell/code, we do'nt know what the actual kernel is, not upfront (unlike native notebooks). 
In inteacive & webviews, we need to get into the lower layers & that where we try to detect the kernel information. This is something i tried to adddress a few weeks ago, and only got it fixed for Native Notebook.

I'm not too keen on fixing that in this PR, 
its just how our code works today.

DonJayamanne: But agree, its ugly, hence my comments (so we know why its ugly, & don't remove it).

src/client/datascience/telemetry/interpreterPackages.ts
DonJayamanne: I.e. pip will have to do for now.

src/client/datascience/jupyter/kernels/kernelSelector.ts
rchiodo: Maybe add an issue to fix it?

rchiodo (approved):

DonJayamanne:
CI seems to have hung

src/client/api/pythonApi.ts
IanMatthewHuff: I'm not sure exactly what you mean by this comment.

DavidKutu (approved):

src/client/datascience/telemetry/interpreterPackageTracker.ts
```diff
@@ -0,0 +1,81 @@
...
+            return;
+        }
+        // Get details of active interpreter for the Uri provided.
+        const activeInterpreter = await this.interpreterService.getActiveInterpreter(undefined);
+        if (!activeInterpreter) {
+            return;
+        }
+        await this.packages.trackPackages(activeInterpreter);
+    }
+    @swallowExceptions()
```
IanMatthewHuff, line 68: If you expect this could fail, wouldn't you also want to swallow exceptions on the other functions here? Such as trackPackagesOfActiveInterpreter? Looks like they are calling the same basic functions, so if one could throw, then they all could.

src/client/datascience/telemetry/interpreterPackageTracker.ts
```diff
@@ -0,0 +1,81 @@
...
+            return;
+        }
+        this.activeInterpreterTrackedUponActivation = true;
+        await this.trackPackagesOfActiveInterpreter();
+    }
+    private async trackPackagesOfActiveInterpreter() {
+        if (!this.pythonExtensionChecker.isPythonExtensionInstalled) {
+            return;
+        }
+        // Get details of active interpreter for the Uri provided.
```
IanMatthewHuff, line 61: Copy paste comment.

src/client/datascience/telemetry/interpreterPackages.ts
```diff
@@ -0,0 +1,114 @@
...
+        const promise = this.getPackageInformation(interpreter);
+        promise.finally(() => {
+            // If this promise was resolved, then remove it from the pending list.
+            if (InterpreterPackages.pendingInterpreterInformation.get(key) === promise) {
+                InterpreterPackages.pendingInterpreterInformation.delete(key);
+            }
+        });
+        InterpreterPackages.pendingInterpreterInformation.set(key, promise);
+    }
+    @captureTelemetry(Telemetry.AttemptedToLaunchRawKernelWithoutInterpreter)
```
IanMatthewHuff, line 79: This looks like a mistake of some type. This isn't launching the raw kernel.

IanMatthewHuff (changes requested):
Had a few comments, but at a minimum it looks like there is a rogue telemetry event in here.

src/client/datascience/telemetry/interpreterPackageTracker.ts
DonJayamanne: Fixd, didnt need them

src/client/datascience/telemetry/interpreterPackageTracker.ts
DonJayamanne: Fixed

src/client/datascience/telemetry/interpreterPackages.ts
DonJayamanne: COpy paste, thanks

Issue #4730, opened 2021-02-10
Hide all but 3 of our icons in the native notebook


rchiodo:
For both stable and insiders hide all but 3.

The 3 are to be determined.

Others can be forced into the overflow menu.

claudiaregio:
Top 3 to be used: 
1. Restart kernel
2. Variable Explorer
3. Export As 
When kernel is in busy state, turn the restart kernel notebook toolbar icon into the STOP EXECUTION RED SQUARE ICON. When kernel is in idle state, turn back into the kernel restart icon.

rchiodo:
Not sure yet about the the icon/command changing.

This might work:

```
                {
                    "command": "jupyter.notebookeditor.restartkernel",
                    "title": "%jupyter.command.jupyter.restartkernel.title%",
                    "group": "navigation@3",
                    "when": "notebookViewType == 'jupyter-notebook' && !jupyter.isbusy"
                },
                {
                    "command": "jupyter.notebookeditor.interruptkernel",
                    "title": "%jupyter.command.jupyter.interruptkernel.title%",
                    "group": "navigation@3",
                    "when": "notebookViewType == 'jupyter-notebook' && jupyter.isbusy"
                },
```

I'd have to code it up to find out.

IanMatthewHuff:
Validated.
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/ae0c3c76-c89f-49a6-87f0-15b0044e5f0c)
Restart, Variable, Export As

PR #4747, opened 2021-02-10
Change number of icons in notebook toolbar

rchiodo:
For #4730

We'll now have:
- REstart/Interrupt depending upon the state of the kernel
- Variable Explorer
- Export As

### DIFF
--- a/package.json
+++ b/package.json
@@ -590,7 +590,10 @@
                 "command": "jupyter.notebookeditor.interruptkernel",
                 "title": "%jupyter.command.jupyter.interruptkernel.title%",
                 "category": "Notebook",
-                "icon": "$(primitive-square)",
+                "icon": {
+                    "light": "resources/light/interrupt.svg",
+                    "dark": "resources/dark/interrupt.svg"
+                },
                 "enablement": "notebookViewType == 'jupyter-notebook' && jupyter.isnotebooktrusted && jupyter.notebookeditor.canInterruptNotebookKernel"
             },
             {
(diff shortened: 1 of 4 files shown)


DavidKutu (approved):

package.json
```diff
@@ -785,55 +788,49 @@
...
+                    "group": "navigation@1",
+                    "when": "notebookViewType == 'jupyter-notebook' && jupyter.notebookeditor.canrestartNotebookkernel && !jupyter.notebookeditor.canInterruptNotebookKernel"
                 },
                 {
                     "command": "jupyter.notebookeditor.interruptkernel",
                     "title": "%jupyter.command.jupyter.interruptkernel.title%",
-                    "group": "navigation@4",
-                    "when": "notebookViewType == 'jupyter-notebook'"
+                    "group": "navigation@1",
+                    "when": "notebookViewType == 'jupyter-notebook' && jupyter.notebookeditor.canInterruptNotebookKernel"
```
IanMatthewHuff, line 810: Maybe jupyter.isnotebooktrusted on interrupt / restart as well? Obviously it should be covered by canrestart or canInterrupt, but provides an extra degree of safety just in case we have a bug with those other contexts so that we don't accidentally enable something when not trusted. Surprised we didn't have the trusted check already.

IanMatthewHuff: Actually they already have the enablements in commands sections? That's odd, which version wins if they are different? I'd assume the more specific one.

rchiodo: The more specific one

I didn't want to disable the command in the command palette

IanMatthewHuff: So is runallcellsabove enabled when not trusted then? Looks like our more specific version here for the editor doesn't have the trusted check

rchiodo: It's there but disabled.

IanMatthewHuff (approved):

PR #4754, opened 2021-02-11
Analyze stderr output from kernel processes wen they die

DonJayamanne:


src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -172,8 +172,9 @@ export class KernelProcess implements IKernelProcess {
         try {
             await Promise.race([
                 tcpPortUsed.waitUntilUsed(this.connection.hb_port, 200, timeout),
+                deferred.promise,
```
DonJayamanne, line 175: Used a promise instead of a cancellation token, else we lose all of the error information.

src/client/datascience/context/telemetry.ts
```diff
@@ -75,13 +75,69 @@ export function getErrorClassification(error: Error) {
         return 'jupyterconnection';
     } else if (isErrorType(error, JupyterInstallError)) {
         return 'jupyterinstall';
+    } else if (isErrorType(error, PythonKernelDiedError)) {
+        return getReasonForKernelToDie(error);
```
DonJayamanne, line 79: Beginning to see a lot of benefit in creating specific error classes for each error.
Like in .NET & other langauges. Feel we should do this more often instead of just throwing  a generic error `new Error('xyz')`.

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -160,7 +160,7 @@ export class KernelProcess implements IKernelProcess {
                         });
                         exitEventFired = true;
                     }
-                    cancelWaiting.cancel();
+                    deferred.reject(error);
```
DonJayamanne, line 163: Better than cancellation, to preserve the original error, else we loose a lot of it when its bubbled up.

src/client/datascience/context/telemetry.ts
```diff
@@ -75,13 +75,110 @@ export function getErrorClassification(error: Error) {
...
     } else if (isErrorType(error, KernelDiedError)) {
         return 'kerneldied';
     } else if (isErrorType(error, FetchError)) {
         return 'fetcherror';
     }
     return 'unknown';
 }
+/**
```
rchiodo, line 87: Not happy with this. We'll only ever identify problems we already know about. Which are theoretically useless once we fix them.

rchiodo: Can we wait until Olivia responds? Or think about another way we can achieve uniqueness that we can track down

rchiodo: Had another idea. 

We could identify the set of site-packages that we know have potential errors thrown in them. Those would be okay to send as they're not user information.

So that list might be:

```
ipykernel
jupyter
zmq
traitlets
sys
os
jupyter-client
nb-client
``` 

At least this would catch more errors.

rchiodo (changes requested):
⏲️

src/client/datascience/context/telemetry.ts
DonJayamanne: Will do that and leave this code as is.

DonJayamanne: I.e. send the hash of the folder & file from the last frame

rchiodo: Sounds good. We can add as many packages as we can think of to that list.

src/client/telemetry/index.ts
```diff
@@ -42,19 +42,13 @@ function isTelemetrySupported(): boolean {
...
-function getErrorClassification(error: Error): string {
-    for (const classifier of errorClassifiers) {
-        const classification = classifier(error);
-        if (classification && classification !== 'unknown') {
-            return classification;
-        }
-    }
-    return 'unknown';
+function populateTelemetryWithErrorInfo(props: Record<string, any>, error: Error) {
+    errorTelemetryUpdaters.forEach(item => item(props, error));
```
DonJayamanne, line 51: Generic way to populate telemetry with details of the 
* Error (type)
* Python stack trace (without PII, hash of file & folder)

src/client/datascience/context/telemetry.ts
DonJayamanne: Oops, not yet completed
I didn't take the folders from site-packages into account.

src/client/common/cancellation.ts
```diff
@@ -3,15 +3,16 @@
...
 import { CancellationToken, CancellationTokenSource } from 'vscode';
+import { BaseError } from './errors';
 import { createDeferred } from './utils/async';
 import * as localize from './utils/localize';
 
 /**
  * Error type thrown when canceling.
  */
-export class CancellationError extends Error {
+export class CancellationError extends BaseError {
```
DonJayamanne, line 13: Modified all of our Errors to ensure they inherit `BaseError`, 
This way we do'nt need to check each error.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -212,10 +208,13 @@ export class Kernel implements IKernel {
                         }
                     } catch (ex) {
                         traceError('failed to create INotebook in kernel', ex);
-                        sendKernelTelemetryEvent(options.document.uri, Telemetry.NotebookStart, stopWatch.elapsedTime, {
-                            failed: true,
-                            failureReason: getErrorClassification(ex)
-                        });
+                        sendKernelTelemetryEvent(
```
DonJayamanne, line 211: Now we just send the error , no need to send custom stuff.

src/client/common/errors/errorUtils.ts
```diff
@@ -49,14 +33,34 @@ export function getErrorMessageFromPythonTraceback(traceback: string) {
...
+    //             File "/Users/donjayamanne/miniconda3/envs/env3/lib/python3.7/site-packages/appnope/_nope.py", line 38, in C
+
+    const lastFrame = traceback
+        .split('\n')
+        .map((item) => item.trim().toLowerCase())
+        .filter((item) => item.length)
+        .reverse()
+        .find(
+            (line) =>
+                line.startsWith('file ') && line.includes(', line ') && line.includes('.pi') && line.includes('.pi')
```
IanMatthewHuff, line 51: .pi here looks like it should be .py

src/client/datascience/jupyter/jupyterExecution.ts
```diff
@@ -301,9 +298,7 @@ export class JupyterExecutionBase implements IJupyterExecution {
                                 );
                             }
                         } else {
-                            sendTelemetryEvent(Telemetry.ConnectFailedJupyter, undefined, {
-                                failureReason: getErrorClassification(err)
-                            });
+                            sendTelemetryEvent(Telemetry.ConnectFailedJupyter, undefined, undefined, err);
```
DonJayamanne, line 301: Even existing `sendTelemetryEvent` can just send the error object (existing function argument) and the details of the error will be included.

src/client/datascience/telemetry/telemetry.ts
```diff
@@ -55,36 +44,6 @@ const trackedInfo = new Map<string, [ResourceSpecificTelemetryProperties, Contex
 const currentOSType = getOSType();
 const pythonEnvironmentsByHash = new Map<string, PythonEnvironment>();
 
-export function getErrorClassification(error: Error) {
```
DonJayamanne, line 58: Removed this gian if block, now all errors inherit BaseError & the informatino is in the error object

src/client/common/errors/errorUtils.ts
```diff
@@ -49,14 +33,34 @@ export function getErrorMessageFromPythonTraceback(traceback: string) {
...
+        return;
     }
-    if (error instanceof WrappedError && error.originalException instanceof expectedType) {
-        return true;
+    const file = lastFrame.substring(0, lastFrame.lastIndexOf('.py')) + '.py';
+    const parts = file.replace(/\\/g, '/').split('/');
+    const indexOfSitePackages = parts.indexOf('site-packages');
+    let packageName =
+        indexOfSitePackages >= 0 && parts.length > indexOfSitePackages + 1 ? parts[indexOfSitePackages + 1] : '';
+    const reversedParts = file.replace(/\\/g, '/').split('/').reverse();
```
IanMatthewHuff, line 61: Could just use parts and reverse that instead of replacing and splitting again, though I doubt it's a perf issue.

src/client/common/errors/types.ts
```diff
@@ -0,0 +1,80 @@
...
+    return error instanceof BaseError ? error.category : 'unknown';
+}
+
+export type ErrorCategory =
+    | 'cancelled'
+    | 'timeout'
+    | 'daemon'
+    | 'zmq'
+    | 'debugger'
+    | 'kerneldied'
```
IanMatthewHuff, line 40: kerneldied 2X

src/client/datascience/context/telemetry.ts
DonJayamanne: Done.

src/client/common/errors/index.ts
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+    return parts.join('.');
+}
+
+/**
+ * Analyze the details of the error such as `stdErr` from the kernel process and
+ * try to determine the cause.
+ */
+function getReasonForKernelToDie(stdErr: string) {
```
IanMatthewHuff, line 80: Final comment to make here, and this is more of a suggestion / thought to talk over. Considering if these should not be a single reason, but more like this:

pattern => tag
For all patterns that match, add all tags

Right now there is only one reason, but multiple patterns could match. Some are just searching for text that could be anywhere in the stack, while some are searching for errors at the end. If we start adding more of these ordering in this file could matter. What is something was looking for a specific string in the stacks, but that same stderr ends with a DLL load failed error. We might want to tag it with both dll load failed and the other issue. But right now it would matter which if statement runs first in this file.

rchiodo (approved):

src/client/common/errors/index.ts
DonJayamanne: Fixed, please check

DonJayamanne: & agree, it should be easy to add them (manage)

IanMatthewHuff (approved):

PR #4760, opened 2021-02-11
Pass resource to track Kernel Metadata against a resource

DonJayamanne:
Addresses the comments from here https://github.com/microsoft/vscode-jupyter/pull/4742#discussion_r574124229


src/client/datascience/jupyter/kernels/kernelSelector.ts
```diff
@@ -147,8 +147,7 @@ export class KernelSelector implements IKernelSelectionUsage {
         notebookMetadata?: nbformat.INotebookMetadata,
         disableUI?: boolean,
         cancelToken?: CancellationToken,
-        ignoreDependencyCheck?: boolean,
-        ignoreTrackingKernelInformation?: boolean
```
DonJayamanne, line 151: Removed the hacks & assumptions.
Just because some of these methods were called doesn't mean the kernel was changed.
We now track the kernel against the URI where these methods are called & kernel information is actually changed/set. No assumptions.

src/client/datascience/telemetry/interpreterPackages.ts
```diff
@@ -86,9 +86,7 @@ export class InterpreterPackages {
         const packageAndVersions = new Map<string, string>();
         // Add defaults.
         interestedPackages.forEach((item) => {
-            if (!packageAndVersions.has(item)) {
-                packageAndVersions.set(item, 'NOT INSTALLED');
```
DonJayamanne, line 90: Should have been hashed (no harm, it was a noop)

src/client/datascience/commands/notebookCommands.ts
```diff
@@ -74,11 +74,15 @@ export class NotebookCommands implements IDisposable {
...
+            const connection = await this.notebookProvider.connect({
+                getOnly: false,
+                disableUI: false,
+                resource: options.resource
+            });
 
             // Select a new kernel using the connection information
             const kernel = await this.kernelSelector.selectJupyterKernel(
-                options.identity,
+                options.resource,
```
DonJayamanne, line 85: Bug fix.

src/client/datascience/ipywidgets/ipyWidgetScriptSource.ts
```diff
@@ -183,6 +183,7 @@ export class IPyWidgetScriptSource implements ILocalResourceUriConverter {
         if (!this.notebook) {
             this.notebook = await this.notebookProvider.getOrCreateNotebook({
                 identity: this.identity,
+                resource: this.identity,
```
joyceerhl, line 186: Why is `resource` the same as `identity`? Or does it not matter in this context?

DonJayamanne: Doesn't matter in this context (basically use the best we have, majority of the time `identify = resoruce`, except in interactive window when you open it without a correspoding py file or when you have multiples.

src/client/datascience/jupyter/serverPreload.ts
```diff
@@ -59,13 +59,18 @@ export class ServerPreload implements IExtensionSingleActivationService {
...
+                disableUI: true,
+                resource: undefined
+            });
 
             // If it didn't start, attempt for local and if allowed.
             if (!providerConnection && !this.configService.getSettings(undefined).disableJupyterAutoStart) {
                 // Local case, try creating one
                 providerConnection = await this.notebookProvider.connect({
                     getOnly: false,
+                    resource: undefined,
```
IanMatthewHuff, line 73: Does this then get updated later? Just making sure that if a server is preloaded that the undefined here was going to stay even when we start up the actual notebook later when opening the file.

IanMatthewHuff: ChangeKernel should trigger the update it looks like?

DonJayamanne: When we start the actual notebook, then we'd be connecting to some kernel, hence at that point with `Jupyter` we'd change the kernel.
And with Raw, we'd end up starting a different kernel.

DonJayamanne: I.e. passing `undefined` is ok for auto start.

IanMatthewHuff (approved):

Issue #4761, opened 2021-02-11
Option to show the native variable view when not in the native editor experiment.


IanMatthewHuff:
Command to show native variable view is available when not in the native experiment. View tries to open and just loads forever. It should not be available here.

rchiodo:
Validated.

PR #4762, opened 2021-02-11
Add new editor context for just if native notebook is active

IanMatthewHuff:
For #4761 

### DIFF
--- a/package.json
+++ b/package.json
@@ -728,7 +728,7 @@
                     "dark": "resources/dark/variable_explorer.svg"
                 },
                 "category": "Jupyter",
-                "enablement": "jupyter.isvscodenotebookactive && !jupyter.variableViewVisible"
+                "enablement": "notebookViewType == jupyter-notebook && !jupyter.variableViewVisible"
             }
         ],
         "menus": {
@@ -1294,7 +1294,7 @@
                     "command": "jupyter.openVariableView",
                     "title": "%jupyter.command.jupyter.openVariableView.title%",
                     "category": "Jupyter",
-                    "when": "jupyter.isvscodenotebookactive"
+                    "when": "notebookViewType == jupyter-notebook"
                 },
                 {
                     "command": "jupyter.selectNativeJupyterUriFromToolBar",
--- a/package.json
+++ b/package.json
@@ -728,7 +728,7 @@
                     "dark": "resources/dark/variable_explorer.svg"
                 },
                 "category": "Jupyter",
-                "enablement": "!jupyter.variableViewVisible"
+                "enablement": "jupyter.isvscodenotebookactive && !jupyter.variableViewVisible"
             }
         ],
         "menus": {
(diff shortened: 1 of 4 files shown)


IanMatthewHuff:
Speaking mostly to address @rchiodo 's question from triage. I'm trying to be careful here with what I change for limited rollout. All the testing for variable view so far have been done with jupyter.isnativeactive. I had assumed (bad on me for not checking 👎 ) that this was for the native editor active, but it's actually native == any type of notebook.

I don't believe that this change here is necessarily the best long term solution. But I do think that it's the right thing for a Tuesday release. I don't have to worry about testing new states where the variable view is opened, but nothing is active or interactive window is active or things like that. Basically just adds a new state to make sure that it's both active notebook editor + active editor is vscode notebooks.

GA release we can make considerations for something like:
1. Show if there is a native notebook document open at all
2. Show anytime the extension is activated

But I wouldn't not want to implement either of those now, as they would open up new untested contexts for the variable view.

rchiodo (approved):

package.json
```diff
@@ -728,7 +728,7 @@
                     "dark": "resources/dark/variable_explorer.svg"
                 },
                 "category": "Jupyter",
-                "enablement": "!jupyter.variableViewVisible"
+                "enablement": "jupyter.isvscodenotebookactive && !jupyter.variableViewVisible"
```
DonJayamanne, line 731: You can use this instead `notebookViewType == jupyter-notebook`
I.e. we don't need a new context variable.

DonJayamanne (changes requested):
We can reuse existing VS Code context variable `notebookViewType == jupyter-notebook`

package.json
IanMatthewHuff: @DonJayamanne not sure if you have a thought on this. But this doesn't seem to be true. In particular on the view enablement notebookViewType == jupyter-notebook doesn't work and the view never shows.

IanMatthewHuff: VSCode bug?

package.json
```diff
@@ -1839,7 +1839,7 @@
                     "type": "webview",
                     "id": "jupyterViewVariables",
                     "name": "Variables",
-                    "when": "jupyter.isnativeactive"
+                    "when": "jupyter.isvscodenotebookactive"
```
IanMatthewHuff, line 1842: @DonJayamanne doesn't seem like I can get rid of this. Tried every variation of notebookViewType == jupyter-notebook here in this when, it never works and the view is never enabled.

IanMatthewHuff: Same enablement does work fine for the commands, so I changed them.

IanMatthewHuff: @DonJayamanne Sorry to bug, but can you take another look here? Would like to get this in and ported to release so that we can test in release.

package.json
DonJayamanne: Yeah, I'd say its a VS Code bug, could be related to the same issue where toolbar icons do not appear until you click on the notebook.

DonJayamanne (approved):

Issue #4771, opened 2021-02-11
.NET extension needs to open a native notebook editor


rchiodo:
If the .NET extension is installed, in order for their kernels to be used, the new editor must be used.

joyceerhl:
Validated. I installed the .NET extension in VS Code insiders, ran their create blank notebook command with our latest debug bits, and got the native notebooks UI with the .NET interactive kernel selected.

PR #4772, opened 2021-02-12
Fix .NET interactive to always use native

rchiodo:
For #4771 

This is to make sure that the new .NET interactive extension can open notebooks and run their kernel. In this situation, the old webview cannot run .NET (as there's no jupyter kernel installed)

### DIFF
--- a/src/client/common/experiments/service.ts
+++ b/src/client/common/experiments/service.ts
@@ -14,6 +14,7 @@ import {
     GLOBAL_MEMENTO,
     IConfigurationService,
     IExperimentService,
+    IExtensions,
     IJupyterSettings,
     IMemento,
     IOutputChannel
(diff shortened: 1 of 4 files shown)


src/client/common/experiments/service.ts
```diff
@@ -95,6 +97,15 @@ export class ExperimentService implements IExperimentService {
             return false;
         }
 
+        // If user has .NET interactive installed, we HAVE to be in the native experiment. See this issue:
+        // https://github.com/microsoft/vscode-jupyter/issues/4771
+        if (
```
IanMatthewHuff, line 102: Are the PMs ok with figuring out this via telemetry somehow? I know they were ok with that for AML.

rchiodo: Yeah we had a meeting with .NET (Claudia was there) and she's okay with using telemetry to figure out retention.

IanMatthewHuff (approved):

DavidKutu (approved):

Issue #4786, opened 2021-02-12
Cannot correctly run a notebook document after you have moved panes


IanMatthewHuff:
I believe there are may ways to repro this, but here is the specific repro that I was using.

1. Start VS Code with no documents open
2. Open Document A, run a cell
3. Open Document B
4. Drag Document B side by side with Document A
5. Try to run a cell in Document B

Cell will spin forever and never start.

DavidKutu:
Validated

PR #4787, opened 2021-02-12
Only clean up editors on close not on Panel dispose

IanMatthewHuff:
For #4786 

### DIFF
--- a/src/client/datascience/notebook/notebookEditor.ts
+++ b/src/client/datascience/notebook/notebookEditor.ts
@@ -97,7 +97,7 @@ export class NotebookEditor implements INotebookEditor {
                 }
             })
         );
-        disposables.push(model.onDidDispose(this._closed.fire.bind(this._closed, this)));
+        disposables.push(model.onDidDispose(this.dispose.bind(this)));
     }
     @captureTelemetry(Telemetry.SyncAllCells)
     public async syncAllCells(): Promise<void> {
--- a/src/client/datascience/notebook/notebookEditorProvider.ts
+++ b/src/client/datascience/notebook/notebookEditorProvider.ts
@@ -184,7 +184,6 @@ export class NotebookEditorProvider implements INotebookEditorProvider {
             return;
         }
         this.trackedVSCodeNotebookEditors.add(editor);
-        this.disposables.push(editor.onDidDispose(() => this.onDidDisposeVSCodeNotebookEditor(editor)));
     }
     private async onDidCloseNotebookDocument(document: NotebookDocument) {
         this.disposeResourceRelatedToNotebookEditor(document.uri);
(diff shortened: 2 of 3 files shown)


src/client/datascience/notebook/notebookEditorProvider.ts
```diff
@@ -184,7 +184,6 @@ export class NotebookEditorProvider implements INotebookEditorProvider {
             return;
         }
         this.trackedVSCodeNotebookEditors.add(editor);
-        this.disposables.push(editor.onDidDispose(() => this.onDidDisposeVSCodeNotebookEditor(editor)));
```
IanMatthewHuff, line 187: I'm trying to make sure that this isn't going to open up cases where we don't dispose. But this was causing some large issues. We remove editors from the editor list either on OnClose or OnDidDispose. But OnDidDispose as per the API is just for the Panel being disposed. This is triggered by something like dragging a notebook editor into a new panel. So just by dragging to a new panel we'll clean up the document and remove it from the editors list. But OnClose / OnOpen is not called on a drag operation like that.

Currently I don't see any cases where we would not get the OnClose event and we just do our cleanup then like we normally would. onDidCloseNotebookDocument calls the same disposeResourceRelatedToNotebookEditor that this function did.

Also this was only on changing the active document. I would think that if we needed to actually catch all disposes than this would need to be onDidOpenNotebookDocument.

rchiodo (approved):

rchiodo:
Good catch. Must have been difficult to find.

DonJayamanne (approved):

joyceerhl (approved):

src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -97,7 +97,7 @@ export class NotebookEditor implements INotebookEditor {
                 }
             })
         );
-        disposables.push(model.onDidDispose(this._closed.fire.bind(this._closed, this)));
+        disposables.push(model.onDidDispose(this.dispose.bind(this)));
```
IanMatthewHuff, line 100: This is the change that I made after checking out how this works shutting down VS. When you shut down VS you do get a dispose on the notebook editor, but you do not get a close. In the notebookEditorProvider the disposeResourceRelatedToNotebookEditor is now only called onClose, not on dispose (since dispose is happening more often, such as moving panes).

disposeResourceRelatedToNotebookEditor basically does the following:
run closeEditor in notebookEditorProvider
dispose editor in notebookEditor
dispose model in baseModel

In the shutdown case the model is disposed via the notebookStorageProvider which puts all models on the dispose queue. In the notebookEditorProvider onEditorOpened we hook up to run the closedEditor function when the editor is closed. And the notebookEditor is set to call the .closed event when the model is disposed. Meaning that on shutdown the following happens:

dispose model in baseModel
disposing model causes closed to file in notebookEditor
run closeEditor in notebookEditorProvider

So basically the exact same things happen here. However I did note that notebookEditor dispose just fires the closed event. But it only hooks up firing the closed event to when the model is disposed not the full dispose. What I have changed this to now is functionally the same, since the only thing that dispose is doing is firing the closed event. But I think this makes it a bit safer in the future if we want to add more to the dispose here.

IanMatthewHuff: @DonJayamanne @rchiodo @joyceerhl Made a tweak on the review. Given that I'm planning on porting this to release wanted to triple check on signoff on it.

rchiodo (approved):

DavidKutu (approved):

IanMatthewHuff:
Test failures due to insiders API update.

PR #4789, opened 2021-02-12
Publish insiders build every day

rchiodo:
This should publish an insiders tied version to the VS code market place every day

### DIFF
--- a/build/ci/scripts/updateEngineToInsiders.js
+++ b/build/ci/scripts/updateEngineToInsiders.js
@@ -1,0 +1,8 @@
+// This file will modify the package.json to have the engine equal to insiders
+var fs = require('fs');
+var packageJson = JSON.parse(fs.readFileSync('./tmp/extension/package.json'));
+
+if (!packageJson.engines.vscode.includes('insider')) {
+    packageJson.engines.vscode = `${packageJson.engines.vscode}-insider`;
+    fs.writeFileSync('./tmp/extension/package.json', JSON.stringify(packageJson, null, 4));
+}
--- a/.github/workflows/publish-insiders.yml
+++ b/.github/workflows/publish-insiders.yml
@@ -1,0 +1,69 @@
+name: Publish For Insiders
+
+on:
+  schedule:
+    # Every day at midnight
+    - cron:  '0 0 * * *'
+  workflow_dispatch:
+    inputs:
+      confirmation:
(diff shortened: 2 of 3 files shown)


build/ci/scripts/updateEngineToInsiders.js
```diff
@@ -0,0 +1,8 @@
+// This file will modify the package.json to have the engine equal to insiders
+var fs = require('fs');
+var packageJson = JSON.parse(fs.readFileSync('./tmp/extension/package.json'));
+
+if (!packageJson.engines.vscode.includes('insider')) {
+    packageJson.engines.vscode = `${packageJson.engines.vscode}-insider`;
```
joyceerhl, line 6: Do we need to account for incrementing semver here?

IanMatthewHuff (approved):
@joyceerhl had a relevant question on versions, but aside from that looks fine to me. So approving if engine versioning doesn't need to change.

build/ci/scripts/updateEngineToInsiders.js
rchiodo: I don't think so. As long is it says '-insider' it will work because ours is set like so:

```
`^1.53.0-insider`
```

That should mean insiders build greater than or equal to 1.53

Wish I could test this though. Going to have to wait till it goes up to the marketplace to make sure it worked. I tested the package.json changes, but that has no effect if you install through a vsix.

rchiodo: Here's the relevant docs on it:
https://code.visualstudio.com/api/working-with-extensions/publishing-extension#visual-studio-code-compatibility

joyceerhl: I ask because the stable semver trails the insiders semver i.e. 1.52 stable and 1.53 insider exist at the same time, so wasn't sure if we wanted to make sure users were on the latest VS Code insiders

rchiodo: Oh you mean always change the semver to the latest? I don't think we should do that until we actually change our usage of the API in the code. At that point we'd change it for stable too. 

So when we make changes for the new notebook API, we'd change the main branch to be ```1.54.0``` which hasn't even shipped in stable yet.

joyceerhl (approved):

PR #4792, opened 2021-02-12
Fix tests in release

rchiodo:
Change tests in release to use the stable version of VS code.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,9 @@ on:
   pull_request:
     branches:
       - main
+      - 'release'
+      - 'release/*'
+      - 'release-*'
   check_run:
     types: [rerequested, requested_action]
   push:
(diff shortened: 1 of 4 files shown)


DonJayamanne (approved):

.github/workflows/main.yml
```diff
@@ -728,7 +728,7 @@ jobs:
           VSC_JUPYTER_REMOTE_NATIVE_TEST: ${{ contains(matrix.test-suite, 'remote') }}
           VSC_JUPYTER_NON_RAW_NATIVE_TEST: ${{ contains(matrix.test-suite, 'nonRawNotebook') }}
           VSC_JUPYTER_CI_RUN_JAVA_NB_TEST: ${{ contains(matrix.test-suite, 'conda') }}
-          VSC_JUPYTER_CI_TEST_VSC_CHANNEL: 'insiders'
+          VSC_JUPYTER_CI_TEST_VSC_CHANNEL: 'stable'
```
joyceerhl, line 731: I think there are a number of places in the tests where we check if the channel is insiders, and we don't even run them if not. Do we need to update all of those as well? I worry the tests are green now because they're all being skipped.

rchiodo: It looks to me like they'd run anyway. There's only one spot that I can find:

```
src\test\initialize.ts
```

In the closeAllNotebooks. However the backup of closeAllEditors would work. Seems like for that spot if it wasn't working, the tests would be failing.

Oh wait, the checks all passed because this branch doesn't have the main.yml running on pull_request. Whoops. Let me change that. Good thought on checking for skipping.

Issue #4726, opened 2021-02-09
Native Notebooks - Survey link for stable experiment


claudiaregio:
Survey link for stable experiment to be rolled out as same cadence as our other surveys. 

Assignees: Ping for survey link

IanMatthewHuff:
Validated. Got the survey with the new extension on stable.

PR #4794, opened 2021-02-12
Refactor surveys and add a new one

DavidKutu:
For #4726


src/client/datascience/dataScienceSurveyBanner.ts
```diff
@@ -87,102 +111,235 @@ export type ShowBannerWithExpiryTime = {
...
+
     private disabledInCurrentSession: boolean = false;
-    private bannerMessage: string = localize.DataScienceSurveyBanner.bannerMessage();
     private bannerLabels: string[] = [
         localize.DataScienceSurveyBanner.bannerLabelYes(),
         localize.DataScienceSurveyBanner.bannerLabelNo()
     ];
-    private readonly showBannerState: IPersistentState<ShowBannerWithExpiryTime>;
-    private readonly surveyLink: string;
+    private readonly showBannerState: IPersistentState<ShowBannerWithExpiryTime>[] = [];
```
rchiodo, line 154: Seems like this should be a map, not an array. It would reenforce that it is being set correctly.

```suggestion
private readonly showBannerState = new Map<BannerType, IPersistentState<ShowBannerWithExpiryTime>>();
```

src/client/datascience/dataScienceSurveyBanner.ts
```diff
@@ -87,102 +111,235 @@ export type ShowBannerWithExpiryTime = {
...
+                state = this.persistentState.createGlobalPersistentState<number>(DSSurveyStateKeys.ExecutionCount, 0);
+                return state.value;
+        }
     }
 
     private async openedNotebook() {
-        const state = this.persistentState.createGlobalPersistentState<number>(DSSurveyStateKeys.OpenNotebookCount, 0);
-        await state.updateValue(state.value + 1);
-        return this.showBanner();
+        const state1 = this.persistentState.createGlobalPersistentState<number>(DSSurveyStateKeys.OpenNotebookCount, 0);
```
rchiodo, line 285: This seems like you could call the same function with a different string 3 times to update the state for all 3.

```
private updateStateAndShowBanner(val: string, banner: BannerType) {
    const state = this.persistentState.createGlobalPersistentState<number>(val, 0);
    await state.updateValue(state.value + 1);
    this.showBanner(banner);
}

File: src/client/datascience/dataScienceSurveyBanner.ts [hunk 3]
```diff
@@ -87,102 +111,235 @@ export type ShowBannerWithExpiryTime = {
...
+    }
+
+    private async kernelStateChanged(kernelStateEvent: KernelStateEventArgs) {
+        // TODO Only on insiders native notebook survey FOR NOW
+        if (kernelStateEvent.state === KernelState.executed) {
+            const state1 = this.persistentState.createGlobalPersistentState<number>(
+                InsidersNotebookSurveyStateKeys.ExecutionCount,
+                0
+            );
+            await state1.updateValue(state1.value + 1);
```
rchiodo, line 311: That function I mentioned above would be called here.

src/client/datascience/dataScienceSurveyBanner.ts
```diff
@@ -87,102 +111,235 @@ export type ShowBannerWithExpiryTime = {
...
+            case BannerType.ExperimentNotebookSurvey:
+                if (
+                    this.applicationEnvironment.channel === 'stable' &&
+                    this.experimentService.inExperiment(Experiments.NativeNotebook)
+                ) {
+                    return this.isEnabledInternal(type);
+                }
+                break;
+            case BannerType.DSSurvey:
+            default:
```
IanMatthewHuff, line 131: Just a minor thing here. I'd break off default here and have it traceError and return false. isEnabled should always be called with a BannerType, so if you want the DSSurvey you should specifically request it. If a new BannerType is added we want someone to explicitly add a new case here, now with default it would go down the DSSurvey path if you forgot to add the case, which you don't want.

DavidKutu: just isEnabled? or all of the switches?

src/client/datascience/dataScienceSurveyBanner.ts
DavidKutu: yes, I'll also do that for getOpenNotebookCount and getExecutionCount

src/client/datascience/dataScienceSurveyBanner.ts
IanMatthewHuff: Yeah, you don't want a default like that here. Only if it has a legit default that you want to happen.

IanMatthewHuff (approved):

rchiodo (changes requested):
⏲️

rchiodo (approved):

PR #4805, opened 2021-02-13
Restyle data viewer and add slice data panel

joyceerhl:
image


joyceerhl:
Should probably give a courtesy heads up to the AML extension team that we're changing the dataset preview UI.

src/datascience-ui/react-common/seti/seti.less
```diff
@@ -0,0 +1,656 @@
+@font-face {
```
rchiodo, line 1: Did you copy this from somewhere? If so we need to add a license for it to the ThirdPartyRespository.txt

package.json
```diff
@@ -1914,6 +1914,7 @@
     },
     "dependencies": {
         "@enonic/fnv-plus": "^1.3.0",
+        "@fluentui/react": "^7.160.1",
```
rchiodo, line 1917: Are you actually using this? I thought you dropped it?

rchiodo (approved):

package.json
joyceerhl: I tried to use this wherever I could, so yes I am using this in places where it was easier to reuse + override styles than it was to write my own.

DonJayamanne: yay for using fluentui/react.... its awesome..

src/client/datascience/data-viewing/types.ts
```diff
@@ -68,6 +68,8 @@ export interface IDataFrameInfo {
     maximumRowChunkSize?: number;
     type?: string;
     originalVariableType?: string;
+    name?: string;
+    fileName?: string;
```
DonJayamanne, line 72: Would be nice to have come comments. What's `FileName` here?
The rest seem to make sense, as they are associated with a variable.

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -132,13 +133,14 @@ export class DebuggerVariables extends DebugLocationTracker
...
-
+        const fileName = this.debugLocation?.fileName ? path.basename(this.debugLocation?.fileName) : undefined;
         // Results should be the updated variable.
         return results
             ? {
                   ...targetVariable,
                   ...JSON.parse(results.result),
-                  maximumRowChunkSize: MaximumRowChunkSizeForDebugger
+                  maximumRowChunkSize: MaximumRowChunkSizeForDebugger,
+                  fileName
```
DonJayamanne, line 143: Isn't this the name of the file where the frame is. Not sure `fileName` is meaningful enough.
Initially i thought it could be the file path for a data frame (source where it was loaded from).
After looking at this code I now konw this is the file path from the frame.

DonJayamanne: I.e. i'd consider re-naming the property (still not clear why we need this & why its associated with the variable - hopefully i'll see this being used later in the code).

src/datascience-ui/data-explorer/mainPanel.tsx
```diff
@@ -148,11 +154,36 @@ export class MainPanel extends React.Component<IMainPanelProps, IMainPanelState>
...
+        if (this.state.fileName) {
+            let breadcrumbText = this.state.variableName;
+            if (this.state.originalVariableShape) {
+                breadcrumbText += ' (' + this.state.originalVariableShape?.join(', ') + ')';
+            }
+            return (
+                <div className="breadcrumb-container control-container">
+                    <div className="breadcrumb">
+                        <div className="icon-python breadcrumb-file-icon" />
+                        <span>{this.state.fileName}</span>
```
DonJayamanne, line 173: Looking at the screen sample, i can't see how or where this is displayed.
Not sure I'd tie the file name to a variable either.

src/datascience-ui/data-explorer/mainPanel.tsx
```diff
@@ -148,11 +154,36 @@ export class MainPanel extends React.Component<IMainPanelProps, IMainPanelState>
...
     };
 
+    private renderBreadcrumb() {
+        if (this.state.fileName) {
+            let breadcrumbText = this.state.variableName;
+            if (this.state.originalVariableShape) {
+                breadcrumbText += ' (' + this.state.originalVariableShape?.join(', ') + ')';
+            }
+            return (
+                <div className="breadcrumb-container control-container">
```
DonJayamanne, line 170: I'd prefer a separate react tsx file for this, perhaps being too much of a purist here, but generally in react, these are treated and craeted as separate components.
FYI - Not a required change.

src/datascience-ui/data-explorer/mainPanel.tsx
```diff
@@ -233,6 +264,8 @@ export class MainPanel extends React.Component<IMainPanelProps, IMainPanelState>
                 const indexColumn = variable.indexColumn ? variable.indexColumn : 'index';
                 const originalVariableType = this.state.originalVariableType ?? variable.type;
                 const originalVariableShape = this.state.originalVariableShape ?? variable.shape;
+                const variableName = this.state.variableName ?? variable.name;
+                const fileName = this.state.fileName ?? variable.fileName;
```
DonJayamanne, line 268: Yeah, its weird for a variable to have a file name. I'd prefer this to be sent as a separate entity.

src/datascience-ui/data-explorer/reactSlickGridFilterBox.tsx
```diff
@@ -3,11 +3,26 @@
 'use strict';
 
 import * as React from 'react';
+import { IIconProps, initializeIcons, SearchBox } from '@fluentui/react';
+
+initializeIcons();
```
DonJayamanne, line 8: I thought this needs to be done one place, if thats the case then I'd move it to the entry point of our UI (so that we dont' have multiples of these)

src/datascience-ui/data-explorer/reactSlickGridFilterBox.tsx
```diff
@@ -18,20 +33,27 @@ export class ReactSlickGridFilterBox extends React.Component<IFilterProps> {
...
         return (
-            <input
-                type="text"
+            <SearchBox
+                iconProps={filterIcon}
+                onChange={this.updateInputValue}
+                onClear={this.clearInputValue}
                 tabIndex={0}
-                aria-label={this.props.column.name}
+                ariaLabel={this.props.column.name}
```
DonJayamanne, line 41: Interesting our linters didn't catch this bug.
How did you come across this?

DonJayamanne: I.e. the change from `aria-label` to `ariaLabel`

src/test/datascience/mountedWebView.ts
```diff
@@ -59,6 +59,9 @@ export class MountedWebView implements IMountedWebView, IDisposable {
     private visible = true;
     private disposedEvent = new EventEmitter<void>();
     private loadFailedEmitter = new EventEmitter<void>();
+    public get viewColumn() {
+        return 1;
```
DonJayamanne, line 63: I'd use the VS Code enums instead of `1`

DonJayamanne (changes requested):
Everything looks ok, except for the `fileName`, not sure it should be part of the variable entity.
If anything I'd try to create a separate object that contains those two such as or similarl/better...
```typescript
type StackVariable = {
	variabel: IDataVariable;
	stackFileName: string;
}
```

src/datascience-ui/data-explorer/mainPanel.tsx
joyceerhl: It's at the very top of the screenshot. Basically Jill wants us to display breadcrumbs in the data viewer webview showing which file the variable was declared in, the same way that text editors in VS Code have breadcrumbs.

src/datascience-ui/data-explorer/reactSlickGridFilterBox.tsx
joyceerhl: This file was previously ignored in our eslintrc.js, but this change was made by eslint after I removed this file from the ignore list

src/datascience-ui/data-explorer/mainPanel.tsx
joyceerhl: Is it really that weird though? The variable must've been declared somewhere. I think it should definitely be renamed but it doesn't seem that weird to me to have `backingFileName` or similar as an attribute associated with the variable info.

src/datascience-ui/data-explorer/mainPanel.tsx
joyceerhl: Yeah I didn't think it was necessary because the breadcrumb render code is just 20 lines of code.

src/datascience-ui/react-common/seti/seti.less
joyceerhl: It's the seti icons that VS Code ships. Needed this for the language icon in the breadcrumbs at the top of the data viewer. This is the one that VS Code has: https://github.com/microsoft/vscode/blob/main/extensions/theme-seti/ThirdPartyNotices.txt

src/test/datascience/mountedWebView.ts
joyceerhl: This is test infra and may not be running in vscode, so I thought it would be better to use `1` here.

src/datascience-ui/react-common/seti/seti.less
joyceerhl: Do we add the license manually, or is there a component governance autogeneration step somewhere? And do we need to add it to both ThirdPartyNotices-Distribution as well as ThirdPartyNotices-Repository?

src/datascience-ui/data-explorer/mainPanel.tsx
DonJayamanne: It is weird, if you look at the debugger prototcol or others, variables are not attached to a file name. We have a file name that can contains a context & that context then contains a list of variables & that list contains the individual variable.
I.e. the mapping is at a much higher level & this seem wrong to me.
But don't want to block this PR for that.

I.e. the request object (sending request to kernel for variables) should already know the file name that is active.
Also if we're displaying variables for Notebooks, would we display the file name as some internal `python file` or the name of the notebook? Today run by line works in webview, if we use run by line & display the variables, what do you see in the file name (breadcrumb)?

DonJayamanne (approved):

src/datascience-ui/data-explorer/mainPanel.tsx
DonJayamanne: Not worth blocking PR

DonJayamanne: But i'd test this with run by line & the others, not sure the name of the file would be as expected/required. I.e. in notebooks I'd always expect to see the name of the notebook if anything.

joyceerhl: The request we send to the kernel doesn't ask for the filename. The filename property is only included on the variable at the point when we send the variable info to the UI for display.

For webview notebooks the file name is the name of the notebook, e.g.
image. This is taken from `notebook.identity.path`.

For RBL I initially went with `debugLocation` because `notebook.identity` is not available when debugging a Python file, but this doesn't work for RBL, the filename would've been '<ipython input>'. Changed debuggerVariables.ts to use the notebook.identity.path when it's available (which it is for Run By Line), and use the debugLocation when it's not (i.e. debugging a Python file). Good catch :) 

image

src/datascience-ui/react-common/seti/seti.less
joyceerhl: Updated Repository notice and manually added this dependency to AzDO CG

src/datascience-ui/data-explorer/reactSlickGridFilterBox.tsx
joyceerhl: Moved to index.tsx, thanks!

joyceerhl:
Made a couple more changes and fixes. Main difference is we are now generating a numeric (as opposed to guid) index column in the UI (as opposed to in our Python scripts) for all data, even if the data already contains an index column. This is to avoid breaking AML with the new filter button changes. AML preview datasets now look like this:

image

PR #4820, opened 2021-02-16
Fix typo in publish command

joyceerhl:
https://github.com/microsoft/vscode-jupyter/actions/runs/570042037 Publishes are failing because of a trailing underscore in the publish command

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -52,7 +52,7 @@ jobs:
         run: az storage blob download --file ${{env.VSIX_NAME_TO_PUBLISH}} --account-name pvsc --container-name ${{env.BLOB_CONTAINER_NAME}} --name ${{ env.VSIX_NAME_UPLOADED_TO_BLOB }}
 
       - name: Publish
-        run: vsce_ publish --packagePath ${{env.VSIX_NAME_TO_PUBLISH}} --pat ${{secrets.VSCE_TOKEN}} --noVerify
+        run: vsce publish --packagePath ${{env.VSIX_NAME_TO_PUBLISH}} --pat ${{secrets.VSCE_TOKEN}} --noVerify
 
       - name: Extract Extension
         shell: bash


IanMatthewHuff (approved):

.github/workflows/publish-release.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         run: az storage blob download --file ${{env.VSIX_NAME_TO_PUBLISH}} --account-name pvsc --container-name ${{env.BLOB_CONTAINER_NAME}} --name ${{ env.VSIX_NAME_UPLOADED_TO_BLOB }}
 
       - name: Publish
-        run: vsce_ publish --packagePath ${{env.VSIX_NAME_TO_PUBLISH}} --pat ${{secrets.VSCE_TOKEN}} --noVerify
+        run: vsce publish --packagePath ${{env.VSIX_NAME_TO_PUBLISH}} --pat ${{secrets.VSCE_TOKEN}} --noVerify
```
DonJayamanne, line 55: Not sure we want this yet. Are we planning on publishing automatically from the pipelines?
This command line will publish directly to the marketplace.
This used to be done when we were in Azure Pipelines, when we could manually approve this stage.
However in github my understanding is that there's no approval stage, hence any commit pushed to releaes will end up pushing a new VSIX to the marketplace.

DonJayamanne (changes requested):
Not sure this is what we want, might want to leave this for now until we sort out how we automate releases

.github/workflows/publish-release.yml
joyceerhl: Isn't this what we wanted with https://github.com/microsoft/vscode-jupyter/pull/4789 ?

rchiodo:
I believe this is what we want. At least Jim agreed to it. However the release bits shouldn't update this way. I didn't notice this PR and fixed the same thing in https://github.com/microsoft/vscode-jupyter/pull/4823

DonJayamanne:
> However the release bits shouldn't update this way


Provided we have a process for publishing release bits, I'm fine. Doesn't sound like we have a process outlined for release yet.

PR #4823, opened 2021-02-16
Fix publishing scripts

rchiodo:
```vsce_``` instead of ```vsce``` was there and I didn't notice it. 

Also made sure to not auto publish the release bits.

### DIFF
--- a/.github/workflows/publish-insiders.yml
+++ b/.github/workflows/publish-insiders.yml
@@ -1,9 +1,9 @@
 name: Publish For Insiders
 
 on:
-  schedule:
+#  schedule:
     # Every day at midnight
-    - cron:  '0 0 * * *'
+#    - cron:  '0 0 * * *'
   workflow_dispatch:
     inputs:
       confirmation:
--- a/.github/workflows/publish-insiders.yml
+++ b/.github/workflows/publish-insiders.yml
@@ -67,5 +67,5 @@ jobs:
           python -c "import os; os.rename('${{env.VSIX_ROOT_NAME}}.zip', '${{env.VSIX_NAME_TO_PUBLISH}}')"
 
       - name: Publish
-        run: vsce_ publish --packagePath ${{env.VSIX_NAME_TO_PUBLISH}} --pat ${{secrets.VSCE_TOKEN}} --noVerify
+        run: vsce publish --packagePath ${{env.VSIX_NAME_TO_PUBLISH}} --pat ${{secrets.VSCE_TOKEN}} --noVerify
 
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,11 +1,6 @@
 name: Publish Extension
 
 on:
-  push:
-    branches:
-      - 'release'
-      - 'release/*'
-      - 'release-*'
   workflow_dispatch:
     inputs:
       confirmation:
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by DavidKutu — APPROVED** [review]

File: .github/workflows/publish-insiders.yml [hunk 1]
```diff
@@ -1,9 +1,9 @@
 name: Publish For Insiders
 
 on:
-  schedule:
+#  schedule:
     # Every day at midnight
-    - cron:  '0 0 * * *'
+#    - cron:  '0 0 * * *'
```
DonJayamanne, line 6: Do we need `workflow_dispatch`
I added that as a way for manually publishing from CI.
I.e. if you run the CI pipline manually and add the message `true` in response to the message `Publish to VSCode marketplace?` then we publish it (i.e. its still automated on CI - via a click of a button).

Or awe we now saying that we'll always pull the extension VSIX down & upload to marketplace manually?
If thats the case we should remove this dispatch

.github/workflows/publish-release.yml
```diff
@@ -1,11 +1,6 @@
 name: Publish Extension
 
 on:
-  push:
-    branches:
-      - 'release'
-      - 'release/*'
-      - 'release-*'
   workflow_dispatch:
```
DonJayamanne, line 4: Do we need `workflow_dispatch`
I added that as a way for manually publishing from CI.
I.e. if you run the CI pipline manually and add the message `true` in response to the message `Publish to VSCode marketplace?` then we publish it (i.e. its still automated on CI - via a click of a button).

Or awe we now saying that we'll always pull the extension VSIX down & upload to marketplace manually?
If thats the case we should remove this dispatch

DonJayamanne: I'm assuming the answer is yes, else theres' no need to rename the `vsce_` command to `vsce`.
FYI - we might want to ensure the `VSC_TOKEN` has been added to github actions.

DonJayamanne (changes requested):
* Remove `workflow_dispatch` for insiders
	I do'nt think we need it, as we're publishing always.
* Waiting for response on comments.

.github/workflows/publish-release.yml
rchiodo: VSC_TOKEN has been added. 

Not sure what you're saying otherwise. The workflow_dispatch would be the 'manual' upload. You just click on the github action to do it.

rchiodo:
> * Remove `workflow_dispatch` for insiders
>   I do'nt think we need it, as we're publishing always.
> * Waiting for response on comments.

Not sure how we're publishing always? The timer is turned off. Jim didn't want the auto publish so I left the workflow_dispatch trigger there.

.github/workflows/publish-insiders.yml
rchiodo: I disagree. I'd like to dispatch from the github action

.github/workflows/publish-release.yml
IanMatthewHuff: Seems fine to me to have the option for workflow_dispatch. Would allow us to pop out a specific build early if we ever wanted to.

IanMatthewHuff (approved):

rchiodo:
@DonJayamanne are you still asking for changes?

Issue #4833, opened 2021-02-16
Interactive Python file commands are missing from command palette


joyceerhl:
Run Below/Run Cell commands are missing from command palette

rchiodo:
I don't repro this?

![image](https://user-images.githubusercontent.com/19672699/108129979-b8d56c80-7063-11eb-9fac-8ae71d8bba47.png)

rchiodo:
Opening a notebook causes this problem.

greazer:
Shouldn't block tomorrow's release. If trivial, let's fix.

joyceerhl:
FWIW we recently changed context key visibility in palette with https://github.com/microsoft/vscode-jupyter/pull/4769/

joyceerhl:
Looked at our source, `jupyter.havenativecells` is set in response to messages from the interactive window. Once you run cells in the IW, the commands reappear.

rchiodo:
This was caused by our latest bits but the fix is going to affect other things. I think we can put this in the point release.

DonJayamanne:
Validated

PR #4834, opened 2021-02-17
Fix command palette for interactive commands

rchiodo:
For #4833 

Root cause is that notebookEditor.activeEditor does not actually update when changing to a python file.

This ripples into a couple of spots though. I don't think we should port this to the release branch.

### DIFF
--- a/src/client/datascience/commands/activeEditorContext.ts
+++ b/src/client/datascience/commands/activeEditorContext.ts
@@ -12,6 +12,7 @@ import { PYTHON_LANGUAGE, UseVSCodeNotebookEditorApi } from '../../common/consta
 import { ContextKey } from '../../common/contextKey';
 import { traceError } from '../../common/logger';
 import { IDisposable, IDisposableRegistry } from '../../common/types';
+import { isNotebookCell } from '../../common/utils/misc';
 import { EditorContexts } from '../constants';
 import { isPythonNotebook } from '../notebook/helpers/helpers';
 import {
@@ -191,7 +192,7 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
         this.updateContextOfActiveNotebookKernel(activeEditor);
     }
     private onDidChangeActiveTextEditor(e?: TextEditor) {
-        this.isPythonFileActive = e?.document.languageId === PYTHON_LANGUAGE;
+        this.isPythonFileActive = e?.document.languageId === PYTHON_LANGUAGE && !isNotebookCell(e.document.uri);
         this.updateNativeNotebookCellContext();
         this.updateMergedContexts();
     }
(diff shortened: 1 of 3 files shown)


src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -189,8 +192,7 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
         this.updateContextOfActiveNotebookKernel(activeEditor);
     }
     private onDidChangeActiveTextEditor(e?: TextEditor) {
-        this.isPythonFileActive =
-            e?.document.languageId === PYTHON_LANGUAGE && !this.notebookEditorProvider.activeEditor;
```
rchiodo, line 193: This is another spot that I believe was incorrect. ```this.notebookEditorProvider.activeEditor``` is always available as long as a notebook was opened.

rchiodo: But this might affect other things that I don't realize. This is why I think we shouldn't port this change.

DavidKutu (approved):

src/client/datascience/commands/activeEditorContext.ts
IanMatthewHuff: I'd have to look a bit deeper myself, but then would the actual fix here be to make notebookEditorProvider.activeEditor undefined if it doesn't actually have editor focus?

rchiodo: Not sure I'd  want to do that in case there was things depending upon it but yes we should probably change that property to use the window.activeNotebookEditor instead.

rchiodo: well come to think of it, if I enter an issue for it, we'll just never fix it. So I'll just change it now.

IanMatthewHuff (approved):

src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -114,9 +115,10 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
         if (!this.inNativeNotebookExperiment) {
             return;
         }
-        this.hasNativeNotebookCells
-            .set((this.notebookEditorProvider.activeEditor?.model?.cellCount || 0) > 0)
-            .ignoreErrors();
+
+        // Separate for debugging.
```
joyceerhl, line 119: I'm not really sure what this comment means.

IanMatthewHuff: Assume that means that rich put this in two steps.
1. Put value into the const
2. Use the const to set the context key
To make it easier to debug by setting a breakpoint on the first of the line. Could be combined into one line a gain. But honestly I find it easier to read as two lines.

joyceerhl: Oh OK, just wasn't sure if it was meant to be committed this way. But if you find it more readable then this is fine the way it is.

DonJayamanne (approved):

Issue #4737, opened 2021-02-10
Update VSCode Notebooks API


DonJayamanne:
All the changes are here (last 4 commits, starting from WIP):
https://github.com/rebornix/vscode-jupyter/commits/rebornix/fixVSCAPI

These changes must be applied to master after we have created a release branch.

* [x] Import from vscode & not from `..../vscode-proposed`
* [x] Update Python extension notebook API
* [x] Update renderers extension

IanMatthewHuff:
Not sure if there is a specific validation on this. But resolving due to notebooks now working in insiders and the API update PRs.

PR #4843, opened 2021-02-17
More Updates to notebook API

DonJayamanne:
For #4737 
More updates

FYI - Tests will still fail, VSC insiders will have the necessary fixes/updates tomorrow 🤞🏾


src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -184,10 +182,10 @@ export class CellExecution {
         await clearCellForExecution(this.editor, this.cell);
         if (!this.isEmptyCodeCell) {
             await chainWithPendingUpdates(this.editor.document, (edit) => {
-                edit.replaceNotebookCellMetadata(this.cell.notebook.uri, this.cell.index, {
-                    ...this.cell.metadata,
-                    runStartTime: new Date().getTime()
-                });
+                const metadata = new NotebookCellMetadata()
```
DonJayamanne, line 185: Concrete classes instead of interfaces in VS Code Notebook API

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -279,18 +277,15 @@ export class CellExecution {
         }
 
         // If there are any errors in the cell, then change status to error.
-        if (hasErrorOutput(this.cell.outputs)) {
+        if (this.cell.outputs.some(isErrorOutput)) {
```
DonJayamanne, line 280: the `hasErrorOutput` method was a single liner now after refactoring, hence removed that one line function.

src/client/datascience/notebook/cellOutputs.ts
```diff
@@ -0,0 +1,379 @@
...
+import { traceWarning } from '../../common/logger';
+import { sendTelemetryEvent } from '../../telemetry';
+import { Telemetry } from '../constants';
+import { CellOutputMimeTypes } from './types';
+
+/**
+ * Metadata we store in VS Code cell output items.
+ * This contains the original metadata from the Jupyter Outputs (nbformat.IOutput).
+ */
+export interface CellOutputMetadata {
```
DonJayamanne, line 17: Created concrete classes for 
* Outputs that display the following jupyter outputs
	* nbformat.IError
	* nbformat.IDisplay
	* nbformat.IUpdateDisplay
	* nbformat.IExecuteReulst
This way its easier to understand the code, else we have functions that cast objects into different types.

* Ensure we support other kernels that add outputs/cells to teh notebook.
E.g. if .NET adds output, we'll need to save them properly in ipynb file. 
Now we save unrecognized outputs as `nbformat.IUnrecognizedOutput`

* I understand we do'nt want classes in the same file, but i kept themn here as they are small & there's a lot of code that gets re-used.
Happy to split into separate files.

src/client/datascience/notebook/cellOutputs.ts
```diff
@@ -0,0 +1,379 @@
...
+    static fromIOutput(output: nbformat.IError) {
+        return new NotebookCellOutputError([
+            new NotebookCellOutputItem(CellOutputMimeTypes.error, {
+                ename: output.ename,
+                evalue: output.evalue,
+                traceback: output.traceback
+            })
+        ]);
+    }
+    public toIOutput(): nbformat.IError {
```
DonJayamanne, line 68: More strongly typed & translation functions are in one place.
This way its easier to manage.
Previously they were scattered. Now we know exactly what metadata goes in & what comes out.

src/client/datascience/notebook/cellOutputs.ts
```diff
@@ -0,0 +1,379 @@
...
+abstract class BaseNotebookCellOutput<T extends nbformat.OutputType> extends NotebookCellOutput {
+    public get cellOutputMetadata() {
+        return this.metadata as CellOutputMetadata | undefined;
+    }
+    constructor(
+        outputs: NotebookCellOutputItem[],
+        public readonly outputType: T,
+        cellOutputMetadata?: CellOutputMetadata
+    ) {
+        super(outputs, cellOutputMetadata);
```
DonJayamanne, line 51: This is new VSC API, they now allow us to store metadta against the output, previous we could only do this in the output item (and we had to resort to using private map to support `transient.display_id`)..


Now Document, cells, outupts, output items - everything in notebooks support metadata...
We are using interfaces to ensur ethey are strongly typed.

src/client/datascience/notebook/integration.ts
```diff
@@ -75,8 +75,8 @@ export class NotebookIntegration implements IExtensionSingleActivationService {
                                 runnable: true,
                                 executionOrder: false,
                                 custom: false,
-                                runState: false,
-                                statusMessage: false
+                                runState: true,
+                                statusMessage: true
```
DonJayamanne, line 79: These should never cause the document to be dirty as they are not persisted in the ipynb file.
These are used only to update the UI (at least in our implementation of notebooks)

src/client/datascience/kernel-launcher/kernelLauncher.ts
```diff
@@ -51,21 +50,6 @@ export class KernelLauncher implements IKernelLauncher {
         @inject(IKernelDependencyService) private readonly kernelDependencyService: IKernelDependencyService
     ) {}
 
-    // This function is public so it can be called when a test shuts down
```
rchiodo, line 54: Why did you remove this? Is it not called anywhere?

rchiodo (approved):

DavidKutu (approved):

src/client/datascience/kernel-launcher/kernelLauncher.ts
DonJayamanne: Yes, not called anywhere

IanMatthewHuff (approved):

PR #4844, opened 2021-02-17
Add no-unnecessary-condition ESLint rule

joyceerhl:
We need to add this rule to our config in order for eslint to apply it when linting.


DavidKutu (approved):

rchiodo:
Are you going to fix all of the errors?

rchiodo (approved):

joyceerhl:
@rchiodo nah I think I'm going to look into running this rule and the  '@typescript-eslint/strict-boolean-expressions' rule on changed files only instead. I don't have enough context on all 300+ errors to know whether changing the check will result in a behavior change.

IanMatthewHuff (approved):

PR #4849, opened 2021-02-17
Final release update

rchiodo:
For #4743


IanMatthewHuff (approved):
Seems good. Assume that you don't need to wait for my command change and Jim's README change as those won't add dependencies?

rchiodo:
> Seems good. Assume that you don't need to wait for my command change and Jim's README change as those won't add dependencies?

Yeah they shouldn't change that stuff.

Issue #4775, opened 2021-02-12
Native Notebook - Don't know how to change kernel. Method is undiscoverable.


claudiaregio:
**For stable experiment only**
When user creates or opens first new notebook create popup that says: 

"Tip: You can change the kernel for your notebook by clicking in the bottom right of the blue global status bar"

DonJayamanne:
Validated

PR #4873, opened 2021-02-19
Add new kernel message

rchiodo:
For #4775


DavidKutu (approved):

src/client/datascience/openNotebookBanner.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    private async openedNotebook(editor: INotebookEditor) {
+        if (
+            !this.pythonExtensionChecker.isPythonExtensionInstalled &&
+            editor.model.metadata &&
+            isPythonNotebook(editor.model.metadata)
+        ) {
+            await this.pythonExtensionChecker.showPythonExtensionInstallRecommendedPrompt();
+        } else if (
+            this.appEnv.channel === 'stable' &&
```
IanMatthewHuff, line 37: Wouldn't insiders also probably want to see this? I'd imagine we're still getting new insiders users, or ones who didn't know about thiat.

IanMatthewHuff (approved):
Code seems fine, but maybe consider showing for insiders as well?

src/client/datascience/openNotebookBanner.ts
rchiodo: @claudiaregio you originally asked for stable only. Do you want to consider this for insiders too?

Issue #1238, opened 2020-08-21
Allow markdown/myst-markdown code blocks to be executed in interactive window.


amichuda:
# Feature: Notebook Editor, Interactive Window, Python Editor cells

## Description

Right now, we can only run the interactive window with a standard `.py` file with markdown commented out. It would be useful to have a way to associate the interactive window with a standard markdown file with code cells, so that those code cells can be run even though it's not a `.py` file. 

For example:

<<---- Begin myfile.md

# My title 

Some text I'm writing

```python
import this

print("hello")
```

------->> End myfile.md

The code cell in this case would have the "Run Cell" prompts appear and you can send that text to the interactive window, even though it's not a `.py` file.

This would be useful for those that use reproducible workflows with [pweave](http://mpastell.com/pweave/) or [codebraid](https://github.com/gpoore/codebraid) to be able to run their code before converting to PDF.

Microsoft Data Science for VS Code Engineering Team: @rchiodo, @IanMatthewHuff, @DavidKutu, @DonJayamanne, @greazer, @joyceerhl

amichuda:
Hi! I hope you had a great new year. I wanted to circle back and see if this was anyone's radar at all, but I understand if the team is busy working on other things!

rchiodo:
@amichuda it's not on our radar at the moment, but we welcome PRs if you want to take a crack at it. I think it would be something like a setting that determines regexes for cells for different languages.

Right now we have these two settings:

```json
"jupyter.codeRegularExpression": "^(#\\s*%%|#\\s*\\<codecell\\>|#\\s*In\\[\\d*?\\]|#\\s*In\\[ \\])"
"jupyter.markdownRegularExpression": "^(#\\s*%%\\s*\\[markdown\\]|#\\s*\\<markdowncell\\>)"
```

It think if we wanted to make it more generic, we'd need one of these settings per language. Maybe something like we do for the variable queries:

```json
"jupyter.codeLensExpressions" : [
   { 
     "extension": ".py",
     "language": "python",
     "codeExpression": "^(#\\s*%%|#\\s*\\<codecell\\>|#\\s*In\\[\\d*?\\]|#\\s*In\\[ \\])",
     "markdownExpression": "^(#\\s*%%\\s*\\[markdown\\]|#\\s*\\<markdowncell\\>)"
   },
   { 
     "extension": ".md",
     "language": "markdown",
     "codeExpression": "```",
     "markdownExpression": ""
   }
]

[...]

Issue #4796, opened 2021-02-12
VS Code Insiders cannot open notebooks


rchiodo:
# Notebooks broken in insiders

Since this version:

![image](https://user-images.githubusercontent.com/19672699/107831658-6a625e00-6d43-11eb-8da3-c677a7b9693e.png)

VS code insiders is making breaking changes to the proposed API the jupyter extension uses to open notebooks. Until we can get a new build out, we recommend going back to stable. 

This issue will track our progress on making a new build just for insiders.

You can also go back to an older build of insiders if you want to continue using it. See directions here on how to do that:
microsoft/vscode#46089 (comment)

peterschmidt85:
I have the same issue.

ShuhuaGao:
Same issue https://github.com/microsoft/vscode-jupyter/issues/4806

ShuhuaGao:
After the latest update of VS Code Insiders, the error message becomes "Unable to open 'IL_BC.ipynb': Cannot read property 'Markdown' of undefined.". 

Version: 1.54.0-insider (user setup)
Commit: a699ffaee62010c4634d301da2bbdb7646b8d1da
Date: 2021-02-15T05:17:48.792Z
Electron: 11.2.3
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Windows_NT x64 10.0.18362

furkantektas:
Same issue here:
`Unable to open 'XXX.ipynb': Cannot read property 'Code' of undefined.`

```
Version: 1.54.0-insider
Commit: a699ffaee62010c4634d301da2bbdb7646b8d1da
Date: 2021-02-15T05:15:52.209Z (5 hrs ago)
Electron: 11.2.3
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Darwin x64 20.3.0

Jupyter versions (conda):

jupyter_client            6.1.11             pyhd8ed1ab_1    conda-forge
jupyter_core              4.7.1            py37hf985489_0    conda-forge
jupyter_server            1.3.0            py37hf985489_0    conda-forge
jupyterlab                3.0.7              pyhd8ed1ab_0    conda-forge
jupyterlab_code_formatter 1.4.4              pyhd8ed1ab_0    conda-forge
jupyterlab_execute_time   2.0.1              pyhd8ed1ab_1    conda-forge
[...]

PR #4876, opened 2021-02-19
generalize code and markdown code lenses to be used with multiple languages

amichuda:
For https://github.com/microsoft/vscode-jupyter/issues/1238

This is still rough, but I wanted to make sure that things were being done correctly (pretty new at typescript)

Please let me know and I can make any changes

**Thank you so much for your patience!**

## Features
- Creates a `codeLensExpression` in the config to allow for both `.py` and `.md` files to be used in interactive window
- Allows for file specific code and markdown Expressions, as well as `defaultCellMarker`
- Can also be extended to more languages?

## Problems

- Units Tests fail for `cellhashprovider.unit.test` 
- Unit tests for `datascience.unit.test` fail for `dataScience.activate() because of this line:

```ts
const codeLensExpressions = this.configuration.getSettings(undefined).codeLensExpressions;
        codeLensExpressions.forEach(c => {
                 vscode.languages.registerCodeLensProvider({ language: c.language }, this.dataScienceCodeLensProvider)
             );
        });
```

-   [ ] Has sufficient logging. **Not sure how to do this**
-   [ ] Has telemetry for enhancements. **Not sure how to do this**


src/client/datascience/cellMatcher.ts
```diff
@@ -16,18 +17,30 @@ export class CellMatcher {
...
+
+        if (language) {
+            language = language;
+        } else if (window.activeTextEditor?.document) {
+            language = window.activeTextEditor?.document.languageId;
+        } else {
+            language = 'python';
+        }
+
+        let codeLens = settings?.codeLensExpressions
```
rchiodo, line 30: Linter should have flagged this I believe. The 'let' should be a 'const'

src/client/datascience/cellMatcher.ts
```diff
@@ -16,18 +17,30 @@ export class CellMatcher {
     private defaultCellMarker: string;
     private defaultCellMarkerExec: RegExp;
 
-    constructor(settings?: IJupyterSettings) {
+    constructor(settings?: IJupyterSettings, language?: string) {
+
+        if (language) {
```
rchiodo, line 22: I think I'd rewrite this like so:

```suggestion
if (!language) {
   language = python
}
```

And don't use the active text editor here. It should be used in the caller. Otherwise this code has an implicit dependency on the window namespace that isn't necessary.

src/client/datascience/jupyter/jupyterImporter.ts
```diff
@@ -105,7 +105,9 @@ export class JupyterImporter implements INotebookImporter {
     };
 
     private get defaultCellMarker(): string {
-        return this.configuration.getSettings().defaultCellMarker || Identifiers.DefaultCodeCellMarker;
+        const language = window.activeTextEditor?.document.languageId
```
rchiodo, line 108: We actually have an abstraction around the VScode 'window' namepace called ```IDocumentManager```. You'd inject an ```IDocumentManager``` into the constructor of the importer and use that here instead of the window namespace.

This abstraction allows us to have tests that can run without VS code being involved.

src/client/datascience/cellMatcher.ts
rchiodo: How do you know that you want to use the active editor? There are cases where this is called on a non active editor

rchiodo: So that's why I'd pass in the language only. Let the caller decided what the language is.

rchiodo:
Looks good except for the cell matcher part with passing the language in.

Thanks for the PR.

rchiodo:
Oh it seems you need to update .eslintrc.js. The linter step is giving this error:

```
Error: Files are being ignored that should be linted: src/test/datascience/cellMatcher.unit.test.ts
src/test/datascience/editor-integration/codewatcher.unit.test.ts
src/datascience-ui/react-common/settingsReactSide.ts
src/client/datascience/datascience.ts
src/client/datascience/cellMatcher.ts
src/client/datascience/editor-integration/codewatcher.ts
src/client/datascience/jupyter/jupyterImporter.ts
```

That means remove those files from the linter ignore list. You'll probably get linter errors then. 

You can run lint on those files if you install the eslint extension (not sure if contributing.md mentions that or not. I'll go fix it if it doesn't)

rchiodo:
The 'VSCode' tests are expected to fail until we can catch up to the new notebook APIs that VS code has been changing (we run our tests with insiders) (See this bug: #4796)

However the other tests should be passing pretty close to 100%. After you fix the build errors, they should go green.

src/client/datascience/cellMatcher.ts
amichuda: Ah so you're saying, don't pass in a default value for language, just let it be an argument in the method and end it there?

rchiodo: Yeah I think so.

rchiodo: Make it the responsibility of the caller to determine it

amichuda: >  that wasn't done for any good reason, just don't know the best way to get the `TextDocument` object in each situation, so i defaulted to that.

rchiodo: No worries. The calling code should be able to make a better decision I think about whether or not to use the active editor to determine the language. You might have to trace back a few levels though (in the callers) to determine what they're actually doing. I think some of the code in cellFactory.ts has some intermediary functions that you might need to pass the language through as well.

src/client/datascience/jupyter/jupyterImporter.ts
amichuda: In this case, do I call the active text editor: 

```
const language = this.DocumentManager.activeTextEditor?.document.languageId
```

where the constructor now has: 

```
@inject(IDocumentManager) private readonly DocumentManager: IDocumentManager
```

rchiodo: Yes something like that. The importer may be called for things other than the active document though. It would be good to check where defaultCellMarker is used (under what circumstances). The caller of defaultCellMarker may have a file.

amichuda: Where can I find the caller of defaultCellMarker?

rchiodo: If you have typescript reference tracking turned on, the IDE can tell you:

image

Or you can search for defaultCellMarker in that same file. It's a private member.

src/client/datascience/cellMatcher.ts
amichuda: So, sorry for the hand holding, but youre asking that I look through all instances of CellMatcher and see where a language argument is needed?

rchiodo: Well all of them should need one, but each caller will get it in different ways. So yes, you'd have to look for all callers of CellMatcher (you should be able to use the same go to references for the constructor to find them).

amichuda:
Getting a `no-floating-promises` error in `datascience.ts`

```ts
    private onSettingsChanged = () => {
        const settings = this.configuration.getSettings(undefined);
        const ownsSelection = settings.sendSelectionToInteractiveWindow;
        const editorContext = new ContextKey(EditorContexts.OwnsSelection, this.commandManager);
        editorContext.set(ownsSelection).catch();
    };

    private onChangedActiveTextEditor() {
        // Setup the editor context for the cells
        const editorContext = new ContextKey(EditorContexts.HasCodeCells, this.commandManager);
        const activeEditor = this.documentManager.activeTextEditor;

        if (activeEditor && activeEditor.document.languageId === PYTHON_LANGUAGE) {
            // Inform the editor context that we have cells, fire and forget is ok on the promise here
            // as we don't care to wait for this context to be set and we can't do anything if it fails
            editorContext.set(hasCells(activeEditor.document, this.configuration.getSettings())).catch();
        } else {
            editorContext.set(false).catch();
        }
    }
```

amichuda:
have to fix a few more things as it turns out...

rchiodo:
> Getting a `no-floating-promises` error in `datascience.ts`
> 
> ```ts
>     private onSettingsChanged = () => {
>         const settings = this.configuration.getSettings(undefined);
>         const ownsSelection = settings.sendSelectionToInteractiveWindow;
>         const editorContext = new ContextKey(EditorContexts.OwnsSelection, this.commandManager);
>         **editorContext.set(ownsSelection).catch();**
>     };
> 
>     private onChangedActiveTextEditor() {
>         // Setup the editor context for the cells
>         const editorContext = new ContextKey(EditorContexts.HasCodeCells, this.commandManager);
>         const activeEditor = this.documentManager.activeTextEditor;
> 
>         if (activeEditor && activeEditor.document.languageId === PYTHON_LANGUAGE) {
>             // Inform the editor context that we have cells, fire and forget is ok on the promise here
>             // as we don't care to wait for this context to be set and we can't do anything if it fails
>             **editorContext.set(hasCells(activeEditor.document, this.configuration.getSettings())).catch();**
>         } else {
>             **editorContext.set(false).catch();**
>         }
>     }
> ```

I think should be fixable using ```.ignoreErrors()``` instead of ```.catch()```

amichuda:
> > Getting a `no-floating-promises` error in `datascience.ts`
> > ```ts
> >     private onSettingsChanged = () => {
> >         const settings = this.configuration.getSettings(undefined);
> >         const ownsSelection = settings.sendSelectionToInteractiveWindow;
> >         const editorContext = new ContextKey(EditorContexts.OwnsSelection, this.commandManager);
> >         **editorContext.set(ownsSelection).catch();**
> >     };
> > 
> >     private onChangedActiveTextEditor() {
> >         // Setup the editor context for the cells
> >         const editorContext = new ContextKey(EditorContexts.HasCodeCells, this.commandManager);
> >         const activeEditor = this.documentManager.activeTextEditor;
> > 
> >         if (activeEditor && activeEditor.document.languageId === PYTHON_LANGUAGE) {
> >             // Inform the editor context that we have cells, fire and forget is ok on the promise here
> >             // as we don't care to wait for this context to be set and we can't do anything if it fails
> >             **editorContext.set(hasCells(activeEditor.document, this.configuration.getSettings())).catch();**
> >         } else {
> >             **editorContext.set(false).catch();**
> >         }
> >     }
> > ```
> 
> I think should be fixable using `.ignoreErrors()` instead of `.catch()`

Perfect.

Another issue, I just noticed is that the  code cell expressions are sent to the interactive window as well, leading to python syntax error. I think it needs to be commented before being sent. 

image

amichuda:
Last merge fixed the issue with `.catch()`.

Now there's an issue with `encryptedStorage`:

```
public async store(service: string, key: string, value: string | undefined): Promise<void> {
        // On CI we don't need to use keytar for testing (else it hangs).
        if (IS_REMOTE_NATIVE_TEST && this.extensionContext.extensionMode !== ExtensionMode.Production) {
            this.testingState.set(`${service}#${key}`, value || '');
            return;
        }

        if (!value) {
            await this.extensionContext.secrets.delete(`${service}.${key}`);
        } else {
            await this.extensionContext.secrets.store(`${service}.${key}`, value);
        }
    }
    public async retrieve(service: string, key: string): Promise<string | undefined> {
        // On CI we don't need to use keytar for testing (else it hangs).
        if (IS_REMOTE_NATIVE_TEST && this.extensionContext.extensionMode !== ExtensionMode.Production) {
            return this.testingState.get(`${service}#${key}`);
        }
        // eslint-disable-next-line
        const val = await this.extensionContext.secrets.get(`${service}.${key}`);
        return val;
    }

```

rchiodo:
> Last merge fixed the issue with `.catch()`.
> 
> Now there's an issue with `encryptedStorage`, "Property `secrets` does not exist on type 'IExtensionContext` :
> 
> ```ts
> public async store(service: string, key: string, value: string | undefined): Promise<void> {
>         // On CI we don't need to use keytar for testing (else it hangs).
>         if (IS_REMOTE_NATIVE_TEST && this.extensionContext.extensionMode !== ExtensionMode.Production) {
>             this.testingState.set(`${service}#${key}`, value || '');
>             return;
>         }
> 
>         if (!value) {
>             await this.extensionContext.secrets.delete(`${service}.${key}`);
>         } else {
>             await this.extensionContext.secrets.store(`${service}.${key}`, value);
>         }
>     }
>     public async retrieve(service: string, key: string): Promise<string | undefined> {
>         // On CI we don't need to use keytar for testing (else it hangs).
>         if (IS_REMOTE_NATIVE_TEST && this.extensionContext.extensionMode !== ExtensionMode.Production) {
>             return this.testingState.get(`${service}#${key}`);
>         }
>         // eslint-disable-next-line
>         const val = await this.extensionContext.secrets.get(`${service}.${key}`);
>         return val;
>     }
> ```

This one is likely a merge required from main. We updated our version of VS code's API that we're using.

amichuda:
Sorry for not getting to this for a long time. My pesky dissertation was getting in the way!
 
I'll try to get to this ASAP, but would also be completely open to having someone push this over the finish line!

rchiodo:
@amichuda I merged from main and it now builds again but I'm not sure the behavior is correct. 

If I have markdown like so:
```md
```python
print('foo')

```
```
It sends the code lens lines along with the python code. The interactive window shows this as an error:

![image](https://user-images.githubusercontent.com/19672699/121742780-2bf06c00-cab5-11eb-8ee6-c22b90189ab4.png)

I'll try debugging it later.

**Comment by rchiodo** [comment]
Had some time to debug. The code here was not stripping your delimiter so it was failing to be valid python code.
https://github.com/microsoft/vscode-jupyter/blob/afddee11c14a4bfdb0bb0c338a1e40efcdd4dd5b/src/client/datascience/jupyter/jupyterNotebook.ts#L901

I made it so it could pick up the language id

File: package.json [hunk 4]
```diff
@@ -1546,23 +1547,24 @@
...
-                    "description": "Cell marker used for delineating a cell in a python file.",
-                    "scope": "resource"
-                },
-                "jupyter.markdownRegularExpression": {
-                    "type": "string",
-                    "default": "^(#\\s*%%\\s*\\[markdown\\]|#\\s*\\<markdowncell\\>)",
-                    "description": "Regular expression used to identify markdown cells. All comments after this expression are considered part of the markdown.",
-                    "scope": "resource"
+                "jupyter.codeLensExpressions": {
+                    "type": "array",
```
IanMatthewHuff, line 1551: Looks like this would trash existing user settings here? I think we might need a migration to ship this.

rchiodo: Yeah that's probably true. I think I'll change it to support both. The old one could be for python. The new one would trump the old one if found.

rchiodo: Should both be supported now.

package.json
```diff
@@ -44,6 +44,7 @@
     ],
     "activationEvents": [
         "onLanguage:python",
+        "onLanguage:markdown",
```
IanMatthewHuff, line 47: Code looks pretty reasonable. But one more thing popping out here. The code expressions are added on activation via the lens manager. But it's only activating on python and markdown files loading. The languages that we add to codeExpressions would need to also register here as activation events. But that feels like it would have to be a dynamic list here. Not sure if that is possible.

IanMatthewHuff (changes requested):

package.json
rchiodo: Something to ask VS code then. Can a user add activation events to their settings.json for an extension.

rchiodo:
There's still work to do here. We need some sort of 'end' marker for markdown cells or it won't get the correct cell ranges (nor have executable code)

amichuda:
Sorry about getting back to this SUPER LATE. What's the best way to continue on this, rebase to master and finish the pull request?

rchiodo:
Yeah a rebase would be the starting point. I don't think it works correctly yet though. At least if my memory is correct, markdown code isn't working because it doesn't have a way to match the 'end' for the markdown code. Normal code cells work because the end is the next cell divider. Markdown code is kinda in the middle of other stuff, so it needs a begin/end pair.

I don't think it should be too hard to match on '```python' for codeStart and '```' for codeEnd in markdown files.

DonJayamanne:
I'm going to close this PR as its been around for a while.
Please feel free to create a new PR once the issues have been addressed and/or the merge conflicts resolved.

PR #4901, opened 2021-02-22
WIP: Refactor kernel finding

rchiodo:
Refactoring kernel finding to use two new classes
- ILocalKernelFinder - searches kernelspecs and interprters
- IRemoteKernelFinder - uses a jupyter connection and searches active sessions and kernelspecs


src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -278,3 +280,92 @@ export function isLocalLaunch(configuration: IConfigurationService) {
 
     return false;
 }
+
+export function findPreferredKernelIndex(
```
rchiodo, line 284: This is the 'generic' algorithm for finding a preferred kernel. Shared by both local and remote. At least so far.

DonJayamanne: Can we retain the preferred kernel.
My concern is if users open notebooks and it takes 10s to find kernels, then users cannot run a cell for 10s.
If instead we return the preferred kernel in 1s, they can run a kernel immediately.
E.g. we run into this often, we open a notebook & just run a cell, but have to wait because VSCode is fetching all of the kernels.

I understand we cannot fix this ritht now, however, with this change it becomes impossible as we have to wait for all kernels to be found first, even before we return the preferred kernel. I.e. we won't be able to make such improvements due to this.

src/client/telemetry/index.ts
```diff
@@ -844,9 +844,9 @@ export interface IEventNamePropertyMapping {
      */
     [Telemetry.KernelLauncherPerf]: undefined | never | TelemetryErrorProperties;
     /**
-     * Total time taken to find a kernel on disc.
+     * Total time taken to find a kernel on disc or on a remote machine.
      */
-    [Telemetry.KernelFinderPerf]: undefined | never;
+    [Telemetry.KernelFinderPerf]: { type: 'remote' | 'local' };
```
DonJayamanne, line 849: Do we need this? Doesnt feel necessary, as we have telemetry tracking whether user is in local or remote.

src/client/telemetry/index.ts
```diff
@@ -375,7 +375,7 @@ export interface IEventNamePropertyMapping {
         hashedName: string;
     };
     [Telemetry.HashedCellOutputMimeTypePerf]: never | undefined;
-    [Telemetry.KernelListingPerf]: never | undefined;
+    [Telemetry.KernelListingPerf]: { type: 'remote' | 'local' };
```
DonJayamanne, line 378: Same as other, i don't think this is necessary, as we have a shared property that tells us whether user is in local or remote.

src/client/datascience/notebook/kernelProvider.ts
```diff
@@ -60,40 +50,25 @@ export class VSCodeKernelPickerProvider implements INotebookKernelProvider {
...
-                        return this._onDidChangeKernels.fire(doc);
-                    }
-                }
-                this._onDidChangeKernels.fire(undefined);
-            },
-            this,
-            disposables
-        );
         this.notebook.onDidChangeActiveNotebookKernel(this.onDidChangeActiveNotebookKernel, this, disposables);
+        this.extensions.onDidChange(this.onDidChangeExtensions, this, disposables);
```
DonJayamanne, line 71: Not sure this is a good idea, just becuase user install `xyz extension` we tell VSCode to reload the list of kernels.

src/client/datascience/jupyter/kernels/helpers.ts
rchiodo: We already wait for all kernels to be returned as far as I can tell. The current code calls ```getKernelForLocalJupyterConnection``` to try and find the preferred kernel. That in turn calls into the kernelFinder to find a matching kernel. That in turn then enumerates all of the kernels.

src/client/datascience/notebook/kernelProvider.ts
```diff
@@ -202,146 +114,79 @@ export class VSCodeKernelPickerProvider implements INotebookKernelProvider {
...
-                true,
-                this.kernelProvider,
-                this.notebook,
-                this.context,
-                this.preferredRemoteKernelIdProvider,
-                this.commandManager
+
+        // If we already have a kernel selected, then set that one as preferred
+        const editor =
+            this.notebook.notebookEditors.find((e) => e.document === document) ||
```
DonJayamanne, line 163: **Suggestion - I would suggest using the following two options:**
Absolutely not necessary, but a good place to discuss perf issues so you too are aware of this.

* Check property `activeNotebookEditor.kernel`
	* Exsiting code
* If this is not available, then create a private weakmap and use document as the key.
	* This is for scenarios where user opens a notebok & opens another notebook
	* Then go back to previous notebook, VSC will NOT set notebookEditor.kernel property
	* I.e. if you open notebook & toggle between notebooks or other files, VS Code will ALWAYS keep calling this.
	* We could improve the perf by caching ourselves against the document.
* I found out on friday that VS Code only sets the Kernel property against Editor only when user runs a cell or edits a notebook.

src/client/datascience/notebook/kernelProvider.ts
```diff
@@ -202,146 +114,79 @@ export class VSCodeKernelPickerProvider implements INotebookKernelProvider {
...
-                    if (!sessionManager) {
-                        throw new Error('Session Manager not available');
+            kernels = await this.localKernelFinder.listKernels(document.uri);
+            preferred = await this.localKernelFinder.findKernel(document.uri, getNotebookMetadata(document), token);
+
+            // We need to filter out those items that are for other extensions.
+            kernels = kernels.filter((r) => {
+                if (r.kind !== 'connectToLiveKernel' && r.kernelSpec) {
+                    if (
+                        r.kernelSpec.metadata?.vscode?.extension_id &&
```
DonJayamanne, line 136: Nice.

src/client/datascience/jupyter/kernels/helpers.ts
DonJayamanne: E.g. when searching for exact match, i do'nt think we need to go through all the kernels, either we find it or we don't & if we dont, then we pick the default , which would then require ALL kernels.

DonJayamanne: Never mind, we can look at the telemetry & see how bad this is.

src/client/datascience/jupyter/kernels/kernelSelections.ts
```diff
@@ -40,223 +19,51 @@ const isSimplePythonDisplayName = /python\s?\d?\.?\d?/;
...
+        const displayName = getDisplayNameOrNameOfKernelConnection(kernel);
+        return {
+            label: displayName,
+            ...kernel,
+            // We don't want descriptions.
+            description: '',
+            selection: kernel
         };
-
-        const liveItems = getSelections().then((items) => (this.localSuggestionsCache = items));
```
DonJayamanne, line 230: General comment - lots of code going away 👍🏾 🎆 👍🏾 🎆 👍🏾 🎆 👍🏾 🎆

DonJayamanne (approved):
:shipit:

src/client/telemetry/index.ts
DonJayamanne: Please remove this.

DonJayamanne:
Oh yes, please remove the telemetry changes.

src/client/datascience/jupyter/kernels/helpers.ts
DonJayamanne: > getKernelForLocalJupyterConnection to try and find the preferred kernel. That in turn calls into the kernelFinder to find a matching kernel. That in turn then enumerates all of the kernels.

Agreed.

Just don't like the fact that users have to wait for all kernels to be discovered for then to run a notebook. Feels like its a serious bottle neck, no point creating an issue either as we don't have the telemetry

PR #4914, opened 2021-02-23
Update renderer API to support the new Output format

DonJayamanne:
Corresponding PR https://github.com/microsoft/vscode-notebook-renderers/pull/32

### DIFF
--- a/src/datascience-ui/ipywidgets/kernel/index.ts
+++ b/src/datascience-ui/ipywidgets/kernel/index.ts
@@ -211,6 +211,7 @@ function convertVSCodeOutputToExecutResultOrDisplayData(
     // Send metadata only for the mimeType we are interested in.
     const customMetadata = request.output.metadata?.custom;
     if (customMetadata) {
+        // Support for Old API
         if (customMetadata[request.mimeType]) {
             metadata[request.mimeType] = customMetadata[request.mimeType];
         }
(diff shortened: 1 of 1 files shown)


DavidKutu (approved):

src/datascience-ui/ipywidgets/kernel/index.ts
```diff
@@ -220,6 +221,17 @@ function convertVSCodeOutputToExecutResultOrDisplayData(
...
         }
+    } else {
+        // New API.
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const outputMetadata = request.output.metadata as Record<string, any> | undefined;
+        if (outputMetadata && outputMetadata[request.mimeType] && outputMetadata[request.mimeType].metadata) {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            Object.assign(metadata, outputMetadata[request.mimeType].metadata);
+            if (request.mimeType in outputMetadata[request.mimeType].metadata) {
+                Object.assign(metadata, outputMetadata[request.mimeType].metadata[request.mimeType]);
```
rchiodo, line 232: This seems weird? 

You assign metadata to the metadata object, and then you reassign it to just the mimetype value? Did you intend to do that?

rchiodo (changes requested):
⏲️

src/datascience-ui/ipywidgets/kernel/index.ts
DonJayamanne: Yes, the VSC renderer API hasn't been updated, hence not sure what we get here, hence just sending everything (spread everything out & send it all).
I have filed an issue on VSC (and a corresponding tracking issue at our end as well) for them to update their renderer types so we know what we're getting in the request object & what needs to be send to the renderer.

rchiodo: Ah okay, part of my confusion was I thought object.assign wholly replaces an object. It just assigns the fields from one object into an already existing object (leaving old fields there).

rchiodo (approved):

src/datascience-ui/ipywidgets/kernel/index.ts
DonJayamanne: Very Hacky, just dump everything and hope it works...

PR #4917, opened 2021-02-23
Disable Insiders extension channel in favour of marketplace

DonJayamanne:
Fix required to get Notebook Tests running back in main CI.


src/test/datascience/notebook/helper.ts
```diff
@@ -192,9 +192,9 @@ export async function canRunNotebookTests() {
     }
     const api = await initialize();
     const appEnv = api.serviceContainer.get<IApplicationEnvironment>(IApplicationEnvironment);
-    const canRunTests = appEnv.extensionChannel !== 'stable';
+    const canRunTests = appEnv.channel === 'insiders';
```
DonJayamanne, line 195: We can run notebook tests only in VS Code insiders for now.
Might want to remove this condition when the API stabilizes, else it will fail in stable with main branch.

src/test/common/experiments/service.unit.test.ts
```diff
@@ -54,7 +54,7 @@ suite('Experimentation service', () => {
     }
 
     function configureApplicationEnvironment(channel: Channel, version: string) {
-        when(appEnvironment.extensionChannel).thenReturn(channel);
+        when(appEnvironment.channel).thenReturn(channel);
```
DonJayamanne, line 57: Insider users are only users in VS Code insiders,

rchiodo (approved):

joyceerhl (approved):

Issue #4845, opened 2021-02-17
Cannot open resource with notebook editor 


TarjinderSingh:
I cannot open Jupyter Notebooks from the explorer window at the moment. I get the following error. However, if I right-click and "Open in Notebook Editor", the Jupyter notebook opens well. Is there a reason or setting for this?

```
Cannot open resource with notebook editor type 'jupyter-notebook', please check if you have the right extension installed or enabled.
```

rchiodo:
Thanks for the bug. The problem should go away if you try a second time. 

Can you check if you have this entry in your settings.json?

```
  "workbench.editorAssociations": [
    {
      "viewType": "jupyter.notebook",
      "filenamePattern": "*.ipynb"
    }
  ],
```

That should only be enabled in the insiders version of VS code. If you're using the stable version of VS code you shouldn't have that value there.

TarjinderSingh:
I do have that entry in my settings.json. I have two Macs that are synced on the same vscode settings, and one is using the insiders version and one is using the stable version. The stable version one is showing this error. Do you think the shared settings between the two is causing the issue?

rchiodo:
> Do you think the shared settings between the two is causing the issue?

Maybe? I thought insiders and stable wouldn't share the same settings but if they did it would cause the problem.

TarjinderSingh:
OK, I commented out that block and it works on my Stable version but not the Insiders one. The insiders one is showing "Cannot read property 'Code' of Undefined".

So I am guessing that my settings are synced between the two and that block is breaking.

rchiodo:
> The insiders one is showing "Cannot read property 'Code' of Undefined".

This is actually this bug here: #4796. Insiders is not currently working for opening notebooks. VS code made some breaking API changes that we haven't caught up to yet.

TarjinderSingh:
OK, so I don't know why:
```
"workbench.editorAssociations": [
    {
      "viewType": "jupyter.notebook",
      "filenamePattern": "*.ipynb"
    }
  ],
 ```
is appearing in my stable build. But I am changing this locally for now in my stable build and I will wait for the Insiders fix soon.

greazer:
If a user has the editorAssociations setting, then opt them into the native notebook editor experiment.

DonJayamanne:
Validated, did use settings sync but must updated the settings manually

PR #4927, opened 2021-02-23
Enlist in native notebook experiment if sync'd settings is forcing it…

rchiodo:
Port fix for #4845 back to main

### DIFF
--- a/src/client/common/experiments/service.ts
+++ b/src/client/common/experiments/service.ts
@@ -8,7 +8,7 @@ import { Memento } from 'vscode';
 import { getExperimentationService, IExperimentationService, TargetPopulation } from 'vscode-tas-client';
 import { sendTelemetryEvent } from '../../telemetry';
 import { EventName } from '../../telemetry/constants';
-import { IApplicationEnvironment } from '../application/types';
+import { IApplicationEnvironment, IWorkspaceService } from '../application/types';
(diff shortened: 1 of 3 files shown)


src/client/common/experiments/service.ts
```diff
@@ -57,6 +58,16 @@ export class ExperimentService implements IExperimentService {
         this._optInto = optInto.filter((exp) => !exp.endsWith('control'));
         this._optOutFrom = optOutFrom.filter((exp) => !exp.endsWith('control'));
 
+        // Custom settings just for native notebook support.
+        const settings = workspaceService.getConfiguration('workbench', undefined);
+        const editorAssociations = settings.get('editorAssociations') as {
+            viewType: string;
+            filenamePattern: string;
+        }[];
+        if (editorAssociations.find((a) => a.viewType && a.viewType.includes('jupyter-notebook'))) {
```
DonJayamanne, line 67: Ewww, all of these hacks make me go 🤮 
But i don't see any other way around it, not complaining about your code, just the hacks we have for native notebooks & experiments (opting in & out...) too complicated

DonJayamanne (approved):

joyceerhl (approved):

src/client/common/experiments/service.ts
rchiodo: I agree. I hate the experiments stuff too.

PR #4939, opened 2021-02-24
Add more logging to determine cause for failures

DonJayamanne:


src/client/api/pythonApi.ts
```diff
@@ -214,20 +216,42 @@ export class PythonInstaller implements IPythonInstaller {
...
-            .then((result) => {
-                if (result === InstallerResponse.Installed) {
-                    this._onInstalled.fire({ product, resource });
-                }
-                return result;
+        let action: 'installed' | 'failed' | 'disabled' | 'ignored' = 'installed';
+        try {
+            const api = await this.apiProvider.getApi();
+            const result = await api.install(ProductMapping[product], resource, cancel);
+            if (result === InstallerResponse.Installed) {
```
DonJayamanne, line 228: We didn't know whether users were presented with prompts to install the missing ipykernel & didn't know whether users installed or not.
I.e. after messages such as `ipykernel not installed` is logged, i still messages indicating cells are being run again, but no messages to indicate whether the kernel was installed or even if the prompt was dismissed.

This will tell us whether a kernel is install or not after the initial prompt.

src/client/common/errors/errors.ts
```diff
@@ -38,6 +45,13 @@ function tagWithWin32Error(stdErr: string, tags: string[] = []) {
         tags.push('win32api');
     }
 }
+function tagWithChildProcessExited(stdErrOrStackTrace: string, tags: string[] = []) {
```
DonJayamanne, line 48: Found a few `unknown` errors & looking at the stack trace I can tell we're getting a few of these `ChildProcess.exithandler`, figured i might as well classify them.
Seems to be a kernel has died, but not yet classifying them as kernel died, as we're just getting a generic node error, however the subcategory/tag will help classify these.

src/client/common/errors/index.ts
```diff
@@ -17,7 +17,11 @@ export function populateTelemetryWithErrorInfo(props: Partial<TelemetryErrorProp
         props.failureCategory = 'fetcherror';
     }
     props.stackTrace = serializeStackTrace(error);
-    const stdErr = error instanceof BaseError ? error.stdErr : '';
+    if (typeof error === 'string') {
```
DonJayamanne, line 20: I can see one telemetry failing, but no errors being sent, absolutely nothing.
I suspect this is because we're rejecting with a string, instead of an error object.
This will help capture those so we can fix it.

E.g. i can see we start a notebook, but we don't even attempt to start the raw kernel, meaning something dies well before then

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
```diff
@@ -233,7 +234,7 @@ export class HostRawNotebookProvider
                 !kernelConnectionMetadata ||
                 (kernelConnectionMetadata?.kind === 'startUsingKernelSpec' && !kernelConnectionMetadata?.kernelSpec)
             ) {
-                notebookPromise.reject('Failed to find a kernelspec to use for ipykernel launch');
+                notebookPromise.reject(new KernelSpecNotFoundError());
```
DonJayamanne, line 237: I believe this is one of the culptrits.
I can see that we attempt to start a notebook, & then we don't even attempt to start the raw kernel, but it just dies prematurely & we have absolutely no idea.

If this is the cause, then we know exactly what to fix. Because rejecting with a string, the telemetry cannot determine the cause of the failure & nothing is logged.

src/client/datascience/telemetry/telemetry.ts
```diff
@@ -76,7 +76,7 @@ export function sendKernelTelemetryWhenDone<P extends IEventNamePropertyMapping,
     stopWatch?: StopWatch,
     properties?: P[E]
 ) {
-    if (eventName === Telemetry.ExecuteCell) {
+    if (eventName === Telemetry.ExecuteCell || eventName === Telemetry.ExecuteNativeCell) {
```
DonJayamanne, line 79: This was missing for native notebooks.

src/client/telemetry/index.ts
```diff
@@ -656,6 +656,10 @@ export interface IEventNamePropertyMapping {
     [Telemetry.UserInstalledPandas]: never | undefined;
     [Telemetry.UserDidNotInstallJupyter]: never | undefined;
     [Telemetry.UserDidNotInstallPandas]: never | undefined;
+    [Telemetry.PythonModuleInstal]: {
+        moduleName: string;
+        action: 'displayed' | 'installed' | 'ignored' | 'disabled' | 'failed';
```
DonJayamanne, line 661: Generic way to track all module isntallation prompts, displayed & actions taken by user.

package.json
```diff
@@ -24,7 +24,7 @@
         "theme": "light"
     },
     "engines": {
-        "vscode": "^1.49.0"
+        "vscode": "^1.53.0"
```
DonJayamanne, line 27: Unrelated, didn't want to create a new PR for this.
A user with this extension can end up in an experiment for native notebooks & they must be on the latest version of VS Code, hence the engine must be udpated.

rchiodo (approved):

Issue #4625, opened 2021-02-03
Refresh buttion for data viewer


Victsz:
After open the data viewer on certain dataframe, there is no way to refresh viewer when the underlying data is updated.

Alsoa minor issue - failed to copy (ctrl+c) text from the data viewer tab.

rchiodo:
Thanks for the suggestion. This is a dupe of #https://github.com/microsoft/vscode-jupyter/issues/1143

Issue #1143, opened 2019-05-16
DataViewer should auto-refresh or provide a button to refresh it.


greazer:

gcambray:
DataViewer in debugging cession is a great improvment
In this context, an auto-refresh check box on top of a refresh button would be ideal
Much needed feature!

tsauliu:
need this feature too, with this, we analyst could replace excel by Pandas + Visual Studio Code

IanMatthewHuff:
Validated. Checked with both a simple flat structure as well as with the slicing feature on to see if the slice was preserved.

PR #4951, opened 2021-02-25
Add refresh button to update data in place

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/1143, https://github.com/microsoft/vscode-jupyter/issues/4625

Per jmew and ronglums, add a refresh button to the editor title menu allowing users to update the data being displayed in the data viewer in place. Also add keybinding and command palette command.

If a slice is currently being applied and the currently requested slice is valid for the variable's new shape, update the data viewer in-place. If a slice is currently being applied and the new shape would invalidate the requested slice, update the data viewer in-place with the default slice expression with preselected indices. Otherwise (i.e. no slice is currently being applied), update the data viewer with the full variable contents.

image

To do:
- [x] Correctly track active data viewers so that we update the correct one
- [x] Figure out how to request full variable refresh (shape, type etc. can change)
- [x] Validate current slice expression against new shape and reuse only if valid


package.json
```diff
@@ -259,6 +259,12 @@
                 "command": "jupyter.selectCell",
                 "key": "ctrl+alt+shift+\\",
                 "when": "editorTextFocus && jupyter.hascodecells && !notebookEditorFocused"
+            },
+            {
+                "command": "jupyter.refreshDataViewer",
+                "key": "ctrl+r",
+                "mac": "cmd+r",
```
joyceerhl, line 266: Open to feedback on what this keybinding should be

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -122,11 +122,10 @@ def _VSCODE_getRowCount(var):
 
 # Function to retrieve a set of rows for a data frame
 def _VSCODE_getDataFrameRows(df, start, end):
-    df = _VSCODE_convertToDataFrame(df)
+    df = _VSCODE_convertToDataFrame(df[start:end])
```
joyceerhl, line 125: This makes viewing N-D data 30% faster (previously we were flattening the whole variable and then throwing everything except the requested rows away)

src/client/datascience/data-viewing/dataViewerFactory.ts
```diff
@@ -46,4 +60,29 @@ export class DataViewerFactory implements IDataViewerFactory, IAsyncDisposable {
...
+
+    private async updateViewStateContext() {
+        // A data viewer's view state has changed. Look through our known viewers to see if any are active
+        let hasActiveViewer = false;
+        this.knownViewers.forEach((viewer) => {
+            if (viewer.active) {
+                hasActiveViewer = true;
+            }
+        });
+        await this.viewContext.set(hasActiveViewer);
```
joyceerhl, line 76: Didn't think it really made sense to track this context in the activeEditorContext since the data viewer is not an editor and the dataViewerFactory already tracks all active data viewers.

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -134,10 +138,8 @@ export class DebuggerVariables extends DebugLocationTracker
             (targetVariable as any).frameId
         );
 
-        let fileName;
-        if (notebook) {
-            fileName = path.basename(notebook.identity.path);
-        } else if (this.debugLocation?.fileName) {
+        let fileName = notebook ? path.basename(notebook.identity.path) : '';
+        if (!fileName && this.debugLocation?.fileName) {
```
joyceerhl, line 142: Found a problem, when debugging in the interactive window `notebook` is undefined, hence should use debugLocation.fileName to populate breadcrumbs

src/datascience-ui/data-explorer/mainPanel.tsx
```diff
@@ -154,33 +156,52 @@ export class MainPanel extends React.Component<IMainPanelProps, IMainPanelState>
...
         );
     };
 
-    private renderBreadcrumb() {
-        if (this.state.fileName) {
-            let breadcrumbText = this.state.variableName;
-            if (this.state.originalVariableShape) {
-                breadcrumbText += ' (' + this.state.originalVariableShape?.join(', ') + ')';
-            }
+    public renderSliceControls = () => {
```
joyceerhl, line 165: Moved rendering SliceControl out of ReactSlickGrid, there really wasn't a reason for it to be a child of the grid

src/datascience-ui/data-explorer/reactSlickGrid.tsx
```diff
@@ -685,7 +667,8 @@ function readonlyCellEditor(this: any, args: any) {
         // to move their cursor within the focused input element
         if (
             (e.keyCode === KeyCodes.LeftArrow && cursorPosition > 0) ||
-            (e.keyCode === KeyCodes.RightArrow && cursorPosition < textLength - 1)
+            (e.keyCode === KeyCodes.RightArrow && cursorPosition < textLength - 1) ||
+            (e.ctrlKey && e.keyCode === KeyCodes.X)
```
joyceerhl, line 671: Fixing a problem that Rich found where ctrl+x deletes the contents of the focused cell. Not destructive (the old contents get restored as soon as cell focus is lost), just a very goofy UX

src/datascience-ui/data-explorer/mainPanel.tsx
```diff
@@ -388,9 +417,7 @@ export class MainPanel extends React.Component<IMainPanelProps, IMainPanelState>
             if (!r) {
                 r = {};
             }
-            if (!r.hasOwnProperty(this.state.indexColumn)) {
-                r[this.state.indexColumn] = this.state.fetchedRowCount + idx;
-            }
+            r[this.state.indexColumn] = this.state.fetchedRowCount + idx;
```
joyceerhl, line 420: Generate index column numbers in the UI. This is necessary with the vscodeDataFrame.py change, since those slices are now indexed relative to themselves and not to the original variable. Tested with AML, this does not mess up their actual index columns. This also doesn't impact data frames since the indices all line up with what would've been generated in the backend.

src/datascience-ui/data-explorer/sliceControl.tsx
```diff
@@ -44,11 +45,16 @@ const styleOverrides = {
...
     root: {
-        ':hover': {
-            color: 'var(--vscode-dropdown-foreground)'
+        color: 'var(--vscode-dropdown-foreground)'
+    },
+    dropdownItems: {
+        ...styleOverrides,
+        selectors: {
+            '@media(min-width: 300px)': {
+                maxHeight: 100
```
joyceerhl, line 54: Limit the dropdown height (this is the recommended way per FluentUI devs)

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -156,12 +160,6 @@ def _VSCODE_getDataFrameInfo(df):
     else:
         columnNames = list(df)
 
-    # Compute the index column. It may have been renamed
```
rchiodo, line 159: I thought this was necessary in the grid so wondering why it was removed?

src/client/datascience/data-viewing/dataViewerFactory.ts
```diff
@@ -46,4 +60,29 @@ export class DataViewerFactory implements IDataViewerFactory, IAsyncDisposable {
...
+    }
+
+    private refreshDataViewer() {
+        // Find the data viewer which is currently active
+        for (const viewer of this.knownViewers) {
+            if (viewer.active) {
```
rchiodo, line 82: Not sure viewer.active is all that accurate unfortunately. I've had problems with interactive window not turning this to false when being hidden.

src/client/datascience/jupyter/jupyterVariables.ts
```diff
@@ -65,9 +65,15 @@ export class JupyterVariables implements IJupyterVariables {
     public async getDataFrameInfo(
         targetVariable: IJupyterVariable,
         notebook?: INotebook,
-        sliceExpression?: string
+        sliceExpression?: string,
+        shouldUpdateCachedInfoCachedInfo?: boolean
```
rchiodo, line 69: Suggestion: Add new method for clearing the cache instead. Makes this method have fewer responsibilities.

rchiodo: Actually that doesn't work as well. What would you clear? If it's on this item it forces a refresh of just this variable.

So ignore the suggestion.

src/datascience-ui/data-explorer/mainPanel.tsx
rchiodo: Thanks this answers my other question.

src/datascience-ui/data-explorer/reactSlickGrid.tsx
```diff
@@ -494,7 +475,8 @@ export class ReactSlickGrid extends React.Component<ISlickGridProps, ISlickGridS
                         buttons: [
                             {
                                 cssClass: 'codicon codicon-filter codicon-button',
-                                handler: this.clickFilterButton
+                                handler: this.clickFilterButton,
+                                tooltip: this.state.showingFilters ? 'Hide filters' : 'Show filters'
```
rchiodo, line 479: These need to be localized

src/datascience-ui/data-explorer/sliceControl.tsx
```diff
@@ -235,6 +269,12 @@ export class SliceControl extends React.Component<ISliceControlProps, ISliceCont
         }
     };
 
+    /**
+     *
+     * @returns A Python slice expression with the 0th index preselected along
+     * the first ndim - 2 axes. For example, for a 5D array with shape
```
rchiodo, line 275: Nice comment. Thanks

rchiodo (changes requested):
Looks good except the loc for the tooltips.

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
joyceerhl: Part of the data viewer redesign changes included generating an index column, and although we were doing this for our Python variables in the backend, the AML data viewer wouldn't have gotten that numeric column unless their backend changed too. The intent here is to always generate the index column in the UI so that AML will also get it.

src/client/datascience/jupyter/jupyterVariables.ts
joyceerhl: Would it be better if I called getFullVariable in the jupyterVariableDataProvider, and then passed that result to getDataFrameInfo in jupyterVariables.ts, since kernelVariables and debuggerVariables do not cache anything? The flag name feels like a misnomer at a minimum.

src/client/datascience/data-viewing/dataViewerFactory.ts
joyceerhl: FWIW I implemented `viewer.active` as a getter whose value is `webPanel.isActive()`, and that in turn is directly equal to the vscode webview panel's `active` property, so this is slightly different from what the interactive window has. I implemented it this way because I noticed that the viewstate in the webPanelHost was wrong. Even after being disposed the viewstate.active property was still `true` because `onDidChangeViewState` doesn't fire when vscode webview panels are disposed. It's possible that is same reason that the interactive window's `active` property is wrong.

joyceerhl:
@rchiodo I reworked the refresh mechanism, tested and found that the previous mechanism only worked if the current variable was sliceable. Sorry for the randomization, I'd appreciate you taking a second look!

src/client/datascience/data-viewing/types.ts
```diff
@@ -77,7 +78,7 @@ export interface IDataFrameInfo {
 
 export interface IDataViewerDataProvider {
     dispose(): void;
-    getDataFrameInfo(sliceExpression?: string): Promise<IDataFrameInfo>;
+    getDataFrameInfo(sliceExpression?: string, shouldUpdateCachedInfo?: boolean): Promise<IDataFrameInfo>;
```
rchiodo, line 81: I think calling this refresh or something like that is better. 'shouldUpdateCachedInfo' means an external caller knows something of the internal implementation. How does an external caller know there's a cache? All the external caller should know is that the user hit the refresh button (or a refresh of the data is required)

rchiodo (approved):

joyceerhl:
@DonJayamanne @DavidKutu do y'all have any feedback on this PR?

Issue #4956, opened 2021-02-25
Wait for all data to be available before sending telemetry


DonJayamanne:
We capture some information about the interpreter and send this along with the telemetry for failures (why kernel failed).
However some information might not be available immediately (its fetched asychronously).

Suggestion:
* When sending the telemetry, wait until such information is available & queue all of the telemetry

E.g.:
* We know we need ipykernel version, at that point we wait for the promise (with a timeout).

IanMatthewHuff:
Verified by looking at properties of new failure event telemetry. 

```
[...]

PR #4981, opened 2021-03-01
Queue & Wait for data to be available before sending telemetry

DonJayamanne:
For #4956

### DIFF
--- a/src/client/datascience/telemetry/telemetry.ts
+++ b/src/client/datascience/telemetry/telemetry.ts
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+import cloneDeep = require('lodash/cloneDeep');
 import { Uri } from 'vscode';
 import { getOSType } from '../../common/utils/platform';
 import { getKernelConnectionId, KernelConnectionMetadata } from '../jupyter/kernels/types';
@@ -290,11 +292,15 @@ function getContextualPropsForTelemetry(
             resourceType
         };
     }
-    if (data) {
-        // Possible the Python package information is now available, update the properties accordingly.
-        updatePythonPackages(data[0]);
+    if (!data) {
+        return;
     }
-    return data ? data[0] : undefined;
+    // Create a copy of this data as it gets updated later asynchronously for other events.
+    // At the point of sending this telemetry we don't want it to change again.
+    const clonedData = cloneDeep(data[0]);
+    // Possible the Python package information is now available, update the properties accordingly.
+    updatePythonPackages(clonedData);
+    return clonedData;
 }
 /**
  * Some information such as interrupt counters & restart counters need to be reset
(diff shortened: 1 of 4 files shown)


src/client/telemetry/index.ts
```diff
@@ -119,16 +120,82 @@ function stringifyProperties(eventName: string, data: Record<string, any>) {
...
+        queuedTelemetry.push({
+            eventName: eventName as string,
+            durationMs,
+            properties,
+            ex,
+            sendOriginalEventWithErrors
+        });
+    }
+
+    sendNextTelemetryItem();
```
DonJayamanne: This is where we queue if stuff has already been queued.

DavidKutu (approved):

src/client/datascience/telemetry/telemetry.ts
```diff
@@ -55,11 +58,17 @@ export function sendKernelTelemetryEvent<P extends IEventNamePropertyMapping, E
         setSharedProperty('userExecutedCell', 'true');
     }
 
-    const addOnTelemetry = getContextualPropsForTelemetry(resource);
-    const props = properties || {};
-    Object.assign(props, addOnTelemetry || {});
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    sendTelemetryEvent(eventName as any, durationMs, props, ex, true);
+    const props = getContextualPropsForTelemetry(resource);
```
rchiodo, line 61: I had another thought. Would it be possible to send properties as promises? Then in the sendTelemetryEvent just check if any properties were promises.

If any were promises, put the event into a queue that submits once all of promises finish
If any event comes in, and the queue has an item, it goes into the queue too.

DonJayamanne: The problem is with our strong typing, we only allow sending telemetry properties that are known.
Her'es a sample 
```typescript
    [Telemetry.KernelEnumeration]: {
        /**
         * Count of the number of kernels found
         */
        count: number;
        /**
         * Boolean indicating if any are python or not
         */
        isPython: boolean;
        /**
         * Indicates how the enumeration was acquired.
         */
        source: 'cli' | 'connection';
    };
```

Here' we'll now need to add a property with a promise.
I.e. we'll need to update `IEventNamePropertyMapping` to contain a new promise.
The current approach makes it unnecessary to update `IEventNamePropertyMapping`

> If any event comes in, and the queue has an item, it goes into the queue too.

This already happens today, i just check if the queue is empty or not.

rchiodo: Wouldn't that be better? Having strong typing for promises? Not sure why that's a downside.

I need to look some more but your current implementation is passing a flag to indicate if should be queued or not?

DonJayamanne: & now we'll have 
```
    [Telemetry.KernelEnumeration]: {
        /**
         * Count of the number of kernels found
         */
        count: number;
        /**
         * Boolean indicating if any are python or not
         */
        isPython: boolean;
        /**
         * Indicates how the enumeration was acquired.
         */
        source: 'cli' | 'connection';
		// New Property.
		isReady: Promise<void>
    };
```

Optionally, we could allows the property `isReady` to all telemetry properties passed into the `sendTelemetryEvent` (& IEventNamePropertyMapping is left untouched) & when we send the telemetry we jsut drop that property.

How does that sound (pretty much what you are suggesting)?

rchiodo: Sure that works too.

rchiodo (approved):

Issue #4926, opened 2021-02-23
Log ICommandManager.executeCommand telemetry


rchiodo:
Seems VS code doesn't track our usage of this API so we need to add this ourselves.

greazer:
Would be good to be able to tell why each call to this function occurs (to understand if the user explicitly did an action to call this or not).

IanMatthewHuff:
Validated.
nfo Jupyter Extension: 2021-03-26 14:12:49: Telemetry Event : DS_INTERNAL.COMMAND_EXECUTED Measures: undefined Props: {"command":"jupyter.clearSavedJupyterUris","installSource":"marketPlace","VSCode.ABExp.Features":"invitationLink;installRecovery;livesharecontinuousaacf;reusableLinks;RemoveKernelToolbarInInteractiveWindowcf;pythonJediLSP;portForwardingServiceEnabled-development;portForwardingServiceEnabled-staging;portForwardingServiceEnabled-production;pythonDiscoveryModule;pythonFindInterpreter;presentWelcomePageByDefault","abexp.assignmentcontext":"vsliv695:30137379;vsins829:30139715;vsliv368cf:30146710;vsreu685:30147344;python383cf:30185419;pythonvspyt602:30280718;vspor879:30202332;vspor708:30202333;vspor363:30204092;pythonvsnew554:30265444;pythonvspyt943:30277000;vspre833:30267464;","isamlcompute":"false","isInsiderExtension":"true","ds_notebookeditor":"native","localOrRemoteConnection":"local","isPythonExtensionInstalled":"true","rawKernelSupported":"true"}

PR #4983, opened 2021-03-01
Telemetry to track the commands executed using ICommandManager

DonJayamanne:
For #4926

### DIFF
--- a/src/client/common/application/commandManager.ts
+++ b/src/client/common/application/commandManager.ts
@@ -36,9 +36,9 @@ export class CommandManager implements ICommandManager {
             (...args: U) => {
                 sendTelemetryEvent(Telemetry.CommandExecuted, undefined, { command: command as string });
                 if (thisArg) {
-                    callback.call(thisArg, ...args);
+                    return callback.call(thisArg, ...args);
                 } else {
-                    callback(...args);
+                    return callback(...args);
                 }
             },
             thisArg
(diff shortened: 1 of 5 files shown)


src/client/common/application/commandManager.ts
```diff
@@ -70,6 +72,7 @@ export class CommandManager implements ICommandManager {
         E extends keyof ICommandNameArgumentTypeMapping,
         U extends ICommandNameArgumentTypeMapping[E]
     >(command: E, ...rest: U): Thenable<T | undefined> {
+        sendTelemetryEvent(Telemetry.CommandExecuted, undefined, { command: command as string });
```
rchiodo, line 75: I was thinking we'd want this on registerCommand too. That way we'd track command palette calls as well.

rchiodo: Or maybe only on registerCommand (otherwise we might have dupes)

DavidKutu (approved):

rchiodo (approved):

src/client/common/application/commandManager.ts
```diff
@@ -5,9 +5,12 @@
 
 import { injectable } from 'inversify';
 import { commands, Disposable, TextEditor, TextEditorEdit } from 'vscode';
+import { Telemetry } from '../../datascience/constants';
+import { sendTelemetryEvent } from '../../telemetry';
 import { ICommandNameArgumentTypeMapping } from './commands';
 import { ICommandManager } from './types';
 
+const commandsToIgnore = new Set<string>(['setContext']);
```
joyceerhl, line 13: Could we rename this variable to capture the fact that it's telemetry specific or add a comment to that effect?

joyceerhl (approved):

src/client/common/application/commandManager.ts
DonJayamanne: Done

Issue #1312, opened 2020-05-24
Warnings are not shown in the Notebook editor


ashtou:
# Bug: Notebook Editor, Interactive Window, Editor cells

Warnings are not shown in the Notebook Editor. 
## Steps to cause the bug to occur

1. If I run the following code in a `.py` from terminal, I get the warning message, but if I run it inside a cell in the Notebook editor, nothing appears.

```python

import warnings  
warnings.warn('Warning Message') 

```

### Your Jupyter and/or Python environment

_Please provide as much info as you readily know_

-   **Jupyter server running:** Local
-   **Extension version:** 2020.5.80290
-   **VS Code version:** 1.45.1
-   **Setting python.jediEnabled:** false
-   **Python version: Python 3.8.0 
-   **OS:** Windows | Mac | Linux (distro): Linux (Ubuntu)
-   **Virtual environment:** conda | venv | virtualenv | N/A | ... venv

## Developer Tools Console Output

Microsoft Data Science for VS Code Engineering Team: @rchiodo, @IanMatthewHuff, @DavidKutu, @DonJayamanne, @greazer

hoechenberger:
Using the VS Code Python extension v2021.1.502429796, I had to use the following settings:

```json
    "jupyter.runStartupCommands": [
        "import warnings", 
        "warnings.simplefilter(action='default')"
    ]
```

I would also like to encourage the developers to take a look at this issue – it's really a big problem for numerous colleagues of mine too, who constantly miss important warnings when working with the interactive console.

IanMatthewHuff:
Moving back to triage.

hoechenberger:
@IanMatthewHuff Not sure if this was intentional or not, but you didn't add the `triage` label :)

IanMatthewHuff:
@hoechenberger We don't actually use that label these days. We do our tracking for what stage issues are in with zenhub. Which sadly isn't publicly visible. But I've moved it back to the triage lane for our team to discuss.

hoechenberger:
@IanMatthewHuff Thanks for the clarification!

joyceerhl:
Validated. Fix should become available in the March release of the Jupyter extension.
 
![image](https://user-images.githubusercontent.com/30305945/109740544-4941a000-7b80-11eb-83fb-cffc1b7dbb34.png)

PR #4986, opened 2021-03-01
Remove special casing for Warnings

DonJayamanne:
For #1312


src/test/datascience/notebook.functional.test.ts
```diff
@@ -1389,20 +1386,6 @@ plt.show()`,
                 }
             });
 
-            test('Notebook launch with PYTHONWARNINGS', async function () {
```
rchiodo, line 1392: Shouldn't this test still pass? I believe the original intent of this was to not have PYTHONWARNINGS cause a kernel to fail to start.

rchiodo (approved):

DavidKutu (approved):

src/test/datascience/notebook.functional.test.ts
DonJayamanne: Oh shoot yes.

PR #4995, opened 2021-03-02
Kernel finding refactor

rchiodo:
Rework how we find kernels.

Two new classes:
- LocalKernelFinder - (based on old KernelFinder class) that is used to find kernels on a local machine. It looks for kernel.json files in all the known jupyter locations and interpreter locations
- RemoteKernelFinder - used to find kernels based on a jupyter connection. Asks jupyter for the kernel list and the active session list

Both of these classes then use the same algorithm to find matches for a notebook's metadata and/or an interpreter. (Algorithm was moved to a separate file)

Unit tests were made to test the different possible conditions that we have for finding kernels.

All other classes dealing with finding kernels were changed to use these classses. Quick picks for kernels just use these classes and then wrap the result in a quick pick item.


src/client/datascience/common.ts
```diff
@@ -150,37 +149,8 @@ export function generateNewNotebookUri(
...
-    if (result && result.stdout) {
-        const trimmed = result.stdout.trim();
-        if (await fs.localDirectoryExists(trimmed)) {
-            return trimmed;
-        }
-        if (await fs.localFileExists(trimmed)) {
-            return trimmed;
-        }
-    }
+export async function getRealPath(expectedPath: string): Promise<string | undefined> {
```
DonJayamanne, line 152: How about we remove this function completely, feels like an unnecessary wrapper around `fs.realpath`. 2 months from now we look at this code we might not understand why we have a wrapper for a simple function.

src/client/datascience/interactive-common/interactiveBase.ts
```diff
@@ -1230,6 +1230,8 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
                                 this.owningResource
                             )
                             .then(noop, noop);
+                    } else {
```
DonJayamanne, line 1233: Sholdn't we re-throw the error here.
At this point, `askForLocalKernel` has returned `undefined` meaning, the user didn't select a kernel and for some reason kernel startutp failed.

I.e. prefer to throw the same  exception (bubble it all the way up), instead of returning `undefined` & then trying to figure out why we have `undefined` & the like.

src/client/datascience/common.ts
rchiodo: Sure good point.

src/client/datascience/interactive-common/interactiveBase.ts
rchiodo: I had that originally. Throwing here makes the error just show up but then it tries again. Breaking here stops the cycle of asking for a new kernel.

src/client/datascience/jupyter/kernels/kernelDependencyService.ts
```diff
@@ -37,12 +40,44 @@ export class KernelDependencyService implements IKernelDependencyService {
...
         if (await this.areDependenciesInstalled(interpreter, token)) {
-            return KernelInterpreterDependencyResponse.ok;
+            return;
+        }
+
+        // Cache the install run
+        let deferred = this.installPromises.get(interpreter.path);
+        if (!deferred) {
+            deferred = createDeferredFromPromise(this.runInstaller(interpreter, token, disableUI));
+            this.installPromises.set(interpreter.path, deferred);
```
DonJayamanne, line 53:
```suggestion
        // Cache the install run
        let promise = this.installPromises.get(interpreter.path);
        if (!promise) {
            promise = this.runInstaller(interpreter, token, disableUI);
            this.installPromises.set(interpreter.path, promise);
```

DonJayamanne: We can do away with the deferred and jsut cache the promise.

rchiodo: Yeah good point.

src/client/datascience/interactive-common/interactiveBase.ts
DonJayamanne: But what part causes the retry, is it inside this loop?
Why not break out of this loop & then throw the error.
I'm not really a fan of `undefined`, its akin to C++, where we return a bogus value & then call some global function to find out what failed (get error number).
I'd prefer to just bomb out with the error information.

DonJayamanne: Else we end up with checking for `undefined` when it should have ideally just failed with an exceptoni.

rchiodo: No it's not inside this loop. It's outside that retries if there's an exception.

rchiodo: I could change the outside to respond to a canceled exception with a different response.

rchiodo: Althought it's difficult to get in this situation. I have to create a kernel spec that is invalid but is still python.

src/client/datascience/jupyter/jupyterExecution.ts
```diff
@@ -233,30 +225,17 @@ export class JupyterExecutionBase implements IJupyterExecution {
                             traceError('Failed to connect to server', ex);
                             if (ex instanceof JupyterSessionStartError && isLocalConnection && allowUI) {
                                 // Keep retrying, until it works or user cancels.
-                                // Sometimes if a bad kernel is selected, starting a session can fail.
-                                // In such cases we need to let the user know about this and prompt them to select another kernel.
-                                const message = localize.DataScience.sessionStartFailedWithKernel().format(
-                                    getDisplayNameOrNameOfKernelConnection(launchInfo.kernelConnectionMetadata),
-                                    Commands.ViewJupyterOutput
+                                const kernelInterpreter = await this.kernelSelector.askForLocalKernel(
```
DonJayamanne, line 228: Should we just exit out of here and throw an error instead.
In Native Notebooks, this isn't going to work, the kernel tha'ts selected & displayed will be different from whats going to be used. I understand this is how things have worked in webview/interactive.

Ideally at least in native notebooks, we should ask the user to select a different kernel.

Or how about we add a telemetry point to see how often this happens.
Because if this happens in NativeNotebooks, we'll always be saving the wrong kernel in the notebook, i suspect it will be the case even with webviews.
I.e. in notebooks they open, its wrong & gets changed, but old kernel is still selected & they save. They open again, its it using the incorrect kernel.

src/client/datascience/jupyter/jupyterSession.ts
```diff
@@ -267,37 +259,4 @@ export class JupyterSession extends BaseJupyterSession {
             this.outputChannel.appendLine(output);
         }
     }
-
-    private async installDependenciesIntoInterpreter(
```
DonJayamanne, line 271: Who, we have a tonne of these... nice to se them go.

src/client/datascience/jupyter/jupyterExecution.ts
rchiodo: Yeah I'd rather not change that logic yet. I'll add the telemetry point.

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -282,6 +327,95 @@ export function isLocalLaunch(configuration: IConfigurationService) {
...
+                index = i;
+                bestScore = score;
+            }
+        }
+    }
+
+    // If still not found, try languages
+    if (index < 0) {
+        index = kernels.findIndex((k) => {
+            const kernelSpecConnection = k;
```
DonJayamanne, line 406:
```suggestion
        index = kernels.findIndex((kernelSpecConnection) => {
```

src/client/datascience/jupyter/kernels/jupyterKernelService.ts
```diff
@@ -0,0 +1,240 @@
...
+        const token = wrapCancellationTokens(cancelToken, tokenSource.token);
+
+        // If we have an interpreter, make sure it has the correct dependencies installed
+        if (kernel.kind !== 'connectToLiveKernel' && kernel.interpreter) {
+            await this.kernelDependencyService.installMissingDependencies(kernel.interpreter, token, disableUI);
+        }
+
+        // If the spec file doesn't exist or is not defined, we need to register this kernel
+        if (kernel.kind !== 'connectToLiveKernel' && kernel.kernelSpec && kernel.interpreter) {
+            if (!kernel.kernelSpec.specFile || !(await this.fs.localFileExists(kernel.kernelSpec.specFile))) {
```
DonJayamanne, line 66: Awesome..

src/client/datascience/kernel-launcher/remoteKernelFinder.ts
```diff
@@ -0,0 +1,160 @@
...
+        if (preferred >= 0) {
+            return kernels[preferred];
+        }
+    }
+
+    // Talk to the remote server to determine sessions
+    @captureTelemetry(Telemetry.KernelListingPerf)
+    public async listKernels(
+        _resource: Resource,
+        connInfo: INotebookProviderConnection | undefined
```
DonJayamanne, line 79: Would be good to accept a cancellation token here, plenty of promises inside this.

src/client/datascience/kernel-launcher/remoteKernelFinder.ts
```diff
@@ -0,0 +1,160 @@
...
+    @traceDecorators.verbose('Find remote kernel spec')
+    @captureTelemetry(Telemetry.KernelFinderPerf)
+    public async findKernel(
+        resource: Resource,
+        connInfo: INotebookProviderConnection | undefined,
+        option?: nbformat.INotebookMetadata | PythonEnvironment,
+        _cancelToken?: CancellationToken
+    ): Promise<KernelConnectionMetadata | undefined> {
+        // Get list of all of the specs
+        const kernels = await this.listKernels(resource, connInfo);
```
DonJayamanne, line 58: Would be good to pass in the `cancelToken` into this.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -69,136 +75,174 @@ export class KernelFinder implements IKernelFinder {
...
-        await this.readCache();
-
-        const searchBasedOnKernelSpecMetadata = this.findKernelSpecBasedOnKernelSpecMetadata(
-            resource,
-            notebookMetadata && notebookMetadata.kernelspec ? notebookMetadata.kernelspec : undefined
-        );
-
-        if (!notebookMetadata || notebookMetadata.kernelspec || !notebookMetadata.language_info?.name) {
-            return searchBasedOnKernelSpecMetadata;
+    public async getKernelSpecRootPath(): Promise<string | undefined> {
```
DonJayamanne, line 134: This doesn't need to be async.

DonJayamanne: Not sure about these paths, I think the correct ones are the ones in line 35,36,
These to my knowledge are the correct global root paths for kernel spec (based on their docs & what we have used today).
```typescript
const winJupyterPath = path.join('AppData', 'Roaming', 'jupyter', 'kernels');
const linuxJupyterPath = path.join('.local', 'share', 'jupyter', 'kernels');
```

rchiodo: The real path has to be passed back, not the appdata roaming one.

src/client/datascience/kernel-launcher/remoteKernelFinder.ts
rchiodo: Yeah good idea

src/client/datascience/interactive-common/interactiveBase.ts
rchiodo: I can't repro this state anymore. I had a python interpreter that we weren't generating the kernelspec correctly for. Now that I fixed that I can't get this specific thing to happen anymore.

Switching kernels already behaved correctly if there was an error. It's only on starting the notebook that this would happen.

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: Yes, but if you look at line 138, 
we're returning `return path.join('/', 'usr', 'share', 'jupyter', 'kernels');`
when it should be one of the two based on the OS:
```
const linuxJupyterPath = path.join('.local', 'share', 'jupyter', 'kernels');
const macJupyterPath = path.join('Library', 'Jupyter', 'kernels');
```

rchiodo: If you look at line 525 in localkernelfinder.ts you'll see that mac and linux always start with /usr/share/jupyter/kernels. That path is the same between them both.

The docs for jupyter mention the same thing:
https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs

I'll add a comment to that effect.

src/client/datascience/jupyter/kernels/jupyterKernelService.ts
```diff
@@ -0,0 +1,240 @@
...
+                await this.registerKernel(kernel, token);
```
DonJayamanne, line 67: I don't see the need for this when we're using raw & starting Python interpreters.
Else we end up registering kernels unnecessarily.

src/client/datascience/jupyter/kernels/jupyterKernelService.ts
```diff
@@ -0,0 +1,240 @@
...
+            }
+            // Special case. If the original spec file came from an interpreter, we may need to register a kernel
+            else if (kernel.interpreter && kernel.kernelSpec.specFile) {
+                // See if the specfile we started with (which might be the one registered in the interpreter)
+                // doesn't match the name of the spec file
+                if (!kernel.kernelSpec.specFile.includes(kernel.kernelSpec.name)) {
```
DonJayamanne, line 73: I believe this is more accurate & also conveys the actual logic.
```suggestion
                if (path.dirname(kernel.kernelSpec.specFile) !== kernel.kernelSpec.name) {
```

src/client/datascience/jupyter/kernels/jupyterKernelService.ts
```diff
@@ -0,0 +1,240 @@
...
+                    // This means the specfile for the kernelspec will not be found by jupyter. We need to
+                    // register it
+                    await this.registerKernel(kernel, token);
```
DonJayamanne, line 76: Again, i think we should do this only when using Jupyter (not raw)

src/client/datascience/jupyter/kernels/jupyterKernelService.ts
rchiodo: path.dirname isn't just the previous dir. That's why the includes is needed.

src/client/datascience/jupyter/kernels/jupyterKernelService.ts
rchiodo: This code is never used by anything but the jupyterSession. That's why it's named JupyterKernelService.

src/client/datascience/jupyter/kernels/jupyterKernelService.ts
rchiodo: It already does that. Hence the name change.

src/client/datascience/jupyter/kernels/jupyterKernelService.ts
DonJayamanne: Sorry, you can take the name from the dir name Or you can do ends with from dir name.
Includes isn't very clear & doesn't convey the right logic either.

rchiodo: Endswith is clearer. I'll do that.

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: Sure, but this crashes and burns, for me (due to permission error, i cannot write to those directories in a mac).
Besides the old code also uses the path i have mentioned as well, in line 525.
I think reading from there is fine, but when writing (creating the file) on a Mac, we must use the user version as it fails for me, & it could fail for others too.

I think we should do the same for Linux as well, as its possible users won't have permissions to write into the `/usr/...` directory

rchiodo: Ah yeah that makes more sense. User folder should be the default because we should be able to write to it.

src/client/datascience/common.ts
```diff
@@ -150,37 +149,14 @@ export function generateNewNotebookUri(
...
 }
 
-export async function getRealPath(
-    fs: IFileSystem,
-    execFactory: IPythonExecutionFactory,
-    pythonPath: string,
-    expectedPath: string
-): Promise<string | undefined> {
-    if (await fs.localDirectoryExists(expectedPath)) {
+export async function tryGetRealPath(expectedPath: string): Promise<string | undefined> {
```
rchiodo, line 152: I put this back to wrap the nodejs realpath func. It was throwing during the functional tests just searching for jupyter paths.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -0,0 +1,609 @@
...
+            traceError(`List kernels failed: ${e} ${e.stack}`);
+            throw e;
+        }
+    }
+
+    // This should return a WRITABLE place that jupyter will look for a kernel as documented
+    // here: https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs
+    public async getKernelSpecRootPath(): Promise<string | undefined> {
+        if (this.platformService.isWindows) {
+            return tryGetRealPath(path.join(this.pathUtils.home, winJupyterPath));
```
DonJayamanne, line 149: I don't see the need for `getRealPath` here, everything in the paths are fully resolved (no relative paths here).

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -0,0 +1,609 @@
...
+        } else if (this.platformService.isMac) {
+            return tryGetRealPath(path.join(this.pathUtils.home, macJupyterPath));
+        } else {
+            return tryGetRealPath(path.join(this.pathUtils.home, linuxJupyterPath));
```
DonJayamanne, line 153: No need for `getRealPath` for any of these.

src/client/datascience/kernel-launcher/localKernelFinder.ts
rchiodo: win path is where we needed the real path before. Installing jupyter through the store python can redirect the main jupyter location to some goofy store spot.

src/client/datascience/kernel-launcher/localKernelFinder.ts
rchiodo: Yeah probably not for the mac/linux paths.

src/client/datascience/kernel-launcher/localKernelFinder.ts
rchiodo: real path isn't for resolving relative paths. It's for windows redirects.

DonJayamanne: Oh yes i forgot that, the weird python App store installs...
& we do have a few rare symlinks in others as well.

Thanks

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -0,0 +1,629 @@
...
+        @inject(IPathUtils) private readonly pathUtils: IPathUtils,
+        @inject(IExtensionContext) private readonly context: IExtensionContext,
+        @inject(IWorkspaceService) private readonly workspaceService: IWorkspaceService,
+        @inject(IEnvironmentVariablesProvider) private readonly envVarsProvider: IEnvironmentVariablesProvider,
+        @inject(IPythonExtensionChecker) private readonly extensionChecker: IPythonExtensionChecker
+    ) {}
+    @traceDecorators.verbose('Find kernel spec')
+    @captureTelemetry(Telemetry.KernelFinderPerf)
+    public async findKernel(
+        resource: Resource,
```
DavidKutu, line 86: looks like this should be optional

DonJayamanne: I'd leave this as is. `undefined` is supported value. 

This way developer needs to make a decision and explicitly pass the value in, instead of accidentally forgetting it.
Fyi, we have had many issues where these (uri) haven't been passed on when they are optional.

src/client/datascience/interactive-common/interactiveBase.ts
IanMatthewHuff: I'd agree with Don that the undefined makes this function a bit messier with returning. Per your last comment do we still need the undefined state? Sounds like the state might not be a thing anymore. Not the biggest deal, just do agree that "createNotebook => Valid notebook or throw" is better than "createNotebook => Valid notebook, or throw, or ?."

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -186,63 +239,7 @@ export function areKernelConnectionsEqual(
...
-        const model2 = cloneDeep(kernelModel2) as ReadWrite<LiveKernelModel>;
-        model2.env = model1.env || {};
-        model2.metadata = model1.metadata || {};
-        return fastDeepEqual(model1, model2);
-    } else if (!kernelModel1 && !kernelModel2) {
-        return true;
-    } else {
-        return false;
-    }
+    return connection1?.id === connection2?.id;
```
IanMatthewHuff, line 242: Yay!

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -282,6 +279,113 @@ export function isLocalLaunch(configuration: IConfigurationService) {
...
+                    spec.path === interpreter.path &&
+                    nbMetadataLanguage === PYTHON_LANGUAGE
+                ) {
+                    // Path match. This is worth more if no notebook metadata as that should
+                    // match first.
+                    score += notebookMetadata ? 1 : 8;
+                }
+
+                // See if the version is the same
+                if (interpreter && interpreter.version && spec && spec.name && nbMetadataLanguage === PYTHON_LANGUAGE) {
```
IanMatthewHuff, line 334: I think this resolves #4941. I was about to take a look at this but figured that I should wait until the refactor. Looks like it should match on a python3 spec name to prefer a python3 interpreter over plain python.

src/client/datascience/jupyter/kernels/jupyterKernelService.ts
```diff
@@ -0,0 +1,240 @@
...
+        // If the spec file doesn't exist or is not defined, we need to register this kernel
+        if (kernel.kind !== 'connectToLiveKernel' && kernel.kernelSpec && kernel.interpreter) {
+            if (!kernel.kernelSpec.specFile || !(await this.fs.localFileExists(kernel.kernelSpec.specFile))) {
+                await this.registerKernel(kernel, token);
+            }
+            // Special case. If the original spec file came from an interpreter, we may need to register a kernel
+            else if (kernel.interpreter && kernel.kernelSpec.specFile) {
+                // See if the specfile we started with (which might be the one registered in the interpreter)
+                // doesn't match the name of the spec file
+                if (!path.dirname(kernel.kernelSpec.specFile).endsWith(kernel.kernelSpec.name)) {
```
IanMatthewHuff, line 73: Not sure if this would manifest as a bug. But it looks like there might be cases where it's not right due to use of endsWith. As I recall the final path segment is the name identifier for a kernelSpec. But this would match like so:

foo/bar/funkybaz === 'baz' would match due to endsWith

I think that we'd want an explicit equals on the last path segment, not a endsWith on the entire path.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -0,0 +1,629 @@
...
+                a.kernelSpec?.display_name === activeInterpreter?.displayName
+            ) {
+                return -1;
+            } else {
+                return 1;
+            }
+        });
+    }
+
+    private findMatchingInterpreters(
```
IanMatthewHuff, line 252: This looks like another spot where (from looking at customer logs) adding logging in the new extra kernel logging option might be helpful.

IanMatthewHuff (approved):
Took a while to get through. Looks like a solid improvement. Cleans up some of our messiest code areas. Had a few suggestions, mostly on some areas where I think that more logging (gated on your new env var) might be helpful.

src/client/datascience/types.ts
```diff
@@ -379,6 +380,7 @@ export interface IJupyterSessionManager extends IAsyncDisposable {
     readonly onRestartSessionCreated: Event<Kernel.IKernelConnection>;
     readonly onRestartSessionUsed: Event<Kernel.IKernelConnection>;
     startNew(
+        resource: Resource,
```
DonJayamanne, line 383: Nice to see this being sent

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -282,6 +279,113 @@ export function isLocalLaunch(configuration: IConfigurationService) {
     return false;
 }
 
+export function findPreferredKernelIndex(
```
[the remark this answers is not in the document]
rchiodo: Good idea. I'll change it.

src/client/datascience/jupyter/kernels/jupyterKernelService.ts
rchiodo: Okay I'll do that.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -0,0 +1,629 @@
...
+        @inject(IExtensionContext) private readonly context: IExtensionContext,
+        @inject(IWorkspaceService) private readonly workspaceService: IWorkspaceService,
+        @inject(IEnvironmentVariablesProvider) private readonly envVarsProvider: IEnvironmentVariablesProvider,
+        @inject(IPythonExtensionChecker) private readonly extensionChecker: IPythonExtensionChecker
+    ) {}
+    @traceDecorators.verbose('Find kernel spec')
+    @captureTelemetry(Telemetry.KernelFinderPerf)
+    public async findKernel(
+        resource: Resource,
+        option?: nbformat.INotebookMetadata | PythonEnvironment,
```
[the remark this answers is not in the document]
rchiodo: Let me double check. Maybe it's only in test code now

src/client/datascience/kernel-launcher/localKernelFinder.ts
rchiodo: I'll add some logging here.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -0,0 +1,654 @@
...
+                        );
+                        return items;
+                    })
+                );
+            }
+
+            this.writeCache().ignoreErrors();
+
+            // ! as the has and set above verify that we have a return here
+            return await this.workspaceToMetadata.get(workspaceFolderId)!;
```
joyceerhl, line 153: Why do we need to do an await here?

rchiodo: Otherwise the catch doesn't work. The promise is not waited on here, but rather on the caller.

src/client/datascience/kernel-launcher/localKernelFinder.ts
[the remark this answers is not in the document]
rchiodo: Yep don't need it. I removed it as suggested.

src/client/datascience/kernel-launcher/localKernelFinder.ts
joyceerhl: Thanks for explaining 😊

rchiodo: This is an easy pattern to miss. (I've had this problem during this refactor in fact).

If you have something like so:

```typescript
try {
    return functionThatReturnsAPromise();
} catch (e) {
   trace(`Swallowing e: ${e}`);
}
```

The catch will never be hit. That's because the promise is returned. Without the await the potential rejection is never thrown in this function but rather in the first spot that awaits the promise.

src/client/datascience/kernel-launcher/remoteKernelFinder.ts
```diff
@@ -0,0 +1,167 @@
+// This class searches for a kernel that matches the given kernel name.
```
joyceerhl, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.
'use strict';


```

src/client/datascience/kernel-launcher/remoteKernelFinder.ts
```diff
@@ -0,0 +1,167 @@
...
+// First it searches on a global persistent state, then on the installed python interpreters,
```
joyceerhl, line 2: Ignore my previous suggestion if you saw it before it was deleted.

I believe this file needs to start with 

```
'use strict';

```

and this comment is continued on line 30 so should move down there

rchiodo: Will do

src/test/datascience/jupyter/interpreter/jupyterInterpreterSubCommandExecutionService.unit.test.ts
```diff
@@ -84,6 +87,9 @@ suite('DataScience - Jupyter InterpreterSubCommandExecutionService', () => {
         when(interperterService.getActiveInterpreter()).thenResolve(activePythonInterpreter);
         when(interperterService.getActiveInterpreter(undefined)).thenResolve(activePythonInterpreter);
     });
+    teardown(() => {
+        sinon.restore();
```
joyceerhl, line 91: Apparently `sinon.restore` is [deprecated and internal](https://github.com/sinonjs/sinon/issues/1390). We're supposed to create a sinon.Sandbox instance and create the stub with `sandbox.stub`, then restore with `sandbox.restore`. I guess if this works it's fine though?

joyceerhl (approved):

Issue #4999, opened 2021-03-02
Add telemetry to check how often we fail to start the right kernel


DonJayamanne:
For local connections we should be able to check if we have started the right Python kernel.
We seem to be having a lot of issues where we're starting the wrong kernel.
Telemetry to track whether its correct or not, just compare the python paths in the interpreter & kernel.

IanMatthewHuff:
Validated.

PR #5000, opened 2021-03-02
Track failure to start right kernel interpreter

DonJayamanne:
For #4999
Added this as we're getting a lot of these lately (conda, etc & this PR makes it very easy to tell how often we get this right, without having to wait for users to file issues).

### DIFF
--- a/src/client/datascience/jupyter/kernels/helpers.ts
+++ b/src/client/datascience/jupyter/kernels/helpers.ts
@@ -2,9 +2,10 @@
 // Licensed under the MIT License.
 'use strict';
 
+import * as uuid from 'uuid/v4';
 import type { Kernel } from '@jupyterlab/services';
 import * as fastDeepEqual from 'fast-deep-equal';
-import { IJupyterKernelSpec } from '../../types';
+import { IJupyterKernelSpec, INotebook } from '../../types';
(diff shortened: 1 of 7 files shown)


src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -278,3 +281,35 @@ export function isLocalLaunch(configuration: IConfigurationService) {
...
+        if (cells.length === 0 || !Array.isArray(cells[0].data.outputs) || cells[0].data.outputs.length === 0) {
+            return;
+        }
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const output: nbformat.IStream = cells[0].data.outputs[0] as any;
+        if (output.name !== 'stdout' && output.output_type !== 'stream') {
+            return;
+        }
+        const sysExecutable = concatMultilineString(output.text).trim().toLowerCase();
+        const match = kernelConnection.interpreter.path.toLowerCase() === sysExecutable;
```
DonJayamanne, line 307: We don't care about case sensitivty for telemetry, 99.9% of the time its unlikely users will have two interpreters with a difference only in the case of the environment name.
I don't think conda would support that, & only applies to venvs, hence don't see the point in supporting that for telemetry.

rchiodo (approved):

PR #5005, opened 2021-03-02
Wait for experiments library to complete loading

DonJayamanne:


rchiodo (approved):

src/client/common/experiments/service.ts
```diff
@@ -113,13 +118,15 @@ export class ExperimentService implements IExperimentService {
         // so we need to perform these checks and send the corresponding telemetry manually.
         switch (this.getOptInOptOutStatus(experiment)) {
             case 'optOut': {
+                await this.experimentationService.isCachedFlightEnabled(experiment);
```
DonJayamanne, line 121: Regardless of whether user has opted in/out manually, we need to check with the experiment service.
This way if a user falls into an experiment, the telemetry will line up with experiments on server.

DonJayamanne: @jakebailey @kimadeline /cc

DonJayamanne: @rchiodo Please re-review.

src/client/common/experiments/service.ts
```diff
@@ -120,6 +125,7 @@ export class ExperimentService implements IExperimentService {
                 return false;
             }
             case 'optIn': {
+                await this.experimentationService.isCachedFlightEnabled(experiment);
```
DonJayamanne, line 128: Regardless of whether user has opted in manually, we need to check with the experiment service.
This way if a user falls into an experiment, the telemetry will line up with experiments on server.

kimadeline (approved):

DavidKutu (approved):

joyceerhl (approved):

Issue #4562, opened 2021-01-29
VariableView goes away if you switch off of an active notebook context


IanMatthewHuff:
Right now the VariableView is keyed to having an active notebook context. And it's keyed to being unavailable without one. This leads to some UI annoyance with it. Take the following scenario.

Open a notebook A
Execute it
Open the variable panel
Open the data viewer (or move to a native notebook that has not been executed yet)
Move back to notebook A in the editor

The variable view is still available and has the right info. But since we moved to a context that hid it, it's not open by default when you move back to the document.

I think there is an easy fix here. Instead of enablement of the view when there is an active notebook running, it should just be enabled when there is a native notebook editor (with or without an active notebook / kernel) open at all in the editor.

DavidKutu:
Its extra annoying if you open settings.

joyceerhl:
Validated. Opened a blank notebook, created some variables, opened the data viewer and verified that the variable view stayed open, opened another blank notebook, created some variables, verified that variable view updated to match the active notebook, switched back to the first blank notebook, verified that variable view updated, and verified that the variable view only went away after closing all notebooks.

joyceerhl:
FYI @DavidKutu I'm still seeing the leading colon :( 

![image](https://user-images.githubusercontent.com/30305945/110262477-ae263d00-7f68-11eb-8834-3ec50306e83b.png)

Issue #4458, opened 2021-01-22
Jupyter Variables tab is not named Jupyter Variables until you click on it


rchiodo:
It was confusing to me what it was for.

![image.png](https://images.zenhubusercontent.com/5bda265b648e32720db70349/a24699cd-51ca-4def-9e69-44ebd99e9973)

DavidKutu:
I can't get a repro of this.

joyceerhl:
Validated, although FYI @DavidKutu I'm still seeing the leading colon :( 
![image](https://user-images.githubusercontent.com/30305945/110262607-0f4e1080-7f69-11eb-9074-572617ba2b41.png)

PR #5006, opened 2021-03-03
Variable view fixes

DavidKutu:
- variable view will stay as long as you have a notebook open (not necesarily active)
- jupyter variables tab will always be named 'Jupyter: Variables'

For #4458, #4562

The new behavior for the Variable View is this:
- It will stay as long as you have a native notebook opened, it goes away if you close them all.
- It doesn't update if you go to another editor (text, interactive window, data viewer, settings), only when you go to another notebook.

It has some inconveniences:

1.
- Open a notebook A, run it and look at the variables
- Open the data viewer
- Open a notebook B, run it and look at the variables
- Go back to the data viewer
- You're now looking at the data viewer with a variable from notebook A, while the variable view shows the variables from notebook B

2.
- Open a notebook A, run it and look at the variables
- Open the interactive window
- Open the variable view in the interactive window
- You have two variable views open, some users might get confused

The first case feels rare enough to don't matter, and the second one could be done with the previews behavior, but it was less likely.


package.json
```diff
@@ -1832,7 +1832,7 @@
             "panel": [
                 {
                     "id": "jupyter-variables",
-                    "title": "Jupyter",
+                    "title": "",
```
rchiodo, line 1835: This seems weird to me. If I drag the panel then it won't have a name anymore.

joyceerhl, line 1835: Are there any side effects of this? Like the tooltip for the notebook icon when the variable explorer panel is docked in the sidebar might be blank if you don't specify a title?

joyceerhl (approved):

package.json
joyceerhl: We raced 😂

news/2 Fixes/4562.md
```diff
@@ -0,0 +1 @@
+Variable view will stay as long as you have a notebook open (not necesarily active).
```
joyceerhl, line 1:
```suggestion
Variable view will stay as long as you have a notebook open (not necessarily active).
```

package.json
DavidKutu: image

It still has the full name, and for @joyceerhl, the tooltip is still there.

joyceerhl (changes requested):
The package.json change won't work. There's a leading colon in the title bar, and if you hover over the notebook icon in the sidebar, there's no tooltip. All other icons do have a tooltip.

image

DavidKutu:
@joyceerhl I don't get the colon, and do get the tooltip
image

joyceerhl:
I checked out your branch and built it, and I get the leading colon even in the terminal panel... wat?

image

DavidKutu:
I don't get that either
image

joyceerhl (approved):
🤷‍♀️ I don't know why I'm getting something different than David when building this branch locally, but I'm not going to block on this. If there's an issue we'll catch it at validate/bug bash.

package.json
IanMatthewHuff: So the ":" in the name was actually just an effect of having both a view container name and a view name. I didn't explicitly add the ":". So if we go with this approach of not giving the view a name, then we could probably just call it "Jupyter Variables" the ":" isn't needed.

IanMatthewHuff: As I recall with the old UI there were cases where the view container had loaded, but not the view itself. Might mean that for a brief period there is some type of empty, nameless view in the UI. At the same time, I don't know of any specific issues with this approach, so doesn't seem like the worst thing in the world to give it a try and see if anything shakes out.

package.json
```diff
@@ -1842,8 +1842,8 @@
                 {
                     "type": "webview",
                     "id": "jupyterViewVariables",
-                    "name": "Variables",
-                    "when": "jupyter.isvscodenotebookactive"
+                    "name": "Jupyter: Variables",
```
IanMatthewHuff, line 1845: Per comment above if we go with this we can just change to "Jupyter Variables" ":" was added by the view container / view separation.

src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -146,6 +148,8 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
         this.updateMergedContexts();
     }
     private updateNativeNotebookContext() {
+        this.hasNativeNotebookOpen.set(this.vscNotebook.notebookDocuments.length > 0).ignoreErrors();
```
IanMatthewHuff, line 151: There is actually an issue here, I believe, and it's a bit of an issue with exiting code as well, so I want to ping @DonJayamanne to check me on this. It looks like in spots like this function both with the added code here and with the existing code that we are looking at the vscNotebook interface. But that interface is just a wrapper around the VS Code interface. We'd be updating our can run above or notebook open context if a native notebook for something like the Markdown notebooks, or .NET Interactive notebooks are open. I think that we might want to watch some events off of the VS Code interface, but for checking a context like this we'd want to be looking at the active .editors off of our Editor provider (notebookEditorProvider) and then checking for any that were of type === 'native'. That would restrict it to just the notebook files that we know about and deal with. And it looks like we should make that change at more places in this file.

IanMatthewHuff (changes requested):
Bit of an open question regarding our usage of the VSCode API to find open editors.

src/client/datascience/commands/activeEditorContext.ts
DavidKutu: Done

src/client/datascience/interactive-ipynb/nativeEditor.ts
```diff
@@ -101,6 +101,10 @@ export class NativeEditor extends InteractiveBase implements INotebookEditor {
         return this.viewState.active;
     }
 
+    public get selection() {
```
DonJayamanne, line 104: Feels unnecessary to add more types into our editor when we do'nt need this and don't support it.
Why not just use the VSC Notebook API directly (instead of making our stuff look like its a VS Code notebook)

DonJayamanne (changes requested):
All ok, but not sure why we're extending our API to make it look like VS Code API.
The goal is to eventually move away from our Notebook Editor, this makes it looks like we want to keep wrapping VSC API with our classes.. seems unnecessary

src/client/datascience/commands/activeEditorContext.ts
DonJayamanne: Oops, i missed this message

DavidKutu:
@DonJayamanne that was because of @IanMatthewHuff's comment here:
https://github.com/microsoft/vscode-jupyter/pull/5006#discussion_r586582009

src/client/datascience/commands/activeEditorContext.ts
DonJayamanne: Yes, i agree with @IanMatthewHuff

DonJayamanne:
@DavidKutu 
I think you have misunderstood @IanMatthewHuff comments.
I think his sugestion was as follows
```
 private updateNativeNotebookContext() {
        this.hasNativeNotebookOpen.set(this.vscNotebook.notebookDocuments.some(isJupyterNotebook)).ignoreErrors();
```

I.e. ensure the flag is set only if the notebooks are ones that we opened i.e. viewType === ''jupyter-notebook'`

@IanMatthewHuff please correct me if i'm wrong.

IanMatthewHuff:
@DonJayamanne I actually did suggesting looking for everything that was 'native' off of the editor provider. Figured that would give us all the native editors that were ours that were open. However your suggestion is better with just checking for jupyter-notebook type off the vsc api. Forgot that we could do that. And I don't want to have to add our own selection on our type.

DavidKutu:
@DonJayamanne @IanMatthewHuff 
Done

IanMatthewHuff (approved):

DonJayamanne (approved):

PR #5007, opened 2021-03-03
Always show data viewer option in debug variables context menu

joyceerhl:
For https://github.com/microsoft/vscode-python/issues/15529

### DIFF
--- a/package.json
+++ b/package.json
@@ -1321,7 +1321,7 @@
                 {
                     "command": "jupyter.showDataViewer",
                     "group": "1_view",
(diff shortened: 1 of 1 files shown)


DavidKutu (approved):

DonJayamanne (approved):

DonJayamanne:
Do we need a change log entry for this, as this is no longer an experiment.

joyceerhl:
@DonJayamanne not sure. We never announced this in the changelog but it has already been announced in the January blog post so I didn't think it was worth mentioning again. I can add it if you think we should though.

PR #5011, opened 2021-03-03
Dont prewarm global interpreter if ZMQ is supported

IanMatthewHuff:
For #5009


IanMatthewHuff:
I ran into this while investigating conda issues. I don't believe this was causing any user visible issues, but it looked like a quick fix so I just clean it up and submitted it.

Basically we have code run on activation that prewarms the env vars for the current global jupyter interpreter. However this is not needed at all for the raw kernel case. So in those cases we'll run python code (to check dependencies, so "import jupyter" "jupyter kernelspec list" ect...) every time our extension activates. Even if our global jupyter interpreter is not set this runs as it considers the current python interpreter as a candidate for the global jupyter interpreter. It was running ignore errors in the background so I don't think that it was causing the conda issue, but it was unneeded extra work / complexity when it's not even used for raw kernel.

rchiodo (approved):

PR #5023, opened 2021-03-03
Upload build early without waiting for tests on release

DonJayamanne:
This is safe at least in this branch, we check the tests & we manually test.


rchiodo:
Maybe eliminate the code coverage step too?

PR #5026, opened 2021-03-03
Change version to 1.55

DonJayamanne:


rchiodo:
I don't think we want to do this until we need to.

rchiodo (changes requested):
⏲️

DonJayamanne:
> I don't think we want to do this until we need to.

How else do we have an insider build of the extension?
Not going to merge this until VS Code releases the new inisder build for 1.54, 
But if we don't do this, i do'nt see how out insiders of the extension will continue?

@rchiodo

rchiodo:
```^1.54.0-insiders``` should pick ```1.55.0-insiders``` too. I don't believe we need to keep updating it every time they up the version.

DonJayamanne:
> d pick 1.55.0-insiders too. I don't believe we need to keep updating it every time they up the version.

I just updated to the latest insiders & its downloading the latest stable version of the extension.
Hence we have no choice but to update this.

rchiodo (changes requested):
The real problem should be the missing ^, not the version number.

You really only need to change it to this:

```^1.54.0-insiders```

DonJayamanne:
> You really only need to change it to this:
> ^1.54.0-insiders

* We don't want users on older version of insiders installing this new version of the extesnion that relies on new API (it will break)
* Tomorrow when we release a new version we're going to have to change this again to ensure it ONLY works in the latest stable version, which would be `1.55.0` (when VSC adds more changes to the API).

I.e. I cannot leave it as `^1.54-insiders`.

rchiodo:
> > You really only need to change it to this:
> > ^1.54.0-insiders
> 
> * We don't want users on older version of insiders installing this new version of the extesnion that relies on new API (it will break)
> * Tomorrow when we release a new version we're going to have to change this again to ensure it ONLY works in the latest stable version, which would be `1.55.0` (when VSC adds more changes to the API).
> 
> I.e. I cannot leave it as `^1.54-insiders`.

Are you saying 1.55 has changes in it already that we need to consume? I didn't think there were any yet.

DonJayamanne:
> u saying 1.55 has changes in it already that we need to consume? I didn't think there were any yet.

Yes, here it is https://github.com/microsoft/vscode-jupyter/pull/5047

rchiodo (approved):

DonJayamanne:
Closing in favor of https://github.com/microsoft/vscode-jupyter/pull/5047

PR #5047, opened 2021-03-05
Account forAPI  changes in display order of output

DonJayamanne:


package.json
```diff
@@ -24,7 +24,7 @@
         "theme": "light"
     },
     "engines": {
-        "vscode": "1.54.0-insider"
+        "vscode": "1.55.0-insider"
```
rchiodo, line 27: Should this have the ```^```?

src/client/datascience/notebook/helpers/helpers.ts
```diff
@@ -346,13 +330,39 @@ function createNotebookCellDataFromCodeCell(cell: nbformat.ICodeCell, cellLangua
...
+    return outputItems.sort((outputItemA, outputItemB) => {
+        const isMimeTypeMatch = (value: string, compareWith: string) => {
+            if (value.endsWith('.*')) {
+                value = value.substr(0, value.indexOf('.*'));
+            }
+            return compareWith.startsWith(value);
+        };
+        const indexOfMimeTypeA = orderOfMimeTypes.findIndex((mime) => isMimeTypeMatch(outputItemA.mime, mime));
+        const indexOfMimeTypeB = orderOfMimeTypes.findIndex((mime) => isMimeTypeMatch(outputItemB.mime, mime));
+        if (indexOfMimeTypeA === indexOfMimeTypeB) {
```
rchiodo, line 361: I believe you could just return 'indexOfMimeTypeA' - 'indexOfMimeTypeB'

rchiodo (approved):

src/client/datascience/notebook/helpers/helpers.ts
DonJayamanne: lol, yes

package.json
DonJayamanne: Not sure we want that,
we want this to be only compatible with 1.55.0-insiders & nothing else.
E.g. if user has `1.55.0` stable, that should not be used.
The API is currently locked into a specific version of VS Code.

DonJayamanne: Also we dno't want insider users in `1.56.0-insiders` to get this VSIX.

DonJayamanne: Let me know what you think about that.

DavidKutu (approved):

PR #5055, opened 2021-03-06
More data viewer style tweaks

joyceerhl:
Found a few issues while testing with different themes, plus some more tweaks to make the data viewer look like the figma mockups.

image


src/datascience-ui/data-explorer/sliceControl.tsx
```diff
@@ -108,16 +124,12 @@ export class SliceControl extends React.Component<ISliceControlProps, ISliceCont
                     </summary>
                     <form onSubmit={this.handleSubmit} className="slice-form">
                         <div className="slice-enablement-checkbox-container">
-                            <input
-                                type="checkbox"
-                                id="slice-enablement-checkbox"
-                                className="slice-enablement-checkbox"
+                            <Checkbox
+                                styles={checkboxStyles}
```
joyceerhl, line 128: Figured out how to style FluentUI controls without hacks, went back and changed this to a FluentUI checkbox

src/datascience-ui/data-explorer/cellFormatter.tsx
```diff
@@ -24,17 +24,14 @@ class CellFormatter extends React.Component<ICellFormatterProps> {
...
-
                 case ColumnType.Number:
                     return this.renderNumber(this.props.value as number);
 
                 default:
                     break;
             }
         }
-        // Otherwise an unknown type or a string
+        // Otherwise an unknown type, boolean, or a string
```
IanMatthewHuff, line 34: Yay for updating comments. thanks.

src/datascience-ui/data-explorer/reactSlickGrid.css
```diff
@@ -76,21 +70,31 @@
...
 .react-grid-header-cell > .slick-sort-indicator-asc::before {
     background: none;
-    font: normal normal normal 16px/1 codicon;
+    font: normal normal normal 10px/1 codicon;
     content: '\eaa1'; /* VS Code arrow-up codicon */
     align-items: center;
+    text-align: right;
+    display: flex;
+    padding: 3px;
+    color: var(--vscode-settings-textInputForeground);
```
IanMatthewHuff, line 84: Bit hard to track through all the CSS, but this color looks a bit suspicious to me. It's from the GUI settings editor specifically.
https://code.visualstudio.com/api/references/theme-color#settings-editor-colors
I would expect it only to be used alongside textInputBackground which I'm not currently seeing. Using it against other background could lead to accessibility issues.

IanMatthewHuff: Whoops, looks like the background is added later in this PR. This should be ok then as long at this icon is showing up under .slice-data. Is that the case?

IanMatthewHuff (approved):
Approved as long as everything using textInputForeground is under textInputBackground elements.

rchiodo (approved):

src/datascience-ui/data-explorer/reactSlickGrid.css
joyceerhl: Hm this is actually used against --vscode-menu-background and it's the token used in the figma mockups

joyceerhl: Switching to menu-foreground after discussing with Jill. Eventually I think we want to use whatever the new table widget is using, but I'll wait till the design for that has stabilized.

IanMatthewHuff: Awesome. Thanks for following up with Jill. These color usage mismatches used to cause lots of issue back in Visual Studio so I try to push being strict on them as much as possible.

DonJayamanne (approved):

src/datascience-ui/data-explorer/reactSlickGrid.css
joyceerhl: I did another pass and found a few more token mismatches for foreground vs background, so fixed those too :)

Issue #4724, opened 2021-02-09
Commands that require installing the python extension spin forever if you don't install it


IanMatthewHuff:
1. Remove python extension
2. Open a notebook
3. Export notebook to HTML
4. When the Python extension page comes up, just close it

Result exporting message never goes away. 
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/624b8935-647d-48ee-9d30-894f2558d581)

PR #5086, opened 2021-03-09
Python api can fail to init if user decides not to install the extension

DavidKutu:
For #4724


src/client/api/pythonApi.ts
```diff
@@ -107,7 +115,7 @@ export class PythonExtensionChecker implements IPythonExtensionChecker {
         return this.extensions.getExtension(this.pythonExtensionId) !== undefined;
     }
 
-    public async showPythonExtensionInstallRequiredPrompt(): Promise<void> {
+    public async showPythonExtensionInstallRequiredPrompt(): Promise<boolean | void> {
```
joyceerhl, line 118: It's not immediately obvious to me what a boolean or void return value means for this function. A comment would be nice, or better yet an enum if these three states are distinct.

src/client/datascience/jupyter/interpreter/jupyterInterpreterService.ts
```diff
@@ -59,15 +59,15 @@ export class JupyterInterpreterService {
...
-        if (!this.getInitialInterpreterPromise) {
-            this.getInitialInterpreterPromise = this.getInitialInterpreterImpl(token).then((result) => {
-                // Set ourselves as a valid interpreter if we found something
-                if (result) {
-                    this.changeSelectedInterpreterProperty(result);
-                }
-                return result;
-            });
-        }
+        // if (!this.getInitialInterpreterPromise) {
```
joyceerhl, line 62: Should this line be deleted altogether?

rchiodo (approved):
Looks good except for the <boolean | void> thing. That's really weird.

src/client/datascience/jupyter/interpreter/jupyterInterpreterService.ts
DavidKutu: That should definitely be deleted XD

src/client/api/pythonApi.ts
```diff
@@ -57,9 +57,11 @@ export class PythonApiProvider implements IPythonApiProvider {
         @inject(IPythonExtensionChecker) private extensionChecker: IPythonExtensionChecker
     ) {}
 
-    public getApi(): Promise<PythonApi> {
-        this.init().catch(noop);
-        return this.api.promise;
+    public async getApi(): Promise<PythonApi | undefined> {
+        if (await this.init().catch(noop)) {
```
DonJayamanne, line 61: Why are we swallowing the exceptions?

src/client/api/pythonApi.ts
```diff
@@ -107,7 +115,7 @@ export class PythonExtensionChecker implements IPythonExtensionChecker {
         return this.extensions.getExtension(this.pythonExtensionId) !== undefined;
     }
 
-    public async showPythonExtensionInstallRequiredPrompt(): Promise<void> {
+    public async showPythonExtensionInstallRequiredPrompt(): Promise<boolean | void> {
         if (this.waitingOnInstallPrompt) {
             return this.waitingOnInstallPrompt;
```
DonJayamanne, line 120: I'd try to go with two states instead of an enum. And ensure we return the right value here as well.

src/client/datascience/interactive-ipynb/nativeEditor.ts
```diff
@@ -811,7 +811,8 @@ export class NativeEditor extends InteractiveBase implements INotebookEditor {
     private async exportAs(): Promise<void> {
         // Export requires the python extension
         if (!this.extensionChecker.isPythonExtensionInstalled) {
-            return this.extensionChecker.showPythonExtensionInstallRequiredPrompt();
+            await this.extensionChecker.showPythonExtensionInstallRequiredPrompt();
```
DonJayamanne, line 814: This change was not necessary. but not saying it needs to be reverted either.

src/client/datascience/jupyter/interpreter/jupyterInterpreterService.ts
```diff
@@ -59,15 +59,15 @@ export class JupyterInterpreterService {
...
+        this.getInitialInterpreterPromise = this.getInitialInterpreterImpl(token).then((result) => {
+            // Set ourselves as a valid interpreter if we found something
+            if (result) {
+                this.changeSelectedInterpreterProperty(result);
```
DonJayamanne, line 66: With the new change, this method will get executed everytime, previously we'd cache the result.
I don't think thats the desired effect.

src/client/api/pythonApi.ts
DonJayamanne: We don't need void, we update `waitingOnInstallPrompt` to contain the result of the installation. 
Then we'd only have a boolean, though I'd still prefer the enum (we can use InstallerResponse)

DonJayamanne (changes requested):

src/client/datascience/jupyter/interpreter/jupyterInterpreterService.ts
DavidKutu: Caching the result was the problem, if you said don't install, then the export button won't ask again and automatically fail. When it should ask again.

src/client/api/types.ts
```diff
@@ -19,13 +19,13 @@ export interface ILanguageServer extends Disposable {
...
 export interface IPythonApiProvider {
-    getApi(): Promise<PythonApi>;
+    getApi(): Promise<PythonApi | undefined>;
     setApi(api: PythonApi): void;
 }
 export const IPythonExtensionChecker = Symbol('IPythonExtensionChecker');
 export interface IPythonExtensionChecker {
     readonly isPythonExtensionInstalled: boolean;
-    showPythonExtensionInstallRequiredPrompt(): Promise<void>;
+    showPythonExtensionInstallRequiredPrompt(): Promise<boolean>;
```
DonJayamanne, line 28: Can we use `InstallerResponse`, as the boolean still doesn't tell us much.
Does it mean it was installed or a message was displayed?

DonJayamanne (changes requested):
Please can we use InstallerResponse instead of boolean see here https://github.com/microsoft/vscode-jupyter/pull/5086#discussion_r590698240

src/client/api/pythonApi.ts
```diff
@@ -107,17 +115,19 @@ export class PythonExtensionChecker implements IPythonExtensionChecker {
         return this.extensions.getExtension(this.pythonExtensionId) !== undefined;
     }
 
-    public async showPythonExtensionInstallRequiredPrompt(): Promise<void> {
+    public async showPythonExtensionInstallRequiredPrompt(): Promise<InstallerResponse> {
         if (this.waitingOnInstallPrompt) {
-            return this.waitingOnInstallPrompt;
+            return InstallerResponse.Waiting;
```
DonJayamanne, line 120: Why can you just return `this.waitingOnInstallPrompt`

& have `this.waitingOnInstallPrompt` get resolved with the value of Installerresponse.

DonJayamanne: Basically returning `Waiting` is unnecessary, we can just wait on this promise, then its not `Waiting` anymore and we'd have a result.

DonJayamanne (changes requested):

DonJayamanne (approved):

Issue #4976, opened 2021-02-28
vscode  does not work with jupyter /Cannot read property 'then' of undefined /jupyter server is not starting


pleabargain:
## Environment data

-   VS Code version: - Version: 1.53.2
Commit: 622cb03f7e070a9670c94bae1a45d78d7181fbd4
Date: 2021-02-11T11:48:44.518Z
Electron: 11.2.1
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Linux x64 5.4.0-66-generic snap
-   Jupyter Extension version (available under the Extensions sidebar): XXX
- Jupyter
ms-toolsai.jupyter
Microsoft
9,948,266
Repository
License
v2021.2.603412351
Jupyter notebook support, interactive programming and computing that supports Intellisense, debugging and more.

-   Python Extension version (available under the Extensions sidebar): XXX
- Python
ms-python.python
Microsoft
v2021.2.582707922
-   OS  Linux distro) and version:ubuntu 20.04
- 
- python3 --version
Python 3.8.5
-   Type of virtual environment used N/A 
-   Jupyter server running: Local 
jupyter server is not starting!

## Expected behaviour
ctrl + enter to run cell
run code

## Actual behaviour
error:
Cannot read property 'then' of undefined /jupyter server is not starting

## Steps to reproduce:

a = "test"
print(a)

no output/ jupyter server does not start
https://i.imgur.com/dHcMDYm.png

![vscode does not work with jupyterPeek 2021-02-28 15-37](https://user-images.githubusercontent.com/640846/109420447-39d52200-79db-11eb-9932-65d992420532.gif)

joyceerhl:
Hi @pleabargain sorry for the trouble. You appear to be clicking on the local/remote kernel picker. Could you instead try clicking on the rightmost text which says 'Not started' in your GIF above? That should allow you to pick a working kernel.

Additionally, could you please copy and paste debug logs from the Jupyter output tab into a new comment below? This will help us identify why the kernel is failing to start for you. To get full logs, add the following setting to your user settings.json:
>"jupyter.logging.level": "debug"

Then reload VSCode for the new setting to take effect, and you'll be able to find debug logs in the Jupyter output tab:
![image](https://user-images.githubusercontent.com/30305945/109401262-8b5dae00-7902-11eb-8b4e-638a31e7de16.png)

Thank you!

pleabargain:
[...]

PR #5091, opened 2021-03-09
Fix issue related to starting Jupyter Server

DonJayamanne:
For #4976

### DIFF
--- a/src/client/datascience/kernel-launcher/localKernelFinder.ts
+++ b/src/client/datascience/kernel-launcher/localKernelFinder.ts
@@ -130,7 +130,7 @@ export class LocalKernelFinder implements ILocalKernelFinder {
                 ) || 'root';
 
             // If we have not already searched for this resource, then generate the search
-            if (workspaceFolderId && !this.workspaceToMetadata.has(workspaceFolderId)) {
+            if (!this.workspaceToMetadata.has(workspaceFolderId)) {
                 this.workspaceToMetadata.set(
                     workspaceFolderId,
                     this.findResourceKernelMetadata(resource, cancelToken).then((items) => {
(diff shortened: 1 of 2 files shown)


DavidKutu (approved):

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -130,7 +130,7 @@ export class LocalKernelFinder implements ILocalKernelFinder {
                 ) || 'root';
 
             // If we have not already searched for this resource, then generate the search
-            if (workspaceFolderId && !this.workspaceToMetadata.has(workspaceFolderId)) {
+            if (!this.workspaceToMetadata.has(workspaceFolderId)) {
```
IanMatthewHuff, line 133: Why did this removal change anything? From the assignment above getWorkspaceFolderIdentifier should return a string. And if the string was '' it should have used 'root', right?

DonJayamanne: Nope, if `worksapceFolderId` is empty, and if user doesn't have a workspace folder open, then it will result in an empty string '', and under that condition we don't have anything, and later in the code we fall over.

This is where things break (line 153)
```return await this.workspaceToMetadata.get(workspaceFolderId)!;```

DonJayamanne: The use of `!` is the root cause.

IanMatthewHuff: Doesn't the || 'root' in line 130 assure that we'd have 'root' instead of ''? '' should be falsy. Or does the value of 'root' break in line 153?

DonJayamanne: Oh yes, let me double check that

DonJayamanne: Yup, i think with the changes Rich has made the original issue should be resolved.

PR #5094, opened 2021-03-09
remove All option from opt in/out of experiments

DavidKutu:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1353,8 +1353,7 @@
                     "default": [],
                     "items": {
                         "enum": [
-                            "CustomEditor",
-                            "All"
+                            "CustomEditor"
                         ]
                     },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1353,8 +1353,7 @@
                     "default": [],
                     "items": {
                         "enum": [
-                            "CustomEditor",
-                            "All"
+                            "CustomEditor"
```
IanMatthewHuff, line 1356: Was it just to remove the autocomplete for this? experiements/services.ts still has the code in it to use 'All' it just would not show up as an autocomplete option.

IanMatthewHuff: Looks good. At least that's assuming that we want to remove the option from people who have already specified it? But that was how I understood it at least.

Issue #4947, opened 2021-02-24
Native Notebooks - Python env for specific notebook is not remembered between sessions


jaymegordo:
## Environment data
Version: 1.54.0-insider
Commit: c47da720f351455681978a48edc4d9a0d095ea0e
Date: 2021-02-24T12:22:35.736Z
Electron: 11.3.0
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Darwin x64 20.3.0

-   VS Code version: 1.54.0-insider
-   Jupyter Extension version: v2021.3.593162453
-   Python Extension version: v2021.3.593199507-dev
-   OS and version: Darwin x64 20.3.0
-   Python and/or Anaconda version: 3.8.1
-   Type of virtual environment used: Conda
-   Jupyter server running: Local

## Expected behaviour

Specific conda env is remembered for .ipynb notebook between vscode restarts/reloads.

## Actual behaviour

After every restart jupyter env reverts to the same conda env instead of the one set for current notebook. (same one every time, seems to be arbitrary, cant find anywhere this is defined).

## Steps to reproduce:

[**NOTE**: Self-contained, minimal reproducing code samples are **extremely** helpful and will expedite addressing your issue]

1. Open my_notebook.ipynb
2. reload window
3. reopen my_notebook.ipynb

## Logs

<details>

<summary>Output for <code>Jupyter</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Jupyter</code>)
</summary>

<p>

```
XXX
```

</p>
</details>

DonJayamanne:
@jaymegordo we udpate the kernel information in the ipynb file only if a cell has been executed.
Please could you execute a cell in the notebook and confirm the environment is still not remembered between sessions.

jaymegordo:
@DonJayamanne yes definitely confirmed the env is not remembered even after executing cells. I have multiple notebooks I'm working in and the env is never remembered after I call "reload window".

I also:
- created a brand new test.ipynb
- changed to different env
- ran cells
- reloaded
- environment switched back to a different one

This issue only started happening within the last couple weeks, but happens on both 1.53.0 and 1.54.0, so I'd guess maybe a recent change in the jupyter extension?

In my case the env always switches back to `/opt/miniconda3/envs/573/bin/python`, but I can't find anywhere that would be defined as a hard default or anything.

[...]

PR #5097, opened 2021-03-09
Ignore errors when getting activated env variables

DonJayamanne:
For #5093 (originally reported in #4947)

### DIFF
--- a/src/client/api/pythonApi.ts
+++ b/src/client/api/pythonApi.ts
@@ -266,7 +266,7 @@ export class EnvironmentActivationService implements IEnvironmentActivationServi
     ): Promise<NodeJS.ProcessEnv | undefined> {
         return this.apiProvider
             .getApi()
-            .then((api) => api.getActivatedEnvironmentVariables(resource, interpreter, true));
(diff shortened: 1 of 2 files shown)


src/client/api/pythonApi.ts
```diff
@@ -266,7 +266,7 @@ export class EnvironmentActivationService implements IEnvironmentActivationServi
     ): Promise<NodeJS.ProcessEnv | undefined> {
         return this.apiProvider
             .getApi()
-            .then((api) => api.getActivatedEnvironmentVariables(resource, interpreter, true));
+            .then((api) => api.getActivatedEnvironmentVariables(resource, interpreter, false));
```
DonJayamanne, line 269: I think this is incorrect, if we do'nt have environment variables, we can still try to run the kernel.
Python extension has logging to capture the failure to get env vars & the necessary telemetry as well.

The rationale is, no point not launching the kernel because we couldn't get the activated environment variables, lets at least try to launch the kernel without that, and it might work in a majority of the case.
& if we have issues we can track the error logs and see that it has failed.

rchiodo (approved):

Issue #5145, opened 2021-03-12
Ensure Python extension does not fail to load if Jupyter extension falls over


DonJayamanne:

IanMatthewHuff:
Validated with the addition of my fix here to clear the status message: https://github.com/microsoft/vscode-jupyter/pull/5239

PR #5146, opened 2021-03-12
Ignore load err that stops Python Ext from loading

DonJayamanne:
For #5145

### DIFF
--- a/src/client/extension.ts
+++ b/src/client/extension.ts
@@ -24,7 +24,7 @@ const stopWatch = new StopWatch();
 //===============================================
 // loading starts here
 
-import { ProgressLocation, ProgressOptions, window } from 'vscode';
+import { EventEmitter, ProgressLocation, ProgressOptions, window } from 'vscode';
 
 import { buildApi, IExtensionApi } from './api';
 import { IApplicationShell } from './common/application/types';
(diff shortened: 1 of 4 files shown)


src/client/extension.ts
```diff
@@ -58,7 +60,19 @@ export async function activate(context: IExtensionContext): Promise<IExtensionAp
         // We want to completely handle the error
         // before notifying VS Code.
         await handleError(ex, durations);
-        throw ex; // re-raise
+        traceError('Failed to active the Jupyter Extension', ex);
```
IanMatthewHuff, line 63: I'm pro this, as I'm big on not busting python. But just to ask. Does this make an odd customer case if they just have Jupyter installed? Like it just fails for the user without visual notification? Wondering about either only doing this if Python is installed or popping up a message (I'm assuming that the re-raise is what would usually show the message) so that the user at least knows what happened.

DonJayamanne: > uming that the re-raise is what would usually show the message) so that the user at least knows what happened.

I think its just to tell VS Code we failed.

`await handleError(ex, durations);`

this is where a message is displayed to the user today.
So if the extension fails to load they'll see a message (exiting code)

joyceerhl (approved):

DavidKutu (approved):

IanMatthewHuff (approved):

PR #5176, opened 2021-03-16
Improvements to how we run tests under different scenarios

DonJayamanne:
Basically a better way to generate different permutations.
Now we can literally have all permutations

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -419,11 +419,11 @@ jobs:
 
       # Install PyTorch with pip
       - name: Install PyTorch on Windows and Linux
-        if: matrix.test-suite != 'conda' && (matrix.os == 'windows-latest' || matrix.os == 'ubuntu-latest')
+        if: matrix.python != 'conda' && matrix.python != 'noPython' && (matrix.os == 'windows-latest' || matrix.os == 'ubuntu-latest')
         run: pip install torch==1.7.1+cpu torchvision==0.8.2+cpu torchaudio===0.7.2 -f https://download.pytorch.org/whl/torch_stable.html
 
       - name: Install PyTorch on Mac
-        if: matrix.test-suite != 'conda' && matrix.os == 'macos-latest'
+        if: matrix.python != 'conda' && matrix.python != 'noPython' && matrix.os == 'macos-latest'
         run: pip install torch torchvision torchaudio
 
       - name: Install secondary conda environment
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test.yml
```diff
@@ -491,43 +491,78 @@ jobs:
         # We're not running CI on macOS for now because it's one less matrix entry to lower the number of runners used,
         # macOS runners are expensive, and we assume that Ubuntu is enough to cover the UNIX case.
         os: [ubuntu-latest]
-        python: [3.8] # Use flaky tests to run against more versions of Python.
+        pythonVersion: [3.8]
+        python: [nonConda, conda]
+        jupyter: [raw, local, remote]
```
DonJayamanne, line 496: Every scenario is now in the matrix, hence we have all possible scenarios.
I have excluded some with comments, e.g. to reduce unnecessary permutations/scenarios.
Optionally we can add os = windows and double all the tests,
but that might not be an efficient way, we can optionally add `includes` as i have done, this way. I have commented out windows as an example
We can add them later, looks like some stuff aren't working on CI in windows.

.github/workflows/build-test.yml
```diff
@@ -729,20 +766,20 @@ jobs:
           VSC_JUPYTER_FORCE_LOGGING: 1
           VSC_PYTHON_FORCE_LOGGING: 1
           VSC_JUPYTER_CI_RUN_NON_PYTHON_NB_TEST: 1
-          VSC_JUPYTER_REMOTE_NATIVE_TEST: ${{ contains(matrix.test-suite, 'remote') }}
-          VSC_JUPYTER_NON_RAW_NATIVE_TEST: ${{ contains(matrix.test-suite, 'nonRawNotebook') }}
-          VSC_JUPYTER_CI_RUN_JAVA_NB_TEST: ${{ contains(matrix.test-suite, 'conda') }}
-          VSC_JUPYTER_CI_IS_CONDA: ${{ contains(matrix.test-suite, 'conda') }}
+          VSC_JUPYTER_REMOTE_NATIVE_TEST: ${{ matrix.jupyter == 'remote' }}
```
DonJayamanne, line 769: Should probably clean this up as well, and have just one single test step, instead of the 3 that we have. however I believe what we have now with the current change, its much simpler to add new permutations & later we can clean this.

Plan was to work on Windows tests, but had to move to other issues.

.github/workflows/build-test.yml
DonJayamanne: All of the followign permutations are now generated with test matrix and we have better control & its much simpler to control the scenarios as well

.github/workflows/build-test.yml
```diff
@@ -491,43 +491,78 @@ jobs:
         # We're not running CI on macOS for now because it's one less matrix entry to lower the number of runners used,
         # macOS runners are expensive, and we assume that Ubuntu is enough to cover the UNIX case.
         os: [ubuntu-latest]
-        python: [3.8] # Use flaky tests to run against more versions of Python.
+        pythonVersion: [3.8]
+        python: [nonConda, conda]
```
DonJayamanne, line 495: We can add others later, pyenv, virtual env, etc (if we want)

.github/workflows/build-test.yml
```diff
@@ -491,43 +491,78 @@ jobs:
...
+        jupyter: [raw, local, remote]
         # integration: Tests with VS Code, Python extension & real Jupyter
         # notebook: Notebook Tests with VS Code, Python extension & real Jupyter
-        test-suite: [integration, notebook, remoteNotebook, nonRawNotebook, notebookWithoutPythonExt, conda, notebookAndWebview]
+        # notebookAndWebview: Tests with VS Code (possible Notebooks), WebViews, Python extension & real Jupyter
+        test-suite: [integration, notebook]
```
DonJayamanne, line 500: The types of tests we want, i think this is yet another way to categorize the tests.
Notebook tests, integration tests, webview tests, IPyWidget Tests, etc.
And each of them will run under different scenarios yet again, under python, conda, non-conda, local, remote, raw, etc.

rchiodo: The actual output here is a bit confusing to compute. 

So when running remote, 3.8 we exclude integration?
And when running 3.8, conda, remote, we exclude notebook?

rchiodo (approved):

.github/workflows/build-test.yml
rchiodo: Can you add a comment explaining what's included/excluded?

DonJayamanne: > So when running remote, 3.8 we exclude integration?

Yes

Basically if we have n*n permuations, i can tell what particular permuation needs to be excluded.
Here I'm saying from all permuations, if there is one that matches OS = Ubuntu, Python 3.8, Jupyter = Remote and test-suite = integration, then exclude that item from the list.

DonJayamanne: Will add comments.

PR #5180, opened 2021-03-16
Changes to execution of Native Notebooks (changes in Proposed API)

DonJayamanne:
For #5183

Next (late monday) week we're going to have to merge this in and get an insider version ready to be shipped.
Tests will fail until we get the latest insider version.

### DIFF
--- a/src/client/api.ts
+++ b/src/client/api.ts
@@ -3,12 +3,13 @@
 
 'use strict';
 
-import { Disposable, Event, NotebookCell, NotebookCellRunState, Uri } from 'vscode';
+import { Disposable, Event, NotebookCell, Uri } from 'vscode';
 import { IPythonApiProvider, PythonApi } from './api/types';
 import { isTestExecution } from './common/constants';
 import { traceError } from './common/logger';
 import { VSCodeNotebookProvider } from './datascience/constants';
 import { IDataViewerDataProvider, IDataViewerFactory } from './datascience/data-viewing/types';
+import { NotebookCellRunState } from './datascience/jupyter/kernels/types';
 import { CreationOptionService } from './datascience/notebook/creation/creationOptionsService';
 import { KernelStateEventArgs } from './datascience/notebookExtensibility';
 import {
@@ -32,7 +33,13 @@ export interface IExtensionApi {
      * @memberof IExtensionApi
      */
     ready: Promise<void>;
+    /**
+     * Do not use this to monitor execution state of cells of Native Notebooks (use VS Code API).
+     */
     readonly onKernelStateChange: Event<KernelStateEventArgs>;
+    /**
+     * Do not use this to register Cell Toolbar icons for Native Notebook.
+     */
     registerCellToolbarButton(
         callback: (cell: NotebookCell, isInteractive: boolean, resource: Uri) => Promise<void>,
         codicon: string,
--- a/src/client/datascience/interactive-common/notebookUsageTracker.ts
+++ b/src/client/datascience/interactive-common/notebookUsageTracker.ts
@@ -6,6 +6,7 @@ import { EventEmitter, notebook, NotebookCellExecutionStateChangeEvent } from 'v
 import { IExtensionSingleActivationService } from '../../activation/types';
 import { IWorkspaceService } from '../../common/application/types';
 import { IDisposableRegistry } from '../../common/types';
+import { noop } from '../../common/utils/misc';
 import { sendTelemetryEvent } from '../../telemetry';
 import { Telemetry } from '../constants';
 import { INotebookEditor, INotebookEditorProvider } from '../types';
@@ -31,7 +32,7 @@ export class NotebookUsageTracker implements IExtensionSingleActivationService {
         // on this though.
         const findFilesPromise = this.workspace.findFiles('**/*.ipynb');
         if (findFilesPromise && findFilesPromise.then) {
-            findFilesPromise.then((r) => (this.notebookCount += r.length));
+            findFilesPromise.then((r) => (this.notebookCount += r.length), noop);
         }
         this.editorProvider.onDidOpenNotebookEditor(this.onEditorOpened, this, this.disposables);
         notebook.onDidChangeCellExecutionState(this.onDidChangeCellExecutionState, this, this.disposables);
--- a/src/client/datascience/interactive-common/notebookUsageTracker.ts
+++ b/src/client/datascience/interactive-common/notebookUsageTracker.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 import { inject, injectable } from 'inversify';
-import { EventEmitter } from 'vscode';
+import { EventEmitter, notebook, NotebookCellExecutionStateChangeEvent } from 'vscode';
 import { IExtensionSingleActivationService } from '../../activation/types';
 import { IWorkspaceService } from '../../common/application/types';
 import { IDisposableRegistry } from '../../common/types';
@@ -17,7 +17,7 @@ import { INotebookEditor, INotebookEditorProvider } from '../types';
 export class NotebookUsageTracker implements IExtensionSingleActivationService {
     protected readonly _onDidChangeActiveNotebookEditor = new EventEmitter<INotebookEditor | undefined>();
     protected readonly _onDidOpenNotebookEditor = new EventEmitter<INotebookEditor>();
-    private readonly executedEditors = new Set<INotebookEditor>();
+    private readonly executedNotebooksIndexedByUri = new Set<string>();
     private notebookCount: number = 0;
     private openedNotebookCount: number = 0;
     constructor(
@@ -34,14 +34,17 @@ export class NotebookUsageTracker implements IExtensionSingleActivationService {
             findFilesPromise.then((r) => (this.notebookCount += r.length));
         }
         this.editorProvider.onDidOpenNotebookEditor(this.onEditorOpened, this, this.disposables);
+        notebook.onDidChangeCellExecutionState(this.onDidChangeCellExecutionState, this, this.disposables);
     }
     public dispose() {
         // Send a bunch of telemetry
         if (this.openedNotebookCount) {
             sendTelemetryEvent(Telemetry.NotebookOpenCount, undefined, { count: this.openedNotebookCount });
         }
-        if (this.executedEditors.size) {
-            sendTelemetryEvent(Telemetry.NotebookRunCount, undefined, { count: this.executedEditors.size });
+        if (this.executedNotebooksIndexedByUri.size) {
+            sendTelemetryEvent(Telemetry.NotebookRunCount, undefined, {
+                count: this.executedNotebooksIndexedByUri.size
+            });
         }
         if (this.notebookCount) {
             sendTelemetryEvent(Telemetry.NotebookWorkspaceCount, undefined, { count: this.notebookCount });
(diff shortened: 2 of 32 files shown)


src/client/api.ts
```diff
@@ -32,7 +33,13 @@ export interface IExtensionApi {
      * @memberof IExtensionApi
      */
     ready: Promise<void>;
+    /**
+     * Do not use this to monitor execution state of cells of Native Notebooks (use VS Code API).
+     */
     readonly onKernelStateChange: Event<KernelStateEventArgs>;
```
DonJayamanne, line 39: @DavidKutu You'll need to update Gather extension to use this new VS Code API to get it work with the Jupyter extension. I.e. use `notebook.onDidChangeCellExecutionState` event

src/client/datascience/interactive-common/notebookUsageTracker.ts
```diff
@@ -17,7 +17,7 @@ import { INotebookEditor, INotebookEditorProvider } from '../types';
 export class NotebookUsageTracker implements IExtensionSingleActivationService {
     protected readonly _onDidChangeActiveNotebookEditor = new EventEmitter<INotebookEditor | undefined>();
     protected readonly _onDidOpenNotebookEditor = new EventEmitter<INotebookEditor>();
-    private readonly executedEditors = new Set<INotebookEditor>();
+    private readonly executedNotebooksIndexedByUri = new Set<string>();
```
DonJayamanne, line 20: Removed reliance on INotbeookEditor, we don't really need that.
Using the new VS Code API.

src/client/datascience/interactive-common/notebookUsageTracker.ts
```diff
@@ -52,8 +55,11 @@ export class NotebookUsageTracker implements IExtensionSingleActivationService {
         if (editor.model?.isUntitled) {
             this.notebookCount += 1;
         }
-        if (!this.executedEditors.has(editor)) {
-            editor.executed((e) => this.executedEditors.add(e), this, this.disposables);
+        if (!this.executedNotebooksIndexedByUri.has(editor.file.fsPath) && editor.executed) {
+            editor.executed((e) => this.executedNotebooksIndexedByUri.add(e.file.fsPath), this, this.disposables);
         }
     }
+    private onDidChangeCellExecutionState(e: NotebookCellExecutionStateChangeEvent): void {
```
DonJayamanne, line 62: Use the new VSC API.

src/test/datascience/interactiveWindow.functional.test.tsx
```diff
@@ -976,7 +976,9 @@ for i in range(100):
             const { window } = await getOrCreateInteractiveWindow(ioc);
             const executed = createDeferred();
             // We have to wait until the execute goes through before we reset.
-            window.onExecutedCode(() => executed.resolve());
+            if (window.onExecutedCode){
```
DonJayamanne, line 979: This event is now optional and only implemented by Interactive or webview & not Native Notebooks

src/client/datascience/utils.ts
```diff
@@ -62,9 +62,7 @@ export function translateCellToNative(
         return {
             index: 0,
             metadata: new NotebookCellMetadata().with({
-                executionOrder: cell.data.execution_count as number,
-                hasExecutionOrder: true,
-                runState: translateCellStateToNative(cell.state)
+                hasExecutionOrder: true
```
DonJayamanne, line 65: This stuff was being returned for Gather, when we fire events with the information of the cell.
Now we can use VS Code API instead of our wrappers.

src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -69,18 +65,13 @@ export class NotebookEditor implements INotebookEditor {
     public get active(): boolean {
         return this.vscodeNotebook.activeNotebookEditor?.document.uri.toString() === this.model.file.toString();
     }
-    public get onExecutedCode(): Event<string> {
```
DonJayamanne, line 72: Removed in favour of VS Code API

src/client/api.ts
```diff
@@ -32,7 +33,13 @@ export interface IExtensionApi {
...
+    /**
+     * Do not use this to register Cell Toolbar icons for Native Notebook.
+     */
     registerCellToolbarButton(
```
DavidKutu, line 43: This is needed for the Interactive window

src/client/api.ts
DavidKutu: This is needed for the Interactive window

src/client/datascience/utils.ts
```diff
@@ -79,42 +77,6 @@ export function translateCellToNative(
     }
 }
 
-export function translateCellFromNative(cell: NotebookCell): ICell {
```
DavidKutu, line 82: I think this is also needed for gather in the interactive window

DavidKutu (changes requested):
I think the translate functions you removed are used to translate the ICell from the Interactive window to a NotebookCell for gather

DonJayamanne:
@DavidKutu 

> I think the translate functions you removed in util.ts are used to translate the ICell from the Interactive window to a NotebookCell for gather

I removed `runState`, as this is not part of VS Code anymore, hence I'm merely removing something that doesn't exist in VS Code anymore.
If we were converting Interactive window ICells this way, then we're going to have to work around this and use the events to pass the run state information of cells. 

VS Code Notebook Cells no longer have run state information in them.

rchiodo (approved):

src/client/api.ts
DonJayamanne: All I'm saying here is Don't use this for Native Notebooks, & if using this for anythign else then continue doing so (I didn't want to mention what this is for or where it can be used)

src/client/api.ts
DonJayamanne: Yes, I agree, i'm only mentioning what this should not be used for.

src/client/datascience/utils.ts
DonJayamanne: Not sure why we would want to convert a VS Code Notebook cell into ICell for interactive window, it goes the other way. I couldn't find any code.
Please let me know what it is, i couldln't find it.

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -695,28 +732,33 @@ export class CellExecution {
...
+                promise = task?.appendOutput([output]);
+            }
+
+            this.endTemporaryTask();
+            // await on the promise at the end, we want to minimize UI flickers.
+            // The way we update output of other cells is to use an existing task or a temporary task.
+            // When using temporary tasks, we end up updating the UI with no execution order and spinning icons.
+            // Doing this causes UI updates, removing the awaits will enure there's no time for ui updates.
+            // I.e. create cell task, perform update, and end cell task (no awaits in between).
+            if (promise) {
```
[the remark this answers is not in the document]
DonJayamanne: > Makes me nervous somebody will forget to do this in the future. Is there some way we can enforce this?

The worst case is we'll have a flicker, it won't break anything.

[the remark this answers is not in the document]
DonJayamanne: I'm concerned the code will get more complicated if we try to fix it by moving it into a single place to avoid the situations you are mentioning.

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -275,77 +273,115 @@ export class CellExecution {
...
+        }
+        // If we have an active task, use that instead of creating a new task.
+        const existingTask = CellExecution.activeNotebookCellExecutionTask.get(this.cell.notebook);
+        if (existingTask) {
+            return existingTask;
+        }
+
+        // Create a temporary task.
+        this.previousResultsToRestore = { ...(this.cell.previousResult || {}) };
+        this.temporaryTask = notebook.createNotebookCellExecutionTask(
```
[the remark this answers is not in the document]
DonJayamanne: Agreed.

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -249,20 +250,17 @@ export class CellExecution {
...
-                const metadata = this.cell.metadata.with({ lastRunDuration: this.stopWatch.elapsedTime });
-                edit.replaceNotebookCellMetadata(this.cell.notebook.uri, this.cell.index, metadata);
-            });
-        }
-        await updateCellWithErrorStatus(this.cell, error);
 
-        this.errorHandler.handleError((error as unknown) as Error).ignoreErrors();
+        await chainWithPendingUpdates(this.cell.notebook, async () => {
+            traceCellMessage(this.cell, 'Update with error state & output');
+            await this.task?.appendOutput([translateErrorOutput(createErrorOutput(error))]);
```
IanMatthewHuff, line 256: Going to look at the API now, but wouldn't an error be a replaceOutput and not an appendOutput?

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -275,77 +273,115 @@ export class CellExecution {
...
-        editor.notifyExecution(this.cell);
+        this.previousResultsToRestore = undefined;
+        this.temporaryTask = undefined;
+    }
+
+    private async completedDueToCancellation() {
+        traceCellMessage(this.cell, 'Completed due to cancellation');
+        if (!this.cell.document.isClosed) {
+            await chainWithPendingUpdates(this.cell.notebook, (edit) => {
+                traceCellMessage(this.cell, 'Update cell statue as idle and message as empty');
```
IanMatthewHuff, line 362: cell status as idle

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -766,8 +812,66 @@ export class CellExecution {
...
+                    data: msg.content.data,
+                    metadata: msg.content.metadata
+                });
+                // If there was no output and still no output, then nothing to do.
+                if (outputToBeUpdated.outputs.length === 0 && newOutput.outputs.length === 0) {
+                    return;
+                }
+                // Compare outputs (at the end of the day everything is serializable).
+                // Hence this is a safe comparison.
+                if (cell.outputs.length === newOutput.outputs.length && fastDeepEqual(cell.outputs, newOutput)) {
```
IanMatthewHuff, line 858: Cell outputs can be huge. Could this equality compare be a perf issue? Say you are dragging around a widget slider that's updating a complex chart in another cell. It's going to be doing a equality comparison on the entire output each drag step.

src/client/datascience/jupyter/kernels/cellExecution.ts
DonJayamanne: We don't want to blow away existing output sent by kernel, that doesn't seem right.
We should still display what was returned by the kernel.

FYI - prior to a cell getting executed all existing output is cleared.

src/client/datascience/jupyter/kernels/cellExecution.ts
DonJayamanne: Damn, it, tried to sneak one past you...

IanMatthewHuff (approved):
Had a question or two but looks legit. I'm glad that you did the meeting on the API before this or some of the changes would have been hard to understand. Both the TemporaryTask and NotebookExecutionStateTracker are kinda painful things that I'd rather not have, but seem like it's what we have to do for now.

src/client/datascience/jupyter/kernels/cellExecution.ts
DonJayamanne: Well, all of the output is already in memory, hence this is basically a comparison of in memory objects (arrays).
So, i don't see much of an issue here. 
I've optimized the comparison, to check each item individually, if something is different, then it will break early.

DonJayamanne:
> Both the TemporaryTask and NotebookExecutionStateTracker are kinda painful things that I'd rather not have, but seem like it's what we have to do for now.

Yup, i think we' all agree this is a horrible work around.

DavidKutu (approved):
After testing it, the code I was talking about was different. What you removed doesn't affect the interactive window nor the custom editor on stable.

joyceerhl (approved):

Issue #5193, opened 2021-03-12
Activation of the Python extension


kieferrm:
Issue Type: <b>Bug</b>

When I startup VS Code I get this in a workspace that exclusively contains markdown files. This is the only window I have open. Why?

![image](https://user-images.githubusercontent.com/4674940/110877623-84ce1f80-828e-11eb-896b-c9d1db62f540.png)

Extension version: 2021.2.636928669
VS Code version: Code - Insiders 1.55.0-insider (Universal) (9b2ee7fb7d970b9a628fdb3545bdd01d67078d16, 2021-03-11T05:15:23.749Z)
OS version: Darwin arm64 20.3.0

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|undefined|
|GPU Status|2d_canvas: enabled<br>gpu_compositing: enabled<br>metal: disabled_off<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>protected_video_decode: unavailable_off<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|2, 3, 2|
|Memory (System)|16.00GB (0.28GB free)|
|Process Argv|--enable-proposed-api lszomoru.lszomoru-proposed-api-sample --crash-reporter-id ef10dfea-3602-4c74-ba1a-109ce7da7825|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv695:30137379
vsins829:30139715
vsliv368:30146709
vsreu685:30147344
python383:30185418
pythonvspyt602:30263608
vspor879:30202332
vspor708:30202333
vspor363:30204092
vstry244:30244315
pythonvsdeb440:30224570
pythonvsded773:30223139
pythonvspyt875:30259475
pythonvsnew554:30265444
pythontb:30258533
vscoresta800:30265947
vspre833:30267464

```

</details>

greazer:
Looks like a dupe of #4965

rchiodo:
This is a dependency issue. Because the typescript notebook extension loads on startup, it also loads the jupyter extension.

greazer:
Ah yes, the other bug was about installing *Python*, sorry. However, the jupyter extension should not need to load the Python extension. The dependency isn't supposed to work that way.

greazer:
[...]

PR #5199, opened 2021-03-18
Prevent unnecessary activation of the Python extension

DonJayamanne:
For #5193

### DIFF
--- a/src/client/api/pythonApi.ts
+++ b/src/client/api/pythonApi.ts
@@ -373,6 +373,9 @@ export class InterpreterService implements IInterpreterService {
         }
     }
     private hookupOnDidChangeInterpreterEvent() {
+        if (this.eventHandlerAdded) {
+            return;
+        }
         this.apiProvider
             .getApi()
             .then((api) => {
--- a/src/client/api/pythonApi.ts
+++ b/src/client/api/pythonApi.ts
@@ -327,7 +327,7 @@ export class InterpreterService implements IInterpreterService {
 
     public get onDidChangeInterpreter(): Event<void> {
         if (this.extensionChecker.isPythonExtensionInstalled) {
-            if (this.extensionChecker.isPythonExtensionActive && this.eventHandlerAdded) {
+            if (this.extensionChecker.isPythonExtensionActive && !this.eventHandlerAdded) {
                 this.hookupOnDidChangeInterpreterEvent();
             }
             if (!this.extensionChecker.isPythonExtensionActive) {
--- a/src/client/api/pythonApi.ts
+++ b/src/client/api/pythonApi.ts
@@ -49,6 +49,10 @@ import {
 @injectable()
 export class PythonApiProvider implements IPythonApiProvider {
     private readonly api = createDeferred<PythonApi>();
+    private readonly didActivePython = new EventEmitter<void>();
+    public get onDidActivePythonExtension() {
+        return this.didActivePython.event;
+    }
 
     private initialized?: boolean;
 
@@ -80,6 +84,7 @@ export class PythonApiProvider implements IPythonApiProvider {
         } else {
             if (!pythonExtension.isActive) {
                 await pythonExtension.activate();
+                this.didActivePython.fire();
             }
             pythonExtension.exports.jupyter.registerHooks();
         }
@@ -105,6 +110,9 @@ export class PythonExtensionChecker implements IPythonExtensionChecker {
 
     public get isPythonExtensionInstalled() {
         return this.extensions.getExtension(this.pythonExtensionId) !== undefined;
+    }
+    public get isPythonExtensionActive() {
+        return this.extensions.getExtension(this.pythonExtensionId)?.isActive === true;
     }
 
     public async showPythonExtensionInstallRequiredPrompt(): Promise<void> {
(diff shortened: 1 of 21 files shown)


src/client/api/pythonApi.ts
```diff
@@ -49,13 +49,32 @@ import {
...
 
     constructor(
         @inject(IExtensions) private readonly extensions: IExtensions,
+        @inject(IDisposableRegistry) private readonly disposables: IDisposableRegistry,
         @inject(IPythonExtensionChecker) private extensionChecker: IPythonExtensionChecker
-    ) {}
+    ) {
+        const previouslyInstalled = this.extensionChecker.isPythonExtensionInstalled;
+        if (!previouslyInstalled) {
+            this.extensions.onDidChange(
```
DonJayamanne, line 67: If the extension is installed, then register the hooks

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -90,12 +91,13 @@ export class LocalKernelFinder implements ILocalKernelFinder {
...
             // Get list of all of the specs
             const kernels = await this.listKernels(resource, cancelToken);
+            const isPythonNb = isPythonNotebook(option);
 
             // Always include the interpreter in the search if we can
             const interpreter =
                 option && isInterpreter(option)
                     ? option
-                    : resource && this.extensionChecker.isPythonExtensionInstalled
+                    : resource && isPythonNb && this.extensionChecker.isPythonExtensionInstalled
```
DonJayamanne, line 100: Use active interpreter only if we know this is a python notebook

src/client/datascience/notebookStorage/nativeEditorStorage.ts
```diff
@@ -220,11 +216,6 @@ export class NativeEditorStorage implements INotebookStorage {
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
             return (notebookData.metadata.language_info.codemirror_mode as any).version;
         }
-        // Use the active interpreter if allowed
-        if (this.extensionChecker.isPythonExtensionInstalled) {
```
DonJayamanne, line 224: This is totally unnecessary, we're just storing this in metadata for language, today we generate all of this information from the kernel, hence we have more accurate information than this.

DonJayamanne: Also this was more stuff that slowed down notebooks, unnecessary when creating blank notebooks.

src/client/api/pythonApi.ts
```diff
@@ -291,16 +325,17 @@ export class InterpreterService implements IInterpreterService {
...
-                .getApi()
-                .then((api) => {
-                    if (!this.eventHandlerAdded) {
-                        this.eventHandlerAdded = true;
-                        api.onDidChangeInterpreter(() => this.didChangeInterpreter.fire(), this, this.disposables);
-                    }
-                })
-                .catch(noop);
+        if (this.extensionChecker.isPythonExtensionInstalled) {
+            if (this.extensionChecker.isPythonExtensionActive && this.eventHandlerAdded) {
```
IanMatthewHuff, line 329: Why is the check for this.eventHandlerAdded here? If eventHandlerAdded === true then hookup is a noop.

src/client/api/types.ts
```diff
@@ -19,12 +19,14 @@ export interface ILanguageServer extends Disposable {
 
 export const IPythonApiProvider = Symbol('IPythonApi');
 export interface IPythonApiProvider {
+    onDidActivatePythonExtension: Event<void>;
```
IanMatthewHuff, line 22: Not fully sure if I'm suggesting this. But does this seem to you like it should be on IPythonExtensionChecker? That has everything relevant to checking if the extension is installed or active.

IanMatthewHuff (changes requested):
Looks solid. Just had one question on the interfaces and one spot that looked like maybe a bug with the event hookup.

src/client/api/pythonApi.ts
```diff
@@ -49,13 +49,32 @@ import {
 @injectable()
 export class PythonApiProvider implements IPythonApiProvider {
     private readonly api = createDeferred<PythonApi>();
+    private readonly didActivatePython = new EventEmitter<void>();
```
joyceerhl, line 52: Do we need to make sure we dispose of this object?

src/client/datascience/kernel-launcher/localKernelFinder.ts
rchiodo: Not sure this works with the interactive window

rchiodo (changes requested):
Waiting on check that active interpreter is still used for interactive window

src/client/api/types.ts
DonJayamanne: No can do, its the PythonApiProvider that activates the Python extension. So only PythonAPIProvider knows if the python extension was activated or not.
VS Code doesn't fire events when extensions load.
Here we're only interested in situations when we load the python extension & other parts of the code can then react to that.

src/client/api/pythonApi.ts
DonJayamanne: Fixed.

src/client/api/pythonApi.ts
DonJayamanne: Fixed

DonJayamanne (commented):
> Waiting on check that active interpreter is still used for interactive window

Don't quite follow this comment.

rchiodo:
See the comment in localKernelFinder.ts

DavidKutu (approved):

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: Aaah, i think i can use the resource to check that, thanks

rchiodo (approved):

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: Fixed.

DonJayamanne:
> See the comment in localKernelFinder.ts

Somehow that comment was missing when i replied. anyways.

IanMatthewHuff (approved):

Issue #4965, opened 2021-02-26
Unwanted message asking to install Python when I am not using python


KwokQPhan:
## Environment data

-   VS Code version: 1.54-Insider
-   Jupyter Extension version (available under the Extensions sidebar): Marketplace
-   Python Extension version (available under the Extensions sidebar): No
-   OS (Windows | Mac | Linux distro) and version: Windows 10
-   Python and/or Anaconda version: No
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): No
-   Jupyter server running: No

## Expected behaviour

No message to install Python.
Is this Python extension or Jupyter?

## Actual behaviour

When I create new Notebook I am asked to install Python, why?
Install Julia Extension
I am asked to install Python?
All in my company gets same message, in SSH worse, always asked to install python.

Why not ask when I try to run a Python notebook instead, I do not have Python install, why ask

## Steps to reproduce:

[**NOTE**: Self-contained, minimal reproducing code samples are **extremely** helpful and will expedite addressing your issue]

1. XXX

## Logs

<details>

<summary>Output for <code>Jupyter</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Jupyter</code>)
</summary>

<p>

```
XXX
```

</p>
</details>

DonJayamanne:
Thanks for filing this issue, 
I believe I know what is going on here. We should be able to address this fairly easily.

joyceerhl:
Validated. Without the Python extension and an updated build of the Jupyter extension, if I open a blank notebook I do not get a prompt to install the Python extension.

PR #5202, opened 2021-03-18
No prompt to install Python when open an exist nb without Kernel

DonJayamanne:
For #4965
If you have a notebook with metadata that has a language of `python`, but no kernel information, we still prompt to install Python.
Not sure how users can create such notebooks, in our extension users can create such notebooks because we always default language to `python` for blank notebooks.
This ensures we don't display prompts if there is no kernel information.

### DIFF
--- a/src/client/datascience/notebook/defaultCellLanguageService.ts
+++ b/src/client/datascience/notebook/defaultCellLanguageService.ts
@@ -7,6 +7,7 @@ import type { nbformat } from '@jupyterlab/coreutils/lib/nbformat';
 import { inject, injectable, named } from 'inversify';
 import { Memento, NotebookCellKind, NotebookDocument } from 'vscode';
 import { IExtensionSingleActivationService } from '../../activation/types';
+import { IPythonExtensionChecker } from '../../api/types';
 import { IVSCodeNotebook } from '../../common/application/types';
 import { PYTHON_LANGUAGE } from '../../common/constants';
 import { traceWarning } from '../../common/logger';
@@ -27,6 +28,7 @@ export class NotebookCellLanguageService implements IExtensionSingleActivationSe
     constructor(
         @inject(IVSCodeNotebook) private readonly vscNotebook: IVSCodeNotebook,
         @inject(IDisposableRegistry) private readonly disposables: IDisposableRegistry,
+        @inject(IPythonExtensionChecker) private readonly pythonExtensionChecker: IPythonExtensionChecker,
         @inject(IMemento) @named(GLOBAL_MEMENTO) private readonly globalMemento: Memento
     ) {}
     /**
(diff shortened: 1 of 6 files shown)


rchiodo (approved):

src/client/datascience/notebook/defaultCellLanguageService.ts
```diff
@@ -38,7 +40,11 @@ export class NotebookCellLanguageService implements IExtensionSingleActivationSe
             metadata?.language_info?.name ||
             (metadata?.kernelspec as IJupyterKernelSpec | undefined)?.language ||
             this.lastSavedNotebookCellLanguage;
-        return translateKernelLanguageToMonaco(jupyterLanguage || PYTHON_LANGUAGE);
+
+        // Default to python language only if the Python extension is installed.
+        const defaultLanguage = this.pythonExtensionChecker.isPythonExtensionInstalled ? PYTHON_LANGUAGE : 'plaintext';
```
DonJayamanne, line 45: I think this wasn't right.

DonJayamanne: If user doesn't have a python kernel at all, no point defaulting to Python

src/client/datascience/notebook/helpers/helpers.ts
```diff
@@ -107,11 +107,8 @@ export function isPythonNotebook(metadata?: nbformat.INotebookMetadata) {
     if (metadata?.language_info?.name && metadata.language_info.name !== PYTHON_LANGUAGE) {
         return false;
     }
-    if (kernelSpec?.language && kernelSpec.language !== PYTHON_LANGUAGE) {
-        return false;
-    }
-    // All other notebooks are python notebooks.
-    return true;
+    // Valid notebooks will have a language information in the metadata.
+    return kernelSpec?.language === PYTHON_LANGUAGE;
```
DonJayamanne, line 111: This too wasn't right, better now

src/client/datascience/notebook/kernelWithMetadata.ts
```diff
@@ -64,18 +65,29 @@ export class VSCodeNotebookKernelMetadata implements VSCNotebookKernel {
...
-
+        // If the notebook is closed, dispose everything.
+        notebook.onDidCloseNotebookDocument(
+            (e) => {
+                if (e === doc) {
+                    disposeAllDisposables(handlerDisposables);
+                }
+            },
+            this,
+            handlerDisposables
```
DonJayamanne, line 82: Better clean up of handlers.

src/client/datascience/notebook/kernelWithMetadata.ts
```diff
@@ -64,18 +65,29 @@ export class VSCodeNotebookKernelMetadata implements VSCNotebookKernel {
...
+        );
         const saveKernelInfo = () => {
             const kernelId = kernelSocket?.options.id;
             if (!kernelId) {
                 return;
             }
             traceInfo(`Updating preferred kernel for remote notebook ${kernelId}`);
             this.preferredRemoteKernelIdProvider.storePreferredRemoteKernelId(doc.uri, kernelId).catch(noop);
-
-            disposeAllDisposables(handlerDisposables);
```
DonJayamanne, line 78: This was wrong

src/client/datascience/notebookStorage/vscNotebookModel.ts
```diff
@@ -83,15 +83,12 @@ export class VSCodeNotebookModel extends BaseNotebookModel {
         return this.document ? this.document.cells.length : this.notebookJson.cells?.length ?? 0;
     }
     public getNotebookData() {
-        if (!this.preferredLanguage) {
-            throw new Error('Preferred Language not initialized');
```
DonJayamanne, line 87: Its possible to not have a preferred langauge, we'll default to plain text.

DonJayamanne: Note: When you create a blank notebook and we return cells with plain text,
& a kernel is selected, then the language of the cells get updated to match the language of the kernel selected.

Tha'ts existing fucntionality.

I.e. if you create a blank notebook and cells are TypeScript, then select Juali kernel, then language of empty cells in blank notebook will change to Julia.

src/client/datascience/notebook/defaultCellLanguageService.ts
DonJayamanne: Else later if you save these notebooks and open it, then we assume its a python notebook and ask the user to install Python.

IanMatthewHuff (approved):

DavidKutu (approved):

Issue #5195, opened 2021-03-17
Update widgets code with the same API changes as renderers extension


IanMatthewHuff:
https://github.com/microsoft/vscode-notebook-renderers/pull/35

PR #5203, opened 2021-03-18
Update widget output params

IanMatthewHuff:
For #5195 

### DIFF
--- a/src/datascience-ui/ipywidgets/kernel/index.ts
+++ b/src/datascience-ui/ipywidgets/kernel/index.ts
@@ -11,6 +11,26 @@ import { SharedMessages } from '../../../client/datascience/messages';
 import { PostOffice } from '../../react-common/postOffice';
 import { WidgetManager } from '../common/manager';
 import { ScriptManager } from '../common/scriptManager';
+
+// Copy of vscode-notebook-renderer old types as of 1.48
+// Keep these so we can support both the old interface and the new interface
+// Interface change here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51675/files
+interface OldNotebookCellOutputMetadata {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    custom?: { [key: string]: any };
+}
+interface OldNotebookOutput {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    data: { [mimeType: string]: any };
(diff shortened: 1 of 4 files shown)


IanMatthewHuff:
@DonJayamanne Is there a way to test this in stable as well? Can't just build and debug in stable as the stable engine isn't 1.55. With the renderers I copied over the client_renderer to the jupyter extension installed in stable to test. Can I just dump my out dir into the stable extension folder? 

I did check manualTestFiles (which has rich outputs + a few widgets) on insiders both with the old renderer code and my new updated renderer code and it looked legit to me.

rchiodo (approved):

DonJayamanne:
> Can't just build and debug in stable as the stable engine isn't 1.55. With the renderers I copied over the client_renderer to the jupyter extension installed in stable to test. Can I just dump my out dir into the stable extension folder?

When we create a new release, we'll pin against the next version of VS Code (which is the current insider).
Hence no need to test in Stable at all.

src/datascience-ui/ipywidgets/kernel/index.ts
```diff
@@ -11,6 +11,26 @@ import { SharedMessages } from '../../../client/datascience/messages';
...
+interface OldNotebookCellOutputMetadata {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    custom?: { [key: string]: any };
+}
+interface OldNotebookOutput {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    data: { [mimeType: string]: any };
+    metadata?: OldNotebookCellOutputMetadata;
+}
+interface OldNotebookOutputEventParams {
```
DonJayamanne, line 27: @IanMatthewHuff I think we can remove this code and there's no need for backwards compatiblity.
The only reason i wanted the backwards compatiblity for renderers, is so that it works with point releases of jupyter extension (point release is for current stable).

DonJayamanne (approved):

src/datascience-ui/ipywidgets/kernel/index.ts
IanMatthewHuff: @DonJayamanne Ahh, that's a good call. Wasn't thinking when I moved that code over. You are right, this will only release with matching stable builds. I'll remove that quick.

Issue #5191, opened 2021-03-17
Updates to Notebook API


DonJayamanne:
* Checking if a notebook has been closed using window.onDidCloseNotebookDocument instead of edito.onDidDispose
* Stdout & stderr streams have been added (we can stop using the text stream)

IanMatthewHuff:
@DonJayamanne Made the onDidDispose change. Is there info about the stderr / stdout streams? Not seeing anything on the API announce page or in the proposed.d.ts

PR #5220, opened 2021-03-19
Notebook API updates

IanMatthewHuff:
For #5191 

### DIFF
--- a/src/client/datascience/notebook/helpers/helpers.ts
+++ b/src/client/datascience/notebook/helpers/helpers.ts
@@ -548,7 +548,7 @@ function translateStreamOutput(output: nbformat.IStream): NotebookCellOutput {
     return new NotebookCellOutput(
         [
             new NotebookCellOutputItem(
-                CellOutputMimeTypes.textStream,
+                output.name === 'stderr' ? CellOutputMimeTypes.stderr : CellOutputMimeTypes.stdout,
                 concatMultilineString(output.text),
                 getOutputMetadata(output)
             )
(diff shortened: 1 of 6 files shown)


IanMatthewHuff:
What it looks like with the new vscode mime types. Red text for stderr
image
Note two outputs for different streams.

DonJayamanne (approved):

src/client/datascience/notebook/helpers/helpers.ts
```diff
@@ -641,12 +641,13 @@ function translateCellDisplayOutput(output: NotebookCellOutput): JupyterOutput {
...
             const outputs = output.outputs
-                .filter((opit) => opit.mime === CellOutputMimeTypes.textStream)
+                .filter((opit) => opit.mime === CellOutputMimeTypes.stderr || opit.mime === CellOutputMimeTypes.stdout)
                 .map((opit) => opit.value as string | string[])
                 .reduceRight<string[]>(
                     (prev, curr) => (Array.isArray(curr) ? prev.concat(...curr) : prev.concat(curr)),
                     []
                 );
+
             result = {
```
IanMatthewHuff, line 651: Minor comment here on the result. The OutputItems now have a mimeType that indicates if they are stderr or stdout. Given that mimeType it's not 100% needed that we still have the streamName in the customMetadata (it's what we previously used to determine this). However I have kept it in. Having the streamName on the Output (a level up from the OutputItems[]) is convenient for something like this where we can just look at it to reconstitute the IStream item from the single Output. Jupyter creates different outputs for stderr and stdout, so you don't have to worry about a case where there are mixed stderr and stdout stream mimetypes in the array of OutputItems.

If we don't like this I could remove the streamName property from metadata and just look at the mimeType of the first item in the output array.

DonJayamanne: > If we don't like this I could remove the streamName property from metadata and just look at the mimeType of the first item in the output array.

Yes thats a good idea, I think we should do that as its cleaner, less crud to carry around.

DonJayamanne: I.e. please remove the property `streamName`.

IanMatthewHuff: Yeah, I'm thinking the same. Will do.

rchiodo (approved):

DonJayamanne (approved):

Issue #5212, opened 2021-03-19
I think that pyvsc-run-isolated.py has a bug (Python 3.x.x 64-bit requires ipykernel to be installed.)


adp-study:
## Environment data

-   VS Code version: 1.54.3
-   Jupyter Extension version (available under the Extensions sidebar): 2021.3.619093157
-   Python Extension version (available under the Extensions sidebar): 3.8.8 64bit
-   OS (Windows | Mac | Linux distro) and version: Windows 10
-   Python and/or Anaconda version: 3.8.8 (it is not anaconda)
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A
-   Jupyter server running: Local

## Expected behaviour

Jupyter will execute python code normally.

## Actual behaviour

When I execute python code with jupyter(Ctrl+Enter), vscode says `python requires ipykernel to be installed`.

If I comment code with in `pyvsc-run-isolated.py` as bellow or python38.zip that extract to site-packages path than Jupyter works perfectly.

```
#sys.path[0] = os.path.join(os.path.dirname(__file__), ".does-not-exist")
```

sys.path has package path to search(in my case sys.path[0] has `d:\\python3\\python38.zip`) but `pyvsc-run-isolated.py` removes it.

Sorry for my pure English.

---
my bad. I didn't know about sys.path[0] behavior.
Only windows build(from www.python.org) works differently.
For more robust code, why don't you check sys.path[0]'s ext. Thank you very much. I love vs-code and jupyter.

## Steps to reproduce:

install python-3.8.8-embed version from python.org(and jupyter package has installed too) then the problem is occurred.

## Logs

<details>

<summary>Output for <code>Jupyter</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Jupyter</code>)
</summary>

<p>

```
 d:\python3\python.exe c:\Users\RyuShin\.vscode\extensions\ms-toolsai.jupyter-2021.3.619093157\pythonFiles\pyvsc-run-isolated.py -c "import ipykernel"
[...]

PR #5221, opened 2021-03-19
Updates to code used to run Python in an isolated manner

DonJayamanne:
For #5212
Updated our code from Core Extension (from what i know all of this stuff will soon go away when Trusted workspaces comes in). I.e. hopefully this will be the last time we need to update this stuff.

Looks a lot of users are having issues with the run isolated script.
Could be causing other issues as well.

### DIFF
--- a/pythonFiles/interpreterInfo.py
+++ b/pythonFiles/interpreterInfo.py
@@ -9,5 +9,6 @@ obj["versionInfo"] = tuple(sys.version_info)
 obj["sysPrefix"] = sys.prefix
 obj["version"] = sys.version
 obj["exe"] = sys.executable
+obj["is64Bit"] = sys.maxsize > 2 ** 32
 
 print(json.dumps(obj))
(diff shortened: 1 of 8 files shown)


rchiodo (approved):

pythonFiles/interpreterInfo.py
```diff
@@ -9,5 +9,6 @@
 obj["sysPrefix"] = sys.prefix
 obj["version"] = sys.version
 obj["exe"] = sys.executable
+obj["is64Bit"] = sys.maxsize > 2 ** 32
```
IanMatthewHuff, line 12: Interesting. I knew about plarform.architecture(), but seems like this is the recommended way.

IanMatthewHuff (approved):

pythonFiles/interpreterInfo.py
DonJayamanne: This was always part of the core code, at some point it was removed and looks like it's getting added back again.

Issue #5022, opened 2021-03-03
Release branches do not need to upload code coverage


rchiodo:
This slows down release builds

IanMatthewHuff:
Validated. Release build here doesn't upload coverage.
https://github.com/microsoft/vscode-jupyter/actions/runs/687458533

PR #5230, opened 2021-03-22
Ignore uploading coverage for release branches

DonJayamanne:
For #5022

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -253,7 +253,7 @@ jobs:
       # Upload unit test coverage reports for later use in the "reports" job.
       - name: Upload unit test coverage reports
         uses: actions/upload-artifact@v2
-        if: always()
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yml
```diff
@@ -253,7 +253,7 @@ jobs:
       # Upload unit test coverage reports for later use in the "reports" job.
       - name: Upload unit test coverage reports
         uses: actions/upload-artifact@v2
-        if: always()
+        if: "(success() || failure()) && !contains(github.ref, 'refs/heads/release')"
```
DonJayamanne, line 256: We won't want to upload coverage when we cancel CI.
Currently always means even if we cancel, it will run this task.
This takes a lot of time when cancelling jobs.

IanMatthewHuff (approved):

DavidKutu (approved):

PR #5239, opened 2021-03-22
Finally to make sure that we claen up our status message

IanMatthewHuff:
For #5145 (followup)

### DIFF
--- a/src/client/extension.ts
+++ b/src/client/extension.ts
@@ -104,24 +104,31 @@ async function activateUnsafe(
     startupDurations: Record<string, number>
 ): Promise<[IExtensionApi, Promise<void>, IServiceContainer]> {
     const activationDeferred = createDeferred<void>();
-    displayProgress(activationDeferred.promise);
-    startupDurations.startActivateTime = startupStopWatch.elapsedTime;
+    try {
+        displayProgress(activationDeferred.promise);
+        startupDurations.startActivateTime = startupStopWatch.elapsedTime;
 
-    //===============================================
-    // activation starts here
+        //===============================================
+        // activation starts here
 
-    const [serviceManager, serviceContainer] = initializeGlobals(context);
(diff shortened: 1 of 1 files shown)


src/client/extension.ts
```diff
@@ -104,24 +104,31 @@ async function activateUnsafe(
...
+        startupDurations.endActivateTime = startupStopWatch.elapsedTime;
+        activationDeferred.resolve();
 
-    const api = buildApi(activationPromise, serviceManager, serviceContainer);
-    return [api, activationPromise, serviceContainer];
+        const api = buildApi(activationPromise, serviceManager, serviceContainer);
+        return [api, activationPromise, serviceContainer];
+    } finally {
+        // Make sure that we clear our status message
+        if (!activationDeferred.completed) {
```
IanMatthewHuff, line 128: Found this while validating 5145 fix. We could allow python to load, but we left up our loading message:
image
Jupyter Extension loading would never go away.

joyceerhl: If we reject here, does VS Code know that our activation has failed, and would this cause extensions depending on us to also fail to activate?

IanMatthewHuff: Double checking now. But I believe this reject is safe to use. This reject is only on the activationDeferred which we only use for turning on or off the window.withProgress status message. The window.withProgress is just a noop on the then, so it doesn't care if the deferred is resolved via resolve or reject, just that it's done. 

The actual error thrown takes the same route as before and is handled up at line 65 of this file where the user is notified and the "fake" api is returned to allow Python to function even if we fail to load. Don already made that fix before, this was just for the status message not being cleared in that scenario.

DonJayamanne (approved):

DavidKutu (approved):

rchiodo (approved):

PR #5245, opened 2021-03-23
Fix scheduled build pipeline

DonJayamanne:
* There was a typo
* Forgot to remove uploads of insider builds after every commit.
* Install VSCE

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -69,11 +69,10 @@ jobs:
         run: |
           echo "vsix_name=ms-toolsai-jupyter-insiders.vsix" >> $GITHUB_ENV
           echo "test_matrix_os=[\"ubuntu-latest\", \"windows-latest\"]" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yml
```diff
@@ -1061,6 +1060,8 @@ jobs:
         with:
           node-version: ${{env.NODE_VERSION}}
 
+      - run: npm i -g vsce
```
rchiodo, line 1063: Should we just add this to our dev dependencies?

rchiodo (approved):

.github/workflows/build-test.yml
IanMatthewHuff: I was about to ask, I thought that it already was. Not sure how build would have worked without this.

DonJayamanne: Not sure, I kinda feel npm ci would take longer.
I'd prefer to leave it out for now

Issue #5130, opened 2021-03-11
Remove jupyter.useNotebookEditor setting


joyceerhl:

IanMatthewHuff:
Validated. Setting is gone now.

Pugio:
Is there an alternative to this? I can't seem to find a way to automatically open '.ipynb' as JSON anymore. Sometimes I just want to see the raw file, and not start a whole notebook editing instance.

gchenfc:
@joyceerhl @IanMatthewHuff Could one of you clarify? I'm not clear on where justifications for planned changes are documented (this issue doesn't have a description?) and I would really like to know of any alternatives.

It's quite frustrating for .ipynb files to automatically open up in notebook form without an option to view them in json form, especially for diff in source control.  And sometimes when the .ipynb file is broken for whatever reason (e.g. because source control did a bad merge because we can't view the file in json form), it can't be fixed in vscode since the jupyter viewer just crashes (I renamed the file to .json, edited, then renamed it back, then repeated until all the mistakes were fixed).

Combined with the fact that I can't seem to uninstall the Jupyter extension without uninstalling the entire Python extension, this is kind of a big inconvenience and I bet a pretty decent chunk of the python coding world might have the same issue?

gchenfc:
#633 just linking in this related issue

PR #5249, opened 2021-03-23
Remove jupyter.useNotebookEditor

DonJayamanne:
For #5130

### DIFF
--- a/package.json
+++ b/package.json
@@ -1604,12 +1604,6 @@
                     "description": "When opening ipynb files, automatically preview the contents in the Interactive window.",
                     "scope": "resource"
                 },
-                "jupyter.useNotebookEditor": {
-                    "type": "boolean",
-                    "default": true,
-                    "description": "Automatically open .ipynb files in the Notebook Editor.",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -1604,12 +1604,6 @@
                     "description": "When opening ipynb files, automatically preview the contents in the Interactive window.",
                     "scope": "resource"
                 },
-                "jupyter.useNotebookEditor": {
```
rchiodo, line 1607: Do we need to mark this as deprecated somehow? Is there a way to signal to people that this setting is no longer supported if they have in their settings.json?

rchiodo (approved):

package.json
IanMatthewHuff: Yeah, our runMagicCommands has an example of this. I believe that we want to use that instead of removing it fully.

DonJayamanne: Do'nt see the point of marking it as deprecated, what's the user going to do. Its not like they can go back to an older behavior.

DavidKutu (approved):

Issue #5253, opened 2021-03-23
Data viewer does not display non-numeric index column values in DataFrames


joyceerhl:
_Originally posted by @mbijjur in https://github.com/microsoft/vscode-jupyter/issues/4886#issuecomment-799949059_

> Hi @mbijjur can you share a code sample and the version of the Jupyter extension that you have? Or is it the same version with either build of VS Code?

It is the same Jupyter extension version. You can try with any simple pandas df example.
column names does not appear when transposed.

![image](https://user-images.githubusercontent.com/31503873/111262164-c4f42000-8649-11eb-857d-b245c67a5b82.png)

IanMatthewHuff:
Validated. Looks good now.
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/34748f74-eab3-4096-bc4b-41faee3516e3)

PR #5254, opened 2021-03-23
Fix display of non-numeric DataFrame index columns

joyceerhl:
For #5253 

The change here is to show the index column in addition to the row number column only for pandas DataFrames and Series objects, because index columns are meaningful for DFs and Series but for ndarrays, lists, dicts and tensors the index column is generated by us as part of the DataFrame conversion code and is hence not meaningful to the user.

DataFrame:
image

Same data as an ndarray (note that there isn't actually an index column being generated on the ndarray itself):
image

### DIFF
--- a/pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
+++ b/pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
@@ -170,6 +170,17 @@ def _VSCODE_getDataFrameInfo(df):
 
     columnTypes = _VSCODE_builtins.list(df.dtypes)
 
+    # Compute the index column. It may have been renamed
+    try:
+        indexColumn = df.index.name if df.index.name else "index"
+    except AttributeError:
+        indexColumn = "index"
+
+    # Make sure the index column exists
+    if indexColumn not in columnNames:
+        columnNames.insert(0, indexColumn)
+        columnTypes.insert(0, "int64")
+
     # Then loop and generate our output json
     columns = []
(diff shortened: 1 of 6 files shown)


pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -170,6 +170,17 @@ def _VSCODE_getDataFrameInfo(df):
 
     columnTypes = _VSCODE_builtins.list(df.dtypes)
 
+    # Compute the index column. It may have been renamed
+    try:
```
IanMatthewHuff, line 174: This looks a bit different from the code in getJupyterVariableDataFrameInfo.py that does something similar with index column. Looks like your version might be safer with the try / except. Should that maybe be updated as well?

joyceerhl: This is actually just restoring code that I removed several PRs ago. Rich pointed out at the time that it was required and now I see why 😊

IanMatthewHuff: Sorry, maybe I didn't say that right. I was saying that this code here looked safer than the code in the other location. Just wondering if getJupyterVariableDataFrameInfo.py should be upgraded with the try / except that you have here.

joyceerhl: Ah good point, will update

Issue #5217, opened 2021-03-19
[Proposed API deprecation] NotebookContentProvider#resolveNotebook


jrieken:
From the proposed API, we will remove the `NotebookContentProvider#resolveNotebook`-function. To establish a communication channel between renderers and extensions use `NotebookKernelProvider#resolveKernel`

IanMatthewHuff:
Our only usage for this is for widget tests that are currently disabled. Will make the change and look at re-enabling those tests.

PR #5255, opened 2021-03-23
Use resolveKernel instead of resolveDocument which is due for deprecation and re-enable and update associated tests.

IanMatthewHuff:
For #5217

### DIFF
--- a/src/client/datascience/notebook/contentProvider.ts
+++ b/src/client/datascience/notebook/contentProvider.ts
@@ -36,23 +36,14 @@ import { NotebookEditorCompatibilitySupport } from './notebookEditorCompatibilit
  */
 @injectable()
 export class NotebookContentProvider implements VSCNotebookContentProvider {
-    public webviews = new Map<string, NotebookCommunication[]>();
     constructor(
         @inject(INotebookStorageProvider) private readonly notebookStorage: INotebookStorageProvider,
         @inject(NotebookEditorCompatibilitySupport)
         private readonly compatibilitySupport: NotebookEditorCompatibilitySupport,
         @inject(IVSCodeNotebook) readonly notebookProvider: IVSCodeNotebook
-    ) {
-        notebookProvider.onDidCloseNotebookDocument(this.onDidCloseNotebook.bind(this));
-    }
-    public async resolveNotebook(document: NotebookDocument, webview: NotebookCommunication): Promise<void> {
-        // Add webviews to list. Used for testing
-        let list = this.webviews.get(document.uri.toString());
-        if (!list) {
-            list = [];
-            this.webviews.set(document.uri.toString(), list);
-        }
-        list.push(webview);
+    ) {}
+    public async resolveNotebook(_document: NotebookDocument, _webview: NotebookCommunication): Promise<void> {
+        return Promise.resolve();
     }
     public async openNotebook(uri: Uri, openContext: NotebookDocumentOpenContext): Promise<NotebookData> {
         if (!this.compatibilitySupport.canOpenWithVSCodeNotebookEditor(uri)) {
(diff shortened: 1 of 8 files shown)


src/client/datascience/notebook/contentProvider.ts
```diff
@@ -36,23 +36,16 @@ import { NotebookEditorCompatibilitySupport } from './notebookEditorCompatibilit
...
-        if (!list) {
-            list = [];
-            this.webviews.set(document.uri.toString(), list);
-        }
-        list.push(webview);
+    ) {}
+    public async resolveNotebook(_document: NotebookDocument, _webview: NotebookCommunication): Promise<void> {
+        // This function is due for deprecation. Associated code has been moved to
+        // NotebookKernelProvider will remove when removed from the API fully
+        return Promise.resolve();
```
IanMatthewHuff, line 48: Better way to remove this? It's due for deprecation, but still in the proposed.d.ts. Didn't think that manually editing the proposed was a good idea, so I just moved out the code and made it a no-op.

src/client/datascience/notebook/kernelProvider.ts
```diff
@@ -50,6 +50,8 @@ import { traceInfo, traceInfoIf } from '../../common/logger';
 
 @injectable()
 export class VSCodeKernelPickerProvider implements INotebookKernelProvider {
+    // Keep a mapping of Document Uri => NotebookCommunication for widget tests
+    public webviews = new Map<string, NotebookCommunication[]>();
```
IanMatthewHuff, line 54: Basically just a move of this webview code from NotebookProvider to KernelProvider.

src/test/datascience/notebook/ipywidget.vscode.test.ts
```diff
@@ -4,97 +4,79 @@
 'use strict';
 
 /* eslint-disable @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires */
-// import * as path from 'path';
-// import * as sinon from 'sinon';
+import * as path from 'path';
```
IanMatthewHuff, line 7: Code change was easy, but the only place that webviews property that I moved from NotebookProvider to KernelProvider was being used was here in these widget tests. And these widget tests were actually being skipped. So I change them to use the new code, and re-enabled them, and fixed up a few issues with them.

Were running on my local machine with this change, will check in CI on the PR.

src/test/datascience/notebook/standard_widgets.ipynb
```diff
@@ -156,15 +156,6 @@
  "metadata": {
   "celltoolbar": "Attachments",
   "file_extension": ".py",
-  "kernelspec": {
```
IanMatthewHuff, line 159: I was having some issues with getting the right kernelspec in the test, so I removed this, so it would just rely on the language info. I believe that I could put this back if desired.

IanMatthewHuff: I noticed that a number of ipynbs that we use in our tests have some local machine specific info in them. Might be worth scrubbing that at some point.

rchiodo (approved):

IanMatthewHuff:
Bleh, CI tests look busted. Will investigate.

src/client/datascience/notebook/kernelProvider.ts
DonJayamanne: Can we change this to use a WeakMap and index it with the `NotebookDocument`
I.e. `new WeakMap<NotebookDocument, NotebookCommunication>`
Feels better than using a string and also if the notebook is disposed, we don't need to remove the item from the map either.

src/test/datascience/notebook/standard_widgets.ipynb
DonJayamanne: Oh, thanks.
In some tests (perhaps one), i remove the metadata.

DonJayamanne (changes requested):
Would prefer if we used a weakmap, better than a map with a string.

DonJayamanne (approved):

DavidKutu (approved):

PR #5260, opened 2021-03-23
Kernel finding refactor (#4995)

DonJayamanne:
Screen Shot 2021-03-23 at 16 14 29


IanMatthewHuff:
I'd say this seems too big for a point fix, but I think that you thought the same thing :(. Should we bug bash the point fix today? Just seems like we were already due for the release that I'm going to run next week. Doesn't seem that far away to me to just wait for that release.

IanMatthewHuff (approved):

joyceerhl:
Were we planning to include this in the point release? My understanding was it was just my data viewer regression fix.

PR #5276, opened 2021-03-24
Avoid calling `reveal` if a webview is already active or visible

mjbvz:
For https://github.com/microsoft/vscode/issues/118764

This prevents a duplicate editor from being opened when `show` is called on a webview that is part of a diff view 

I gave this a quick test but don't know all potential custom editor base notebook scenarios so it will need more testing from the folks that work on this extension

### DIFF
--- a/src/client/common/application/webviewPanels/webviewPanel.ts
+++ b/src/client/common/application/webviewPanels/webviewPanel.ts
@@ -29,8 +29,18 @@ export class WebviewPanel extends Webview implements IWebviewPanel {
 
     public async show(preserveFocus: boolean) {
         await this.loadPromise;
-        if (this.panel) {
-            this.panel.reveal(this.panel.viewColumn, preserveFocus);
+        if (!this.panel) {
+            return;
+        }
+
+        if (preserveFocus) {
+            if (!this.panel.visible) {
+                this.panel.reveal(this.panel.viewColumn, preserveFocus);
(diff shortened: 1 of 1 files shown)


DavidKutu (approved):

src/client/common/application/webviewPanels/webviewPanel.ts
```diff
@@ -29,8 +29,18 @@ export class WebviewPanel extends Webview implements IWebviewPanel {
 
     public async show(preserveFocus: boolean) {
         await this.loadPromise;
-        if (this.panel) {
-            this.panel.reveal(this.panel.viewColumn, preserveFocus);
+        if (!this.panel) {
+            return;
+        }
+
+        if (preserveFocus) {
```
IanMatthewHuff, line 36: I read the explanation here: https://github.com/microsoft/vscode/issues/118764#issuecomment-806195383. So the change makes sense. I might not fully grasp the API, but it does seem like this functionality should maybe be inside the reveal function? Like if !perserveFocus and the panel is already active then reveal should just be a no-op?

In any case, this works, so thanks for the fix.

IanMatthewHuff (approved):

mjbvz:
I fixed the upstream VS Code issue so you should be able to revert the change now if you like. Let me know if there are still any problems in the current VS Code insiders

PR #5297, opened 2021-03-25
Fix problems found while debugging synapse connections

rchiodo:
While debugging the connection to synapse last night there were some settings we had that Synapse wasn't allowing. Jupyter seems to be more flexible in this regard, but this tightening doesn't hurt (I tested with Azure ML and it still works).

### DIFF
--- a/src/client/datascience/jupyter/jupyterRequest.ts
+++ b/src/client/datascience/jupyter/jupyterRequest.ts
@@ -15,7 +15,7 @@ export function createAuthorizingRequest(getAuthHeader: () => any) {
             const authorizationHeader = getAuthHeader();
             const keys = Object.keys(authorizationHeader);
             keys.forEach((k) => origHeaders.append(k, authorizationHeader[k].toString()));
-            origHeaders.append('Content-Type', 'application/json');
+            origHeaders.set('Content-Type', 'application/json');
 
             // Rewrite the 'append' method for the headers to disallow 'authorization' after this point
             const origAppend = origHeaders.append.bind(origHeaders);
--- a/src/client/datascience/jupyter/jupyterSession.ts
+++ b/src/client/datascience/jupyter/jupyterSession.ts
@@ -244,11 +244,12 @@ export class JupyterSession extends BaseJupyterSession {
         }
 
         // Create our session options using this temporary notebook and our connection info
-        const options: Session.IOptions = {
-            path: backingFile?.path || '.',
-            kernelName: getNameOfKernelConnection(kernelConnection) || '',
+        const options: Session.IOptions = { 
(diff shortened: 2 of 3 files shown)


src/client/datascience/jupyter/jupyterRequest.ts
```diff
@@ -15,7 +15,7 @@ export function createAuthorizingRequest(getAuthHeader: () => any) {
             const authorizationHeader = getAuthHeader();
             const keys = Object.keys(authorizationHeader);
             keys.forEach((k) => origHeaders.append(k, authorizationHeader[k].toString()));
-            origHeaders.append('Content-Type', 'application/json');
+            origHeaders.set('Content-Type', 'application/json');
```
rchiodo, line 18: Append means there might already be a content-type set. We want to force application/json. Synapse was saying bad request when there were two content-types.

src/client/datascience/jupyter/jupyterSession.ts
```diff
@@ -244,11 +244,12 @@ export class JupyterSession extends BaseJupyterSession {
         }
 
         // Create our session options using this temporary notebook and our connection info
-        const options: Session.IOptions = {
-            path: backingFile?.path || '.',
-            kernelName: getNameOfKernelConnection(kernelConnection) || '',
+        const options: Session.IOptions = { 
+            path: backingFile?.path || `${uuid()}.ipynb`, // Name has to be unique
```
rchiodo, line 248: backingFile doesn't work on synapse (contents api is not supported), so this was always empty (which is an invalid file path).

src/client/datascience/jupyter/jupyterSession.ts
```diff
@@ -244,11 +244,12 @@ export class JupyterSession extends BaseJupyterSession {
...
+            kernelName: getNameOfKernelConnection(kernelConnection) || '', // TODO: This can't be empty. See https://github.com/microsoft/vscode-jupyter/issues/5290
```
rchiodo, line 249: Kernel name is needed too, but I added a separate bug for that.

src/client/datascience/jupyter/jupyterSession.ts
```diff
@@ -244,11 +244,12 @@ export class JupyterSession extends BaseJupyterSession {
...
             name: uuid(), // This is crucial to distinguish this session from any other.
-            serverSettings: serverSettings
+            serverSettings: serverSettings,
+            type: 'notebook'
```
rchiodo, line 252: type 'notebook' is required for synapse. Jupyter just assumes this.

IanMatthewHuff (approved):

DonJayamanne (approved):

PR #5303, opened 2021-03-25
Merge updates to Change log from Release branch

DonJayamanne:


IanMatthewHuff (approved):
These were cherry-picked fixes right? Don't we usually just have a follow up merge to main after full release? Not that I mind, but I didn't think that we had to pull these back to main one at a time.

Issue #5080, opened 2021-03-09
Random bytes generation has been skipped


DonJayamanne:
We need to re-enable this test.

PR #5308, opened 2021-03-25
Enable smoke test for randomBytesTest along with Windows unit tests

DonJayamanne:
For #5080

### DIFF
--- a/src/test/smoke/datascience.smoke.test.ts
+++ b/src/test/smoke/datascience.smoke.test.ts
@@ -30,7 +30,6 @@ suite('Smoke Tests', () => {
     teardown(closeActiveWindows);
 
     test('Random bytes generation', async function () {
-        return this.skip(); // Failing on windows. Tracked by 4444
         // We do have a unit test testing this, however create a smoke test to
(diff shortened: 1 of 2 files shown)


src/test/smoke/datascience.smoke.test.ts
```diff
@@ -3,50 +3,28 @@
...
         await setAutoSaveDelayInWorkspaceRoot(1);
     });
     setup(initializeTest);
     suiteTeardown(closeActiveWindows);
     teardown(closeActiveWindows);
 
-    test('Random bytes generation', async function () {
-        return this.skip(); // Failing on windows. Tracked by 4444
-        // We do have a unit test testing this, however create a smoke test to
-        // ensure that the bundling of the native node modules worked
```
rchiodo, line 35: Does this not still apply? It's verifying the bundling worked.

DonJayamanne: Oh, ok, i thought we had unit tests and that was sufficient.

DonJayamanne: Will revert.

PR #5320, opened 2021-03-25
Logging for kernel shutdown when closing Notebooks

DonJayamanne:
Added logging to detect when a kernel is shutdown & when a notebook is closed.
After adding this logging, things are now magically working.
Lets leave this in here, if this crops up again we'll be able to tell whether our events get fired or not.


rchiodo (approved):

IanMatthewHuff:
I mean, I'm happy it works, but doesn't seem like much of a root cause. If logging fixed it then a timing issue of some type?

Issue #4644, opened 2021-02-04
Run Cells Above & Below should be an icon in Cell Toolbar


DonJayamanne:
<img width="652" alt="Screen Shot 2021-02-03 at 18 01 04" src="https://user-images.githubusercontent.com/1948812/106834226-dfe67400-6649-11eb-81d5-bc52d6aa3ffd.png">

Other icons we see today for cells..
![Screen Shot 2021-02-04 at 09 52 46](https://user-images.githubusercontent.com/1948812/106935583-62197b80-66d0-11eb-80b1-494b9ba18d37.png)

As a user when I'm on a cell, and I want to run all cells above the current cell or below, then I need to move my mouse all the way up to the toolbar to click on those corresponding icons.

Ideally these icons should be on the Cell Toolbar, i.e. the icons should be next to the cell i'm working on. Just like Run Cell, Stop Cell, Delete cell icons are next to the cell.

jaymegordo:
@DonJayamanne are all these options in an unreleased version? I'm on the newest vscode-insiders with native notebooks, but my menu bar icons only look like this:
<img width="283" alt="Screen Shot 2021-02-04 at 10 43 15" src="https://user-images.githubusercontent.com/29611875/106940085-f5a17b00-66d5-11eb-82c7-049516b94de5.png">

I ask because I've really been hoping for the "run all cells above" functionality back in the nn. Thanks!

claudiaregio:
The run all above/below was placed in the notebook toolbar to mirror the notebook toolbar from stable. The icons could be moved to the cell toolbar but could be potentially resolved by VS Code. https://github.com/microsoft/vscode/issues/105183

claudiaregio:
Also waiting on the response to this first: https://github.com/microsoft/vscode/issues/114927

greazer:
Need to make sure we get SVG's added so that we can do this in the next stable release of VS Code.

joyceerhl:
Validated:
![image](https://user-images.githubusercontent.com/30305945/113433378-f1bb6d80-9393-11eb-8884-b21eaae19b87.png)

PR #5385, opened 2021-04-01
Move run above and below buttons from the notebook toolbar to the cell toolbar

DavidKutu:
For #4644

### DIFF
--- a/package.json
+++ b/package.json
@@ -287,20 +287,14 @@
                 "command": "jupyter.notebookeditor.runallcellsabove",
                 "title": "%DataScience.runAbove%",
                 "category": "Notebook",
-                "icon": {
-                    "light": "resources/light/runabove.svg",
-                    "dark": "resources/dark/runabove-inverse.svg"
-                },
-                "enablement": "notebookViewType == jupyter-notebook && jupyter.isnotebooktrusted && jupyter.notebookeditor.canRunCellsAboveInNativeNotebook"
+                "icon": "$(run-above)",
+                "enablement": "notebookViewType == jupyter-notebook && jupyter.isnotebooktrusted"
             },
             {
                 "command": "jupyter.notebookeditor.runcellandallbelow",
                 "title": "%DataScience.runBelow%",
                 "category": "Notebook",
-                "icon": {
-                    "light": "resources/light/runbelow.svg",
-                    "dark": "resources/dark/runbelow-inverse.svg"
-                },
+                "icon": "$(run-below)",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -867,6 +863,16 @@
...
             "commandPalette": [
+                {
+                    "command": "jupyter.notebookeditor.runallcellsabove",
+                    "title": "%DataScience.runAbove%",
+                    "when": "false"
+                },
+                {
+                    "command": "jupyter.notebookeditor.runcellandallbelow",
+                    "title": "%DataScience.runBelow%",
+                    "when": "false"
```
DavidKutu, line 874: It doesn't make a lot of sense to me to have cell level commands on the command palette. It could work with the selected cell if there is one, but as a user I don't think it makes sense. That's why I removed them.

Let me know if you disagree.

src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -70,10 +69,6 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
             EditorContexts.IsInteractiveOrNativeActive,
             this.commandManager
         );
-        this.canRunCellsAboveInNativeNotebook = new ContextKey(
-            EditorContexts.canRunCellsAboveInNativeNotebook,
```
DavidKutu, line 74: This context doesn't serve a purpose anymore. Since you can click the run above/below buttons on a cell, the selected cell doesn't matter.

This made it so run above was disabled on all cells, until you selected a cell that's not the first one.

I looked into it, but there's no way of having a context key at the cell level to just disable runAbove on the first cell. The only consequence is that runAbove is available on the first cell and clicking it does nothing.

We could ask vscode to create a context for the cell index. Let me know what you think.

package.json
```diff
@@ -287,20 +287,14 @@
                 "command": "jupyter.notebookeditor.runallcellsabove",
                 "title": "%DataScience.runAbove%",
                 "category": "Notebook",
-                "icon": {
-                    "light": "resources/light/runabove.svg",
-                    "dark": "resources/dark/runabove-inverse.svg"
-                },
-                "enablement": "notebookViewType == jupyter-notebook && jupyter.isnotebooktrusted && jupyter.notebookeditor.canRunCellsAboveInNativeNotebook"
+                "icon": "$(run-above)",
```
DavidKutu, line 290: These icons are the same, might as well use the vs code ones.

rchiodo (approved):

Issue #5242, opened 2021-03-23
Kernel Interrupt command should accept an argument


DonJayamanne:
When we invoke the command `jupyter.notebookeditor.interruptkernel` we make the assumption the document is the active editor.

Assumptions leads to bugs,
We shouldn't make this assumption, instaed we should pass the document as an argument to the command

DonJayamanne:
Validated

PR #5411, opened 2021-04-06
Pass `NotebookDocument` when invoking `jupyter.notebookeditor.interruptkernel`

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/5242


src/client/datascience/types.ts
```diff
@@ -590,6 +591,7 @@ export interface INotebookEditor extends Disposable, IInteractiveBase {
      * Once VSC Notebook is stable, this property can be removed.
      */
     readonly type: 'old' | 'custom' | 'native';
+    readonly document: NotebookDocument | undefined;
```
DonJayamanne, line 594: I'd remove this. 
Basically what this does is we're prolonging the life of `INotebookEditor`.
I'd prefer we don't use such wrappers where possible.
E.g. WebView & Interactive windows don't need this property (yes I can see that from `undefined`.
But we're just adding more stuff to some wrappers and it just keeps getting used elsewheree.

src/client/datascience/interactive-ipynb/nativeEditorCommandListener.ts
```diff
@@ -97,10 +99,11 @@ export class NativeEditorCommandListener implements IDataScienceCommandListener
...
 
-    private interruptKernel() {
-        const activeEditor = this.provider.activeEditor;
-        if (activeEditor) {
-            activeEditor.interruptKernel().ignoreErrors();
+    private interruptKernel(document: NotebookDocument | undefined) {
+        const target =
+            this.provider.editors.find((editor) => editor.document === document) ?? this.provider.activeEditor;
+        if (target) {
+            target.interruptKernel().ignoreErrors();
```
DonJayamanne, line 106: Why not just look for an editor with the same Uri.
If. this.provider.activeEditor.uri === document.uri, then we have what we need. else just iterate through the editors & look for the Uri.

This way you don't need to expose the Document in the Editor object.

DonJayamanne (changes requested):
We don't to add `document` property to INotebookditor

src/client/datascience/interactive-ipynb/nativeEditorCommandListener.ts
joyceerhl: Yeah that was the other idea I had, but I saw that `document` was already public readonly in VSCNotebook and thought adding it to the interface was cleaner. If we don't want to use `INotebookEditor` I'll make your suggested change.

src/client/datascience/interactive-ipynb/nativeEditorCommandListener.ts
```diff
@@ -34,7 +34,9 @@ export class NativeEditorCommandListener implements IDataScienceCommandListener
             commandManager.registerCommand(Commands.NotebookEditorRemoveAllCells, () => this.removeAllCells())
         );
         this.disposableRegistry.push(
-            commandManager.registerCommand(Commands.NotebookEditorInterruptKernel, () => this.interruptKernel())
+            commandManager.registerCommand(Commands.NotebookEditorInterruptKernel, (document: NotebookDocument) =>
```
rchiodo, line 37: When you run this from the command palette doesn't it pass the uri of the file as the first argument? At least execSelectionInInteractive window did that (as in the command palette always passes the URI of the active editor)

joyceerhl: It doesn't seem to pass anything when invoked from the command palette:

image

rchiodo: Might depend upon what has focus when the palette comes up?

IanMatthewHuff: Yeah, I was at least semi-sure that it was the active editor document. It's annoying to track down what parameters are passed for these I've found. I spelunked around in VS Code source before, but I've never found it explicit in the API anywhere that I looked.

joyceerhl: I just tried invoking execSelectionInteractive from the command palette with a selection in a Python file, and I don't see anything being passed as an arg to the handler 😕  I do see the URI being passed when invoked from the context menu, but that's a different scenario. Ian, do you remember how you got that to happen for you?

image


Similarly after setting focus to the notebook cell or the notebook editor, I'm still not seeing the URI getting passed. However at minimum the parameter can clearly be undefined so I should update the type in the handler.

rchiodo (approved):

DonJayamanne (approved):

Issue #5419, opened 2021-04-06
Running the contributed command: 'jupyter.selectjupyteruri' failed


clayms:
Issue Type: <b>Bug</b>
Recent update completely broke Jupyter.

Selecting <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>p</kbd> then `Jupyter: Specify local or remote Jupyter server for connections` gives the following error:
```
Command 'Jupyter: Specify local or remote Jupyter server for connections' resulted in an error (Running the contributed command: 'jupyter.selectjupyteruri' failed.)
```

![image](https://user-images.githubusercontent.com/12793783/113743280-80502780-96d1-11eb-860d-4ea4941066dc.png)

Trying to run python code with <kbd>Shift</kbd>+<kbd>Enter</kbd> gives the following three errors simultaneously:

1. `Unable to open 'Interactive - microsoft/vscode-python#1': Webview overlay is disposed.`
2. `e.toLowerCase is not a function`
3.  `e.toLowerCase is not a function`

![image](https://user-images.githubusercontent.com/12793783/113743662-ea68cc80-96d1-11eb-9c11-729dca1a940c.png)

Extension version: 2021.3.680753044
VS Code version: Code 1.55.0 (c185983a683d14c396952dd432459097bc7f757f, 2021-03-30T16:01:55.261Z)
OS version: Windows_NT x64 10.0.18363

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-6300U CPU @ 2.40GHz (4 x 2496)|
|GPU Status|2d_canvas: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>protected_video_decode: unavailable_off<br>rasterization: enabled<br>skia_renderer: enabled_on<br>video_decode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|15.88GB (6.38GB free)|
|Process Argv|--crash-reporter-id f652528b-5678-4b12-9ca0-3de6b1e1135f|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vsreu685:30147344
python383cf:30185419
vspor879:30202332
vspor708:30202333
vspor363:30204092
[...]

PR #5426, opened 2021-04-06
Delete extension context secrets if we get an error when getting them.

DavidKutu:
And small fixes on error handling.

For #5419

### DIFF
--- a/src/client/common/application/encryptedStorage.ts
+++ b/src/client/common/application/encryptedStorage.ts
@@ -34,8 +34,14 @@ export class EncryptedStorage implements IEncryptedStorage {
         if (IS_REMOTE_NATIVE_TEST && this.extensionContext.extensionMode !== ExtensionMode.Production) {
             return this.testingState.get(`${service}#${key}`);
         }
-        // eslint-disable-next-line
-        const val = await this.extensionContext.secrets.get(`${service}.${key}`);
-        return val;
+        try {
+            // eslint-disable-next-line
+            const val = await this.extensionContext.secrets.get(`${service}.${key}`);
+            return val;
+        } catch (e) {
+            // If we get an error trying to get a secret, it might be corrupted. So we delete it.
+            await this.extensionContext.secrets.delete(`${service}.${key}`);
+            return;
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


src/client/common/application/encryptedStorage.ts
```diff
@@ -34,8 +34,14 @@ export class EncryptedStorage implements IEncryptedStorage {
...
-        // eslint-disable-next-line
-        const val = await this.extensionContext.secrets.get(`${service}.${key}`);
-        return val;
+        try {
+            // eslint-disable-next-line
+            const val = await this.extensionContext.secrets.get(`${service}.${key}`);
+            return val;
+        } catch (e) {
+            // If we get an error trying to get a secret, it might be corrupted. So we delete it.
+            await this.extensionContext.secrets.delete(`${service}.${key}`);
```
DavidKutu, line 43: According to this issue https://github.com/microsoft/vscode/issues/120558, an encryption change in Windows requires us to re enter every secret.

I checked and this is the only place in the repo where .get() gets called. So if we get an error while getting, we can delete and then enter the secret again as usual.

Thanks @IanMatthewHuff for the help.

src/client/common/errors/errorUtils.ts
```diff
@@ -41,7 +41,9 @@ export function getLastFrameFromPythonTraceback(
     }
     //             File "/Users/donjayamanne/miniconda3/envs/env3/lib/python3.7/site-packages/appnope/_nope.py", line 38, in C
 
-    const lastFrame = traceback
+    // This parameter might be either a string or a string array
+    const fixedTraceback: string = Array.isArray(traceback) ? traceback[0] : traceback;
```
DavidKutu, line 45: It makes more sense to change the parameter type, but the issue is that JS allows it to be either an array or a string. This works for both cases.

rchiodo (approved):

src/client/common/application/encryptedStorage.ts
DavidKutu: We considered what other secret could be affected by this, but it looks like its only the remote uri's.

Issue #5368, opened 2021-03-31
Install ipykernel is coming up during auto start. It shouldn't.


rchiodo:
Turn off native notebook
Create notebook with kernel that requires ipykernel to be installed
Reload VS code
Open notebook
As you open the notebook the 'You need to install ipykernel' comes up.

joyceerhl:
Validated. In custom editor when I first open a notebook connected to a kernel that does not have ipykernel installed, I don't get prompted to install ipykernel. I only get prompted when I run a cell. (Tested with new blank notebook and with opening an existing notebook.)

One thing that is sort of odd is that the kernel status is reported as Busy, and the kernel name isn't displayed.

![image](https://user-images.githubusercontent.com/30305945/114088747-3189c580-986a-11eb-973c-50207ceccd3b.png)

PR #5440, opened 2021-04-07
Stop asking users to install ipykernel on autostart

DavidKutu:
only do it when a cell is ran.

For #5368

### DIFF
--- a/src/client/datascience/interactive-common/interactiveBase.ts
+++ b/src/client/datascience/interactive-common/interactiveBase.ts
@@ -902,12 +902,13 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
         }
     }
 
-    protected async createNotebookIfProviderConnectionExists(disableUI: boolean): Promise<void> {
+    protected async createNotebookIfProviderConnectionExists(disableUI?: boolean): Promise<void> {
         // Check to see if we are already connected to our provider
         const providerConnection = await this.notebookProvider.connect({
             getOnly: true,
             resource: this.owningResource,
-            metadata: this.notebookMetadata
+            metadata: this.notebookMetadata,
+            disableUI
         });
 
         if (providerConnection) {
--- a/src/client/datascience/interactive-common/interactiveBase.ts
+++ b/src/client/datascience/interactive-common/interactiveBase.ts
@@ -229,8 +229,9 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
         this.disposables.push(this.jupyterVariables.refreshRequired(this.refreshVariables.bind(this)));
 
         // If we have already auto started our server then we can go ahead and try to create a notebook on construction
+        // Disable the UI to avoid errors before the user runs a cell
         setTimeout(() => {
-            this.createNotebookIfProviderConnectionExists().ignoreErrors();
+            this.createNotebookIfProviderConnectionExists(true).ignoreErrors();
         }, 0);
     }
 
@@ -887,7 +888,7 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
     // ensureNotebook can be called apart from ensureNotebookAndServer and it needs
     // the same protection to not be called twice
     // eslint-disable-next-line @typescript-eslint/member-ordering
-    protected async ensureNotebook(serverConnection: INotebookProviderConnection, disableUI: boolean): Promise<void> {
+    protected async ensureNotebook(serverConnection: INotebookProviderConnection, disableUI = false): Promise<void> {
         if (!this.notebookPromise) {
             this.notebookPromise = this.ensureNotebookImpl(serverConnection, disableUI);
         }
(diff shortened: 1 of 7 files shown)


src/client/datascience/interactive-common/interactiveBase.ts
```diff
@@ -911,7 +911,7 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
 
         if (providerConnection) {
             try {
-                await this.ensureNotebook(providerConnection);
+                await this.ensureNotebook(providerConnection, true);
```
rchiodo, line 914: Not sure why this one is true and the other one is false?

DavidKutu: I found it by testing

rchiodo: That makes it hard to reason about this later. Is there a more concrete way to follow this up to the decision about using auto start or not?

DavidKutu: A clue I got was that since this function already hardcodes `disableUI` to false, that one call should also hardcode a false (it makes sense this is called when running a cell). And there's only one more place where `ensureNotebook` is called.

image

DavidKutu: Maybe some comments would help?

rchiodo: In your picture I'm not sure that disableUI should be false or not. Because I don't know how that decision was made.

That also means on the spot where you have true, you should also have disableUI: true for ensureConnection.

I think it would just be clearer if the disableUI flag was pushed even higher. Like createNotebookIfProviderConnectionExists is called in the ctor, that one seems obvious that you'd want to disable the UI. The other spot where's it called, maybe not?

IanMatthewHuff: Could maybe make the default to false for the disableUI parameter? It seems like it's the exceptional case the disable the UI. So you could call the true out here with a comment as to why, and the other spots don't need a random false drifting around.

@rchiodo would that be cleaner?

rchiodo: I'd still want to know why it's true at this spot. I think the true spot isn't always correct either. Based on how ```createNotebookIfProviderConnectionExists``` is called. It should be true on the first call (in the ctor). But should it be for notebookProvider.connectionMade? Only if notebookProvider was called with disableUI set to true itself.

rchiodo (changes requested):
I think createNotebookIfProviderConnectionExists should be changed to add a flag and it should only be set to true in the ctor call.

This also means in the same function the call to this.notebookProvider.connect should also pass the disable UI flag.

src/client/datascience/interactive-common/interactiveBase.ts
```diff
@@ -901,7 +902,7 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
         }
     }
 
-    protected async createNotebookIfProviderConnectionExists(): Promise<void> {
+    protected async createNotebookIfProviderConnectionExists(disableUI: boolean): Promise<void> {
         // Check to see if we are already connected to our provider
         const providerConnection = await this.notebookProvider.connect({
             getOnly: true,
```
rchiodo, line 908: This call here needs the 'disableUI' flag too.

src/client/datascience/interactive-common/notebookProvider.ts
```diff
@@ -176,7 +176,8 @@ export class NotebookProvider implements INotebookProvider {
     }
 
     private fireConnectionMade() {
-        this._connectionMade.fire();
+        // Disable the UI to avoid errors before the user runs a cell
```
rchiodo, line 179: I don't think this  should always be true. It should be based on the options passed to the connect call.

rchiodo: In the spots where this is bound, you can likely bind it to the options passed into the caller.

rchiodo: Sorry it looks like there are two more spots. In src\client\datascience\interactive-common\notebookProvider.ts where the connectionMade event is bound to the fireConnectionMade function. You should bind the first parameter to the options.DisableUI setting.

rchiodo: Line 86 and 95 in notebookProvider.ts

src/client/datascience/interactive-ipynb/nativeEditor.ts
```diff
@@ -448,7 +448,8 @@ export class NativeEditor extends InteractiveBase implements INotebookEditor {
 
     protected async createNotebookIfProviderConnectionExists() {
```
rchiodo, line 449: Doesn't this overload of the function need it too?

rchiodo (approved):

Issue #5436, opened 2021-04-07
Native notebooks - can run cells even when untrusted


rchiodo:
Open notebook in jupyter classic
Open in insiders. It should say untrusted. Close trust notification without accepting yes.

I can still run a cell.

rchiodo:
If run is clicked we shouldn't run but instead put up the error message.

IanMatthewHuff:
Validated. I'm blocked from execution after closing the trust prompt.

PR #5464, opened 2021-04-09
Cell execution requests should block on trust prompt in untrusted notebooks

joyceerhl:
For #5436

### DIFF
--- a/src/client/datascience/interactive-ipynb/trustCommandHandler.ts
+++ b/src/client/datascience/interactive-ipynb/trustCommandHandler.ts
@@ -9,7 +9,7 @@ import { IExtensionSingleActivationService } from '../../activation/types';
 import { IApplicationShell, ICommandManager } from '../../common/application/types';
 import { ContextKey } from '../../common/contextKey';
 import '../../common/extensions';
-import { traceInfo } from '../../common/logger';
+import { traceError, traceInfo } from '../../common/logger';
 import { IDisposableRegistry } from '../../common/types';
 import { swallowExceptions } from '../../common/utils/decorators';
 import { DataScience } from '../../common/utils/localize';
@@ -65,10 +65,14 @@ export class TrustCommandHandler implements IExtensionSingleActivationService {
                 sendTelemetryEvent(Telemetry.TrustAllNotebooks);
                 break;
             case DataScience.trustNotebook():
-                if (model instanceof VSCodeNotebookModel) {
-                    await this.trustNativeNotebook(model);
-                } else {
-                    await this.trustNotebook(model);
+                try {
+                    if (model instanceof VSCodeNotebookModel) {
+                        await this.trustNativeNotebook(model);
+                    } else {
+                        await this.trustNotebook(model);
+                    }
+                } catch (e) {
+                    traceError('Error while trusting notebook', e);
                 }
                 break;
             case DataScience.doNotTrustNotebook():
--- a/src/client/datascience/interactive-ipynb/trustCommandHandler.ts
+++ b/src/client/datascience/interactive-ipynb/trustCommandHandler.ts
@@ -40,15 +40,15 @@ export class TrustCommandHandler implements IExtensionSingleActivationService {
         this.disposables.push(this.commandManager.registerCommand(Commands.NotebookTrusted, () => noop(), this));
     }
     @swallowExceptions('Trusting notebook')
-    private async onTrustNotebook(uri?: Uri) {
+    private async onTrustNotebook(uri?: Uri): Promise<boolean> {
         uri = uri ?? this.editorProvider.activeEditor?.file;
         if (!uri) {
-            return;
+            return true;
         }
 
         const model = await this.storageProvider.getOrCreateModel({ file: uri });
         if (model.isTrusted) {
-            return;
+            return true;
         }
         traceInfo('Display prompt to trust notebook');
         const selection = await this.applicationShell.showErrorMessage(
@@ -77,6 +77,7 @@ export class TrustCommandHandler implements IExtensionSingleActivationService {
             default:
                 break;
         }
+        return selection === DataScience.trustNotebook();
     }
     private async trustNotebook(model: INotebookModel) {
         // Update model trust
--- a/src/client/datascience/notebook/kernelWithMetadata.ts
+++ b/src/client/datascience/notebook/kernelWithMetadata.ts
@@ -63,6 +63,13 @@ export class VSCodeNotebookKernelMetadata implements VSCNotebookKernel {
      * resolved, the cell will be put back into the Idle state.
      */
     public async executeCellsRequest(document: NotebookDocument, ranges: NotebookCellRange[]): Promise<void> {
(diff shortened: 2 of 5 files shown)


src/client/datascience/notebook/kernelWithMetadata.ts
```diff
@@ -71,7 +78,7 @@ export class VSCodeNotebookKernelMetadata implements VSCNotebookKernel {
                     ranges.some((range) => range.start <= cell.index && cell.index < range.end)
             );
         traceInfo(`Execute Cells request ${cells.length} ${cells.map((cell) => cell.index).join(', ')}`);
-        await cells.map((cell) => this.executeCell(document, cell));
+        await Promise.all(cells.map((cell) => this.executeCell(document, cell)));
```
joyceerhl, line 81: I'm not sure about this change, but the original code we had here looked a little odd to me. We were awaiting on an array of promises, which TypeScript flagged as a no-op:
image

I think the intent was to await on Promise.all of that array so I went ahead and made the change.

src/client/datascience/notebook/kernelWithMetadata.ts
```diff
@@ -63,6 +63,13 @@ export class VSCodeNotebookKernelMetadata implements VSCNotebookKernel {
      * resolved, the cell will be put back into the Idle state.
      */
     public async executeCellsRequest(document: NotebookDocument, ranges: NotebookCellRange[]): Promise<void> {
+        // When we receive a cell execute request, first ensure that the notebook is trusted.
+        // If it isn't already trusted, block execution until the user trusts it.
```
DonJayamanne, line 67: Previously, one could not execute a cell of the notebook was not trusted.
Looks like something used now changes that's causing the run to be enabled.

I personally think that's a good thing. However I'd like  to understand what changed. Did something in vscode change?

Because right now we're not really fixing a big were fixing a regressing, and there's no choice change in here that indicates something broke...

Also would be good to have a test for this.

joyceerhl: Yeah, still draft status, been trying to get tests passing locally before I push :) 

I'm not entirely sure what changed to be honest. I imagine there could have been an unintentional change while refactoring, and we didn't have tests to enforce this specific scenario. 

> I personally think that's a good thing. 

FWIW I agree, we talked about this at triage and said the same, but it's still not working as designed. I'm also looking forward to being able to deprecate our trust support when VS Code releases trusted workspaces (I'm seeing their WIP API for trusted notebooks in the proposed API already)

rchiodo (approved):

src/client/datascience/interactive-ipynb/trustCommandHandler.ts
```diff
@@ -77,6 +77,7 @@ export class TrustCommandHandler implements IExtensionSingleActivationService {
             default:
                 break;
         }
+        return selection === DataScience.trustNotebook();
```
IanMatthewHuff, line 80: Are trustNativeNotebook and trustNotebook guaranteed to work or throw? Just making sure here as we're basing the return on the users selection, not the result of the trust operation.

joyceerhl: Good catch (no pun intended). The part where we stringify the notebook contents for trusting may throw if a circular reference is detected in the notebook JSON (which shouldn't happen). The rest of those methods as currently implemented won't throw but there's nothing enforcing that right now, so I'll add a catch for errors from calling those methods.

src/client/datascience/interactive-ipynb/trustCommandHandler.ts
```diff
@@ -65,10 +65,14 @@ export class TrustCommandHandler implements IExtensionSingleActivationService {
...
-                } else {
-                    await this.trustNotebook(model);
+                try {
+                    if (model instanceof VSCodeNotebookModel) {
+                        await this.trustNativeNotebook(model);
+                    } else {
+                        await this.trustNotebook(model);
+                    }
+                } catch (e) {
+                    traceError('Error while trusting notebook', e);
```
DonJayamanne, line 75: Why would this fail?

src/client/datascience/notebook/kernelWithMetadata.ts
DonJayamanne: Yikes, good find.

src/client/datascience/notebook/kernelWithMetadata.ts
```diff
@@ -63,6 +63,13 @@ export class VSCodeNotebookKernelMetadata implements VSCNotebookKernel {
...
+        const isTrusted = await this.commandManager.executeCommand(Commands.TrustNotebook, document.uri);
```
DonJayamanne, line 68: i'd first check if the notebook is already trusted. if not trusted, then perform this async operation.

DonJayamanne:
```
        const model = this.storageProvider.get(document.uri);
        if (model && model.isTrusted === false) {
			const isTrusted = await this.commandManager.executeCommand(Commands.TrustNotebook, document.uri);
```

Else calling a method to trust a notebook when its already trusted seems wrong.

DonJayamanne (changes requested):
Change to invoke trust command only if not trusted.

src/client/datascience/notebook/kernelWithMetadata.ts
joyceerhl: I'm not sure I understand why that's necessary. This command already checks if the model is trusted and returns true if it is, so we'd be performing the same check twice if the document wasn't trusted. This would also necessitate injecting the NotebookStorageProvider into the VSCNotebookKernel, which seems like a weird dependency for VSCNotebookKernel to have to take.

src/client/datascience/interactive-ipynb/trustCommandHandler.ts
joyceerhl: Currently they shouldn't, but there's nothing enforcing that either trustNativeNotebook or trustNotebook cannot throw.

src/client/datascience/notebook/kernelWithMetadata.ts
DonJayamanne: > his command already checks if the model is trusted and returns true if it is, so we'd be performing the same check twice if the document wasn't trusted

I don't like the idea of calling a command `trustNotebook` when its already trusted. We should call methods `TrustNotebook` only if something is not trusted.

But, now i see that this is not somethign I should block the PR on. hence approving (as its just a personal preference).

DonJayamanne (approved):

src/client/datascience/notebook/kernelWithMetadata.ts
joyceerhl: Yeah the command name threw me too. Perhaps it should be renamed to `Commands.PromptToTrustNotebook`?

DonJayamanne: I've leave it as is for now, this will go away soon, when VS Code has trust.

PR #5499, opened 2021-04-12
Updates as a result of changes to Proposed NB API

DonJayamanne:
For https://github.com/microsoft/vscode-jupyter/issues/5491

### DIFF
--- a/types/vscode-proposed/index.d.ts
+++ b/types/vscode-proposed/index.d.ts
@@ -173,17 +173,45 @@ export interface NotebookDocument {
     readonly uri: Uri;
     readonly version: number;
 
+    /** @deprecated Use `uri` instead */
     // todo@API don't have this...
     readonly fileName: string;
 
     readonly isDirty: boolean;
     readonly isUntitled: boolean;
-    readonly cells: ReadonlyArray<NotebookCell>;
+
+    /**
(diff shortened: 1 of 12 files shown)


joyceerhl (approved):
Out of curiosity why do we have
1. types/vscode-proposed/index.d.ts
2. types/vscode.proposed.d.ts
3. typings/vscode-proposed/index.d.ts?

I see all three being updated in this PR.

DonJayamanne:
> Out of curiosity why do we have

Good question. I have no idea. I think this is stuff coming from Python extension.

types/vscode-proposed/index.d.ts
```diff
@@ -173,18 +173,46 @@ export interface NotebookDocument {
     readonly uri: Uri;
     readonly version: number;
 
+    /** @deprecated Use `uri` instead */
```
IanMatthewHuff, line 176: Why do we seem to have three proposed API files? I only updated types/vscode.proposed.d.ts in my last API change.

IanMatthewHuff (approved):

types/vscode-proposed/index.d.ts
DonJayamanne: I'll remove the others in the separate PR.

Issue #5495, opened 2021-04-12
Interpreter information stored in kernelspec not used when starting Kernels


DonJayamanne:
* Select kernel pointing to virtual env
* Run kenrel & it points to something else

I've been using this notebook & kernel for months.
Kernel has been around for almost a year.
For some reason now its not getting picked

DonJayamanne:
Assigning to myself to identify the cause.

dgerzhoy:
I've found this same thing happening to me. It defaults to an environment that doesn't have IPython installed, and fails, even though I have the correct environment selected (and run) in the workspace.

If this tidbit helps: I am able to go into htop and kill all the processes related to vscode, and then when I restart vscode, it works again.

(I am running the notebook on linux, through vscode remote from a windows laptop)

DonJayamanne:
@dgerzhoy Please could you install the latest VS Code Insiders, then yo'ull automatically get a version of the Jupyter extesnion with the latest build bits. If you cannot repro it there then it means it has been fixed, else you're experiencing a whole new issue, which means its time to (please) create a seprate issue.

rchiodo:
Seems to be validated.

PR #5500, opened 2021-04-12
Use interpreter info stored in kernelspec.json

DonJayamanne:
For #5495

Not sure if we need this in the point release, but this is a regression.

### DIFF
--- a/src/client/datascience/kernel-launcher/localKernelFinder.ts
+++ b/src/client/datascience/kernel-launcher/localKernelFinder.ts
@@ -224,7 +224,6 @@ export class LocalKernelFinder implements ILocalKernelFinder {
                                 return interpreter;
                             });
                     }
-                    // No interpreter found. If python, use the active interpreter anyway
                     const result: KernelSpecConnectionMetadata = {
                         kind: 'startUsingKernelSpec',
                         kernelSpec: k,
--- a/src/client/datascience/kernel-launcher/localKernelFinder.ts
+++ b/src/client/datascience/kernel-launcher/localKernelFinder.ts
@@ -11,7 +11,7 @@ import { IWorkspaceService } from '../../common/application/types';
 import { PYTHON_LANGUAGE } from '../../common/constants';
 import { traceDecorators, traceError, traceInfo, traceInfoIf } from '../../common/logger';
 import { IFileSystem, IPlatformService } from '../../common/platform/types';
-import { IExtensionContext, IPathUtils, Resource } from '../../common/types';
+import { IExtensionContext, IPathUtils, ReadWrite, Resource } from '../../common/types';
 import { IEnvironmentVariablesProvider } from '../../common/variables/types';
 import { IInterpreterService } from '../../interpreter/contracts';
 import { PythonEnvironment } from '../../pythonEnvironments/info';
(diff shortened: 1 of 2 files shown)


src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -447,7 +464,13 @@ export class LocalKernelFinder implements ILocalKernelFinder {
                 ? interpreter?.displayName || kernelJson.display_name
                 : kernelJson.display_name;
 
-        const kernelSpec: IJupyterKernelSpec = new JupyterKernelSpec(kernelJson, specPath, interpreter?.path);
+        const kernelSpec: IJupyterKernelSpec = new JupyterKernelSpec(
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            kernelJson as any,
+            specPath,
+            // Interpreter information may be saved in the metadata (if this is a kernel spec created/registered by us).
```
DonJayamanne, line 471: This was missing.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -189,35 +189,52 @@ export class LocalKernelFinder implements ILocalKernelFinder {
...
+                    filteredInterpreters = filteredInterpreters.filter((i) => !matchingInterpreters.includes(i));
+
+                    // Return our metadata that uses an interpreter to start
+                    return result;
+                } else {
+                    let interpreter = k.language === PYTHON_LANGUAGE ? activeInterpreter : undefined;
+                    // If the interpreter information is stored in kernelspec.json then use that to determine the interpreter.
+                    if (
+                        k.language === PYTHON_LANGUAGE &&
+                        this.extensionChecker.isPythonExtensionInstalled &&
```
DonJayamanne, line 216: Even if we had interpreter information in the kernelspec file, we weren't using that. This fixes that.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -189,35 +189,52 @@ export class LocalKernelFinder implements ILocalKernelFinder {
...
+                        k.metadata?.interpreter?.path !== activeInterpreter?.path
+                    ) {
+                        interpreter = await this.interpreterService
+                            .getInterpreterDetails(k.metadata?.interpreter?.path)
+                            .catch((ex) => {
+                                traceError(`Failed to get interpreter details for Kernel Spec ${k.specFile}`, ex);
+                                return interpreter;
+                            });
+                    }
+                    // No interpreter found. If python, use the active interpreter anyway
```
rchiodo, line 227: Comment here is a little confusing. It's more like use the closest match, if not found then use the active interpreter.

rchiodo (approved):

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: Oops, old comment

IanMatthewHuff (approved):

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -189,35 +189,51 @@ export class LocalKernelFinder implements ILocalKernelFinder {
...
-                };
-
-                // If interpreters were found, remove them from the interpreter list we'll eventually
-                // return as interpreter only items
-                filteredInterpreters = filteredInterpreters.filter((i) => !matchingInterpreters.includes(i));
+        const kernelMetadata = await Promise.all(
+            kernelSpecs.map(async (k) => {
+                // Find the interpreter that matches. If we find one, we want to use
+                // this to start the kernel.
+                const matchingInterpreters = this.findMatchingInterpreters(k, interpreters);
```
rchiodo, line 196: findMatchingInterpreters is supposed to look at the kernel metadata. How come it didn't find it?

rchiodo: Meaning your fix should be redundant.

rchiodo (changes requested):
⏲️

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: Because the interpreter is a virtual environment in a workspace that not visible in another workspace.

E.g. Worksapce A - Virtual Env - PythonABC
Open sub directory in Worksapce A or anotehr worksapce, now the Virtual Env Python ABC is not listed as a python environment. But this is listed as a valid kernel because we have regsitered this kernel.

So, when you go into anotehr worksapce folder and open an existing notebook or or a new notebook & select an existing kernel, it will not work.

DonJayamanne: i.e. workspace virtual envs don't appear in all worksapces in the interpreter list, but kernels pecs do (as they are registered gloablly)

rchiodo: So getInterpreterDetails will work even though it's not an allowed interpreter?

rchiodo: If that's true, it seems like we should be calling that in findMatchingInterpreters?

joyceerhl (commented):
It's somewhat worrying that we managed to regress this. Should we add a test?

src/client/datascience/kernel-launcher/localKernelFinder.ts
rchiodo: Actually no where you have it seems better to me. Maybe a comment saying the reasons why this can happen though would be nice. 

The call to findResourceInterpreters doesn't find interpreters that are global.

DonJayamanne: > So getInterpreterDetails will work even though it's not an allowed interpreter?

Its not correct to say `allowed interrpeter`, its basically an interpreter that wasn't discovered.

DonJayamanne: Will update comments.

rchiodo (approved):

DonJayamanne:
> It's somewhat worrying that we managed to regress this. Should we add a test?

I don't think this is a regression now that i think about it.
Adding a test will be difficult. Will add this seprately.

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: Done

DonJayamanne:
@joyceerhl  i created an issue here https://github.com/microsoft/vscode-jupyter/issues/5504
To add the tests, will work on that next (still not sure how to setup this scenario but will address that in the test PR).

Issue #5481, opened 2021-04-11
Native Notebooks - Green Check + Timer after kernel restart should be removed


claudiaregio:
## Steps to Reproduce:
1. Open notebook
2. Run some cells
3. Clear outputs or restart kernel

## Actual Behavior: 
**If you clear all outputs:** 
Green check mark remains even though execution count is wiped

![image](https://user-images.githubusercontent.com/19276747/114314962-c16d8080-9ab1-11eb-9722-513edd79faa1.png)

This is especially weird if you delete the code before you clear all outputs. 
![image](https://user-images.githubusercontent.com/19276747/114315299-38574900-9ab3-11eb-9ec4-64c30fcc7398.png)

**If you restart kernel:** 
This looks identical to if I had actually just run it

![image](https://user-images.githubusercontent.com/19276747/114315270-12ca3f80-9ab3-11eb-8340-5b5488bb7973.png)

## Expected Behavior: 
No green check mark or timer in cell status bar when kernel is restarted or all outputs cleared

roblourens:
This is now done correctly for clearing outputs. But for restarting, that would have to be done on the extension side with a fake execution event. I will give it back to you @DonJayamanne to decide whether it's right to clear those when the kernel is restarted. I have no idea how to do that in the extension codebase. But if you want me to tackle it you can point me to the right code

DonJayamanne:
Jupyter has the following options:
* Restart
* Restart and Clear Outputs

The latter ends up clearing the execution count and the output. We do not have such an option.
I don't think we should remove the green tick and not clear the execution count, that would seem an incomplete behaviour.
I think we should instead have the first option (current behaviour which lines up with Jupyter) and the second option (new restart option)

This means we'd need a different restart button or a split dropdown button or the like.

Given the fact that this issue has been open for more than a year and we've only had 2 similar issues, I'm inclined to close this due to a lack of interest from users.

PR #5507, opened 2021-04-13
Clear native notebook cell check mark and timer on kernel restart

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/5481

clearonrestart


rchiodo (approved):

src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -380,4 +382,28 @@ export class NotebookEditor implements INotebookEditor {
...
+                        cell.kind,
+                        cell.document.getText(),
+                        cell.document.languageId,
+                        outputs,
+                        cell.metadata,
+                        // Retain execution order to match Jupyter behavior
+                        { executionOrder: cell.latestExecutionSummary?.executionOrder }
+                    );
+                });
+                edit.replaceNotebookCells(editor.document.uri, new NotebookRange(0, this.document.cellCount), newCells);
```
DonJayamanne, line 405: @rebornix @roblourens Is there a better way to achieve this, rather than having to delete & re-create all of the cells.

roblourens: I assume that if it was running previously you have an execution task, can't you use that and end the execution?

roblourens: Oh, you are trying to clear the state after it has finished executing. Do you want to clear outputs too? If so you can run the clear outputs command. Or, you can create execution tasks and complete them immediately with no success or time. Trying to figure out what a better way to do this would look like and I don't really see how it fits into the model. I think that doing it with execution tasks would be reasonable though. Basically that's just how execution-related state is manipulated.

roblourens: If we can think of a general-purpose reason to have this, we could add a "clear execution state" command (like "clear outputs" but doesn't touch outputs)

DonJayamanne: > I think that doing it with execution tasks would be reasonable though. Basically that's just how execution-related state is manipulated.

Yup thats what I thought as well. 
In general I'm assuming thats better than using the replace cells (as that could have other perf implications as well, on the UI & live share, & other areas).

Thanks @roblourens

joyceerhl: JFYI, we don't want to clear outputs in order to match Jupyter behavior.

DonJayamanne:
@joyceerhl are we planning on merging this PR?
Personally i think we should, if the concern is that solution isn't great, then we might want to discuss this in the VSC Sync meeting.
I think we need to fix this to be consistent withJupyter.

joyceerhl:
@DonJayamanne Thanks for the reminder, I was waiting on @claudiaregio for her thoughts on whether we want this implemented given that there's no direct analogue for the cell status concept in classic Jupyter.  If we are all in agreement that we should include this I'm happy to merge it.

Issue #5509, opened 2021-04-13
Better telemetry when interpreter doesn't match the kernel


DonJayamanne:
We track this today, but we don't know if:
* User ran a cell
* Whether its notebook or interactive or native or custom, etc...

Basically we don't have a lot of information we have with other events.
Using the existing infrastructure for this.

IanMatthewHuff:
I don't have a specific repro for this, but validating on the implementation of the telemetry event.

PR #5510, opened 2021-04-13
Better logs when interpreter does not match kernel

DonJayamanne:
For #5509

### DIFF
--- a/src/client/datascience/jupyter/kernels/helpers.ts
+++ b/src/client/datascience/jupyter/kernels/helpers.ts
@@ -32,6 +32,8 @@ import { traceInfo, traceInfoIf } from '../../../common/logger';
 import { getInterpreterHash } from '../../../pythonEnvironments/info/interpreter';
 import { getTelemetrySafeVersion } from '../../../telemetry/helpers';
 import { IS_CI_SERVER } from '../../../../test/ciConstants';
+import { trackKernelResourceInformation } from '../../telemetry/telemetry';
+import { Uri } from 'vscode';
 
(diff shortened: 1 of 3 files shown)


src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -464,10 +465,7 @@ export async function sendTelemetryForPythonKernelExecutable(
         }
         const sysExecutable = concatMultilineString(output.text).trim().toLowerCase();
         const match = kernelConnection.interpreter.path.toLowerCase() === sysExecutable;
-        sendTelemetryEvent(Telemetry.PythonKerneExecutableMatches, undefined, {
-            match: match ? 'true' : 'false',
-            kernelConnectionType: kernelConnection.kind
-        });
+        trackKernelResourceInformation(Uri.file(file), { interpreterMatchesKernel: match });
```
DonJayamanne, line 468: With the existing infrastructure we can tell:
* If its a notebook vs interactive window
* If its custom vs native notebook
* If user ran the cell vs auto start
All three are crucial and this wasn't easy to get with the previous event.

Removing the old item, once we add this, we can deprecate the old event.

IanMatthewHuff (approved):

DavidKutu (approved):

DonJayamanne:
@IanMatthewHuff @DavidKutu  I reverted the code that removed the old telemetry, will need that to compare in the future (see if we're getting better)

rchiodo (approved):

Issue #5527, opened 2021-04-15
Update to the new CellStatusBar API breaking change due to hit today


IanMatthewHuff:
https://github.com/microsoft/vscode/issues/93265#issuecomment-819738397

PR #5530, opened 2021-04-15
Notebook Dev breaking changes

IanMatthewHuff:
For #5527


src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -296,8 +296,8 @@ export class CellExecution {
         if (this.isEmptyCodeCell) {
             this.task?.end({});
         } else if (success === 'success' || success === 'failed') {
-            this.lastRunDuration = this.stopWatch.elapsedTime;
-            this.task?.end({ duration: this.lastRunDuration, success: success === 'success' });
+            this.endTime = new Date().getTime();
```
IanMatthewHuff, line 299: Peng was right. I did this endTime convert quickly and I'm not fully sure that it's handling the elapsed time correctly. Let me re-look quick.

src/client/datascience/notebook/statusBarProvider.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * This class controls the status messages that we see on our Notebook cells
+ */
+@injectable()
+export class StatusBarProvider implements VSCNotebookCellStatusBarItemProvider {
+    // Weakmap, as we don't own the lifetime of either of these, just a helpful mapping for tests
+    private static cellStatusMappings = new WeakMap<NotebookCell, NotebookCellStatusBarItem>();
+
+    // Allow test code to see what status is currently on any cell
+    public static getCellStatusBarItem(cell: NotebookCell): NotebookCellStatusBarItem | undefined {
```
IanMatthewHuff, line 26: Instead of adding to the existing tracker I just put the tracker for tests on there with the same basic scheme. Seemed easier than having to throw an event from here and handle it elsewhere since there is no notebook API for when cells get new status items.

src/test/datascience/notebook/executionService.vscode.test.ts
```diff
@@ -203,7 +204,8 @@ suite('DataScience - VSCode Notebook - (Execution) (slow)', function () {
         expect(cell.latestExecutionSummary?.executionOrder).to.be.greaterThan(0, 'Execution count should be > 0');
         assert.equal(NotebookCellStateTracker.getCellState(cell), NotebookCellExecutionState.Idle, 'Incorrect State');
         assert.isFalse(hasErrorOutput(cell.outputs), 'Incorrect State');
-        assert.equal(cell.metadata.statusMessage, '', 'Incorrect Status message');
+        const statusBarItem = StatusBarProvider.getCellStatusBarItem(cell);
```
IanMatthewHuff, line 207: Got a clean test run on `notebook/dev` with this.

IanMatthewHuff:
@rebornix You don't have to review, but adding you on as we've been chatting about these changes some.

rchiodo (approved):

src/client/datascience/notebook/statusBarProvider.ts
```diff
@@ -0,0 +1,43 @@
...
+        return StatusBarProvider.cellStatusMappings.get(cell);
+    }
+
+    // For any NotebookCell, check its output to see if we need to put up a status message
+    provideCellStatusBarItems(
```
DonJayamanne, line 31: Won't we need to trigger an event letting VS Code know about potential changes?
* @rebornix 
Or will VS Code always call this method when ever there are changes to notebooks?

Instances when VS Code should call this:
* Execution has completed (outputs changed)
* Execution has gone into pending state (outputs changed)
* User clears outputs (outputs changed)

Basically when ever outputs change we need to ensure we either notify VS Code to call this or VS Code should call this when outputs change.

@IanMatthewHuff @rebornix  do we know if VS Code will call when outputs change?
If not then we'll need to trigger the event in this provider in all of the above cases.

Not sure about the scenario when user changes outputs (@rebornix do we have events when user cleas outputs?)
There are other places where we want to clea other metadata when user clears outputs.

DonJayamanne: This isn't a showstopper, however we'd need to ensure this works as expected.

DonJayamanne (approved):

src/client/datascience/notebook/statusBarProvider.ts
IanMatthewHuff: @DonJayamanne What I did check was the following.

* Run a cell without an error (no status message)
* Create an error in the cell and run it (error status message show)
* Fix the error and run again (no status message)

I'll check for output clearing now.

src/client/datascience/jupyter/kernels/cellExecution.ts
IanMatthewHuff: Actually I think this is ok. Stopwatch still used to send durations for telemetry, we start it right where we set start time and just get an endTime here now, should be legit.

src/client/datascience/notebook/statusBarProvider.ts
DonJayamanne: > Fix the error and run again (no status message)

Is this purely by chance of is this how VS Code works. I.e. does it query the methods when ever cell output or status changes? I'd prefer we know that for a fact instead of leaving this to chance.
@rebornix Please could you let us know when VS Code will query this method to get the latest status?

DavidKutu (approved):

src/client/datascience/notebook/statusBarProvider.ts
roblourens: It will be called

- When a cell begins or finishes executing (really when any internal metadata changes)
- When output changes
- When language changes
- When the cell scrolls into view
- When any provider fires `onDidChangeCellStatusBarItems`

I should have jsdoc that spells this out

src/client/datascience/notebook/statusBarProvider.ts
```diff
@@ -0,0 +1,43 @@
...
+        cell: NotebookCell,
+        _token: CancellationToken
+    ): ProviderResult<NotebookCellStatusBarItem[]> {
+        // Get our message from the cell output and create a basic message
+        const statusMessage = getCellStatusMessageBasedOnFirstCellErrorOutput(cell.outputs);
+        const statusItem = new NotebookCellStatusBarItem(statusMessage, NotebookCellStatusBarAlignment.Left);
+
+        // Save our mapping
+        StatusBarProvider.cellStatusMappings.set(cell, statusItem);
```
roblourens, line 40: Is this only used for test code?

IanMatthewHuff: @roblourens Yes, that was what I added to use instead of the .statusMessage in our tests.

src/client/datascience/notebook/statusBarProvider.ts
DonJayamanne: Awesome, then we don't need to do anything, thanks for the confirmation.
Easy simple powerful api.....

IanMatthewHuff:
Pushing over test failures as we expect these tests to fail. The live version of VS code - insiders that CI uses is out of date from the proposed API that we have here. We'll need to run CI in the morning after VS code has updated.

PR #5553, opened 2021-04-19
Update cell status bar items provider registration selector

rebornix:
Update the cell status bar items provider registration API.


DavidKutu:
Looks good, but a change in the proposed api (executionOrder  being readonly) breaks some stuff.

DonJayamanne (approved):

Issue #5559, opened 2021-04-20
Jupyter specific commands incorrectly show up for all notebook editors


mjbvz:
**Repro**
1. With Jupyter installed 
1. Open a [GitHub issues notebook](https://github.com/microsoft/vscode-github-issue-notebooks)
2. In the command platte, run `Notebook: collapse all notebook cells`

**bug**

<img width="686" alt="Screen Shot 2021-04-19 at 5 46 33 PM" src="https://user-images.githubusercontent.com/12821956/115320677-2e7ab900-a137-11eb-8875-957d9796a774.png">

mjbvz:
I believe you need to use a more specific context key here: https://github.com/microsoft/vscode-jupyter/blob/main/package.json#L1178

It's also confusing that these commands are under the generic `Notebook` command group. It made me think they were coming from VS Code at first

DonJayamanne:
> confusing that these commands are under the generic Notebook command group. It made me think they were coming from VS Code at first

Last we discussed this. We decided to change Jupyter to notebook.
I think it should start as notebook.
Most of the Jupyter users will be using notebook command to create cells, switch kernel,  etc, and those are prefixed `Notebook` having others prefixed `Jupyter` is confusing.

Either it's all consistently Jupyter or notebook, the later is the oblivious choice.
I think hiding the command for other notebooks will address the problem you had.

For users of ipynb (Jupyter) they wouldn't care whether it's a command contributed by vscode or an extension, they just called about notebooks.

joyceerhl:
Validated. The expand/collapse all notebook cells commands are now hidden when a GitHub issues notebook is active.

PR #5574, opened 2021-04-20
Hide jupyter commands from other types of notebooks

DavidKutu:
For #5559

### DIFF
--- a/news/2 Fixes/5559.md
+++ b/news/2 Fixes/5559.md
@@ -1,0 +1,1 @@
+Hide jupyter commands from other types of notebooks.
--- a/package.json
+++ b/package.json
@@ -652,17 +652,17 @@
             {
                 "command": "jupyter.notebookeditor.removeallcells",
                 "title": "%jupyter.command.jupyter.notebookeditor.removeallcells.title%",
-                "category": "Jupyter"
+                "category": "Notebook"
             },
             {
                 "command": "jupyter.notebookeditor.expandallcells",
                 "title": "%jupyter.command.jupyter.notebookeditor.expandallcells.title%",
-                "category": "Jupyter"
+                "category": "Notebook"
             },
             {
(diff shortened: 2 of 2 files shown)


rchiodo (changes requested):
I think we need a when clause instead

rchiodo:
I believe we use 'Notebook' on purpose. Isn't this going to make all of those commands prefixed with 'Jupyter' instead? That seems more confusing to me.

DavidKutu:
@rchiodo I could leave only the changes to the when clause, that's enough to fix the issue.
As of the name `Jupyter` or `Notebook`, both are going to be confusing to a group of people. I don't mind leaving one or the other.

rchiodo (approved):

joyceerhl:
@rchiodo I believe Matt said in #5559 that our using 'Notebook' is confusing, because the category also indicates where the command is coming from. E.g. 'Notebook' means 'VS Code-contributed' while 'Jupyter' means 'Jupyter extension-contributed'. Since this keybinding is contributed by the Jupyter extension I think it makes sense to prefix it with 'Jupyter'. I'm also wondering if Peng/Matt have further thoughts on this.

DonJayamanne:
@joyceerhl I don't think we should change the command group to Jupyter.
I have replied her https://github.com/microsoft/vscode-jupyter/issues/5559#issuecomment-823749451

It's something we discussed as a team (I think we even have an issue for this), I believe it came out of a QP study

DonJayamanne:
@joyceerhl I've replied on the issues.
I don't think we should change the command group to Jupyter.
I recall discussing this with our team, as a result of a QP. Im certain we even have an issue for this (and why we changed from Jupyter to notebook).

DavidKutu:
@joyceerhl that's a good topic to bring up in our next sync meeting. I'd vote to have everything we contribute in the  `Jupyter` group.

Issue #5535, opened 2021-04-16
Too many flake tests failing


IanMatthewHuff:
Almost every single run for the last week or so has failing tests in it.

Kernel starting
Notebook restarting with same kernel
Interrupt 

Seem to be three of the main culprits. Marking as high priority as we can't PR and release confidently in this state.

DonJayamanne:
Hmm, are these flaky or always failing.
These didn't seem to fall last week or earlier.

IanMatthewHuff:
Not sure how far back they started failing (is there a way to see test history? Used to be able to see that in AzDO) but all of these tests were failing at least once in the two runs before the breaking changes.

https://github.com/microsoft/vscode-jupyter/runs/2356143894?check_suite_focus=true
https://github.com/microsoft/vscode-jupyter/runs/2351318190?check_suite_focus=true

I actually don't mind driving this after the kernel push stuff, but it's not good. I just took a look and the last fully green 'Build and Test' run that I see (excluding runs in or targeting the release branch) was eight days ago. Over a week since the last all green PR run.

https://github.com/microsoft/vscode-jupyter/actions/runs/731108365

DonJayamanne:
Oh no not implying that the recent changes broke the tests.
Agreed, we should get them green, end this is how we'll send up back to where we were until early this year, constantly having flaky tests

IanMatthewHuff:
Ohh, I'm not worried about what change caused it (looking now my change might have caused a different new failure). Just mentioning that I'm unsure of when things started to flake, and expressing that I want to get it cleaned up so that we are pushing green. Red gets momentum, something starts to fail, we push anyways, we see the previous failure and push over it again. I do it plenty myself. Gotta get strict with myself.

IanMatthewHuff:
We are still cleaning up a few tests, but I believe the above listed ones are all green. Validated.

Issue #5585, opened 2021-04-21
Fix tests after kernel push update.


IanMatthewHuff:
Tests were broken to unblock kernel push.

joyceerhl:
@IanMatthewHuff were you already working on this? If not I can take this on.

IanMatthewHuff:
@joyceerhl please grab it if you have cycles. I did one cleanup pass here, but it's still pretty bad. https://github.com/microsoft/vscode-jupyter/pull/5589

joyceerhl:
I think we can close this as a dupe of #5535, it's the same set of tests that are failing (apart from ipywidgets, which are legitimately broken).

PR #5602, opened 2021-04-22
Fix/skip flakey tests

DavidKutu:
For #5535

So far 'Failed to set notebook controllers' is still failing (Ian's looking into that).
The functional tests fail because we can't install pytorch.

All skipped tests link to issues and work in real life.


joyceerhl:
FYI @IanMatthewHuff I think #5535 is the same set of test failures (interrupt, restart kernel) that I was originally looking at for https://github.com/microsoft/vscode-jupyter/issues/5585 I'll unassign myself from https://github.com/microsoft/vscode-jupyter/issues/5585 to avoid duplicating work.

src/test/datascience/notebook/contentProvider.vscode.test.ts
```diff
@@ -201,13 +201,21 @@ suite('DataScience - VSCode Notebook - (Open)', function () {
 
         const originalJsonStr = (await fs.readFile(templateIPynb, { encoding: 'utf8' })).trim();
         const originalJson: nbformat.INotebookContent = JSON.parse(originalJsonStr);
-        assert.deepEqual(JSON.parse(model.getContent()), originalJson, 'Untrusted notebook json content is invalid');
+        assert.deepEqual(
+            JSON.parse(model.getContent()).cells,
+            originalJson.cells,
```
DavidKutu, line 206: This was failing because we change the kernel in the metadata. That content is not worth comparing, so lets just compare the cells.

src/client/datascience/interactive-ipynb/nativeEditorCommandListener.ts
```diff
@@ -100,14 +99,18 @@ export class NativeEditorCommandListener implements IDataScienceCommandListener
...
+        if (document) {
+            const target =
+                this.provider.activeEditor?.file.toString() === document.toString()
+                    ? this.provider.activeEditor
+                    : this.provider.editors.find((editor) => editor.file.toString() === document.toString());
+            if (target) {
+                target.interruptKernel().ignoreErrors();
+            }
+        } else {
+            this.provider.activeEditor?.interruptKernel().ignoreErrors();
```
DavidKutu, line 113: We weren't doing anything if the command came from the command palette.

.github/workflows/build-test.yml
```diff
@@ -451,7 +451,7 @@ jobs:
           python ./pythonFiles/install_debugpy.py
           conda env update --file ./build/conda-functional-requirements.yml
           python -m pip install --upgrade -r ./build/conda-functional-requirements.txt
-          conda install pytorch torchvision torchaudio cpuonly -c pytorch
+          conda install pytorch==1.8.0 torchvision==0.9.0 torchaudio==0.8.0 cpuonly -c pytorch
```
DavidKutu, line 454: Functional tests are failing because we can't install pytorch.
This didn't really fix it, but I think its safer to stay in a constant stable version and update it when necessary, instead of risking pytorch coming out with a broken release.

DonJayamanne:
> FYI @IanMatthewHuff I think #5535 is the same set of test failures (interrupt, restart kernel) that I was originally looking at for #5585 I'll unassign myself from #5585 to avoid duplicating work.

I'll look into the interrupt test failures. Since these have been skipped here. I'll pick that up (https://github.com/microsoft/vscode-jupyter/issues/5713).

src/test/datascience/notebook/contentProvider.vscode.test.ts
DonJayamanne: This is not correct, if the user didn't make any changes but just trusted the notebook, then we should not be making changes. 
I.e. just because a notebook is trusted doesn't mean we should modify the file.
I think its a bug.

src/client/common/application/commands.ts
```diff
@@ -84,7 +84,7 @@ export interface ICommandNameArgumentTypeMapping extends ICommandNameWithoutArgu
     ['workbench.action.files.saveAs']: [Uri];
     ['workbench.action.files.save']: [Uri];
     ['undo']: [];
-    [DSCommands.NotebookEditorInterruptKernel]: [NotebookDocument];
+    [DSCommands.NotebookEditorInterruptKernel]: [Uri];
```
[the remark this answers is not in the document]
DonJayamanne: Good pick

DonJayamanne (changes requested):
Blocking on the content provider.
If user doesnt do anything the content should not change, i.e. just because the user opened a file doesn't mean we should modify it.
Eitehr skip that test or we should fix it or file a seprate issue for that.
Either way, its a breaking test, something that used to work.

src/client/common/application/commands.ts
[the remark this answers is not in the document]
joyceerhl: Should this be `NotebookDocument | Uri` then? And the command handler would need to distinguish between the two possible arg types as well.

This must've been what @rchiodo and @IanMatthewHuff were referring to in https://github.com/microsoft/vscode-jupyter/pull/5411#discussion_r607986199 when they asked about the command palette resolving this to a URI. It's not the command palette, it's the notebook toolbar.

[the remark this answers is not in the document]
joyceerhl: Oh never mind, just saw https://github.com/microsoft/vscode-jupyter/pull/5602/files#diff-0e9199b84e8377970515123cf9f3d1c2e60d1a1379507cbade651126ae15934eR195

src/test/datascience/notebook/contentProvider.vscode.test.ts
DavidKutu: When would it fair to change a notebook's metadata? I get not when opening or trusting. How about when a kernel started? If so, that's our default behavior.

DonJayamanne: > w about when a kernel started? If so, that's our default behavior.

Yes that's find, but thiis test doesn't do anything like that, this is just trusting a notebook, hence metadata should not change.

src/test/datascience/notebook/executionService.vscode.test.ts
```diff
@@ -242,7 +242,11 @@ suite('DataScience - VSCode Notebook - (Execution) (slow)', function () {
         );
         assertHasTextOutputInVSCode(displayCell, 'foo', 0, true);
     });
-    test('Clearing output while executing will ensure output is cleared', async () => {
+    test('Clearing output while executing will ensure output is cleared', async function () {
+        // https://github.com/microsoft/vscode-jupyter/issues/5713
+        // The pending cells are always timing out on interrupt, and it might not be up to us
+        // We handle it by asking the user to restart the kernel instead
+        return this.skip();
```
DonJayamanne, line 249: Please remove this, i have fixed the tests

src/test/datascience/notebook/contentProvider.vscode.test.ts
DonJayamanne: I think this is ok David. 
Let's fix this test, i can file a separate issue for what I'm describing

DonJayamanne (approved):
Please remove the skips for the interrupt tests

src/test/datascience/notebook/contentProvider.vscode.test.ts
DavidKutu: Here's the new issue: https://github.com/microsoft/vscode-jupyter/issues/5721

PR #5619, opened 2021-04-23
Remove kernel cache as its not used

DonJayamanne:
We're not reading from the cache, hence removed it (fewer unnecessary disk writes).

### DIFF
--- a/src/client/datascience/kernel-launcher/localKernelFinder.ts
+++ b/src/client/datascience/kernel-launcher/localKernelFinder.ts
@@ -648,5 +648,4 @@ export class LocalKernelFinder implements ILocalKernelFinder {
 
         return kernelSpecFiles;
     }
-
 }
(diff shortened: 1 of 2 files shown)


rchiodo (approved):
How often do we read from the list of kernelspecs? Maybe we should be using the cache?

DonJayamanne:
@rchiodo 
Its read a lot, however we cache the promise https://github.com/microsoft/vscode-jupyter/blob/fad999e807ce74ae415cd27aeb23da9d5abe30d0/src/client/datascience/kernel-launcher/localKernelFinder.ts#L149-L149

Hence for every session, we only ever do this once per workspace.

IanMatthewHuff (approved):

rchiodo:
> @rchiodo
> Its read a lot, however we cache the promise
> 
> https://github.com/microsoft/vscode-jupyter/blob/fad999e807ce74ae415cd27aeb23da9d5abe30d0/src/client/datascience/kernel-launcher/localKernelFinder.ts#L149-L149
> 
> Hence for every session, we only ever do this once per workspace.

Seems like that would be problematic when people add new environments. But I guess we have that issue with interpreters too.

DavidKutu (approved):

DonJayamanne:
>  But I guess we have that issue with interpreters too.

Would be as simple as invalidating this cache when we detect new enviroments from Python extension.

PR #5631, opened 2021-04-26
Remove unnecessary event

DonJayamanne:
General cleanup

### DIFF
--- a/src/client/datascience/ipywidgets/ipyWidgetScriptSource.ts
+++ b/src/client/datascience/ipywidgets/ipyWidgetScriptSource.ts
@@ -36,21 +36,11 @@ export class IPyWidgetScriptSource implements ILocalResourceUriConverter {
     public get postMessage(): Event<{ message: string; payload: any }> {
         return this.postEmitter.event;
     }
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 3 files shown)


src/client/datascience/ipywidgets/ipyWidgetScriptSource.ts
```diff
@@ -143,7 +133,7 @@ export class IPyWidgetScriptSource implements ILocalResourceUriConverter {
         if (!this.uriConversionPromises.has(key)) {
             this.uriConversionPromises.set(key, createDeferred<Uri>());
             // Send a request for the translation.
-            this.postInternalMessageEmitter.fire({
```
rchiodo, line 146: I may be remembering wrong but I believe the internal event didn't go to the webview (just the listeners)? I think this just means the webview will get a message it doesn't care about.

rchiodo (approved):

src/client/datascience/ipywidgets/ipyWidgetScriptSource.ts
DonJayamanne: Based on the code, it didn't make any difference, both were identical, hence removed it.
I.e. absolutely 0 change

Issue #5628, opened 2021-04-26
Interactive window not using Active Interpreter  when running a cell


linette-zyy:
## Environment data
![image](https://user-images.githubusercontent.com/59192491/116025133-5e1f4900-a682-11eb-899b-4c40d153808e.png)

## Steps to reproduce:
1. Open the file manualTestFile.py in VSCode
2. Select conda Interpreter as current environment
3. On the first cell click `Run Cell And All Below`
4. Observes

## Actual behaviour
An unexcepted error message "Python 3.9.2 64-bit requires ipykernel to be installed." pops up.
![image](https://user-images.githubusercontent.com/59192491/116025898-1ef1f780-a684-11eb-8af6-3c319b35a8ce.png)

## Excepted result
Interactive Window should open, show connection information, and execute cells

## Jupyter log
```
Info 2021-04-26 03:42:28: Loading webview. View is notset
Info 2021-04-26 03:42:28: Loading web view...
Info 2021-04-26 03:42:28: Webview panel created.
Info 2021-04-26 03:42:28: Waiting for jupyter server and web panel ...
Info 2021-04-26 03:42:28: Checking for server existence.
Info 2021-04-26 03:42:28: Checking for server usability.
Info 2021-04-26 03:42:28: Starting notebook server.
Info 2021-04-26 03:42:28: Connecting to history server
Info 2021-04-26 03:42:28: Getting kernel specs for history server
Info 2021-04-26 03:42:28: Launching history server
Info 2021-04-26 03:42:28: Starting Notebook
Info 2021-04-26 03:42:28: findPreferredKernel score for Python 3.9.2 64-bit is 1
Info 2021-04-26 03:42:28: findPreferredKernel score for Python 3.8.5 64-bit ('base': conda) is 1
[...]

PR #5632, opened 2021-04-26
Use active interpreter in Interactive Window

DonJayamanne:
For #5628

### DIFF
--- a/src/client/datascience/jupyter/kernels/helpers.ts
+++ b/src/client/datascience/jupyter/kernels/helpers.ts
@@ -34,6 +34,7 @@ import { getTelemetrySafeVersion } from '../../../telemetry/helpers';
 import { IS_CI_SERVER } from '../../../../test/ciConstants';
 import { trackKernelResourceInformation } from '../../telemetry/telemetry';
 import { Uri } from 'vscode';
+import { getResourceType } from '../../common';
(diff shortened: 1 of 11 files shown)


src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -362,8 +375,8 @@ export function findPreferredKernel(
                     !notebookMetadata && // If we don't have metadata, only then should we compare against the interpreter.
                     isKernelRegisteredByUs(spec) === 'newVersion' &&
                     isPythonKernelSpec(spec) &&
-                    interpreter &&
-                    spec.name === getInterpreterKernelSpecName(interpreter)
+                    preferredInterpreter &&
+                    spec.name === getInterpreterKernelSpecName(preferredInterpreter)
                 ) {
                     // This is a perfect match.
                     score += 100;
```
rchiodo, line 382: I believe this path was supposed to do the same thing. How come it doesn't go down this path?

rchiodo: No never mind. Spec.name wouldn't match.

rchiodo (approved):

IanMatthewHuff (approved):

Issue #5633, opened 2021-04-26
Widgets do not get downloaded in CodeSpaces


DonJayamanne:

rchiodo:
Waiting for codespaces/vscode to resolve CORS issue.

joyceerhl:
We should check if abovementioned CORS issue has been fixed.

IanMatthewHuff:
Looks good. Got the third party warning and it worked:
![image](https://user-images.githubusercontent.com/812783/135511615-dae86731-ac7e-48ef-9092-30143b4b4001.png)

PR #5634, opened 2021-04-26
Use `download` package to download widget scripts

DonJayamanne:
For #5633
This might end up fixing other issues we've had with SSH, etc.

I think our old approach doesn't when downloading over HTTPS or the like (no idea, but this worked when I tested in codespaces).


src/test/datascience/ipywidgets/cdnWidgetScriptSourceProvider.unit.test.ts
```diff
@@ -140,7 +134,7 @@ suite('DataScience - ipywidget - CDN', () => {
 
                 assert.deepEqual(value, { moduleName: 'HelloWorld' });
                 // Should not make any http calls.
-                verify(httpClient.exists(anything())).never();
+                // verify(httpClient.exists(anything())).never();
```
DonJayamanne, line 137: Wanted to leave this commented, this way when the tests are fixed we add these checks back.
FYI - The tests are disabled and i dont want to fix the tests now (the'res already an issue for that)

rchiodo (approved):

IanMatthewHuff (approved):

src/test/datascience/ipywidgets/cdnWidgetScriptSourceProvider.unit.test.ts
IanMatthewHuff: Do we need the issue linked here so we can associate the change needed when the issue is fixed?

PR #5635, opened 2021-04-26
More Kernel test fixing

IanMatthewHuff:
For #


src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -201,6 +207,12 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
 
         if (targetController) {
             targetController.updateNotebookAffinity(document, NotebookControllerAffinity.Preferred);
+
+            // When we set the target controller we don't actually get a selected event from our controllers
```
IanMatthewHuff, line 211: @DonJayamanne Can video quick to chat as well. But I don't really love this fix. It's coming up per the convo here with vs code:
https://github.com/microsoft/vscode/issues/122028#issuecomment-825773020
Basics is that when we set our preferredController it is getting changes, but the onNotebookControllerSelected event only happens when the users changes the controller via the UI, the initial set doesn't send this event.

Here I'm basically forcing this message when we updateNotebookAffinity with a target preferred controller. This fixed the bulk of tests for me as we get a message that our kernel is selected now. But I still feel that the right fix would be that VS code sends a message when the selected controller changes regardless of what changes it.

IanMatthewHuff: Since you've looked the new kernel push stuff as well was wondering what your thought was.

DavidKutu (approved):

IanMatthewHuff:
Raw looks much better with this, but seeing issues with local (non-ZMQ) and remote tests. Looking into that now.

src/client/datascience/notebook/notebookControllerManager.ts
DonJayamanne: > But I still feel that the right fix would be that VS code sends a message when the selected controller changes regardless of what changes it.

Agreed, please could you file a big against vscode for this.
When I spoke to @jrieken  about this he mentioned that the event will get fired for this scenario.

DonJayamanne: I'm concerned about this fix and the lack of this fix (work around) as well.
Basically the code doesn't work correctly without this, and having a workaround feels wrong for such a crucial workflow, and I don't want workarounds (specially such be api/native notebooks).
Ie work around could have issues as well.

DonJayamanne (approved):
Good pick

src/client/datascience/notebook/notebookControllerManager.ts
jrieken: Calling `updateNotebookAffinity` does _not_ automatically associate a controller for that notebook. It only means for this notebook the controller is preferred and therefore more likely to be suggested.

DonJayamanne: FYI - We need a way to know what kernel was selected. In our tests we have just one provider, hence if we make this the preferred, I do'nt see why its not getting selected (when it should be) as the controller.

jrieken: > hence if we make this the preferred, I do'nt see why its not getting selected (when it should be) as the controller.

We never know that it is the only controller and it would mean that we always select the very first controller.

IanMatthewHuff: @jrieken I feel like I need more info to understand how this affinity setting is working. Say I have A, B, and C controllers registered on the system. I'm hooked up to all of their onDidChangeNotebookAssociation events in our top level NotebookControllerManager. When I open up a document we hook the document open event (async) and ask our kernel finding code to find the preferred kernel for this document. When it's found, we set it as preferred via updateNotebookAffinity.

This process (after your fix from friday) appears to get us the correct item in the kernel selection. However during this process onDidChangeNotebookAssociation is never hit. I would expect that when a document is initially opened if A, B, or C is associated with it during that initial open that onDidChangNotebookAssociation would fire for that controller. As you mentioned here:

> Calling updateNotebookAffinity does not automatically associate a controller for that notebook. It only means for this notebook the controller is preferred and therefore more likely to be suggested.

Setting affinity isn't a guarantee of selection. So if setting that is not a guarantee then we need to know somehow when something is actually selected for that document and what it is. My workaround here is pretending that the controller that we update affinity for is selected, but I don't believe that would always be the case. But if nothing fires onDidChangeNotebookAssociation when a document is opened and the initial selection is made, how would we know what is actually selected?

Sorry if I'm not explaining that well. Maybe to boil it down more, if I know that A, B, and C are the only notebook controllers registered by any extension on the system period. Then when a document is initially opened then I'm in a state where I'm not actually sure which of those three is selected even if I can suggest one as preferred.

jrieken: Notebook controller selection (what causes the `onDidChangeNotebookAssociation`-event) is a user action. Extensions can use the affinity API to become more prominent in UI for that user action. Users can explicitly select a controller via the picker or implicitly when running with a suggested controller.

> So if setting that is not a guarantee then we need to know somehow when something is actually selected for that document and what it is.

That is the `onDidChangeNotebookAssociation`-event, but again there is no guarantee that this will ever fire for your controller.

> how would we know what is actually selected?

How would we know? Again, it is a user action and for new notebooks or new controllers we simply don't know. We are in the same boat here and it is not that we hold anything back from you. 


Maybe it would help if you explain what you are trying to achieve? Why is it so important for "force" select a specific controller?

IanMatthewHuff: > Maybe it would help if you explain what you are trying to achieve? Why is it so important to "force" select a specific controller?

Sure, so these are the main reasons that we need.

1. ipynbs have kernel metadata in them about what kernel they are expecting to run with. Customers can often have a lot of these options on their system. Having the right kernel selected and ready to go (as opposed to just at the top of a list of options) is important for them to be able to launch right away.
2. For performance we do lots of pre-warming when kernels are selected, but before they actually run. When we don't know the initial selection we can't do this prewarming as we don't know what controller is actually active. If we are aware of the initially selected controller (and it's one of ours) we can start this prewarming. Not having this would be a pretty hefty hit to our first run startup time for notebooks. 
3. Less important as this is test only, but our test infrastructure is based heavily around starting up notebooks (often with different metadata in them) and then waiting to see when the correct kernel is selected, and then executing. Right now a good chunk of our tests are broken (without the hacky fix that I'm considering adding here) as the notebook loads and we now are not sure when that initial controller will get selected, or what that controller is.

IanMatthewHuff:
Squashing in as this improves tests quite a bit. Will file new issue to look at "remote" and "local" tests which still appear to have issues with initial kernel selection.

Issue #5489, opened 2021-04-12
Selected Kernel does not point to the right interpreter


DonJayamanne:
Originally reported by @muser840 
https://github.com/microsoft/vscode-jupyter/issues/5214#issuecomment-817510157

Just to confirm that Jupyter uses the wrong python, which is confirmed by
``` python
import sys
print(sys.executable)
! which python
```

For me the output of the above prints:
```
/usr/bin/python3
/home/user/venv/bin/python
```

It should be the same! These is some mess with switching between python versions by the extension. Somehow switching to old and already deleted environments make it work. That makes extension totally unusable...

The problem is present on versions:
v2021.5.702919634
v2021.6.732501275

update: sometimes it randomly starts to work...

DonJayamanne:
@muser840

Please could you provide the following information:
-   VS Code version: XXX
-   Jupyter Extension version (available under the Extensions sidebar): XXX
-   Python Extension version (available under the Extensions sidebar): XXX
-   OS (Windows | Mac | Linux distro) and version: XXX
-   Python and/or Anaconda version: XXX
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): XXX
-   Attempt to repro the issue & then copy the output from the `Jupyter` output panel.

bardal:
I think I see the same issue:

Version: 1.55.2 (system setup)
Commit: 3c4e3df9e89829dce27b7b5c24508306b151f30d
Date: 2021-04-13T09:35:57.887Z
Electron: 11.3.0
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Windows_NT x64 6.1.7601
Jupyter Extension : v2021.5.745244803
Python 3.8.6 64-bit
Environment: venv

With venv selected - Jupyter Notebook output:
import sys
print(sys.version)
print(sys.executable)
3.8.6 (tags/v3.8.6:db45529, Sep 23 2020, 15:52:53) [MSC v.1927 64 bit (AMD64)]
C:\Program Files\Python38\python.exe

Command line - same venv activated:
(reporting-env) D:\projects\utils>python
Python 3.8.6 (tags/v3.8.6:db45529, Sep 23 2020, 15:52:53) [MSC v.1927 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> print(sys.executable)
D:\projects\reporting-env\Scripts\python.exe

[...]

PR #5644, opened 2021-04-27
Use the right interpreter and kernel with notebooks

DonJayamanne:
For #5489 

This has two fixes
* Ensure we select the right interpreter associated with a kernelspec
* Ensure we pick the right kernelspec based on notebook metadata

### DIFF
--- a/src/client/datascience/kernel-launcher/localKernelFinder.ts
+++ b/src/client/datascience/kernel-launcher/localKernelFinder.ts
@@ -204,9 +204,7 @@ export class LocalKernelFinder implements ILocalKernelFinder {
 
                     // If interpreters were found, remove them from the interpreter list we'll eventually
                     // return as interpreter only items
-                    filteredInterpreters = filteredInterpreters.filter(
-                        (i) => matchingInterpreter === i || this.fs.areLocalPathsSame(matchingInterpreter.path, i.path)
-                    );
+                    filteredInterpreters = filteredInterpreters.filter((i) => matchingInterpreter === i);
 
                     // Return our metadata that uses an interpreter to start
                     return result;
--- a/src/client/datascience/kernel-launcher/localKernelFinder.ts
+++ b/src/client/datascience/kernel-launcher/localKernelFinder.ts
@@ -193,18 +193,20 @@ export class LocalKernelFinder implements ILocalKernelFinder {
             kernelSpecs.map(async (k) => {
                 // Find the interpreter that matches. If we find one, we want to use
                 // this to start the kernel.
-                const matchingInterpreters = this.findMatchingInterpreters(k, interpreters);
-                if (matchingInterpreters && matchingInterpreters.length) {
+                const matchingInterpreter = this.findMatchingInterpreter(k, interpreters);
+                if (matchingInterpreter) {
                     const result: PythonKernelConnectionMetadata = {
                         kind: 'startUsingPythonInterpreter',
                         kernelSpec: k,
-                        interpreter: matchingInterpreters[0],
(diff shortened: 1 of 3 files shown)


src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -276,62 +278,63 @@ export class LocalKernelFinder implements ILocalKernelFinder {
...
-                    `Kernel ${kernelSpec.name} is not python based so does not have an interpreter.`
-                );
-                return false;
-            }
-
-            // 1. Check if current interpreter has the same path
+    ): PythonEnvironment | undefined {
+        // If we know for a fact that the kernel spec is a Non-Python kernel, then return nothing.
+        if (kernelSpec.language && kernelSpec.language !== PYTHON_LANGUAGE) {
+            return;
```
rchiodo, line 287: Why remove the tracing?

rchiodo (approved):

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: I'll add that back.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -193,18 +193,20 @@ export class LocalKernelFinder implements ILocalKernelFinder {
...
-                        id: getKernelId(k, matchingInterpreters[0])
+                        interpreter: matchingInterpreter,
+                        id: getKernelId(k, matchingInterpreter)
                     };
 
                     // If interpreters were found, remove them from the interpreter list we'll eventually
                     // return as interpreter only items
-                    filteredInterpreters = filteredInterpreters.filter((i) => !matchingInterpreters.includes(i));
+                    filteredInterpreters = filteredInterpreters.filter(
+                        (i) => matchingInterpreter === i || this.fs.areLocalPathsSame(matchingInterpreter.path, i.path)
```
IanMatthewHuff, line 208: This local path check doesn't look right to me. We have PythonEnvironments here, right? As far as I know (not the expert here) you can't match those by path. With conda without specifying a specific python multiple conda envs can share one local interpreter path.

IanMatthewHuff (changes requested):
Sorry, maybe I'm off on this (and this looks like it was an existing issue as well) but it looks to me like there is an issue here with using interpreter paths to match, when I believe those can be shared by conda python environments at least.

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: Fixed

IanMatthewHuff (approved):
Approving, though I would mention that it looks like that local path checking might be an issue in the existing code as well? Not fully sure from a quick review, but I think that we might be making bad assumptions for conda here.

DonJayamanne:
> ough I would mention that it looks like that local path checking might be an issue in the 

I removed that change.

IanMatthewHuff:
> > ough I would mention that it looks like that local path checking might be an issue in the
> 
> I removed that change.

@DonJayamanne Sorry, maybe I wasn't clear. I don't mean in your change. I mean the existing code that didn't change seems to also be looking at localpath in a way that could cause issues for conda environments.

DonJayamanne:
@IanMatthewHuff Please could you create an issue for that so we can discuss that, could definitely be a problem for users

IanMatthewHuff:
> @IanMatthewHuff Please could you create an issue for that so we can discuss that, could definitely be a problem for users

Added #5650

PR #5656, opened 2021-04-27
Bump prismjs

joyceerhl:
This just hacks the package-lock.json to use prismjs@1.23.0 until the upstream dependency is fixed. This is temporary and really only necessary in the release branch. Writing a postInstall script doesn't suffice to ensure that the package-lock.json is preserved, because other npm tasks run and modify the package-lock after the postInstall script runs on an npm install. 

This also needs to be ported to release

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24973,14 +24973,14 @@
             "dependencies": {
                 "hastscript": "^5.0.0",
                 "parse-entities": "^1.1.2",
-                "prismjs": "~1.17.0"
+                "prismjs": "1.23.0"
             }
         },
         "node_modules/refractor/node_modules/prismjs": {
-            "version": "1.17.1",
-            "resolved": "https://registry.npmjs.org/prismjs/-/prismjs-1.17.1.tgz",
-            "integrity": "sha512-PrEDJAFdUGbOP6xK/UsfkC5ghJsPJviKgnQOoxaDbBjwc8op68Quupwt1DeAFoG8GImPhiKXAvvsH7wDSLsu1Q==",
-            "optionalDependencies": {
+            "version": "1.23.0",
+            "resolved": "https://registry.npmjs.org/prismjs/-/prismjs-1.23.0.tgz",
+            "integrity": "sha512-c29LVsqOaLbBHuIbsTxaKENh1N2EQBOHaWv7gkHN4dgRbxSREqDnDbtFJYdpPauS4YCplMSNCABQ6Eeor69bAA==",
+            "requires": {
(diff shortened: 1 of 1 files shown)


rchiodo:
The package-lock is really far off from what I expected. How did you generate it?

rchiodo:
How will we redo this later? We need something after postinstall I think then.

joyceerhl:
@rchiodo by running npm install. The package-lock changed significantly in https://github.com/microsoft/vscode-jupyter/commit/e00adb7e434693338ed0511ed171957e7bb625fb. I think @DonJayamanne and I may have different npm versions or something? The package-locks that we generate have different versions.

rchiodo:
> @rchiodo by running npm install. The package-lock changed significantly in [e00adb7](https://github.com/microsoft/vscode-jupyter/commit/e00adb7e434693338ed0511ed171957e7bb625fb). I think @DonJayamanne and I may have different npm versions or something? The package-locks that we generate have different versions.

By running npm install without any specific changes? So you mean to fix this bug we just had to run npm install? The package lock looks like a whole mess of stuff went backwards.

Example:

image
(left side is your changes)

IanMatthewHuff:
Yeah, it is a bit scary how much the lock file is changing. Question on this. Should we do this just in the release branch? If that's the only place that it's needed might be better?

joyceerhl:
@rchiodo sorry, no, I mean that I'm not sure the package-lock should look the way it does in main. Don's PR added a lockfile with "lockfileVersion": 2 and mine is still "lockfileVersion": 1. 

@DonJayamanne what version of npm are you using? I'm on npm 6.13. npm 7+ results in v2 lockfiles. Maybe the version doesn't matter all that much?

rchiodo:
That seems better but would still like something to enforce it? Maybe a CI task (if we can't get postinstall to fix it)

joyceerhl:
Sorry, forgot to change the base to release-2021.5. I'm also submitting a PR to nteract to bump their dependency on react-syntax-highlighter (which pulls prismjs in). If getting that accepted / released takes too long I'll work on a CI task for main. Does that sound OK @rchiodo?

rchiodo (approved):

joyceerhl:
Maybe a pre-commit hook would be better for this? I couldn't find an explanation of the difference between npm i and npm ci (besides the obvious part of the latter never writing to package-lock). But npm i runs something else after our postInstall script which undoes any changes I could add in the postInstall step. It looks like it might be audit or something?

In any case this package-lock hack is temporary until we can bump the dependency.

PR #5691, opened 2021-04-29
Merge release back into main

joyceerhl:


package.json
```diff
@@ -24,7 +24,7 @@
         "theme": "light"
     },
     "engines": {
-        "vscode": "1.56.0-insider"
+        "vscode": "1.56.0"
```
IanMatthewHuff, line 27: Can't fully remember our process on this engine number. Don't we keep it as -insider in main?

rchiodo: Yeah should be 1.57.0-insider

joyceerhl: I think we discussed this at bug bash and said we don't need to bump this to 1.57.0-insider right away, as that will block publishing nightly insiders builds until 1.57.0-insider becomes available

joyceerhl: But yes good point, closing this PR since I can't fix this directly in the release branch (which is the source of this PR)

PR #5700, opened 2021-04-29
Merge release back into main

joyceerhl:
For #5398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,6 @@
 
 1. Manage contributed Jupyter kernels registration.
    ([#4490](https://github.com/Microsoft/vscode-jupyter/issues/4490))
-1. Hide the Jupyter Server status bar item when a third party extension kernel is used.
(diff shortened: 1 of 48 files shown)


DonJayamanne (approved):

joyceerhl:
@rchiodo @IanMatthewHuff do we want to bump to 1.57.0-insider? This means no insider builds will go out until 1.57-insider becomes available.

DonJayamanne:
> ant to bump to 1.57.0-insider? This means no insider builds will go out until 1.57-insider becomes available.

Why not leave it as is? I.e why do we have to bump this?

DonJayamanne:
Lets discuss this in triage. Even I'm not sure what needs to be done here (we need to ensure users don't get this build in stable(

joyceerhl:
@DonJayamanne pointed out that if we bump main to 1.57-insider now, we're going to have to patch the package.json locally in order to develop against main bits. I forgot about that.

PR #5714, opened 2021-04-30
Better telemetry for languages

DonJayamanne:


DavidKutu (approved):

src/client/datascience/constants.ts
```diff
@@ -26,6 +26,10 @@ export const KnownKernelLanguageAliases = new Map<string, string>([
     ['c++12', 'c++'],
     ['c++14', 'c++']
 ]);
+/**
+ * This will get updated with the list of VS Code languages.
+ * This way, we can send those via telemetry, instead of having to hardcode the languages.
+ */
```
IanMatthewHuff, line 32: The comment here helps, but this seems a bit strange. It's a const array in our datascience/constants file. I don't think that I would expect for anything in this file to be changing dynamically.

IanMatthewHuff: I guess I can't really think of a better location for this off hand. So approving. Just doesn't seem to fit with the rest of this class.

IanMatthewHuff (approved):

src/client/datascience/constants.ts
DonJayamanne: Didn't want to change it a `let xyz`, then it means `xyz` can be updated.

IanMatthewHuff: Yeah, I get that we can add or remove items from it and that it should not be a let. Just more meant that nothing else in this file is getting dynamically updated. If I was using an array from this file I'd be surprised if it was dynamically changed at runtime. I think it works though.

DonJayamanne: Happy to change it to something else, or move it out of the `constants.ts` file if that works.

Issue #4758, opened 2021-02-11
Install kernel dependencies with the -U flag


DonJayamanne:
When install ipykernel, etc we should update the depednencies.
We have had plenty of cases where users had it installed, but versions were incorrect.
They had to uninstall & re-install.

For instance I had an issue with ipykernel in a conda environment.
Had an old version of ipython, hence when installing ipykernel, an old compatible version got installed. However that version of ipython is too old, similarly its possible the versio of ipykernel is also old.

The solution was `conda install ipykernel --update-deps --force-reinstall`
We might want to do the same for `pip install ipykernel -U --force-reinstall`

# How to test/validate
* Install latest dev version of Python extension with the changes (here's the PR in Python ext https://github.com/microsoft/vscode-python/pull/16133)
* Use conda & non conda
* Get notebooks working
* Uninstall pyzmq (a dependency), 
* Now ipykernel is installed, but it doesn't work, hence re-installing ipykernel will not work
* Try now, and it should force re-install

DonJayamanne:
We might also want to consider makign it possible to pass in the argument `--force-reinstall`.
We found at least in 2 occasions that this works.
hence we might have scenarios where we have -U as well as `--force-reinstall`.

TLDR - ensure we can pass more than just the one argument.

DonJayamanne:
This works, see here https://github.com/microsoft/vscode-jupyter/issues/5689  (user tried this option to get unblocked recently)

DonJayamanne:
We should probably use `--update-deps --force-reinstall` only when first install of `ipykernel` fails.
Probably worth discussing with team.

DonJayamanne:
Discussed solution:
* if ipykernel is available (pip list), then pass flags to Python to force update & force re-install

joyceerhl:
I do get a prompt for conda and non conda environments, but for conda the install command fails.

Conda:
![image](https://user-images.githubusercontent.com/30305945/118434606-307e5c00-b692-11eb-9846-0a9d70fc638a.png)

Conda install errors out:
![image](https://user-images.githubusercontent.com/30305945/118434733-6ae7f900-b692-11eb-8da4-4e67847d8e17.png)

Non-conda (pip):
![image](https://user-images.githubusercontent.com/30305945/118434881-abe00d80-b692-11eb-9f3f-cfbce1bbed84.png)

DavidKutu:
validated, in both the jupyter and python extensions

PR #5738, opened 2021-05-03
Update/reinstall if module was installed or exists

DonJayamanne:
For #4758

When re-installing for the second time ask user to re-install/update the package.
Requires a corresponding change in Python extension (that will be done seprately - separate repo, hence separate PR)

When ipykernel is detected (already found but we cannot import it), or if we have previously installed this, then the following error message is displayed;
Screen Shot 2021-05-03 at 16 34 04


DonJayamanne:
@IanMatthewHuff thanks, i now know not to ignore the failures.

DavidKutu (approved):

src/client/common/installer/productInstaller.ts
```diff
@@ -1,34 +1,74 @@
...
+    }
+    const packageName = translateProductToModule(product);
+    const packageVersionPromise = InterpreterPackages.getPackageVersion(interpreter, packageName)
+        .then((version) => (typeof version === 'string' ? 'found' : 'notfound'))
+        .catch((ex) => traceError('Failed to get interpreter package version', ex));
+    try {
+        // Dont wait for too long we don't want to delay installation prompt.
+        const version = await Promise.race([sleep(500), packageVersionPromise]);
+        if (typeof version === 'string') {
+            return version === 'found';
```
IanMatthewHuff, line 65: Do you want to add it to the memento here since we found it?

DonJayamanne: Wasn't sure if it was a good idea.
What if user uninst, re-created the environment, and testing for us also gets more messy.

IanMatthewHuff (approved):

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -283,28 +290,40 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+            // Hook up to if this NotebookController is selected or de-selected
+            controller.onNotebookControllerSelected(this.handleOnNotebookControllerSelected, this, this.disposables);
+
+            // We are disposing as documents are closed, but do this as well
+            this.disposables.push(controller);
+
+            return controller;
+        } catch (ex) {
+            // We know that this fails when we have xeus kernels installed (untill that's resolved thats one instance when we can have duplicates).
+            sendTelemetryEvent(
```
DonJayamanne, line 318: If you install the xeus kernel, then everythings falls over. You cannot run any notebook, because we end up with duplicate kernels in our code.
This ensures we never end up in a situtaion where we're completely broken because of one bad apple.
Have a separate issue for handling xeus python kernels.

@IanMatthewHuff @rchiodo

rchiodo (approved):

src/client/common/installer/productInstaller.ts
```diff
@@ -66,9 +114,12 @@ export abstract class BaseInstaller {
...
         product: Product,
         resource?: InterpreterUri,
-        cancel?: CancellationToken
+        cancel?: CancellationToken,
+        reInstallAndUpdate?: boolean
     ): Promise<InstallerResponse> {
-        return this.serviceContainer.get<IPythonInstaller>(IPythonInstaller).install(product, resource, cancel);
+        return this.serviceContainer
+            .get<IPythonInstaller>(IPythonInstaller)
+            .install(product, resource, cancel, reInstallAndUpdate);
```
joyceerhl, line 122: I think this flag needs to be added to vscode-python too, because `IPythonInstaller` wraps the contributed installer from the Python extension?

joyceerhl: Ah yes you mentioned that in the issue description

joyceerhl (approved):

src/client/datascience/notebook/notebookControllerManager.ts
IanMatthewHuff: Good catch @DonJayamanne I should install that kernel as well just to have for testing.

PR #5747, opened 2021-05-04
Hide the run above/below buttons in native notebook cells

DonJayamanne:
Hiding in release (not main)
Disabling auto start in release (not main)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,6 @@
 
 ### Fixes
 
-1. Move the run cells above and below buttons from the notebook toolbar to the cell toolbar.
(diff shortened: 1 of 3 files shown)


joyceerhl (approved):
We may also want to remove the changelog entry about `1. Move the run cells above and below buttons from the notebook toolbar to the cell toolbar. ([#4644](https://github.com/Microsoft/vscode-jupyter/issues/4644))`

DavidKutu (approved):

IanMatthewHuff (approved):

joyceerhl:
@DonJayamanne are we not disabling autostart anymore?

DonJayamanne:
>  are we not disabling autostart anymore?

I thought the tests were failing because of that, then found that that was not the case.

PR #5754, opened 2021-05-04
Better comparisons of the interpreter paths when starting kernels

DonJayamanne:


src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -497,11 +499,31 @@ export async function sendTelemetryForPythonKernelExecutable(
...
+            sendTelemetryEvent(Telemetry.PythonKerneExecutableMatches, undefined, {
+                match: match ? 'true' : 'false',
+                kernelConnectionType: kernelConnection.kind
+            });
+            trackKernelResourceInformation(Uri.file(file), { interpreterMatchesKernel: match });
+            return;
+        }
+
+        // The interpreter paths don't match, possible we have a synlink or similar.
+        // Lets try to get the path from the interpreter using the exact same code we send to the kernel.
```
DonJayamanne, line 512: Comments added why this might be better.

rchiodo (approved):

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -497,11 +499,31 @@ export async function sendTelemetryForPythonKernelExecutable(
         }
         const sysExecutable = concatMultilineString(output.text).trim().toLowerCase();
         const match = kernelConnection.interpreter.path.toLowerCase() === sysExecutable;
-        sendTelemetryEvent(Telemetry.PythonKerneExecutableMatches, undefined, {
-            match: match ? 'true' : 'false',
-            kernelConnectionType: kernelConnection.kind
+        if (match) {
```
DonJayamanne, line 502: With the previous kernel refactor we seemed to have regressed a in one case where we weren't starting the right interpreter.
That was fixed & will go out in the next release.

However its possible there are other scenarios when this still doesn't work.
Or the telemetry is wrong, i.e. algorithm used to compare interpreters is incorrect.
Improved the code check the paths.

DavidKutu (approved):

IanMatthewHuff (approved):

Issue #5753, opened 2021-05-04
Breaking: Update notebook messaging APIs in webview


IanMatthewHuff:
Update Jupyter extension for:
https://github.com/microsoft/vscode/issues/122944
https://github.com/microsoft/vscode/commit/d6393457912d2a9cb1df6f9067499ebff828e2bf

joyceerhl:
Validated as ipywidgets etc work in native notebooks in Insiders.

PR #5775, opened 2021-05-05
Update Jupyter to new notebook preload message api

IanMatthewHuff:
For #5753 


src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -160,32 +160,27 @@ export class VSCodeNotebookController implements Disposable {
...
-                                  this.context.extensionPath,
-                                  'out',
-                                  'datascience-ui',
-                                  'ipywidgetsKernel',
-                                  'require.js'
-                              )
-                          )
-                      }
-                  ]
+                    new NotebookKernelPreload(Uri.file(
```
IanMatthewHuff, line 163: Minor change from new proposed. Needed to use the new class here.

src/datascience-ui/ipywidgets/renderer/index.ts
```diff
@@ -30,4 +29,4 @@ function initialize(api: NotebookRendererApi<any>) {
     api.onWillDestroyOutput(disposeOutput);
 }
 
-initialize(acquireNotebookRendererApi(JupyterIPyWidgetNotebookRenderer));
+initialize(acquireNotebookRendererApi());
```
IanMatthewHuff, line 32: Another change from proposed, correct name is automatically passed.

src/datascience-ui/react-common/postOffice.ts
```diff
@@ -74,38 +84,70 @@ export class PostOffice implements IDisposable {
     }
 
     public acquireApi(): IVsCodeApi | undefined {
```
rchiodo, line 86: Seems like this could be split up more? Or generalized into two different functions instead of multiple if points? I know some of it might be duped in each function, but I believe there's two distinct paths here?

- postKerneMessage = native case
- acquireVsCodeApi = webview case

rchiodo: I'm suggesting:

- acquireApi (highest level)
- acquireNativeApi (for native case when new function - isNative - returns true). This function would do the work for bnding for the native case.
- acquireWebViewApi - This function would do the code we used to have?

joyceerhl (commented):
Apologies if you already did this, but I believe you may need to update to npm7 to generate a v2 package-lock if you haven't already done so.

src/datascience-ui/react-common/postOffice.ts
IanMatthewHuff: Were you thinking like just break this function up into the kernel / vs code specific functions? Or more like factor our a full common interface. Full interface feels kinda heavy, but it might read better.

rchiodo: I think two different functions would be good enough and make it clearer which path you were on.

DavidKutu (approved):

src/datascience-ui/react-common/postOffice.ts
IanMatthewHuff: I was actually starting down the path of an interface / class refactor. What do you think about this? @rchiodo

rchiodo (approved):

DonJayamanne (approved):

IanMatthewHuff:
> Apologies if you already did this, but I believe you may need to update to npm7 to generate a v2 package-lock if you haven't already done so.

Thanks, updated to latest 7 version and regenerated.

PR #5777, opened 2021-05-05
Update extension license

greazer:


rchiodo:
I believe this means we have to change the package.json to point to an MIT license too. Otherwise the extension will still ship with the marketplace license? Maybe you still want that.

DavidKutu:
It should be ok if we still have the license file. I'd ask CELA. Otherwise, somehow mention both MIT and 'look at license.txt' on the package.json

DavidKutu (approved):

PR #5781, opened 2021-05-05
Changes to cell execution task API

DonJayamanne:


joyceerhl (approved):

DavidKutu (approved):

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -118,7 +127,8 @@ export class CellExecution {
         private readonly applicationService: IApplicationShell,
         private readonly kernelConnection: Readonly<KernelConnectionMetadata>,
         disposables: IDisposableRegistry,
-        extensionContext: IExtensionContext
+        extensionContext: IExtensionContext,
+        private readonly controller: NotebookController
```
IanMatthewHuff, line 131: It think this works fine, just wondering why not just look it up off the manager from the cell uri?

IanMatthewHuff: Actually never mind, that's a bad idea. That could change, and the execution is specific to a given connection. Ignore my comment :).

IanMatthewHuff (approved):

src/client/datascience/jupyter/kernels/cellExecution.ts
DonJayamanne: > execution is specific to a given connection. Ignore my comment :).

Yes. Initially I did what you suggested. But each kernel/execution is tied to a specific connection (hence specific controller). 

It was much easier to just pass this, rather than passing the manager all around and then getting the controller (also ended up ina cyclic reference issues)

Issue #5798, opened 2021-05-06
Modal dialog for ipykernel installs & better messages


DonJayamanne:
When running a notebook without ipykernel the user will be presented with the same message today, but modal diaglog as follows:
<img width="447" alt="Screen Shot 2021-05-06 at 16 12 24" src="https://user-images.githubusercontent.com/1948812/117376441-fb9a2a00-ae85-11eb-8794-ad27562c81f5.png">

Should probably change the message, doesn't seem to be descriptive enought. Feels way too technical.
Someone starting jupyter notebooks or not aware of ipykernel will just ignore this.

It should ideally be along the lines:
`In order to run notebooks IPykernel needs to be installed, do you want to install ipykernel?`

rchiodo:
I still think it should reference the kernel though. And if it's too long, nobody will read it.

claudiaregio:
Going to keep it very simple

"IPykernel REQUIRED for this action" 
Options: Cancel, Install

rchiodo:
Yeah but that doesn't tell me where it's going, so I don't know if I want to answer that or not.

If it's too simple then I'm more afraid.

DonJayamanne:
> "IPykernel REQUIRED for this action"

Isn't the assumption that users know what `IPYKernel` is, or does it not matter.

DonJayamanne:
> t should reference the kernel though. And if it

Does this matter, i would have thought they'd need to know that its tied to their current operation (running cell in notebook/interaciev window).

I doubt if anyone looks at the kernel information **(though thats very useful for diagnostic purposes)** - because we fail to install in the right one.

rchiodo:
At least the kernel name is something they can see elsewhere. And ipykernel can be looked up on google to find out what it is.

If we just say

'We need to install something - Yes or No'

Who is going to say Yes? My first question is what are you installing and second, where?

DonJayamanne:
Agreed, tha'ts why my suggestion 
` order to run notebooks IPykernel needs to be installed, do you want to install ipykernel?`
But could be long

claudiaregio:
I do not think most users will know or really care too much what it is. The important message that needs to be given to users is "you are blocked if you do not install this". 

We can keep the name so if users are curious they can do research to see what it is. I do not think this is the located where we need to explain to users in depth what is happening and why, it will be too long and they will not read it. 

[...]

PR #5799, opened 2021-05-06
Modal dialog for prompt to install ipykernel

DonJayamanne:
For #5798


IanMatthewHuff (changes requested):
Look like the linked issue says it should be behind an experiment?

DonJayamanne:
> ook like the linked issue says it should be behind an experiment?

That was my initial plan, but I don't think we need to do that.
Based on @claudiaregio some users are not seeing this message (validated in a QP)

claudiaregio:
I would like to create an experiment just so we can track improvement from our other version (show how many people are taking an action on the exp vs. how many took it on the old one)

Issue #5806, opened 2021-05-07
Support both old and new `workbench.editorAssociations` setting formats


joyceerhl:
Old layout will not get IntelliSense in settings.json anymore. It's now an object instead of an array. However we can't assume that we are loaded in a version of VS Code with the right setting format, so we should support both.

kkukshtel:
Just peeking in here as a fellow extension author - what's the new workbench.editorAssociations format? My extension also broke on 1.56 and I can't find any documentation on why the following no longer works:

```
{
    "workbench.editorAssociations": [
    {
        "viewType": "depot.data",
        "filenamePattern": "*.cmod"
    }]
}
```

joyceerhl:
@kkukshtel The new format is:

```
    "workbench.editorAssociations": {
        "*.cmod": "depot.data"
    }
```

i.e. the filenamePattern is the key and the viewType is the value.

That being said, the workbench.editorAssociations setting change should only be in VS Code 1.57.0-insider. VS Code stable should still be able to work with the old format. VS Code Insider is also supposed to be able to understand the old format and the new format--the settings.json IntelliSense will just flag the old setting as an error, but it's totally benign for now. I've also been told that tomorrow's VS Code Insider build will start automatically migrating users to the new format.

kkukshtel:
Gotcha, thank you! And yeah I changed the setting in 1.56 and it seemed to break the editor association settings even more so tha change definitely didn't accidentally sneak into the release.

I'm trying to track down the causes of this https://github.com/microsoft/vscode/issues/123617 and thought it may have been related to what you posted here but seems like not.

DavidKutu:
validated

Issue #5804, opened 2021-05-07
Unable to open Notebooks in Insiders


sztal:
Issue Type: <b>Bug</b>

It seems that the last update crashed Jupyter notebook extension.

It cannot find `jupyter.opennotebook` function/object presumably need for opening `.ipynb` files in the Notebook Editor. Moreover, there seems to be some even deeper problem going in the core of the extension which probably cannot load some sort of a configuration file (or so it seems according to the console log in Developer Tools. Below I paste the javascript erorr trace generated in my  particular case and raised at `workbench.desktop.main.js:62`:

```
[Extension Host] extension activation failed TypeError: o.getConfiguration(...).get(...).find is not a function
    at new m (/home/sztal/.vscode-insiders/extensions/ms-toolsai.jupyter-2021.6.819764235/out/client/extension.js:52:683169)
    at Object.t.resolveInstance (/home/sztal/.vscode-insiders/extensions/ms-toolsai.jupyter-2021.6.819764235/out/client/extension.js:52:263270)
    at /home/sztal/.vscode-insiders/extensions/ms-toolsai.jupyter-2021.6.819764235/out/client/extension.js:52:262475
    at Object.t.resolve (/home/sztal/.vscode-insiders/extensions/ms-toolsai.jupyter-2021.6.819764235/out/client/extension.js:52:262738)
    at /home/sztal/.vscode-insiders/extensions/ms-toolsai.jupyter-2021.6.819764235/out/client/extension.js:52:253835
    at e._get (/home/sztal/.vscode-insiders/extensions/ms-toolsai.jupyter-2021.6.819764235/out/client/extension.js:52:253601)
    at e.get (/home/sztal/.vscode-insiders/extensions/ms-toolsai.jupyter-2021.6.819764235/out/client/extension.js:52:252313)
    at a.get (/home/sztal/.vscode-insiders/extensions/ms-toolsai.jupyter-2021.6.819764235/out/client/extension.js:90:509426)
    at /home/sztal/.vscode-insiders/extensions/ms-toolsai.jupyter-2021.6.819764235/out/client/extension.js:52:245341
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at /home/sztal/.vscode-insiders/extensions/ms-toolsai.jupyter-2021.6.819764235/out/client/extension.js:37:825686
[...]

PR #5810, opened 2021-05-07
Migrate editorAssociations setting to new format

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/5804, https://github.com/microsoft/vscode-jupyter/issues/5806

### DIFF
--- a/src/client/common/application/customEditorService.ts
+++ b/src/client/common/application/customEditorService.ts
@@ -5,6 +5,10 @@ import { inject, injectable } from 'inversify';
 import * as vscode from 'vscode';
 import { IExtensionSingleActivationService } from '../../activation/types';
 import { Telemetry } from '../../datascience/constants';
+import {
+    NewEditorAssociationSetting,
+    ensureUpdatedEditorAssociationSettingFormat
+} from '../../datascience/notebook/integration';
 import { sendTelemetryEvent } from '../../telemetry';
 
 import { UseCustomEditorApi } from '../constants';
@@ -48,32 +52,22 @@ export class CustomEditorService implements ICustomEditorService, IExtensionSing
         let updateType: 'added' | 'removed' | undefined;
         try {
             const settings = this.workspace.getConfiguration('workbench', undefined);
-            const editorAssociations = settings.get('editorAssociations') as {
-                viewType: string;
-                filenamePattern: string;
-            }[];
+            const editorAssociations = settings.get('editorAssociations'); // At this point we don't know if this is the old or new format
+            const updatedSettings = ensureUpdatedEditorAssociationSettingFormat(
+                editorAssociations
+            ) as NewEditorAssociationSetting;
 
             // Update the settings.
-            if (
-                this.useCustomEditorApi &&
-                (editorAssociations.length === 0 || !editorAssociations.find((item) => item.viewType === ViewType))
-            ) {
-                editorAssociations.push({
-                    viewType: ViewType,
-                    filenamePattern: '*.ipynb'
-                });
+            if (this.useCustomEditorApi && updatedSettings['*.ipynb'] !== ViewType) {
+                updatedSettings['*.ipynb'] = ViewType;
                 updateType = 'added';
-                await settings.update('editorAssociations', editorAssociations, vscode.ConfigurationTarget.Global);
+                await settings.update('editorAssociations', updatedSettings, vscode.ConfigurationTarget.Global);
                 sendTelemetryEvent(Telemetry.UpdateCustomEditorAssociation, undefined, { type: 'added' });
             }
 
             // Revert the settings.
-            if (
-                !this.useCustomEditorApi &&
-                Array.isArray(editorAssociations) &&
-                editorAssociations.find((item) => item.viewType === ViewType)
-            ) {
-                const updatedSettings = editorAssociations.filter((item) => item.viewType !== ViewType);
+            if (!this.useCustomEditorApi && updatedSettings['*.ipynb'] === ViewType) {
+                delete updatedSettings['*ipynb'];
                 updateType = 'removed';
                 await settings.update('editorAssociations', updatedSettings, vscode.ConfigurationTarget.Global);
                 sendTelemetryEvent(Telemetry.UpdateCustomEditorAssociation, undefined, { type: 'removed' });
--- a/src/client/datascience/notebook/integration.ts
+++ b/src/client/datascience/notebook/integration.ts
@@ -125,39 +125,48 @@ export class NotebookIntegration implements IExtensionSingleActivationService {
         await settings.update('cellToolbarLocation', toolbarSettings, ConfigurationTarget.Global);
     }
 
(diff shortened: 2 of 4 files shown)


joyceerhl:
Just realized the custom editor viewType might get parsed as an object because it's '.' delimited. Checking that now

joyceerhl:
Actually I think we may have to support both settings, i.e. we can't do a migrate, since we don't know which version of the setting the current build of VS Code we're installed in supports.

src/client/common/application/customEditorService.ts
```diff
@@ -48,35 +49,54 @@ export class CustomEditorService implements ICustomEditorService, IExtensionSing
...
+                    const updatedSettings = [
+                        ...editorAssociations,
+                        {
+                            viewType: ViewType,
+                            filenamePattern: '*.ipynb'
+                        }
+                    ];
+                    updateType = 'added';
+                    await settings.update('editorAssociations', updatedSettings, vscode.ConfigurationTarget.Global);
+                    sendTelemetryEvent(Telemetry.UpdateCustomEditorAssociation, undefined, { type: 'added' });
```
greazer, line 70: Why not just use updateType here?

joyceerhl: No reason other than this is how the code was originally written 😁

src/client/common/application/customEditorService.ts
```diff
@@ -48,35 +49,54 @@ export class CustomEditorService implements ICustomEditorService, IExtensionSing
...
+                }
+
+                // Revert the settings.
+                if (!this.useCustomEditorApi && hasCustomEditorAssociation) {
+                    const updatedSettings = editorAssociations.filter((item) => item.viewType !== ViewType);
+                    updateType = 'removed';
+                    await settings.update('editorAssociations', updatedSettings, vscode.ConfigurationTarget.Global);
```
greazer, line 77: Seems like these two lines "await settings.update(...)" and "sendTelemetryEvent(Telemetry...)" could be pulled out and called just once throughout this whole function. Not a huge deal though.

src/client/common/application/customEditorService.ts
greazer: So the only reason this variable exists is to have something to pass into the catch block? Seems like a waste of a good variable. :D

src/client/common/application/customEditorService.ts
joyceerhl: I don't believe we want to unconditionally send the telemetry event or update the settings object outside of these four cases. For example, if we should be using custom editor and the association is already present, the settings object hasn't changed so shouldn't be updated, and the telemetry event represents a settings update and no update has taken place so shouldn't be sent.

src/client/datascience/notebook/integration.ts
```diff
@@ -125,39 +125,60 @@ export class NotebookIntegration implements IExtensionSingleActivationService {
...
-                editorAssociations.length === 0 ||
-                !editorAssociations.find((item) => isJupyterNotebook(item.viewType)))
-        ) {
-            editorAssociations.push({
-                viewType: 'jupyter-notebook',
-                filenamePattern: '*.ipynb'
-            });
-            await settings.update('editorAssociations', editorAssociations, ConfigurationTarget.Global);
-        }
+        if (Array.isArray(editorAssociations)) {
```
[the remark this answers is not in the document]
joyceerhl: The logic does differ in the following ways:
1. Here we're checking for the native notebooks viewtype and the native notebooks experiment, while in the customEditorService we're checking for the custom editor viewtype and custom editor experiment, so they're almost the opposite of each other.
2. In the customEditorService we send a custom editor-specific telemetry event with custom properties based on which case we fall into.

It did bother me that we have the same logic structure in two different places, but I didn't attempt to refactor the logic into a shared function when I put up the original PR because the custom editor code is temporary and I thought it might hurt readability by stripping out context. But I've added 076bdb7b6ad676a42629953d4231c9217dd73159 to this PR which pulls most of the shared logic into a common function, `updateEditorAssociationSettings`. I'm going to run through my 'test plan' in this PR description in the morning, but I'd also appreciate feedback on whether this change impacts readability.

src/client/common/application/customEditorService.ts
joyceerhl: I believe this is resolved by 076bdb7b6ad676a42629953d4231c9217dd73159.

src/client/datascience/notebook/integration.ts
```diff
@@ -125,39 +131,92 @@ export class NotebookIntegration implements IExtensionSingleActivationService {
...
+            noop
+        );
+    }
+
+    private async disableNotebooks() {
+        await this.enableDisableEditorAssociation(false);
+    }
+}
+
+export type NewEditorAssociationSetting = { [glob: string]: string | undefined };
```
DonJayamanne, line 151:
```suggestion
export type NewEditorAssociationSetting = Record<string, string>;
```

DonJayamanne: I don't see how the value can be undefined.

joyceerhl:
@lramos15 VS Code Stable 1.56.0 does not recognize the new format, but I imagine that if VS Code does a recovery build, 1.56.1 will support the new format. Can we guarantee that your latest change to convert the settings will go out together with any stable build that contains support for the new format? If so that significantly simplifies things for us.

lramos15:
@joyceerhl Yes we can guarantee that the stable builds that ship with the new format also ship with the conversion logic and eventually (3-4 releases down the line), we will remove the old setting completely as anyone who has launched VS Code since that release will have the new format. The release which will support the new format is 1.57. I'm debating whether or not to roll back the setting change until right before the 1.57 release as I understand that you are currently straddling the insiders / stable line with stable being old format and insiders being the new format.

joyceerhl:
@lramos15 Rolling back won't be necessary, we have a release branch which targets the stable build and our main branch targets insiders. If you keep your change in Insiders, tomorrow's Insiders build will support the new format and migrate users to the new format. We will also publish an Insiders-compatible build in our automated nightly pipeline that supports the new format. This should unblock new Jupyter users in Insiders (https://github.com/microsoft/vscode-jupyter/issues/5804).

joyceerhl:
JFYI @lramos15 we just published a VS Code Insiders-compatible build of the Jupyter extension containing the changes from this PR which should unblock our customers today (if the setting is the old format, we convert it to the new one, thereafter we deal with the new format). Thanks for following up with us, we'll also test with tomorrow's build.

lramos15:
@joyceerhl Sorry for not updating here, it appears we had some failing tests which prevented yesterdays insiders from going out (the one that you see released yesterday was just security fixes). It appears the team is working on a build right now and the test issue has been resolved.

joyceerhl:
@lramos15 I can confirm your change works in the latest Insiders and migrates the editorAssociation setting from the old to the new format😊 Thanks again for following up!

PR #5834, opened 2021-05-11
Update npm dependencies

DonJayamanne:


src/test/datascience/extensionapi/exampleextension/ms-ai-tools-test/package-lock.json
```diff
@@ -1,4657 +0,0 @@
-{
```
DonJayamanne, line 1: Easier to remove than maintianing this

gulpfile.js
```diff
@@ -52,6 +52,50 @@ gulp.task('checkNativeDependencies', (done) => {
     }
     done();
 });
+gulp.task('checkNpmDependencies', (done) => {
```
rchiodo, line 55: What's this for? Seems odd. Why do we need to check these? Did you force change the package-lock again?

DonJayamanne: Yes, and wrote a script to ensure it doesn't regress.
Else we need to update a tonne of dependencies.
I'd rather create a seprate issue to update all packages to latest versions than do that now.

rchiodo: Could you add a comment as to why you added this into the gulpfile?

package-lock.json
```diff
@@ -58349,7 +58326,7 @@
                 "express": "^4.16.3",
                 "filesize": "^3.6.1",
                 "gzip-size": "^5.0.0",
-                "lodash": "^4.17.15",
+                "lodash": ">=4.17.21",
```
rchiodo, line 58329: I don't think these are necessary? Doesn't ```^4.17.15``` mean that it will accept any ```4``` version? So your change to the package.json should cause the rest of these dependencies to just use the version we ship.

rchiodo: Meaning you don't need the hand editing here?

DonJayamanne: Done.

gulpfile.js
DonJayamanne: Sure

rchiodo (approved):

gulpfile.js
DonJayamanne: Done

IanMatthewHuff (approved):

PR #5844, opened 2021-05-11
Fixes to flaky execution tests

DonJayamanne:
In some cases the cells have already started (hence we weren't validating the right thing)

### DIFF
--- a/src/test/datascience/notebook/helper.ts
+++ b/src/test/datascience/notebook/helper.ts
@@ -529,10 +529,31 @@ export async function waitForExecutionInProgress(cell: NotebookCell, timeout: nu
 export async function waitForQueuedForExecution(cell: NotebookCell, timeout: number = defaultTimeout) {
     await waitForCondition(
         async () => {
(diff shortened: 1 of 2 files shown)


rchiodo (approved):

DavidKutu (approved):

src/test/datascience/notebook/helper.ts
```diff
@@ -529,10 +529,31 @@ export async function waitForExecutionInProgress(cell: NotebookCell, timeout: nu
 export async function waitForQueuedForExecution(cell: NotebookCell, timeout: number = defaultTimeout) {
     await waitForCondition(
         async () => {
-            return NotebookCellStateTracker.getCellState(cell) === NotebookCellExecutionState.Pending;
+            return (
```
IanMatthewHuff, line 532: Isn't this function still just Pending? It looks the same as the condition below.

IanMatthewHuff (changes requested):

IanMatthewHuff (approved):

PR #5845, opened 2021-05-11
Remove redundant test

DonJayamanne:
So we can have green tests


IanMatthewHuff (approved):

joyceerhl (approved):
Why is this now redundant?

Issue #5746, opened 2021-05-04
Run Above/Below commands not visible in command palette for Native Notebooks


DonJayamanne:
Regression (available in stable today)

joyceerhl:
Validated in latest insiders:
![image](https://user-images.githubusercontent.com/30305945/118432181-ce6f2800-b68c-11eb-939b-3ef10682989a.png)

PR #5856, opened 2021-05-12
Restore run above/below commands

DonJayamanne:
For #5746

### DIFF
--- a/news/2 Fixes/5746.md
+++ b/news/2 Fixes/5746.md
@@ -1,0 +1,1 @@
+Restore the `Run Above/Below` cells command in `Command Palette`.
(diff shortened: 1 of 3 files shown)


joyceerhl (commented):
Was the original removal of these command from the command palette a PM ask? /cc @claudiaregio

DonJayamanne:
> Was the original removal of these command from the command palette a PM ask?

I didn't think so.
Also would be inconsistent and would be an accessibility issue.
Today we have all cell related commands in comman palette (run cell, run all cells, copy/delete cells, etc), hence I think this should be there as well.

IanMatthewHuff (approved):

PR #5873, opened 2021-05-13
Error out when no python or no kernels

DonJayamanne:
This is a temporary fix (@jmew is working on a spec related to this scenario).
THis is a part of a fix to address `KernelSpecNotFound` errors.

* In webviews, we used to just bomb out `Kernel spec not found` and user doesn't know what to do.
* In Notebooks, we don't have any kernels, and again user doestn't know what to do.

Need to add tests, but wanted to get this out sooner

@claudiaregio /cc

**Solution:**
* Display a meaningful error message
* If running python notebooks & theres' no python, then we display a specific error related to python
* If using native notebooks, we output an error in the cell as well.
Screen Shot 2021-05-13 at 11 49 54
Screen Shot 2021-05-13 at 11 50 18
Screen Shot 2021-05-13 at 11 58 21


claudiaregio:
These look good to me!

DavidKutu (approved):

Issue #5874, opened 2021-05-13
Blank notebooks not using active interpreter


DonJayamanne:

DavidKutu:
Blank notebooks don't use the active interpreter, rather they ask me to select a kernel. 
I think that behavior is ok, but I'm putting this back just to be sure.

DonJayamanne:
> I'm putting this back just to be sure.

The first kernel in the list should be the active interpret

DavidKutu:
validated

PR #5875, opened 2021-05-13
Use active interpreter for blank notebooks

DonJayamanne:
For #5874

### DIFF
--- a/src/client/datascience/jupyter/kernels/helpers.ts
+++ b/src/client/datascience/jupyter/kernels/helpers.ts
@@ -510,7 +510,11 @@ export function findPreferredKernel(
                 // If ther'es no kernelspec in the metadata (e.g. blank notebooks),
                 // & its a python notebook (language in the notebook metatadata will be Python),
                 // Then give preference to the preferred (active) interpreter.
-                if (!notebookMetadata?.kernelspec && preferredInterpreter && spec.interpreterPath === preferredInterpreter?.path) {
+                if (
+                    !notebookMetadata?.kernelspec &&
+                    preferredInterpreter &&
+                    spec.interpreterPath === preferredInterpreter?.path
+                ) {
                     score += 10;
                 }
             }
--- a/src/client/datascience/jupyter/kernels/helpers.ts
+++ b/src/client/datascience/jupyter/kernels/helpers.ts
@@ -361,6 +361,8 @@ export function findPreferredKernel(
         if (kernelMatchingPreferredInterpreter) {
             return kernelMatchingPreferredInterpreter;
         }
+        // Telemetry to see if this happens in the real world, this should not be possible.
+        sendTelemetryEvent(Telemetry.FailedToFindKernelSpecInterpreterForInteractive);
     }
 
     // If still not found, look for a match based on notebook metadata and interpreter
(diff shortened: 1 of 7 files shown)


rchiodo (approved):

src/test/datascience/kernel-launcher/localKernelFinder.unit.test.ts
```diff
@@ -106,7 +106,6 @@ import { IExtensions } from '../../../client/common/types';
                 prerelease: ['0']
             }
         };
-        const activeInterpreter = python3Interpreter;
```
DonJayamanne, line 109: This was the bug, in our tests we were always using the first interpreter.

src/client/datascience/notebook/helpers/helpers.ts
```diff
@@ -107,7 +107,7 @@ export function isPythonNotebook(metadata?: nbformat.INotebookMetadata) {
     }
 
     // Valid notebooks will have a language information in the metadata.
-    return kernelSpec?.language === PYTHON_LANGUAGE;
+    return kernelSpec?.language === PYTHON_LANGUAGE || metadata?.language_info?.name === PYTHON_LANGUAGE;
```
DonJayamanne, line 110: This was the fix.

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -504,6 +506,13 @@ export function findPreferredKernel(
                         )} is ${score}`
                     );
                 }
+
+                // If ther'es no kernelspec in the metadata (e.g. blank notebooks),
+                // & its a python notebook (language in the notebook metatadata will be Python),
+                // Then give preference to the preferred (active) interpreter.
+                if (!notebookMetadata?.kernelspec && preferredInterpreter && spec.interpreterPath === preferredInterpreter?.path) {
```
DonJayamanne, line 513: & this was required

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -361,6 +361,8 @@ export function findPreferredKernel(
         if (kernelMatchingPreferredInterpreter) {
             return kernelMatchingPreferredInterpreter;
         }
+        // Telemetry to see if this happens in the real world, this should not be possible.
+        sendTelemetryEvent(Telemetry.FailedToFindKernelSpecInterpreterForInteractive);
```
DonJayamanne, line 365: Added telemetry to understand if users run into this. Looking at the code, it doesn't make sense how this case would hit

src/client/datascience/notebook/helpers/helpers.ts
IanMatthewHuff: Does the isPythonNotebook helper function work here?

IanMatthewHuff (approved):
Looks legit. Had one comment that we might have a helper function that would have possibly fixed this language check. If that one works, then I'd use it.

src/client/datascience/notebook/helpers/helpers.ts
DonJayamanne: This is the `isPythonNotebook` helper function (hence its a yes)

IanMatthewHuff: Teaches me to not be careful where I was looking. I thought this was an independent check that wasn't using the helper.

DonJayamanne: But I'm glad you are ensuring (in code reviews) we have a common functions instead of duplicating the logic elsewhere..
So, kudos to you, awesome reviewing...

Issue #4585, opened 2021-02-01
Variables are not sortable.


rchiodo:
I know we never implemented this, but being able to click on the 'Name' column seems kinda crucial.

IanMatthewHuff:
Agree :(. I do think that after this is rolling out to stable I want to take a good long look at what the AZNB team did with their variable view. Might be better to sync up with them as opposed to sinking more time into this.

IanMatthewHuff:
Fixed by Vandy

PR #5876, opened 2021-05-13
Sort variables in variable explorer by column

vandyliu:
For #4585

### DIFF
--- a/src/datascience-ui/interactive-common/variableExplorer.tsx
+++ b/src/datascience-ui/interactive-common/variableExplorer.tsx
@@ -69,6 +69,7 @@ interface IGridRow {
 interface IVariableExplorerState {
     containerHeight: number;
     gridHeight: number;
+    variables: IJupyterVariable[];
 }
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
@@ -102,12 +103,14 @@ export class VariableExplorer extends React.Component<IVariableExplorerProps, IV
 
         this.state = {
             containerHeight: this.props.containerHeight,
-            gridHeight: this.props.gridHeight
+            gridHeight: this.props.gridHeight,
+            variables: this.props.variables
         };
 
         this.handleResizeMouseMove = this.handleResizeMouseMove.bind(this);
         this.setInitialHeight = this.setInitialHeight.bind(this);
         this.saveCurrentSize = this.saveCurrentSize.bind(this);
+        this.sortRows = this.sortRows.bind(this);
 
         this.gridColumns = [
             {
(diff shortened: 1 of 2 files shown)


src/datascience-ui/interactive-common/variableExplorer.tsx
```diff
@@ -169,13 +174,26 @@ export class VariableExplorer extends React.Component<IVariableExplorerProps, IV
         }
     }
 
+    public componentWillReceiveProps(nextProps: IVariableExplorerProps) {
+        this.setState({
+              containerHeight: nextProps.containerHeight,
+              gridHeight: nextProps.gridHeight,
+              variables: nextProps.variables
```
joyceerhl, line 181: If `this.props.variables` changes while a sort is active on the variable explorer's contents, would this cause the variable order to reset to an unordered state? I believe the user would expect their existing sort order to be preserved as variables change.

src/datascience-ui/interactive-common/variableExplorer.tsx
```diff
@@ -498,4 +516,22 @@ export class VariableExplorer extends React.Component<IVariableExplorerProps, IV
             this.props.showDataExplorer(row.buttons.variable, row.buttons.numberOfColumns);
         }
     };
+
+    private sortRows(sortColumn: string, sortDirection: 'ASC' | 'DESC' | 'NONE') {
```
rchiodo, line 520: You should probably save the sortColumn/direction so that on refresh you can resort. Although that would mean sorting when set the state.

rchiodo: Hmm, maybe not. Sorting would require fetching all rows so if you resorted on variable refresh that would likely make all variable refreshes fetch every row.

rchiodo: Hmm, thinking about this more, it feels like the variables need to be sorted when requested from the kernel. This way you can still keep the paging algorithm

rchiodo: And then you could save the sort column/direction, but instead of sorting in the UI, you'd make a new variable request to the server.

src/datascience-ui/interactive-common/variableExplorer.tsx
vandyliu: good point, I'll take a look into it

src/datascience-ui/interactive-common/variableExplorer.tsx
vandyliu: Ahh, that makes sense. I'll try that. Thanks :)

rchiodo: Something to keep in mind, the code you'll send to the kernel will likely be specific to python. Try to keep the python semantics out of the internal request as far down as you can go. This will make it easier to add support for say C#/Julia/R in the future. 

For example, you wouldn't want the sort column message from the UI to have any python specific stuff in it.

Issue #5823, opened 2021-05-10
"Run All Cells Above" icon gone and shortcut not working


MattWolf74:
-   VS Code version: 1.56.0
-   Jupyter Extension version (available under the Extensions sidebar): 2021.6.811652604
-   Python Extension version (available under the Extensions sidebar): 2021.5.829140558
-   OS (Windows | Mac | Linux distro) and version: Windows 10 Pro 10.0.19042 Build 19042
-   Python and/or Anaconda version: 3.8.7 64 bit
-   Type of virtual environment used : N/A
-   Jupyter server running:  Local

Issue: 
The "Run All Above Cells" Icon is gone and the shortcut NOT working anymore. It was there along with the other Jupyter icons on the top right in VS Code. Neither does the shortcut work. What is going on? Have the shortcuts changed and why is there no mention in the change-logs of a change to this important feature? 

Can someone please help to make running all the above cells work again? It's hugely important for my  code writing as its very cumbersome to have to manually execute every cell above a location

MattWolf74:
Thanks, will work with the insider in the meantime. Any idea when the next stable release can be expected? Thanks

> Sorry for the inconvenience @MattWolf74. We had to remove them because they had an issue.
> 
> They are back now on the cell toolbar in insiders, please use that in the mean time. They will be back on stable on our next release.

DavidKutu:
My apologies, we should've put it on the change log.

Our next release should be on the first week of June. But I'll see if I can push this to a point release sooner.

DavidKutu:
I checked, and unfortunately, the buttons are not in stable because of an issue with VS Code (the fix is live on Insiders).
We'll have the buttons back on the cell toolbar on the next release of VS Code and the extension.
Please use insiders in the mean time.

Sorry for the trouble.

joyceerhl:
In the latest Jupyter extension Insiders build the run all cells above/below buttons are available in the cell toolbar and the command palette command has been reinstated:

![image](https://user-images.githubusercontent.com/30305945/118433100-18f1a400-b68f-11eb-9057-2f3d45cd0916.png)

As David mentioned, we will ship this fix when VS Code 1.57 stable releases, which should be at the start of June. Apologies again for the inconvenience.

MattWolf74:
That is great, thanks for the update.

Issue #5737, opened 2021-05-03
Run Cells Above and Run Cells Below using incorrect (not selected) kernel


IanMatthewHuff:
It appears to be using a different kernel versus the selected on for performing run cells above and run cells below.

IanMatthewHuff:
'debugEnv' is selected and cells are being run using that. Run above / run below are using something different. You can see in this picture that it's using groovy kernel.

![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/e6784dec-a272-46d0-9be5-6cb64fab2740)

DonJayamanne:
@IanMatthewHuff in light of https://github.com/microsoft/vscode/issues/122877
Not sure whether we should fix this issue.

Even if we fixed it, its possible for the user to select a different kernel when the hit the run button. I'd vote we hide the buttons for now.

DavidKutu:
Dupe of #5497

DonJayamanne:
To be validated on 7th May

DavidKutu:
![image](https://user-images.githubusercontent.com/20347396/121616723-4c1e1d80-ca18-11eb-87e3-3c1cba93d99d.png)
Validated (same python)

PR #5883, opened 2021-05-13
reenable run above and below

DavidKutu:
For #5823

### DIFF
--- a/package.json
+++ b/package.json
@@ -841,13 +841,13 @@
                     "command": "jupyter.notebookeditor.runallcellsabove",
                     "title": "%DataScience.runAbove%",
                     "group": "inline/cell@0",
-                    "when": "notebookViewType == 'jupyter-notebookxxx' && jupyter.isnotebooktrusted"
+                    "when": "notebookViewType == 'jupyter-notebook' && jupyter.isnotebooktrusted"
                 },
                 {
                     "command": "jupyter.notebookeditor.runcellandallbelow",
                     "title": "%DataScience.runBelow%",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -841,13 +841,13 @@
                     "command": "jupyter.notebookeditor.runallcellsabove",
                     "title": "%DataScience.runAbove%",
                     "group": "inline/cell@0",
-                    "when": "notebookViewType == 'jupyter-notebookxxx' && jupyter.isnotebooktrusted"
+                    "when": "notebookViewType == 'jupyter-notebook' && jupyter.isnotebooktrusted"
```
DavidKutu, line 844: We are already using the controller instead of the command to run above and below (https://github.com/microsoft/vscode-jupyter/blob/e2207488952e8a71693d044a811712fd760c66e5/src/client/datascience/notebook/notebookEditor.ts#L319).
This should be the only change needed

joyceerhl: There was a reason we dropped this. Wasn't the existing implementation contributing to https://github.com/microsoft/vscode-jupyter/issues/5737?

DonJayamanne: Aaah yes, you are right, i guess we either ship this with the bug or don't ship it and remove the functionality. which ones the less evil...
Something to be discussed. (to me its obvious, the Push API broken this like plenty of other things 😢 )

DavidKutu: Yea. Don recommended using the controller, I don't know if that would regress to the original issue.
@DonJayamanne what do you think?

PR #5886, opened 2021-05-14
Sort variables by name in variable explorer WITH test added.

vandyliu:
For #4585

10,000 variables test

Attempt 3

Will try to add 'type' column to be sortable afterwards


src/client/datascience/jupyter/kernelVariables.ts
```diff
@@ -418,6 +418,18 @@ export class KernelVariables implements IJupyterVariables {
...
+            type SortableColumn = "name" | "type";
+            const sortColumn = request.sortColumn as SortableColumn;
+            const comparer = (a: IJupyterVariable, b: IJupyterVariable): number => {
+                if (!request.sortAscending) {
+                    return (a[sortColumn] < b[sortColumn]) ? 1 : -1;
+                } else {
+                    return (a[sortColumn] > b[sortColumn]) ? 1 : -1;
+                }
+            };
+            result.pageResponse.sort(comparer);
```
rchiodo, line 430: This isn't going to work. It's only sorting a single page at a time. The page retrieved might even be the wrong one.

Imagine you had 10,000 variables, let's say with names like:

A1
A2
A3
.
.
B1
B2
B3
.

And you fetched the first page. It will return the A1 page. You then sort descending. It still fetch the first page, you'll just have that page sorted descending, but what the user wanted to see was the Z900 to Z800 set of variables.

rchiodo: You have to push the sorting all the way into the python code.

rchiodo: And that's rather complicated because it requires reworking how variables are fetched (we only fetch names and then with a separate request we fetch types).

rchiodo (changes requested):
⏲️

rchiodo:
I think here

https://github.com/microsoft/vscode-jupyter/blob/82b2b5bac0b9bda7914e0623607e623e7bff319f/src/client/datascience/jupyter/kernelVariables.ts#L371

I think the request there is going to change. You won't be getting the list of names, but instead getting all of the data for a page at a time from the server. 

Some piece of python code that will:

- Retrieve the list of variables and their types
- Sort based on name or type
- Get the values for the page requested (as they're visible)
- Returning just that page

You might be able to do this instead (I'm not 100% sure this will work)

- Send request to server to retrieve sorted list of variables with name and type information (not values)
- Replace the code in the kernelVariableProvider to retrieve this list based on if the sort order changes
- Keep the same code that already gets the value for each variable in the page requested

src/client/datascience/jupyter/kernelVariables.ts
vandyliu: Ahh, I see where I went wrong. Thanks for all your input :) 
I shall try again

rchiodo:
Looks great to me. Will you do type in a subsequent PR? Seems like a good idea.

rchiodo (approved):

vandyliu:
> Looks great to me. Will you do type in a subsequent PR? Seems like a good idea.

Yeah, I'm gonna try to add `type`, but wanted do the name one first to keep PRs smaller

joyceerhl (approved):

IanMatthewHuff (approved):
Looks great! Maybe next step on this we should add a test for this in our CI.

DavidKutu (approved):

vandyliu:
@IanMatthewHuff I added a test for it on this PR. 
If it's good, can someone merge 😁

rchiodo:
> @IanMatthewHuff I added a test for it on this PR.
> If it's good, can someone merge 😁

I believe you have linting errors. If you look at the linting task it should tell you what the problem is. 

Alternatively if you install ESLint in VS Code and give it permission to run, it should catch the errors when you save a file locally.

rchiodo:
It looks like the linting error is this:

```
Error: Files are being ignored that should be linted: src/datascience-ui/native-editor/redux/actions.ts
src/datascience-ui/history-react/redux/actions.ts
```

That means in our .eslintrc file you have to remove ```src/datascience-ui/history-react/redux/actions.ts``` from the ignore list.

We didn't start using eslint until somewhat recently and some of the files were skipped as we didn't want to fix all eslint errors at once. As we edit them the build task makes sure the linter will run on newly edited files. After you fix the ignore problem, you might have to fix linter errors in that file too.

rchiodo:
@DavidKutu @IanMatthewHuff @joyceerhl Are those errors expected now? I believe they were some sort of kernel finding problem which shouldn't have been caused by @vandyliu's changes.

joyceerhl:
The local notebook failures seem new. I'm not seeing those fail on the latest main run or on the latest PR run from 1hr ago. Try rerunning?

IanMatthewHuff:
> @DavidKutu @IanMatthewHuff @joyceerhl Are those errors expected now? I believe they were some sort of kernel finding problem which shouldn't have been caused by @vandyliu's changes.

@rchiodo The one failure in nonConda remote notebook was expected. The failure in nonConda 'local' was not expected from what I had last seen.

DavidKutu:
The nonConda local errors seem unrelated. They were already fixed but I guess something might've broken them again.
I just merged something and they passed.

Please try merging main here to see if they still fail.
If they do, go ahead and merge and file a bug and link to this run (https://github.com/microsoft/vscode-jupyter/pull/5886/checks?check_run_id=2634150447)

Issue #5310, opened 2021-03-25
Update LSP packages to latest stable


DonJayamanne:
Address PR https://github.com/microsoft/vscode-jupyter/pull/5169

PR #5946, opened 2021-05-20
Update LSP Packages

DavidKutu:
For #5310


src/test/datascience/reactHelpers.ts
```diff
@@ -414,7 +414,13 @@ export function setUpDomEnvironment() {
             content += 'export function getCSSBasedConfiguration() { return CSSBasedConfiguration.INSTANCE; };\n';
             mod._compile(content, filename);
         } else {
-            _oldLoader(mod, filename);
+            try {
+                _oldLoader(mod, filename);
+            } catch (e) {
```
DavidKutu, line 419: The new LS Packages have issues when loading them to the DOM for tests.
They're not needed so I let them fail. Let me know if we want them loaded for that.

joyceerhl (approved):

rchiodo (approved):

src/test/datascience/reactHelpers.ts
```diff
@@ -414,7 +414,13 @@ export function setUpDomEnvironment() {
...
+                console.error(e);
+                console.error(mod);
+                console.error(filename);
```
DonJayamanne, line 422:
```suggestion
                console.error(`Failed to load module for tests ${mod} in ${filename}`, e);
```

src/test/datascience/reactHelpers.ts
DonJayamanne: Hmm, what if something else fails? can we instead check if `e` is something related to LS package and then ignore it, and if its something else we re-throw the error.

DonJayamanne (changes requested):
I'd prefer to check the error before swallowing it (i.e. only ignore LS specific errors & others should be re-thrown)

IanMatthewHuff (approved):

DonJayamanne (approved):

PR #5947, opened 2021-05-20
Track raw sessions getting disposed while starting kernels

joyceerhl:
Notebook start is failing with raw sessions getting disposed. Currently we don't have enough information to identify the root cause. We need to understand why they're getting disposed as this impacts kernel startup success.


IanMatthewHuff (approved):
Looks solid to me. Might end up being useful telemetry in general.

src/client/datascience/raw-kernel/rawJupyterSession.ts
```diff
@@ -203,15 +203,16 @@ export class RawJupyterSession extends BaseJupyterSession {
...
     ): Promise<void> {
         // REmove our process exit handler. Kernel is shutting down on purpose
         // so we don't need to listen.
         if (this.processExitHandler) {
             this.processExitHandler.dispose();
             this.processExitHandler = undefined;
         }
-        return super.shutdownSession(session, statusHandler, isRequestToShutDownRestartSession).then(() => {
+        return super.shutdownSession(session, statusHandler, isRequestToShutdownRestartSession).then(() => {
+            sendTelemetryEvent(Telemetry.RawKernelSessionShutdown, undefined, { isRequestToShutdownRestartSession });
```
rchiodo, line 215: Does this capture the callstack here? I think that's part of what we need to identify why this is occurring.

rchiodo (changes requested):

rchiodo (approved):

Issue #5927, opened 2021-05-19
Native Notebooks - Getting started w/ Python


claudiaregio:
By default (for both new notebooks or opened notebooks) set the kernel to Python 3. 
When user tries to run a cell, create modal dialog:
"Python Extension required to run Python notebooks. Install?"

DonJayamanne:
Blcoked here https://github.com/microsoft/vscode/issues/124535

PR #5973, opened 2021-05-24
Remove dummy controllers

DonJayamanne:
Reverting due to https://github.com/microsoft/vscode/issues/124535
(tracked here #5927)

### DIFF
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -5,7 +5,7 @@ import { inject, injectable } from 'inversify';
 import { CancellationToken, NotebookControllerAffinity } from 'vscode';
 import { CancellationTokenSource, EventEmitter, NotebookDocument } from 'vscode';
 import { IExtensionSyncActivationService } from '../../activation/types';
-import { IApplicationShell, ICommandManager, IVSCodeNotebook } from '../../common/application/types';
+import { ICommandManager, IVSCodeNotebook } from '../../common/application/types';
 import { PYTHON_LANGUAGE } from '../../common/constants';
 import { traceError, traceInfo, traceInfoIf } from '../../common/logger';
 import {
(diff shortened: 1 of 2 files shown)


src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -195,7 +182,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
                     traceInfo('Find preferred kernel cancelled');
                     return;
                 }
-                await this.createOrDeletePlaceholderPythonKernel();
```
rchiodo, line 198: If I understand the issue correctly the id that the dummy kernel has needs to be updated in order for it to work after you reload VS code?

Perhaps this isn't true? Can we just use the same id for the dummy kernel forever? Save it in user memento or something.

DonJayamanne: >  use the same id for the dummy kernel forever? Save it in user memento or something.

Yes agreed, we could (i didnt want to add that code here), if i did go down that path, I'd be  adding the solution for pre-loads where we have a dummy solution.

DonJayamanne: I.e. i'd like to fix this problem once we have a better understanding of the problem & solution. Right now it seems like i'm going down the path of fixing (coming up with a solution) as we implement something.

rchiodo: So you're backing out the idea of a dummy kernel entirely, and waiting to do this until we decide on how to handle preload kernels?

rchiodo (approved):

src/client/datascience/notebook/notebookControllerManager.ts
IanMatthewHuff: Current implementation feels super reactionary to me, so I'm super pro backing off and approaching this with more of a plan.

IanMatthewHuff (approved):

Issue #5745, opened 2021-05-04
Support Workspace Trust


DonJayamanne:
Filed here https://github.com/microsoft/vscode-python/issues/16126
I'm filing this in case they forget our extension.

Hello :wave: I'm from the VS Code team.

Recently, we have been exploring a security feature we refer to as _Workspace Trust_. This feature is intended to centralize and unify a security conscious decision required by a variety of VS Code features. With workspace trust, the user will be able to declare whether or not they trust the folder that is opened in VS Code before these features are executed.

### Why you should care

Your extension is _incredibly popular_ with VS Code users! We want to make sure that those users have a delightful experience with workspace trust and that includes extension authors deciding how much of their extension is supported in an untrusted workspace.

### Workspace Trust experience

You can enable the feature with the following setting `security.workspace.trust.enabled`. Once enabled, you will see the following dialog when opening folders in VS Code.

![Workspace Trust Startup Dialog](https://media.githubusercontent.com/media/microsoft/vscode-docs/vnext/release-notes/images/1_56/trust-dialog.png)

This dialog is important for allowing the user to make a decision early and understand the impact of their decision. Once you understand the feature, you may want to customize when to display the dialog using the setting `security.workspace.trust.startupPrompt`.

You can follow the development of Workspace Trust and provide feedback in [issue #106488](https://github.com/microsoft/vscode/issues/106488).

### Workspace trust API

First off, all of what I am about to say can be found in [issue #120251](https://github.com/microsoft/vscode/issues/120251). That issue will include discussion of the feature and any updates to the feature.

The Workspace Trust extension API is now in stable. This allowed us to release the first cut of our guide for onboarding your extension to Workspace Trust. The API is small, so here is a quick look.

You can declare your extension to provide complete, partial or no support in untrusted workspaces using the `untrustedWorkspaces` capability in `package.json`.

The following example declares that the extension is supported completely in untrusted workspaces. In this case, the extension is enabled in untrusted workspaces.

```json
"capabilities": {
  "untrustedWorkspaces": {
    "supported": true
  }
}
```

The next example declares that the extension is not supported in untrusted workspaces. In this case, the extension is disabled in untrusted workspaces.

```json
"capabilities": {
  "untrustedWorkspaces": {
    "supported": false
  }
}
```

The third option is to declared `limited` support. There are three tools provided to you when you select the `limited` option.

[...]

PR #6023, opened 2021-05-28
Upgrade notebook trust to VS Codes workspace trust feature

DonJayamanne:
For #5745

* Interactive window features are disabled
	* Cannot create/open one
* Intefactive window (Python file code lenses) disabled
	* Cannor run/debug
* Import/export, Selecting Python interpreter for Jupyter, etc disabled
	* You need Python or Python Extension (nbconvert, etc)
* Removed bycrupt, Digest store, trust service & similar stuff
* INotebookModel.isTrusted is still left in place
	* I'm now returning `workspace.isTrusted` in its place (less code changes for WebView)
* In 3 parts of our code I've ensured we throw errors if we try to accidentally execute code
	* This should not happen, but if it does we've got a fallback.
* When trusted, code lenses appear in interactive
* When trusted native notebooks work as expected
* When trusted, webview notebooks will have their buttons re-enabled.


src/client/common/process/processFactory.ts
```diff
@@ -16,9 +17,14 @@ export class ProcessServiceFactory implements IProcessServiceFactory {
...
         @inject(IProcessLogger) private readonly processLogger: IProcessLogger,
         @inject(IBufferDecoder) private readonly decoder: IBufferDecoder,
-        @inject(IDisposableRegistry) private readonly disposableRegistry: IDisposableRegistry
+        @inject(IDisposableRegistry) private readonly disposableRegistry: IDisposableRegistry,
+        @inject(IWorkspaceService) private readonly workspace: IWorkspaceService
     ) {}
     public async create(resource?: Uri): Promise<IProcessService> {
+        // This should never happen, but if it does ensure we never run code accidentally in untrusted workspaces.
+        if (!this.workspace.isTrusted) {
+            throw new Error('Workspace not trusted');
```
IanMatthewHuff, line 26: Normally I would say this looks like it might end up user visible, so it would need to be localized. If the comment here is correct (we should never get here) ok as is, but if we think this might show up we should localize it.

src/client/datascience/commands/commandRegistry.ts
```diff
@@ -66,26 +72,18 @@ export class CommandRegistry implements IDisposable {
...
         @inject(IJupyterVariables) @named(Identifiers.DEBUGGER_VARIABLES) private variableProvider: IJupyterVariables,
         @inject(UseVSCodeNotebookEditorApi) private readonly useNativeNotebook: boolean,
-        @inject(NotebookCreator) private readonly nativeNotebookCreator: NotebookCreator
+        @inject(NotebookCreator) private readonly nativeNotebookCreator: NotebookCreator,
+        @inject(IWorkspaceService) private readonly workspace: IWorkspaceService
     ) {
         this.disposables.push(this.serverSelectedCommand);
         this.disposables.push(this.notebookCommands);
         this.dataViewerChecker = new DataViewerChecker(configService, appShell);
+        if (!this.workspace.isTrusted) {
```
IanMatthewHuff, line 81: This seems minor, but to be safe. It looks like might leave a gap for a possible double registration. If the workspace is trusted in between the constructor and register being called it looks like it would try to register twice. Should the event be hooked up at the point of register() being called?

IanMatthewHuff (approved):
Awesome work Don. Had a couple of minor comments, but that was it.

src/client/common/process/processFactory.ts
DonJayamanne: > f the comment here is correct (we should never get here) ok as is,

Yes, thats how it should work. Even if it doesn't work as expected, then we should fix the root cause instead of updating this message.

src/client/datascience/commands/commandRegistry.ts
DonJayamanne: >  If the workspace is trusted in between the constructor and register being called it looks like it would try to reg

Cant happen as this is all part of the synchronous code flow (extension loading is synchronous).

IanMatthewHuff: Sounds good as is then. Just want to avoid ever showing an unlocalized message to users.

rchiodo (approved):

Issue #6025, opened 2021-05-28
Remove run above and below buttons from cell toolbar


joyceerhl:
These are now implemented in VS Code core:

![image](https://user-images.githubusercontent.com/30305945/120284617-48003c00-c271-11eb-9d7f-8278aea44872.png)

claudiaregio:
Summary of discussion: Leave the VS Code drop down actions, change our run all above/below icons to use their commands, add telemetry that shows usage of both of those icons.

IanMatthewHuff:
I'd say validated, but it's actually that these are now not provided by VS Code and we put our back in. But I validated that as well.

joyceerhl:
Actually, we ended up removing ours. At least for the 1.57 stable release, the default layout settings disable the consolidated run button, so VS Code's run above/below buttons end up in the cell toolbar.

IanMatthewHuff:
Should have paid more attention to the adding and removing. I thought the cell toolbar ones were back to us.

PR #6026, opened 2021-05-28
Remove our Run Above/Below commands

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/6025

VS Code has implemented this in core. It's controlled by a setting and disabled by default till Tuesday. This must only be merged on Tuesday.


DonJayamanne (approved):

IanMatthewHuff (approved):

joyceerhl:
Should we migrate any keybindings that users have customized for our commands since we are deleting them?

IanMatthewHuff:
> Should we migrate any keybindings that users have customized for our commands since we are deleting them?

I guess I think this might be more trouble than it's worth. This is native only commands right? For insiders customers it doesn't seem too bad to just have them update. My 2 cents at least. Avoid putting more code in.

news/1 Enhancements/6025.md
```diff
@@ -0,0 +1 @@
+Replace 'Run cells above' and 'Run cell and below' commands and cell toolbar buttons with VS Code's built-in 'Execute Above Cells' and 'Execute Cell And Below' commands and unified run button.
```
rchiodo, line 1: I'm wondering if 'interrupt' is going to work correctly or not. We used to be able to cancel all the queued cells for run all when interrrupting. I'm guessing that doesn't work anymore.

rchiodo: We probably will need some special logic in the executeCell commands to handle it.

rchiodo (approved):

joyceerhl:
> This is native only commands right?

@IanMatthewHuff yes, but I believe these commands may be available to stable users in native notebooks as well.

Issue #5590, opened 2021-04-21
Adding additional objects to Data Viewer; e.g. xarray arrays?


max-sixty:
Currently Data Viewer works with a few objects — pandas' DataFrames, numpy arrays, TF & PyTorch tensors. It's v impressive! Would it be possible to add more types of objects?

It would be great to get xarray in there too (disclaimer: I'm a core dev). A couple of approaches:
1. Look for objects with `__array__` methods and call that to return a numpy array. This extension doesn't need to know anything about the array for this to work, but we only get numpy functionality.
2. Add actual xarray support, including the ability to select labels, as well as indices

Here's an example of `np.asarray` & `__array__` working:

```python
In [6]: import xarray as xr

In [11]: da = xr.DataArray(np.random.rand(2,3,4), dims=list('abc'), coords=dict(a=['x','y'], b=['m','n','o']))

In [12]: da
Out[12]:
<xarray.DataArray (a: 2, b: 3, c: 4)>
array([[[0.91948492, 0.20846841, 0.6842219 , 0.16919421],
        [0.34729456, 0.59948279, 0.63013757, 0.82908032],
        [0.18657731, 0.24992941, 0.87905069, 0.93960824]],

       [[0.12899889, 0.30758554, 0.7182392 , 0.76364721],
        [0.68879206, 0.81600394, 0.10102972, 0.13856853],
        [0.43866809, 0.60140471, 0.47634698, 0.37118161]]])
Coordinates:
  * a        (a) <U1 'x' 'y'
  * b        (b) <U1 'm' 'n' 'o'
Dimensions without coordinates: c

In [13]: np.asarray(da)
Out[13]:
array([[[0.91948492, 0.20846841, 0.6842219 , 0.16919421],
        [0.34729456, 0.59948279, 0.63013757, 0.82908032],
        [0.18657731, 0.24992941, 0.87905069, 0.93960824]],

       [[0.12899889, 0.30758554, 0.7182392 , 0.76364721],
        [0.68879206, 0.81600394, 0.10102972, 0.13856853],
        [0.43866809, 0.60140471, 0.47634698, 0.37118161]]])

In [14]: da.__array__()   # same as above
Out[14]:
array([[[0.91948492, 0.20846841, 0.6842219 , 0.16919421],
        [0.34729456, 0.59948279, 0.63013757, 0.82908032],
        [0.18657731, 0.24992941, 0.87905069, 0.93960824]],

[...]

PR #6027, opened 2021-05-28
Adding xarray DataArrays to Data Viewer

vandyliu:
For #5590
Pretty much based off of Joyce's comment 


pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -105,6 +105,8 @@ def _VSCODE_convertToDataFrame(df, start=None, end=None):
         df = _VSCODE_convertTensorToDataFrame(df, start, end)
     elif hasattr(vartype, "__name__") and vartype.__name__ == "ndarray":
         df = _VSCODE_convertNumpyArrayToDataFrame(df, start, end)
+    elif hasattr(df, "__array__") and hasattr(vartype, "__name__") and vartype.__name__ == "DataArray":
+        df = _VSCODE_convertNumpyArrayToDataFrame(df.__array__(), start, end)
```
joyceerhl, line 109: This might be slow, especially when opening the data viewer from the debug variables context menu, since the debug adapter rate-limits how many rows we can fetch at a time (we set this limit to 100 but it really depends on the number of columns in the data as well). This would result in us converting the entire xarray but only returning the [start:end] rows in the request. Suggestion: try opening some 2D and higher dimensional xarrays in our data viewer with the current implementation, see if it's significantly slower. It may be that converting only the rows we need makes this faster (I don't know whether that's the case).

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -139,11 +145,7 @@ def _VSCODE_getDataFrameRows(df, start, end):
     # Turn into JSON using pandas. We use pandas because it's about 3 orders of magnitude faster to turn into JSON
     try:
         df = df.replace(
-            {
-                _VSCODE_np.inf: "inf",
-                -_VSCODE_np.inf: "-inf",
-                _VSCODE_np.nan: "nan",
-            }
+            {_VSCODE_np.inf: "inf", -_VSCODE_np.inf: "-inf", _VSCODE_np.nan: "nan",}
```
vandyliu, line 148: This part was untouched, just reformatted with black.

pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
vandyliu: I see what you're saying, and I saw you mention that doing `df[start:end].__array__()` might be better, which makes sense to me. It was definitely faster on (1000, 1000, 1000) xarrays.

IanMatthewHuff (approved):

rchiodo (approved):

PR #6042, opened 2021-05-31
Extract out python-specific code out of KernelVariables and into PythonVariablesRequester (and Sort variable explorer by types)

vandyliu:
For #4585

This changes some things in KernelVariables so not entirely how badly this would affect other kernels. This addition may not be worth it depending on that. Any/all comments appreciated

Summary:
- Created interface: IKernelVariableRequester
- Created class PythonVariableRequester (which extends the interface above)
- Code that used to be in KernelVariables (and was Python-specific) is now in PythonVariableRequester, and KernelVariables calls IKernelVariableRequester functions


src/client/datascience/jupyter/kernelVariables.ts
```diff
@@ -471,22 +471,50 @@ export class KernelVariables implements IJupyterVariables {
...
             const cells = await notebook.execute(query.query, Identifiers.EmptyFileName, 0, uuid(), token, true);
             const text = this.extractJupyterResultText(cells);
-
-            // Apply the expression to it
             const matches = this.getAllMatches(query.parser, text);
+            const matchesAsStr = matches.map((v) => `'${v}'`);
+
+            // VariableTypesFunc takes in list of vars and the corresponding var names
+            const results = await notebook.execute(
+                `print(${GetVariableInfo.VariableTypesFunc}([${matches}], [${matchesAsStr}]))`,
```
rchiodo, line 493: This is obviously python specific. The query.query was intended to support other languages. I think we could at least extract out the python part into a separate class? Maybe make an interface for getting names and types, like IKernelVariableRequester or something.

So instead of having getVariableNamesandTypesFromKernel be a function, there would be an interface with this single method on it. 

Right now we'd just support python, but this interface could be expanded (or suppled by other extensions) for other languages later.

pythonFiles/vscode_datascience_helpers/getVariableInfo/vscodeGetVariableInfo.py
```diff
@@ -58,3 +58,16 @@ def _VSCODE_getVariableProperties(var, listOfAttributes):
         if hasattr(var, attr)
     }
     return _VSCODE_json.dumps(result)
+
+
+def _VSCODE_getVariableTypes(vars, varnames):
```
rchiodo, line 63: How long does this take to execute for say 10000 variables?

rchiodo (changes requested):
⏲️

pythonFiles/vscode_datascience_helpers/getVariableInfo/vscodeGetVariableInfo.py
rchiodo: I was wondering if we're making another request for types, could we do the list and the types at the same time?

rchiodo: Internally this would essentially call the same function we do now to list the names (you'd have to look up what the magic %_who_ls does in ipykernel)

src/client/datascience/jupyter/kernelVariables.ts
vandyliu: There are other functions that do some python specific things. I'm thinking about putting those into the PythonVariableRequester as well.

pythonFiles/vscode_datascience_helpers/getVariableInfo/vscodeGetVariableInfo.py
vandyliu: think I got something that works well

rchiodo (approved):
Looks good. Like the python variable separation. Should make supporting other languages a lot easier.

IanMatthewHuff (approved):

PR #6070, opened 2021-06-02
Better data when fail to find kernel connection

DonJayamanne:


src/client/datascience/kernel-launcher/remoteKernelFinder.ts
```diff
@@ -165,13 +166,15 @@ export class RemoteKernelFinder implements IRemoteKernelFinder {
                     )}`
                 );
 
+                sendKernelListTelemetry(resource, items);
```
DonJayamanne, line 169: Now we don't need to wait for users to select a kernel, else we only have part of the data.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -145,8 +145,11 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             const controllers = await this.createNotebookControllers(connections);
 
             // Send telemetry related to fetching the kernel connections
-            // KERNELPUSH: undefined works for telemetry?
-            sendNotebookControllerCreateTelemetry(undefined, controllers, stopWatch);
```
DonJayamanne, line 149: This was incorrect, we were sending `undefined`, hence the resource was not a notebook, and the mapping of kernel information for the notebooks was not showing up in telemetry.

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
```diff
@@ -201,6 +207,14 @@ export class HostRawNotebookProvider
                 !kernelConnectionMetadata ||
                 (kernelConnectionMetadata?.kind === 'startUsingKernelSpec' && !kernelConnectionMetadata?.kernelSpec)
             ) {
+                sendTelemetryEvent(Telemetry.KernelSpecNotFoundError, undefined, {
+                    resourceType: getResourceType(resource),
+                    language: getTelemetrySafeLanguage(getLanguageInNotebookMetadata(notebookMetadata)),
+                    kernelConnectionProvided: !!kernelConnection,
+                    notebookMetadataProvided: !!notebookMetadata,
+                    hasKernelSpecInMetadata: !!notebookMetadata?.kernelspec,
+                    kernelConnectionFound: !!kernelConnectionMetadata
```
DonJayamanne, line 216: Need to find out why these errors are being throw for webview notebooks.
Its possible with the recent fixes, the number of these cases would go down, but would like to ensure we have better data in place.

src/client/datascience/telemetry/kernelTelemetry.ts
```diff
@@ -35,35 +34,7 @@ export function sendKernelListTelemetry(
         }
     });
     trackKernelResourceInformation(resource, counters);
-    sendKernelTelemetryEvent(resource, Telemetry.KernelCount, stopWatch.elapsedTime, counters);
-}
-
-export function sendNotebookControllerCreateTelemetry(
```
DonJayamanne, line 41: Removed duplicate function, no different from the other, just an argument, hence just reused.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -145,8 +145,11 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+            sendKernelListTelemetry(
+                Uri.file('test.ipynb'), // Give a dummy ipynb value, we need this as its used in telemetry to determine the resource.
```
rchiodo, line 149: Would it be better  ```'if undefined use 'test.ipynb'```  was in the ```sendKernelListTelemetry```? It seems like it would know better that undefined is not supported.

rchiodo (approved):

src/client/datascience/notebook/notebookControllerManager.ts
DonJayamanne: I don't want to do that, as we have plenty of places in interactive window where we end up with `undefined` for the `resource`.
I.e. this is backwards compatible, `resource` can be undefined & only in the case of interactive window (with notebooks we always have ipynb).

PR #6072, opened 2021-06-02
Use constants for storing default nb formats

DonJayamanne:
I've managed to get a notebook document that cannot be opened in jupyter.
Found that we're hardcoding the version numbers, moved into a single constant.

`orig_nbformat` is updated by jupyter when converting from one version into another.
Since we're creating notebooks, they're always version `4` & not `3`.

This doesn't fix how i ended up with a document that cannot be opened, but cleans it up.
I have a notebook with major & minor as `2`. Not sure how.

### DIFF
--- a/src/client/datascience/interactive-window/interactiveWindow.ts
+++ b/src/client/datascience/interactive-window/interactiveWindow.ts
@@ -27,6 +27,7 @@ import {
 } from '../../common/types';
 import { createDeferred, Deferred } from '../../common/utils/async';
 import * as localize from '../../common/utils/localize';
+import { noop } from '../../common/utils/misc';
 import { EXTENSION_ROOT_DIR } from '../../constants';
 import { captureTelemetry, sendTelemetryEvent } from '../../telemetry';
 import { Commands, defaultNotebookFormat, EditorContexts, Identifiers, Telemetry } from '../constants';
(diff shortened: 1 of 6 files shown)


src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -533,7 +534,7 @@ export class InteractiveWindow extends InteractiveBase implements IInteractiveWi
         }
 
         // Pull out the metadata from our active notebook
-        const metadata: nbformat.INotebookMetadata = { orig_nbformat: 3 };
```
rchiodo, line 536: I thought we had done this on purpose. That the 4.2 would be too new. Swear there was a bug where somebody complained about not supporting 3.0.

DonJayamanne: `orig_format` is not the same as nbformat.
Orig_format is used by Jupyter to keep track of what it used to be.
E.g. if someone opens an existig notebook that was of format 3, and it gets upgraded to 4, then this will contain `orig_format` of 3.

All notebooks we create via `create blank notebook` are of format 4.2 today, we don't support anything other than that (that's existig code where we set nbformat to `4`).
This change makes it more clear.

rchiodo (approved):

PR #6074, opened 2021-06-02
Fixes to breaking tests

DonJayamanne:
For #6104

### DIFF
--- a/src/client/common/application/notebook.ts
+++ b/src/client/common/application/notebook.ts
@@ -39,9 +39,14 @@ export class VSCodeNotebook implements IVSCodeNotebook {
     }
     public get activeNotebookEditor(): NotebookEditor | undefined {
         if (!this.useNativeNb) {
+            console.error('Not using native');
             return;
         }
         try {
+            console.error(`window.visibleNotebookEditors.length = ${window.visibleNotebookEditors.length}`);
+            console.error(`workspace.notebookDocuments.length = ${workspace.notebookDocuments.length}`);
+            console.error(`window.activeNotebookEditor = ${window.activeNotebookEditor}`);
+            console.error(`window.activeTextEditor = ${window.activeTextEditor?.document?.uri?.toString()}`);
             return window.activeNotebookEditor;
         } catch {
             return undefined;
--- a/src/client/common/application/notebook.ts
+++ b/src/client/common/application/notebook.ts
@@ -18,9 +18,7 @@ import {
     window,
     workspace
 } from 'vscode';
-import { IS_CI_SERVER } from '../../../test/ciConstants';
 import { UseVSCodeNotebookEditorApi } from '../constants';
-import { traceErrorIf } from '../logger';
 import { IDisposableRegistry } from '../types';
 import { IApplicationEnvironment, IVSCodeNotebook, NotebookCellChangedEvent } from './types';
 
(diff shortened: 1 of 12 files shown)


src/client/datascience/commands/notebookCommands.ts
```diff
@@ -48,17 +44,8 @@ export class NotebookCommands implements IDisposable {
...
-            );
-            if (cells.length) {
-                const cmd =
-                    cells[0].metadata.outputCollapsed === true
-                        ? 'notebook.cell.expandCellOutput'
-                        : 'notebook.cell.collapseCellOutput';
-                this.commandManager.executeCommand(cmd).then(noop, noop);
-            }
+        if (this.notebookEditorProvider.activeEditor?.toggleOutput) {
+            this.notebookEditorProvider.activeEditor.toggleOutput();
```
DonJayamanne, line 48: I think its cleaner to have it outside this class (that's what we do for other code).

src/test/datascience/notebook/creation/notebookCreation.vscode.test.ts
```diff
@@ -44,7 +44,13 @@ suite('DataScience - VSCode Notebook - (Creation Integration)', function () {
         sinon.restore();
         await closeNotebooksAndCleanUpAfterTests(disposables);
     });
-    teardown(async () => closeNotebooksAndCleanUpAfterTests(disposables));
```
DonJayamanne, line 47: This is a duplicate, and we weren't running the other one.

src/client/common/application/notebook.ts
```diff
@@ -39,11 +41,17 @@ export class VSCodeNotebook implements IVSCodeNotebook {
     }
     public get activeNotebookEditor(): NotebookEditor | undefined {
         if (!this.useNativeNb) {
+            console.error('Not using native');
             return;
         }
         try {
+            console.error(`window.visibleNotebookEditors.length = ${window.visibleNotebookEditors.length}`);
```
DonJayamanne, line 48: To be removed.

src/client/common/application/notebook.ts
```diff
@@ -18,7 +18,9 @@ import {
     window,
     workspace
 } from 'vscode';
+import { IS_CI_SERVER } from '../../../test/ciConstants';
```
IanMatthewHuff, line 21: I don't believe that a constant like this would be an issue either way. But I've seen issues before where we referenced a check from src/test in product code and it caused an issue by causing the test code to initialize. I believe that isCI from src/client/common/constants is what we would want here. It's the same thing, but not referencing into test source from product code.

Looking at the code I see we do this in other spots, so if you want to leave this for now I can file a code-heath issue to clean up all references of this in a later checkin.

src/client/common/application/notebook.ts
IanMatthewHuff: 😆 I've got a PR with all the same logging in. Does this PR have a fix for the creation issue? I wasn't seeing that offhand. Checking since I was looking at that as well.

DonJayamanne: Looks like more tests are failing because of this, i just found that out (my tests were failing because of this as well)
I thought my tests was failing due to some popup (but that wasn't the case), the notebook wasn't getting created.

IanMatthewHuff: I had a PR here that does fix the verify cells test, but I think it's a hack. 
https://github.com/microsoft/vscode-jupyter/pull/6069/files
I explicitly check for one visible document being open, and then I force it active. But I'm not sure this is a good fix as I don't know yet what was making it inactive.

IanMatthewHuff: At least in that case the notebook was getting created, but it's getting set off of active by something. Doesn't happen locally.

src/client/common/application/notebook.ts
DonJayamanne: I'll remove this

DonJayamanne: Actually it was already removed.

src/client/common/application/notebook.ts
DonJayamanne: Damn it, will need to debug this in VSCode and notify VSCode where the actual bug is

DavidKutu (approved):

src/client/datascience/notebook/creation/notebookCreator.ts
```diff
@@ -45,6 +48,7 @@ export class NotebookCreator {
             label: JVSC_EXTENSION_DisplayName
         });
         const placeHolder = DataScience.placeHolderToSelectOptionForNotebookCreation();
+        traceInfoIf(IS_CI_SERVER, `Display quick pick for creation of notebooks ${items.length}`);
```
DonJayamanne, line 51: Damn it, will remove this in another PR.

IanMatthewHuff: @DonJayamanne  it's fine I think that I need to remove them all as a code-health issue. I don't believe it matters now. I'll file an issue.

PR #6085, opened 2021-06-02
change version and API port

DavidKutu:
And port https://github.com/microsoft/vscode-jupyter/pull/6089

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "jupyter",
     "displayName": "Jupyter",
-    "version": "2021.7.0-rc",
+    "version": "2021.6.0-rc",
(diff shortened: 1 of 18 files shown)


joyceerhl (approved):

package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "jupyter",
     "displayName": "Jupyter",
-    "version": "2021.7.0-rc",
+    "version": "2021.6.0-rc",
```
IanMatthewHuff, line 4: Change log is 6.1, do we need that here as well?

DavidKutu: I could, I don't think it matters

IanMatthewHuff: Suppose we update it before release in any case.

IanMatthewHuff (approved):

Issue #6077, opened 2021-06-02
Should not start in untrusted `Attempting to start a server because of preload conditions`


DonJayamanne:

IanMatthewHuff:
Validated. Turned on preload in a new workspace. Tried with both ZMQ on and ZMQ off. No preload when opening a notebook:
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/294d79c3-9cc4-4b14-afb6-29a88dcea128)

PR #6088, opened 2021-06-02
Disable kernel auto startup in untrusted workspace

DonJayamanne:
For #6077

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -194,7 +194,7 @@ jobs:
 
   vsc_api_check:
     needs: pick_environment
-    name: Compile with latest Proposed API
+    name: VSC Stable & Proposed API
     runs-on: ${{ matrix.os }}
     if: github.repository == 'microsoft/vscode-jupyter'
     strategy:
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -242,11 +242,11 @@ jobs:
           path: ./out
           key: ${{runner.os}}-${{env.CACHE_OUT_DIRECTORY}}-${{hashFiles('src/**')}}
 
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1900,7 +1900,7 @@
         "build-ipywidgets-compile": "tsc -p ./src/ipywidgets && rimraf ./out/tsconfig.tsbuildinfo && node ./src/ipywidgets/scripts/copyfiles.js",
         "build-ipywidgets-webpack": "cross-env NODE_OPTIONS=--max_old_space_size=9096 webpack --config ./src/ipywidgets/webpack.config.js",
         "checkDependencies": "gulp checkDependencies",
-        "postinstall": "npm run download-api && node ./build/ci/postInstall.js",
```
DonJayamanne, line 1903: This should be done either as a separate job (optional that can fail).
Else we're all unable to submit PRs even when VS Code Insiders hasn't been updated.

.github/workflows/build-test.yml
```diff
@@ -192,6 +192,67 @@ jobs:
       - name: Check dependencies
         run: npm run checkDependencies
 
+  vsc_api_check:
```
DonJayamanne, line 195: Created a separate Job that will test compilation against the latest type definition files.
This way we can have CI running and perfectly green until VS Code actually ships stuff.

DonJayamanne: For instance, I should have be able to run CI right now with the current VSC Insiders.

DonJayamanne: This will allow us to have both

package.json
DonJayamanne: @joyceerhl @IanMatthewHuff /cc

IanMatthewHuff (approved):

PR #6111, opened 2021-06-03
Fix test failures resulting from VSCodes Notebook Start page

DonJayamanne:
Fixes a few of the test failure's we're having, all seem to have the same root cause (hopefully tha'ts the case).

### DIFF
--- a/src/client/datascience/notebook/creation/notebookCreator.ts
+++ b/src/client/datascience/notebook/creation/notebookCreator.ts
@@ -3,10 +3,8 @@
 
 import { inject, injectable } from 'inversify';
 import { QuickPickItem } from 'vscode';
-import { IS_CI_SERVER } from '../../../../test/ciConstants';
 import { IApplicationShell } from '../../../common/application/types';
 import { JVSC_EXTENSION_DisplayName, JVSC_EXTENSION_ID, PYTHON_LANGUAGE } from '../../../common/constants';
-import { traceInfoIf } from '../../../common/logger';
 import { DataScience } from '../../../common/utils/localize';
 import { sendTelemetryEvent } from '../../../telemetry';
 import { Telemetry, VSCodeNotebookProvider } from '../../constants';
@@ -23,7 +21,6 @@ export class NotebookCreator {
 
     public async createNewNotebook() {
         if (this.creationOptionsService.registrations.length === 0) {
-            console.error('Create using createNew');
             await this.editorProvider.createNew();
             return;
         }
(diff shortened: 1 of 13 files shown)


src/client/datascience/notebook/creation/notebookCreator.ts
```diff
@@ -48,7 +45,6 @@ export class NotebookCreator {
             label: JVSC_EXTENSION_DisplayName
         });
         const placeHolder = DataScience.placeHolderToSelectOptionForNotebookCreation();
-        traceInfoIf(IS_CI_SERVER, `Display quick pick for creation of notebooks ${items.length}`);
```
DonJayamanne, line 51: Sorry @IanMatthewHuff I thought i removed this in the previous PR, removing here.

src/test/datascience/notebook/creation/notebookCreation.vscode.test.ts
```diff
@@ -33,6 +38,7 @@ suite('DataScience - VSCode Notebook - (Creation Integration)', function () {
         creationOptions = api.serviceContainer.get<CreationOptionService>(CreationOptionService);
         vscodeNotebook = api.serviceContainer.get<IVSCodeNotebook>(IVSCodeNotebook);
         creationOptions.clear();
+        await workAroundVSCodeNotebookStartPages();
```
DonJayamanne, line 41: Will add this to other tests, once i know which ones are failing.

src/client/datascience/notebook/creation/notebookCreator.ts
IanMatthewHuff: NBD I filed a code health bug to just clean this usage up all up. No need to sweat it now.

src/test/datascience/notebook/helper.ts
```diff
@@ -416,6 +416,26 @@ export async function stopJupyterServer() {
...
+    if (workedAroundVSCodeNotebookStartPage) {
+        return;
+    }
+    workedAroundVSCodeNotebookStartPage = true;
+    const { editorProvider } = await getServices();
+    await closeActiveWindows();
+
+    // Open a notebook, VS Code will open the start page (wait for 5s for VSCode to react & open it)
+    await editorProvider.createNew();
+    await sleep(5_000);
```
rchiodo, line 435: Blech. Every test takes 5 seconds longer? I guess it's only temporary. 

I'm wondering if there's a way to just wait for the editor open to fire?

rchiodo (approved):

src/test/datascience/notebook/helper.ts
DonJayamanne: The only way is to see if we first have an active notebook editor, and then we don't have an 
 active notebook editor.
I've filed an issue to remove this hack (will be easier once webview/custom goes away).

DonJayamanne: FYI https://github.com/microsoft/vscode-jupyter/issues/6113
Yes, all of this sux...

DavidKutu (approved):

src/test/datascience/notebook/helper.ts
IanMatthewHuff: My hack was to see if we had a visibleEditor, but no active editor, then showNotebookDocument to force it back to showing. That didn't require a wait.

PR #6114, opened 2021-06-03
Remove unwanted Github action workflows

DonJayamanne:


rchiodo:
What do we use to publish now? Is it always manual?

Not sure what the reason is to remove these.

joyceerhl:
I think the publish happens here now:

https://github.com/microsoft/vscode-jupyter/blob/5c2cf0c4cb56d16a5f2124a7e5a72ce72b9266b1/.github/workflows/build-test.yml#L1095-L1170

rchiodo (approved):

PR #6133, opened 2021-06-04
Show run above/below buttons if `notebook.consolidatedRunButton` is enabled

joyceerhl:
VS Code isn't shipping notebook profiles this release. We removed these buttons on the premise that VS Code's implementation of run above/below would appear in the cell toolbar when `notebook.consolidatedRunButton` is set to false (which is the case for the Jupyter profile but not for the default profile). We need to make sure users continue to have these buttons (otherwise we're removing functionality that they previously had access to).

Solution here is to keep our run above/below contributions to the cell toolbar but show them only when the consolidated run button is enabled (since in that case VS Code's buttons won't show in the toolbar and there won't be duplicates).

Alternative is to migrate the `notebook.consolidatedRunButton` setting ourselves, but I believe @roblourens would prefer we didn't configure this setting globally for Jupyter users because the setting applies for all notebooks and the notebook layout story is supposed to take other notebooks into consideration.

There is a limitation to this PR: while the buttons will continue to appear regardless of the setting value, they are bound to different underlying commands, which could be confusing for users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,6 @@
    ([#5954](https://github.com/Microsoft/vscode-jupyter/issues/5954))
 1. Tweak variable view fit and finish to match VS Code.
    ([#5955](https://github.com/Microsoft/vscode-jupyter/issues/5955))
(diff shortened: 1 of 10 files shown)


joyceerhl:
Closing as VS Code will be disabling `notebook.consolidatedRunButton` by default for this release.

Issue #5984, opened 2021-05-25
Allow for LiveKernels to be detected and added / removed as NotebookControllers


IanMatthewHuff:
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/2a350ca3-3b8b-44d1-90c3-108d753b67be)

IanMatthewHuff:
Looks like this is an actual functional issue that might take a bit of work.

When we connect to the remote machine we add our notebook kernels for installed kernels. But we also look for live running kernels. This test does the following:
Starts server
Starts a kernel with a document
Closes the document
Re-opens the document
Connects to the kernel that was previously left running
Runs a cell

But right now we only add NotebookControllers at first document open in the remote case.

There is a basic solution here, where on document open we just run our kernel finder again and addin anything missing from last time we ran it. This should pick up the new running kernels. However I did think that there might be an issue with the VS Code controller API where .dispose wouldn't remove a controller from the list. Will investigate.

IanMatthewHuff:
When this work is done must turn back on and verify the test:
"Use same kernel when re-opening notebook"
I've also put this issue in a comment in the code.

rchiodo:
Validated

PR #6181, opened 2021-06-07
Update live kernel models on remove document open

IanMatthewHuff:
For #5984 

### DIFF
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -60,6 +60,9 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
     // Promise to resolve when we have loaded our controllers
     private controllersPromise?: Promise<VSCodeNotebookController[]>;
 
+    // Listing of the controllers that we have registered
+    private registeredControllers: VSCodeNotebookController[] = [];
+
     private cancelToken: CancellationTokenSource | undefined;
     private readonly isLocalLaunch: boolean;
     constructor(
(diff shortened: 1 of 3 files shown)


DavidKutu (approved):

DonJayamanne (approved):

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -231,6 +257,9 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         document: NotebookDocument,
         token: CancellationToken
     ): Promise<KernelConnectionMetadata | undefined> {
+        // Don't look for a preferred kernel until we have finished our initial controller load
```
rchiodo, line 260: This seems out of place here. It has nothing to do with finding a preferred kernel. Perhaps the caller of this should make this call?

IanMatthewHuff: @rchiodo Sorry I had just been waiting for Don's review to push, so I pushed before this comment came in. I actually agree with you, it's a precondition for setting the preferredController, but it's not a precondition for finding a preferred kernel. I'll do a follow up PR.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -173,35 +183,51 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+        } else {
+            this.findPreferredController(document, preferredSearchToken.token).finally(() => {
                 // Make sure that we clear our finding in progress when done
                 this.findPreferredInProgress.delete(document);
             });
+        }
+    }
+
+    // Find the preferred controller for the given notebook document
+    private async findPreferredController(document: NotebookDocument, cancelToken: CancellationToken) {
```
rchiodo, line 211: Name feels off to me. It's setting the controller mapping, right? That's the goal of this function? Maybe setController?

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -60,6 +60,9 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
     // Promise to resolve when we have loaded our controllers
     private controllersPromise?: Promise<VSCodeNotebookController[]>;
 
+    // Listing of the controllers that we have registered
+    private registeredControllers: VSCodeNotebookController[] = [];
```
rchiodo, line 64: I don't think this is necessary. You only use it one function. And in that function you have to wait on the controllers promise anyway. You could just use the result of getNotebookControllers instead.

rchiodo (changes requested):
⏲️

Issue #6081, opened 2021-06-02
Interactive Window variable window doesn't work right when debugging a cell.


greazer:
- Open a python file with cell markers 
- Run a couple with some variables.
- Open the IW variables pane
- Debug into a cell with variables that have been initialized
- Hover over one of the variables.

**Actual**
The variables pane in the IW will change to reflect just the hovered variable.

**Expected**
The variables pane in the IW either doesn't change at all. It should show the known variables in the cell OR (easier solution) restore the code that pointed users to look at the VS Code debugger variables window instead.

This repro's in latest Insiders as well as Stable 1.56.2, indicating this started happening at least a couple releases ago. It is a regression.

joyceerhl:
I think the right fix here is to disable the variable explorer when debugging. This is what we used to do.

IanMatthewHuff:
Validated. IW looks fine (I'm hovering in this case):
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/a5c1c256-3b6c-4f24-8573-08f8a0f96004)
Nice work @vandyliu

PR #6186, opened 2021-06-07
Fix variable explorer when debugging and hovering

vandyliu:
For #6081

**Update: Rich helped me solve it.
When you go over the local variable, it sends a "hover" "update variables" message, and then gets the internal attributes of a variable like shown below in the 2nd picture. Now we check that the "update variables" message is triggered from a "scopes" message, which is when the variables update because you ran some code, etc.
15kjpCvLZf**

I tried digging into the problem first. It seems that when you hover over a variable in the IW, a new message is sent and `onDidMessageSend` is called with `message.type === 'response' && message.command === 'variables'`, so the variables are updated with that message. However the variables in the message are actually the attributes of the variable that is being hovered over, so the variables in the variable explorer get replaced by all of those variable's attributes. 
Eg. 
Code_-_Insiders_ZUaPIERIrj

Seems like a fix would be to have it so no messages send when you hover over a variable or be able to differentiate between a variable hover and a normal variable message. 

Not sure how much code to remove though for debugger variables.


rchiodo:
Is your intent to fix the real problem later? It seems like you're close to figuring it out.

vandyliu:
> Is your intent to fix the real problem later? It seems like you're close to figuring it out.

I'd love to fix the real problem now honestly, but I am pretty heavily blocked. My original intentions were to try my best to figure it out and if not, disable it (like in this PR) and as stated in #6081. Then make an issue for it to be investigated later on. Maybe we can chat tomorrow on it and see if I can get a little closer?

rchiodo:
> > Is your intent to fix the real problem later? It seems like you're close to figuring it out.
> 
> I'd love to fix the real problem now honestly, but I am pretty heavily blocked. My original intentions were to try my best to figure it out and if not, disable it (like in this PR) as stated in #6081. Then make an issue for it to be investigated later on. Maybe we can chat tomorrow on it and see if I can get a little closer?

Sounds good. Ping me on Teams.

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -207,17 +210,53 @@ export class DebuggerVariables extends DebugLocationTracker
...
+            message.type === 'response' &&
+            message.command === 'variables' &&
+            message.body &&
+            this.currentSeqNumsForVariables.has(message.request_seq)
+        ) {
             // If using the interactive debugger, update our variables.
             // eslint-disable-next-line
             // TODO: Figure out what resource to use
+
+            // Only update variables if it came from a "scopes" command and not a "hover"
```
IanMatthewHuff, line 253: Thanks for the comment here. I need to brush up on the DAP 😆

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -207,17 +210,53 @@ export class DebuggerVariables extends DebugLocationTracker
...
     // This special DebugAdapterTracker function listens to messages sent from the debug adapter to VS Code
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     public onDidSendMessage(message: any) {
         super.onDidSendMessage(message);
         // When the initialize response comes back, indicate we have started.
         if (message.type === 'response' && message.command === 'initialize') {
             this.debuggingStarted = true;
-        } else if (message.type === 'response' && message.command === 'variables' && message.body) {
+        } else if (message.type === 'response' && message.command === 'scopes' && message.body && message.body.scopes) {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
IanMatthewHuff, line 234: Do you need this eslint disable here?

src/client/datascience/jupyter/debuggerVariables.ts
IanMatthewHuff: Looks legit to me but might want to wait for the debugger expert signoff.

IanMatthewHuff (approved):

src/client/datascience/jupyter/debuggerVariables.ts
vandyliu: 😅 Missed that one, ty!

rchiodo (approved):

DavidKutu (approved):

PR #6199, opened 2021-06-08
Remove upload to Azure

DonJayamanne:
This job is failing, and we don't need this (blocks release process)

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1001,29 +1001,6 @@ jobs:
           path: .nyc_output
           retention-days: 1
 
-  upload:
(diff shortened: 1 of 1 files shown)


IanMatthewHuff (approved):

joyceerhl (approved):
At one point I think we were going to have the Python extension's smoke tests pull our latest CI builds, but for now the Python tests just install Jupyter from the marketplace. So this seems OK to do. https://github.com/microsoft/vscode-python/blob/c1a8d94c6b8caa658aa99109111a3ee3551c8fe0/src/test/standardTest.ts#L42-L46

Issue #6185, opened 2021-06-07
Sometimes it's very slow to open an ipynb file (native notebooks)


greazer:

greazer:
May be related to #5193

DonJayamanne:
This is actually a regression.
Previously I fixed this, now we're activating the Python extension every time when opening a notebook.
I think we need to get this fixed and MUST add the tests to ensure we don't regress again.
Previously we'd load kernels only when opening a notebook, now we load kernels as soon as the extension activates.

I.e. we now activate Python extension EVERYTIME, & this is a regression from previous behavior

Basically when we load list of kernels:
* If python extension is installed, then we actiavte the python extension and get list of interpreters
* Event if we're opening a C# notebook.

I think we need to refactor the kernel provider:
**Short term**
* Provide non-python kernels first
* After the document has been opened, then activate the python extension and load the python kernels
* Note: Even if we moved python extension code into an npm module we'd still run into the same issue (we'd still load Python interpreters from within Jupyter, and perform all of the discovery all over again - i.e. moving Python into npm doesn't buy much, a little yes)

I think the problem is the fact that we're activating Python extension before opening the notebook.
We need to delay that until the notebook has been opened.

PR #6201, opened 2021-06-08
Do not activate Python before opening nb

DonJayamanne:
For #6185

### DIFF
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -101,11 +101,6 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
 
         // Be aware of if we need to re-look for kernels on extension change
         this.extensions.onDidChange(this.onDidChangeExtensions, this, this.disposables);
-
-        if (this.isLocalLaunch) {
-            // Pre-warm fetching local kernels, for remote connections fetch as and when needed.
-            this.getNotebookControllers().catch(traceError);
-        }
     }
 
     // Look up what NotebookController is currently selected for the given notebook document


src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -101,11 +101,6 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
 
         // Be aware of if we need to re-look for kernels on extension change
         this.extensions.onDidChange(this.onDidChangeExtensions, this, this.disposables);
-
-        if (this.isLocalLaunch) {
-            // Pre-warm fetching local kernels, for remote connections fetch as and when needed.
-            this.getNotebookControllers().catch(traceError);
```
rchiodo, line 107: Not sure how this slows things down? Is it the python extension activation?

rchiodo: I guess it feels like we need to do this so that running a cell doesn't take forever. I'm wondering how this slows down the open as it should be a promise that's going to at least going to go off and look on disk for kernelspecs.

DonJayamanne: Yes,
This causes pyton extension actiavtion, hence slows down extenison host and opening of the notebook.
We alraedy have code that gets this information after we get the notebook opened event.

IanMatthewHuff (approved):

src/client/datascience/notebook/notebookControllerManager.ts
DonJayamanne: > t should be a promise that's going to at least going to go off and look on disk for kern

Python extnesion spins up processes (process spawning is a sync process), all of this and other code (such as loading python extension js is synchronous), that blocks the extenssion host hence slows the current process of opening a notebook.

PR #6206, opened 2021-06-08
LiveKernelModel - Review followup

IanMatthewHuff:
@rchiodo Had good feedback on a PR just after I pushed it, so added a follow up PR here (though this is point release, so I need to port both).

### DIFF
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -40,6 +40,7 @@ import { InterpreterPackages } from '../telemetry/interpreterPackages';
 import { sendTelemetryEvent } from '../../telemetry';
 import { NotebookCellLanguageService } from './cellLanguageService';
 import { sendKernelListTelemetry } from '../telemetry/kernelTelemetry';
+import { testOnlyMethod } from '../../common/utils/decorators';
 /**
  * This class tracks notebook documents that are open and the provides NotebookControllers for
  * each of them
@@ -113,7 +114,9 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         }
     }
 
-    public getNotebookControllers(): VSCodeNotebookController[] {
+    // Function to expose currently registered controllers to test code only
+    @testOnlyMethod()
+    public registeredNotebookControllers(): VSCodeNotebookController[] {
         return this.registeredControllers;
     }
 
@@ -234,7 +237,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         // can happen if a document is opened quick and we have not yet loaded our controllers
         await this.loadNotebookControllers();
 
-        const targetController = this.getNotebookControllers().find((value) => {
+        const targetController = this.registeredControllers.find((value) => {
             // Check for a connection match
             return areKernelConnectionsEqual(kernelConnection, value.connection);
         });
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -58,7 +58,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
     }>;
 
     // Promise to resolve when we have loaded our controllers
-    private controllersPromise?: Promise<VSCodeNotebookController[]>;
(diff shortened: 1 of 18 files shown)


src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -257,9 +257,6 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         document: NotebookDocument,
         token: CancellationToken
     ): Promise<KernelConnectionMetadata | undefined> {
```
IanMatthewHuff, line 259: I was basically doubling up on this call. Instead just rely on the call in setPreferredController, findPreferredConnection doesn't need controllers to operate.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -447,6 +444,9 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
 
         // If we have any out of date connections, dispose of them
         disposedControllers.forEach((controller) => {
```
IanMatthewHuff, line 446: @rchiodo you mentioned not needing to save registeredControllers, but we both add (a few lines above here) and remove from this set after the initial load, so I think that we do need to save it off and not just use the promise results.

rchiodo: Then it seems that getNotebookControllers is incorrect. If somebody else calls that function they won't get the registered controllers as the list is modified outside of that function.

IanMatthewHuff: Bleh, there is indeed a grody discrepancy there.

So maybe this:

- getNotebookControllers => loadNotebookControllers - Responsible for performing the initial load of controllers. Returns an empty promise so it can be awaited (tests use this) but just assigns results to the registered list and doesn't return the list.
- getNotebookControllers - New function. Non-async to just return the current set of registered controllers. Believe that we only would need this for tests.

IanMatthewHuff: Looks like the remote test I reenabled is failing anyways so gotta investigate.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -113,15 +113,18 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         }
     }
 
+    public getNotebookControllers(): VSCodeNotebookController[] {
```
rchiodo, line 116: Seems like this should be async and await the load? Just to make sure. Otherwise there's an implicit dependency between the two.

IanMatthewHuff: @rchiodo Yeah, I'm just playing around a bit in draft now. So right now this function is only used for tests and as I see it, it would always reflect the current state of registered controllers. I don't actually want it to trigger the load, it's just the current state of what is registered for tests.

So you could do in a test:
1. getNotebookControllers check that it's 0
2. open a document
3. Wait for selection
4. getNotebookControllers check that it's the expected number

DavidKutu (approved):

src/client/datascience/notebook/notebookControllerManager.ts
rchiodo: Hmm, still not happy with the public part of it then. I wish typescript had InternalsVisibleTo. Can you put a comment saying this is only for tests?

rchiodo (approved):

src/client/datascience/notebook/notebookControllerManager.ts
IanMatthewHuff: @rchiodo We've talked over this some for other test only interfaces. I think that I've used a specific test only interface in other places. Let me look that up. Might be better here along with a comment. Will update.

IanMatthewHuff: Ahh, now I remember what I did. For WebViewHost I created a decorator that enforces test only execution. That + a test interface should work here.

IanMatthewHuff: Added the decorator on the function + comment on the interface.

rchiodo (approved):

PR #6252, opened 2021-06-10
Do not assume we're the controller

DonJayamanne:


src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -245,12 +245,6 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         if (targetController) {
             traceInfo(`TargetController found ID: ${targetController.id} for document ${document.uri.toString()}`);
             await targetController.updateNotebookAffinity(document, NotebookControllerAffinity.Preferred);
-
-            // When we set the target controller we don't actually get a selected event from our controllers
-            // to get around that when we see affinity here 'force' an event as if a user selected it
-            this.handleOnNotebookControllerSelected({ notebook: document, controller: targetController }).catch(
```
DonJayamanne, line 251: @IanMatthewHuff Not sure why we're doing this.
Basically if we're setting the preferred controller we shodj'nt have to do this.
I found that with .NET extension installed, we run code in `handleNotebookControllerSelected` which is incorrect (VS Code remembers what the last kernel was , assume it was .NET kernel, but we're running code as though user selected ours, when ,NET is selected).

Basically we're running code when we shouldn't be.
Hopefully the API has changed enough to make this code unnecessary anymore.

DonJayamanne: I'll file a bug with the actual issue (what happens when .NET extension is installed), then we can decide if we need this fix in a a point release.

IanMatthewHuff: @DonJayamanne Previous discussion: 
https://github.com/microsoft/vscode-jupyter/pull/5635#discussion_r620661779

Basically from when we were going back and forth on selection vs. preferred ect...Firing this event allowed us to keep doing our auto start  and other things that expected to happen on selection now that selection was happening at a different point.

IanMatthewHuff: Per the discussion then I think that we can remove it now. I don't think that either you or I really wanted it in. It was when we expected a "selected" event and it allow us to unblock tests and the rest of our functionality. With the API change since then (+ the test hack to force selection) I don't think it's needed. 

Auto start might still be an open question, but I think that has issues right now anyways.

IanMatthewHuff (approved):

rchiodo (approved):

PR #6293, opened 2021-06-15
Ensure our contributed notebook keybindings respect `jupyter.enableKeyboardShortcuts` setting

joyceerhl:
This also fixes the issue we found where ctrl+enter inserts a new cell below the current cell when in command mode. This was caused by a built-in keybinding:

image

Solution is to ensure our contributed ctrl+enter keybinding (for executing the current cell without advancing and leaving the cell in command mode) is active when the notebook cell list has focus, not just when a cell is in edit mode.

While looking at this I also found that the when clauses used for VS Code's ctrl+enter and shift+enter keybindings check the cell execution state and active kernel count. This didn't seem at all necessary to me so I removed those checks from the when clauses. Looking for feedback on whether that's a reasonable approach.

This also needs to be ported to main, but I'm creating this directly into the release branch first so as not to hold up the point release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -174,22 +174,22 @@
             },
             {
                 "key": "ctrl+enter",
-                "when": "editorTextFocus && inputFocus && notebookEditorFocused && notebookType == jupyter-notebook && config.jupyter.enableKeyboardShortcuts",
+                "when": "editorTextFocus && inputFocus && notebookEditorFocused && notebookType == jupyter-notebook && config.jupyter.enableKeyboardShortcuts || notebookCellListFocused && notebookCellType == 'code' && notebookType == jupyter-notebook && config.jupyter.enableKeyboardShortcuts",
                 "command": "jupyter.notebookeditor.keybind.executeCell"
             },
             {
                 "key": "ctrl+enter",
-                "when": "inputFocus && notebookEditorFocused && !editorHasMultipleSelections && !editorHasSelection && !editorHoverVisible && notebookCellType == 'markup'",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -174,22 +174,22 @@
             },
             {
                 "key": "ctrl+enter",
-                "when": "editorTextFocus && inputFocus && notebookEditorFocused && notebookType == jupyter-notebook && config.jupyter.enableKeyboardShortcuts",
+                "when": "editorTextFocus && inputFocus && notebookEditorFocused && notebookType == jupyter-notebook && config.jupyter.enableKeyboardShortcuts || notebookCellListFocused && notebookCellType == 'code' && notebookType == jupyter-notebook && config.jupyter.enableKeyboardShortcuts",
```
rchiodo, line 177: What's the priority for an || statement? There's no parenthesis here so not sure if the or applies to just the two items next to it or the whole statement after?

rchiodo (approved):

DavidKutu (approved):

IanMatthewHuff (approved):

package.json
joyceerhl: I don't believe parentheses are supported in when clauses. At least I tried the following and it didn't work:

```
"when": "((editorTextFocus && inputFocus && notebookEditorFocused) || notebookCellListFocused && notebookCellType == 'code')) && notebookType = jupyter-notebook && config.jupyter.enableKeyboardShortcuts"
```

The priority is not documented, but I would be very surprised if && didn't trump || in precedence.

joyceerhl: I believe this is the spot for parsing when expressions. Looks like VS Code splits on '||', then '&&', so unless I'm misreading and this code isn't the right spot, it seems like our when clauses should do what we expect https://github.com/microsoft/vscode/blob/5b73c885890e95af601df0f968194efeab1b70cd/src/vs/platform/contextkey/common/contextkey.ts#L128-L144

PR #6296, opened 2021-06-15
Data Wrangler (feature PR)

vandyliu:
This is a feature branch that I will be merging PRs into, then in the end merging it to main.

In the original PR, there are the data wrangler parts, slick grid, webview, the command palette commands. 
I have removed the side panels so it is a little easier to review.

I put `data-wrangler` under `data-viewer` because data-wrangler uses quite a few stuff from `data-viewer` but I'm open to comments/opinions.

PR review video availabe in Data Wrangler Feature Branch PR Call teams chat


src/datascience-ui/data-explorer/data-wrangler/reactSlickGrid.tsx
```diff
@@ -0,0 +1,802 @@
...
+                    {/* <Resizable 
+                        style={{display: "flex", alignItems: "top", justifyContent: "left", flexDirection: "column", zIndex: 99998 }}
+                        handleClasses={{ right: "resizable-span" }}
+                        defaultSize={{ width: '60%', height }}
+                        enable={{ left:false, top:false, right:true, bottom:false, topRight:false, bottomRight:false, bottomLeft:false, topLeft:false }}
+                        > */}
+                    <div className="react-grid-container" style={style} ref={this.containerRef}></div>
+                    <div className="react-grid-measure" ref={this.measureRef} />
+                    <Resizable>
+                        {/* <Resizable
```
vandyliu, line 339: Side Panel (will be in another PR)

src/client/common/application/customEditorService.ts
```diff
@@ -36,10 +36,11 @@ export class CustomEditorService implements ICustomEditorService, IExtensionSing
         if (this.useCustomEditorApi) {
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
             return (vscode.window as any).registerCustomEditorProvider(viewType, provider, options);
-        } else {
+        } else if (!this.useCustomEditorApi) {
```
rchiodo, line 39: Weird? How can a boolean be both not true and not false?

src/client/datascience/commands/commandRegistry.ts
```diff
@@ -552,3 +562,26 @@ export class CommandRegistry implements IDisposable {
         }
     }
 }
+
+export function getImportCodeForFileType(filepath: string) {
```
rchiodo, line 566: Seems like a weird spot for this? Not sure what it has to do with the commandRegistry?

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,685 @@
...
+    IDropNaRequest,
+    OpenDataWranglerSetting,
+    IDataWranglerJupyterVariableDataProvider,
+    IDataWranglerJupyterVariableDataProviderFactory
+} from './types';
+
+const PREFERRED_VIEWGROUP = 'JupyterDataWranglerPreferredViewColumn';
+const dataWranglerDir = path.join(EXTENSION_ROOT_DIR, 'out', 'datascience-ui', 'viewers');
+@injectable()
+export class DataWrangler extends WebviewPanelHost<IDataWranglerMapping> implements IDataWrangler, IDisposable {
```
rchiodo, line 72: I'm wondering if this class and the DataViewer can share anything? They are pretty similar.

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,685 @@
...
+            }
+        }
+    }
+
+    private renameColumn(currentVariableName: string, req: IRenameColumnsRequest): IHistoryItem {
+        this.variableCounter += 1;
+        const newVariableName = `df${this.variableCounter}`;
+        const code = `${newVariableName} = ${currentVariableName}.rename(columns={ "${req.oldColumnName}": "${req.newColumnName}" })\n`;
+        const historyItem = {
+            transformation: `Renamed column "${req.oldColumnName}" to "${req.newColumnName}"`,
```
rchiodo, line 565: I believe these 'actions' need to be localized.

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,685 @@
...
+            case DataWranglerCommands.GetHistoryItem:
+                this.getHistoryItem(payload.args.index).ignoreErrors();
+                break;
+
+            case DataWranglerCommands.Describe:
+                void this.getColumnStats(payload.args.columnName);
+                break;
+        }
+
+        const dataCleaningMode = this.configService.getSettings().dataCleaningMode;
```
rchiodo, line 520: You might abstract out this two pronged approach into separate classes that implement an interface? DataWranglerCommandHandler or something like that? One that executes in a notebook. One that is standalone?

rchiodo: Or just two functions at least with the same signature.

rchiodo: I think the commands could also be a map from command to function (instead of the big switch statement), each function returning a historyItem or undefined if no further history update required.

src/client/datascience/data-viewing/data-wrangler/dataWranglerJupyterVariableDataProvider.ts
```diff
@@ -0,0 +1,188 @@
...
+
+import { Identifiers } from '../../constants';
+import { IJupyterVariable, IJupyterVariables, INotebook } from '../../types';
+import { DataViewerDependencyService } from '../dataViewerDependencyService';
+import { ColumnType, IDataFrameInfo, IRowsResponse, IColsResponse } from '../types';
+import { traceError } from '../../../common/logger';
+import { IDataWranglerJupyterVariableDataProvider } from './types';
+
+@injectable()
+export class DataWranglerJupyterVariableDataProvider implements IDataWranglerJupyterVariableDataProvider {
```
rchiodo, line 16: Is this much different than IJupyterVariables? Why not add the new methods to IJupyterVariables and just use those? 

That would have the added benefit of supporting doing this during debugging too (although that's not a likely use case)

src/client/datascience/data-viewing/data-wrangler/dataWranglerProvider.ts
```diff
@@ -0,0 +1,266 @@
...
+        if (dataCleaningMode == '') {
+            const qpoptions: QuickPickOptions = {
+                ignoreFocusOut: false,
+                matchOnDescription: true,
+                matchOnDetail: true
+            };
+
+            const qpitems = [
+                {
+                    label: 'Open Just the Data Wrangler',
```
rchiodo, line 120: Localization for these strings.

src/datascience-ui/data-explorer/data-wrangler/mainPanel.tsx
```diff
@@ -0,0 +1,632 @@
...
+    //                     ) : undefined}
+    //                     <span>{breadcrumbText}</span>
+    //                 </div>
+    //             </div>
+    //         );
+    //     }
+    // }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    public handleMessage = (msg: string, payload?: any) => {
```
rchiodo, line 245: I wonder if we can abstract out part of this. A lot of the webviews have to handle the onigasm stuff. Seems like we don't need to repeat it in each.

src/datascience-ui/data-explorer/data-wrangler/reactSlickGrid.tsx
```diff
@@ -0,0 +1,802 @@
...
+    handleRefreshRequest(): void;
+}
+
+interface ISlickGridState {
+    grid?: Slick.Grid<ISlickRow>;
+    showingFilters?: boolean;
+    fontSize: number;
+}
+
+export class ReactSlickGrid extends React.Component<ISlickGridProps, ISlickGridState> {
```
rchiodo, line 97: Yeah this has to be reusable. It looks identical to the one in the data viewer.

rchiodo (changes requested):
Looks pretty good. I think there are some spots for code reuse though. Open to discussion on them of course. Or some other way to achieve reuse.

src/client/datascience/commands/commandRegistry.ts
```diff
@@ -477,14 +482,18 @@ export class CommandRegistry implements IDisposable {
         }
     }
 
-    private async createNewNotebook(): Promise<void> {
+    private async createNewNotebook(): Promise<INotebookEditor | undefined> {
         if (this.useNativeNotebook) {
-            await this.nativeNotebookCreator.createNewNotebook();
+            return await this.nativeNotebookCreator.createNewNotebook();
```
IanMatthewHuff, line 487: Isn't return await redundant? Thought there was an eslint rule for that.

package.json
```diff
@@ -1825,6 +1873,16 @@
                     }
                 ],
                 "priority": "option"
+            },
+            {
+                "viewType": "jupyter-data-wrangler",
+                "displayName": "Data Wrangler",
+                "selector": [
+                    {
+                        "filenamePattern": "*.csv"
```
IanMatthewHuff, line 1882: Are the other file types needed here? Not sure until I read further in the review.

src/client/datascience/commands/commandRegistry.ts
```diff
@@ -552,3 +562,26 @@ export class CommandRegistry implements IDisposable {
...
 }
+
+export function getImportCodeForFileType(filepath: string) {
+    const fileExtension = path.extname(filepath);
+    let code = 'import pandas as pd\n';
+    switch (fileExtension) {
+        case '.csv':
+            code += `df = pd.read_csv(r"${filepath}")`;
+            break;
+        case '.xlsx': // TODO dependency check for openpyxl
```
IanMatthewHuff, line 573: TODOs are ok, but are they being tracked somewhere? They always make me nervous going in, easy to forget. If you have more might want to tag them for easy searching. Like (DataWranglerTODO).

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
vandyliu: Added to localize.ts and package.nls.json

src/client/datascience/commands/commandRegistry.ts
vandyliu: Good point, removed them

package.json
vandyliu: Will just be implementing csv for now

src/client/datascience/commands/commandRegistry.ts
vandyliu: Yep, it is and removed all instances of it. Ty

src/client/datascience/data-viewing/data-wrangler/dataWranglerJupyterVariableDataProvider.ts
vandyliu: I guess I was scared of adding more methods to the interface in case other components relied on them and would have to update them, but I see your point and will just add it there.

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
vandyliu: They are definitely very similar. So are some of the other classes, I really really want to figure out the best way to re-use code.  Here are my thoughts: abstract classes, but it won't work because the classes `extend` the same class but the `Mapping` is different. Same for the React classes with the props. Unless I made it so they used the same mappings, and made the data wrangler ones optional. I could also have a dataViewingHelper (probably needs a better name) class that both dataViewer and dataWrangler call when the code is the same but there will still probably be a bit of duplicate code. Any other ideas or thoughts on the above? They are very welcomed and thank you for your review. I appreciate it a lot.

rchiodo: DataWrangler could probably be a subclass of DataViewer on the extension side? Its handleMessage might have some extra message handlers but otherwise they seem identical. 

On the react side, the message handling could be abstracted into a base class for both (and others) or you could use some object to handle the messages.  Actually I think the tmlanguage and onigasm stuff is already abstracted using redux, but maybe the datawrangler/dataviewer don't bother with redux. If they don't, then just creating a common base class or some other object to handle those messages seems like it would work.

vandyliu: Alright, so I did the `extends` implementation
I decided to just add DataWranglerMessages to IDataViewerMapping
I played around with the constructor because I wasn't sure about the best way to get the different values into WebviewPanelHost<IDataViewerMapping> (Eg. title, directory strings). What do you think?

rchiodo: Couldn't DataWrangler extend DataViewer? They're very similar

vandyliu: Yep, that's what I did in my most recent commits.

src/client/datascience/jupyter/kernelVariables.ts
```diff
@@ -152,6 +152,20 @@ export class KernelVariables implements IJupyterVariables {
         return {};
     }
 
+    public async getDataFrameColumn(
```
rchiodo, line 155: 👍

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
vandyliu: I see what you mean with the separate classes implementing an interface. It would especially be smart once other forms are added like WITH_INTERACTIVE_WINDOW, etc. It started getting tricky due to the async/awaits and disposables. For the two functions, I believe that doesn't work in typescript because of it getting compiled to javascript (or I may be misinterpreting what you mean). I do like the command to function idea, but there are some async/await functions which makes it a little awkward. I can make them all async. Do you think that would be worthwhile exploring?

rchiodo: Making them all async sounds fine to me. Non async functions that are marked that way don't infer a penalty AFAIK. They just run synchronously.

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,502 @@
...
+
+            default:
+                break;
+        }
+
+        // Some messages will be handled by DataViewer
+        super.onMessage(message, payload);
+    }
+
+    private async requestTmLanguage(languageId: string = PYTHON_LANGUAGE) {
```
rchiodo, line 226: We must be doing this somewhere else? Ah yes we are. InteractiveBase. I think you can move this up one level to WebviewPanelHost and then that would work for this code too.

Then all of the monaco init stuff will be in one spot.

rchiodo: Meaning change InteractiveBase and this code to not have these monaco related functions, but instead move them to WebviewPanelHost.

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,502 @@
...
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private async handleCommand(payload: { command: string; args: any }) {
+        const notebook = (this.dataProvider as IJupyterVariableDataProvider).notebook;
+        let code = '';
+        const currentVariableName = (await this.dataFrameInfoPromise)!.name ?? '';
+        let newVariableName = currentVariableName ?? '';
+
+        // Get and run data wrangler command
+        const cmd = this.commands.get(payload.command as DataWranglerCommands);
```
rchiodo, line 316: Much nicer 👍

src/client/datascience/data-viewing/data-wrangler/dataWranglerProvider.ts
```diff
@@ -0,0 +1,157 @@
...
+import { Identifiers } from '../../constants';
+import { INotebookProvider, IJupyterVariables, IJupyterVariableDataProviderFactory } from '../../types';
+import { DataViewerChecker } from '../../interactive-common/dataViewerChecker';
+import { IConfigurationService } from '../../../common/types';
+import { IDataWranglerFactory } from './types';
+import { DataScience } from '../../../common/utils/localize';
+import { IDataViewerDataProvider } from '../types';
+
+@injectable()
+export class DataWranglerProvider implements IDataWranglerProvider, IExtensionSingleActivationService {
```
rchiodo, line 25: can you put a comment on the top of each class explaining what it's for? For example it's not clear immediately the difference between DataWranglerProvider and DataWranglerFactory

src/datascience-ui/data-explorer/data-wrangler/controls/toolbar.tsx
```diff
@@ -0,0 +1,85 @@
...
+                        className="codicon codicon-go-to-file codicon-button"
+                        style={{ verticalAlign: 'middle' }}
+                        title="Open as Python script"
+                    />
+                    <span style={{ verticalAlign: 'middle', paddingLeft: '4px', paddingBottom: '4px' }}>
+                        Open as Python script
+                    </span>
+                    {/* <div className="codicon codicon-notebook codicon-button" title="Open in Notebook" /> */}
+                </div>
+                {/* <div
```
rchiodo, line 57: Why is a bunch of this commented out? Do we intend to enable this in the future?

src/datascience-ui/data-explorer/data-wrangler/mainPanel.tsx
rchiodo: The state transitions in the datawrangler may warrant using redux instead of handling everything in the react code, but I think we can leave that till later/next pr.

src/test/datascience/data-viewing/dataWrangler.unit.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        await dataWrangler.showData(instance(dataProvider), title);
+
+        verify(dataProvider.getDataFrameInfo(anything(), anything())).once();
+    });
+    test('Data Wrangler calls data provider dispose', async () => {
+        await dataWrangler.showData(instance(dataProvider), title);
+        dataWrangler.dispose();
+
+        verify(dataProvider.dispose()).once();
+    });
```
rchiodo, line 91: I think we should have an integration test for it? Maybe I missed that.

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
vandyliu: The IMapping part is a little tricky, I'll message you about it.

src/test/datascience/data-viewing/dataWrangler.unit.test.ts
vandyliu: added a simple one

src/datascience-ui/data-explorer/data-wrangler/controls/toolbar.tsx
vandyliu: still wasn't sure if they were going to be included or not but got confirmation and removed them now

package.json
```diff
@@ -2044,20 +2108,25 @@
         "node-stream-zip": "^1.6.0",
         "onigasm": "^2.2.2",
         "pdfkit": "^0.12.1",
+        "plotly.js": "^1.58.4",
```
IanMatthewHuff, line 2111: Is this including plotly twice? This might just be one minified and one-non minified version.

vandyliu: yep, and fixed ty

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,461 @@
...
+            case DataWranglerMessages.SubmitCommand:
+                this.handleCommand(payload).ignoreErrors();
+                break;
+
+            case DataWranglerMessages.RefreshDataWrangler:
+                this.refreshData().ignoreErrors();
+                break;
+
+            case InteractiveWindowMessages.LoadTmLanguageRequest:
+                void this.requestTmLanguage(payload);
```
rchiodo, line 205: Shouldn't these be handled by the super.handleMessage?

src/client/datascience/data-viewing/data-wrangler/dataWranglerFactory.ts
```diff
@@ -0,0 +1,132 @@
...
+        const editor = window.activeTextEditor;
+        if (!editor) return;
+        const document = editor.document;
+        if (!document) return;
+        const position = editor.selection;
+        if (!position) return;
+        // See if a variable exists
+        // Look for an active data wrangler
+        if (this.knownDataWranglers.size === 0) {
+            // Create a new data wrangler
```
rchiodo, line 118: Seems like this case isn't actually being handled?

src/client/datascience/data-viewing/data-wrangler/dataWranglerProvider.ts
rchiodo: Can you put this same comment on all of the new classes you added?

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
```diff
@@ -47,10 +49,14 @@ export class JupyterVariableDataProvider implements IJupyterVariableDataProvider
...
                     normalizedType = ColumnType.String;
             }
-            return {
+            const col = {
                 key: column.key,
                 type: normalizedType
-            };
+            } as { key: string; type: ColumnType; describe?: string };
+            if (column.describe) {
+                col['describe'] = column.describe;
```
rchiodo, line 57: What happens if this column isn't here? Do you need a default value for this?

src/client/datascience/interactive-common/interactiveBase.ts
```diff
@@ -329,6 +331,10 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
                 this.handleMessage(message, payload, this.showDataViewer);
                 break;
 
+            case InteractiveWindowMessages.ShowDataWrangler:
+                this.handleMessage(message, payload, this.showDataWrangler);
+                break;
+
             case InteractiveWindowMessages.GetVariablesRequest:
                 this.handleMessage(message, payload, this.requestVariables);
                 break;
```
rchiodo, line 340: I was thinking the requestTmLanguage here would be in the base class. So you don't need to handle the message here.

rchiodo: the ```super.onMessage``` would handle it instead.

rchiodo (approved):
Just some minor comments about adding comments on all new classes and maybe not needing to handle the monaco messages in the sub classes.

But looks good to me. Feel free to submit.

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
vandyliu: Not exactly sure what you meant. Do you mean if describe doesn't exist? If that's the case, it can be undefined, I could also make it the empty string

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
vandyliu: yep

src/client/datascience/data-viewing/jupyterVariableDataProvider.ts
rchiodo: Yeah ```if (column.describe)``` is undefined, so ```col['describe']``` never gets created. Will that break something else later?

vandyliu: I think it's okay because it is potentially undefined from the type definition. I can play it safe and give it an empty string though.

src/client/datascience/interactive-common/interactiveBase.ts
vandyliu: WebviewPanelHost or WebViewHost would be better?
WebviewPanelHost extends WebViewHost
WebViewHost also has 4 other onMessages already: SharedMessages.Started, CssMessages.GetCssRequest, CssMessages.GetMonacoThemeRequest, and InteractiveWindowMessages.GetHTMLByIdResponse

Leaning towards WebViewHost but I don't have enough knowledge about it

rchiodo: I think the lower you can push this stuff the better.

rchiodo (approved):

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,821 @@
...
+// Runs the transformations, communicates with the data wrangler UI through onMessage and postMessage
+@injectable()
+export class DataWrangler extends DataViewer implements IDataWrangler, IDisposable {
+    private existingDisposable: Disposable | undefined;
+    private historyList: IHistoryItem[] = [];
+    private sourceFile: string | undefined;
+    private commands = new Map<
+        DataWranglerCommands,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (args: any, currentVariableName: string) => Promise<IHistoryItem | void>
```
pwang347, line 69: nit: if you want typed arguments you could possibly do something like

```
export type DataWranglerCommandArgs<T extends DataWranglerCommands> = 
  T extends DataWranglerCommands.Describe ?
  IDescribeArgs :
  T extends DataWranglerCommands.Drop ?
  IDropArgs :
  any;

private commands = new Map<
        DataWranglerCommands,
        <T>(args: DataWranglerCommandArgs<T>, currentVariableName: string) => Promise<IHistoryItem | void>
```

another option would be to add the command to the args interface as well, and have a type that is like
```
export type DataWranglerCommandArgs = IDescribeArgs | IDropArgs;
```

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,821 @@
...
+                    DataWranglerMessages.SetSidePanels,
+                    wantedPanels as SidePanelSections[]
+                ).ignoreErrors();
+            }
+
+            // Use Data Viewer logic to show initial data
+            const dataFrameInfo = await this.showInitialData(title);
+            this.sourceFile = dataFrameInfo.sourceFile;
+
+            this.historyList.push({
```
pwang347, line 162: nit: should we use the `addToHistory` helper here?

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,821 @@
...
+                description: DataScience.dataWranglerImportDescription(),
+                code: `import pandas as pd\r\ndf = pd.read_csv(r'${this.sourceFile ?? ''}')\n`,
```
pwang347, line 164: nit: can we have a helper for multiline codegen e.g. something like
```
createCode([
  "import pandas as pd",
  `df = pd.read_csv(r'${this.sourceFile ?? '''}')`
], { lineEnding: <override default?> })
```

pwang347: I think we should probably default to the system line ending, ie from `TextDocument.eol`

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,821 @@
...
+        this.commands.set(DataWranglerCommands.CoerceColumn, this.coerceColumn.bind(this));
+        this.commands.set(DataWranglerCommands.ReplaceAllColumn, this.replaceAllColumn.bind(this));
+        this.commands.set(DataWranglerCommands.RemoveHistoryItem, this.removeHistoryItem.bind(this));
+        this.commands.set(DataWranglerCommands.ExportToCsv, this.exportToCsv.bind(this));
+        this.commands.set(DataWranglerCommands.RespondToPreview, this.respondToPreview.bind(this));
+
+        this.onDidDispose(this.dataWranglerDisposed, this);
+    }
+
+    public async showData(
```
pwang347, line 138: can this be merged with `showInitialData`?

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,821 @@
...
+                handled = true;
+                break;
+
+            case DataWranglerMessages.RefreshDataWrangler:
+                this.refreshData().ignoreErrors();
+                handled = true;
+                break;
+
+            case CssMessages.GetMonacoThemeRequest:
+                void this.handleMonacoThemeRequest(payload);
```
pwang347, line 241: is the void here a typo?

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,821 @@
...
+        }
+    }
+
+    private addToHistory(newHistoryItem: IHistoryItem) {
+        this.historyList.push(newHistoryItem);
+        this.postMessage(DataWranglerMessages.UpdateHistoryList, this.historyList).ignoreErrors();
+    }
+
+    private getCode() {
+        return this.historyList.map((item) => item.code).join('\n');
```
pwang347, line 261: same nit, maybe use the document EOL

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,821 @@
...
+
+    private async exportToCsv(_req: undefined, currentVariableName: string) {
+        const notebook = (this.dataProvider as IJupyterVariableDataProvider).notebook;
+        const fileInfo = await this.applicationShell.showSaveDialog({
+            saveLabel: DataScience.dataWranglerSaveCsv(),
+            filters: { CSV: ['csv'] }
+        });
+        if (fileInfo) {
+            const code = `${currentVariableName}.to_csv(path_or_buf=r'${fileInfo.fsPath}', index=False)`;
+            await notebook?.execute(code, '', 0, uuid(), undefined, false);
```
pwang347, line 272: I would recommend writing a helper over this function, since you really only care about the `code` part as a parameter

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,821 @@
...
+    private async handleCommand(payload: { command: DataWranglerCommands; args: any }) {
+        const notebook = (this.dataProvider as IJupyterVariableDataProvider).notebook;
+        let codeToRun;
+        const currentVariableName = (await this.dataFrameInfoPromise)!.name ?? '';
+        let newVariableName = currentVariableName ?? '';
+        let historyItem: IHistoryItem | void;
+
+        // Get and run data wrangler command
+        const cmd = this.commands.get(payload.command);
+        if (cmd) {
```
pwang347, line 305: nit: maybe do
```
if (!cmd) {
  return;
}

...
```

so we know there's nothing left to check for that case below

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,821 @@
...
+            historyItem = await cmd(payload.args, currentVariableName);
+            if (historyItem !== undefined) {
+                codeToRun = historyItem.isPreview ? historyItem.previewCode : historyItem.code;
```
pwang347, line 308: nit: might be good to add a small comment here

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,821 @@
...
+        const dataCleanCode = this.getCode();
+        const notebookEditor = await this.commandManager.executeCommand(Commands.CreateNewNotebook);
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const blankCell = (notebookEditor as any).document.cellAt(0) as NotebookCell;
+        await updateCellCode(blankCell, dataCleanCode);
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private async handleCommand(payload: { command: DataWranglerCommands; args: any }) {
+        const notebook = (this.dataProvider as IJupyterVariableDataProvider).notebook;
```
pwang347, line 297: curious, can we just store this value once and make sure it's not undefined? do we need to keep getting references and checking for undefined?

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
```diff
@@ -0,0 +1,821 @@
...
+            if (this.existingDisposable) {
+                this.existingDisposable.dispose();
+            }
+            if (newVariableName) {
+                await this.updateWithNewVariable(newVariableName);
+            }
+        }
+
+        if (historyItem) {
+            if (historyItem?.shouldAdd) {
```
pwang347, line 325: nit: we don't need these `?`s after the check above

pwang347 (commented):
will continue review in a bit

src/client/datascience/data-viewing/data-wrangler/dataWrangler.ts
vandyliu: I believe it changes when `updateWithNewVariable` is called

Issue #6034, opened 2021-05-28
Ship require.js with our renderers & kernel pre-loads


DonJayamanne:
https://github.com/microsoft/vscode/pull/124864
Today we ship require.js with our preloads for CodeSpaces, we need to ship it always and also include it into our renderers

IanMatthewHuff:
Picking this up. From the linked PR I should be able to test by removing the if statement from vscode core and making sure that things still work.

IanMatthewHuff:
Getting into this now. It's a bit odd, I commented out the ipynb workaround in vscode core before making changes. My assumption was that something would have broken at that point using our current bits, but all the basics seem to be working fine.

DavidKutu:
validated requirejs is preloaded

PR #6318, opened 2021-06-17
Add require to preload

IanMatthewHuff:
For #6034 

### DIFF
--- a/news/2 Fixes/6034.md
+++ b/news/2 Fixes/6034.md
@@ -1,0 +1,1 @@
+Ship require.js with our notebook preloads and renderers.
--- a/src/client/datascience/notebook/vscodeNotebookController.ts
+++ b/src/client/datascience/notebook/vscodeNotebookController.ts
@@ -4,7 +4,6 @@
 import { join } from 'path';
 import {
     Disposable,
-    env,
     EventEmitter,
     ExtensionMode,
     languages,
@@ -15,7 +14,6 @@ import {
     NotebookDocument,
     NotebookEditor,
     NotebookRendererScript,
-    UIKind,
     Uri
 } from 'vscode';
 import { ICommandManager, IVSCodeNotebook, IWorkspaceService } from '../../common/application/types';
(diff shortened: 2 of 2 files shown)


IanMatthewHuff:
Per this issue here we need this in our preloads and renderers:
https://github.com/microsoft/vscode-jupyter/issues/6034

To test this I built VS Code with the special casing to still add the loader for ipynb removed. I tested without any changes first and simple widget rendering broke trying to load the widget (as expected). Making this change here fixes that.

I wanted to loop in @DonJayamanne with a draft here as I don't consider myself a JS module loading expert. Even with no changes and the loader removed I didn't see any issues with our renderers. Per the issue here require.js was needed to support amd module loading: https://github.com/microsoft/vscode/issues/121245. However my change here (and the change in Jupyter) https://github.com/microsoft/vscode-notebook-renderers/commit/f245b5a7dee2ed61b8685b0c476ae959028fc009 moved the renderers over to loading as a Javascript module. So I don't believe that we need a change for the renderers here, but I wanted to check my understanding there. My example test file with plotly output worked fine before and after removing the loader code from vscode core.

Note: I'm prepping a PR to remove the Jupyter special case for VS Code core using the initial change from Matt.

DavidKutu (approved):

DonJayamanne (approved):

rchiodo:
@IanMatthewHuff is there a wait on this? I wanted to try it to see if allowed the 'notebook' renderer type for plotly to work again. See bug #6334

IanMatthewHuff:
@rchiodo Sorry just caught in the four day weekend. I need to check quick to see if the VS Code changes need to happen at the same time or not. Let me try it quick with current main. If it works there I'll push right away.
https://github.com/microsoft/vscode/pull/126939

IanMatthewHuff:
@rchiodo I did a merge, not sure about it fixing the issue though. This just replaces vscode's loader with our loader which should be the same I think?

rchiodo:
I'll try it out when the build comes out. The JS for plotly is failing because it's looking for the global 'require.undef'. I think adding require.js to our preloads should add that. It would explain why it works in the old webview notebook editor, it explicitly added require.js.

Issue #6215, opened 2021-06-08
Ensure user's ipykernel supports debugging


rchiodo:
Not all versions of ipykernel support debugging. 

This would be something to ask the user to upgrade.

IanMatthewHuff:
Validated per our iteration plan testing.

Issue #6207, opened 2021-06-08
Spike on using DAP for ipython


rchiodo:
Have to figure out how we talk to ipython directly instead of just using debugpy to inject a debugger.

Can we do this through the Jupyter REST api? (meaning remote)
Can we do this in raw kernel or is jupyter required?

https://github.com/microsoft/vscode-simple-jupyter-notebook

IanMatthewHuff:
Validated per iteration test plan.

PR #6332, opened 2021-06-18
Debugging (run by line) spike

DavidKutu:
For #6207
For #6215
For #6376


src/client/debugger/jupyter/debuggingCellMap.ts
```diff
@@ -0,0 +1,41 @@
+import { injectable } from 'inversify';
+import { NotebookCell, NotebookDocument, Uri } from 'vscode';
+import { noop } from '../../common/utils/misc';
+import { ICell, IDebuggingCellMap, INotebookExecutionLogger } from '../../datascience/types';
+
+@injectable()
+export class DebuggingCellMap implements IDebuggingCellMap, INotebookExecutionLogger {
```
IanMatthewHuff, line 7: Just a draft here, but this class could use some class comments about what it's responsible for.

src/client/datascience/types.ts
```diff
@@ -272,6 +273,7 @@ export const INotebookExecutionLogger = Symbol('INotebookExecutionLogger');
 export interface INotebookExecutionLogger extends IDisposable {
     preExecute(cell: ICell, silent: boolean): Promise<void>;
     postExecute(cell: ICell, silent: boolean, language: string, resource: Uri): Promise<void>;
+    nativePostExecute(cell: NotebookCell): Promise<void>;
```
DonJayamanne, line 276: Can we make this an optional method, this way you don't need to modify all other classes.


```suggestion
    nativePostExecute?(cell: NotebookCell): Promise<void>;
```

src/client/datascience/baseJupyterSession.ts
```diff
@@ -63,6 +63,12 @@ export abstract class BaseJupyterSession implements IJupyterSession {
         }
         return this.onStatusChangedEvent.event;
     }
+    public get onIOPubMessageSignal(): Event<KernelMessage.IIOPubMessage> {
```
IanMatthewHuff, line 66: Might want to remove the "Signal" from the name here, it's only a Signal down in the Juypter code doesn't really make sense to have that on the name here.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    public constructor(
+        @inject(IKernelProvider) private kernelProvider: IKernelProvider,
+        @inject(INotebookProvider) private notebookProvider: INotebookProvider,
+        @inject(IDebuggingCellMap) private debuggingCellMap: IDebuggingCellMap,
+        @inject(INotebookControllerManager) private readonly notebookControllerManager: INotebookControllerManager
+    ) {}
+
+    public async activate() {
+        vscode.debug.breakpoints; // start to fetch breakpoints
```
IanMatthewHuff, line 73: In the API this looks like a non-async array, why are the results just ignored here?

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -0,0 +1,213 @@
...
+
+        this.disposables.push(
```
IanMatthewHuff, line 75: I don't see the dispose pattern or interface being implemented to go along with this. Probably just IDisposableRegistry.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -0,0 +1,213 @@
...
+                if (dbg) {
+                    await dbg.stop();
+                }
+                this.fixBreakpoints(document);
+            }),
+
+            // factory for kernel debug adapters
+            vscode.debug.registerDebugAdapterDescriptorFactory('kernel', {
+                createDebugAdapterDescriptor: async (session) => {
+                    const activeDoc = vscode.window.activeNotebookEditor!.document;
```
IanMatthewHuff, line 100: Using the ! here is breaking type safety, activeDoc should just be able to handle undefined or document cases.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -0,0 +1,213 @@
...
+                            identity: debug.document.uri,
+                            getOnly: true
+                        });
+                        if (notebook && notebook.session) {
+                            debug.resolve(session);
+                            return new vscode.DebugAdapterInlineImplementation(
+                                new KernelDebugAdapter(session, debug.document, notebook.session, this.debuggingCellMap)
+                            );
+                        } else {
+                            void vscode.window.showInformationMessage('run the kernel');
```
IanMatthewHuff, line 117: Might be good to mark spots like this (assuming this is a temp error message) with TODOs so you can add localized messages before final PRs.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -0,0 +1,213 @@
...
+                    void vscode.window.showErrorMessage('No active notebook document to debug');
+                }
+            })
+        );
+    }
+
+    private fixBreakpoints(doc: vscode.NotebookDocument) {
+        const map = new Map<string, vscode.Uri>();
+
+        doc.getCells().forEach((cell, ix) => {
```
IanMatthewHuff, line 139: Minor suggestion, but ix feels like a funny abbreviation. I'd just use index. Makes the loop much more readable to me.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -0,0 +1,213 @@
...
+                if (editor) {
+                    void this.startDebugging(editor.document);
+                } else {
+                    void vscode.window.showErrorMessage('No active notebook document to debug');
+                }
+            })
+        );
+    }
+
+    private fixBreakpoints(doc: vscode.NotebookDocument) {
```
IanMatthewHuff, line 136: More of these can come later, but this is the type of function that could use a function comment, even at the draft stage. It's easier to parse the code if you know what the intent of the function is. And FixBreakpoints isn't super clear (fix from what? are they broken?)

src/client/debugger/jupyter/debuggingManager.ts
IanMatthewHuff: Also we should use our specific messaging classes.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -0,0 +1,213 @@
...
+        for (const [doc, dbg] of this.notebookToDebugger.entries()) {
+            if (document.uri.toString() === doc.uri.toString()) {
+                return dbg;
+            }
+        }
+    }
+
+    private async ensureKernelIsRunning(doc: vscode.NotebookDocument): Promise<void> {
+        await this.notebookControllerManager.loadNotebookControllers();
+        const controller = this.notebookControllerManager.getSelectedNotebookController(doc);
```
IanMatthewHuff, line 200: You should take a quick peek at Joyce's current review / Don's recent reviews. I believe this has been removed, so might need to be re-added or this changed.

src/client/common/installer/productInstaller.ts
```diff
@@ -126,6 +126,54 @@ export abstract class BaseInstaller {
...
+        // User may have customized the module name or provided the fully qualified path.
+        const interpreter = isResource(resource) ? undefined : resource;
+        const uri = isResource(resource) ? resource : undefined;
+        const executableName = this.getExecutableNameFromSettings(product, uri);
+
+        const pythonProcess = await this.serviceContainer
+            .get<IPythonExecutionFactory>(IPythonExecutionFactory)
+            .createActivatedEnvironment({ resource: uri, interpreter, allowEnvironmentFetchExceptions: true });
+
+        const output = pythonProcess.execObservable(['-m', 'pip', 'show', executableName], {});
```
rchiodo, line 139: Does this work for conda? Meaning will running pip show give you the correct information in a conda environment?

src/client/datascience/jupyter/kernels/kernelDependencyService.ts
```diff
@@ -80,6 +80,21 @@ export class KernelDependencyService implements IKernelDependencyService {
     public areDependenciesInstalled(interpreter: PythonEnvironment, _token?: CancellationToken): Promise<boolean> {
         return this.installer.isInstalled(Product.ipykernel, interpreter).then((installed) => installed === true);
     }
+
+    // The requirement for debugging is ipykernel v6 or newer
+    public areDebuggingDependenciesInstalled(
+        interpreter: PythonEnvironment,
+        _token?: CancellationToken
+    ): Promise<boolean> {
+        return this.installer.getVersion(Product.ipykernel, interpreter).then((version) => {
```
rchiodo, line 89: Would async/await be more readable here? You forgot a 'catch' clause I think. Might be easier with try, await getVersion, parse versoin, catch - return false

src/client/debugger/jupyter/debuggingManager.ts
rchiodo: This seems really weird to do on extension activation? Fetch breakpoints?

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -0,0 +1,213 @@
...
+                        } else {
+                            void vscode.window.showInformationMessage('run the kernel');
+                        }
+                    }
+                    // Should not happen, debug sessions should start only from the cell toolbar command
+                    return;
+                }
+            }),
+
+            vscode.commands.registerCommand('jupyter.debugNotebook', () => {
```
rchiodo, line 125: I think we have commands in a 'commands.ts' file? So that this string is one spot.

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+
+        // after attaching, send a 'debugInfo' request
+        // reset breakpoints and continue stopped threads if there are any
+        // we do this in case the kernel is stopped when we attach
+        // This might happen if VS Code or the extension host crashes
+        if (message.type === 'request' && (message as DebugProtocol.Request).command === 'attach') {
+            await this.debugInfo();
+        }
+
+        // after disconnecting, hide the breakpoint margin
```
rchiodo, line 106: Can you elaborate a bit in this comment? I assume detaching is the same as hitting the 'debug' button on the toolbar. Can the user also cause this by just hitting the stop button during debugging? 

Seems weird to me. Stopping during debugging should just interrrupt the cell I think.

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+                control.onIOPub = (msg) => {
+                    this.getMessageSourceAndHookIt(msg.content as DebugProtocol.ProtocolMessage, (source) => {
+                        if (source && source.path) {
+                            const cell = this.fileToCell.get(source.path);
+                            if (cell) {
+                                source.name = path.basename(cell.document.uri.path);
+                                if (cell.index >= 0) {
+                                    source.name += `, Cell ${cell.index + 1}`;
+                                }
+                                source.path = cell.document.uri.toString();
```
rchiodo, line 156: I think this is common with the code above? Meaning can you make a separate function or lambda?

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    /**
+     * Dump content of given cell into a tmp file and return path to file.
+     */
+    private async dumpCell(uri: string): Promise<void> {
+        const cell = this.notebookDocument.getCells().find((c) => c.document.uri.toString() === uri);
+        if (cell) {
+            try {
+                const response = await this.session.customRequest('dumpCell', { code: cell.document.getText() });
```
rchiodo, line 213: I guess this means the kernel has to support this request. I guess it's okay cause ipykernel is probably the only reasonable case for debugging, but maybe we should flag this somehow so we don't forget it when we try getting debugging working elsewhere.

src/client/common/installer/productInstaller.ts
```diff
@@ -126,6 +126,54 @@ export abstract class BaseInstaller {
             .install(product, resource, cancel, reInstallAndUpdate);
     }
 
+    public async getVersion(product: Product, resource?: InterpreterUri): Promise<string | undefined> {
```
IanMatthewHuff, line 129: Did using the pythonAPI function not work here?

src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -173,6 +180,31 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
         }
         this.updateContextOfActiveNotebookKernel(activeEditor);
     }
+    private async updateDebugContext(activeEditor?: INotebookEditor, interpreter?: PythonEnvironment) {
```
IanMatthewHuff, line 183: Where is the second parameter case (interpreter) coming from? The only usage of this looks to just be passed an editor.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -0,0 +1,228 @@
...
+                void vscode.commands.executeCommand('notebook.toggleBreakpointMargin', doc);
+            } catch (err) {
+                void vscode.window.showErrorMessage(`Can't start debugging (${err})`);
+            }
+        }
+    }
+
+    private getDebuggerByUri(document: vscode.NotebookDocument): Debugger | undefined {
+        for (const [doc, dbg] of this.notebookToDebugger.entries()) {
+            if (document.uri.toString() === doc.uri.toString()) {
```
IanMatthewHuff, line 207: Per-discussion here: https://github.com/microsoft/vscode-jupyter/pull/6358#discussion_r657594246 I believe we can just compare on documents.

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+const debugRequest = (message: DebugProtocol.Request): KernelMessage.IDebugRequestMsg => {
+    return {
+        channel: 'control',
+        header: {
+            msg_id: randomBytes(8).toString('hex'),
+            date: new Date().toISOString(),
+            version: '5.2',
+            msg_type: 'debug_request',
+            username: 'vscode',
+            session: randomBytes(8).toString('hex')
```
IanMatthewHuff, line 22: This might work for the spike, but for JMP messages session should be unique per session. It's not random per message. You can get the id of an existing session off the ISession on the IJupyterSession.

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+                                    source.name += `, Cell ${cell.index + 1}`;
+                                }
+                                source.path = cell.document.uri.toString();
+                            }
+                        }
+                    });
+
+                    this.sendMessage.fire(msg.content);
+                };
+                control.onStdin = (msg) => {
```
IanMatthewHuff, line 163: Does anything actually come in on stdin?

package.json
```diff
@@ -1954,7 +1966,13 @@
                     "when": "jupyter.hasNativeNotebookOpen"
                 }
             ]
-        }
+        },
+        "debuggers": [
+            {
+                "type": "kernel",
+                "label": "Kernel Debug"
```
DonJayamanne, line 1973: II think this should be called `Jupyter Python` or `Python Kernel` or similar, as it's specific to python not just any kernel.
This will matter in full debugging, user needs to select the right debug config (based on the label)

src/client/common/installer/productInstaller.ts
DonJayamanne: Why not use `-m ipython --version` or the existing api as mentioned by ian

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+    private async dumpCellsThatRanBeforeDebuggingBegan() {
+        this.cellMap.getCellsAnClearQueue(this.notebookDocument).forEach(async (cell) => {
+            await this.dumpCell(cell.document.uri.toString());
+        });
+    }
+
+    /**
+     * Dump content of given cell into a tmp file and return path to file.
+     */
+    private async dumpCell(uri: string): Promise<void> {
```
IanMatthewHuff, line 209: Who has ownership of the file here? Looks like ipykernel returns back a sourcePath for the file created in a success message. But I'm not seeing cleanup on the ipykernel side.

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+                console.log(err);
+            }
+        }
+    }
+
+    private async debugInfo(): Promise<void> {
+        const response = await this.session.customRequest('debugInfo');
+
+        // If there's breakpoints at this point, delete them
+        if (response.breakpoints.lenght > 0) {
```
IanMatthewHuff, line 227: This looks misspelled, "length." Sorry in a PR I can't see the type easily is it an any? Not sure why this would build.

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+            this.jupyterSession.requestDebug({
+                seq: 0,
+                type: 'request',
+                command: 'setBreakpoints',
+                arguments: {
+                    source: {
+                        path: response.breakpoints[0].source
```
IanMatthewHuff, line 234: Will they all have the same source here?

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: updated on the cleanup commit

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: cleanup commit

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: cleanup commit

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: cleanup commit

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: this still works :)

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: cleanup commit

src/client/datascience/jupyter/kernels/kernelDependencyService.ts
DavidKutu: cleanup commit
although, I will check if we can accomplish this with the python api.

There's the issue that `python -m ipykernel --version` returns the ipython version instead of the ipykernel version. Hopefully the python api deals with that.

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: normally nothing needs to happen here. But since vscode has a command that shows and hides the breakpoint margin for notebooks, we need to hide it at some point. And the best time to do it is on the disconnect command.

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: cleanup commit

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: What do you mean by flagging? With a comment?

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: cleanup commit

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: yeah, nothing, I'll remove it

package.json
DavidKutu: I changed it to Python Kernel on the cleanup commit

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: cleanup commit
I created types for these responses to avoid that

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: cleanup commit
Small change, but instead of deleting the breakpoints on the kernel, I instead sent a message to VS Code to register them.

src/client/common/installer/productInstaller.ts
DavidKutu: This is the reason:
image

I think the python api uses the bottom command, and it gives the version of something different (its ipython) and that wouldn't work. I'll check their code, but for this PR we might want to do it ourselves. 

And fix it for Conda of course.

src/client/common/installer/productInstaller.ts
DavidKutu: This is the reason:
image

I think the python api uses the bottom command, and it gives the version of something different (its ipython) and that wouldn't work. I'll check their code, but for this PR we might want to do it ourselves. 

And fix it for Conda of course.

DavidKutu:
# To do

- Add context for switching kernels and notebook opened
- cache valid kernels so the button enables faster
- don't make up a random session ids
- delete temp files (the kernel is not doing this)
- use python api for checking ipykernel version
- use IDisposableRegistry
- check if we need the vscode.debug.breakpoints on activate

src/client/common/installer/productInstaller.ts
IanMatthewHuff: I'm not a python expert (could consider asking the python folks). But an easier way if we can't use the python -m version check or python api then might be to activate the environment and do like this:

python -c "import ipykernel; print(ipykernel.__version__)"

DavidKutu: I'll try it, although it would be nice to not have to activate it

src/client/common/installer/productInstaller.ts
joyceerhl: For modules (which ipykernel should be treated as), I believe the Python API uses 
```
python -c 'import ipykernel; print(ipykernel.__version__)'
```

which would yield the right result

image

https://github.com/microsoft/vscode-python/blob/362eb34cafcbc513919eb4d78ea0521a0ed6738a/src/client/common/process/internal/python.ts#L104-L112

src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -173,6 +179,31 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
...
+        this.canDebug.set(false).ignoreErrors();
+
+        if (activeEditor) {
+            this.notebookProvider
+                .getOrCreateNotebook({ identity: activeEditor.file, resource: activeEditor.file, getOnly: true })
+                .then(async (nb) => {
```
rchiodo, line 188: why not just an await here instead of chaining stuff?

src/client/debugger/jupyter/kernelDebugAdapter.ts
rchiodo: Yeah with a comment here and maybe an issue? ipykernel supports 'dumpcell' but does xeus?

rchiodo (approved):

DavidKutu:
@rchiodo for your second comment, I added this at the top of the file
https://github.com/microsoft/vscode-jupyter/blob/e3bdfd1bc5260004b668db2bcc338cc193b81eb4/src/client/debugger/jupyter/kernelDebugAdapter.ts#L89

I put it there because the whole debug adapter need support for those 2 special messages.

PR #6342, opened 2021-06-21
Remove `setController` Method

DonJayamanne:
Something that came out of another PR,

### DIFF
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -177,7 +177,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
     }
 
     // When a document is opened we need to look for a perferred kernel for it
-    private onDidOpenNotebookDocument(document: NotebookDocument) {
+    private async onDidOpenNotebookDocument(document: NotebookDocument) {
         // Restrict to only our notebook documents
         if (document.notebookType !== JupyterNotebookView || !this.workspace.isTrusted) {
             return;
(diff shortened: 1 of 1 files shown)


src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -187,53 +187,43 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
-        } else {
-            this.setController(document, preferredSearchToken.token).finally(() => {
-                // Make sure that we clear our finding in progress when done
-                this.findPreferredInProgress.delete(document);
-            });
-        }
-    }
-
-    // Set the controller for this notebook document
-    private async setController(document: NotebookDocument, cancelToken: CancellationToken) {
```
DonJayamanne, line 211: This method doesn't set the controller at all, if anything it finds the preferred controller and sets the preferred controller.
We also have an existing method named `setPreferredController`.
Also, it wasn't clear to me my we were clearing ` this.controllerMapping.delete(document);` in line `214`.

I felt that it makes more sense to have this code inline in the event handler. 
To me the code reads better, but if others disagree please reject this PR

**Again, this is NOT a required PR, just a simple codehealth/refactor.**

rchiodo: Looks good to me. Should probably wait for Ian as he wrote this originally.

IanMatthewHuff: The main reason that this function was structured as it is, is that I believe we discussed wanted to keep the onDidOpen / onDidClose sync and not have awaits in them. However I can't track down that initial discussion on that. If aync on the doc functions is not a problem I'd prefer this as well.

DonJayamanne: Looks like we're ok, please can we get this approved.

IanMatthewHuff (approved):

Issue #5861, opened 2021-05-12
Return a default controller as fast as possible to enable run button on native notebooks


joyceerhl:
Background
- Kernel push currently doesn't show run button until a controller is available for the notebook
- 90% of users experience kernel search time taking 5 seconds or less
- Our goal is to allow notebooks to load ASAP and allow users to run code ASAP
- Listing interpreters is the slow part of kernel search at the moment

Ideas for making first-time controller search as fast as possible
- Generate a dummy preferred controller and hotswap it out for the real preferred kernel when the user attempts to execute code 
	- If we can't find a kernel that matches what the notebook metadata says, we should ask the user if the one we found instead is OK. Classic Jupyter doesn't prompt--it picks randomly
	- If Python extension has an active interpreter and notebook metadata only says 'Python 3', use that one
	- Scenario: notebook metadata says 'Python 3' --> when we find an interpreter e.g. Python 3.9, the kernel picker will change from 'Python 3' to 'Python 3.9'
	- We only put up our dummy kernel when there aren't any other extensions which might be able to provide a better kernel, e.g. we won't show our dummy kernel for .NET interactive notebook
- Generate a partial list of kernels vs doing a full search

For subsequent activation we can also cache interpreters/kernelspecs (we currently don't do any caching)
- However there are potential issues with when we invalidate the cache e.g. for users whose cache is months old

Other ideas
- Get Python extension to move all its interpreter discovery into an npm package
	- Assumption here is that Python extension activation is the bottleneck
        - But we need the active interpreter which relies on VS Code settings so npm package might not work either

DonJayamanne:
Note: This is also imporatnt for getting started scneario:
* User install Juptyer extension for first time
* User opens a notebook & immediately goes into kernel pcker
* At this point we should have added the `Python` kernel so they can see it (instead of not having anything)

DonJayamanne:
Blocked here https://github.com/microsoft/vscode/issues/124535

DonJayamanne:
* (5) Add a placeholder Python Kernel
    * Whether we have Python or not
    * If no python, then running cells will prompt users to install Python
    * Once Python is installed, we can update this to point to a real kernel (see further below)
* (1) Fetch non-python kernels first & list them as controllers
    * (2) Fetch preferred kernel from this list (for non-Python kernels)
* (3) Fetch Python kernels
    * (4) Fetch preferred kernel from this list (for non-Python kernels)
[...]

PR #6345, opened 2021-06-21
Split python vs non-python kernel search

DonJayamanne:
Part of #5861
This PR merely splits the search between Python vs non_python.
Functionally there are no changes, more of a refactor (preparation for more work).

Need to test this and ensure things are ok, else the PR seems to be ready for review.

### DIFF
--- a/src/client/datascience/jupyter/kernels/helpers.ts
+++ b/src/client/datascience/jupyter/kernels/helpers.ts
@@ -37,6 +37,7 @@ import { Uri } from 'vscode';
 import { getResourceType } from '../../common';
 import { IPythonExecutionFactory } from '../../../common/process/types';
 import { SysInfoReason } from '../../interactive-common/interactiveWindowTypes';
+import { isDefaultPythonKernelSpecName } from '../../kernel-launcher/localPythonkernelFinder';
 
 // Helper functions for dealing with kernels and kernelspecs
 
@@ -364,7 +365,7 @@ export function findPreferredKernel(
     traceInfo(
         `Find preferred kernel for ${resource?.toString()} with metadata ${JSON.stringify(
             notebookMetadata || {}
-        )} & preferred interpreter ${preferredInterpreter || {}}`
+        )} & preferred interpreter ${JSON.stringify(preferredInterpreter || {})}`
     );
     let index = -1;
 
@@ -493,6 +494,18 @@ export function findPreferredKernel(
 
                 // See if the display name already matches.
                 if (spec.display_name && spec.display_name === notebookMetadata?.kernelspec?.display_name) {
+                    score += 16;
+                }
+                // See if the name of the environments match (kernel name == environment name).
+                // At this point we dont care about version numbers of the Python environments.
+                // E.g. assume user opens notebook with metadata pointing to kernelspec with the name `condaPytoch`,
+                // & the user has such an environment (with the same name), then its a match.
+                if (
+                    metadata.interpreter?.envName &&
+                    metadata.interpreter?.envName === notebookMetadata?.kernelspec?.name &&
+                    nbMetadataLanguage === PYTHON_LANGUAGE &&
+                    !notebookMetadata?.kernelspec?.name.toLowerCase().match(isDefaultPythonKernelSpecName)
+                ) {
                     score += 16;
                 }
 
--- a/src/client/datascience/jupyter/kernels/helpers.ts
+++ b/src/client/datascience/jupyter/kernels/helpers.ts
@@ -37,7 +37,7 @@ import { Uri } from 'vscode';
 import { getResourceType } from '../../common';
 import { IPythonExecutionFactory } from '../../../common/process/types';
 import { SysInfoReason } from '../../interactive-common/interactiveWindowTypes';
-import { isDefaultPythonKernelSpecName } from '../../kernel-launcher/localPythonkernelFinder';
+import { isDefaultPythonKernelSpecName } from '../../kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder';
 
 // Helper functions for dealing with kernels and kernelspecs
 
--- a/src/client/datascience/kernel-launcher/jupyterPaths.ts
+++ b/src/client/datascience/kernel-launcher/jupyterPaths.ts
@@ -29,6 +29,8 @@ export class JupyterPaths {
      */
     public async getKernelSpecRootPath(): Promise<string | undefined> {
         if (this.platformService.isWindows) {
+            // On windows the path is not correct if we combine those variables.
+            // It won't point to a path that you can actually read from.
             return tryGetRealPath(path.join(this.pathUtils.home, winJupyterPath));
         } else if (this.platformService.isMac) {
             return path.join(this.pathUtils.home, macJupyterPath);
--- a/src/client/datascience/kernel-launcher/jupyterPaths.ts
+++ b/src/client/datascience/kernel-launcher/jupyterPaths.ts
@@ -1,0 +1,98 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+'use strict';
+
+import { inject, injectable } from 'inversify';
+import * as path from 'path';
+import { CancellationToken } from 'vscode';
+import { IPlatformService } from '../../common/platform/types';
+import { IPathUtils } from '../../common/types';
+import { IEnvironmentVariablesProvider } from '../../common/variables/types';
+import { tryGetRealPath } from '../common';
+
+const winJupyterPath = path.join('AppData', 'Roaming', 'jupyter', 'kernels');
+const linuxJupyterPath = path.join('.local', 'share', 'jupyter', 'kernels');
+const macJupyterPath = path.join('Library', 'Jupyter', 'kernels');
+export const baseKernelPath = path.join('share', 'jupyter', 'kernels');
+
+@injectable()
+export class JupyterPaths {
+    constructor(
+        @inject(IPlatformService) private platformService: IPlatformService,
+        @inject(IPathUtils) private readonly pathUtils: IPathUtils,
+        @inject(IEnvironmentVariablesProvider) private readonly envVarsProvider: IEnvironmentVariablesProvider
+    ) {}
+
+    /**
+     * This should return a WRITABLE place that jupyter will look for a kernel as documented
+     * here: https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs
+     */
+    public async getKernelSpecRootPath(): Promise<string | undefined> {
+        if (this.platformService.isWindows) {
+            return tryGetRealPath(path.join(this.pathUtils.home, winJupyterPath));
+        } else if (this.platformService.isMac) {
+            return path.join(this.pathUtils.home, macJupyterPath);
+        } else {
+            return path.join(this.pathUtils.home, linuxJupyterPath);
+        }
+    }
+    /**
+     * This list comes from the docs here:
+     * https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs
+     */
+    public async getKernelSpecRootPaths(cancelToken?: CancellationToken): Promise<string[]> {
+        // Paths specified in JUPYTER_PATH are supposed to come first in searching
+        const paths: string[] = await this.getJupyterPathPaths(cancelToken);
+
+        if (this.platformService.isWindows) {
+            const winPath = await this.getKernelSpecRootPath();
+            if (winPath) {
+                paths.push(winPath);
+            }
+
+            if (process.env.ALLUSERSPROFILE) {
+                paths.push(path.join(process.env.ALLUSERSPROFILE, 'jupyter', 'kernels'));
+            }
+        } else {
+            // Unix based
+            const secondPart = this.platformService.isMac ? macJupyterPath : linuxJupyterPath;
+
+            paths.push(
+                path.join('/', 'usr', 'share', 'jupyter', 'kernels'),
+                path.join('/', 'usr', 'local', 'share', 'jupyter', 'kernels'),
+                path.join(this.pathUtils.home, secondPart)
+            );
+        }
+
+        return paths;
+    }
+
+    /**
+     * Find any paths associated with the JUPYTER_PATH env var. Can be a list of dirs.
+     * We need to look at the 'kernels' sub-directory and these paths are supposed to come first in the searching
+     * https://jupyter.readthedocs.io/en/latest/projects/jupyter-directories.html#envvar-JUPYTER_PATH
+     */
+    private async getJupyterPathPaths(cancelToken?: CancellationToken): Promise<string[]> {
+        const paths: string[] = [];
+        const vars = await this.envVarsProvider.getEnvironmentVariables();
+        if (cancelToken?.isCancellationRequested) {
+            return [];
+        }
+        const jupyterPathVars = vars.JUPYTER_PATH
+            ? vars.JUPYTER_PATH.split(path.delimiter).map((jupyterPath) => {
+                  return path.join(jupyterPath, 'kernels');
+              })
+            : [];
+
+        if (jupyterPathVars.length > 0) {
+            jupyterPathVars.forEach(async (jupyterPath) => {
+                const realPath = await tryGetRealPath(jupyterPath);
+                if (realPath) {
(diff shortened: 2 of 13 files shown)


src/client/datascience/kernel-launcher/jupyterPaths.ts
```diff
@@ -0,0 +1,98 @@
...
+        @inject(IEnvironmentVariablesProvider) private readonly envVarsProvider: IEnvironmentVariablesProvider
+    ) {}
+
+    /**
+     * This should return a WRITABLE place that jupyter will look for a kernel as documented
+     * here: https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs
+     */
+    public async getKernelSpecRootPath(): Promise<string | undefined> {
+        if (this.platformService.isWindows) {
+            return tryGetRealPath(path.join(this.pathUtils.home, winJupyterPath));
```
DonJayamanne, line 32: I wish we knew exactly why we had to use `tryGetRealPythonpath`.
If we did, then everything in this class is a simple constant.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -191,53 +191,42 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         const preferredSearchToken = new CancellationTokenSource();
         this.findPreferredInProgress.set(document, preferredSearchToken);
 
-        if (!this.isLocalLaunch) {
```
DonJayamanne, line 194: Please ignore this code, blocked by https://github.com/microsoft/vscode-jupyter/pull/6342
If that PR isn't merged, I'll revert the changes.
The changes in this file are in https://github.com/microsoft/vscode-jupyter/pull/6342

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -147,33 +105,27 @@ export class LocalKernelFinder implements ILocalKernelFinder {
...
-                );
-            }
+            const [nonPythonKernelSpecs, pythonRelatedKernelSpecs] = await Promise.all([
+                this.nonPythonkernelFinder.listKernels(resource, cancelToken),
+                this.extensionChecker.isPythonExtensionInstalled
+                    ? this.pythonKernelFinder.listKernels(resource, cancelToken)
+                    : []
+            ]);
+
+            const kernels = [...nonPythonKernelSpecs, ...pythonRelatedKernelSpecs].filter(({ kernelSpec }) => {
```
DonJayamanne, line 115: The results from Python & Non Python are merged here.
Hence we should still get the same list of kernels.

src/client/datascience/kernel-launcher/jupyterPaths.ts
rchiodo: On windows the path is not correct if you combine those variables. It won't point to a path that you can actually read from.

src/client/datascience/kernel-launcher/jupyterPaths.ts
```diff
@@ -0,0 +1,98 @@
...
+        const paths: string[] = await this.getJupyterPathPaths(cancelToken);
+
+        if (this.platformService.isWindows) {
+            const winPath = await this.getKernelSpecRootPath();
+            if (winPath) {
+                paths.push(winPath);
+            }
+
+            if (process.env.ALLUSERSPROFILE) {
+                paths.push(path.join(process.env.ALLUSERSPROFILE, 'jupyter', 'kernels'));
```
rchiodo, line 54: I think you might need tryGetRealPath on all paths returned on windows.

src/client/datascience/kernel-launcher/jupyterPaths.ts
DonJayamanne: Thanks, will add that comment

src/client/datascience/kernel-launcher/localNonPythonKernelFinder.ts
```diff
@@ -0,0 +1,119 @@
...
+import { IFileSystem } from '../../common/platform/types';
+import { Resource } from '../../common/types';
+import { getKernelId } from '../jupyter/kernels/helpers';
+import { KernelSpecConnectionMetadata, PythonKernelConnectionMetadata } from '../jupyter/kernels/types';
+import { IJupyterKernelSpec } from '../types';
+import { LocalKernelFinderBase } from './localFinderBase';
+import { JupyterPaths } from './jupyterPaths';
+
+// This class searches for a kernel that matches the given kernel name.
+// First it searches on a global persistent state, then on the installed python interpreters,
```
IanMatthewHuff, line 17: Comment here is not right for this class.

src/client/datascience/kernel-launcher/jupyterPaths.ts
DonJayamanne: Lets do that separately

DonJayamanne: I'll file an issue for ath.

rchiodo (approved):

src/client/datascience/kernel-launcher/jupyterPaths.ts
DonJayamanne: Not sure we need to do that at all, I thought the environment variable contains the fully qualified path.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -147,33 +105,38 @@ export class LocalKernelFinder implements ILocalKernelFinder {
...
+                this.nonPythonkernelFinder.listKernels(resource, cancelToken),
+                this.extensionChecker.isPythonExtensionInstalled
+                    ? this.pythonKernelFinder.listKernels(resource, cancelToken)
+                    : []
+            ]);
+
+            // If python extension is installed, then ignore Python kernels kernels returned by kernelspec finder.
+            if (!this.extensionChecker.isPythonExtensionInstalled) {
+                nonPythonKernelSpecs = nonPythonKernelSpecs.filter(({ kernelSpec }) => {
+                    if (kernelSpec.language === PYTHON_LANGUAGE) {
```
IanMatthewHuff, line 118: I think this works, but it is possible for non-python kernel specs to be installed in paths relative to Python interpreters. For example IJava does this if you follow their default instructions. But in this case it feels ok to not list them.

src/client/datascience/kernel-launcher/jupyterPaths.ts
DonJayamanne: Added an issue here https://github.com/microsoft/vscode-jupyter/issues/6352

IanMatthewHuff (approved):
There is a case about non-python kernels installed relative to python interpreter paths. But I don't believe that is a very common case at all, might not be worth supporting.

src/test/datascience/kernel-launcher/localKernelFinder.unit.test.ts
```diff
@@ -559,6 +569,31 @@ import { IExtensions } from '../../../client/common/types';
                     });
                     assert.equal(kernel?.interpreter, activeInterpreter);
                 });
+                test('Return conda interpreter if we have conda env name as kernelspec name in notebook metadata', async () => {
```
DonJayamanne, line 572: @rchiodo @IanMatthewHuff Added a new test & fixed this.
Not sure how we missed this, but the tests prove that the current search doesn't work properly (basically the preferred interpreter is not as expected).

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -364,7 +365,7 @@ export function findPreferredKernel(
     traceInfo(
         `Find preferred kernel for ${resource?.toString()} with metadata ${JSON.stringify(
             notebookMetadata || {}
-        )} & preferred interpreter ${preferredInterpreter || {}}`
+        )} & preferred interpreter ${JSON.stringify(preferredInterpreter || {})}`
```
DonJayamanne, line 368: This wasn't getting logged properly

src/test/datascience/kernel-launcher/localKernelFinder.unit.test.ts
IanMatthewHuff: @DonJayamanne Sorry where is the fix in code? I'm not seeing it off hand.

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: I probably should rename this file/class to indicate the fact that this only searches the global paths. Probably add comments as well.
The other python finder searches for java kernels found in python interpreters like the one you are referring to.

Open to suggestions for names for the two times

IanMatthewHuff: Ahh, sorry. I was reading that wrong. You are right that the python extension finder would find them, just the name was throwing me off. Honestly I'm not sure if there is better naming. So:
LocalNonPythonKernelFinder sounds like "finds non-python kernels"
LocalPythonKernelFinder sounds like "finds python kernels" 
Which is sorta right on one level. 

Could be an option:
LocalJupyterKernelFinder => For the global paths might work better? It totally finds python kernels, the tossing out of them happens after they are returned from the finder only if python is install. Then the name of the other class could just stay the same.

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -495,6 +496,18 @@ export function findPreferredKernel(
                 if (spec.display_name && spec.display_name === notebookMetadata?.kernelspec?.display_name) {
                     score += 16;
                 }
+                // See if the name of the environments match (kernel name == environment name).
+                // At this point we dont care about version numbers of the Python environments.
+                // E.g. assume user opens notebook with metadata pointing to kernelspec with the name `condaPytoch`,
+                // & the user has such an environment (with the same name), then its a match.
+                if (
```
DonJayamanne, line 503: @IanMatthewHuff this is the code that fixes searching for kernels by name.

DonJayamanne:
@rchiodo @IanMatthewHuff  had to completely refactor this as the old implementation was incorrect.
Once its ready i'd appreciate a second set of eyes once again.
Trying to fix some tests at this stage.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -147,33 +106,25 @@ export class LocalKernelFinder implements ILocalKernelFinder {
...
-                                undefined,
-                                4
-                            )}`
-                        );
-                        return items;
-                    })
-                );
-            }
+            let [globalKernelSpecs, pythonRelatedKernelSpecs] = await Promise.all([
+                this.nonPythonkernelFinder.listKernelSpecs(false, cancelToken),
```
IanMatthewHuff, line 110: The name changes should trickle down to the local variables as well.

src/client/datascience/kernel-launcher/localKernelSpecFinderBase.ts
```diff
@@ -0,0 +1,196 @@
...
+                    if (!distinctKernelMetadata.has(kernelSpec.id)) {
+                        distinctKernelMetadata.set(kernelSpec.id, kernelSpec);
+                    }
+                });
+
+                return Array.from(distinctKernelMetadata.values()).sort((a, b) => {
+                    if (a.kernelSpec?.display_name === b.kernelSpec?.display_name) {
+                        return 0;
+                    } else {
+                        return 1;
```
IanMatthewHuff, line 75: Unsure what the intent of the sort here is, items stay in the same position if display names are equal and then b always moved down the sort regardless of contents?

src/client/datascience/kernel-launcher/localKernelSpecFinderBase.ts
```diff
@@ -0,0 +1,196 @@
...
+            interpreter?.path || kernelJson?.metadata?.interpreter?.path
+        );
+
+        // Some registered kernel specs do not have a name, in this case use the last part of the path
+        kernelSpec.name = kernelJson?.name || path.basename(path.dirname(specPath));
+
+        // Possible user deleted the underlying kernel.
+        const interpreterPath = interpreter?.path || kernelJson?.metadata?.interpreter?.path;
+        if (
+            isKernelRegisteredByUs(kernelSpec) &&
```
IanMatthewHuff, line 156: I think this was from a previous change (a bit hard to tell since there is all the new code), but why would this be gated on isKernelRegisteredByUs? If there is an interpreter path and it doesn't exist do we want the kernel spec?

src/client/datascience/kernel-launcher/localKnownPathKernelSpecFinder.ts
```diff
@@ -0,0 +1,111 @@
...
+import { IFileSystem } from '../../common/platform/types';
+import { getKernelId } from '../jupyter/kernels/helpers';
+import { KernelSpecConnectionMetadata, PythonKernelConnectionMetadata } from '../jupyter/kernels/types';
+import { IJupyterKernelSpec } from '../types';
+import { LocalKernelSpecFinderBase } from './localKernelSpecFinderBase';
+import { JupyterPaths } from './jupyterPaths';
+import { PYTHON_LANGUAGE } from '../../common/constants';
+
+/**
+ * This class searches for kernels on the file system in well known paths documented by Jupyter.
```
IanMatthewHuff, line 17: Good class comment, make the usage more clear.

src/client/datascience/kernel-launcher/localKnownPathKernelSpecFinder.ts
```diff
@@ -0,0 +1,111 @@
...
+        super(fs, workspaceService);
+    }
+    /**
+     * @param {boolean} includePythonKernels Include/exclude Python kernels in the result.
+     */
+    public async listKernelSpecs(
+        includePythonKernels: boolean,
+        cancelToken?: CancellationToken
+    ): Promise<(KernelSpecConnectionMetadata | PythonKernelConnectionMetadata)[]> {
+        return this.listKernelsWithCache(includePythonKernels ? 'IncludePython' : 'ExcludePython', async () => {
```
IanMatthewHuff, line 37: So the cache key can just be a constant here since it's the global search? That make sense.

src/client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder.ts
```diff
@@ -0,0 +1,373 @@
...
+                        // In such an instance the interpreter information is stored in the kernelspec.json file.
+                        if (
+                            k.language === PYTHON_LANGUAGE &&
+                            k.metadata?.interpreter?.path &&
+                            k.metadata?.interpreter?.path !== activeInterpreter?.path
+                        ) {
+                            interpreter = await this.interpreterService
+                                .getInterpreterDetails(k.metadata?.interpreter?.path)
+                                .catch((ex) => {
+                                    traceError(`Failed to get interpreter details for Kernel Spec ${k.specFile}`, ex);
```
IanMatthewHuff, line 167: Looks like we just proceed with the interpreter here, should we do that? I would think that if we failed to get details here then we don't have a valid interpreter, but I'm not fully sure if that's the case for getInterpreterDetails.

IanMatthewHuff (approved):

src/client/datascience/kernel-launcher/localKernelSpecFinderBase.ts
DonJayamanne: This was old code, let me fix that.

src/client/datascience/kernel-launcher/localKernelSpecFinderBase.ts
DonJayamanne: Yes, that's better, fixed.

rchiodo (approved):

src/client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder.ts
DonJayamanne: This is old code, but yes, you are right.
Its nice to see this code is getting cleaned up (i.e. minor issues being picked up & resolved).

IanMatthewHuff: @DonJayamanne Yeah, sorry for picking on old code here, I know that it's not stuff that you added. Just this review gives a chance to look at it more as you noted.

PR #6358, opened 2021-06-23
Implement variable view and debug cell for native interactive window

joyceerhl:


src/client/datascience/notebook/helpers/helpers.ts
```diff
@@ -48,9 +48,9 @@ export function isJupyterNotebook(document: NotebookDocument): boolean;
 export function isJupyterNotebook(viewType: string): boolean;
 export function isJupyterNotebook(option: NotebookDocument | string) {
     if (typeof option === 'string') {
-        return option === JupyterNotebookView;
+        return option === JupyterNotebookView || option === InteractiveWindowView;
     } else {
-        return option.notebookType === JupyterNotebookView;
+        return option.notebookType === JupyterNotebookView || option.notebookType === InteractiveWindowView;
```
joyceerhl, line 53: The changes in this file combined with the package.json interactive/toolbar contribution allow us to reuse literally all our existing variable view code for the interactive window.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -474,18 +484,20 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
+        line: number,
+        id?: string,
+        data?: nbformat.ICodeCell | nbformat.IRawCell | nbformat.IMarkdownCell,
+        debugInfo?: { runByLine: boolean; hashFileName?: string },
         _cancelToken?: CancellationToken
     ): Promise<boolean> {
+        // Ensure we have a controller to execute code against
+        // and a NotebookDocument to add the NotebookCell to
         const notebookDocument = await this.tryGetMatchingNotebookDocument();
+        await this.initialControllerSelected.promise;
```
joyceerhl, line 497: @rebornix This is the fix for the problem where clicking Run Cell sometimes results in the code not being executed.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -500,29 +512,106 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
+                observable.subscribe(
+                    async (cells: ICell[]) => {
+                        // Combine the cell data with the possible input data (so we don't lose anything that might have already been in the cells)
+                        const combined = cells.map(this.combineData.bind(undefined, data));
+
+                        // Then send the combined output to the UI
+                        const converted = (combined[0].data as nbformat.ICodeCell).outputs.map(
+                            cellOutputToVSCCellOutput
+                        );
+                        await temporaryExecution.replaceOutput(converted);
```
joyceerhl, line 584: This code for interactive window debugging is identical to the interactiveBase.ts version, except here we update the `NotebookCellOutput` using VS Code APIs instead of posting `ICell`s to the webview.

Note, I'm making an assumption here that we'll only ever get an `ICell` array of length one back from the observable change callback. Is it possible for a cell execution to update the output of more than one cell?

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -75,7 +75,7 @@ export class Kernel implements IKernel {
     get kernelSocket(): Observable<KernelSocketInformation | undefined> {
         return this._kernelSocket.asObservable();
     }
-    private notebook?: INotebook;
+    public notebook?: INotebook;
```
joyceerhl, line 78: Exposing this property on the `IKernel` isn't ideal but allows us to reuse basically all the existing debugging code, since `IKernel` doesn't have the concept of `executeObservable`. I really didn't want to go down the path of forwarding more `INotebook` methods in `IKernel` for consumption in the interactive window. Do folks care either way?

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -500,29 +510,106 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
+        } else {
+            const notebook = this.kernel?.notebook;
+            if (!notebook) {
+                return false;
+            }
+
+            try {
+                const finishedAddingCode = createDeferred<void>();
+
+                // Before we try to execute code make sure that we have an initial directory set
```
rchiodo, line 544: How come you don't need to do this for the non debug case?

src/client/datascience/jupyter/kernels/kernel.ts
rchiodo: LGTM

rchiodo (approved):

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
joyceerhl: The non debug case just uses `notebook.cell.execute`. I found that this `Deferred` was necessary for debugging a cell to ensure the debug session doesn't prematurely terminate.

src/client/datascience/editor-integration/codeLensFactory.ts
```diff
@@ -354,14 +354,7 @@ export class CodeLensFactory implements ICodeLensFactory, IInteractiveWindowList
...
         // Do not generate interactive window codelenses for TextDocuments which are part of NotebookDocuments
-        // and do not generate debug codelenses for new interactive window (not properly supported yet)
-        if (
-            (commandName.includes('debug') &&
-                workspace.getConfiguration('interactive.experiments')?.get('enable') === true) ||
-            workspace.notebookDocuments.find(
-                (notebook) => notebook.uri.toString() === document.notebook?.uri.toString()
-            )
-        ) {
+        if (workspace.notebookDocuments.find((notebook) => notebook.uri.toString() === document.uri.toString())) {
```
DonJayamanne, line 357: Isn't this the same? I feel its better.
```suggestion
        if (workspace.notebookDocuments.find((notebook) => notebook === document)) {
```

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
DonJayamanne: > Is it possible for a cell execution to update the output of more than one cell?

Yes. I think we should try to use the executeCell in `IKernel`.
Any reason why we cant use that instead of the executeObservable.
(after all tat's what we're doing in the non-debug case).

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -565,6 +655,30 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
+                    ...cell.data,
+                    metadata: {
+                        ...oldData.metadata,
+                        ...cell.data.metadata
+                    }
+                }
+            };
+            // Workaround the nyc compiler problem.
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            return (result as any) as ICell;
```
DonJayamanne, line 676: yes scary when we cast to `any`, will only find out issues during run time
This shouldn't be a problem when we get rid of ICell

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
```diff
@@ -209,6 +231,27 @@ export class NativeInteractiveWindowProvider implements IInteractiveWindowProvid
         });
     }
 
+    private update() {
+        const windows: NativeInteractiveWindow[] = [];
+
+        this._windows.forEach((win) => {
+            const notebookDocument = workspace.notebookDocuments.find(
+                (document) => win.notebookUri?.toString() === document.uri.toString()
```
DonJayamanne, line 239: Can we not use the following
```suggestion
                (document) => win.notebookUri === document
```

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -121,7 +128,8 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
         private _notebookUri: Uri | undefined, // This remains the same for the lifetime of the InteractiveWindow object
```
DonJayamanne, line 128: Can we refactor this contstructor.
Today `_notebookUri` is undefined & it only gets set inside the contructor after the interactive window is opened via the command `interactive.open`.
I'd prefer we did that out side the contructor, this way 

as follows:
```typescript
const notebookUri = await this.commandManager
            .executeCommand('interactive.open', ViewColumn.Beside)
            .then((result) => result.notebookUri);


const xyx = new NativeInteractiveWindow(...., notebookUri).

```

This way, the property `NativeInteractiveWindow.notebookUri` is never undefined.
Else it gets too complicated to figure out & handle situations where this can be undefined & not... 

I'd prefer if we could just create the interactive window outside this class (in the `nativeInteractiveWindowProvider.create` method).
Makes it more cleaner & precise.

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
DonJayamanne: With the change I suggested we wontn' end up with situations where `notebookUri` is undefined.

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: I think this was only so that you can use the `executeAsObservable`, but there are other issues there (side effects of running a cell that can update other cells).
hence we might have to just go with the standard IKernel.executeCell method.

DonJayamanne (changes requested):
Blocking on the `notebooKUri`, this should be initialized with a non-nullable value (i.e. create interaactive window in the provider & then pass the Uri as a ctor argument).

joyceerhl:
@DonJayamanne agreed re: the refactor, I attempted that early last week and had to revert it because it caused issues with catching the controller selection and change event and updating the cached kernel. I'll try that again.

We cannot use `executeCell` here, it doesn't work for debug cell. I started down the path of trying to unify the IKernel --> KernelExecution --> CellExecutionQueue --> CellExecution --> BaseJupyterSession codepath with the debug codepath, and adding a `debugCell` method, but wasn't able to get it working after 2-3 hours and it also seemed like I was trying to graft an unrelated concept (debugging) onto an existing abstraction, which increased complexity overall. It seems likely we will need to adopt CellExecutionQueue for the interactive window regardless (especially since we have now accepted that users will have to pick a kernel for the interactive window the first time round, before which they may have tried to execute multiple cells), but I didn't want to handle that right away in this PR.

src/client/datascience/editor-integration/codeLensFactory.ts
joyceerhl: IIRC they are not guaranteed to be the same object, so `document.uri.toString()` is the recommended way to check for document equality.

rebornix: There should be no duplicated documents (`TextDocument`, `NotebookDocument`) in the exthost.

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
rebornix: I'll suggest we either check `document` or `uri.toString()`.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
rebornix: +1 on this. The lifecyle of the `NativeInteractiveWindow` should probably be tied to `NotebookDocument` and then the Uri is never undefined.

rebornix: Maybe introduce a static `NativeInteractiveWindow.create()` if we want to embed the logic of iw opening inside `NativeInteractiveWindow`

src/client/datascience/editor-integration/codeLensFactory.ts
IanMatthewHuff: There is a scheme vscode-notebook-cell for the URI on NotebookCell textDocuments. That might be an easier check.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -500,29 +514,109 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
-        });
-        return true;
+        if (!debugInfo) {
+            await this.commandManager.executeCommand('notebook.cell.execute', {
+                ranges: [{ start: notebookDocument.cellCount - 1, end: notebookDocument.cellCount }],
+                document: notebookDocument.uri,
+                autoReveal: true
+            });
+        } else {
+            const notebook = this.kernel?.notebook;
```
IanMatthewHuff, line 541: Just a minor suggestion so feel free to ignore. But this function was already longish, so adding in the debugger case (which is quite long) makes the function and standard non-debug case a bit harder to read and follow, might be nice to subfunction it.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
IanMatthewHuff: I feel like we could us a no-explicit-any pass at some point, but not sure if we'd ever find the time. Looks like we have 350+ instances.

src/client/datascience/notebook/helpers/helpers.ts
IanMatthewHuff: Great, it's nice when something comes together like that. Sign that we structured something right.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -130,20 +136,31 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
+        const controller = this.notebookControllerManager.getSelectedNotebookController(this.notebookDocument);
+        if (controller !== undefined) {
+            this.registerKernel(this.notebookDocument, controller);
+        }
+
+        // Ensure we hear about any controller changes so we can update our cache accordingly
         this.notebookControllerManager.onNotebookControllerSelected(
             (e: { notebook: NotebookDocument; controller: VSCodeNotebookController }) => {
-                if (this._notebookUri !== undefined && e.notebook.uri.toString() !== this._notebookUri.toString()) {
+                if (e.notebook.uri.toString() !== this.notebookDocument.uri.toString()) {
```
DonJayamanne, line 151:
```suggestion
                if (e.notebook !== this.notebookDocument) {
```

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -130,20 +136,31 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
                     return;
                 }
 
                 // Clear cached kernel when the selected controller for this document changes
-                const controllerChangeListener = e.controller.controller.onDidChangeSelectedNotebooks(
+                const controllerChangeListener = (
+                    this.notebookController || e.controller
```
DonJayamanne, line 157: Why not just leave this as `e.controller`?
Why give preference to `this.notebookController`?

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -130,20 +136,31 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
+                ).controller.onDidChangeSelectedNotebooks(
                     (selectedEvent: { notebook: NotebookDocument; selected: boolean }) => {
                         // Controller was deselected for this InteractiveWindow's NotebookDocument
                         if (
                             selectedEvent.selected === false &&
-                            this._notebookUri !== undefined &&
-                            selectedEvent.notebook.uri.toString() === this._notebookUri.toString()
+                            selectedEvent.notebook.uri.toString() === this.notebookDocument.uri.toString()
```
DonJayamanne, line 163:
```suggestion
                            selectedEvent.notebook === this.notebookDocument
```

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -183,15 +195,10 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
     // point when we try to access it. Calling this method may cause this
     // InteractiveWindow object to dispose itself
     private async tryGetMatchingNotebookDocument(): Promise<NotebookDocument | undefined> {
-        await this.interactiveOpenPromise;
-
         const notebookDocument = workspace.notebookDocuments.find(
-            (document) => this._notebookUri?.toString() === document.uri.toString()
+            (document) => this.notebookDocument.uri.toString() === document.uri.toString()
```
DonJayamanne, line 199:
```suggestion
            (document) => this.notebookDocument === document
```

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
```diff
@@ -97,14 +121,23 @@ export class NativeInteractiveWindowProvider implements IInteractiveWindowProvid
...
+    protected async create(resource: Resource, mode: InteractiveWindowMode): Promise<NativeInteractiveWindow> {
         // Ensure all our controllers are registered with VS Code
         await this.notebookControllerManager.loadNotebookControllers();
 
+        const { notebookUri } = (await this.commandManager.executeCommand(
+            'interactive.open',
+            ViewColumn.Beside
+        )) as INativeInteractiveWindow;
+        const notebookDocument = workspace.notebookDocuments.find(
+            (doc) => doc.uri.toString() === notebookUri.toString()
```
DonJayamanne, line 133:
```suggestion
            (doc) => doc === notebookUri
```

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
```diff
@@ -209,6 +243,27 @@ export class NativeInteractiveWindowProvider implements IInteractiveWindowProvid
         });
     }
 
+    private update() {
+        const windows: NativeInteractiveWindow[] = [];
+
+        this._windows.forEach((win) => {
+            const notebookDocument = workspace.notebookDocuments.find(
```
DonJayamanne, line 250: Not sure this is possible.
If we're holding onto a reference of a Notebook, how is it that it that we're holding onto a reference of that object when this notebook no longer exists.

I understand its a simple block of code, but I'd prefer to avoid such code that account for scenarios that should tehcnically never occur.
I.e. if we handle the onDidCloseNotebook document, then the interactive window would get diposed.

DonJayamanne: Also I think it might be better to use expose the `NotebookDocument isntead of `NotebookUri` as you're comparing Uris, when its more efficient to compare the object reference.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
DonJayamanne: This doesn't make sense, I can't understand how we'd have the object reference yet not have the notebook in the list of `workspace.notebookDocuments`.

Can't we just check `this.notebookDocument.isClosed` in this method and return undefined.
& optionally add an event handler to the `workspace.onDidCloseNotebookDocument(() => this.dispose())` (if the same notebook has been closed).

DonJayamanne (changes requested):
Sorry but the disposing looks a little funky, 
We have code that i think can be cleaned up, i don't think we can end up in situations where we have a notebook, but the object no longer exists in the list of notebooks. If that were to happen then the notebook has already been disposed/closed and we have events to handle that & a propety to check if the `notebook.isClosed`.

I.e. i dont think we need to look for the document in the list of notebooks
We can just check if its closed, after all we 're not storing the reference in the interactive window class (instead of the Uri, as was done previously).

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
joyceerhl: > if we handle the onDidCloseNotebook document, then the interactive window would get diposed.

That's exactly what this code is doing. This is the function that gets called when `workspace.onDidCloseNotebookDocument` fires, and it ensures that we dispose our NativeInteractiveWindow wrapper objects if the underlying NotebookDocument has been disposed (which can absolutely happen).

joyceerhl: Maybe you're suggesting that inside the NativeInteractiveWindow, we should listen for onDidCloseNotebookDocument, and call dispose on ourselves if the closed NotebookDocument is the one we hold a reference to? That's how I originally implemented it, but Peng and I discussed and agreed it seemed weird for an object to be disposing itself. If anything an object's manager/provider should be handling disposal. That's why we landed on this pattern.

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
joyceerhl: This is comparing a Uri to a NotebookDocument? Not sure this works.

src/client/datascience/editor-integration/codeLensFactory.ts
DonJayamanne: > IIRC they are not guaranteed to be the same object, so document.uri.toString() is the recommended way to check for document equality.

Whoa, @rebornix is that correct. We use references so over today in Jupyter, weakmaps in a number of places. Ie we awesome there's only one insurance of the same document, hence same ref.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -130,21 +135,30 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
             this._submitters.push(owner);
         }
 
+        this.initialControllerSelected = createDeferred<void>();
+
+        // Immediately try to find a selected controller for our NotebookDocument,
+        // as it's possible that a selection event fired before our ctor was able to run
+        const controller = this.notebookControllerManager.getSelectedNotebookController(this.notebookDocument);
```
DonJayamanne, line 142: @joyceerhl is this necessary? I  can't think of a situation when `controller` here would ever be not undefined. for a controller to be associaed with an interactive window, the user has to run a cell, but to run a cell the window has to appear and user has to select a controller.
Personally i'd prefer to get rid of `getSelecedNotebookController`, i thought i removed this in the main branch (I've added this back in my other repo).

joyceerhl: The problem is that the selection event can fire before the interactive window ctor runs. It's either we maintain the mapping in the notebookControllerManager, or we maintain the same mapping in the nativeInteractiveWindowProvider and have the provider set the controller on the nativeInteractiveWindow.

> to run a cell the window has to appear and user has to select a controller

Yes, but the controller may not have been selected by the time that the interactive window's addCode method gets called.

Basically what we want to do is ensure that execution requests are not dropped even when a controller hasn't been selected yet, which can absolutely happen. Personally it makes more sense to me to wait for the controller to be selected in the interactiveWindow, versus exposing that information say in the commandListener and making the commandListener responsible for checking if a controller has been selected and queuing execution requests until the controller has been selected.

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
DonJayamanne: > listen for onDidCloseNotebookDocument, and call dispose on ourselves if the closed NotebookDocument is the on

Yes, thats right.

> to be disposing itself. If anything an object's manager/provider should be handling disposal. That's why we landed on this pattern.

But then why can't we have a manager listen to the dispose event and then dispose the interactive windows. 
Anyways, looks like this has been discussed, lets leave this as is then.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
DonJayamanne: Hmm, lets chat tomorrow.

DonJayamanne (approved):

src/client/datascience/editor-integration/codeLensFactory.ts
joyceerhl: Just to close the loop here, you can't directly compare URI objects for equality since they're not guaranteed to be the same object, but comparing TextDocuments and NotebookDocuments should be safe

DonJayamanne:
Can work on other changes later

src/client/datascience/editor-integration/codeLensFactory.ts
DonJayamanne: Nope they are guaranteed to the be same object, thats how it works today.
We use instances in Native Notebooks, instead of the Uri. VS Code never re-create the notebook object.

PR #6366, opened 2021-06-23
Move logic to individual controllers

DonJayamanne:
VS Code doesn't have a concept of a controller manager. I think we had this prior to kernel push.
In VS Code each controller has its own event whether its selected/de-selected.
Looking at the code we're trying to move this back to controller manager (kind of the way it was preivously) & i feel that's incorrect, 
If VS Code has events in the controllers, then I think we should handle that in the controller and do what ever we need to do there, rather than adding an abstraction & propagating the events to the manager & then doing the stuff in the manager.

Also, the manager creates the controller, but the controller calls methods on the Manager via callbacks (test method i added) & some otehr public methods to get the controller associated with a document..
I've removed that as well.

This way, the relationship is simple, mnanager creates controllers and can call methods in controllers.
But controllers cannot access methods in manager. (else its a cyclic dependency)

### DIFF
--- a/src/client/datascience/kernel-launcher/kernelDaemonPreWarmer.ts
+++ b/src/client/datascience/kernel-launcher/kernelDaemonPreWarmer.ts
@@ -4,17 +4,12 @@
 'use strict';
 
 import { inject, injectable } from 'inversify';
-import { NotebookDocument, NotebookCell } from 'vscode';
+import { NotebookDocument } from 'vscode';
 import { IPythonExtensionChecker } from '../../api/types';
 import { IVSCodeNotebook } from '../../common/application/types';
-import { PYTHON_LANGUAGE } from '../../common/constants';
 import '../../common/extensions';
 import { IConfigurationService, IDisposableRegistry, Resource } from '../../common/types';
 import { swallowExceptions } from '../../common/utils/decorators';
-import { isUntitledFile } from '../../common/utils/misc';
-import { isPythonKernelConnection } from '../jupyter/kernels/helpers';
-import { getNotebookMetadata, isJupyterNotebook, isPythonNotebook } from '../notebook/helpers/helpers';
-import { INotebookControllerManager } from '../notebook/types';
 import {
     IInteractiveWindowProvider,
     INotebookCreationTracker,
@@ -36,8 +31,7 @@ export class KernelDaemonPreWarmer {
         @inject(IRawNotebookSupportedService) private readonly rawNotebookSupported: IRawNotebookSupportedService,
         @inject(IConfigurationService) private readonly configService: IConfigurationService,
         @inject(IVSCodeNotebook) private readonly vscodeNotebook: IVSCodeNotebook,
-        @inject(IPythonExtensionChecker) private readonly extensionChecker: IPythonExtensionChecker,
-        @inject(INotebookControllerManager) private readonly controllerManager: INotebookControllerManager
+        @inject(IPythonExtensionChecker) private readonly extensionChecker: IPythonExtensionChecker
     ) {}
     public async activate(_resource: Resource): Promise<void> {
         // Check to see if raw notebooks are supported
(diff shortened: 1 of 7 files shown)


src/client/datascience/kernel-launcher/kernelDaemonPreWarmer.ts
```diff
@@ -88,23 +82,23 @@ export class KernelDaemonPreWarmer {
...
-            })
-        ) {
-            await this.preWarmKernelDaemonPool();
-        }
+    private async onDidOpenNotebookDocument(_doc: NotebookDocument): Promise<void> {
+        // // It could be anything, lets not make any assumptions.
+        // if (isUntitledFile(doc.uri) || !isJupyterNotebook(doc)) {
+        //     return;
+        // }
+        // const kernelConnection = this.controllerManager.getSelectedNotebookController(doc)?.connection;
```
DonJayamanne, line 90: This doesn't work, as in the controller manager, the document is opened and setting the preferred is done asynchronously. Also there's no such thing as a selected controller anymore.

I think we have an issue to track auto start of native notebook kernels, somethign to do with some UI (progress indicator) not appearing corrrectly.

either way, I've disabled this as it wont' work.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -99,17 +90,11 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
         // Sign up for document either opening or closing
         this.notebook.onDidOpenNotebookDocument(this.onDidOpenNotebookDocument, this, this.disposables);
-        this.notebook.onDidCloseNotebookDocument(this.onDidCloseNotebookDocument, this, this.disposables);
 
         // Be aware of if we need to re-look for kernels on extension change
         this.extensions.onDidChange(this.onDidChangeExtensions, this, this.disposables);
     }
 
-    // Look up what NotebookController is currently selected for the given notebook document
-    public getSelectedNotebookController(document: NotebookDocument): VSCodeNotebookController | undefined {
```
DonJayamanne, line 109: Was only used in auto start, hence removed.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -342,54 +323,11 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             traceError(`Failed to create notebook controller for ${kernelConnection.id}`, ex);
         }
     }
-    /**
-     * In our tests, preferred controllers are setup as the active controller.
-     *
-     * This method is called on when running tests, else in the real world,
-     * users need to select a kernel (preferred is on top of the list).
-     */
-    private async setAsActiveControllerForTests(controller: VSCodeNotebookController, notebook: NotebookDocument) {
```
DonJayamanne, line 351: Moved into controller.
The controller was calling this method via a callback. I.e. manager creates controller, controller calls a method in manager via a callback, yuck...

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -468,90 +406,4 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             controller.dispose();
         });
     }
-
-    private async notebookKernelChanged(document: NotebookDocument, controller: VSCodeNotebookController) {
```
DonJayamanne, line 472: Moved all of this code into the controller, the controller has an event for whether its seleced or not, hence moved it closer to where the evnt is fired, instead of adding our own layer (our own event)..

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -67,6 +80,11 @@ export class VSCodeNotebookController implements Disposable {
     get onDidReceiveMessage() {
         return this.controller.onDidReceiveMessage;
     }
+    @testOnlyMethod()
+    public isAssociatedWithDocument(doc: NotebookDocument) {
+        return this.associatedDocuments.has(doc);
+    }
+    private readonly associatedDocuments = new WeakSet<NotebookDocument>();
```
DonJayamanne, line 87: Instead of `controllerMapping`, the controller will have its own property to keep track of whether a document is associated with this controller or not.

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -77,15 +95,14 @@ export class VSCodeNotebookController implements Disposable {
         private readonly kernelProvider: IKernelProvider,
         private readonly preferredRemoteKernelIdProvider: PreferredRemoteKernelIdProvider,
         private readonly context: IExtensionContext,
-        private readonly notebookControllerManager: INotebookControllerManager,
         private readonly pathUtils: IPathUtils,
         disposableRegistry: IDisposableRegistry,
         private readonly languageService: NotebookCellLanguageService,
         private readonly workspace: IWorkspaceService,
-        private readonly setAsActiveControllerForTests: (
```
DonJayamanne, line 85: Removed this hack, calling back into the controller

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -77,15 +95,14 @@ export class VSCodeNotebookController implements Disposable {
         private readonly kernelProvider: IKernelProvider,
         private readonly preferredRemoteKernelIdProvider: PreferredRemoteKernelIdProvider,
         private readonly context: IExtensionContext,
-        private readonly notebookControllerManager: INotebookControllerManager,
```
DonJayamanne, line 80: Remvoed this cyclic dependency

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -265,9 +295,8 @@ export class VSCodeNotebookController implements Disposable {
                 return;
             }
 
-            const documentConnection = this.notebookControllerManager.getSelectedNotebookController(doc);
```
DonJayamanne, line 268: This cyclic dependency is also gone now.

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -77,15 +95,14 @@ export class VSCodeNotebookController implements Disposable {
...
-        private readonly notebookControllerManager: INotebookControllerManager,
         private readonly pathUtils: IPathUtils,
         disposableRegistry: IDisposableRegistry,
         private readonly languageService: NotebookCellLanguageService,
         private readonly workspace: IWorkspaceService,
-        private readonly setAsActiveControllerForTests: (
-            controller: VSCodeNotebookController,
-            notebook: NotebookDocument
-        ) => Promise<void>
+        private readonly localOrRemoteKernel: 'local' | 'remote',
```
DonJayamanne, line 102: This controller only needs to know whether its a local or remove controller.
This way, if tomorrow we support both together, we don't need to make too many changes.
I.e. don't assume that we can only even have one type at a time...

IanMatthewHuff (approved):

Issue #6354, opened 2021-06-22
Resolve DataFrame horizontal scrollbar issue on Jupyter side


IanMatthewHuff:
Per the issue here: https://github.com/microsoft/vscode/issues/123816

Jupyter extension will look at providing the styles to size dataframe table output. Will also look at if this needs to be released in a point fix.

rebornix:
<img width="539" alt="image" src="https://user-images.githubusercontent.com/876920/123042550-d9b02480-d3ab-11eb-9b65-fbc24878d926.png">

With `width: 100%` removed on table, it's actually an easy fix in the core https://github.com/microsoft/vscode/commit/d111d967b72748bbd125d3392eeebc52718a7c07

CWen001:
@rebornix Sorry to have a question here. What color theme is used in this screencast? I have tried a few light color themes and the pandas output is not that pleasing. Yours looks great.

PR #6367, opened 2021-06-23
Update styles to support H-Scroll with dataframe tables

IanMatthewHuff:
For #6354 


src/client/datascience/notebook/cssStyles/tableStyles.js
```diff
@@ -0,0 +1,10 @@
+// This file is a preload just for current release bits, it updates vscode core styles + dataframe styles to allow
+// for horizontal scrolling in native notebook table elements
+var tableStyleSheet = document.createElement('style');
+tableStyleSheet.innerHTML = `#container > div > div:not(.preview) > div {overflow-x: overlay;};`;
+document.body.appendChild(tableStyleSheet);
+var dataframes = document.getElementsByClassName('dataframe');
```
IanMatthewHuff, line 6: @rebornix helped to sort this out, but in stable the styles are a bit different so we had to expand the fix here to also set width and overflowX on the divs that wrape the dataframes. Without this the overflow-x: overlay wasn't getting picked up until the element was resized or a new layout was triggered.

IanMatthewHuff: This only has to update the dataframes currently on the page, as new dataframes created by execution correctly pick up the overlay style.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 2021.6.9999 (23 June 2021)
```
IanMatthewHuff, line 3: Maybe the 9999 is silly here? Just following the structure until we get to our next major release and we can drop the extra 9999s and get back on the month schedule.

src/client/datascience/notebook/cssStyles/tableStyles.js
rebornix: > Without this, the overflow-x: overlay wasn't getting picked up until the element was resized or a new layout was triggered

what we did here is basically triggering the browser to do a real scrollHeight/offsetHeight check.

DonJayamanne: This seems like a total hack, soemthing I've done a lot of in IE6/IE8 days.
To my knowledge this could be bad in a large document, with large number of outputs with the same class. Unlikely yes, but could have some serious repercussions.

@rebornix is this because the pre-loads are loaded after the HTML is rendered? Sounds like thats the reason.

DonJayamanne: Which also explains why the UI first doesn't have scroll bars, then later after 1-2 seconds we see the scrollbars.

I'd prefer to just add the styles into the oiutput html if thats the case, as that provides a better UX, else this is too clunky for me & hackish as well (by iterating through all `dataframe` classes, which again is an unnecessary hardcoding)

IanMatthewHuff: It is hackish, but this change is intended not to be backported and only be for the release branch to cover the next two weeks of experiment rollout. VSCode main already has a fix for this in place for the next release. @rebornix and I chatted about some longer term fixing around getting jupyter styling in HTML as well. Which might be either through output html editing or a jupyter specific HTML rendering mime type.

Editing the output HTML would have to be done both on execution and on document loading as well so that the output would work for files that were not yet executed. We'd also have to mark it specifically so that we can remove it from saved output. That also feels hackish to me in that the controllers are still modifying kernel output and are controlling UI display which should be the domain of renderers. 

The reason for the dataframe sizing isn't explicitly for pre-loads happening after the HTML is rendered. In main (prior to @rebornix 's change) the dataframe code isn't required, just adding the new overflow style.

IanMatthewHuff: Options that we have:
1. Take this hackish change now for release only
2. Only keep the style addition from this change and remove the dataframe sizing. This (IMO) makes it safer as it's just a CSS update. Only issue is the scrollbar is not picked up until you resize / reexecute / whatever. Might be ok to fill the gap for now in a safer fashion. Also for release only. 
3. Rule this preload approach out and edit the output html. I believe this is not super hard, but it also doesn't feel long term correct to me, and riskier to take as a point release.

DonJayamanne (approved):

DavidKutu (approved):

src/client/datascience/notebook/cssStyles/tableStyles.js
rebornix: It's a hack. Considering we are not releasing a dot release for VS Code for next 2 weeks, it can unblock us from bumping the ratio. I'm not too concerned about the performance penalty as we are triggering the browser to recalcuate the height, which they should do.

I'm not in favor of tweaking outputs as it will involve code changes for both file opening and saving, which is more risky than the current option now (having impact almost on every user other than just users/notebooks with`dataframe` outputs).

Issue #6353, opened 2021-06-22
Don't notify the user to reload after installing the extension


digitarald:
Python (or Jupyter?) does its own reload after install prompt:

![image](https://user-images.githubusercontent.com/8599/122977344-2236f600-d34a-11eb-8830-cc687ac598cf.png)

(the screenshot is from usertesting but I had no luck reproducing this for myself)

It would be great if it was left to VS Code to handle the reload, so extensions don't run extra notifications. If there are edge cases that VS Code doesn't handle well or if the UX isn't obvious enough, we should improve those in VS Code to make this easier for all extensions.

IanMatthewHuff:
@digitarald Thanks for filing. I do think we could probably handle this better. I believe there are a few edge cases that might not work in a specific open notebook or Interactive Window document. But this messaging is pretty non-standard. We might be able to winnow out those edge cases or if not, just go ahead and have VS Code do a full reload.

greazer:
We should eliminate this message box from needing to be shown by scanning our source for calls to the Python extension and testing to see whether those scenarios are working without having to reload VS Code. If so, then we can remove the message box.

digitarald:
A proposal I discussed with @isidorn was that if the extension determines that it needs to reload, it should be handled by VS Code. Worst case would be that a user syncs multiple extensions into a fresh VS Code instance and multiple extensions prompting for restart.

Tyriar:
I just installed the python extension to try debug a python file without a launch config and I got this notification, I don't have any notebooks open so I don't know what it's talking about.

IanMatthewHuff:
This was a duplicate of a previous issue that I cleaned up. The new logic is only to show this message if there is an active notebook or interactive window python kernel. In which case the warning seems reasonable.

PR #6371, opened 2021-06-23
Support listing non-python kernels early

DonJayamanne:
* List non-pyton kernels early
* Non-Python & Python kernels are loaded separately (asynchronously)
	* As we get new kernels they are registered
	* The assumption is non-python kernels will be fetched very quickly
* At the same time, if users install Pytohn extension after opening a notebook, the changes now detect this and re-discover the kernels
	* Was easy as i could use the same infrastructure

recording (1)

### DIFF
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -12,8 +12,7 @@ import {
     IDisposableRegistry,
     IExtensionContext,
     IExtensions,
-    IPathUtils,
-    Resource
+    IPathUtils
 } from '../../common/types';
 import { StopWatch } from '../../common/utils/stopWatch';
 import { Telemetry } from '../constants';
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -51,7 +51,6 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
     // Listing of the controllers that we have registered
     private registeredControllers = new Map<string, VSCodeNotebookController>();
 
-    private cancelToken: CancellationTokenSource | undefined;
     private readonly isLocalLaunch: boolean;
     private wasPythonInstalledWhenFetchingControllers?: boolean;
     constructor(
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -17,11 +17,7 @@ import {
 } from '../../common/types';
 import { StopWatch } from '../../common/utils/stopWatch';
 import { Telemetry } from '../constants';
-import {
-    areKernelConnectionsEqual,
-    getDisplayNameOrNameOfKernelConnection,
-    isLocalLaunch
-} from '../jupyter/kernels/helpers';
+import { getDisplayNameOrNameOfKernelConnection, isLocalLaunch } from '../jupyter/kernels/helpers';
 import { IKernelProvider, KernelConnectionMetadata } from '../jupyter/kernels/types';
 import { ILocalKernelFinder, IRemoteKernelFinder } from '../kernel-launcher/types';
 import { PreferredRemoteKernelIdProvider } from '../notebookStorage/preferredRemoteKernelIdProvider';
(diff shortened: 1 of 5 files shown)


src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -52,10 +50,11 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
     private controllersPromise?: Promise<void>;
 
     // Listing of the controllers that we have registered
-    private registeredControllers: VSCodeNotebookController[] = [];
+    private registeredControllers = new Map<string, VSCodeNotebookController>();
```
DonJayamanne, line 53: Changed to a Map, this way we can discover Non-Python kernels in parallel (as we get new items we just push it into this map & easy to check if an item exists, by id).

Same list is used to re-fetch controllers when Pyton extension is installed (re-run controller discovery without having to worry about duplicate controllers). Easy fix.

DonJayamanne: This ensures we don't need to reload VS Code (at least for kernel discovery, partly fixes https://github.com/microsoft/vscode-jupyter/issues/6353).
I think there are other areas that might require similar fixes. I fixed this as it was super simple

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -98,64 +98,61 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
-                .then((controllers) => {
-                    // Just assign here as this is our initial set of controllers
-                    // anything that adds or updates should make sure the initial load has happened first
-                    this.registeredControllers = controllers;
-                })
+            const stopWatch = new StopWatch();
+
+            this.controllersPromise = Promise.all([
+                this.loadNotebookControllersImpl(true),
+                this.loadNotebookControllersImpl(false)
```
DonJayamanne, line 111: Fetch both in parallel, if non-python kernels finish first, then those kernels will be registerd early (validatd by adding a simple sleep of 10s in pyton discovery)

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -259,58 +256,49 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
-                    this.disposables
-                );
-
-                // We are disposing as documents are closed, but do this as well
-                this.disposables.push(controller);
-
-                return controller;
-            });
+            ]
+                .filter(([id]) => !this.registeredControllers.has(id))
```
DonJayamanne, line 270: This way we can try to create the controllers without having to worry about duplicates.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -259,58 +256,49 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+                    // Hook up to if this NotebookController is selected or de-selected
+                    controller.onNotebookControllerSelected(
+                        this.handleOnNotebookControllerSelected,
+                        this,
+                        this.disposables
+                    );
+
+                    // We are disposing as documents are closed, but do this as well
+                    this.disposables.push(controller);
+                    this.registeredControllers.set(controller.id, controller);
```
DonJayamanne, line 300: once created we add it here, rather than returning it and trhying to add it elsewhere..

src/test/datascience/kernel-launcher/localKernelFinder.unit.test.ts
```diff
@@ -33,8 +33,9 @@ import { IExtensions } from '../../../client/common/types';
 import { LocalKnownPathKernelSpecFinder } from '../../../client/datascience/kernel-launcher/localKnownPathKernelSpecFinder';
 import { JupyterPaths } from '../../../client/datascience/kernel-launcher/jupyterPaths';
 import { LocalPythonAndRelatedNonPythonKernelSpecFinder } from '../../../client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder';
+import { getInterpreterHash } from '../../../client/pythonEnvironments/info/interpreter';
 
-[false, true].forEach((isWindows) => {
+[false].forEach((isWindows) => {
```
IanMatthewHuff, line 38: Did you mean to remove the true here?

IanMatthewHuff (approved):

src/test/datascience/kernel-launcher/localKernelFinder.unit.test.ts
DonJayamanne: Oops, was testing

DonJayamanne: Fixed

PR #6375, opened 2021-06-24
Allow users to opt into preview native interactive window experience

joyceerhl:
Per discussion with @rebornix and @claudiaregio, we would like to roll out the new interactive window experience for Jupyter VS Code Insiders users as soon as possible to get early feedback. One blocker right now is that IntelliSense will only be available for users who are opted into the Python Daily insiders channel. We are planning to flip the default value of the setting introduced in this PR to `true` for Python Daily insiders users as soon as our IntelliSense PR lands in vscode-python main.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1453,6 +1453,12 @@
                     "description": "Enables/disables A/B tests.",
                     "scope": "machine"
                 },
+                "jupyter.enableNativeInteractiveWindow": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Enables the preview native interactive window experience. Set to false to opt back into the webview interactive window experience.",
+                    "scope": "machine"
+                },
                 "jupyter.logging.level": {
                     "type": "string",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1453,6 +1453,12 @@
                     "description": "Enables/disables A/B tests.",
                     "scope": "machine"
                 },
+                "jupyter.enableNativeInteractiveWindow": {
```
rchiodo, line 1456: How come we're not using the experiments framework for this? Is it because we don't need experiment data on this? It's really just a setting? 

Not sure if we still have those type of experiments. Experiments that are not backed by the experiment service.

rchiodo (approved):

package.json
joyceerhl: Yeah, this is just to target Insiders users for initial feedback, we're not going to rollout to Stable yet so experiment data isn't really necessary.

Issue #5650, opened 2021-04-27
FindMatchingInterpreter matching on interpreter path


IanMatthewHuff:
Noticed this in a semi-related PR here.

https://github.com/microsoft/vscode-jupyter/pull/5644#pullrequestreview-646103184

In FindMatchingInterpreter it's matching based on the path of the interpreter. I believe this might be an issue as "Interpreters" here is actually PythonEnvironments, and for Conda I believe you can have multiple environments with the same interpreter path if you don't specify the python version when creating the env.

rchiodo:
We need to rethink how we uniquely identify PythonEnvironments.

greazer:
We should add telemetry to see how many users fall into this category. Not necessary for June though.

IanMatthewHuff:
Repro for this.

Create a conda env. Then you can create named conda kernelspec like this:
(dupe1) Ians-MacBook-Pro:~ ianhuff$ python -m ipykernel install --user --name dupeEnvUno --display-name "Dupe Env Uno"
Installed kernelspec dupeEnvUno in /Users/ianhuff/Library/Jupyter/kernels/dupeenvuno
(dupe1) Ians-MacBook-Pro:~ ianhuff$ python -m ipykernel install --user --name dupeEnvDos --display-name "Dupe Env Dos"
Installed kernelspec dupeEnvDos in /Users/ianhuff/Library/Jupyter/kernels/dupeenvdos

These will all get startUsingPythonInterpreter kernelConnectionMetadata and they will all have the same path. Find preferredKernel in helpers.ts matches only on path in a couple of places. If using old webview you run with "Dupe Env Uno" and save the file. The next time you open it you might get your kernel switched to the main conda env (dupe1) or "Dupe Env Dos" it's all ordering dependent.

The interactive window part usually works since the current interpreter is first in the list, so it usually gets matched before the other ones, but that is ordering dependent to some degree as well.

rchiodo:
These are just completely skipped now. Seems wrong to me? Kernelspec might have a custom env.

rchiodo:
This works fine if the custom kernelspecs are any different than the default kernelspec.

PR #6379, opened 2021-06-25
Pass active interpreter's controller id to VS Code when creating native interactive windows

joyceerhl:
# Background
I had a lengthy discussion with @greazer and later @rebornix today regarding honoring the user's active Python interpreter selection when starting the interactive window. Previously @DonJayamanne and I discussed requiring users to use the kernel picker UI to select a kernel, as opposed to forcing a selection automatically, because even though we are setting the preferred affinity for the controller, that only causes it to show up at the top of the kernel picker list rather than autoselecting it for the NotebookDocument. 

Interactive window usage is driven entirely by a script file tied to an active Python interpreter (in future, active Julia/R environment). From the user's standpoint, they have explicitly told us which Python environment to use via their active interpreter selection, and we would be disregarding that preference by requiring them to use the kernel picker UI.

# Description of changes
This change allows us to preserve the current UX for Python users working in interactive #%% files with the interactive window. Peng and I have added support in VS Code (https://github.com/microsoft/vscode/commit/1750b2b231b47dfc2373c7ef2ee55a1a93e0c230) and the Jupyter extension (this PR) to pass the controller ID to VS Code when creating interactive windows. If this ID is not undefined, i.e. the Python extension is installed, there is an active Python interpreter selected, and a controller has been registered for the interpreter, then VS Code will use this controller for the interactive window that it creates when handling `interactive.open`. If this ID is undefined, VS Code will then look for its own cached controller association for the interactive window's embedded notebook editor and use that instead. 

In future, particularly when we are looking at supporting a #%%-file driven scripting experience for arbitrary languages in the interactive window, we may wish to discuss first-class VS Code support for an extension-level 'active execution environment', in order to eliminate the disconnect between language extension-contributed active environments and the kernel associated with a given notebook. Alternatively, the Jupyter extension can query other language extensions which also contribute controllers to VS Code for their active environments, in order to ensure that the active environment is honored when starting an interactive window. However, this is out of scope for the initial preview of the interactive window.

### DIFF
--- a/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
+++ b/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
@@ -591,11 +591,14 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
         // Ensure we have a controller to execute code against
         // and a NotebookDocument to add the NotebookCell to
         const notebookDocument = this.notebookDocument;
+        if (!this.notebookController) {
+            await this.commandManager.executeCommand('notebook.selectKernel');
+        }
         await this.initialControllerSelected.promise;
         await this.kernelLoadPromise;
 
         // ensure editor is opened/focused
-        await this.commandManager.executeCommand('interactive.open', undefined, notebookDocument.uri);
+        await this.commandManager.executeCommand('interactive.open', undefined, notebookDocument.uri, this.notebookController!.id);
 
         // Strip #%% and store it in the cell metadata so we can reconstruct the cell structure when exporting to Python files
         const settings = this.configuration.getSettings();
--- a/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
+++ b/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
@@ -589,8 +589,6 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
 
     private async addNotebookCell(code: string): Promise<NotebookCell> {
         // Ensure we have a controller to execute code against
-        // and a NotebookDocument to add the NotebookCell to
-        const notebookDocument = this.notebookDocument;
         if (!this.notebookController) {
             await this.commandManager.executeCommand('notebook.selectKernel');
         }
@@ -601,7 +599,7 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
         await this.commandManager.executeCommand(
             'interactive.open',
             undefined,
-            notebookDocument.uri,
+            this.notebookDocument.uri,
             this.notebookController!.id
         );
 
@@ -624,14 +622,14 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
         );
         notebookCellData.metadata = { interactiveWindowCellMarker };
         edit.replaceNotebookCells(
-            notebookDocument.uri,
-            new NotebookRange(notebookDocument.cellCount, notebookDocument.cellCount),
+            this.notebookDocument.uri,
+            new NotebookRange(this.notebookDocument.cellCount, this.notebookDocument.cellCount),
             [
                 notebookCellData // TODO generalize to arbitrary languages and cell types
             ]
         );
         await workspace.applyEdit(edit);
-        return notebookDocument.cellAt(notebookDocument.cellCount - 1);
+        return this.notebookDocument.cellAt(this.notebookDocument.cellCount - 1);
     }
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any, no-empty,@typescript-eslint/no-empty-function
--- a/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
+++ b/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
@@ -598,7 +598,12 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
         await this.kernelLoadPromise;
 
         // ensure editor is opened/focused
-        await this.commandManager.executeCommand('interactive.open', undefined, notebookDocument.uri, this.notebookController!.id);
+        await this.commandManager.executeCommand(
+            'interactive.open',
+            undefined,
+            notebookDocument.uri,
+            this.notebookController!.id
+        );
 
         // Strip #%% and store it in the cell metadata so we can reconstruct the cell structure when exporting to Python files
         const settings = this.configuration.getSettings();
--- a/src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
+++ b/src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
@@ -29,6 +29,7 @@ import * as localize from '../../common/utils/localize';
 import { noop } from '../../common/utils/misc';
 import { IServiceContainer } from '../../ioc/types';
 import { IExportDialog } from '../export/types';
+import { getInterpreterKernelSpecName } from '../jupyter/kernels/helpers';
 import { IKernelProvider } from '../jupyter/kernels/types';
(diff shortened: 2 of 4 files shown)


src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -591,11 +591,19 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
         // Ensure we have a controller to execute code against
         // and a NotebookDocument to add the NotebookCell to
         const notebookDocument = this.notebookDocument;
+        if (!this.notebookController) {
```
rebornix, line 594: we don't need to call this anymore considering we have passed the id to the `interactive.open` command. Keeping this will trigger the kernel picker sometimes, which we don't want to see.

joyceerhl: @rebornix If the id is undefined, e.g. there is no selected Python interpreter, and there is no cached association in VS Code, I believe we would still need to prompt the user to select a kernel for the first time.

DavidKutu (approved):

rebornix (approved):

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
IanMatthewHuff: Is this state here ok if the users just hits escape in the kernel picker and doesn't pick anything?

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
```diff
@@ -125,10 +128,30 @@ export class NativeInteractiveWindowProvider implements IInteractiveWindowProvid
...
         await this.notebookControllerManager.loadNotebookControllers();
 
+        // Fetch the active interpreter and use the matching controller
+        const api = await this.pythonApi.getApi();
+        const activeInterpreter = await api.getActiveInterpreter();
+        const preferredController = this.notebookControllerManager
+            .registeredNotebookControllers()
+            .find((controller) => {
+                return (
+                    controller.connection.interpreter?.path === activeInterpreter?.path &&
```
IanMatthewHuff, line 138: Looking up now, but I don't believe this will be correct. Multiple conda envs can share the same interpreter path.

IanMatthewHuff (changes requested):
I believe the activeInterpreter => connection check isn't correct as is. Looking that up myself now.

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
joyceerhl: Yep you're totally right

IanMatthewHuff: In startUsingPythonInterpreter type connections I believe that we save the entire env. displayName and envType are optional, but could help to distinguish here.

IanMatthewHuff: Hmm, I was just looking more at this, I can't actually recall the way that we were seeing conda envs with shared interpreter paths. I recall discussing that but I was unable to make one myself.

IanMatthewHuff: Ok, I did actually create one. Looking at the code there might actually be a bug here for existing interactive window. Digging in now.

joyceerhl: I pulled this logic from https://github.com/microsoft/vscode-jupyter/blob/a127a52f71c0bb10a1aa0979a310d871fdeb521d/src/client/datascience/jupyter/kernels/helpers.ts#L388 Does this need to be updated as well?

joyceerhl: This existing issue might be the same problem: https://github.com/microsoft/vscode-jupyter/issues/5650

IanMatthewHuff: Yup that's the location that I ran into looking at your code here. I'm building a quick repro to test.

IanMatthewHuff (approved):

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
joyceerhl: Yeah, we simply won't execute anything until a controller is selected for this window.

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
```diff
@@ -244,6 +255,45 @@ export class NativeInteractiveWindowProvider implements IInteractiveWindowProvid
...
+        if (preferredController === undefined) {
+            preferredController = interpreterControllers.find((controller) => {
+                return (
+                    controller.connection.kind === 'startUsingPythonInterpreter' &&
+                    controller.connection.interpreter?.path === activeInterpreter?.path
+                );
+            });
+        }
+
+        return preferredController !== undefined ? `${JVSC_EXTENSION_ID}/${preferredController.id}` : undefined;
```
rchiodo, line 294: This seems weird that you need to prefix this with JVSC extension id. Shouldn't the ID already include that if it's necessary?

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
```diff
@@ -244,6 +255,45 @@ export class NativeInteractiveWindowProvider implements IInteractiveWindowProvid
...
+                controller.controller.notebookType === InteractiveWindowView
+        );
+
+        // First try to match on the kernelSpec name. This is supposed to uniquely
+        // identify the active interpreter.
+        const kernelSpecName = getInterpreterKernelSpecName(activeInterpreter);
+        let preferredController = interpreterControllers.find((controller) => {
+            return (
+                controller.connection.kind === 'startUsingPythonInterpreter' &&
+                controller.connection.kernelSpec.name === kernelSpecName
```
rchiodo, line 276: I don't think we're matching on this anymore. The name is essentially the 'path' to the kernelspec in the jupyter kernelspec folder. I mean this would match in the old code, but now this should always be python3 AFAIK.

The match you have after this seems like a better match. But it should include path and name, not just path.

IanMatthewHuff: @rchiodo I pointed Joyce this direction here, not sure if I'm pointing things wrong. But here we are matching against an active interpreter. In the kernel finder code we getInterpreterKernelSpecName in the path to create kernelspecs to use for installed interpreters. So this generated name seemed like the safest match.

IanMatthewHuff: Might be because the interpreter that I have is registered via the old style? Looking now.

rchiodo: Yeah the old style used to generate a goofy hash for this name. We hopefully don't do this anymore.

IanMatthewHuff: @rchiodo This is the kernel connection metadata for an new conda env that I just created:
image
I think the hash is still there, but I'm looking at the implementation now, and I'm not sure if what I recommended is the right thing. The InterpreterKernelSpec name might just use the path for calculation of the name.

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
joyceerhl: Yeah we're working around a vscode-internal problem here.

PR #6386, opened 2021-06-26
Fix round-trip running and exporting markdown cells in native interactive window

joyceerhl:
Original implementation didn't account for needing to strip leading # before execution, and stripped # before export (which would result in a `# H1` heading becoming plain `H1` after export).

Also properly implemented firing of `closed` event inside the `IInteractiveWindow` and removed our hack in the `IInteractiveWindowProvider`.

### DIFF
--- a/src/client/datascience/cellFactory.ts
+++ b/src/client/datascience/cellFactory.ts
@@ -6,13 +6,14 @@ import '../common/extensions';
 import * as uuid from 'uuid/v4';
 import { NotebookDocument, Range, TextDocument, Uri } from 'vscode';
 
-import { parseForComments } from '../../datascience-ui/common';
+import { appendLineFeed, parseForComments } from '../../datascience-ui/common';
 import { createCodeCell, createMarkdownCell } from '../../datascience-ui/common/cellFactory';
 import { IJupyterSettings, Resource } from '../common/types';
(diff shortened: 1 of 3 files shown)


src/client/datascience/cellFactory.ts
```diff
@@ -194,19 +195,21 @@ export function generateCellsFromDocument(document: TextDocument, settings?: IJu
...
+    notebookDocument: NotebookDocument,
+    magicCommandsAsComments: boolean
+): ICell[] {
+    return notebookDocument
+        .getCells()
+        .filter((cell) => !cell.metadata.isSysInfoCell)
+        .map((cell) => {
+            // Reinstate cell structure + comments from cell metadata
+            let code = cell.document.getText().splitLines();
+            if (cell.metadata.interactiveWindowCellMarker !== undefined) {
```
IanMatthewHuff, line 208: Had to look this up quick (I missed it on the earlier review). Is this just to prevent showing the #%% line in the interactive window code block?

IanMatthewHuff (approved):

src/client/datascience/cellFactory.ts
joyceerhl: Yes, we need to remove it before sending it to the interactive window, then restore it when exporting.

PR #6390, opened 2021-06-28
Fix controller/active interpreter matching logic, implement `show`, run correct range of cells with `notebook.cell.execute`

joyceerhl:


rchiodo (approved):

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
```diff
@@ -245,37 +244,19 @@ export class NativeInteractiveWindowProvider implements IInteractiveWindowProvid
...
+            .find((controller) => {
                 return (
+                    // We register each of our kernels as two controllers
+                    // because controllers are currently per-viewtype. Find
+                    // the one for the interactive viewtype for now
+                    controller.controller.notebookType === InteractiveWindowView &&
                     controller.connection.kind === 'startUsingPythonInterpreter' &&
-                    controller.connection.interpreter?.path === activeInterpreter?.path
+                    controller.connection.interpreter?.path === activeInterpreter?.path &&
+                    controller.connection.interpreter.displayName === activeInterpreter.displayName
```
IanMatthewHuff, line 257: Is envName preferable over displayName? https://github.com/microsoft/vscode-python/blob/846eee870e8b7bab38172600836faedb5fb80166/src/client/activation/activationService.ts#L343

Not sure I full understand the depth of when one of the other will be available. But using a display name for any type of checking seems dicy.

IanMatthewHuff (approved):
Looks like for an interpreter display name is calculated with envName, so Display name should work fine here.

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
joyceerhl: `envName` is undefined for base environments while `displayName` doesn't seem to be, and I don't know enough about conda environments to know if you can have 2 base environments associated with the same Python executable but without an `envName`, but that seems hypothetically possible, which is why I ended up going with `displayName` here. But per your review comment below there would be a `displayName` collision for that case.

Issue #6391, opened 2021-06-28
Pending work for basic debugging


DavidKutu:
Work items pending from https://github.com/microsoft/vscode-jupyter/pull/6332

- ~~Add context for switching kernels and notebook opened~~
- ~~cache valid kernels so the button enables faster~~
- ~~don't make up a random session ids~~
- ~~delete temp files (the kernel is not doing this)~~
- ~~use python api for checking ipykernel version~~
- ~~use IDisposableRegistry~~
- ~~check if we need the vscode.debug.breakpoints on activate~~

PR #6396, opened 2021-06-28
fix canDebug contextkey

DavidKutu:
-Change canDebug flag when a controller is selected

For first item from #6391

### DIFF
--- a/src/client/datascience/commands/activeEditorContext.ts
+++ b/src/client/datascience/commands/activeEditorContext.ts
@@ -192,7 +192,7 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
     }) {
         const activeDoc = this.vscNotebook.activeNotebookEditor?.document;
         if (activeDoc === notebook) {
-            this.updateDebugContext(controller.connection.interpreter);
+            this.updateDebugContext(controller.connection.interpreter).ignoreErrors();
         }
     }
     private async updateDebugContext(interpreter?: PythonEnvironment) {
--- a/src/client/datascience/commands/activeEditorContext.ts
+++ b/src/client/datascience/commands/activeEditorContext.ts
@@ -92,6 +92,7 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
     }
     public dispose() {
         this.disposables.forEach((item) => item.dispose());
+        this.debugReadyInterpreters.clear();
     }
     public async activate(): Promise<void> {
         this.docManager.onDidChangeActiveTextEditor(this.onDidChangeActiveTextEditor, this, this.disposables);
--- a/src/client/datascience/commands/activeEditorContext.ts
+++ b/src/client/datascience/commands/activeEditorContext.ts
@@ -202,6 +202,9 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -141,7 +147,7 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
         this.updateNativeNotebookContext();
         this.updateNativeNotebookCellContext();
         this.updateMergedContexts();
-        void this.updateDebugContext(e);
+        void this.updateDebugContext(e?.notebook?.getMatchingInterpreter());
```
IanMatthewHuff, line 150: For an async like this we usually use .ignoreErrors() versus putting the void on, just for consistency in this function.

src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -179,25 +185,30 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
...
-                this.canDebug
-                    .set(await this.dependencyService.areDebuggingDependenciesInstalled(interpreter))
-                    .ignoreErrors();
+    private onNotebookControllerSelected({
+        controller
+    }: {
+        notebook: NotebookDocument;
+        controller: VSCodeNotebookController;
+    }) {
+        this.updateDebugContext(controller.connection.interpreter);
```
IanMatthewHuff, line 194: I think that you also need a check on if this is the same as the activeNotebookEditor document. It would usually be the case but there isn't a guarantee that this message is coming in from the active document. You might accidentally update with the interpreter from a different controller on a non-active notebook editor.

src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -88,6 +92,7 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
     }
     public dispose() {
         this.disposables.forEach((item) => item.dispose());
+        this.debugReadyInterpreters.clear();
```
IanMatthewHuff, line 95: You should only need to worry about disposables here. This is just a Map, it's going away in any case.

src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -179,25 +185,30 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
...
+        notebook: NotebookDocument;
+        controller: VSCodeNotebookController;
+    }) {
+        this.updateDebugContext(controller.connection.interpreter);
+    }
+    private async updateDebugContext(interpreter?: PythonEnvironment) {
+        if (!interpreter) {
+            this.canDebug.set(false).ignoreErrors();
+        } else {
+            let flag = this.debugReadyInterpreters.get(interpreter);
```
IanMatthewHuff, line 200: I'm actually a bit hesitant about this caching as is. If they don't have the right version installed they are going to install it from the terminal. Then it's going to continue to warn them. I'd like to get this in so we can test with it, but it's worth a follow up issue. Maybe the check could return a cached value and then start the full check each time.

IanMatthewHuff (approved):
I'm a bit leary on the caching. Might want to file an issue on that as it could lead to customer annoyance. But I'd like to get this in for testing for Endgame.

Issue #6457, opened 2021-06-30
Nothing happens when I click on collapse all


sandy081:
Testing https://github.com/microsoft/vscode/issues/127378

Here is my interactive window and clicking on collapse all action does not do anything

![image](https://user-images.githubusercontent.com/10746682/124008844-295ea380-d9dd-11eb-9ebc-5e5a6a94bc7b.png)

May be I do not have any collapsible cells here? If so, what is the way to have such cells.

IanMatthewHuff:
Validated.
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/75ec2569-fda6-4e6d-b78a-d1cd3b270f88)

Issue #6418, opened 2021-06-29
Interactive window makes cell markers disappear when a cell doesn't run


rchiodo:
Testing microsoft/vscode#127378

Open python file
Add cell marker
Execute cell

Execution is failing for me because of this error:
https://github.com/microsoft/vscode/issues/127525

But then cell markers disappear?

![image](https://user-images.githubusercontent.com/19672699/123848461-99c9cf80-d8cc-11eb-9c83-91cfd7323a69.png)

rebornix:
I'm seeing the same with webview interactive window.

joyceerhl:
Probably a dupe of https://github.com/microsoft/vscode-jupyter/issues/6882

PR #6469, opened 2021-07-01
Fix interactive window debug cell, codelens tracking, go to cell, expand/collapse all from toolbar when focus is not on notebook editor

joyceerhl:
For #6418, #6423, #6426, #6457

Debug Cell problem was because executeObservable has side effects (updating execution count) that the `notebook.cell.execute` codepath doesn't.

Go to cell problem was because notebook cell metadata wasn't populated.

Cell markers were caused by interactive window editor taking focus when it shouldn't (required fix in core).

Collapse cell problem was because clicking on a toolbar button does not set focus to the interactive window, and the command handler relied on the `activeNotebookEditor` instead of using the `interactive/toolbar` context.

Will be porting this PR to release branch.

### DIFF
--- a/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
+++ b/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
@@ -272,7 +272,13 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
         return result;
     }
 
-    private async submitCodeImpl(code: string, file: string, line: number, isDebug: boolean, notebookCell: NotebookCell) {
+    private async submitCodeImpl(
+        code: string,
+        file: string,
+        line: number,
+        isDebug: boolean,
+        notebookCell: NotebookCell
+    ) {
         const notebook = this.kernel?.notebook;
         if (!notebook) {
             return false;
@@ -297,9 +303,7 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
             const owningResource = this.owningResource;
             const id = uuid();
             const observable = this.kernel!.notebook!.executeObservable(code, file, line, id, false);
-            const temporaryExecution = this.notebookController!.controller.createNotebookCellExecution(
-                notebookCell
-            );
+            const temporaryExecution = this.notebookController!.controller.createNotebookCellExecution(notebookCell);
             temporaryExecution?.start();
 
             // Sign up for cell changes
@@ -484,7 +488,7 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
         throw new Error('Method not implemented.');
     }
 
-    public expandAllCells() {
+    public async expandAllCells() {
         const edit = new WorkspaceEdit();
         this.notebookDocument.getCells().forEach((cell, index) => {
             const metadata = {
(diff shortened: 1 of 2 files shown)


src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -225,16 +225,11 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
-        await this.addNotebookCell(code);
+        const notebookCell = await this.addNotebookCell(code, file, line);
         try {
-            await this.commandManager.executeCommand('notebook.cell.execute', {
-                ranges: [{ start: this.notebookDocument.cellCount - 1, end: this.notebookDocument.cellCount }],
-                document: this.notebookDocument.uri,
-                autoReveal: true
-            });
-            return true;
+            return this.submitCodeImpl(code, file.fsPath, 0, false, notebookCell);
```
IanMatthewHuff, line 232: I was actually having to look this up myself. But this try catch block now seems funny. If you're directly returning the promise from submitCodeImpl I don't believe that the try catch block is relevant anymore. Looks to me like a case of this here: https://stackoverflow.com/questions/38708550/difference-between-return-await-promise-and-return-promise/42750371#42750371

src/client/datascience/interactive-window/nativeInteractiveWindowCommandListener.ts
```diff
@@ -422,15 +430,19 @@ export class NativeInteractiveWindowCommandListener {
         }
     }
 
-    private expandAllCells() {
-        const interactiveWindow = this.interactiveWindowProvider.activeWindow;
+    private expandAllCells(uri?: Uri) {
+        const interactiveWindow = uri
+            ? this.interactiveWindowProvider.windows.find((window) => window.notebookUri!.toString() === uri.toString())
```
IanMatthewHuff, line 435: From the code it looks like notebookUri is always defined, but the "!" feels a tiny bit less typesafe here since it is optional on the interface. Would ? accomplish the same thing since undefined won't equal uri.toString()?

IanMatthewHuff: Ohh, looks like you have ? below for exportAs (was just reading on more). Can we use that in these locations as well.

src/client/datascience/interactive-window/nativeInteractiveWindowCommandListener.ts
```diff
@@ -454,7 +466,7 @@ export class NativeInteractiveWindowCommandListener {
 
     private export(uri?: Uri) {
         const interactiveWindow = uri
-            ? this.interactiveWindowProvider.windows.find((window) => window.notebookUri!.toString() === uri.toString())
+            ? this.interactiveWindowProvider.windows.find((window) => window.notebookUri?.toString() === uri.toString())
```
rchiodo, line 469: Maybe this 'find matching interactive window' could be on the intearctiveWindowProvider? And then everyone of these functions would use that code instead?

rchiodo (approved):

DavidKutu (approved):

IanMatthewHuff (approved):

src/client/datascience/interactive-window/nativeInteractiveWindowCommandListener.ts
joyceerhl: Unfortunately we're still using the `IInteractiveWindowProvider` interface which is shared for webview, so I'd have to put that method on the old provider too (and it doesn't make sense there). I tracked this item in a list of todos for webview code cleanup https://github.com/microsoft/vscode-jupyter/issues/6488 feel free to add more there as well :)

Issue #6479, opened 2021-07-01
Native Interactive - Collapse shouldn't collapse markdown cells.


greazer:
The webview IW did not do this, and in general, I don't think it's valuable to do. The reason we had expand/collapse at all in the old IW is because we default to showing a collapsed form of the source by default. Therefore there are times when the user wanted to expand all of the source look for something and then collapse all of it again.

Issue #6477, opened 2021-07-01
Native Interactive - Prepend cell marker (#%%) to copied code when Copy Code button is clicked.


greazer:
The idea behind adding the ability to copy a code cell that exists in the interactive window is that it can be used to take a particular piece of code that was being iterated on directly in the IW and paste it into the script the customer is building. Therefore, it's not unreasonable to assume the user will want to just "Copy Code" and paste the copied cell code *as a marked cell* in the script. Having to manually add the marker can be an annoyance. 

If we did this, then the "Copy Code" button should perhaps be called "Copy Cell". Note that if the user does not want the cell marker they can always do a selection (including ctrl-a to select all) within the Monaco editor and use Ctrl+c or Cmd+c. It's also easier to delete a line with an unwanted cell marker than it is to add one. 

Note that the webview based IW did not support this feature and it also did not readily support selecting lines of code from the code section of each cell. I believe *I* was the requestor for this functionality, so you could say I'm just a little biased with regard to how I think it should work. :D

joyceerhl:
Would you expect to see #%% on cells created from the interactive window input editor in addition to cells run from a Python file? FWIW I agree with @IanMatthewHuff I wouldn't want to see #%% at all. Could be a setting based on user feedback

PR #6480, opened 2021-07-01
Fix copy code cell marker, collapsing markdown cells

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/6479, https://github.com/microsoft/vscode-jupyter/issues/6477

### DIFF
--- a/package.json
+++ b/package.json
@@ -804,8 +804,8 @@
                 "category": "Jupyter"
             },
             {
-                "command": "jupyter.interactive.copyCode",
-                "title": "%DataScience.interactiveCopyCode%",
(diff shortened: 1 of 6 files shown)


IanMatthewHuff (approved):
Code looks fine. I'm tempted to say that I'm not sure that I would want the marker in when copying the code out. I would often be pasting just that section possibly into another existing file or cell section. But @greazer has strong feelings on this it seems :).

joyceerhl:
Yeah it should probably say Copy Cell, not Copy Code

rchiodo (approved):

PR #6503, opened 2021-07-01
Don't pass registrationData twice when registering notebook serializer

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/6501

### DIFF
--- a/src/client/common/application/notebook.ts
+++ b/src/client/common/application/notebook.ts
@@ -114,10 +114,7 @@ export class VSCodeNotebook implements IVSCodeNotebook {
         serializer: NotebookSerializer,
         options?: NotebookDocumentContentOptions
     ): Disposable {
-        return workspace.registerNotebookSerializer(notebookType, serializer, options, {
-            displayName: 'Jupyter',
-            filenamePattern: ['*.ipynb']
-        });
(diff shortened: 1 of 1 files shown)


rchiodo (approved):

src/client/common/application/notebook.ts
```diff
@@ -114,10 +114,7 @@ export class VSCodeNotebook implements IVSCodeNotebook {
         serializer: NotebookSerializer,
         options?: NotebookDocumentContentOptions
     ): Disposable {
-        return workspace.registerNotebookSerializer(notebookType, serializer, options, {
-            displayName: 'Jupyter',
-            filenamePattern: ['*.ipynb']
-        });
+        return workspace.registerNotebookSerializer(notebookType, serializer, options);
```
IanMatthewHuff, line 117: That's odd. Had to spelunk around in vscode core to see why this was an issue. Wonder why that option is even here? https://github.com/microsoft/vscode/blob/7adcbfdd68925543df362b74c7026017d0af3b24/src/vs/workbench/api/browser/mainThreadNotebook.ts#L119

IanMatthewHuff (approved):

Issue #6467, opened 2021-06-30
remove 'fixBreakpoints' code


DavidKutu:
this is being dealt with in VS Code

PR #6505, opened 2021-07-01
debugging work

DavidKutu:
For #6467
For #6391 its the following:
- don't make up a random session ids
- delete temp files (the kernel is not doing this)
- use IDisposableRegistry
- check if we need the vscode.debug.breakpoints on activate

### DIFF
--- a/src/client/debugger/jupyter/kernelDebugAdapter.ts
+++ b/src/client/debugger/jupyter/kernelDebugAdapter.ts
@@ -189,8 +189,8 @@ export class KernelDebugAdapter implements DebugAdapter {
         this.cellToFile.forEach((tempPath) => {
             const norm = path.normalize(tempPath);
             const dir = path.dirname(norm);
-            this.fs.deleteLocalFile(norm);
-            this.fs.deleteLocalDirectory(dir);
+            void this.fs.deleteLocalFile(norm);
+            void this.fs.deleteLocalDirectory(dir);
         });
     }
 
--- a/src/client/debugger/jupyter/kernelDebugAdapter.ts
+++ b/src/client/debugger/jupyter/kernelDebugAdapter.ts
@@ -189,8 +189,12 @@ export class KernelDebugAdapter implements DebugAdapter {
         this.cellToFile.forEach((tempPath) => {
             const norm = path.normalize(tempPath);
             const dir = path.dirname(norm);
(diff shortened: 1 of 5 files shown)


src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -181,6 +184,14 @@ export class KernelDebugAdapter implements DebugAdapter {
     dispose() {
         this.messageListener.forEach((ml) => ml.dispose());
         this.messageListener.clear();
+
+        // clean temp files
+        this.cellToFile.forEach((tempPath) => {
+            const norm = path.normalize(tempPath);
+            const dir = path.dirname(norm);
+            void this.fs.deleteLocalFile(norm);
```
rchiodo, line 192: Do you need to try catch on any of these?

rchiodo (approved):

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: good idea

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -19,8 +19,10 @@ import { IDebuggingCellMap, IJupyterSession } from '../../datascience/types';
 import { Kernel, KernelMessage } from '@jupyterlab/services';
 import { ICommandManager } from '../../common/application/types';
 import { traceError } from '../../common/logger';
+import { IFileSystem } from '../../common/platform/types';
 
 const debugRequest = (message: DebugProtocol.Request): KernelMessage.IDebugRequestMsg => {
+    const sessionId = message.arguments?.__sessionId ? message.arguments.__sessionId : randomBytes(8).toString('hex');
```
IanMatthewHuff, line 25: Hang on a second, shouldn't this be the Jupyter session ID?

DavidKutu: this will be sent to jupyter, and if the message came from jupyter the casting should put it there. But I'll test and make sure.

DavidKutu: so, sessionId only comes in the attach message. I think I should save it and send the same one but I'm not sure. I'll investigate.

Issue #6456, opened 2021-06-30
Restart kernel should be a modal dialog


sandy081:
Testing https://github.com/microsoft/vscode/issues/127378

Restart kernel shall be a modal dialog because user has to take an action to continue...

![image](https://user-images.githubusercontent.com/10746682/124008394-a9d0d480-d9dc-11eb-8c1d-2b748885af19.png)

joyceerhl:
We should discuss potentially moving the restart kernel command into VS Code core, as this is common to many kernels.

DonJayamanne:
here's the upstream issue with that feature request https://github.com/microsoft/vscode/issues/118783

joyceerhl:
Closing in favor of https://github.com/microsoft/vscode/issues/118783

PR #6517, opened 2021-07-02
Make restart kernel dialog modal

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/6456


rchiodo (approved):

joyceerhl:
Deferring this change because we probably want restart kernel and any notifications to be handled by VS Code core instead.

Issue #5833, opened 2021-05-11
Native Notebooks - Selecting base env from kernel picker does not activate


claudiaregio:
## Environment data

-   VS Code version: 1.56.0 (but opted into native notebook)
-   Jupyter Extension version (available under the Extensions sidebar): v2021.6.811652604
-   Python Extension version (available under the Extensions sidebar): v2021.5.829140558
-   Type of virtual environment used: conda
-   Jupyter server running: Local

## Steps to reproduce:

[**NOTE**: Self-contained, minimal reproducing code samples are **extremely** helpful and will expedite addressing your issue]

1. Open VS Code + notebook
2. Select anaconda base env from kernel picker 
3. Does not activate - keep getting error that package is not found although it shows up in the base env package list

Issue is solved by opening an anaconda prompt in the base environment and starting up vs code from there (code . )

DonJayamanne:
Based on the feedback from @claudiaregio this is blocked on VS Code https://github.com/microsoft/vscode-jupyter/issues/5872

DonJayamanne:
@claudiaregio does this now work for you?

claudiaregio:
This is working for me in insiders now!!!

claudiaregio:
Reopening as this is not working in stable post new release (1.57.0)

rchiodo:
This should be fixed I believe.

PR #6518, opened 2021-07-02
Fix problems with path not getting picked up during kernel launch

rchiodo:
For #5833

It seems 'Path' is not the only way that PATH environments can be set on Windows. Especially when picking up environments from .env files or from conda activations.

I was able to repro locally and I would end up with something like so:

```json
{
  "mergedEnv": {
     "Path": "C:\\users\\rich\\anaconda;c:\\OtherIMport",
     "PATH: "C:\\Windows\\System32
  }
}
```

Also strangely I had to downgrade to npm 6 to get npm install to work.


package.json
```diff
@@ -2290,6 +2290,7 @@
         "terser-webpack-plugin": "^3.1.0",
         "thread-loader": "^2.1.3",
         "transform-loader": "^0.2.4",
+        "trim": "^1.0.1",
```
rchiodo, line 2293: Not sure why but this was necessary in order for npm package to run.

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -341,7 +341,7 @@ export class KernelProcess implements IKernelProcess {
             traceInfo(`Launching Raw Kernel & not daemon ${this.launchKernelSpec.display_name} # ${executable}`);
             const [executionService, env] = await Promise.all([
                 this.processExecutionFactory.create(this.resource),
-                this.kernelEnvVarsService.getEnvironmentVariables(this.resource, this.launchKernelSpec)
+                this.kernelEnvVarsService.getEnvironmentVariables(this.resource, undefined, this.launchKernelSpec)
```
rchiodo, line 344: Interpreter not possible in this situation since this is a non python launch. I'll add a comment.

rchiodo:
Somebody else will likely have to merge this as I'm OOF next week.

src/client/datascience/kernel-launcher/kernelDaemon.ts
```diff
@@ -73,6 +73,8 @@ export class PythonKernelDaemon extends BasePythonDaemon implements IPythonKerne
         this.monitorOutput();
 
         if (this.preWarmed) {
+            // Making an assumption that the prewarmed kernel has the same environment.
+            traceInfo(`Using prewarmed kernel ...`);
```
IanMatthewHuff, line 77: Doesn't quite seem related, but looks like good logging, so leaving in.

src/client/common/variables/environment.ts
```diff
@@ -69,11 +54,16 @@ export class EnvironmentVariablesService implements IEnvironmentVariablesService
             return vars;
         }
 
-        const variable = vars ? vars[variableName] : undefined;
+        // It's been shown that the 'path' variable can have multiple casing even on the same platform
+        // depending upon where the environment variable comes from (kernelspec might have 'PATH' whereas windows might use 'Path')
+        const variableNameLower = variableName.toLowerCase();
+        const matchingKey = vars ? Object.keys(vars).find((k) => k.toLowerCase() == variableNameLower) : undefined;
+        const variable = vars && matchingKey ? vars[matchingKey] : undefined;
+        const setKey = matchingKey || variableName;
```
IanMatthewHuff, line 62: Triple checking this bit here, but looks right to me. The current path casing in the target environment block (as long as it exists) wins out for the final casing (so PATH stays as PATH, Path stays as Path) but new values of any casing get appended to it.

src/client/datascience/kernel-launcher/kernelProcess.ts
IanMatthewHuff: Un-resolving comment here. This looks correct here, but the reading is a bit funny (to me) since the daemon service is not always supported. However for the non-deamon cases the PythonKernelLauncherDaemon does return a non-daemon exeObs, so this here is the non-interpreter case.

IanMatthewHuff (approved):

src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
```diff
@@ -27,66 +23,60 @@ export class KernelEnvironmentVariablesService {
...
+        kernelSpec: IJupyterKernelSpec
+    ) {
         let kernelEnv = kernelSpec.env && Object.keys(kernelSpec.env).length > 0 ? kernelSpec.env : undefined;
         if (!kernelSpec.interpreterPath) {
             traceInfo(
                 `No custom variables for Kernel as interpreter path is not defined for kernel ${kernelSpec.display_name}`
             );
             return kernelEnv;
         }
-        const interpreter = await this.interpreterService
```
DonJayamanne, line 40: This is required for non-python environments like Java & the like that belong to conda environments.
Basically we need to actiavte the conda environment & then run the non-python kernels, tha'ts what the error message in line 43 means.
We really need to fix/re-enable the java kernel test on CI, that would have caught this

src/client/datascience/kernel-launcher/kernelProcess.ts
DonJayamanne: > in this situation since this is a non python launch. I'll add a comment.

This isn't correct, we need the interpreter here
If launching a java kernel from a conda environment, we need the associated conda environment to be actiavted.

That's whats happening here https://github.com/microsoft/vscode-jupyter/pull/6518/files#diff-c2b62e495d0dabb0ddd0d2418166b8d017dc981428b3916a3ea8e989eeb6a803L43

DonJayamanne (changes requested):

src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
IanMatthewHuff: @DonJayamanne Sounds good, I can update this. Also reenabling the Java test seems like a great candidate for debt week. I don't see an explicitly disabled test for that in non-python kernels. Maybe we need to add one? Does the beakerX java kernel run from a conda env?

src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
```diff
@@ -18,75 +17,79 @@ export class KernelEnvironmentVariablesService {
...
+    ) {
         let kernelEnv = kernelSpec.env && Object.keys(kernelSpec.env).length > 0 ? kernelSpec.env : undefined;
-        if (!kernelSpec.interpreterPath) {
-            traceInfo(
-                `No custom variables for Kernel as interpreter path is not defined for kernel ${kernelSpec.display_name}`
-            );
-            return kernelEnv;
+
+        // If an interpreter was not explicitly passed in, check for an interpreter path in the kernelspec to use
+        if (!interpreter) {
```
IanMatthewHuff, line 36: I kept the addition of the interpreter parameter on this function. For the case where we are launching with a known pythonEnv it seems better to use the actual launch interpreter that we will be using as opposed to the one in the kernelSpec interpreterPath. but added back the old code path for any case where the launch interpreter is not already calculated.

src/test/common/variables/kernelEnvVarsService.unit.test.ts
```diff
@@ -0,0 +1,119 @@
...
+                .returns(() => Promise.resolve({ PATH: 'foobaz' }));
+
+            const vars = await kernelVariablesService.getEnvironmentVariables(undefined, interpreter, kernelSpec);
+            const processPath = Object.keys(process.env).find((k) => k.toLowerCase() == 'path');
+            expect(processPath).to.not.be.undefined;
+            expect(vars).to.not.be.undefined;
+            expect(vars![processPath!]).to.be.equal(`foobar${path.delimiter}foobaz`);
+        });
+
+        test('KernelSpec interpreterPath used if interpreter is undefined', async () => {
```
IanMatthewHuff, line 96: Added quick test case for interpreter = undefined case.

src/test/common/variables/kernelEnvVarsService.unit.test.ts
```diff
@@ -0,0 +1,119 @@
...
+                .returns(() => Promise.resolve({ pATh: 'foobar' }));
+            customVariablesService
+                .setup((c) => c.getCustomEnvironmentVariables(TypeMoq.It.isAny()))
+                .returns(() => Promise.resolve({ PATH: 'foobaz' }));
+
+            const vars = await kernelVariablesService.getEnvironmentVariables(undefined, interpreter, kernelSpec);
+            const processPath = Object.keys(process.env).find((k) => k.toLowerCase() == 'path');
+            expect(processPath).to.not.be.undefined;
+            expect(vars).to.not.be.undefined;
+            expect(vars![processPath!]).to.be.equal(`foobar${path.delimiter}foobaz`);
```
IanMatthewHuff, line 93: Fix here in this update to use path.delimiter as test was failing on my machine due to : versus ;.

src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
DonJayamanne: > citly disabled test for that in non-python kernels. Maybe we need to add one? Does the beakerX java kernel run from a conda env?

I thought we had a beakerx java kernel in our tests. I'm certain we have it.

IanMatthewHuff: We had one. It just opens the notebook though. I don't see a disabled test to actually run it. I'll add an issue for this test.

IanMatthewHuff: Added #6544

DonJayamanne (approved):

PR #6519, opened 2021-07-02
Allow notebook serializer bits to be used for web

joyceerhl:
`Buffer` isn't defined in web workers

### DIFF
--- a/src/client/datascience/notebook/helpers/helpers.ts
+++ b/src/client/datascience/notebook/helpers/helpers.ts
@@ -445,7 +445,7 @@ function translateDisplayDataOutput(
     const data: Record<string, any> = output.data || {};
     // eslint-disable-next-line
     for (const key in data) {
-        items.push(new NotebookCellOutputItem(convertJupyterOutputToBuffer(key, data[key]), key));
+        items.push(convertJupyterOutputToBuffer(key, data[key]));
     }
 
     return new NotebookCellOutput(sortOutputItemsBasedOnDisplayOrder(items), metadata);
@@ -557,29 +557,30 @@ function convertOutputMimeToJupyterOutput(mime: string, value: Uint8Array) {
         return '';
     }
     try {
-        const stringValue = Buffer.from(value as Uint8Array).toString('utf8');
+        const stringValue = new TextDecoder().decode(value.buffer.slice(value.byteOffset));
         if (mime === CellOutputMimeTypes.error) {
-            traceInfo(`Concerting ${mime} from ${stringValue}`);
             return JSON.parse(stringValue);
         } else if (mime.startsWith('text/') || textMimeTypes.includes(mime)) {
             return splitMultilineString(stringValue);
         } else if (mime.startsWith('image/') && mime !== 'image/svg+xml') {
             // Images in Jupyter are stored in base64 encoded format.
             // VS Code expects bytes when rendering images.
-            return Buffer.from(value).toString('base64');
+            // https://developer.mozilla.org/en-US/docs/Glossary/Base64#solution_1_%E2%80%93_escaping_the_string_before_encoding_it
+            return btoa(encodeURIComponent(stringValue).replace(/%([0-9A-F]{2})/g, function(_match, p1) {
+                return String.fromCharCode(Number.parseInt('0x' + p1));
+            }));
         } else if (mime.toLowerCase().includes('json')) {
             return stringValue.length > 0 ? JSON.parse(stringValue) : stringValue;
         } else {
             return stringValue;
         }
     } catch (ex) {
-        traceError(`Failed to convert ${mime} output from a buffer ${typeof value}, ${value}`, ex);
         return '';
     }
 }
-function convertJupyterOutputToBuffer(mime: string, value: unknown): Buffer {
+function convertJupyterOutputToBuffer(mime: string, value: unknown): NotebookCellOutputItem {
     if (!value) {
-        return Buffer.from('');
+        return NotebookCellOutputItem.text('', mime);
     }
     try {
         if (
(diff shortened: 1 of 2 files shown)


src/client/datascience/notebook/notebookSerializer.ts
```diff
@@ -27,7 +27,9 @@ export class NotebookSerializer implements VSCNotebookSerializer {
         @inject(NotebookCellLanguageService) private readonly cellLanguageService: NotebookCellLanguageService
     ) {}
     public deserializeNotebook(content: Uint8Array, _token: CancellationToken): NotebookData {
-        const contents = Buffer.from(content).toString();
+        let contents = new TextDecoder().decode(content.buffer);
+		// Decoded string appears to contain $dataToNotebook + unicode junk? Skip past the junk
```
rchiodo, line 31: I wonder why Buffer skips it? Shouldn't it have the same problem?

rchiodo: Is there a start index perhaps on the content?

joyceerhl: It might be a web worker extension host problem moreso than a TextDecoder vs Buffer problem. $dataToNotebook is a function in VS Code source

rchiodo: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/byteOffset?

rchiodo (approved):

src/client/datascience/notebook/helpers/helpers.ts
```diff
@@ -557,51 +557,54 @@ function convertOutputMimeToJupyterOutput(mime: string, value: Uint8Array) {
         return '';
     }
     try {
-        const stringValue = Buffer.from(value as Uint8Array).toString('utf8');
+        const stringValue = new TextDecoder().decode(value.buffer.slice(value.byteOffset));
```
DonJayamanne, line 560:
```suggestion
        const stringValue = new TextDecoder().decode(value);
```

src/client/datascience/notebook/helpers/helpers.ts
```diff
@@ -557,51 +557,54 @@ function convertOutputMimeToJupyterOutput(mime: string, value: Uint8Array) {
...
         if (mime === CellOutputMimeTypes.error) {
-            traceInfo(`Concerting ${mime} from ${stringValue}`);
             return JSON.parse(stringValue);
         } else if (mime.startsWith('text/') || textMimeTypes.includes(mime)) {
             return splitMultilineString(stringValue);
         } else if (mime.startsWith('image/') && mime !== 'image/svg+xml') {
             // Images in Jupyter are stored in base64 encoded format.
             // VS Code expects bytes when rendering images.
-            return Buffer.from(value).toString('base64');
+            // https://developer.mozilla.org/en-US/docs/Glossary/Base64#solution_1_%E2%80%93_escaping_the_string_before_encoding_it
```
DonJayamanne, line 568: Can we use an npm package for somethign like this

src/client/datascience/notebook/helpers/helpers.ts
```diff
@@ -557,51 +557,54 @@ function convertOutputMimeToJupyterOutput(mime: string, value: Uint8Array) {
...
         } else {
             return stringValue;
         }
     } catch (ex) {
-        traceError(`Failed to convert ${mime} output from a buffer ${typeof value}, ${value}`, ex);
         return '';
     }
 }
-function convertJupyterOutputToBuffer(mime: string, value: unknown): Buffer {
+function convertJupyterOutputToBuffer(mime: string, value: unknown): NotebookCellOutputItem {
```
DonJayamanne, line 583: Might want to rename the function.

src/client/datascience/notebook/notebookSerializer.ts
```diff
@@ -27,7 +27,7 @@ export class NotebookSerializer implements VSCNotebookSerializer {
         @inject(NotebookCellLanguageService) private readonly cellLanguageService: NotebookCellLanguageService
     ) {}
     public deserializeNotebook(content: Uint8Array, _token: CancellationToken): NotebookData {
-        const contents = Buffer.from(content).toString();
+        const contents = new TextDecoder().decode(content.buffer.slice(content.byteOffset));
```
DonJayamanne, line 30:
```suggestion
        const contents = new TextDecoder().decode(content);
```

src/client/datascience/notebook/helpers/helpers.ts
```diff
@@ -557,51 +557,54 @@ function convertOutputMimeToJupyterOutput(mime: string, value: Uint8Array) {
...
         } else {
             // For everything else, treat the data as strings (or multi-line strings).
             value = Array.isArray(value) ? concatMultilineString(value) : value;
-            return Buffer.from(value as string);
+            return NotebookCellOutputItem.text(value as string, mime);
         }
     } catch (ex) {
-        traceError(`Failed to convert ${mime} output to a buffer ${typeof value}, ${value}`, ex);
-        return Buffer.from('');
+        return NotebookCellOutputItem.error(ex);
```
DonJayamanne, line 607: Not sure about this. I guess its ok, but we might want to change the error message.
If a user is expecting `x`, and now sees `errorl....` they won't know that its due to a failure in loading the notebook.
Personally i'd prefer to not display the output, instead of displaying an error output which would be even more confusing (also when saving the notebook, it will be saved as an error output, which would be wrong).

DonJayamanne (changes requested):

src/client/datascience/notebook/notebookSerializer.ts
joyceerhl: This is actually required, because `content.byteOffset` is not 0 (IIRC it's 3). If we ignore the `byteOffset` the resultant string is not parseable (will thrown JSON parse error)

joyceerhl:
Closing in favor of built-in ipynb support

PR #6561, opened 2021-07-07
Promote Julia extension for Julia notebooks

DonJayamanne:
Will need to be merged & released once Julia is ready to ship their extension

### DIFF
--- a/src/client/common/experiments/service.ts
+++ b/src/client/common/experiments/service.ts
@@ -115,7 +115,8 @@ export class ExperimentService implements IExperimentService {
         // https://github.com/microsoft/vscode-jupyter/issues/4771
         if (
             experiment === ExperimentGroups.NativeNotebook &&
-            this.extensions.getExtension('ms-dotnettools.dotnet-interactive-vscode')
+            (this.extensions.getExtension('ms-dotnettools.dotnet-interactive-vscode') ||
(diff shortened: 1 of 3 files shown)


joyceerhl (approved):

src/client/common/experiments/service.ts
```diff
@@ -115,7 +115,8 @@ export class ExperimentService implements IExperimentService {
         // https://github.com/microsoft/vscode-jupyter/issues/4771
         if (
             experiment === ExperimentGroups.NativeNotebook &&
-            this.extensions.getExtension('ms-dotnettools.dotnet-interactive-vscode')
+            (this.extensions.getExtension('ms-dotnettools.dotnet-interactive-vscode') ||
```
IanMatthewHuff, line 118: This looks like it could look at distinct values from extensionsThatSupportJupyterKernelLanguages possibly if we just wanted to add the extension names in one place. Or maybe just have the two extension names as constants.

IanMatthewHuff (approved):

src/client/common/experiments/service.ts
DonJayamanne: I'm not too concerned about this, as we'll be going 100% with native editor soon

DavidKutu (approved):

Issue #6496, opened 2021-07-01
Native Interactive - After executing cell script, it's missing the Goto [#] codelens


greazer:
- Execute any cell in a python script
- After it runs, you should see a new "Goto [#]" that when clicked will scroll the interactive window to that cell.

joyceerhl:
This is happening because for native interactive we aren't updating the `cellExecutionCounts` in the `this.notebookData` map here:

https://github.com/microsoft/vscode-jupyter/blob/2253b23239520a89310ec944f51f54db9d783b14/src/client/datascience/editor-integration/codeLensFactory.ts#L481-L485

So this check always fails:

https://github.com/microsoft/vscode-jupyter/blob/2253b23239520a89310ec944f51f54db9d783b14/src/client/datascience/editor-integration/codeLensFactory.ts#L500-L507

Also we haven't implemented the ScrollToCell handler for native interactive: 

https://github.com/microsoft/vscode-jupyter/blob/d06d8d8b0c2e8d7db45fcb6dc4434cdb313e6271/src/client/datascience/interactive-window/nativeInteractiveWindow.ts#L504-L506

PR #6597, opened 2021-07-09
Fix Go To codelens not appearing

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/6496

This PR requires upstream changes to VS Code to fully work (https://github.com/microsoft/vscode/pull/128274, https://github.com/microsoft/vscode/pull/128373), because extensions need a way to activate an already-visible interactive window (and then thereafter we can use `activeNotebookEditor.revealRange` to scroll to the right cell). The go-to-cell behavior in the current webview interactive window activates the interactive window panel when scrolling to the matching cell.

### DIFF
--- a/src/client/datascience/editor-integration/codeLensFactory.ts
+++ b/src/client/datascience/editor-integration/codeLensFactory.ts
@@ -272,6 +272,11 @@ export class CodeLensFactory implements ICodeLensFactory, IInteractiveWindowList
         args.notebook.onDisposed(() => {
             this.notebookData.delete(key);
         });
+        if (args.notebook.onDidFinishExecuting !== undefined) {
+            args.notebook.onDidFinishExecuting((cell: ICell) => {
+                this.updateExecutionCounts(args.identity, cell);
+            });
+        }
     }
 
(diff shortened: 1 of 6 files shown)


src/client/datascience/editor-integration/codeLensFactory.ts
```diff
@@ -272,6 +272,11 @@ export class CodeLensFactory implements ICodeLensFactory, IInteractiveWindowList
         args.notebook.onDisposed(() => {
             this.notebookData.delete(key);
         });
+        if (args.notebook.onDidFinishExecuting !== undefined) {
+            args.notebook.onDidFinishExecuting((cell: ICell) => {
+                this.updateExecutionCounts(args.identity, cell);
+            });
+        }
```
joyceerhl, line 279: Unfortunately codeLensFactory.ts uses `ICell` all over the place. We should rewrite the whole thing to just work with `NotebookCell`s and change how we execute cells for the interactive window to execute `NotebookCell`s rather than using the old `executeObservable` codepath.

DavidKutu (approved):

PR #6607, opened 2021-07-09
use python api for checking ipykernel version and cache kernels

DavidKutu:
For #6391
- cache valid kernels so the button enables faster
- use python api for checking ipykernel version


src/client/api/pythonApi.ts
```diff
@@ -316,6 +317,19 @@ export class PythonInstaller implements IPythonInstaller {
...
     }
+
+    public async isProductVersionCompatible(
+        product: Product,
+        semVerRequirement: string,
+        resource?: InterpreterUri
+    ): Promise<ProductInstallStatus> {
+        try {
+            const api = await this.apiProvider.getApi();
+            return await api.isProductVersionCompatible(product, semVerRequirement, resource);
```
DonJayamanne, line 328: You don't need the await
```suggestion
            return api.isProductVersionCompatible(product, semVerRequirement, resource);
```

src/client/api/pythonApi.ts
```diff
@@ -316,6 +317,19 @@ export class PythonInstaller implements IPythonInstaller {
...
+        } catch (ex) {
```
DonJayamanne, line 329: If we're not doing anything with the exception, then ther'es no need to capture this.

src/client/api/types.ts
```diff
@@ -92,6 +92,14 @@ export type PythonApi = {
         cancel?: CancellationToken,
         reInstallAndUpdate?: boolean
     ): Promise<InstallerResponse>;
+    /**
+     * IInstaller
+     */
+    isProductVersionCompatible(
+        product: Product,
+        semVerRequirement: string,
+        resource?: InterpreterUri
```
DonJayamanne, line 101: I would change this to `PythonEnvironment` instead of `InterpreterUri`, thats an old approach, and very vague that has caused a lot of issues, We should be more specific about the target interpreter.

src/client/datascience/jupyter/kernels/kernelDependencyService.ts
```diff
@@ -87,13 +90,19 @@ export class KernelDependencyService implements IKernelDependencyService {
...
     ): Promise<boolean> {
         try {
-            const version = await this.installer.getVersion(Product.ipykernel, interpreter);
-            if (version) {
-                const versionSplit = version.split('.');
-                const mainVersionNumber = Number(versionSplit[0]);
-                return mainVersionNumber >= 6;
+            const result = await this.pythonInstaller.isProductVersionCompatible(
+                Product.ipykernel,
+                '>=6.0.0',
```
DonJayamanne, line 95: This isn't a valid semver.
We should ideally just pass `6.0.0` and then on the other end use the necessary API to check the minimum version by prefixing wqith `>=` or the like.

DavidKutu: It is valid and it works.
There's an example here: https://docs.npmjs.com/cli/v6/using-npm/semver#ranges

src/client/datascience/jupyter/kernels/kernelDependencyService.ts
```diff
@@ -87,13 +91,16 @@ export class KernelDependencyService implements IKernelDependencyService {
         _token?: CancellationToken
     ): Promise<boolean> {
         try {
-            const version = await this.installer.getVersion(Product.ipykernel, interpreter);
-            if (version) {
-                const versionSplit = version.split('.');
-                const mainVersionNumber = Number(versionSplit[0]);
-                return mainVersionNumber >= 6;
+            const installer = await this.serviceContainer.get<IPythonInstaller>(IPythonInstaller);
+            const result = await installer.isProductVersionCompatible(Product.ipykernel, '>=6.0.0', interpreter);
```
DonJayamanne, line 95: I would change this to `6.0.0` and change the code at the other end to add the `>=` else its incorrect to pass `>=6.0.0`, i.e. if you are saying that minimum is `6.0.0`, then why do we need `>=`.

DonJayamanne: The text `>=` is actually part of the code used to check for the version compatiblity. (its very specific to the tool & mechanism used to check the version).

DavidKutu: What other end? The python extension? That doesn't make sense, the function exposed asks for the semVerRequirement. It doesn't make sense to give it half the string in one place and the rest in another.

Passing `6.0.0` is wrong, that means 'be exactly `6.0.0`', that's not what we want. From their documentation, the parameter should be `>=6.0.0` which means, `6.0.0` or above.

DavidKutu: passing just `6.0.0` doesn't work, it returns 'needs upgrade'

DonJayamanne: The parameter is `minimum` version, and that's it should be, we need to pass the minimum version as the semver value.
Using `>=` is a way to check via pip whether the version is minimum or not. If you were using npm, then `>=` doens't work, similarly it could differ from different tools.
Again, the requirement is `what is the minimum version`, not, what is the query that needs to be run to check the minimum version.

you are providing the arguments to some CLI argument, instead of providing the minimum version value as semver. Hence that's incorrect.

DonJayamanne: I.e. in Python you need to change the following ` if (semver.satisfies(version, semVerRequirement)) {` as follows:
` if (semver.satisfies(version, ${`>=${semVerRequirement}`)) {`

and change the argument `semVerRequirement` to be `mimimumVersion: Semver` (this way its not just any arbitrary string, but a proper version.
Else someone looking at this might pass `6.0.0` and assume that works.

DavidKutu: I understand your comment, but I don't think this is confusing. The parameter clearly reads `semVerRequirement` and not `semVer` or `minimumVersion`.

Also, doing what you're asking means loosing functionality of `isProductVersionCompatible`. It would only be able to check for a minimum version, and the time might come when we need to check a maximum version, an exact version, or a range. We can do all that with the function as it is right now.

rchiodo (approved):

PR #6642, opened 2021-07-12
Refactor preferred controller for interactive

DonJayamanne:


src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
```diff
@@ -238,19 +236,8 @@ export class NativeInteractiveWindowProvider implements IInteractiveWindowProvid
     }
 
     private async getControllerForInteractiveWindow(): Promise<string | undefined> {
-        // Fetch the active interpreter and use the matching controller
-        const api = await this.pythonApi.getApi();
```
DonJayamanne, line 242: The logic for what's the preferred should be in the manager (that's where it is today for native notebooks).

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -86,6 +87,16 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         this.disposables.push(this._onNotebookControllerSelected);
         this.isLocalLaunch = isLocalLaunch(this.configuration);
     }
+    public async getInteractiveController(): Promise<VSCodeNotebookController | undefined> {
+        // Fetch the active interpreter and use the matching controller
+        const api = await this.pythonApi.getApi();
+        const activeInterpreter = await api.getActiveInterpreter();
```
DonJayamanne, line 93: This change also makes it easier for https://github.com/microsoft/vscode-jupyter/issues/5861
Else we have to change a few places, when ideally we should have the controllers created & return from a single place.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -86,6 +87,16 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+
+        if (!activeInterpreter) {
+            return;
+        }
+        this.getOrCreateController(activeInterpreter, InteractiveWindowView)?.id;
```
joyceerhl, line 98:
```suggestion
        return this.getOrCreateController(activeInterpreter, InteractiveWindowView);
```

joyceerhl (changes requested):
https://github.com/microsoft/vscode-jupyter/pull/6642#discussion_r668214535

rchiodo (approved):

src/client/datascience/notebook/notebookControllerManager.ts
joyceerhl: We need this change, otherwise the controller will never be returned from `getInteractiveController`.

joyceerhl (approved):

Issue #6209, opened 2021-06-08
Run to cursor for first entry on run by line


rchiodo:
This changes our logic for 'injecting' a breakpoint by making it go into the middle of the cell. Not sure how ipykernel handles breakpoints though. This might be tricky to inject a breakpoint in the middle.

Issue #6208, opened 2021-06-08
Add icon and command for run by line


rchiodo:
Just adding the icon to the toolbar and enabling when it is allowed.

PR #6655, opened 2021-07-13
Run by Line

DavidKutu:
For #6208, #6209


package.json
```diff
@@ -24,7 +24,7 @@
         "theme": "light"
     },
     "engines": {
-        "vscode": "1.59.0-insider"
+        "vscode": "^1.58.0"
```
rchiodo, line 27: I believe this is a merge issue?

package.json
```diff
@@ -281,7 +281,28 @@
                 "title": "Debug",
                 "icon": "$(bug)",
                 "category": "Jupyter",
-                "enablement": "notebookKernelCount > 0 && !jupyter.notebookeditor.debuggingInProgress && config.jupyter.experimental.debugging"
+                "enablement": "notebookKernelCount > 0 && !jupyter.notebookeditor.debuggingInProgress && !jupyter.notebookeditor.runByLineInProgress && config.jupyter.experimental.debugging"
```
rchiodo, line 284: Could the 'debuggingInProgress' context key be true when run by line is in progress? So that the enablement clause here can be shorter?

package.json
DavidKutu: yes, sorry

package.json
DavidKutu: It can't, by we don't want users to be able to press RBL while debugging and vice versa.
Also we don't want them to be able to press RBL if they're already RBLing.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -27,15 +34,16 @@ class Debugger {
...
+        const name = cell ? `${document.uri.toString()}?RBL=${cell.index}` : document.uri.toString();
         this.session = new Promise<DebugSession>((resolve, reject) => {
             this.resolveFunc = resolve;
             this.rejectFunc = reject;
 
             debug
                 .startDebugging(undefined, {
                     type: DataScience.pythonKernelDebugAdapter(),
-                    name: `${path.basename(document.uri.toString())}`,
+                    name: `${path.basename(name)}`,
```
rchiodo, line 46: path.basename works with ```?``` in it? I mean I'd think that would work but seems like an invalid file name?

DavidKutu: it does work, I use it to differentiate between run by line and normal debugging.

I'm open to change it if you have a suggestion

DavidKutu: nvm, I'll use it the same, but will use path.basename only on the doc URI

rchiodo (approved):

Issue #5864, opened 2021-05-13
What do we do if user does not have Python when attempting to run Python notebooks


DonJayamanne:
@claudiaregio /cc
Looks like we're running into this a lot

rchiodo:
Ideas?
- Use piodide. Yeah okay this is silly but might be good enough?
- Offer to install python (well bring up a link to python.org)
- Warn them that the kernel won't work

DonJayamanne:
> arn them that the kernel won't work

Shouldn't we do the same for other non-python kernels?

rchiodo:
> Shouldn't we do the same for other non-python kernels? Let them know that they do not have a kernel installed that can handle this (only for non-pytohn kernels)

Yeah that sounds like a good idea.

joyceerhl:
@DonJayamanne in this case, wouldn't the user not have any kernel selected for the current notebook? How is this a kernel start issue then?

greazer:
We need at minimum an error message that gives some direction to the user. Especially if they don't have Python installed at all.

DonJayamanne:
This will require a lot of changes, hence moving back to backlog and leaving assigned so we discuss the technical implementation.

rchiodo:
We could
- always return the python default kernel
- always have the same id for this default kernel (per URI?) same guid forever?
- update this with the state of the actual preferred one when we find it

DonJayamanne:
* If python is not installed have a kernel named `Python 3`
    * When user selects this and attempts to run the cell they'll be asked to either install Python extension or Python runtime
    * After installing Python extension (if python runtime is installed) user should not have to reload VS Code (the interpreters should be listed)
    * After installing Python, user will need to reload VS Code
    * TODO: File for API when this is detected by core extension.
* For non python kernels, if the kernelspec cannot be found, then add a placeholder for that language
    * User will know that their kernel isn't available when they attempt to run. Hence no point stating the obvious via some message.

DonJayamanne:
Upstream feature requests:
* https://github.com/microsoft/vscode-python/issues/16676
* https://github.com/microsoft/vscode-python/issues/16675

PR #6657, opened 2021-07-13
Display placeholder kernel if Python not found

DonJayamanne:
For #5864


src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -184,11 +195,47 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+        if (!listLocalNonPythonKernels) {
+            if (connections.some((item) => isPythonKernelConnection(item))) {
+                this.removeNoPythonControllers();
+            } else {
+                this.regsiterNoPythonControllers();
+            }
+        }
     }
+    private removeNoPythonControllers() {
+        this.notbeookNoPythonController?.dispose();
```
rchiodo, line 210: Does it get updated after the python extension is installed/interpreters are found?

rchiodo (approved):

src/client/datascience/notebook/notebookControllerManager.ts
DonJayamanne: Yes, existing code (if you don't have python installed & then install it, we re-query list of kernels)

Issue #6603, opened 2021-07-09
Always Scroll On new cell setting not working in new native Interactive mode


jonvlcs07:
-   VS Code version: 1.58.0
-   Jupyter Extension version: 2021.8.0

In the new native interactive mode with the setting `Jupyter: Always Scroll On New Cell` enabled, when I'm running cells it doesn't scroll down to newest cell in the interactive window as it should.

joyceerhl:
We actually need a way to scroll to show the output, not the cell input.

![image](https://user-images.githubusercontent.com/30305945/125119612-38f97e80-e0a6-11eb-8c6e-a6fb0d50d0c8.png)

I don't believe we have a way to do this yet:

![image](https://user-images.githubusercontent.com/30305945/125119706-54648980-e0a6-11eb-8438-331e3f0f9b93.png)

StevenLi-DS:
This is missing indeed and the notebook UI not scroll at all.

joyceerhl:
@jonvlcs07 This should be fixed in tonight's build, please let us know if it works for you 😊

jonvlcs07:
Hey, it's still not scrolling down to the latest output in the interactive window when `jupyter.enableNativeInteractiveWindow` is set and I'm running cells.

Keeptg:
> 
> 
> Hey, it's still not scrolling down to the latest output in the interactive window when `jupyter.enableNativeInteractiveWindow` is set and I'm running cells.

It doesn't work to me too!

joyceerhl:
Sorry, I should have been more specific. It should be fixed in the latest Jupyter build for VS Code Insiders, or our next stable release at the beginning of August.

rzhao271:
This works on Insiders.

Issue #6494, opened 2021-07-01
Native Interactive (and WebView) - Run several cells at one to start an interactive session, some cells don't run and are out of order


greazer:
I believe this is a dupe, but I'm not sure if that dupe was readily reproable. This one seems to be.
- Open a Python script file with several cells that are distinguishable from each other, ala
```
# %%
print('1')
# %%
print('2')
# %%
print('3')
# %%
print('4')
```
- Run each line of code quickly via Shift+Enter
- Wait until processing is done.

Result
Some cells not run and definitely out of order. See...
![image.png](https://images.zenhubusercontent.com/5bda244bb6c4f1227d4a2d8e/d2e7b492-5d6c-4a48-b092-4679d64f602b)
-

joyceerhl:
I was able to fix the problem where some cells don't run, however the out of order problem is easily reproducible even with webview interactive window:

![image](https://user-images.githubusercontent.com/30305945/125558649-2f518d02-3921-4147-a196-a17c59302edc.png)

PR #6658, opened 2021-07-14
Fix race conditions when running multiple cells in native interactive window from codelens

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/6494 and partially https://github.com/microsoft/vscode-jupyter/issues/6603


src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
```diff
@@ -109,69 +110,83 @@ export class NativeInteractiveWindowProvider implements IInteractiveWindowProvid
...
+        if (resource === undefined) {
+            return this.createInteractiveWindowPromise(resource, mode);
+        }
 
-        return result;
+        const existingPromise = this.mapOfResourcesToInteractiveWindowPromises.get(resource.toString());
+        // If there's a resource and no existing promise, create a new one
+        if (existingPromise === undefined) {
+            const promise = this.createInteractiveWindowPromise(resource, mode);
+            this.mapOfResourcesToInteractiveWindowPromises.set(resource.toString(), promise);
```
joyceerhl, line 184: Previously interactive window creation was asynchronous and could take a while, so if you ran a bunch of cells it was possible for the previous interactive window to still be in the process of initializing, and this caused us to try and create two interactive windows for the same submitting file, which uncovered a race condition in VS Code where the editor input might not have a notebook model initialized yet. This ensures that we reuse the existing init promise if a bunch of requests from the same file come in at once.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -664,14 +664,13 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
         };
-        edit.replaceNotebookCells(
-            this.notebookDocument.uri,
-            new NotebookRange(this.notebookDocument.cellCount, this.notebookDocument.cellCount),
-            [
-                notebookCellData // TODO generalize to arbitrary languages and cell types
-            ]
-        );
-        await workspace.applyEdit(edit);
+        await chainWithPendingUpdates(this.notebookDocument, (edit) => {
```
joyceerhl, line 667: Cells were getting inserted into the `NotebookDocument` out of order, needed `chainWithPendingUpdates` to enforce edits in the right sequence

joyceerhl:
There's still a problem which also exists for the webview interactive window that I don't fully understand where, even though `addCode` is getting invoked with the right sequence of code cells, `executeObservable` seems to receive them out of order, i.e. if I do `print(1) ... print(4)`, the actual Jupyter session is told to `print(2); print(3); print(4); print(1)`, so the execution counts are all wrong because the execution requests were received in the wrong order. Might need your help with this @rchiodo

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
```diff
@@ -109,69 +110,83 @@ export class NativeInteractiveWindowProvider implements IInteractiveWindowProvid
...
+                return result;
+            });
+    }
 
-        // fire created event
-        this._onDidCreateInteractiveWindow.fire(result);
+    protected async create(resource: Resource, mode: InteractiveWindowMode): Promise<NativeInteractiveWindow> {
+        // If there's no resource, just create a new one
+        if (resource === undefined) {
+            return this.createInteractiveWindowPromise(resource, mode);
```
rchiodo, line 177: Could this cause the same race? You might create two interactive windows?

Wondering if the undefined case should also have an entry in the map

rchiodo (approved):

rchiodo:
> There's still a problem which also exists for the webview interactive window that I don't fully understand where, even though `addCode` is getting invoked with the right sequence of code cells, `executeObservable` seems to receive them out of order, i.e. if I do `print(1) ... print(4)`, the actual Jupyter session is told to `print(2); print(3); print(4); print(1)`, so the execution counts are all wrong because the execution requests were received in the wrong order. Might need your help with this @rchiodo

My guess is that the async code in addCode is causing one of them to wait but not another.

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
joyceerhl: If resource is undefined that means there's no backing script file, i.e. being created from command palette, in which case we would always want to create a new interactive window, no?

rchiodo: Looking at the old code, it was async too, but the creation of the window was based entirely on the mode. Why does the controller need to be created before the window?

rchiodo: The old code would just await the async parts after creation. Maybe we can live chat instead?

joyceerhl:
@rchiodo I'll fix up the promise chaining issue we talked about which impacts webview and explore refactoring the ctor in a separate PR

Issue #6465, opened 2021-06-29
Support error link clicking in interactive window


rchiodo:
Testing microsoft/vscode#127378

Create an error like so:

![image](https://user-images.githubusercontent.com/19672699/123853697-9df8eb80-d8d2-11eb-86d2-1538ff6714b7.png)

The old IW would allow a user to click on the line number to jump to where the exception was thrown.

greazer:
The case being made for this for IW and not necessarily native notebooks is because the user may be working with a large python file that tends to have a lot more code that a notebook.

PR #6660, opened 2021-07-14
Support error links in builtin interactive window

rebornix:
For #6465.

Added a renderer for error output, which will render error/traceback with `ansi-to-html`. We add a modified traceback info to the outputs' transient metadata by https://github.com/microsoft/vscode-jupyter/blob/b368ea9326c13d6497e0f070d89d810d4ff86416/src/client/datascience/editor-integration/cellhashprovider.ts#L134-L147

which will be used by the new output renderer to use as the traceback data for rendering.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1938,6 +1938,15 @@
                 ]
             },
             {
+                "id": "jupyter-error-renderer",
+                "entrypoint": "./out/datascience-ui/error-renderer/index.js",
+                "displayName": "Jupyter Error Renderer",
+                "requiresMessaging": "always",
+                "mimeTypes": [
+                    "application/vnd.code.notebook.error"
+                ]
+            },
+            {
                 "id": "jupyter-notebook-renderer",
                 "entrypoint": "./out/client_renderer/renderers.js",
                 "displayName": "Jupyter Notebook Renderer",
--- a/package.json
+++ b/package.json
@@ -1939,7 +1939,7 @@
             },
             {
                 "id": "jupyter-error-renderer",
-                "entrypoint": "./out/datascience-ui/error-renderer/index.js",
+                "entrypoint": "./out/datascience-ui/errorRenderer/errorRenderer.js",
                 "displayName": "Jupyter Error Renderer",
                 "requiresMessaging": "always",
                 "mimeTypes": [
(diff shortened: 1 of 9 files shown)


rebornix:
https://github.com/microsoft/vscode-jupyter/pull/6660/files#diff-f1216ee90d35222d31d042066ee9a29193385fd273457e4ff166ddaf0002d397R160-R251 is from https://github.com/microsoft/vscode-jupyter/blob/16d7ba06b56a7a2a7650e2ae7d29cefcfdd670e8/src/client/datascience/interactive-common/linkProvider.ts#L25-L25 which should be removed once we delete all interactive window web impl.

joyceerhl (approved):

rchiodo (approved):

package.json
```diff
@@ -1937,6 +1937,15 @@
                     "application/vnd.jupyter.widget-view+json"
                 ]
             },
+            {
+                "id": "jupyter-error-renderer",
+                "entrypoint": "./out/datascience-ui/errorRenderer/errorRenderer.js",
+                "displayName": "Jupyter Error Renderer",
+                "requiresMessaging": "always",
+                "mimeTypes": [
+                    "application/vnd.code.notebook.error"
```
DonJayamanne, line 1946: What will happen when irun a Jupyter notebook? Will the hyperlinks for errors work there?
I don't think it will, shouldn't we disable the links or get it working, else it's weird to have links and do nothing.
If it works or doesn't have links for error lines then fine, else I think it's a problem.

DonJayamanne (changes requested):
I think we should consider what we do when in notebooks. Not a fan of it not working in notebook output.

package.json
rebornix: > I think we should consider what we do when in notebooks. Not a fan of it not working in notebook output.

It works in notebook, if `preHandleIOPub` is for all kernel IOs, not just interactive window.

And also if the outputs don't have a `outputItem.metadata.transient: string[]`, we will use the raw error data directly, so it has a default fall back. But we may want to give it a better namespace, like `outputItem.metadata.transient.modifiedTraceback`.

rebornix: good catch though, the prehandleoppub code is not robust enough that it doesn't fail gracefully when it fails to find the file link

what it should render:

```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-2-659abc318634> in <module>
----> 1 print(fo)

NameError: name 'fo' is not defined
```

what it renders:

```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
 in 
----> 1 print(fo)

NameError: name 'fo' is not defined
```

DonJayamanne (approved):

PR #6675, opened 2021-07-15
Scrolling for outputs in native interactive window

joyceerhl:
For the rest of https://github.com/microsoft/vscode-jupyter/issues/6603

`setTimeout` when doing `editor.revealRange` to scroll to outputs during/after execution

### DIFF
--- a/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
+++ b/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
@@ -346,7 +346,8 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
         const id = uuid();
         const editor = window.visibleNotebookEditors.find((editor) => editor.document === this.notebookDocument);
 
-        // Compute isAtBottom based on last notebook cell before adding a notebook cell
+        // Compute isAtBottom based on last notebook cell before adding a notebook cell,
+        // since the notebook cell we're going to add is by definition not visible
         const isLastCellVisible = editor?.visibleRanges.find((r) => {
             return r.end === this.notebookDocument.cellCount - 1;
         });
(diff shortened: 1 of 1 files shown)


src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -615,7 +626,10 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
         const editor = window.visibleNotebookEditors.find((editor) => editor.document === this.notebookDocument);
         if (editor) {
             const notebookRange = new NotebookRange(notebookCell.index, notebookCell.index + 1);
-            editor.revealRange(notebookRange, NotebookEditorRevealType.InCenterIfOutsideViewport);
+            // This will always try to reveal the whole cell--input + output combined
+            setTimeout(() => {
+                editor.revealRange(notebookRange, NotebookEditorRevealType.Default);
+            }, 200); // Rendering output is async so the output is not guaranteed to immediately exist
```
rchiodo, line 632: Would zero work as well here? Not sure about 200. Seems rather arbitrary.

rchiodo (approved):

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
joyceerhl: 200 is more likely to work than 0, though note this doesn't work perfectly because outputs render asynchronously. For this to work properly we would need some kind of event to fire signaling that output has fully rendered

rchiodo: > event to fire signaling that output has fully rendered

Rendered or finished executing?

joyceerhl: Rendered. We know when the cell has finished executing. The problem is that when we do `replaceOutputs`, the outputs are not guaranteed to have completely rendered when `replaceOutputs` resolves.

Issue #6534, opened 2021-07-05
ipykernel 6.0.1 breaks interactive debugger


HDavidEspinosa:
Issue Type: <b>Bug</b>

I am working with anaconda as my Python distro and in my environment I updated ipykernel to the 6.0.1 version today.

As soon as I did this the interactive debbugger started popping out the following error:
Could not load source 'c:\Users\david\AppData\Local\Temp\ipykernel_17508\2969028694.py': Unable to retrieve source for C:\Users\david\AppData\Local\Temp\ipykernel_17508\2969028694.py.

Downgraded ipykernel to 5.5.5 all went back to normal.

Something in this upgrade is breaking the debugger when running through the interactive interpreter.

Extension version: 2021.6.999662501
VS Code version: Code 1.57.1 (507ce72a4466fbb27b715c3722558bb15afa9f48, 2021-06-17T13:28:07.755Z)
OS version: Windows_NT x64 10.0.19043
Restricted Mode: No

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz (12 x 2592)|
|GPU Status|2d_canvas: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>rasterization: enabled<br>skia_renderer: enabled_on<br>video_decode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|15.82GB (7.02GB free)|
|Process Argv|. --crash-reporter-id b5fdfc3b-a477-44a1-8dbc-9ec2aadb185b|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vsreu685:30147344
python383:30185418
pythonvspyt700cf:30270857
pythonvspyt602:30300191
vspor879:30202332
vspor708:30202333
vspor363:30204092
vswsl492cf:30256860
pythonvspyt639:30300192
pythontb:30283811
pythonvspyt551cf:30311713
vspre833:30321513
pythonptprofiler:30281270
vshan820:30294714
pythondataviewer:30285071
vscus158:30321503
pythonvsuse255:30323308
vscorehov:30309549
vscod805cf:30301675
pythonvspyt200cf:30331938
vscextlangct:30333562
[...]

PR #6691, opened 2021-07-16
Fix interactive window debugging with ipykernel6

DonJayamanne:
For #6534

### DIFF
--- a/src/client/datascience/interactive-common/interactiveBase.ts
+++ b/src/client/datascience/interactive-common/interactiveBase.ts
@@ -109,7 +109,6 @@ import { DataViewerChecker } from './dataViewerChecker';
 import { InteractiveWindowMessageListener } from './interactiveWindowMessageListener';
 import { serializeLanguageConfiguration } from './serialization';
 import { sendKernelTelemetryEvent, trackKernelResourceInformation } from '../telemetry/telemetry';
-import { IPythonExecutionFactory, IPythonExecutionService } from '../../common/process/types';
 
 export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindowMapping> implements IInteractiveBase {
     public get notebook(): INotebook | undefined {
@@ -177,8 +176,7 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
         private readonly notebookProvider: INotebookProvider,
         useCustomEditorApi: boolean,
         private selector: KernelSelector,
-        private serverStorage: IJupyterServerUriStorage,
-        private readonly pythonExecFactory: IPythonExecutionFactory
+        private serverStorage: IJupyterServerUriStorage
     ) {
         super(
             configuration,
@@ -678,7 +676,7 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
                     if (debugInfo.runByLine && debugInfo.hashFileName) {
                         await this.jupyterDebugger.startRunByLine(this._notebook, debugInfo.hashFileName);
                     } else if (!debugInfo.runByLine) {
-                        const promise = this._notebook.execute(
+                        await this._notebook.execute(
                             `import os;os.environ["IPYKERNEL_CELL_NAME"] = '${file.replace(/\\/g, '\\\\')}'`,
                             file,
                             0,
@@ -686,22 +684,7 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
                             undefined,
                             true
                         );
-                        const interpreter = this._notebook.getKernelConnection()?.interpreter;
-                        let execService: IPythonExecutionService | undefined;
-                        if (interpreter) {
-                            execService = await this.pythonExecFactory.createActivatedEnvironment({
-                                interpreter: this._notebook.getKernelConnection()?.interpreter
-                            });
-                        }
-                        let ipykernelVersion: string | undefined;
-                        if (execService) {
-                            const result = await execService
-                                .exec(['-c', 'import ipykernel;print(ipykernel.__version__)'], { env: process.env })
-                                .catch(noop);
-                            ipykernelVersion = result ? (result.stdout || result.stderr || '').trim() : undefined;
-                        }
-                        await promise;
-                        await this.jupyterDebugger.startDebugging(this._notebook, ipykernelVersion);
+                        await this.jupyterDebugger.startDebugging(this._notebook);
                     } else {
                         throw Error('Missing hash file name when running by line');
                     }
--- a/src/client/datascience/interactive-common/interactiveBase.ts
+++ b/src/client/datascience/interactive-common/interactiveBase.ts
@@ -678,6 +678,14 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
                     if (debugInfo.runByLine && debugInfo.hashFileName) {
                         await this.jupyterDebugger.startRunByLine(this._notebook, debugInfo.hashFileName);
                     } else if (!debugInfo.runByLine) {
+                        const promise = this._notebook.execute(
+                            `import os;os.environ["IPYKERNEL_CELL_NAME"] = '${file.replace(/\\/g, '\\\\')}'`,
+                            file,
+                            0,
+                            uuid(),
+                            undefined,
+                            true
+                        );
                         const interpreter = this._notebook.getKernelConnection()?.interpreter;
                         let execService: IPythonExecutionService | undefined;
                         if (interpreter) {
@@ -692,6 +700,7 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
                                 .catch(noop);
                             ipykernelVersion = result ? (result.stdout || result.stderr || '').trim() : undefined;
                         }
+                        await promise;
                         await this.jupyterDebugger.startDebugging(this._notebook, ipykernelVersion);
                     } else {
                         throw Error('Missing hash file name when running by line');
--- a/src/client/datascience/interactive-common/interactiveBase.ts
+++ b/src/client/datascience/interactive-common/interactiveBase.ts
@@ -109,6 +109,7 @@ import { DataViewerChecker } from './dataViewerChecker';
 import { InteractiveWindowMessageListener } from './interactiveWindowMessageListener';
 import { serializeLanguageConfiguration } from './serialization';
 import { sendKernelTelemetryEvent, trackKernelResourceInformation } from '../telemetry/telemetry';
+import { IPythonExecutionFactory, IPythonExecutionService } from '../../common/process/types';
 
 export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindowMapping> implements IInteractiveBase {
     public get notebook(): INotebook | undefined {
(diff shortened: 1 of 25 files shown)


DavidKutu (approved):

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -88,7 +88,8 @@ export class Kernel implements IKernel {
     private readonly kernelExecution: KernelExecution;
     private startCancellation = new CancellationTokenSource();
     constructor(
-        public readonly uri: Uri,
+        public readonly notebookUri: Uri,
+        public readonly resourceUri: Uri,
```
DonJayamanne, line 92: Today in our notebook provider we have the distinction between IdentityUri & ResourceUri.
This change merely ensures we support that in the IKernel object as well.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -215,8 +216,8 @@ export class Kernel implements IKernel {
                     try {
                         traceInfo(`Starting Notebook in kernel.ts id = ${this.kernelConnectionMetadata.id}`);
                         this.notebook = await this.notebookProvider.getOrCreateNotebook({
-                            identity: this.uri,
-                            resource: this.uri,
+                            identity: this.notebookUri,
+                            resource: this.resourceUri,
```
DonJayamanne, line 220: Now, we're passing the right `resource`, else we'd be passing the Uri of the Interactive Notebook Uri instead of the the Python file Uri (i.e. works the way it did with inteactive webview)

src/client/datascience/jupyter/kernels/types.ts
```diff
@@ -143,16 +143,23 @@ export interface IKernel extends IAsyncDisposable {
...
+    resourceUri?: Uri;
+};
 export const IKernelProvider = Symbol('IKernelProvider');
 export interface IKernelProvider extends IAsyncDisposable {
     /**
-     * Get hold of the active kernel for a given Uri (Notebook or other file).
+     * Get hold of the active kernel for a given Notebook.
      */
-    get(uri: Uri): IKernel | undefined;
+    get(notebook: NotebookDocument): IKernel | undefined;
```
DonJayamanne, line 159: Didn't want to keep the Uri, its too ambiguous, 
this way we're very explicit.

src/client/datascience/kernel-launcher/kernelLauncherDaemon.ts
```diff
@@ -51,14 +51,20 @@ export class PythonKernelLauncherDaemon implements IDisposable {
...
-        const [daemon, wdExists, env] = await Promise.all([
+        let [daemon, wdExists, env] = await Promise.all([
             this.daemonPool.get(resource, kernelSpec, interpreter),
             fs.pathExists(workingDirectory),
             this.kernelEnvVarsService.getEnvironmentVariables(resource, interpreter, kernelSpec)
         ]);
 
         // Check to see if we have the type of kernelspec that we expect
         const args = kernelSpec.argv.slice();
+        if (resource && resource.fsPath.toLowerCase().endsWith('.py')) {
```
DonJayamanne, line 62: Without the change to KernelProvider, this resource will be the Ingeractive Notebook Uri when using native itneractive window (ie. this used to work only in webview interactive, now works everywhere)

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -257,9 +259,10 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
     }
 
     private registerKernel(notebookDocument: NotebookDocument, controller: VSCodeNotebookController) {
-        const kernel = this.kernelProvider.getOrCreate(notebookDocument.uri, {
+        const kernel = this.kernelProvider.getOrCreate(notebookDocument, {
```
DonJayamanne, line 262: No ambiguity with the argument now, we pass the document, else if it was uri, one could accidentally provide the resource Uri (owning uri)

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -376,7 +379,21 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
+                const interpreter = notebook.getKernelConnection()?.interpreter;
+                let execService: IPythonExecutionService | undefined;
+                if (interpreter) {
+                    execService = await this.pythonExecFactory.createActivatedEnvironment({
+                        interpreter: notebook.getKernelConnection()?.interpreter
+                    });
+                }
+                let ipykernelVersion: string | undefined;
+                if (execService) {
+                    const result = await execService
```
joyceerhl, line 391: Can we use the `isProductVersionCompatible` Python extension API function instead of reimplementing here? IIRC David added it to the API?

DonJayamanne: No we can't, its not a compatibility check, we need the version information.

DonJayamanne: My plan is to enable caching as well (separate pr), also not sure we need to move stuff into python extension (else there's too much indirection. If jupyter is knows about what version is required and checking the version to enable something, why move that logic into another extension, also makey shipping the two difficult, i.e. if we ship this change, we MUST have python change, else things will never work)

src/client/datascience/kernel-launcher/kernelLauncherDaemon.ts
rchiodo: THis isn't going to work when the interactive window is set to 'single' or is multiple and you're sharing the IW amongst different files. 

Only the first python file will be debuggable.

rchiodo: I think we need to change the env value before every run of the cell if debugging.

rchiodo: We could likely do that in the cellHashProvider in the preExecute function. It could either add a line to the cell to do it or execute silently before the execute. The extra line would just make the offset for the breakpoint one larger.

rchiodo (changes requested):
⏲️

src/client/datascience/kernel-launcher/kernelLauncherDaemon.ts
rchiodo: Oh I guess putting it in the cell doesn't work because it needs it before it runs the cell. So it would need to execute some code ahead of time.

DonJayamanne: I'll try execution of the code silently.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
joyceerhl: `isProductVersionCompatible` just means whether the installed version satisfies a given semver constraint (point taken that the name could be improved). We can check if `isProductVersionCompatible` with the min version required for the new ipykernel behavior i.e. 6.0.x. If `false` that means use the old way, if `true` then we go down the codepath you added here. 

The reason I'm suggesting reusing `isProductVersionCompatible` is so that we don't duplicate logic for determining the version of `ipykernel` in the Jupyter extension.

src/client/datascience/jupyter/jupyterDebugger.ts
```diff
@@ -275,12 +277,14 @@ export class JupyterDebugger implements IJupyterDebugger, ICellHashListener {
 
     private buildSourceMap(fileHash: IFileHashes): ISourceMapRequest {
         const sourceMapRequest: ISourceMapRequest = { source: { path: fileHash.file }, pydevdSourceMaps: [] };
-
+        const isIPyKernel6 = this.ipykernelVersion?.toLowerCase().trim().startsWith('6') === true;
```
joyceerhl, line 280: We'd have to remember to update this when ipykernel 7 goes out, which is another reason I'd advocate for using `isProductVersionCompatible`.

src/client/datascience/kernel-launcher/kernelLauncherDaemon.ts
DonJayamanne: Agreed, done.

src/client/datascience/interactive-common/interactiveBase.ts
```diff
@@ -676,7 +678,30 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
                     if (debugInfo.runByLine && debugInfo.hashFileName) {
                         await this.jupyterDebugger.startRunByLine(this._notebook, debugInfo.hashFileName);
                     } else if (!debugInfo.runByLine) {
-                        await this.jupyterDebugger.startDebugging(this._notebook);
+                        const promise = this._notebook.execute(
+                            `import os;os.environ["IPYKERNEL_CELL_NAME"] = '${file.replace(/\\/g, '\\\\')}'`,
```
DonJayamanne, line 682: This is a harmless env variable for older Ipyhotn, hence i think we can just run this everytime.
My plan is to optimize all of this with heavy caching (cache version checks & depdnencies, etc rather than checking everytime https://github.com/microsoft/vscode-jupyter/issues/6694)
I'd rather do that after this PR goes, i.e. work on optimization seprately

DonJayamanne: @rchiodo /cc

rchiodo: Maybe this code coudl be in a separate function/file? I know the dupe of this in the interactive window will go away but there's quite a bit here that's making the enclosing function rather large.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
DonJayamanne: Oh yes, we can do that.

rchiodo:
Just waiting on the suggestion that Joyce made. Otherwise looks good to me

src/client/datascience/interactive-common/interactiveBase.ts
DonJayamanne: Moved most of the code, hence this enclosing function now has just one new line (moved all fo the other code out).

DonJayamanne:
> ust waiting on the suggestion that Joyce made. Otherwise looks good to me

@joyceerhl thanks for the explanation, initially i didn't quite understand that. It makes perfect sense. Updated the PR. Please check
@rchiodo it should be simpler now.

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -379,12 +379,16 @@ export class VSCodeNotebookController implements Disposable {
         // Before we start the notebook, make sure the metadata is set to this new kernel.
         await updateNotebookDocumentMetadata(document, this.documentManager, selectedKernelConnectionMetadata);
 
+        if (document.notebookType === InteractiveWindowView) {
+            // Possible its an interactive window, in that case we'll create the kernel manually.
+            return;
```
joyceerhl, line 384: Not sure I follow this part here?

DonJayamanne: We don't have the rsource information to create a kernel.
With interactive window, we have a *.py file, that needs to be passed when creating a kernel, else its not associated with any py file & thats not correct.

rchiodo (approved):

src/client/datascience/notebook/vscodeNotebookController.ts
joyceerhl: I don't think we should skip initialization here, I'm already working on a code change to not cache the kernel in the nativeInteractiveWindow.

DonJayamanne: > I don't think we should skip initialization her

Unless you have a way to pass the path of the Python file, then please feel free to restore this change. Else we cannot have this, as that will end up with the wrong information in the kernel.
I'll submit a subsequent PR that will make this resource information mandatory, that will make the problem i'm highlighting more obvious & we won't create kernels with the wrong information.

PR #6724, opened 2021-07-19
Fix markdown cells cancelling execution in native interactive window

joyceerhl:
Accessing `outputs` causes a TypeError here, so the execution promise never completes, which is why execution halts at a markdown cell. Note, I will refactor the NativeInteractiveWindow to not use observables at all and instead use the native notebooks codepath, this is just to unblock the scenario @greazer found

### DIFF
--- a/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
+++ b/src/client/datascience/interactive-window/nativeInteractiveWindow.ts
@@ -400,7 +400,7 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
             observable.subscribe(
                 (cells: ICell[]) => {
                     // Then send the combined output to the UI
-                    const converted = (cells[0].data as nbformat.ICodeCell).outputs.map(cellOutputToVSCCellOutput);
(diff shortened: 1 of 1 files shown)


src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -400,7 +400,7 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
             observable.subscribe(
                 (cells: ICell[]) => {
                     // Then send the combined output to the UI
-                    const converted = (cells[0].data as nbformat.ICodeCell).outputs.map(cellOutputToVSCCellOutput);
+                    const converted = (cells[0].data as nbformat.ICodeCell).outputs?.map(cellOutputToVSCCellOutput);
                     void temporaryExecution.replaceOutput(converted).then(() => {
```
rchiodo, line 404: Does replaceOutput do nothing if converted is undefined?

rchiodo (approved):

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
joyceerhl: Good catch, updated to not perform a cast here, as that messes with TS's ability to flag problems like this

PR #6725, opened 2021-07-19
Move `interactive.open` call back into `NativeInteractiveWindow` ctor

joyceerhl:
This reverses course on a previous recommendation, but is necessary in order to prevent races instantiating an interactive window. Basically we need to create our `NativeInteractiveWindow` wrapper without awaiting on anything (added a comment to that effect in the provider). This means moving `getInteractiveController` and `executeCommand('interactive.open', ...)` into the `NativeInteractiveWindow` ctor, and also means `notebookDocument` can now be undefined and we have to await on `this.loadPromise` to access the `notebookDocument` connected to the interactive window


src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -222,7 +169,64 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
         });
     }
 
-    private openFile(fileUri: string) {
+    private createReadyPromise(): Promise<NotebookDocument> {
+        return this.notebookControllerManager
```
rchiodo, line 173: Could you just use awaits here? The 'then' flow doesn't seem necessary.

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
```diff
@@ -95,7 +94,7 @@ export class NativeInteractiveWindowProvider implements IInteractiveWindowProvid
         let result = this.get(resource, mode) as IInteractiveWindow;
         if (!result) {
             // No match. Create a new item.
-            result = await this.create(resource, mode);
+            result = this.create(resource, mode);
         }
```
rchiodo, line 98: Don't you have to await the ready promise here?

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -256,6 +260,47 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
...
+        // Immediately try to find a selected controller for our NotebookDocument,
+        // as it's possible that a selection event fired before our ctor was able to run
+        const controller = this.notebookControllerManager.getSelectedNotebookController(notebookDocument);
+        if (controller !== undefined) {
+            this.registerKernel(notebookDocument, controller);
+            this.initialControllerSelected.resolve();
+        }
+
+        // Ensure we hear about any controller changes so we can update our cache accordingly
+        this.notebookControllerManager.onNotebookControllerSelected(
```
rchiodo, line 273: I thought you didn't need to register the kernel anymore. You just need the notebook document? Then you always ask for the active kernel for that document.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
```diff
@@ -122,6 +128,8 @@ export class NativeInteractiveWindow implements IInteractiveWindowLoadable {
     private kernel: IKernel | undefined;
     private kernelLoadPromise: Promise<void> | undefined;
     private initialControllerSelected: Deferred<void>;
+    private _readyPromise: Promise<NotebookDocument>;
+    private notebookDocument: NotebookDocument | undefined;
```
rchiodo, line 132: Seems like you should't catch the document here. Just use the readyPromise everywhere. Or maybe call it the 'documentPromise' or somehting.

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
joyceerhl: Yup, I'm deferring kernel changes to a separate PR (want to keep this one focused on the ctor refactor)

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
joyceerhl: Awaiting here causes races because the first getOrCreate takes longer to resolve than the second. I think it would be better to do the await in the codeWatcher.

rchiodo: Sorry was suggesting somethign different. The create should put the item into the map, the await happens after you have the result already. That's what the old IW did. It created the item and then awaited.

rchiodo: So that subsequent calls into this function get the same item but also await on the same promise.

joyceerhl: Yeah, that seems to cause races (cells get inserted into the IW in the wrong order). Not sure why.

rchiodo: But putting that same await into the code window doesn't? 

They should be functionally equivalent.

rchiodo: If it was the same as this
https://github.com/microsoft/vscode-jupyter/blob/adde5480d2ddca1d463db1c26e31f87505497de5/src/client/datascience/interactive-window/interactiveWindowProvider.ts#L122

It seems like it should behave like the old window

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
joyceerhl: We need this in order to support synchronously finding a matching interactive window based on this.notebookDocument.uri, in cases where having an undefined notebookUri doesn't impact correctness, e.g. getting the active interactive window. If there's no notebookDocument in that case, by definition the UI isn't up, and we read activeWindow to set context keys so awaiting on the ready promise in such scenarios is unnecessary.

src/client/datascience/interactive-window/nativeInteractiveWindowProvider.ts
joyceerhl: Dumb bug, thanks Rich :)

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
rchiodo: Let me make sure I understand. The notebookDocument.URI is necessary synchonously so that the context keys can update without having to await?

rchiodo (approved):

src/client/datascience/interactive-window/nativeInteractiveWindow.ts
joyceerhl: Wow, sorry, I meant we read the activeWindow to handle commands (in the command listener). No context keys involved. But the commands don't make sense without the UI being up so there shouldn't be a difference if we read notebookUri without awaiting on readyPromise versus awaiting and getting notebookDocument.uri.

PR #6744, opened 2021-07-21
Try again. Still not working on my machine

rchiodo:
Zeromq is still failing on postinstall.

This makes zeromq failures not break anything. Hopefully zeromq will still end up in the product though. PR should determine that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -102,8 +102,7 @@
                 "vscode-tas-client": "^0.1.17",
                 "vsls": "^0.3.1291",
                 "winston": "^3.2.1",
-                "ws": "^6.0.0",
-                "zeromq": "^6.0.0-beta.6"
+                "ws": "^6.0.0"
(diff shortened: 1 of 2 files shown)


joyceerhl:
@rchiodo What version of node do you have? I have windows-nvm and `nvm install 14.16.0`, and zmq install works just fine.

rchiodo:
> @rchiodo What version of node do you have? I have windows-nvm and `nvm install 14.16.0`, and zmq install works just fine.

I don't know if it's the version of node. Node-gyp-build is just thinking it needs to rebuild on my machine. I recently installed a bunch of OS updates, so that might be it as well.

This essentially skips the post install script for zeromq (although if it fails, it isn't there for me, so I have to install it with ```--ignore-scripts``` as a second step)

DonJayamanne (approved):

Issue #6735, opened 2021-07-21
Ctrl+P fails to load Jupyter Notebook (ipynb) files


Astlaan:
Issue Type: <b>Bug</b>

After the latest Jupyter update (which included the latest aesthetic/layout changes), using the Ctrl+P shortcut to open an ipynb file appears not to work. When I do this, a new tab opens, but   stays blank, with the loading animation (the "travelling line" below the tab list) playing indefinitely.

There seems to be no problem in opening ipynb files with VS Code when done directly from the Windows Explorer.

Extension version: 2021.8.1046824664
VS Code version: Code 1.58.2 (c3f126316369cd610563c75b1b1725e0679adfb3, 2021-07-14T22:10:15.214Z)
OS version: Windows_NT x64 10.0.19042
Restricted Mode: No

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz (8 x 1690)|
|GPU Status|2d_canvas: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>rasterization: enabled<br>skia_renderer: enabled_on<br>video_decode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|15.73GB (3.42GB free)|
|Process Argv|G:\\Code\\workspace.code-workspace --crash-reporter-id c4587ea1-65de-4590-a2c4-cb48d86b4f8b|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vsreu685:30147344
python383:30185418
vspyt653:30270858
pythonvspyt602:30300191
vspor879:30202332
vspor708:30202333
vspor363:30204092
pythonvspyt639:30300192
pythontb:30283811
pythonvspyt551:30311712
vspre833:30321513
pythonptprofiler:30281270
vshan820:30294714
vstes263:30335439
pythondataviewer:30285071
vscus158:30321503
pythonvsuse255:30340121
vscod805:30301674
pythonvspyt200:30340761
vscextlang:30333561
binariesv615:30325510
vsccppwtct:30329789
bridge0708:30335490

```

</details>

szz65:
[...]

PR #6745, opened 2021-07-21
Make custom editor work even in native case

rchiodo:
For #6735

Seems custom editor wasn't even registered when the native notebook is being used. @rebornix suggested I just make the custom editor work even when native is the default. This should just fix all of the problems people are hitting.


package.json
```diff
@@ -2388,6 +2387,7 @@
     },
     "optionalDependencies": {
         "canvas": "^2.7.0",
-        "playwright-chromium": "^1.10.0"
+        "playwright-chromium": "^1.10.0",
+        "zeromq": "^6.0.0-beta.6"
```
rchiodo, line 2391: This should go away when I rebase on the other PR I have out. This is so I can run npm install on my machine

joyceerhl (approved):

DonJayamanne (approved):

DavidKutu (approved):

PR #6752, opened 2021-07-21
variable window in rbl

DavidKutu:
For #6701


src/client/datascience/jupyterDebugService.ts
```diff
@@ -252,6 +253,27 @@ export class JupyterDebugService implements IJupyterDebugService, IDisposable {
         this.onClose();
     }
 
+    public requestKernelDebugAdapterVariables(msg: DebugProtocolMessage): void {
+        switch ((msg as any).type) {
+            case 'request':
+                this.debugAdapterTrackers.forEach((d) => {
```
roblourens, line 259: I don't understand why you need to forward messages to the debug adapter trackers manually, isn't vscode doing that when you register them for the session?

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -242,8 +242,8 @@ export class DebuggerVariables extends DebugLocationTracker
         } else if (
             message.type === 'response' &&
             message.command === 'variables' &&
-            message.body &&
-            this.currentSeqNumsForVariables.has(message.request_seq)
+            message.body
+            // this.currentSeqNumsForVariables.has(message.request_seq)
```
rchiodo, line 246: This messes up variable hover. Vandy added this to fix variable hover.
https://github.com/microsoft/vscode-jupyter/pull/6186/files

rchiodo (changes requested):
⏲️

src/client/datascience/jupyterDebugService.ts
DavidKutu: we ended up removing this

src/client/datascience/variablesView/variableView.ts
```diff
@@ -78,6 +78,7 @@ export class VariableView extends WebviewViewHost<IVariableViewPanelMapping> imp
         this.notebookWatcher.onDidExecuteActiveNotebook(this.activeNotebookExecuted, this, this.disposables);
         this.notebookWatcher.onDidChangeActiveNotebook(this.activeNotebookChanged, this, this.disposables);
         this.notebookWatcher.onDidRestartActiveNotebook(this.activeNotebookRestarted, this, this.disposables);
+        this.variables.refreshRequired(this.sendRefreshMessage, this, this.disposables);
```
DonJayamanne, line 81: Somethings isn't right here.

Either this was completely broken or we have some weird code.
I'm not sure how this worked in the previous WebView version of Run by line & interactive window.
Did that old code perform their own request in some other place? Or do we have some other pattern or way to make this request (

I.e. i'd like to understand how the old code worked and why we need this here.

src/client/datascience/variablesView/variableView.ts
```diff
@@ -215,4 +216,22 @@ export class VariableView extends WebviewViewHost<IVariableViewPanelMapping> imp
...
+
+            const args: IJupyterVariablesRequest = {
+                executionCount: this.notebookWatcher.activeNotebookExecutionCount,
+                sortColumn: 'name',
+                startIndex: 0,
+                sortAscending: true,
+                pageSize: 5,
+                refreshCount: 1
+            };
+            await this.requestVariables(args);
```
DonJayamanne, line 234: Weird
* this.variables.refreshRequired(this.sendRefreshMessage, this, this.disposables);
	So, lets assume `variables` object triggers the event `refreshRequired`
* The handler will then do something and call a method `requestVariables`
* The `requestVariables` method then calls a method in `variables` object

I.e. its just a loop, variables fires & event indicating variables need to be refreshed & then something else calls a method on that same object.

Or is the event name incorrect? Does the event mean the variables are stale?

@rchiodo @DavidKutu  not sure you can help here, but @IanMatthewHuff  is out.
If not sure, then we can ignore this. but the flow seems to be weird

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -226,7 +212,19 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDis
...
+                ? `${path.basename(doc.uri.toString())}?RBL=${cell.index}`
+                : path.basename(doc.uri.toString());
+            const justMyCode = cell ? true : false;
+            const config: DebugConfiguration = {
+                type: DataScience.pythonKernelDebugAdapter(),
+                name: name,
+                request: 'attach',
+                internalConsoleOptions: 'neverOpen',
+                justMyCode: justMyCode,
+                __document: doc.uri.toString()
```
DonJayamanne, line 225: Some comments here would be great, as `__document` is not a known property in the debug configuration.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -226,7 +212,19 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDis
     private async startDebugging(doc: NotebookDocument, cell?: NotebookCell, options?: DebugSessionOptions) {
         let dbg = this.notebookToDebugger.get(doc);
         if (!dbg) {
-            dbg = new Debugger(doc, cell, options);
+            const name = cell
+                ? `${path.basename(doc.uri.toString())}?RBL=${cell.index}`
```
DonJayamanne, line 216: Why `RBL=` Does it mean anything? Why not just `xyz.ipynb?Cell=123`
`RBL` just looks weird

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -313,6 +356,19 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter {
             }
         });
 
+        if ((message as DebugProtocol.StackTraceResponse).command === 'stackTrace') {
```
DonJayamanne, line 359: Would be good to have comments around what these methods are doing (or at least at the class level).
When i read this code, I wasn't sure what we were doing & why, i had to build my own notebook debugger (personal extension) to understand out what we were doing & why...

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -208,6 +212,45 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter {
         });
     }
 
+    private RunByLineStackTrace(): void {
```
DonJayamanne, line 215:
```suggestion
    private runByLineStackTrace(): void {
```

Lower case

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -118,6 +118,10 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter {
             if (content.event === 'stopped') {
                 this.runByLineThreadId = content.body.threadId;
                 this.runByLineSeq = content.seq;
+
+                if (this.isRunByLine) {
+                    this.RunByLineStackTrace();
```
DonJayamanne, line 123: Not sure why this is called here..

DonJayamanne: Some comments would help (one line comment)

DonJayamanne: I know this is old code `  if (this.session.name.includes('RBL=')) {`
But would be good to have a constant instaed of hardcoding `RBL` i wouldn't have known that its used in this manner at all.
I commented elsewhere suggesting we not use `RBL`

DonJayamanne: Instead of using `RBL` why don't we add a custom configuration property item into the debugger. Just like we use `__document`, that way we dont have to look for particular strings.

DonJayamanne (approved):

src/client/datascience/variablesView/variableView.ts
rchiodo: I believe we need this here because the old code did a refresh when the debugger updated the variables pane. So interactive debug would auto refresh whenever the debugger refreshed. 

Run by line in the webview didn't have this problem.

David and I chatted about something similar to this. The debuggerVariables (or perhaps a standalone class) could be modified to force variable refresh whenver the debugger hits a stop event. I think that would also alleviate this problem.

src/client/datascience/variablesView/variableView.ts
rchiodo: I agree, this seems like it could be refactored. The responsibilities of the different classes are getting muddied in my head.

DonJayamanne (approved):

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -226,16 +230,29 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDis
...
             this.notebookToDebugger.set(doc, dbg);
 
             try {
                 await dbg.session;
 
-                if (!cell) {
-                    // toggle the breakpoint margin
-                    void this.commandManager.executeCommand('notebook.toggleBreakpointMargin', doc);
-                }
+                // if (cell) {
```
DonJayamanne, line 251: Please can we remove this code

src/client/datascience/jupyter/jupyterVariables.ts
```diff
@@ -92,8 +91,8 @@ export class JupyterVariables implements IJupyterVariables {
         );
     }
 
-    private async getVariableHandler(notebook?: INotebook): Promise<IJupyterVariables> {
-        if (this.debuggerVariables.active && (!notebook || notebook.status === ServerStatus.Busy)) {
+    private async getVariableHandler(_notebook?: INotebook): Promise<IJupyterVariables> {
```
IanMatthewHuff, line 94: This function is just local to this class so can just remove the parameter if we are not using it, no need to _. But at the same time is it ok that we are missing that (!notebook || notebook.status === ServerStatus.Busy) clause?

src/client/datascience/jupyter/debuggerVariables.ts
DavidKutu: done

src/client/datascience/variablesView/variableView.ts
DavidKutu: this was removed

src/client/datascience/variablesView/variableView.ts
DavidKutu: this was also removed

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: done

src/client/datascience/jupyter/jupyterVariables.ts
DavidKutu: It doesn't work for the ipykernel debugger, but It might be necessary for the old debugger. I need to check

DavidKutu: Thinking about it, the old stuff (webview notebooks and interactive window) won't be available starting next month, so that doesn't matter. The native interactive window seems to not be hooked up yet (I checked in main), I'll check that in the next iteration.

So I'll remove the parameter and its mentions.

roblourens: Not hooked up to the variables window? It works for me in the interactive window

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -158,7 +172,32 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDis
...
                                     this.fs
                                 );
+                                this.disposables.push(
+                                    adapter.onDidSendMessage((msg: DebugProtocolMessage) => {
+                                        if ((msg as DebugProtocol.VariablesResponse).command === 'variables') {
+                                            this._onDidFireVariablesEvent.fire();
+                                        }
+                                        switch ((msg as DebugProtocol.ProtocolMessage).type) {
+                                            case 'request':
+                                                this.debugAdapterTrackers.forEach((d) => {
```
rchiodo, line 182: Not sure why this necessary? Shouldn't debug adapter trackers just get called anyway?

src/client/datascience/jupyter/debuggerVariableRegistration.ts
```diff
@@ -9,16 +9,20 @@ import { PYTHON_LANGUAGE } from '../../common/constants';
...
 export class DebuggerVariableRegistration implements IExtensionSingleActivationService, DebugAdapterTrackerFactory {
     constructor(
         @inject(IJupyterDebugService) @named(Identifiers.MULTIPLEXING_DEBUGSERVICE) private debugService: IDebugService,
+        @inject(IDebuggingManager) private readonly debuggingManager: IDebuggingManager,
         @inject(IDisposableRegistry) private disposables: IDisposableRegistry,
         @inject(IJupyterVariables) @named(Identifiers.DEBUGGER_VARIABLES) private debugVariables: DebugAdapterTracker
     ) {}
     public activate(): Promise<void> {
         this.disposables.push(this.debugService.registerDebugAdapterTrackerFactory(PYTHON_LANGUAGE, this));
+        this.disposables.push(this.debuggingManager.registerDebugAdapterTrackerFactory(pythonKernelDebugAdapter, this));
```
rchiodo, line 25: Seems like if you use the debugservice, registerDebugAdapterTrackerFactory, you wouldn't need the custom debug adapter tracker stuff.

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: It doesn't get called :/
jupyterDebugService works the same way

src/client/datascience/jupyter/debuggerVariableRegistration.ts
DavidKutu: It would be a lot of work to do it that way. debugService is meant to be used for debugpy, as opposed to debuggingManager which uses debugpy in ipykernel.

It makes more sense this way. (And it also makes sense to remove debugService, stop using debugpy directly, and change the interactive window to also use ipykernel)

src/client/debugger/jupyter/debuggingManager.ts
rchiodo: JupyterDebugService is different. It's the entire debugger for run by line for webviews. Because we couldn't switch VS code into debug mode.

rchiodo: In your situation, the real debugger is there.

src/client/datascience/jupyter/debuggerVariableRegistration.ts
rchiodo: Going to disagree. You're reinventing something that should already work.

DebugService works by multiplexing. It supports the old run by line and regular debugging. We actually don't need that anymore as it can just use the real VS code debug service instead.

rchiodo: DebugService isn't used for debugpy. It's used for the old run by line. The debug cell path for the interactive window is just injecting debugpy and then basically doing the same thing you're doing in your kernel adapter. It does an attach to the kernel.

IanMatthewHuff: I feel like the ownership is a bit muddled in my head as well from when I chatted with you David. The main core issue was just that debuggerVariables needed to be getting the debug messages incoming into it, right?

IanMatthewHuff: So is the change here just detaching debuggerVariables from debug service and just hooking up to the vscode debug messages directly? Like we chatted about this would invalidate the old non-ipykernel 6.0 debugging functionality, but we might be ok with that at this point.

rchiodo: First step was to use the debugAdapterTracker stuff instead of custom passing messages to the jupyter kernel and routing messages to the debug adapter's manually.

Second step (after this PR I think) would be to remove the JupyterDebugSession, MultiplexingDebugSession, and JupyterDebugger and just use what David has here for run by line/notebook debugging for interactive window debugging too.

rchiodo (approved):

IanMatthewHuff (approved):

PR #6754, opened 2021-07-21
Load kernels from cache and kernel for active interpreter

DonJayamanne:
Part of #5861
* Load controller ASAP for Active Interpreter
* Load controllers from cache (whilst fetching real controllers & updating the list after we get them)

# First time users
* We load active interpreter as kernel fast
* We load all other kernels (as we do today, which is a little slow)
# Users coming back
* We load kernels from cache
* We load kernels without cache in the background, and as its made available, we add those.
	* I.e. if a user as a new environment, and that gets picked up by the kernel discovery it will be added subsequently to the list of kernels (but that's a little slower).

### DIFF
--- a/src/client/datascience/kernel-launcher/localKernelFinder.ts
+++ b/src/client/datascience/kernel-launcher/localKernelFinder.ts
@@ -30,6 +30,7 @@ import { IFileSystem } from '../../common/platform/types';
 import { noop } from '../../common/utils/misc';
 import { createPromiseFromCancellation } from '../../common/cancellation';
 
+const GlobalKernelSpecsCacheKey = 'JUPYTER_GLOBAL_KERNELSPECS';
 // This class searches for a kernel that matches the given kernel name.
 // First it searches on a global persistent state, then on the installed python interpreters,
 // and finally on the default locations that jupyter installs kernels on.
@@ -160,7 +161,7 @@ export class LocalKernelFinder implements ILocalKernelFinder {
 
         const kernels = this.filterKernels(nonPythonKernelSpecs.concat(pythonRelatedKernelSpecs));
         this.lastFetchedKernelsWithoutCache = kernels;
-        this.globalState.update('JUPYTER_GLOBAL_KERNELSPECS', kernels).then(noop, (ex) => {
+        this.globalState.update(GlobalKernelSpecsCacheKey, kernels).then(noop, (ex) => {
             console.error('Failed to update global kernel cache', ex);
         });
         return kernels;
@@ -171,7 +172,7 @@ export class LocalKernelFinder implements ILocalKernelFinder {
     ): Promise<LocalKernelConnectionMetadata[]> {
         const values = this.lastFetchedKernelsWithoutCache.length
             ? this.lastFetchedKernelsWithoutCache
-            : this.globalState.get<LocalKernelConnectionMetadata[]>('JUPYTER_GLOBAL_KERNELSPECS', []);
+            : this.globalState.get<LocalKernelConnectionMetadata[]>(GlobalKernelSpecsCacheKey, []);
         const validValues: LocalKernelConnectionMetadata[] = [];
         const promise = Promise.all(
             values.map(async (item) => {
--- a/src/client/datascience/kernel-launcher/localKernelFinder.ts
+++ b/src/client/datascience/kernel-launcher/localKernelFinder.ts
@@ -178,6 +178,7 @@ export class LocalKernelFinder implements ILocalKernelFinder {
                 let somethingIsInvalid = false;
                 const promises: Promise<void>[] = [];
                 if (item.interpreter?.path) {
+                    // Possible the interpreter no longer exists, in such cases, exclude this cached kernel from the list.
                     promises.push(
                         this.fs
                             .localFileExists(item.interpreter.path)
@@ -190,6 +191,7 @@ export class LocalKernelFinder implements ILocalKernelFinder {
                     );
                 }
                 if (item.kind === 'startUsingKernelSpec' && item.kernelSpec?.specFile) {
+                    // Possible the kernelspec file no longer exists, in such cases, exclude this cached kernel from the list.
                     promises.push(
                         this.fs
                             .localFileExists(item.kernelSpec.specFile)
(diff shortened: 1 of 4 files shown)


src/client/datascience/kernel-launcher/types.ts
```diff
@@ -58,8 +58,15 @@ export interface ILocalKernelFinder {
         option?: nbformat.INotebookMetadata,
         cancelToken?: CancellationToken
     ): Promise<LocalKernelConnectionMetadata | undefined>;
-    listNonPythonKernels(cancelToken?: CancellationToken): Promise<LocalKernelConnectionMetadata[]>;
-    listKernels(resource: Resource, cancelToken?: CancellationToken): Promise<LocalKernelConnectionMetadata[]>;
+    listNonPythonKernels(
+        cancelToken?: CancellationToken,
+        useCache?: 'useCache' | 'ignoreCache'
```
DonJayamanne, line 63: Didn't want to use a boolean, i feel this is more descriptive

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -269,6 +286,10 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         traceInfoIf(isCI, `Clear controller mapping for ${document.uri.toString()}`);
         const loadControllersPromise = this.loadNotebookControllers();
 
+        if (isPythonNotebook(getNotebookMetadata(document)) && this.extensionChecker.isPythonExtensionInstalled) {
+            // If we know we're dealing with a Python notebook, load the active interpreter as a kernel asap.
+            this.createActiveInterpreterController().catch(noop);
```
rchiodo, line 291: Does this need to be dedupified later? Won't it be returned by the code after this again?

DonJayamanne: Agreed, will add a simple flag to ensure we don't do this again.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -122,17 +159,71 @@ export class LocalKernelFinder implements ILocalKernelFinder {
         ]);
 
         const kernels = this.filterKernels(nonPythonKernelSpecs.concat(pythonRelatedKernelSpecs));
-        sendKernelListTelemetry(resource, kernels);
+        this.lastFetchedKernelsWithoutCache = kernels;
+        this.globalState.update('JUPYTER_GLOBAL_KERNELSPECS', kernels).then(noop, (ex) => {
```
rchiodo, line 163: Probably makes sense to turn this string into a constant so we don't have any typos in other uses.

src/client/datascience/notebook/notebookControllerManager.ts
DonJayamanne: Fixed.

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -122,17 +159,71 @@ export class LocalKernelFinder implements ILocalKernelFinder {
...
+                                }
+                            })
+                            .catch(noop)
+                    );
+                }
+                if (item.kind === 'startUsingKernelSpec' && item.kernelSpec?.specFile) {
+                    // Possible the kernelspec file no longer exists, in such cases, exclude this cached kernel from the list.
+                    promises.push(
+                        this.fs
+                            .localFileExists(item.kernelSpec.specFile)
```
rchiodo, line 197: I'm not sure this is true? Do we ever create 'startUsingKernelSpec' without a specfile?

rchiodo: As in I don't see this condition from before.

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: oops

src/client/datascience/kernel-launcher/localKernelFinder.ts
rchiodo: The code in the jupyterKernelService seems to rely on the fact that the specFile doesn't exist.

rchiodo (changes requested):
⏲️

src/client/datascience/kernel-launcher/localKernelFinder.ts
DonJayamanne: Its possible the user deleted the kernelspec file.

> As in I don't see this condition from before.

Thats correct, thats because we always serach and return only kernelspecs that exist on disc.
However now that I'm caching it, its possibel the user subsequently uninstalls/deletes that kernelspec from disc. In which case, the item in the cache is no longer valid (as it doesn't exist).

rchiodo: So yes we only search for ones on disk, but I thought we generate ones too, where the file may not have ever existed. So this would destroy those (as well as removing the ones you intend).

DonJayamanne: > here the file may not have ever existed. So this would destroy those (as well as removing the ones you intend).

In that case, aren't we starting from `PythonInterpreter` & not from kernel spec.

DonJayamanne: I.e. if we're starting from a kernel spec, then the spec file must exist & this code ensures we only deal with those kernels where `kind= startUsingKernelSpec`

rchiodo: > aren't we starting from PythonInterpreter & not from kernel spec.

I can't say what the type would end up as. I mean you can test it and find out. Or maybe just look at all the spots where it's 'startUsingKernelSpec' and verify they always need a kernelspec. I thought we always create a 'default' one that may or may not have an actual kernelspec on disk.

DonJayamanne: > where it's 'startUsingKernelSpec' and verify they always need a kernelspec.

Yup, i checked, and can confirm that it must exist.

> thought we always create a 'default' one that may or may not have an actual kernelspec on disk.

Yes, in that case the `kind=DefaultKernelConnectionMetadata`
Hence the code i have should work correctly.

DonJayamanne: > thought we always create a 'default' one that may or may not have an actual kernelspec on disk.

Yes, thats for when we start Python interpreters (we have a bogus kernelspec, which will get created only when we register the kernel), Again, i'm not interested in those (not interested when we start using Python interpreter).

rchiodo (approved):

PR #6756, opened 2021-07-22
Remove notebook serializer from Jupyter extension

joyceerhl:
With https://github.com/microsoft/vscode/commit/0e6388c47adfeec9de3424585df3040ba13ef1fd we can now remove the jupyter-notebook serializer we're registering from the Jupyter extension. They can coexist, so rolling this out doesn't need to be carefully coordinated--if a duplicate serializer is registered VS Code just logs an error in the dev console.


src/test/datascience/notebook/contentProviderMain.vscode.test.ts
```diff
@@ -1,150 +0,0 @@
...
-    setup(async () => {
-        const vscNotebooks = mock<IVSCodeNotebook>();
-        when(vscNotebooks.onDidSaveNotebookDocument).thenReturn(new EventEmitter<NotebookDocument>().event);
-        const memento = mock<Memento>();
-        when(memento.get(anything())).thenReturn();
-        languageService = mock<NotebookCellLanguageService>();
-        contentProvider = new NotebookSerializer(instance(languageService));
-    });
-    teardown(() => disposeAllDisposables(disposables));
-    test('Return notebook with 2 cells', async () => {
```
joyceerhl, line 37: @roblourens You may wish to add these tests for the ipynb extension

src/client/datascience/notebook/notebookEditorProvider.ts
```diff
@@ -131,28 +132,9 @@ export class NotebookEditorProvider implements INotebookEditorProvider {
...
-            nbformat_minor: defaultNotebookFormat.minor
-        };
-        if (options?.contents) {
-            Object.assign(nbJson, JSON.parse(options.contents));
-        }
-        if (options?.defaultCellLanguage) {
-            if (!nbJson.metadata) {
-                nbJson.metadata = nbJson.metadata || { orig_nbformat: 4 };
-            }
-            if (nbJson.metadata.language_info) {
```
roblourens, line 147: Do you need to keep this metadata?

DonJayamanne: Yes, we need that, that's how we pick the preferred kernel (we indirect the metadata)

src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -98,8 +97,10 @@ export class NotebookEditor implements INotebookEditor {
         return getNotebookMetadata(this.document);
     }
     onExecutedCode?: Event<string> | undefined;
-    public getContent() {
-        return this.serializer.serializeNotebookDocument(this.document);
+    public async getContent(): Promise<string> {
+        await workspace.saveAll(false);
```
DonJayamanne, line 101: Shouldn't we just save the active notebook editor instead of everything. Else user will get provided to save other temporary untiled files as well, which would be undesirable (eg generally I have a few untitled files)

src/client/datascience/notebook/notebookEditorProvider.ts
DonJayamanne: I think we need to restore that code

src/client/datascience/notebook/notebookEditor.ts
roblourens: What is the export command supposed to do? I run it and nothing happens. Also I think the new method won't work if you export a notebook with unsaved changes. 

I don't want to share code with npm. Does it make sense for vscode to own the export commands?

joyceerhl: Export runs `nbconvert` which is a Python module.

src/client/datascience/notebook/notebookEditorProvider.ts
```diff
@@ -135,7 +138,18 @@ export class NotebookEditorProvider implements INotebookEditorProvider {
...
         const doc = await this.vscodeNotebook.openNotebookDocument(JupyterNotebookView, data);
+        const metadata = cloneDeep(doc.metadata);
+        if (metadata.language_info !== undefined) {
+            metadata.language_info.name = options?.defaultCellLanguage;
+        } else {
+            metadata.language_info = {
+                name: options?.defaultCellLanguage
+            };
+        }
+        edit.replaceNotebookMetadata(doc.uri, metadata);
```
DonJayamanne, line 151: Why not add the metadata to the `data (NotebookDdata)` in line `141`. then we don't need to edit the notebook document.

joyceerhl: Ahh, I thought `metadata` was frozen. Looks like it's only frozen for `NotebookDocument` not `NotebookData`

src/client/datascience/notebook/notebookEditor.ts
```diff
@@ -98,8 +97,14 @@ export class NotebookEditor implements INotebookEditor {
         return getNotebookMetadata(this.document);
     }
     onExecutedCode?: Event<string> | undefined;
-    public getContent() {
-        return this.serializer.serializeNotebookDocument(this.document);
+    public async getContent(): Promise<string> {
+        const cells = generateCellsFromNotebookDocument(this.document, false);
```
DonJayamanne, line 101: This code is weird, in other places we save the notebook and get the content from the file system, here we convert the notebook to JSON ourselves.
Why can't we do this in other places as well, instead of getting the user to save the notbook to disck (i.e.have a consisten approach).

src/client/datascience/commands/exportCommands.ts
```diff
@@ -69,8 +69,10 @@ export class ExportCommands implements IDisposable {
             : this.notebookProvider.activeEditor;
 
         if (editor) {
+            await workspace.saveAll(false);
+            const contents = await this.fs.readFile(editor.file);
             const interpreter = editor.notebook?.getMatchingInterpreter();
```
DonJayamanne, line 74: Why can't we use `generateCellsFromNotebookDocument` or similar.

joyceerhl: Because we don't have a NotebookDocument to work with in webview interactive window and webview notebooks. Eventually we can do that, but not yet

src/client/datascience/notebook/notebookEditor.ts
joyceerhl: Because there's no NotebookDocument for webview cases

src/client/datascience/commands/exportCommands.ts
joyceerhl: I don't think Peng and I are planning to remove the ability to opt out of native interactive window till next iteration

joyceerhl: Oh wait, 🤦‍♀️ this shouldn't be here

src/client/datascience/notebook/notebookEditor.ts
joyceerhl: Sorry, misunderstood your question, we should've used editor.getContent() in the other place too

DonJayamanne: This doesn't need to be async

src/client/datascience/notebook/notebookEditorProvider.ts
```diff
@@ -131,28 +132,15 @@ export class NotebookEditorProvider implements INotebookEditorProvider {
...
-                nbJson.metadata.language_info.name = options.defaultCellLanguage;
-            } else {
-                nbJson.metadata.language_info = { name: options.defaultCellLanguage };
+        // contents will be ignored
+        const language = options?.defaultCellLanguage ?? PYTHON_LANGUAGE;
+        const cell = new NotebookCellData(NotebookCellKind.Code, '', language);
+        const data = new NotebookData([cell]);
+        data.metadata = {
+            language_info: {
+                name: language
```
DonJayamanne, line 141: The metadadata is invalid, we need to set the other values as well (`cells`, `nbformat`, etc).
If we're setting the metadata it has to be a whole, not a partial (in other parts we access the metadata and its type is `nbformat.INotebookContent` and none of them are optional, hence `cells` must exist.

Also, the metadata is not stored in the right place. using `data.metadata` is not correct.
I think we used to set it in custom as follows:
`data.metadata = {custom: nbJson}`

**Background:** In the past the only way to store metadta was to use the `metadata.custom` property, only later was the VS Code API changed to allow any arbitrary properties. but we didn't change that as others were already using this & we didn't want to break it (existing code & .NET extension)

I think this will fix it
We should instead have 
```typescript
const nbJson: nbformat.INotebookContent = {
    cells: [],
    metadata: { orig_nbformat: defaultNotebookFormat.major },
    nbformat: defaultNotebookFormat.major,
    nbformat_minor: defaultNotebookFormat.minor
};
data.metadata = {custom: nbJson}
```

DonJayamanne (changes requested):

src/client/datascience/types.ts
```diff
@@ -634,7 +634,7 @@ export interface INotebookEditor extends Disposable, IInteractiveBase {
     interruptKernel(): Promise<void>;
     restartKernel(): Promise<void>;
     syncAllCells(): Promise<void>;
-    getContent(): string;
+    getContent(): Promise<string>;
```
DonJayamanne, line 637: Can be synchronous

src/client/datascience/notebook/notebookEditor.ts
DonJayamanne, line 101: I don't think this function is the right one.  We should be using the one we had in the serializer, this function doesn't support converting Notebook cells to ipynb cells with the output. **The output is ignored**.

DonJayamanne (changes requested):
Found a few small issues.

src/client/datascience/commands/exportCommands.ts
rebornix: Since we are disallowing users using custom editor, we should probably stop supporting the webview IW as well.

src/client/datascience/notebook/notebookEditorProvider.ts
joyceerhl: We should probably revisit the custom metadata property. I assume that this was from when .NET Interactive built on top of the Jupyter extension for creating new Jupyter files. I think .NET Interactive would ideally depend on the serializer extension, because the custom property sounds like a workaround that is no longer necessary. The New File... contribution point shipped in the 1.59 release so I believe it would also be possible for .NET Interactive to contribute to that directly https://code.visualstudio.com/updates/v1_59#_contribution-point-for-new-file-menu

DavidKutu (approved):

rchiodo (approved):

DonJayamanne (approved):

PR #6769, opened 2021-07-23
Disable autostarting restart sessions

DonJayamanne:
Perf improvement


rchiodo (approved):

joyceerhl (commented):
Are we disabling autostarting restart sessions, or simply not creating them?

DonJayamanne:
> autostarting restart sessions, or simply not creating them?

We're not auto staring the restart session. 
It will be started only when we need it.

PR #6783, opened 2021-07-23
Add workflow for validating testplan items

JacksonKearl:
Ref https://github.com/microsoft/vscode-internalbacklog/issues/2286

### DIFF
--- a/.github/workflows/test-plan-item-validator.yml
+++ b/.github/workflows/test-plan-item-validator.yml
@@ -1,0 +1,24 @@
+name: Test Plan Item Validator
+on:
+  issues:
+    types: [edited, labeled]
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    if: contains(github.event.issue.labels.*.name, 'testplan-item') || contains(github.event.issue.labels.*.name, 'invalid-testplan-item')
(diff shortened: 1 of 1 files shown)


.github/workflows/test-plan-item-validator.yml
```diff
@@ -0,0 +1,24 @@
+name: Test Plan Item Validator
+on:
+  issues:
+    types: [edited, labeled]
+
```
joyceerhl, line 5: We locked down actions' access to our GitHub token some time ago. Depending on what this action does you might need to specify `permissions` for it to work: https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token e.g. https://github.com/microsoft/vscode-jupyter/blob/bec563336a7cc7eed837835918fbd282b58469f1/.github/workflows/codeql-analysis.yml#L14-L15

joyceerhl (approved):

rchiodo (approved):

.github/workflows/test-plan-item-validator.yml
JacksonKearl: Thanks, added scopes

Issue #6779, opened 2021-07-23
Native interactive window export ignores outputs


joyceerhl:
Oversight; we should use the serializer. Thanks @DonJayamanne

Issue #6488, opened 2021-07-01
Remove webview notebook editor code


joyceerhl:
(WIP)

- [x] Delete `interactiveWindow.ts`, `interactiveWindowProvider.ts`, `interactiveWindowCommandListener.ts`, `interactiveBase.ts`

joyceerhl:
Discussion with @rebornix @rchiodo @DonJayamanne 

This iteration (Stage 0)
* Remove Live Share layers

Stage 1
Remove the UI 
Remove the INotebookEditorProvider
Remove INotebook (TBD? Native notebooks codepath also depends on INotebook)

Stage 2
Remove the ICell interface
* Rewrite export 
* Remove INotebookExecutionLogger preExecute and postExecute in favor of nativePostExecute (take NotebookCell instead of ICell)
How do we refactor execution so that things can be shared
What are we going to do for other languages 
Jupyter ext/Python ext --> npm module

Split out npm modules for
* Kernel discovery (trivial, look on disk? Need Python to discover certain kernels)
* Open sockets and start kernels
* Translating Jupyter outputs to VS Code outputs


--- Interactive window discussion ---

We need to be able to execute code silently against the underlying kernel so we cannot use pure VS Code APIs to execute code in the interactive window
* CWD
* __file__

Interactive window debug cell
* INotebookExecutionLoggers provide cell hashes and increment execution count, must preserve loggers for native interactive window port

DonJayamanne:
Closing as this work has been completed

PR #6784, opened 2021-07-23
Fix native interactive window export to preserve outputs and omit sys info cells

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/6779

@DonJayamanne FYI, we cannot simply use the notebook serializer to convert NotebookDocument to bytes/string like I initially thought, because we do all kinds of stuff to manipulate the notebook cells before serializing and writing it to disk e.g. inserting change directory, and we would need to rewrite all of that functionality for NotebookCell/NotebookDocument (I started down this path and had touched 10 files after an hour). 

@rebornix had a proposal that we should rewrite our export to deal with nbformat directly rather than ICells, which I agree with. In the meantime this PR is a localized fix for the original problem (export dropping notebook cell outputs) to avoid introducing regressions for native interactive window, but when we create an npm package with shared code for converting NotebookData to nbformat and delete ICell https://github.com/microsoft/vscode-jupyter/issues/6488, this will go away

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -350,7 +350,6 @@ module.exports = {
         'src/datascience-ui/react-common/codicon/codicon.ts',
         'src/datascience-ui/react-common/errorBoundary.tsx',
         'src/datascience-ui/common/main.ts',
-        'src/datascience-ui/common/cellFactory.ts',
         'src/datascience-ui/common/index.ts',
         'src/datascience-ui/startPage/index.tsx',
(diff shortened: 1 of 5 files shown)


DonJayamanne:
> sal that we should rewrite our export to deal with nbformat directly rather than ICells, 

Agreed, we'd first need to get rid of the noise and the consider building a better solution.

DonJayamanne (approved):

PR #6786, opened 2021-07-24
Functional tests for native interactive window

joyceerhl:
For https://github.com/microsoft/vscode-jupyter/issues/6516


joyceerhl:
I have more tests I want to add to this file, but I'd like to check this initial set in first. These tests have already caught some bugs.

src/client/datascience/editor-integration/codeLensFactory.ts
```diff
@@ -184,7 +184,7 @@ export class CodeLensFactory implements ICodeLensFactory {
             };
             this.notebookData.set(e.cell.notebook.uri.toString(), data);
         }
-        if (data) {
+        if (data && metadata.interactive !== undefined) {
```
DonJayamanne, line 187: Please can we use `getInteractiveCellMetadata` function.
Else tomorrow if we change the types/names its impossible to find out the usages.
I.e. use types as opposed to `any`

DonJayamanne (approved):

src/client/datascience/editor-integration/codeLensFactory.ts
DonJayamanne: E.g. when I had to change `executionId` to `id`, it wasn't easy to find all the usages, and easy to miss usages and break the code.

Issue #5996, opened 2021-05-26
Kernel restart lets cells run after restart


grapland0:
```
Version: 1.57.0-insider (user setup)
Commit: a18ea9c9ec25c59e599ba3c2eb668315546dd269
Date: 2021-05-26T05:14:06.346Z
Electron: 12.0.9
Chrome: 89.0.4389.128
Node.js: 14.16.0
V8: 8.9.255.25-electron.0
OS: Windows_NT x64 10.0.19043
```
```
Name: Jupyter
Id: ms-toolsai.jupyter
Description: Jupyter notebook support, interactive programming and computing that supports Intellisense, debugging and more.
Version: 2021.6.879154572
Publisher: Microsoft
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter
```

![recording-no-reliable-interrupt](https://user-images.githubusercontent.com/11003371/119704543-9f834f80-be0c-11eb-9c0b-0638722294fe.gif)

rchiodo:
Thanks for the bug. It looks like restart is not killing all cells.

grapland0:
Thank you! This only happens after the last 1~2 versions of vscode insider / jupyter ext updates.

PR #6826, opened 2021-07-26
Rework how restart works for native notebooks

rchiodo:
Fixes: #5996 

Root cause is the mechanism for how restart works was not signalling the cell execution. Refactored how we do restart to match how we do interrrupt.

### DIFF
--- a/src/client/datascience/jupyter/kernels/kernel.ts
+++ b/src/client/datascience/jupyter/kernels/kernel.ts
@@ -177,6 +177,9 @@ export class Kernel implements IKernel {
         const restartPromise = this.kernelExecution.restart(notebookDocument, this._notebookPromise);
         await restartPromise;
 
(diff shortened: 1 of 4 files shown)


src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -172,18 +172,13 @@ export class Kernel implements IKernel {
         if (this.restarting) {
             return this.restarting.promise;
         }
-        if (this.notebook) {
-            this.restarting = createDeferred<void>();
-            try {
-                await this.notebook.restartKernel(this.launchTimeout);
-                await this.initializeAfterStart(SysInfoReason.Restart, notebookDocument);
```
rchiodo, line 179: Oh I likely still need the sysinfo request

DonJayamanne (approved):

PR #6915, opened 2021-07-29
Fix call to getLocString to not have any default value

rchiodo:


rchiodo:
I'm going to enter a bug for getLocString to not have a default value. It should say something like 'loading' or something if the value isn't there.

src/datascience-ui/native-editor/nativeEditor.tsx
```diff
@@ -109,7 +109,7 @@ export class NativeEditor extends React.Component<INativeEditorProps> {
 ${buildSettingsCss(this.props.settings)}`}</style>
                 </div>
                 <div className="custom-editor-banner">
-                    <BannerTransform data={getLocString('DataScience.customEditorBanner', 'This is going away')} />
+                    <BannerTransform data={getLocString('DataScience.customEditorBanner', '')} />
```
joyceerhl, line 112: Sorry to keep harping on this, but if getLocString fails for whatever reason wouldn't this cause the banner to display empty? I always thought the default value needed to exactly match the entry in package.nls.json.

rchiodo: Then we don't realize we have a bug. I'd rather have a customer report the bug. 

Cause if it shows that, that means it's not being translated. Which is the real problem.

rchiodo: If I did that here, it would show english for half a second and then show the translated bits

joyceerhl: Right, I meant something like a failure to read the package.nls.json file contents off disk or similar

rchiodo: Yep or like this one where there's a second delay. If I had matched the text here, Jim would never have found the bug.

DavidKutu (approved):

Issue #5662, opened 2021-04-27
Execution of Notebooks are very slow compared to Classic Jupyter


DonJayamanne:
Run 200 cells in classic Jupyter = 0.5s
Run 200 cells in custom Editor = 9s
Run 200 cells in Native Editor = 35s

Last year I recall making some changes to ensure the perf was faster

**Suggestions:**
* Synchronize messages only when we have widgets (I did this last year, either the code has been accidentally removed or still there & not workign, or other)
* Don't wait for VS Code to update the output

DonJayamanne:
We're back at 9-10s in Native notebooks, back to what it was with Custom Editor Notebooks.

Not sure we'll be able to get to 0.5s, as we run each cell sequentially instead of sending all cells to Jupyter Kernel (i.e. we run a cell, check if it fails, then proceed to the next cell).
Considering this only decreases perf when running a large number of cells, I think this is acceptable (200 cells in 10s is accpetable with this architecture),

DonJayamanne:
Dup of #6941

PR #6946, opened 2021-08-02
Minimize unnecessary message deserialization

DonJayamanne:
Part of https://github.com/microsoft/vscode-jupyter/issues/5662
Simple change, deserialization is a little slower, hence using plain string operations over deserialization.

### DIFF
--- a/src/client/datascience/ipywidgets/ipyWidgetMessageDispatcher.ts
+++ b/src/client/datascience/ipywidgets/ipyWidgetMessageDispatcher.ts
@@ -219,7 +219,7 @@ export class IPyWidgetMessageDispatcher implements IIPyWidgetMessageDispatcher {
     private async mirrorSend(data: any, _cb?: (err?: Error) => void): Promise<void> {
         // If this is shell control message, mirror to the other side. This is how
         // we get the kernel in the UI to have the same set of futures we have on this side
-        if (typeof data === 'string') {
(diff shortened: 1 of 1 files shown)


src/client/datascience/ipywidgets/ipyWidgetMessageDispatcher.ts
```diff
@@ -275,12 +275,22 @@ export class IPyWidgetMessageDispatcher implements IIPyWidgetMessageDispatcher {
         let message;
 
         if (!this.isUsingIPyWidgets) {
-            if (!message) {
+            // Lets deserialize only if we know we have a potential case
+            // where this message contains some data we're interested in.
+            let mustDeserialize = false;
+            if (typeof data === 'string') {
+                mustDeserialize = data.includes(WIDGET_MIMETYPE) || data.includes(Identifiers.DefaultCommTarget);
```
rchiodo, line 282: This is a pretty big assumption. How do you know that all messages for a widget will fit into these subcategories? And what about when we start supporting synapse?

What's the perf improvement of this change? Seems risky to me.

rchiodo (approved):

src/client/datascience/ipywidgets/ipyWidgetMessageDispatcher.ts
rchiodo: Misread the section below. THis is just like the previous change.

PR #6955, opened 2021-08-03
Remove Live share code (based on Davids work)

DonJayamanne:


DonJayamanne:
Cannot use Davids branch as it removes code we need (I think we have merged some code incorrectly)
E.g. the followign code has been removed in errorHandler.ts, & that's wrong.

PR #6967, opened 2021-08-04
First pass at removing webview code

DonJayamanne:
This is not the final PR, 
**Here I'm removing** 
* Initially focusing on the UI code
* some of the redundant NPM packages (used by webview code)
* some of the webview code (ensuring existing stuff compiles, & tests run)
	* 	E.g. we need variable viewer, data frame viewer, etc
* Subsequent PRs will remove code from the extension host  & then later clean up...

Probably 2-3 more PRs to come


src/client/datascience/interactive-common/interactiveBase.ts
```diff
@@ -321,14 +318,6 @@ export abstract class InteractiveBase extends WebviewPanelHost<IInteractiveWindo
                 this.handleMessage(message, payload, this.requestVariables);
```
rchiodo, line 318: What is this file still used for? Seems like it can go away?

DonJayamanne: Please see comments, I'm going to remove the code from extension in a separate PR.
Else this PR gets way too big & too difficult, this way I can just remove the UI code & the bundle & later focus on extension & then focus on cleaning up everything (3 prs).
Primarily because I find that easier, else I'd still be working on cleaning up extension code (this PR gets the UI out)

joyceerhl (approved):

PR #6968, opened 2021-08-04
Fixes to CI

DonJayamanne:


package.json
```diff
@@ -2123,6 +2123,7 @@
         "detect-indent": "^6.0.0",
         "diff-match-patch": "^1.0.0",
         "download": "^8.0.0",
+        "encoding": "^0.1.13",
```
DonJayamanne, line 2126: node-fetch requires `encoding`.
Somehow this isn't getting installed today. I'm guessing something changed (some dependency) that used to automatically pull in `encoding`, & now doesn't. 
Hence we don't end up with `encoding` package & things fall over.

node-fetch 3.0 fixes this by having an explicit dependency.

rchiodo (approved):

PR #6973, opened 2021-08-05
Remove Old Notebook Editor & Interactive Window code

DonJayamanne:
For https://github.com/microsoft/vscode-jupyter/issues/6488


DavidKutu (approved):

DavidKutu:
The new interactive window was using interactiveBase.ts IIRC. There might be some changes needed there.

joyceerhl:
> The new interactive window was using interactiveBase.ts IIRC. There might be some changes needed there.

Nope, the native interactive window does not reference interactiveBase.ts

DonJayamanne:
> The new interactive window was using interactiveBase.ts IIRC. There might be some changes

Btw, I have tested Native Notebooks, Native Interactive Window, debugging interactive window & debugging notebooks.
hence My assuption is most of it seems to be ok.

src/client/common/application/notebook.ts
```diff
@@ -36,52 +36,35 @@ export class VSCodeNotebook implements IVSCodeNotebook {
...
-            >();
-            this.onDidChangeActiveNotebookEditor = this.createDisposableEventEmitter<NotebookEditor | undefined>();
-            this.onDidOpenNotebookDocument = this.createDisposableEventEmitter<NotebookDocument>();
-            this.onDidCloseNotebookDocument = this.createDisposableEventEmitter<NotebookDocument>();
-            this.onDidChangeVisibleNotebookEditors = this.createDisposableEventEmitter<NotebookEditor[]>();
-            this.onDidSaveNotebookDocument = this.createDisposableEventEmitter<NotebookDocument>();
-            this.onDidChangeNotebookDocument = this.createDisposableEventEmitter<NotebookCellChangedEvent>();
-        }
+        this.addEventHandlers();
+        this.canUseNotebookApi = true;
```
joyceerhl, line 60: Can we remove this property altogether?

DonJayamanne: Oops, missed, that

DonJayamanne: I'm sure there will be a lot more...

src/client/telemetry/importTracker.ts
```diff
@@ -164,12 +143,12 @@ export class ImportTracker implements IExtensionSingleActivationService, INotebo
...
-
-    private onOpenedOrClosedNotebook(e: INotebookEditor) {
-        if (e.file) {
-            this.scheduleCheck(e.file.fsPath, this.checkNotebook.bind(this, e));
+        } else if (
+            document.notebook &&
+            isJupyterNotebook(document.notebook) &&
+            document.languageId === PYTHON_LANGUAGE
+        ) {
+            this.scheduleDocument(document);
```
joyceerhl, line 151: I assume this is the fix for package import telemetry not showing up?

src/test/datascience/activation.unit.test.ts
```diff
@@ -2,122 +2,120 @@
 // Licensed under the MIT License.
 
 'use strict';
+export const __ = '';
```
joyceerhl, line 5: Why is this test commented out?

DonJayamanne: Simple, if you check the code, we have Notebook Editor stuff in there.

src/client/telemetry/importTracker.ts
DonJayamanne: **This fix is for when running cells**, we're already sending telemetry for Opening & Closing notebook.s
I think they may be checking telemetry in Python repo (coz if they're talking to Luciana and looking at their telelemtry, well then it comes from Python repo).

& Possible they haven't looked at our repo for the telemtry (explains why they do'nt see telemetry for open/close of notebooks)

DonJayamanne: Oh and I wanted to ensure we only look for Jupyter notebooks with Python cells, previously it was any notebook.

src/test/datascience/activation.unit.test.ts
joyceerhl: Sure but doesn't the test still apply for native notebooks?

src/test/smoke/datascience.smoke.test.ts
```diff
@@ -39,32 +39,32 @@ suite('Smoke Tests', () => {
...
+    //     console.log('Step0');
+    //     await sleep(1_000);
+    //     console.log('Step1');
+    //     await vscode.commands.executeCommand<void>('jupyter.runallcells', textDocument.uri);
+    //     console.log('Step2');
+    //     const checkIfFileHasBeenCreated = () => fs.pathExists(outputFile);
+    //     console.log('Step3');
+    //     await waitForCondition(checkIfFileHasBeenCreated, timeoutForCellToRun, `"${outputFile}" file not created`);
+    //     console.log('Step4');
+    // }).timeout(timeoutForCellToRun);
```
joyceerhl, line 67: This test should still be valid?

joyceerhl (approved):

src/test/smoke/datascience.smoke.test.ts
DonJayamanne: Yes, but its failing, will fix after I merge

joyceerhl: You might need this change https://github.com/microsoft/vscode-jupyter/pull/6786/files#diff-0e9199b84e8377970515123cf9f3d1c2e60d1a1379507cbade651126ae15934eR420-R423 We have test code that forces a kernel selection, and that code will force a selection for the wrong controller (the one registered for jupyter-notebook). This causes execution to fail because the controller won't be correctly associated with the notebook viewtype. We need to update that code to not force a selection for the interactive window, as it gets started with the right controller id anyway

PR #6974, opened 2021-08-05
Run by Line: disconnect at last line

DavidKutu:
For #6858


src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -313,6 +335,25 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter {
             }
         });
 
+        if ((message as DebugProtocol.StackTraceResponse).command === 'stackTrace') {
+            (message as DebugProtocol.StackTraceResponse).body.stackFrames.forEach((sf) => {
+                // Check if we're stopped at the last line
+                let currentCell: NotebookCell | undefined;
+                this.notebookDocument.getCells().forEach((cell) => {
+                    const index = sf.source?.path!.indexOf('#ch');
```
DonJayamanne, line 343: this should be a helper function, give a stack we should find the corresponding cell.
I'm assuming we have similar code today elsewhere.
Also no idea what `#ch` is here.

DonJayamanne, line 343: This code can go outside the forEach loop, this doesn't change, hence no need to get index & fragment everytime for each cell.

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -116,8 +117,11 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter {
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
             const content = msg.content as any;
             if (content.event === 'stopped') {
-                this.runByLineThreadId = content.body.threadId;
-                this.runByLineSeq = content.seq;
+                if (this.isRunByLine) {
```
DonJayamanne, line 120: Does this code work if i have a cell as follows:
```
def do_something():
    print("Hello")


do_something()
```

Now, when I'm at the last line, will I be able to step in?

DonJayamanne: Or what if i add a breakpoint at line `print("Hello")`
& then hit `continue`, surely it shouldn't stop the run by line, but should step into that code.

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: I'll put it on a helper. The `#ch` is to find the fragment in the stackFrame

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: No, I don't remember how it used to work, I'll check

DavidKutu: Looks like run by line in the webview doesn't work on stable. I'll go for it anyway.

DavidKutu: The new approach covers that scenario

rchiodo (approved):

PR #7019, opened 2021-08-09
Enable export to HTML/Python tests

joyceerhl:


joyceerhl:
Was looking at some other failures on a different PR and it looks like these tests haven't run for a year. I think we do want these tests so reenabling.

Issue #1652, opened 2020-08-09
Implement full notebook debugger


tyoc213:
Feature: Notebook Editor

## Been able to debug ipynb cells

I know I can debug "cells" from within a .py file as

![image](https://user-images.githubusercontent.com/506234/89743763-dea60700-da6b-11ea-872b-d676bcff57fc.png)

Bu there is no way to launch debug like this from a cell in a ipynb

![image](https://user-images.githubusercontent.com/506234/89743781-0006f300-da6c-11ea-9904-a8f0c250cee7.png)

It would love to launch the debugger "normally" with a button, if the breakpoints are not respected in the cells, at less, if they are in other .py that code has access at the moment because "in the working directory" would love that those breakpoints are respected as normal .py.

Microsoft Data Science for VS Code Engineering Team: @rchiodo, @IanMatthewHuff, @DavidKutu, @DonJayamanne, @greazer, @joyceerhl

tyoc213:
Is this already implemented or public in latest version of vscode?

joyceerhl:
@tyoc213 Hi! Full cell debugging is not yet available in the preview native notebooks editor, but Run By Line (a lite debugger that allows you to execute code cells and view variable state line by line) is available in the webview based notebook editor. Run By Line feature parity in the preview native notebooks editor is on our roadmap. This is what Run By Line looks like in the webview based notebook editor:

<img width="607" alt="Screen Shot 2020-12-31 at 10 17 18 AM" src="https://user-images.githubusercontent.com/30305945/103421476-77d7d600-4b51-11eb-8bdb-b06bb72b6e89.png">

nerkulec:
If I were to contribute to help resolve this particular issue, where would I start? I've already read the 'contributing guide'.

rchiodo:
> If I were to contribute to help resolve this particular issue, where would I start? I've already read the 'contributing guide'.

What were you thinking of contributing?

Our plans for debugging are:
- Interactive window will continue to support debug cell - this is working today
- Stable VS code will show the Run by Line support that @joyceerhl outlined above - this is working today
- Insiders VS code will eventually get Run by line as well. - this requires help from the VS code team at the moment.

nerkulec:
> > If I were to contribute to help resolve this particular issue, where would I start? I've already read the 'contributing guide'.
> 
> What were you thinking of contributing?
> 
> Our plans for debugging are:
> 
> * Interactive window will continue to support debug cell - this is working today
[...]

PR #7059, opened 2021-08-11
Add "Debug Cell"

roblourens:
For microsoft/vscode#130524

And #1652

- Add a "Debug Cell" command
- Remove the "Debug" button from the toolbar

A little flakier than RBL, I keep having this issue where I send the 'disconnect' request but the Debug Session does not actually terminate, seen that before? Seems to work fine for RBL.

### DIFF
--- a/src/client/debugger/jupyter/debuggingManager.ts
+++ b/src/client/debugger/jupyter/debuggingManager.ts
@@ -218,7 +218,7 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
                 if (!cell) {
                     const range = editor?.selections[0];
                     if (range) {
-                        cell = editor.document.cellAt(range.start);
+                        cell = editor?.document.cellAt(range.start);
                     }
                 }
 
--- a/src/client/debugger/jupyter/debuggingManager.ts
+++ b/src/client/debugger/jupyter/debuggingManager.ts
@@ -20,7 +20,7 @@ import {
 import * as path from 'path';
 import { IKernelProvider } from '../../datascience/jupyter/kernels/types';
 import { IDisposable } from '../../common/types';
-import { KernelDebugAdapter } from './kernelDebugAdapter';
+import { IKernelDebugAdapterConfig, KernelDebugAdapter, KernelDebugMode } from './kernelDebugAdapter';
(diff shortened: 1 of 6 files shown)


rchiodo (approved):

DavidKutu:
The Lint test is failing because you need to run prettier on:
src/client/debugger/jupyter/debuggingManager.ts
src/client/debugger/jupyter/kernelDebugAdapter.ts

IanMatthewHuff (approved):

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -234,22 +244,40 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
...
-    private async startDebugging(doc: NotebookDocument, cell?: NotebookCell, options?: DebugSessionOptions) {
+    private async startDebuggingCell(doc: NotebookDocument, mode: KernelDebugMode.Cell | KernelDebugMode.RunByLine, cell: NotebookCell) {
+        const config: IKernelDebugAdapterConfig = {
+            type: pythonKernelDebugAdapter,
+            name: path.basename(doc.uri.toString()),
+            request: 'attach',
+            internalConsoleOptions: 'neverOpen',
+            justMyCode: true,
+            // add the doc uri to the config
+            __document: doc.uri.toString(),
```
roblourens, line 255: @DavidKutu is this used anywhere? I couldn't actually find a usage

roblourens:
> I keep having this issue where I send the 'disconnect' request but the Debug Session does not actually terminate, seen that before? Seems to work fine for RBL.

@DavidKutu any idea about that?

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: Yeah, I guess I removed it. We send the NotebookDocument as a parameter anyway. I'll remove it on my next PR.

DavidKutu:
I don't see how you're sending the disconnect. Is it with the normal debugging widget?
I haven't had issues with neither normal debugging nor run by line

roblourens:
The `disconnect` method gets called in the same way it is for RBL when the cell execution is finished. If I click the disconnect button, it works correctly

PR #7067, opened 2021-08-11
Interactive window should use `IKernel.executeCell`

### LINKED ISSUE #6983: Kernel Interrupt not working in interactive window for cells submitted from a #%% file
Version: 1.59.0 (system setup)
Commit: 379476f0e13988d90fab105c5c19e7abc8b1dea8
Date: 2021-08-04T23:13:12.822Z
Electron: 13.1.7
Chrome: 91.0.4472.124
Node.js: 14.16.0
V8: 9.1.269.36-electron.0
OS: Windows_NT x64 10.0.22000

1. Run a long running notebook using a python script with #%% cells, run it via a codelens then try to stop it with Interrupt Kernel. 

It doesn't stop. No notification that it's even trying to stop it as is expected.

joyceerhl:
This PR fixes https://github.com/microsoft/vscode-jupyter/issues/6983 and https://github.com/microsoft/vscode-jupyter/issues/7081

Interrupt kernel doesn't work for cells submitted from a #%% file because they're not part of the cell execution queue, and the interrupt command bails early if it sees that there's no available execution queue. The fix was to rewrite the native interactive window addCode/debugCode codepath to use a unified `IKernel.executeCell` codepath. This entailed modifying the CellHashProvider to work with NotebookCells so that debugging would work correctly. 

One complication is that interactive window debugging relies on us injecting `breakpoint()` into the code that is submitted to the kernel, which was previously accomplished with the `ICell` model. Since the `CellExecution` and `KernelExecution` classes deal only with `vscode.NotebookCell`s, this meant we needed a way to represent the modified source code in the `KernelExecution` path. For now this is in the metadata since this change is already quite large, but in future we should consider introducing the notion of an `InteractiveCell` object that extends `vscode.NotebookCell` and ensuring that the `KernelExecution` path can handle that.


rchiodo (approved):

DavidKutu (approved):

src/client/datascience/types.ts
```diff
@@ -937,7 +936,7 @@ export interface ICellHashProvider {
     getHashes(): IFileHashes[];
     getExecutionCount(): number;
     incExecutionCount(): void;
-    generateHashFileName(cell: ICell, executionCount: number): string;
+    onBeforeCellExecute(notebookCell: NotebookCell): Promise<void>;
```
DonJayamanne, line 939: I'd rename this to `addCellHash` or something.
We know why we're calling this method, its to generate a hash. Hvaing a method named `onBeforeCellExecute` means its just an event handler & we have no idea what its doing.

I.e. i'd prefer us to be more explicit about the action we're performing. Else this is similar to a logger, kind of a black box, but that's not the case (i.e. we know what this class is meant to be used for).

DonJayamanne (changes requested):
Please rename the method.

DonJayamanne (approved):

PR #7070, opened 2021-08-11
Fix interrupt kernel for interactive window

joyceerhl:
Discussed with @DonJayamanne  and while https://github.com/microsoft/vscode-jupyter/pull/7067 is the right fix, it's too risky for the recovery build and would also break the webview interactive window). Since #6983 is a recovery build candidate, this PR is a temporary workaround for the root cause (which is that the interactive window uses executeObservable and that results in the IKernel not being aware of cells executed that way when doing IKernel.interrupt)

### DIFF
--- a/src/client/datascience/jupyter/kernels/kernelExecution.ts
+++ b/src/client/datascience/jupyter/kernels/kernelExecution.ts
@@ -87,6 +87,35 @@ export class KernelExecution implements IDisposable {
             traceInfo('Restore notebook state to idle after completion');
         }
     }
+
+    /**
+     * This is a temporary method to support interrupting kernels when code cells are executed
+     * outside of the kernelExecution/cellExecution codepath. Do not use!
+     */
+    public async interruptInteractiveKernel(document: NotebookDocument, notebookPromise?: Promise<INotebook>): Promise<InterruptResult> {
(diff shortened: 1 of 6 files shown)


src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -87,6 +87,35 @@ export class KernelExecution implements IDisposable {
             traceInfo('Restore notebook state to idle after completion');
         }
     }
+
+    /**
+     * This is a temporary method to support interrupting kernels when code cells are executed
+     * outside of the kernelExecution/cellExecution codepath. Do not use!
+     */
+    public async interruptInteractiveKernel(document: NotebookDocument, notebookPromise?: Promise<INotebook>): Promise<InterruptResult> {
```
rchiodo, line 95: What's the difference between this method and interrupt()? Seems almost the same.

joyceerhl: It creates a deferred if there's no execution queue and lets the execution time out.

rchiodo (approved):

IanMatthewHuff (approved):

DavidKutu (approved):

PR #7083, opened 2021-08-11
Clicking the wrong RBL button should be a no-op

### LINKED ISSUE #6868: Run by line: Fails in two notebooks side by side
Testing microsoft/vscode#129467

- One notebook open already
- Open another to the side
- Attempt to verify "Test run by line in two notebooks simultaneously"
- :bug: Unexpected behavior, see recording below

recording (41)

If I should be testing two notebooks simultaneously in a different way, or if I can provide any additional info/logs, please just let me know.

### LINKED ISSUE #6905: Unexpected cell switching behavior in run by line
Testing https://github.com/microsoft/vscode/issues/129467

recording (62)

roblourens:
Fix #6868
And fix #6905

Ideally we would hide the RBL buttons on cells that are not doing RBL, but we really don't have a way to do that with context keys right now, so I just make the ones on other notebooks or cells no-op

### DIFF
--- a/src/client/debugger/jupyter/debuggingManager.ts
+++ b/src/client/debugger/jupyter/debuggingManager.ts
@@ -197,19 +197,15 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
 
             this.commandManager.registerCommand(DSCommands.RunByLineContinue, (cell: NotebookCell) => {
                 const adapter = this.notebookToDebugAdapter.get(cell.notebook);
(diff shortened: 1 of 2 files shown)


rchiodo (approved):

DavidKutu (approved):

roblourens:
Is it ok that some failed? Seems like random failures

DavidKutu:
Yes, they're unrelated

Issue #7016, opened 2021-08-09
Restarting notebook changes jupyter root directory


darim1:
## Environment data

-   VS Code version: 1.59
-   Jupyter Extension version (available under the Extensions sidebar): v2021.8.1195043623
-   Python Extension version (available under the Extensions sidebar): v2021.8.1105858891
-   OS (Windows | Mac | Linux distro) and version: Windows 10 
-   Python and/or Anaconda version: 3.9.2
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): pipenv
-   Jupyter server running: Local 

## Expected behaviour

Restarting the kernel changes the root directory according to the source code file location.

## Actual behaviour

Restarting the kernel keeps specified directory, specified via extension setting "Notebook File Root"

## Steps to reproduce:

1. Create project in workspace "workspace/project/
2. Configure source file: "workspace/project/blabla/my_file.py, containing code cell
3. Configure extension setting "Notebook File Root" to /workspace/project
4. Run code cell => root should be /workspace/project
5. Restart kernel => root will be /workspace/project/blabla

## Logs

```
Info 2021-08-09 17:59:08: Setting setActiveController for vscode-interactive:Interactive-1.interactive
Info 2021-08-09 17:59:08: Intiailize notebook communications for editor vscode-interactive:Interactive-1.interactive
Info 2021-08-09 17:59:08: Resolving notebook UI Comms (resolve) for vscode-interactive:Interactive-1.interactive
Info 2021-08-09 17:59:08: initialize CommonMessageCoordinator
Info 2021-08-09 17:59:08: [32mPosting message to Notebook UI IPyWidgets_onKernelChanged
Info 2021-08-09 17:59:08: IPyWidgetScriptSource.initialize
Info 2021-08-09 17:59:08: Process Execution: > d:\work\kov2-ibr\git\mock-controller\.venv\Scripts\python.exe -m pip list
> d:\work\kov2-ibr\git\mock-controller\.venv\Scripts\python.exe -m pip list
Info 2021-08-09 17:59:08: IPyWidgetMessageDispatcher.initialize
Info 2021-08-09 17:59:08: Created and initailized CommonMessageCoordinator
Info 2021-08-09 17:59:08: Coordinator already attached for vscode-interactive:Interactive-1.interactive
Info 2021-08-09 17:59:08: Dispose kernel vscode-interactive:Interactive-1.interactive
[...]

Issue #7086, opened 2021-08-12
Notebook initialize code (startup commands, custom directory) is not applied on native notebook restart


IanMatthewHuff:
On restart of native notebooks the restartKernel function of the notebook is not being applied. This means that things like custom working directories and custom startup commands are not applied on restart as expected.

joyceerhl:
This was on purpose to fix https://github.com/microsoft/vscode-jupyter/issues/5996, because the old restartKernel method would wait for pending execution to complete instead of cancelling them all the way that kernel interrupt did /cc @rchiodo

jaymegordo:
@IanMatthewHuff would jupyter startup commands eg:
```
    "jupyter.runStartupCommands": [
        "%alias_magic t timeit",
        "%load_ext autoreload",
        "%autoreload 2"
    ],
```

Fall under what you're talking about? I'm just noticing these don't work now in the native interactive window (I'm gonna use NIW from now on haha). Just wondering if you guys are aware or I should open a separate issue.

Not having `autoreload` makes the NIW pretty hard to use tbh so hopefully @joyceerhl 's PR resolves it?

jaymegordo:
Ahh yes, the startup commands do work on first open of IW, but fail after a "restart". Thats why I was only noticing it sometimes.

IanMatthewHuff:
@joyceerhl Yeah, I was looking at it a bit. In the old restart the cell cancellations are mixed in with the initialization steps. We'd want to separate out just the init code so that we can run the re-initialization without mixing up the old and new methods of cell cancellation. The init function could probably just take any old JupyterNotebook as an input.

rchiodo:
Remove restart and just replace with kernel start.

rchiodo:
questions about whether we wait for full refactor or not.

joyceerhl:
This has been fixed https://github.com/microsoft/vscode-jupyter/blob/2e1c221cdc3ec1128d55886711677f77d699e552/src/client/datascience/jupyter/kernels/kernel.ts#L359-L377

PR #7089, opened 2021-08-12
Rerun kernel initialization after restarting kernel session

joyceerhl:
This PR fixes https://github.com/microsoft/vscode-jupyter/issues/7016 and https://github.com/microsoft/vscode-jupyter/issues/7086

### DIFF
--- a/src/client/datascience/jupyter/kernels/kernelExecution.ts
+++ b/src/client/datascience/jupyter/kernels/kernelExecution.ts
@@ -150,9 +150,7 @@ export class KernelExecution implements IDisposable {
         }
 
         // Restart the active execution
-        await (this._restartPromise
-            ? this._restartPromise
-            : (this._restartPromise = this.restartExecution(document, notebook.session)));
+        await (this._restartPromise ? this._restartPromise : (this._restartPromise = this.restartExecution(notebook)));
 
         // Done restarting, clear restart promise
         this._restartPromise = undefined;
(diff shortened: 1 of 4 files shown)


src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -268,11 +266,14 @@ export class KernelExecution implements IDisposable {
...
-    private async restartExecution(_document: NotebookDocument, session: IJupyterSession): Promise<void> {
+    private async restartExecution(notebook: INotebook): Promise<void> {
         // Just use the internal session. Pending cells should have been canceled by the caller
-        return session.restart(this.interruptTimeout).catch((exc) => {
+        await notebook.session.restart(this.interruptTimeout).catch((exc) => {
             traceWarning(`Error during restart: ${exc}`);
         });
+
+        // Reinitialize the kernel after a session restart
+        await notebook.runInitialSetup();
```
DonJayamanne, line 276: This was never required for Native notebooks, now we have code thats specific to interactive widow.
I think we should put an `if` condition for this and ensure this only runs for Intearctive window, as we have initialization for Native Notebooks already in the IKernel clas.

DonJayamanne (changes requested):
Interactive window specific code should be gated (temporarily).
Once we move interactive widow to use IKernel then this code will not be necessary or we should move the initialization to the IKernel class.

src/client/datascience/jupyter/kernels/kernelExecution.ts
joyceerhl: This is required for native notebooks, and this is how restart used to work until 2 weeks ago. See https://github.com/microsoft/vscode-jupyter/pull/6826/files#diff-7d0d36cd88827937fc9458d80a807484935c9d2b5d63501a6ec98eb1e5351c16L178

DonJayamanne: I don't see any old code that called `notebook.runInitialSetup`
That's new, Or an I missing something

joyceerhl: That's a new method I added to expose the initialization logic which used to run as part of INotebook.restartKernel

rchiodo (approved):

src/client/datascience/jupyter/kernels/kernelExecution.ts
DonJayamanne: > PR is to move the initialization logic to IKernel.initializeAfterStart. What do you think?

Yes, thats much better, that way all the initialization is in one place.

DonJayamanne (approved):

IanMatthewHuff (approved):

joyceerhl:
Closing as I think we want to rework how restart works and start a new kernel instance instead so that the INotebook / IKernel object doesn't need to remember whether it's been restarted.

PR #7100, opened 2021-08-12
Remove all awaits when updating cell outputs

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/7101

### DIFF
--- a/src/client/datascience/raw-kernel/rawSocket.ts
+++ b/src/client/datascience/raw-kernel/rawSocket.ts
@@ -51,6 +51,7 @@ export class RawSocket implements IWebSocketLike, IKernelSocket, IDisposable {
     ) {
         // Setup our ZMQ channels now
         this.channels = this.generateChannels(connection);
+        console.log(this.serialize.length);
     }
 
     public dispose() {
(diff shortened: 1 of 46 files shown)


DonJayamanne:
Here is what is failing today with these changes:
# 1. Outputs are lost
Run the following cell, the expected outupt is `999\n Done: xxxx`
Sometimes, we don't get anything in the cell output.
If you are unable to repro, please change the counter from `1000` to `10000` or `3000` or other.
No idea what the magic number is, for some reason `1000` repros the bug, but `10,000` does not!
```python
from timeit import default_timer as timer
from IPython import display

start = timer()

for i in range(1000):
    display.clear_output()
    print(i)

end = timer()
print(f"Done: {end - start}")
```

# 2. UI not updated with changes to the model.
Create first cell with the following code:
```python
from IPython.display import Markdown
dh = display(display_id=True)
```
Run the above cell, then run the following cell:
```python
dh.update(Markdown("foo"))
```

After you run the second cell, the first cell should have an output `foo`.
Collapse the first cell & expand it, and you'll see `foo` appear, basically its a refresh issue.

src/client/datascience/raw-kernel/rawSocket.ts
```diff
@@ -51,6 +51,7 @@ export class RawSocket implements IWebSocketLike, IKernelSocket, IDisposable {
     ) {
         // Setup our ZMQ channels now
         this.channels = this.generateChannels(connection);
+        console.log(this.serialize.length);
```
rchiodo, line 54: Did you mean to leave this here?

src/client/datascience/raw-kernel/rawSocket.ts
```diff
@@ -159,10 +160,48 @@ export class RawSocket implements IWebSocketLike, IKernelSocket, IDisposable {
 
         // Wire up all of the different channels.
         const result: IChannels = {
-            iopub: this.generateChannel(connection, 'iopub', () => new zmq.Subscriber()),
-            shell: this.generateChannel(connection, 'shell', () => new zmq.Dealer({ routingId })),
-            control: this.generateChannel(connection, 'control', () => new zmq.Dealer({ routingId })),
-            stdin: this.generateChannel(connection, 'stdin', () => new zmq.Dealer({ routingId }))
+            iopub: this.generateChannel(
```
rchiodo, line 163: Maybe a comment about the highWatermark stuff so next person knows why this is here?

src/test/datascience/notebook/executionService.mock.vscode.test.ts
```diff
@@ -0,0 +1,221 @@
...
+        const kernel = createKernelWithMockJupyterSession(cell.notebook, session);
+
+        // Wait till execution count changes and status is success.
+        await Promise.all([
+            kernel?.executeCell(cell),
+            waitForTextOutput(cell, '123412341234', 0, false),
+            waitForExecutionCompletedSuccessfully(cell)
+        ]);
+    });
+    test('Execute a cell that prints and clears the output 100s of times (once each iteration)', async () => {
```
rchiodo, line 195: Should we add some sort of perf analysis to this one? Jupyter is out of the loop so all perf problems should be in the UI, right?

src/client/datascience/raw-kernel/rawSocket.ts
DonJayamanne: oops

DonJayamanne: Fixed

rchiodo (approved):
This is freaking awesome but scary. I'd love to bug bash the hell out of this. Find all the issues where you added the execution queueing etc and make sure they still work.

src/test/datascience/notebook/executionService.mock.vscode.test.ts
DonJayamanne: I want to do that in a separate PR, 
* THis PR is large as is
* Got some more perf fixes (unnecessary serialization/deserialization)
* Perf related to string concatenation (handling streams & ANSI codes)
* Test, tests, tests. test

DonJayamanne:
@rchiodo 

> This is freaking awesome but scary. I'd love to bug bash the hell out of this. Find all the issues where you added the execution queueing etc and make sure they still work.

I didnt find any bug that our tests didn't cath, i.e. I'm glad we have some very extensive tests. Identified 4 bugs in VS Code as a result of this refactor via our tests.

I.e. I'm feeling pretty confident at this stage, but would like to add more tests.
Its very satisfying to see tests catch real bugs abd block PRs, specially with large refactoring such as this.

DavidKutu (approved):

Issue #6913, opened 2021-07-29
SVG Plot viewer perf fix


DonJayamanne:
* Stop generating SVG
* Display plot viewer with PNGs
* Ensure exporting notebooks to PDF works even without SVG

IanMatthewHuff:
One question here to examine is what do customers want from the plot viewer. We have feedback that they want it, but do they really care about zoom, pan, ect? Does just a basic image list work for 90% of users? Also are we worried enough about the perf issues here that we implement 1 - 3 quickly while we work on a bigger solution.

IanMatthewHuff:
Yeah, that sounds like a useful side feature. We can support images or SVGs. But we could still have an opt-in feature to generate SVGs alongside plot images automatically. I don't think that any of the solutions here would preclude that, and it's a good idea.

PR #7140, opened 2021-08-14
Default plots to PNG output, and support showing PNGs or SVGs in the existing PlotViewer control

IanMatthewHuff:
For #6913 


package.json
```diff
@@ -1809,8 +1809,8 @@
                 },
                 "jupyter.enablePlotViewer": {
                     "type": "boolean",
-                    "default": true,
-                    "description": "Modify plot output so that it can be expanded into a plot viewer window.",
+                    "default": false,
```
IanMatthewHuff, line 1812: We might need to discuss this. I feel that customers might still want the old behavior, so I left the setting in, but defaulted it to false instead of true, so they can opt into getting the extra SVGs if they want. My code will prefer SVG output to any PNG output if present.

src/client/datascience/constants.ts
```diff
@@ -659,10 +659,10 @@ export namespace CodeSnippets {
...
-    export const MatplotLibInitSvg = `import matplotlib\n%matplotlib inline\n${Identifiers.MatplotLibDefaultParams} = dict(matplotlib.rcParams)\n%config InlineBackend.figure_formats = {'svg', 'png'}`;
-    export const MatplotLibInitPng = `import matplotlib\n%matplotlib inline\n${Identifiers.MatplotLibDefaultParams} = dict(matplotlib.rcParams)\n%config InlineBackend.figure_formats = {'png'}`;
-    export const ConfigSvg = `%config InlineBackend.figure_formats = {'svg', 'png'}`;
-    export const ConfigPng = `%config InlineBackend.figure_formats = {'png'}`;
+    export const MatplotLibInitSvg = `import matplotlib\n%matplotlib inline\n${Identifiers.MatplotLibDefaultParams} = dict(matplotlib.rcParams)\n%config InlineBackend.figure_formats = ['svg', 'png']`;
+    export const MatplotLibInitPng = `import matplotlib\n%matplotlib inline\n${Identifiers.MatplotLibDefaultParams} = dict(matplotlib.rcParams)\n%config InlineBackend.figure_formats = ['png']`;
+    export const ConfigSvg = `%config InlineBackend.figure_formats = ['svg', 'png']`;
+    export const ConfigPng = `%config InlineBackend.figure_formats = ['png']`;
```
IanMatthewHuff, line 665: Looks like this has been broken for a long time / forever? For multiples it seems to take {} or [], but for the singular it can't:
image
Docs are a little funny on this, they say set, but the example uses a list?
image

Guessing we didn't notice this before as png is Jupyter default I'm pretty sure. So it would fail to run, but then just get PNG anyways from the Jupyter default.

src/client/datascience/jupyter/jupyterNotebook.ts
```diff
@@ -821,6 +823,16 @@ export class JupyterNotebookBase implements INotebook {
             // get a request to update style
             await this.executeSilently(matplobInit, cancelToken);
 
+            const useDark = this.applicationService.activeColorTheme.kind === ColorThemeKind.Dark;
```
IanMatthewHuff, line 821: This code was in the kernel.ts version, but not here so I added it as it looks correct for native theming.

src/client/datascience/jupyter/jupyterNotebook.ts
```diff
@@ -329,11 +329,13 @@ export class JupyterNotebookBase implements INotebook {
...
+                //(!settings || settings.enablePlotViewer) &&
+                //!isResourceNativeNotebook(this._resource, this.vscNotebook, this.fs)
+                //? CodeSnippets.ConfigSvg
+                //: CodeSnippets.ConfigPng;
                 const configInit =
-                    (!settings || settings.enablePlotViewer) &&
-                    !isResourceNativeNotebook(this._resource, this.vscNotebook, this.fs)
-                        ? CodeSnippets.ConfigSvg
-                        : CodeSnippets.ConfigPng;
+                    settings && settings.enablePlotViewer ? CodeSnippets.ConfigSvg : CodeSnippets.ConfigPng;
```
IanMatthewHuff, line 333: Removed the !isResourceNativeNotebook as Native notebooks can now use the setting to opt into getting the SVGs if they want. 

This code also doesn't look to have worked before. The check for isNativeNotebook was here, but not in the kernel.ts version. So this init would run first and set to [png] then the init in the native kernel.ts code would run and set to [png, svg] so native notebooks were always getting the SVGs unless the setting was off.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -436,33 +433,6 @@ export class Kernel implements IKernel {
             });
         }
     }
-    private async initializeMatplotlib(): Promise<void> {
```
IanMatthewHuff, line 439: This init code was running in two places. Also, the code was similar but different enough that it was actually getting different results from the two different calls so the results were conficting in the notebook. Locking on the notebook version for now as I believe that is correct for as long as we are still using INotebook, seems related to the refresh discussion as well.

rchiodo (approved):

package-lock.json
```diff
@@ -13548,11 +13548,22 @@
             }
         },
         "image-size": {
-            "version": "0.5.5",
-            "resolved": "https://registry.npmjs.org/image-size/-/image-size-0.5.5.tgz",
-            "integrity": "sha1-Cd/Uq50g4p6xw+gLiZA3jfnjy5w=",
-            "dev": true,
-            "optional": true
+            "version": "1.0.0",
```
DonJayamanne, line 13551: Any idea why this was updated, when package.json isn't updated?
Or did you forget to use `npm ci` instead of `npm i/npm install`

package.json
```diff
@@ -2095,6 +2095,7 @@
         "glob": "^7.1.2",
         "hash.js": "^1.1.7",
         "iconv-lite": "^0.4.21",
+        "image-size": "^1.0.0",
```
rchiodo, line 2098: @DonJayamanne image-size was added here.

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: > n the notebook. Locking on the notebook version for now as I believe that is correct for as long as we are still us

Please can we move the initialization to `Kernel`, as the plan is to move all of INotebook initialization into this.

DonJayamanne (changes requested):
Please move initialization into Kernel

DavidKutu (approved):

src/client/datascience/jupyter/jupyterNotebook.ts
```diff
@@ -287,93 +283,6 @@ export class JupyterNotebookBase implements INotebook {
         return this.session ? this.session.waitForIdle(timeoutMs) : Promise.resolve();
     }
```
IanMatthewHuff, line 285: Intialize code (working dir / startup commands / plot init) now all moved from Notebook into Kernel.

src/client/datascience/jupyter/kernels/kernel.ts
IanMatthewHuff: @DonJayamanne moved the init code (working dir / startup commands / plot init) into kernel.

IanMatthewHuff:
@DavidKutu @rchiodo Tagged for a re-review since I made a big bigger change to move init code only into Kernel versus Notebook.

src/client/datascience/jupyter/jupyterNotebook.ts
IanMatthewHuff: Gave a quick peek with notebooks and IW. Startup command, new working directory, and plots all looked good after a notebook start as well as after a kernel restart.

rchiodo (approved):

src/client/datascience/jupyter/jupyterNotebook.ts
IanMatthewHuff: @DonJayamanne could you take a re-look? It's still blocked on the change requested.

DonJayamanne (approved):

DavidKutu (approved):

PR #7150, opened 2021-08-16
Propagate kernel restart to all parts of extension (#7129)

DonJayamanne:


src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -187,6 +188,7 @@ export class KernelExecution implements IDisposable {
         // Restart the active execution
         await (this._restartPromise ? this._restartPromise : (this._restartPromise = this.restartExecution(notebook)));
 
+        (notebook as JupyterNotebookBase).fireRestart();
```
rchiodo, line 191: This seems like the wrong spot to me.

restartExecution seems better because:

- It would only fire when the kernel actually restarts (instead of every time the button is hit)
- It would work for the temporary code that joyce added for the interactive window too.

rchiodo (changes requested):
⏲️

rchiodo (approved):

DavidKutu (approved):

PR #7205, opened 2021-08-19
Remove redundant call

DonJayamanne:
Found that we were executing the same code twice (came across this when adding some logging to identify cause of a flaky test).

### DIFF
--- a/src/client/datascience/jupyter/kernels/kernel.ts
+++ b/src/client/datascience/jupyter/kernels/kernel.ts
@@ -364,10 +364,6 @@ export class Kernel implements IKernel {
             traceInfoIf(isCI, 'Step I');
             await this.disableJedi();
             traceInfoIf(isCI, 'Step J');
-            if (this.resourceUri) {
-                await this.notebook.setLaunchingFile(this.resourceUri.fsPath);
-                traceInfoIf(isCI, 'Step K');
-            }
 
             // For Python notebook initialize matplotlib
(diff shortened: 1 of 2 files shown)


src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -364,10 +364,6 @@ export class Kernel implements IKernel {
             traceInfoIf(isCI, 'Step I');
             await this.disableJedi();
             traceInfoIf(isCI, 'Step J');
-            if (this.resourceUri) {
-                await this.notebook.setLaunchingFile(this.resourceUri.fsPath);
```
DonJayamanne, line 368: In lin 361, we're already setting the working directory (code is identical), hence no need to call notebook API again.

rchiodo: Except on line 361, the resourceUri.fsPath isn't passed, so the logic will be different.

rchiodo (changes requested):
:timer:

DavidKutu (approved):

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: Good one, fixed.

IanMatthewHuff (approved):

rchiodo (approved):

PR #7219, opened 2021-08-21
Log debug adapter traffic

### LINKED ISSUE #6468: Log Debug Adapter traffic
This will make it easier to test performance and other issues. Like this one https://github.com/microsoft/vscode/issues/127720.
Put it on the Jupyter output or on its own output.

roblourens:
Add a "verbose" logging level
Fix #6468

### DIFF
--- a/src/client/common/types.ts
+++ b/src/client/common/types.ts
@@ -186,7 +186,7 @@ export interface IWatchableJupyterSettings extends IJupyterSettings {
     readonly onDidChange: Event<void>;
 }
 
-export type LoggingLevelSettingType = 'off' | 'error' | 'warn' | 'info' | 'debug';
+export type LoggingLevelSettingType = 'off' | 'error' | 'warn' | 'info' | 'debug' | 'verbose';
 
 export interface ILoggingSettings {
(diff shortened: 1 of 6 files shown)


DavidKutu (approved):

src/client/common/types.ts
```diff
@@ -186,7 +186,7 @@ export interface IWatchableJupyterSettings extends IJupyterSettings {
     readonly onDidChange: Event<void>;
 }
 
-export type LoggingLevelSettingType = 'off' | 'error' | 'warn' | 'info' | 'debug';
+export type LoggingLevelSettingType = 'off' | 'error' | 'warn' | 'info' | 'debug' | 'verbose';
```
DonJayamanne, line 189: I personally think we should clean all of this, I don't know if anyone uses them. Feels rather redundant to have `warn|info|debug`, I think `off|error|verbose` are sufficient.

DonJayamanne (approved):

src/client/common/types.ts
DonJayamanne: Something for debt week

Issue #7218, opened 2021-08-20
Bug cleanup in Run by Line


DavidKutu:
High priority

- [x] context keys are sometimes stuck in the "running" state
  - Seems that we run the cell but don't hit the breakpoint, and don't reset the keys. See logs below
- [x] RBL button doesn't show up sometimes
- [x] Sometimes I am blocked by an error with starting the kernel. I get a modal error about a missing debug adapter, and an infinite "Starting run by line" notification
- [x] When stepping over a call to another file, the other file is still opened
   - don't send stopped events to vscode when we're outside the notebook. Check stacktrace request to see if we get response right away.

Lower priority

- [x] show the "starting run by line" notification only the first time
- [x] set global context keys based on active editor, so I can run two RBL sessions in two notebooks
- [x] Context keys are reset when any other debug session ends
   - debug.onDidTerminateDebugSession, check for our own sessions
- [x] setting for optionally showing variable explorer

roblourens:
[worked.log](https://github.com/microsoft/vscode-jupyter/files/7024276/worked.log)
[didntwork.log](https://github.com/microsoft/vscode-jupyter/files/7024277/didntwork.log)

DavidKutu:
add setting for bringing up the variable explorer

DavidKutu:
don't send stopped events to vscode when we're outside the notebook. Check stacktrace request to see if we get response right away.

PR #7226, opened 2021-08-23
RBL bug fixes

roblourens:
Here are a few fixes for assorted RBL bugs, described in #7218. I can split these into multiple PRs if you want, but they are small changes. You can see them in the individual commits.

- Update context keys on startup so I don't have to close and reopen the notebook to get the RBL button to show up
- Forward all debug events to client, some were being lost
- Avoid sending messages to the wrong place (like requests to client)
- Don't block the configurationDone event on execution, the target didn't receive it
- Only show the "Starting" popup when RBL will be slow to start
- Don't leak listeners from the debug adapter, which led to strange behavior
- Don't delete the temp directory containing ipykernel cell contents, this spammed errors and might work around a bug where RBL gets stuck and won't start

### DIFF
--- a/src/client/debugger/jupyter/kernelDebugAdapter.ts
+++ b/src/client/debugger/jupyter/kernelDebugAdapter.ts
@@ -25,7 +25,7 @@ import * as path from 'path';
 import { IJupyterSession } from '../../datascience/types';
 import { KernelMessage } from '@jupyterlab/services';
 import { ICommandManager } from '../../common/application/types';
-import { traceError } from '../../common/logger';
+import { traceError, traceVerbose } from '../../common/logger';
 import { IFileSystem } from '../../common/platform/types';
 import { IKernelDebugAdapter } from '../types';
 import { IDisposable } from '../../common/types';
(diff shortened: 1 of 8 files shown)


src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -350,26 +300,24 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter, ID
...
             });
 
             if (control) {
                 control.onReply = (msg) => this.controlCallback(msg.content as DebugProtocol.ProtocolMessage);
-                control.onIOPub = (msg) => this.controlCallback(msg.content as DebugProtocol.ProtocolMessage);
             }
         } else if (message.type === 'response') {
             // responses of reverse requests
-            const response = debugResponse(message as DebugProtocol.Response, this.jupyterSession.sessionId);
+            const response = message as DebugProtocol.Response;
```
DonJayamanne, line 316: Much cleaner

DonJayamanne (approved):

PR #7228, opened 2021-08-23
Ensure the build numbers are updated correctly

DonJayamanne:


gulpfile.js
```diff
@@ -198,10 +198,10 @@ async function updateBuildNumber(args) {
             versionParts.length > 2 ? versionParts[2].replace(/(\d+)/, args.buildNumber) : args.buildNumber;
         const newVersion =
             versionParts.length > 1
-                ? `${versionParts[0]}.${versionParts[1]}.${versionParts[0].substring(
```
DonJayamanne, line 201: This was the problem we were using [0].
Also, using the last n characters is incorrect, we should use the first n characters, ignoring the last 3 characters.

For instance a few weeks ago we had `1149009458` and this week we have `1158009458`
Based on the old algorithm, the last weeks build is later than this weeks build.

gulpfile.js
```diff
@@ -198,10 +198,10 @@ async function updateBuildNumber(args) {
...
+                ? `${versionParts[0]}.${versionParts[1]}.${versionParts[2].substring(
                       0,
                       3
-                  )}${buildNumberPortion.substring(3)}`
+                  )}${buildNumberPortion.substring(0, buildNumberPortion.length - 3)}`
```
joyceerhl, line 204: This build number is generated from the GITHUB_RUN_ID--is it possible that this generates collisions (i.e. the last 3 digits of the ID are incremented but the first n-3 are not)? 

https://github.com/microsoft/vscode-jupyter/blob/b83bac0a6101797a45c393163bb1c546c823b5bb/.github/actions/build-vsix/action.yml#L15

DonJayamanne: > ble that this generates collisions (i.e. the last 3 digits of the ID are incremented between runs but the first n-3 are not)?

Not for biulds that happen every night/daily.

joyceerhl (approved):

PR #7234, opened 2021-08-23
Fix notebook export for already-open notebook documents

### LINKED ISSUE #7233: Export notebook doesn't put up the export quickpick for opening an existing notebook
1. Open a notebook
2. Reload VS Code. Notebook is still open
3. Click on the notebook toolbar export button
4. 🐛 export does nothing, quickpick isn't shown
5. Close and reopen notebook, then export works

Root cause is the notebookEditorProvider wasn't aware of notebookDocuments that were already open before the extension activated, so running the command does nothing because activeEditor is undefined and the editor list is empty

https://github.com/microsoft/vscode-jupyter/blob/b019cc9ce852a2de38df45f7db1beba3e503e6cb/src/client/datascience/commands/exportCommands.ts#L66-L70

https://github.com/microsoft/vscode-jupyter/blob/b019cc9ce852a2de38df45f7db1beba3e503e6cb/src/client/datascience/commands/exportCommands.ts#L89-L94

joyceerhl:
Fix #7233

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "jupyter",
     "displayName": "Jupyter",
-    "version": "2021.8.200",
+    "version": "2021.8.203",
     "description": "Jupyter notebook support, interactive programming and computing that supports Intellisense, debugging and more.",
     "publisher": "ms-toolsai",
     "enableProposedApi": true,
--- a/package.json
+++ b/package.json
@@ -1411,7 +1411,8 @@
                         "error",
                         "warn",
                         "info",
-                        "debug"
+                        "debug",
+                        "verbose"
                     ],
                     "description": "The logging level the extension logs at.",
                     "scope": "machine"
--- a/package.json
+++ b/package.json
@@ -722,6 +722,7 @@
                 "title": "%DataScience.exportDialogTitle%",
                 "shortTitle": "%DataScience.exportAsNotebook.shorttitle%",
                 "icon": "$(save-as)",
+                "enablement": "notebookType == interactive",
                 "category": "Jupyter"
             },
             {
@@ -731,6 +732,7 @@
                     "light": "resources/light/export_to_python.svg",
                     "dark": "resources/dark/export_to_python.svg"
                 },
+                "enablement": "notebookType == interactive",
                 "category": "Jupyter"
             },
             {
--- a/src/client/common/application/commands.ts
+++ b/src/client/common/application/commands.ts
@@ -143,7 +143,7 @@ export interface ICommandNameArgumentTypeMapping extends ICommandNameWithoutArgu
     [DSCommands.ExportToHTML]: [string | undefined, Uri | undefined, string | undefined, PythonEnvironment | undefined];
     [DSCommands.ExportToPDF]: [string | undefined, Uri | undefined, string | undefined, PythonEnvironment | undefined];
     [DSCommands.Export]: [string | undefined, Uri | undefined, string | undefined, PythonEnvironment | undefined];
-    [DSCommands.NativeNotebookExport]: [Uri | undefined];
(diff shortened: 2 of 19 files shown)


package.json
```diff
@@ -731,6 +732,7 @@
                     "light": "resources/light/export_to_python.svg",
                     "dark": "resources/dark/export_to_python.svg"
                 },
+                "enablement": "notebookType == interactive",
```
joyceerhl, line 735: These commands were showing up in the command palette when they shouldn't

src/client/common/application/commands.ts
```diff
@@ -143,7 +143,7 @@ export interface ICommandNameArgumentTypeMapping extends ICommandNameWithoutArgu
     [DSCommands.ExportToHTML]: [string | undefined, Uri | undefined, string | undefined, PythonEnvironment | undefined];
     [DSCommands.ExportToPDF]: [string | undefined, Uri | undefined, string | undefined, PythonEnvironment | undefined];
     [DSCommands.Export]: [string | undefined, Uri | undefined, string | undefined, PythonEnvironment | undefined];
-    [DSCommands.NativeNotebookExport]: [Uri | undefined];
+    [DSCommands.NativeNotebookExport]: [Uri | undefined | { notebookEditor: { notebookUri: Uri } }];
```
joyceerhl, line 146: The context that's passed from the notebookToolbar is not a simple Uri. Not sure if this was a recent change in VS Code. There might be other command handlers that are impacted by this.

src/client/datascience/notebook/notebookEditorProvider.ts
```diff
@@ -76,6 +76,10 @@ export class NotebookEditorProvider implements INotebookEditorProvider {
         @inject(IExtensions) private readonly extensions: IExtensions
     ) {
         disposables.push(this);
+        // Register notebook documents that are already open e.g. after a workspace reload
+        for (const notebookDocument of this.vscodeNotebook.notebookDocuments) {
+            void this.onDidOpenNotebookDocument(notebookDocument);
+        }
```
joyceerhl, line 82: This is the actual fix--notebookEditorProvider wasn't aware of notebookDocuments that were already loaded.

src/client/datascience/commands/exportCommands.ts
```diff
@@ -63,9 +63,10 @@ export class ExportCommands implements IDisposable {
         this.disposables.push(disposable);
     }
 
-    private async nativeNotebookExport(uri?: Uri) {
-        const editor = uri
-            ? this.notebookProvider.editors.find((item) => this.fs.arePathsSame(item.file, uri))
+    private async nativeNotebookExport(uri?: Uri | { notebookEditor: { notebookUri: Uri } }) {
+        const notebookUri = uri !== undefined && 'notebookEditor' in uri ? uri.notebookEditor.notebookUri : uri;
```
DonJayamanne, line 67:
```suggestion
        const notebookUri = isUri(uri) ? uri : uri.notebookEditor.notebookUri;
```

DonJayamanne: You can use the helper `isUri` instead, you'd need to import that into this file.

DonJayamanne (approved):

DonJayamanne:
@joyceerhl please could you update your code with main, this PR looks funky

src/client/datascience/notebook/notebookEditorProvider.ts
DonJayamanne: So we've always had this issue! Whoa

joyceerhl: Actually, I debugged the release branch in VS Code 1.59.1, and it looks like VS Code has stopped firing `onDidOpenNotebookDocument` for already-open NotebookDocuments in Insiders, so previously this code wasn't even necessary. i.e. this was a behavior change in VS Code, not the Jupyter extension.

Issue #7156, opened 2021-08-16
Running code from the interactive window input box causes ipython-input temp file to open for debugging


joyceerhl:
Originally reported by @greazer 

1. Open a Python file with a #%% cell in it    
2. Shift+Enter to execute
3. Run code from the interactive window input box <-- this seems to be messing with sourcemaps or similar
4. Debug cell
5. 🐛 temp file containing our injected source code appears

This is likely the execution count getting messed up because we're not accounting for the cell submitted from the interactive window input box

PR #7245, opened 2021-08-23
Ensure executionCount is incremented for cells submitted from interactive window input box

joyceerhl:
Fix https://github.com/microsoft/vscode-jupyter/issues/7156

We were previously only incrementing the executionCounts for interactive cells submitted from a Python file. Need to do this for cells submitted from the interactive window input box too. No news entry because this was caused by https://github.com/microsoft/vscode-jupyter/commit/2b3fdb7ef5eae0bae50adf1b126060b09b857278, which was checked in between stable releases

### DIFF
--- a/src/client/datascience/jupyter/kernels/kernel.ts
+++ b/src/client/datascience/jupyter/kernels/kernel.ts
@@ -35,6 +35,7 @@ import { CodeSnippets, Identifiers, Telemetry } from '../../constants';
 import { sendKernelTelemetryEvent, trackKernelResourceInformation } from '../../telemetry/telemetry';
 import { getNotebookMetadata } from '../../notebook/helpers/helpers';
 import {
+    ICellHashProvider,
     IDataScienceErrorHandler,
     IJupyterServerUriStorage,
     INotebook,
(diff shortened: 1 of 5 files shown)


src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -136,6 +138,9 @@ export class Kernel implements IKernel {
     public async executeCell(cell: NotebookCell): Promise<void> {
         const stopWatch = new StopWatch();
         const notebookPromise = this.startNotebook({ disableUI: false, document: cell.notebook });
+        if (cell.notebook.notebookType === InteractiveWindowView) {
```
DonJayamanne, line 141: This is a very weird pr, no part of the code makes me think that any of these changes will update the execution count 
My only guess is that when we add a cell using the cell hash provided it's a real Block of code, else it's treated as silent and we end up with no execution count.

Is that right?
If that's the case then cell hash provided does a whole lot more than it seems to.
No need to change this, but I would have never guessed, and to me is a pretty weird way to wire things up.

DonJayamanne (approved):
Eew

src/client/datascience/jupyter/kernels/kernel.ts
joyceerhl: Yeah the executionCount is state that's maintained inside the cellHashProvider. I agree that it's pretty opaque. FWIW, this used to be done in an INotebookExecutionLogger, I think my change to explicitly call addCellHash made it more explicit that the cellHashProvider is getting called, so the code now scans funny.

PR #7344, opened 2021-08-30
Remove execute silently from notebooks

### LINKED ISSUE #7267: Bogus paths displayed in interactive window errors messages
Run a cell in the interactive window and ensure you have some errors in the code.
The stack trace will display some werid python file, instead of displaying the original python file.

This would apply to any error message display in interactive window (run time errors, or even when interrupting - actually they are one and the same).

Used to work in webview interactive window.

## Webview
Screen Shot 2021-08-24 at 11 36 00

## Native

Screen Shot 2021-08-24 at 11 36 08

DonJayamanne:
Part of https://github.com/microsoft/vscode-jupyter/issues/7267

Part 1 of fix #7267
(this just removes executeSilently completely & a lot of ICell stuff)


rchiodo (approved):
Seems okay, just not sure why? Not sure how removing the 'silent' flag fixes the bug?

Issue #7071, opened 2021-08-11
Native interactive window should support kernel-based hover provider


joyceerhl:
This doesn't currently work because the hover provider is a NotebookExecutionLogger.

JacksonKearl:
I don't seem to get any hovers in interactive window's input, is this expected?

PR #7348, opened 2021-08-30
Fixes to cell hash provider related code

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/7347, https://github.com/microsoft/vscode-jupyter/issues/7071, https://github.com/microsoft/vscode-jupyter/issues/7267, https://github.com/microsoft/vscode-jupyter/issues/7356

Will need to add tests as part of debt week

* Remove loggers, don't need them anymore.
* Removed code where we were exposing this as an API, we also had comments not to use this (was only used by Gather, now that gather has been updated to use Native VSC API, we don't need this anymore)

### DIFF
--- a/src/client/datascience/interactive-window/nativeInteractiveWindowCommandListener.ts
+++ b/src/client/datascience/interactive-window/nativeInteractiveWindowCommandListener.ts
@@ -478,7 +478,6 @@ export class NativeInteractiveWindowCommandListener implements IDataScienceComma
             // Scroll to cell in the one that has the cell. We need this so
             // we don't activate all of them.
             // eslint-disable-next-line @typescript-eslint/prefer-for-of
-            possibles.forEach((item) => item.scrollToCell(id));
             for (let i = 0; i < possibles.length; i += 1) {
                 if (await possibles[i].hasCell(id)) {
                     possibles[i].scrollToCell(id);
--- a/src/client/datascience/interactive-window/nativeInteractiveWindowCommandListener.ts
+++ b/src/client/datascience/interactive-window/nativeInteractiveWindowCommandListener.ts
@@ -478,6 +478,7 @@ export class NativeInteractiveWindowCommandListener implements IDataScienceComma
             // Scroll to cell in the one that has the cell. We need this so
             // we don't activate all of them.
             // eslint-disable-next-line @typescript-eslint/prefer-for-of
+            possibles.forEach((item) => item.scrollToCell(id));
             for (let i = 0; i < possibles.length; i += 1) {
                 if (await possibles[i].hasCell(id)) {
                     possibles[i].scrollToCell(id);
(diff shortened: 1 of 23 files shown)


src/client/datascience/interactive-window/nativeInteractiveWindowCommandListener.ts
```diff
@@ -478,6 +478,7 @@ export class NativeInteractiveWindowCommandListener implements IDataScienceComma
             // Scroll to cell in the one that has the cell. We need this so
             // we don't activate all of them.
             // eslint-disable-next-line @typescript-eslint/prefer-for-of
+            possibles.forEach((item) => item.scrollToCell(id));
```
rchiodo, line 481: This seems like it would scroll to multiple cells and then go back and scroll to the first one found? Did you intend to leave this here?

src/client/datascience/notebookExtensibility.ts
```diff
@@ -1,8 +1,4 @@
-import { injectable } from 'inversify';
-import { Event, EventEmitter, NotebookCell, Uri } from 'vscode';
-import { noop } from '../common/utils/misc';
-import { ICell, INotebookExecutionLogger, INotebookExtensibility } from './types';
-import { translateCellToNative } from './utils';
+import { NotebookCell, Uri } from 'vscode';
```
rchiodo, line 1: Maybe rename this file? Or just move types somewhere else?

src/client/datascience/interactive-window/nativeInteractiveWindowCommandListener.ts
DonJayamanne: oops, good catch. was meant to be removed.

src/client/datascience/types.ts
```diff
@@ -247,13 +244,7 @@ export interface IHoverProvider extends HoverProvider {}
...
 
 export const INotebookExecutionLogger = Symbol('INotebookExecutionLogger');
-export interface INotebookExecutionLogger extends IDisposable {
-    preExecute(cell: ICell, silent: boolean): Promise<void>;
-    postExecute(cell: ICell, silent: boolean, language: string, resource: Uri): Promise<void>;
-    nativePostExecute?(cell: NotebookCell): Promise<void>;
-    onKernelRestarted(resource: Uri): void;
-    preHandleIOPub?(msg: KernelMessage.IIOPubMessage): KernelMessage.IIOPubMessage;
-}
+export interface INotebookExecutionLogger extends IDisposable {}
```
rchiodo, line 247: Can't this just be eliminated now?

rchiodo (approved):

DavidKutu (approved):

src/client/datascience/notebookExtensibility.ts
DonJayamanne: I'd like to leave that for debt, i just want to get this PR over and done with and get it into the release branch.
Hope thats ok with you.

src/client/datascience/types.ts
DonJayamanne: I wanted to leave that for debt, there are places where this is used.
Was too much work to try and remove its usages (I only remvoed the methods so that we don't accidentally use them, that's kinda what happened, the methods exist, and we're assuming its all ok, but since nothing is wired up, it will not work).

DonJayamanne:
> Can you add some unit tests for this stuff?

Want to do that separately, later today (debt)

Issue #7385, opened 2021-09-01
Remote Notebook kernels do not get loaded unless you open another notebook


DonJayamanne:
Open vscode, and open a notebook
Connect to remote
Reload VS Code
Note: Notebook is opened
Wait for 10 minutes, & remote kernels will not appear
Next, open another notebook and the remote kernels will now appear

DonJayamanne:
Fixed

PR #7386, opened 2021-09-01
Refresh list of remote kernels if a notebook is already open

DonJayamanne:
For #7385

### DIFF
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -112,9 +112,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         // Sign up for document either opening or closing
         this.notebook.onDidOpenNotebookDocument(this.onDidOpenNotebookDocument, this, this.disposables);
         // If the extension activates after installing Jupyter extension, then ensure we load controllers right now.
-        if (this.isLocalLaunch) {
-            this.notebook.notebookDocuments.forEach((notebook) => this.onDidOpenNotebookDocument(notebook).catch(noop));
-        }
+        this.notebook.notebookDocuments.forEach((notebook) => this.onDidOpenNotebookDocument(notebook).catch(noop));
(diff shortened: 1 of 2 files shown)


joyceerhl:
VS Code used to fire onDidOpenNotebookDocument for already-open notebook documents. This behavior changed in this last iteration and is probably the reason for this problem, I also fixed a problem with export not working that Rong found which was caused by this. I think we should check with VS Code re: whether this was an intentional change

rchiodo (approved):

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -112,9 +112,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         // Sign up for document either opening or closing
         this.notebook.onDidOpenNotebookDocument(this.onDidOpenNotebookDocument, this, this.disposables);
         // If the extension activates after installing Jupyter extension, then ensure we load controllers right now.
-        if (this.isLocalLaunch) {
-            this.notebook.notebookDocuments.forEach((notebook) => this.onDidOpenNotebookDocument(notebook).catch(noop));
-        }
+        this.notebook.notebookDocuments.forEach((notebook) => this.onDidOpenNotebookDocument(notebook).catch(noop));
```
IanMatthewHuff, line 115: Change makes sense, do we know why it was gated on isLocalLaunch in the first place? That doesn't seem relevant to me. Just making sure we're not regressing on some reason this was only local before.

IanMatthewHuff (approved):

DavidKutu (approved):

src/client/datascience/notebook/notebookControllerManager.ts
DonJayamanne: Back then I wasn't sure how remote worked, and didn't want to break remote.
I think I fixed this at the last minute before a release, hence didn't want to make unnecessary changes

PR #7401, opened 2021-09-02
RBL tests

DavidKutu:
For #6243

### DIFF
--- a/src/test/datascience/debugger.vscode.test.ts
+++ b/src/test/datascience/debugger.vscode.test.ts
@@ -1,0 +1,160 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+'use strict';
+import * as sinon from 'sinon';
+import { DebugProtocol } from 'vscode-debugprotocol';
+import { ICommandManager, IVSCodeNotebook } from '../../client/common/application/types';
+import { IDisposable } from '../../client/common/types';
+import { Commands } from '../../client/datascience/constants';
+import { IVariableViewProvider } from '../../client/datascience/variablesView/types';
+import { IExtensionTestApi } from '../common';
+import { initialize, IS_REMOTE_NATIVE_TEST, IS_WEBVIEW_BUILD_SKIPPED } from '../initialize';
+import {
+    canRunNotebookTests,
+    closeNotebooks,
+    closeNotebooksAndCleanUpAfterTests,
+    createEmptyPythonNotebook,
+    insertCodeCell,
+    prewarmNotebooks,
+    workAroundVSCodeNotebookStartPages,
+    waitForEvent,
+    getCellOutputs
+} from './notebook/helper';
+import { verifyViewVariables } from './variableView/variableViewHelpers';
+import { ITestVariableViewProvider } from './variableView/variableViewTestInterfaces';
+import { traceInfo } from '../../client/common/logger';
+import { sleep } from '../core';
+import { IDebuggingManager } from '../../client/debugger/types';
+import { assert } from 'chai';
+import { DebugSession } from 'vscode';
+import { KernelDebugAdapter } from '../../client/debugger/jupyter/kernelDebugAdapter';
+import { ITestWebviewHost } from './testInterfaces';
+
+suite('DataScience - Debugging', function () {
+    let api: IExtensionTestApi;
+    const disposables: IDisposable[] = [];
+    let commandManager: ICommandManager;
+    let variableViewProvider: ITestVariableViewProvider;
+    let vscodeNotebook: IVSCodeNotebook;
+    let debuggingManager: IDebuggingManager;
+    this.timeout(120_000);
+    suiteSetup(async function () {
+        traceInfo(`Start Test Suite`);
+        this.timeout(120_000);
+        api = await initialize();
+
+        // We need to have webviews built to run this, so skip if we don't have them
+        if (IS_WEBVIEW_BUILD_SKIPPED) {
+            console.log('Debugging tests require webview build to be enabled (for the variable view)');
+            return this.skip();
+        }
+
+        // Don't run if we can't use the native notebook interface
+        if (IS_REMOTE_NATIVE_TEST || !(await canRunNotebookTests())) {
+            return this.skip();
+        }
+        await workAroundVSCodeNotebookStartPages();
+        await closeNotebooksAndCleanUpAfterTests(disposables);
+        await sleep(5_000);
+        await prewarmNotebooks();
+        sinon.restore();
+        commandManager = api.serviceContainer.get<ICommandManager>(ICommandManager);
+        const coreVariableViewProvider = api.serviceContainer.get<IVariableViewProvider>(IVariableViewProvider);
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        variableViewProvider = (coreVariableViewProvider as any) as ITestVariableViewProvider; // Cast to expose the test interfaces
+        debuggingManager = api.serviceContainer.get<IDebuggingManager>(IDebuggingManager);
+        vscodeNotebook = api.serviceContainer.get<IVSCodeNotebook>(IVSCodeNotebook);
+        traceInfo(`Start Test Suite (completed)`);
+    });
+    setup(async function () {
+        traceInfo(`Start Test ${this.currentTest?.title}`);
+        sinon.restore();
+
+        // Create an editor to use for our tests
+        await createEmptyPythonNotebook(disposables);
+        traceInfo(`Start Test (completed) ${this.currentTest?.title}`);
+    });
+    teardown(async function () {
+        traceInfo(`Ended Test ${this.currentTest?.title}`);
+        await closeNotebooks(disposables);
+        await closeNotebooksAndCleanUpAfterTests(disposables);
+        traceInfo(`Ended Test (completed) ${this.currentTest?.title}`);
+    });
+
+    // Cleanup after suite is finished
+    suiteTeardown(() => closeNotebooksAndCleanUpAfterTests(disposables));
+
+    test('Run by Line - Full Workflow', async function () {
+        // set up
+        await insertCodeCell('a=1\na', { index: 0 });
+        const doc = vscodeNotebook.activeNotebookEditor?.document!;
+        const cell = doc.getCells()![0]!;
+
+        // Start Run by Line
+        await commandManager.executeCommand(Commands.RunByLine, cell);
+
+        // Check that a debugging session is created
+        const session = debuggingManager.getDebugSession(doc);
+        assert.isOk<DebugSession | undefined>(session, 'Session not started');
+
+        // Check that the debug adapter is created
+        const debugAdapter = debuggingManager.getDebugAdapter(doc);
+        assert.isOk<KernelDebugAdapter | undefined>(debugAdapter, 'DebugAdapter not started');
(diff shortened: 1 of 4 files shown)


src/test/datascience/notebook/helper.ts
```diff
@@ -763,3 +767,37 @@ export async function hijackPrompt(
...
+    });
+}
+
+export async function waitForEvent<T>(
+    eventType: string,
+    debugAdapter: KernelDebugAdapter,
+    timeout = env.uiKind === UIKind.Desktop ? 5000 : 15000
+): Promise<T> {
+    setTimeout(() => {
+        return;
```
rchiodo, line 795: I don't think this timeout will do anything. It should be throwing an exception or rejecting a promise.

DavidKutu: I'll throw an exception

src/test/datascience/notebook/helper.ts
```diff
@@ -763,3 +767,37 @@ export async function hijackPrompt(
...
+    });
+}
+
+export async function waitForEvent<T>(
+    eventType: string,
+    debugAdapter: KernelDebugAdapter,
+    timeout = env.uiKind === UIKind.Desktop ? 5000 : 15000
+): Promise<T> {
+    setTimeout(() => {
+        throw new Error(`Debugging Timeout: ${eventType} message took too long to arrive`);
```
roblourens, line 795: You might want to clear the timeout to avoid throwing in the successful case. It might show up as an unhandled error in a test log.

roblourens: Also, you should return a rejected promise from waitForEvent when this triggers to make sure you get a test failure with the right error message.

src/test/datascience/debugger.vscode.test.ts
```diff
@@ -0,0 +1,160 @@
...
+            return this.skip();
+        }
+
+        // Don't run if we can't use the native notebook interface
+        if (IS_REMOTE_NATIVE_TEST || !(await canRunNotebookTests())) {
+            return this.skip();
+        }
+        await workAroundVSCodeNotebookStartPages();
+        await closeNotebooksAndCleanUpAfterTests(disposables);
+        await sleep(5_000);
```
roblourens, line 58: Why a sleep?

src/test/datascience/debugger.vscode.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        assert.isTrue(stack.stackFrames.length > 0, 'has frames');
+        assert.equal(stack.stackFrames[0].source?.path!, cell.document.uri.toString(), 'Stopped at the worng path');
+
+        // continue to the next line
+        await commandManager.executeCommand(Commands.RunByLineContinue, cell);
+
+        // Wait for the stoped event
+        msg = await waitForEvent<DebugProtocol.StoppedEvent>('stopped', debugAdapter!);
+        // Wait for variable view to be updated
+        await sleep(1000);
```
roblourens, line 120: Is there an event on the variable view that you can listen to?

src/test/datascience/debugger.vscode.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        const coreVariableView = await variableViewProvider.activeVariableView;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const variableView = (coreVariableView as any) as ITestWebviewHost;
+        const htmlResult = await variableView?.getHTMLById('variable-view-main-panel');
+        const expectedVariables = [{ name: 'a', type: 'int', length: '', value: '1' }];
+        verifyViewVariables(expectedVariables, htmlResult);
+
+        // Stop run by line and check that the cell ran
+        await commandManager.executeCommand(Commands.RunByLineStop);
+        await sleep(1000);
```
roblourens, line 132: You can wait on the vscode event onDidUpdateCellOutputs

src/test/datascience/debugger.vscode.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        // Check that the debug adapter is created
+        const debugAdapter = debuggingManager.getDebugAdapter(doc);
+        assert.isOk<KernelDebugAdapter | undefined>(debugAdapter, 'DebugAdapter not started');
+
+        // Wait for the stoped event
+        await waitForEvent<DebugProtocol.StoppedEvent>('stopped', debugAdapter!);
+
+        // Interrupt kernel and check that the cell didn't finish running
+        await commandManager.executeCommand(Commands.InterruptKernel, { notebookEditor: { notebookUri: doc.uri } });
+        assert.isTrue(getCellOutputs(cell).includes('KeyboardInterrupt'));
```
roblourens, line 158: You might need to wait on the cell output to be updated?

roblourens: Or, it might always be done by the time the Interrupt command is done, if that code awaits on the output updates. You might not want to depend on that though. You can wait on the onDidUpdateCellOutputs event here too

src/test/datascience/notebook/helper.ts
rchiodo: Yeah this exception isn't going to end up anywhere. You could just the already existing WaitForCondition

```suggestion
    return WaitForCondition(async () => {
       const msg: any = await asPromise(debugAdapter.onDidSendMessage);
       return msg.event == eventType;
    },
    timeout);
```

DavidKutu: That wouldn't get the message back

rchiodo: Yeah sorry, the message is part of waitForCondition too. I just didn't type it in.

src/test/datascience/debugger.vscode.test.ts
DavidKutu: it was tricky, but yeah there is

src/test/datascience/debugger.vscode.test.ts
DavidKutu: where is this event? its not on the d.ts

src/test/datascience/debugger.vscode.test.ts
DavidKutu: to give time for setup, I'm not 100% sure why 5 seconds and not 1 for example. That is on all out tests.

DonJayamanne: I could only find one test that does this, not everywhere, I think we should remove this.

src/test/datascience/debugger.vscode.test.ts
DonJayamanne: You can use the existing api we have waitForCondition (please see other tests)

@roblourens Fyi That's just using a settimeout, not the best,  back when this was written the events wouldn't fire, however can update that later,

src/test/datascience/debugger.vscode.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    let vscodeNotebook: IVSCodeNotebook;
+    let debuggingManager: IDebuggingManager;
+    this.timeout(120_000);
+    suiteSetup(async function () {
+        traceInfo(`Start Test Suite`);
+        this.timeout(120_000);
+        api = await initialize();
+
+        // We need to have webviews built to run this, so skip if we don't have them
+        if (IS_WEBVIEW_BUILD_SKIPPED) {
```
DonJayamanne, line 49: You can move these above initialize

src/test/datascience/debugger.vscode.test.ts
DonJayamanne: You can use the existing api we have waitForCondition, see how are check for output in other tests

DonJayamanne (commented):
I'd like to see a test for the complex scenarios like debugger handling last lines, and last lines calling functions and the like.
I recall in a code review for a big fix I identified a scenario that ended up requiring a lot of changes, hence two tests for the original bug and another for that complex use case.

DonJayamanne (commented):
Would also like to see tests for step in (function in same cell)
Tests for  restart debugger after we finish the first time (can we restart multiple times), 
This is crucial for the bug that was identified and fixed (reattach)
Tests to ensure we don't lose kernel state when starting the debugger.

src/test/datascience/notebook/helper.ts
```diff
@@ -792,7 +792,9 @@ export async function waitForEvent<T>(
     timeout = env.uiKind === UIKind.Desktop ? 5000 : 15000
 ): Promise<T> {
     setTimeout(() => {
-        throw new Error(`Debugging Timeout: ${eventType} message took too long to arrive`);
+        clearTimeout(timeout);
+        console.log(`Test failing --- Debugging Timeout: ${eventType} message took too long to arrive`);
+        return Promise.reject();
```
roblourens, line 797: This is being returned from `setTimeout` and not going anywhere, you need a promise returned from `waitForEvent`.

rchiodo: I'm not sure what happend to the suggestion I made before? waitForCondition could be used here. It would make this function super simple.

src/test/datascience/notebook/helper.ts
roblourens: Not sure what you mean @rchiodo, how can waitForCondition return a value?

roblourens: We need waitForEvent to return the event, only way I can think of is to capture the message in the function closure and return it after awaiting waitForCondition

rchiodo: No you'd use waitForCondition in this function to wait for the message. Inside the function, just set a variable.

Like this:

```typescript
export async function waitForEvent<T>(
   eventType: string,
   debugAdapter: KernelDebugAdapter
   timeout = env.uiKind === UIKind.Desktop ? 5000 : 15000
): Promise<T> {
   let msg: T | undefined = undefined;
   await waitForCondition(async () => { 
    msg: any = await asPromise(debugAdapter.onDidSendMessage);
           msg = await asPromise(debugAdapter.onDidSendMessage);
	       return msg.event == eventType;
   }, `Debugging Timeout: ${eventType} message took too long to arrive`, timeout);
  return msg;
}
```

roblourens: That's what I was describing. But it's not the right approach because waitForCondition has a 10ms delay between loops so it's possible to miss an event (that might be possible even without the delay just because of async-ness) I implemented something simpler

src/test/datascience/notebook/helper.ts
```diff
@@ -793,3 +795,40 @@ export async function hijackPrompt(
...
+        });
+    });
+}
+
+export async function waitForDebugEvent<T>(
+    eventType: string,
+    debugAdapter: KernelDebugAdapter,
+    timeout = env.uiKind === UIKind.Desktop ? 5000 : 15000
+): Promise<T> {
+    return await asPromise(
```
rchiodo, line 825: Could you skip the await here?. Just return the ```asPromise<T>```

rchiodo (approved):

roblourens:
Stabilized and resolved feedback but am still adding more tests

roblourens:
> Tests to ensure we don't lose kernel state when starting the debugger.

What would be the right way to test that/

roblourens:
Any idea why I would get lots of `rejected promise not handled within 1 second: Error: Unknown webview handle:webviewview-jupyterviewvariables` when running these tests? The variables webview stuff seems to work

PR #7410, opened 2021-09-03
Fix interpreter tests

DonJayamanne:
For https://github.com/microsoft/vscode-jupyter/issues/7403

Looks like we have two interpreter paths that end up pointing to the exact same interpreter:
* /opt/hostedtoolcache/Python/3.8.11/x64/python
* /opt/hostedtoolcache/Python/3.8.11/x64/bin/python

This PR ensures we have a common way to compare interpreter paths and normalize them so that they are treated the same.


.github/actions/create-venv-for-tests/action.yml
```diff
@@ -21,13 +21,17 @@ runs:
 
         python -m venv .venvkernel
         source .venvkernel/bin/activate
+        python --version
+        python -c "import sys;print(sys.executable)"
```
DonJayamanne, line 25: Useful, this is how i figured some of the paths were different.

DonJayamanne: No harm in leaving this.

src/client/api/pythonApi.ts
```diff
@@ -345,7 +349,24 @@ export class InterpreterService implements IInterpreterService {
...
+                .then((items) => {
+                    const current = interpreterCacheForCI.get(resource?.toString() || '');
+                    const itemToStore = items;
+                    if (
+                        current &&
+                        (itemToStore === current || JSON.stringify(itemToStore) === JSON.stringify(current))
+                    ) {
+                        return;
+                    }
+                    interpreterCacheForCI.set(resource?.toString() || '', itemToStore);
```
DonJayamanne, line 364: Looks like Python extension stuff doesn't get logged, this was crucial to figuring out what was going on.
I've enabled this just for CI.
We can remove this, i'm not fussed. I thought it would be useful

DonJayamanne: I've used the logs from Python extension a lot in the past, hence feel this is useful, but NOT necessary.

Can remove if anyone things we don't need this,

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -451,6 +461,10 @@ export function findPreferredKernel(
             let subScore = 0;
 
             if (spec) {
+                traceInfoIf(isCI, `Checking kernel Spec ${JSON.stringify(spec)}`);
+                traceInfoIf(isCI, `isPythonKernelSpec(spec) = ${isPythonKernelSpec(spec)}`);
+                traceInfoIf(isCI, `isKernelRegisteredByUs(spec) = ${isKernelRegisteredByUs(spec)}`);
```
DonJayamanne, line 466: More logging just for CI, found that if the version numbers were 3.8.10 and 3.8.11, we'd end up returuning 3.8.10 as the correct one when 3.8.11 is whats exected.
Hence this logging was very useful.

Again, this only gets logged in CI

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -499,36 +516,21 @@ export function findPreferredKernel(
...
-                    // Search for a digit on the end of the name. It should match our major version
-                    const match = /\D+(\d+)/.exec(spec.name);
-                    if (match && match !== null && match.length > 0) {
-                        // See if the version number matches
-                        const nameVersion = parseInt(match[1][0], 10);
-                        if (nameVersion && nameVersion === preferredInterpreter.version.major) {
-                            score += 4;
-                        }
-                    }
-                }
```
DonJayamanne, line 528: Moved this code down to where we compare major versions (keep both together)

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -576,13 +603,43 @@ export function findPreferredKernel(
...
+                                isCI,
+                                'Increased score by +3 for matching raw version in notebook metadata interpreter'
+                            );
+                            score += 3;
+                        }
+                    } else if (
+                        metadata.interpreter?.version?.raw &&
+                        metadata.interpreter?.version?.raw === preferredInterpreter?.version?.raw
+                    ) {
+                        traceInfoIf(isCI, 'Increased score by +3 for matching raw version in preferred interpreter');
```
DonJayamanne, line 640: Cam across a scenario where 3.8.10 was picked over 3.8.11 (beacuse both have 3 and 3.8.10 was on the top of the list hence that got picked when both had the same score).

src/client/pythonEnvironments/info/interpreter.ts
```diff
@@ -76,5 +78,42 @@ export async function getInterpreterInfo(
...
+}
+
+/**
+ * Sometimes on CI, we have paths such as (this could happen on user machines as well)
+ *  - /opt/hostedtoolcache/Python/3.8.11/x64/python
+ *  - /opt/hostedtoolcache/Python/3.8.11/x64/bin/python
+ *  They are both the same.
+ * This function will take that into account.
+ */
+export function areInterpreterPathsSame(path1: string = '', path2:string = '', ostype = getOSType(), fs?: IFileSystem){
```
DonJayamanne, line 92: Added new helpers to compare paths, ensured the hash also takes this into account.

src/client/api/pythonApi.ts
IanMatthewHuff: Don't see an issue with leaving it in myself.

.github/workflows/build-test.yml
```diff
@@ -556,15 +558,15 @@ jobs:
...
-      # Upload unit test coverage reports for later use in the "reports" job.
-      - name: Upload unit test coverage reports
-        uses: actions/upload-artifact@v2
-        if: "(success() || failure()) && !contains(github.ref, 'refs/heads/release')"
-        with:
-          name: ${{runner.os}}-${{env.COVERAGE_REPORTS}}
-          path: .nyc_output
-          retention-days: 1
+    #   # Mosts tests are commented, hence nothing runs, thus no coverage (if not commented, CI falls over as this code expects coverage files)
+    #   # Upload unit test coverage reports for later use in the "reports" job.
```
DonJayamanne, line 562: This throws error, and functional atests (old style) are almost non-existent, hence don't see the need to break CI just for coverage.
Once we've fixed our functional tests we  can re-vist this

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -576,13 +603,43 @@ export function findPreferredKernel(
...
+
+                    // If the versions match exactly, then increase the score by another 1
+                    if (
+                        typeof notebookMetadata === 'object' &&
+                        'interpreter' in notebookMetadata &&
+                        notebookMetadata.interpreter &&
+                        typeof notebookMetadata.interpreter === 'object' &&
+                        'hash' in notebookMetadata.interpreter &&
+                        metadata.kind === 'startUsingPythonInterpreter' &&
+                        notebookMetadata.interpreter
```
IanMatthewHuff, line 623: notebookMetadata.interpreter looks repeated here. Also is the 'hash' needed?

IanMatthewHuff (approved):

src/client/datascience/jupyter/kernels/helpers.ts
DonJayamanne: Thanks

DavidKutu (approved):

src/test/common.ts
```diff
@@ -139,6 +139,11 @@ async function setGlobalPathToInterpreter(pythonPath?: string): Promise<void> {
     await pythonConfig.update('defaultInterpreterPath', pythonPath, true);
     await disposePythonSettings();
 }
+export async function disableExperimentsInPythonExtension(): Promise<void> {
+    const vscode = require('vscode') as typeof import('vscode');
+    const pythonConfig = vscode.workspace.getConfiguration('python', (null as any) as Uri);
+    await pythonConfig.update('experiments.enabled', false, ConfigurationTarget.Global);
```
DonJayamanne, line 145: This should fix the regressions in Python extension.

src/client/pythonEnvironments/info/interpreter.ts
```diff
@@ -76,5 +78,42 @@ export async function getInterpreterInfo(
...
+    const norm1 = getNormalizedInterpreterPath(path1, ostype);
+    const norm2 = getNormalizedInterpreterPath(path2, ostype);
+    return norm1 === norm2 || (fs ? fs.areLocalPathsSame(norm1, norm2) : false);
+}
+/**
+ * Sometimes on CI, we have paths such as (this could happen on user machines as well)
+ *  - /opt/hostedtoolcache/Python/3.8.11/x64/python
```
rchiodo, line 99: I'm not sure why these are the same? They're the same on your machine, but why everywhere this pattern is found? I would assume on the CI machine these paths are symlinked to each other?

DonJayamanne: > same? They're the same on your machine, but why everywhere this pattern is found? I would assume on the CI machine these paths are symlinked to each other?



Yes, if its `/bin/xyz` and you have `xyz` then they are the same. I.e. python doesn't hvae multiple different interpreter executables in the same directory. They have `python3.5.exe` and python.exe` in the same directory and both are the same.

& `/bin/`  is only used on linux, on windows its always `scripts/....`
Also we're looking at python file names, hence windows will always be excluded as well.

DonJayamanne: > found? I would assume on the CI machine these paths are symlinked to each other?

I think so, i didn't try to figure that out. This is very new, we didn't see this in the past.
Looking back at older runs i couldn't find this. Either ther'es a change in some python package or CI or the python extension is not finding these symlinked items.

rchiodo: I guess I'm just nervous this will break something that used to work.

Somebody has say

```
~/mystuff/python
~/mystuff/bin/python
```

And those aren't the same. I think you're saying this isn't normal. You wouldn't have bin/python and python at the same level.

DonJayamanne: > ou wouldn't have bin/python and python at the same level.

Yes. 
Because if `mystuff/python` is the python environment, then you won't have a different version interpreter under `mystuff/bin/python` (thats very very unlikely, from what i know, python doesn't setup interpreters that way).

rchiodo (approved):

PR #7416, opened 2021-09-03
RBL code cleanup

### LINKED ISSUE #7346: RBL/Debugging code cleanup
- Look at how the debug adapter tracker running the variables view interacts with the KernelDebugAdapter
- Split KernelDebugAdapter into multiple classes that only have one job
  - https://github.com/microsoft/vscode-jupyter/pull/7416
- Move the DebugAdapterDescriptorFactory into its own class (depends on variables stuff, and not setting global context keys for debug)
- We send disconnect multiple times which triggers errors from the kernel

roblourens:
Fix #7346

This moves most of the RBL and cell debug specific code out of KernelDebugAdapter into controller classes. There is some left that will be a separate change for the variables changes we discussed in #7346. 

-   [x] ~~Has a [news entry](https://github.com/Microsoft/vscode-jupyter/tree/main/news) file (remember to thank yourself!).~~
-   [x] ~~[Test plan](https://github.com/Microsoft/vscode-jupyter/blob/main/.github/test_plan.md) is updated as appropriate.~~
-   [x] ~~[`package-lock.json`](https://github.com/Microsoft/vscode-jupyter/blob/main/package-lock.json) has been regenerated by running `npm install` (if dependencies have changed).~~


src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -110,30 +104,16 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter, ID
...
-
                 if (anyMsg.header.msg_type === 'debug_event') {
-                    if (anyMsg.content.event === 'stopped') {
-                        this.threadId = anyMsg.content.body.threadId;
-                        if (await this.handleStoppedEvent()) {
-                            this.trace('intercepted', JSON.stringify(anyMsg.content));
-                            return;
-                        }
+                    this.trace('event', JSON.stringify(msg));
+                    if (!(await this.delegate?.willSendEvent(anyMsg))) {
```
IanMatthewHuff, line 114: Nice feels much better this way.

IanMatthewHuff (approved):

DavidKutu (approved):

src/client/debugger/jupyter/debugControllers.ts
```diff
@@ -0,0 +1,173 @@
...
+            if (await this.handleStoppedEvent(this.lastPausedThreadId!)) {
+                this.trace('intercepted', JSON.stringify(anyMsg.content));
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public async willSendRequest(request: DebugProtocol.Request): Promise<void> {
```
DonJayamanne, line 85: Not sure i like this `willSendrequest`.
`will` implies there's a question, and I'd expect the response to be `true/false` or similar.
But I'm not seeing a return value, hence this looks more like a command to do somethign as opposed to a method used to inquire somethign.

DonJayamanne: Never mind, read the comments in interface.
Do we need an interface

roblourens: Should I call them `onWill...`?

roblourens: And the fact that one returns a boolean while the other doesn't may or may not be bad, I can't decide

Issue #6881, opened 2021-07-28
Support for Remote in IW window


ghylander:
Issue Type: <b>Bug</b>

i have a .py file i run in an interactive window with cells
first i start a jupyter server in a WSL, then connect to it
when i try to run any cell, i get the error "Invalid response: 501 Not Implemented"

the server is started in the exact same working directory as the visual studio workspace

this doesnt always happen, i have a second jupyter server running in a different directory serving a different workspace and i can run interactive files no problem. Both use the same WSL and interpreter within it

VS Code version: Code - Insiders 1.59.0-insider (e437d424c5e6de97632f5513229c7ebc2cf27a32, 2021-07-28T05:13:47.114Z)
OS version: Windows_NT x64 10.0.19042
Restricted Mode: No

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-1035G1 CPU @ 1.00GHz (8 x 1190)|
|GPU Status|2d_canvas: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>rasterization: enabled<br>skia_renderer: enabled_on<br>video_decode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|11.70GB (1.06GB free)|
|Process Argv|--crash-reporter-id 4285aea8-8024-49ba-97f2-9d60622a70f0|
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (11)</summary>

Extension|Author (truncated)|Version
---|---|---
prettier-vscode|esb|8.1.0
kite|kit|0.147.0
python|ms-|2021.8.1054534874-dev
vscode-pylance|ms-|2021.7.6-pre.1
jupyter|ms-|2021.8.1070546638
remote-containers|ms-|0.189.0
remote-ssh|ms-|0.65.7
remote-ssh-edit|ms-|0.65.7
remote-wsl|ms-|0.58.2
vscode-remote-extensionpack|ms-|0.21.0
powershell-preview|ms-|2021.7.0

(2 theme extensions excluded)

</details><details>
<summary>A/B Experiments</summary>

```
vsliv695:30137379
vsins829:30139715
vsliv368:30146709
vsreu685:30147344
python383cf:30185419
pythonvspyt602:30291494
[...]

PR #7452, opened 2021-09-07
Support remote for IW

DonJayamanne:
For #6881

Was much easier than i thought.

### DIFF
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -215,8 +215,8 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
 
         // Find the default kernel `python` if we can find one
         // If not available, then return anything thats a python kernel
-        let defaultPythonKernel: VSCodeNotebookController | undefined;
         let defaultPython3Kernel: VSCodeNotebookController | undefined;
+        let defaultPythonKernel: VSCodeNotebookController | undefined;
         let defaultPythonLanguageKernel: VSCodeNotebookController | undefined;
         controllers.forEach((item) => {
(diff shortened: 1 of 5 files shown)


rchiodo (approved):

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -200,6 +204,38 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+                defaultPython3Kernel = item;
+            } else if (
+                item.connection.kind === 'startUsingKernelSpec' &&
+                item.connection.kernelSpec.language === PYTHON_LANGUAGE
+            ) {
+                defaultPythonLanguageKernel = item;
+            }
+        });
+
+        return defaultPythonKernel || defaultPython3Kernel || defaultPythonLanguageKernel || controllers[0];
```
IanMatthewHuff, line 237: For the default kernel would we maybe want defaultPython3Kernel to have priority over defaultPythonKernel? Feel like that would be better on mac / linux systems.

joyceerhl (approved):

src/client/datascience/notebook/notebookControllerManager.ts
DonJayamanne: @rchiodo @IanMatthewHuff i'll be updating this code, as this is assuming IW always works with python, and I do believe we used to support other languages.

rchiodo: It used to support other languages from the kernel picker AFAIK, but not on startup.

IanMatthewHuff (approved):

DavidKutu (approved):

Issue #7094, opened 2021-08-12
Review all settings, commands, activation events in package.json


joyceerhl:
Some of these look like they can be removed

Already marked deprecated for >1 year
* `jupyter.runMagicCommands`
* `jupyter.autoPreviewNotebooksInInteractivePane`

Should be deprecated
* `jupyter.liveShareConnectionTimeout` -- we have removed all our code to support Live Share 
* `jupyter.allowLiveShare`  -- we have removed all our code to support Live Share 
* `jupyter.allowImportFromNotebook` -- not used anywhere
* `jupyter.ignoreVscodeTheme` -- can't support this
* `jupyter.textOutputLimit` -- should be a core VS Code setting
* `jupyter.maxOutputSize` -- should be a core VS Code setting
* `jupyter.colorizeInputBox` -- we have discussed input box discoverability in https://github.com/microsoft/vscode/issues/128766 and https://github.com/microsoft/vscode/issues/126415#issuecomment-875721427, and we found that using background colors to draw attention to an area visually poses potential accessibility problems and is inconsistent with other treatments in VS Code. We have opted to render the input box as a notebook cell instead, which should solve the visibility issues that prompted the original colorization treatment. Therefore this setting can be deprecated as it will not be an option moving forward
* `jupyter.showCellInputCode` -- we cannot support this without significant changes in VS Code
* `jupyter.allowInput` -- controls whether to show an input box in the interactive window. We cannot support this without changes in VS Code 
* `jupyter.searchForJupyter` -- controls whether to search for a Jupyter-capable interpreter if the user-selected one doesn't work. This has not been supported for ~1 year

Issue #7168, opened 2021-08-17
Polish interactive window


joyceerhl:
We shipped a native interactive window experience in the July release. This issue tracks polishing the interactive window experience.

Done
- [x] Variable explorer should be populated when focus is in script file https://github.com/microsoft/vscode-jupyter/issues/6993 
- [x] Fix interrupt/restart https://github.com/microsoft/vscode-jupyter/issues/7081 
- [x] Provide some solution for `esc` and `ctrl+u` clearing the interactive window input box https://github.com/microsoft/vscode-jupyter/issues/7157
- [x] Context keys are not correctly set https://github.com/microsoft/vscode-jupyter/issues/7158
- [x] Support or deprecate missing settings https://github.com/microsoft/vscode-jupyter/issues/7031, https://github.com/microsoft/vscode-jupyter/issues/7094
- [x] Execution count incorrect https://github.com/microsoft/vscode-jupyter/issues/7156
- [x] Code dedented on IW export https://github.com/microsoft/vscode-jupyter/issues/7088
- [x] No execution time set https://github.com/microsoft/vscode-jupyter/issues/7162
- [x] https://github.com/microsoft/vscode-jupyter/issues/7013
- [x] https://github.com/microsoft/vscode-jupyter/issues/6881
- [x] https://github.com/microsoft/vscode-jupyter/issues/7303
- [x] Delete cell usability: 
 https://github.com/microsoft/vscode-jupyter/issues/7188, https://github.com/microsoft/vscode-jupyter/issues/7250, https://github.com/microsoft/vscode-jupyter/issues/6475
- [x] https://github.com/microsoft/vscode-jupyter/issues/7288
- [x] https://github.com/microsoft/vscode-jupyter/issues/7269
- [x] https://github.com/microsoft/vscode-jupyter/issues/7476
- [x] https://github.com/microsoft/vscode-jupyter/issues/7366
- [x] https://github.com/microsoft/vscode-jupyter/issues/7280

Upstream
- https://github.com/microsoft/vscode/issues/131659
- https://github.com/microsoft/vscode/issues/132064

Deferred
- Rework command queuing https://github.com/microsoft/vscode-jupyter/issues/7164, https://github.com/microsoft/vscode/issues/130295

ihnorton:
[...]

PR #7455, opened 2021-09-08
Remove deprecated settings and mark more settings as deprecated

joyceerhl:
Fix https://github.com/microsoft/vscode-jupyter/issues/7094 (part of https://github.com/microsoft/vscode-jupyter/issues/7168)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1444,13 +1444,14 @@
                 "jupyter.enableKeyboardShortcuts": {
                     "type": "boolean",
                     "default": true,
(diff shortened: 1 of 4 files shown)


DavidKutu (approved):

package.json
```diff
@@ -1622,12 +1610,14 @@
                 "jupyter.allowLiveShare": {
                     "type": "boolean",
                     "default": true,
+                    "deprecationMessage": "This setting is deprecated as Live Share support for the interactive window no longer exists in the Jupyter extension.",
```
rchiodo, line 1613: We should probably enter a task on us to reimplement this. The liveshare team has a bug where people are complaining about this being removed: https://github.com/MicrosoftDocs/live-share/issues/4032

joyceerhl: I opened https://github.com/microsoft/vscode-jupyter/issues/7460

Issue #7288, opened 2021-08-25
Changing the kernel doesn't work for Interactive


wwydmanski:
## Environment data

-   VS Code version: 1.59.1
-   Jupyter Extension version (available under the Extensions sidebar): v2021.8.1236758218
-   Python Extension version (available under the Extensions sidebar): v2021.8.1159798656
-   OS (Windows | Mac | Linux distro) and version: Darwin x64 20.5.0
-   Python and/or Anaconda version: Python 3.8.8
-   Type of virtual environment used: conda
-   Jupyter server running: Local 

## Expected behaviour

After switching the kernel (from 3.8.5 to 3.8.8), the button "Run cell" should execute the code in the 'Interactive' window using the newly selected Python.

## Actual behaviour

The editor proceeds to next cell, but the 'Interactive' window tries to execute it using the original kernel. 

Running the code directly from the `Type code here and press Enter (...)` works as expected.

## Steps to reproduce:

Make two Python installations available to the jupyter - a default one without ipykernel installed, and another one with an ipykernel. Change the kernel in the 'Interactive' window.

Everything works well if the default python has the IPyKernel installed.

## Logs

<details>

<summary>
IPyKernel error
</summary>

<p>

```
Error 2021-08-25 12:38:03: DataScience Error [a [Error]: IPyKernel not installed into interpreter Python 3.8.5 64-bit:/usr/bin/python3
	at E.handleKernelDependencyResponse (/home/MCB/wwydmanski/.vscode-server/extensions/ms-toolsai.jupyter-2021.8.1236758218/out/client/extension.js:90:247249)
	at E.installMissingDependencies (/home/MCB/wwydmanski/.vscode-server/extensions/ms-toolsai.jupyter-2021.8.1236758218/out/client/extension.js:90:246271)] {
  category: 'noipykernel',
  reason: 1
}
```

</p>
</details>

joyceerhl:
This is caused by the promise chaining we do. Pending promises aren't cleared when the kernel is changed, which causes all cells submitted from the Run Cell codelens to wait forever on this line: 

https://github.com/microsoft/vscode-jupyter/blob/7f1c9d0bda398d5ae5cf175932ad808f85c7fa63/src/client/datascience/interactive-window/interactiveWindow.ts#L404

PR #7469, opened 2021-09-08
Fix changing interactive window kernel when the active interpreter does not have ipykernel installed

joyceerhl:
Fix https://github.com/microsoft/vscode-jupyter/issues/7288

### DIFF
--- a/src/client/datascience/interactive-window/interactiveWindow.ts
+++ b/src/client/datascience/interactive-window/interactiveWindow.ts
@@ -173,14 +173,14 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         const controllerId = preferredController ? `${JVSC_EXTENSION_ID}/${preferredController.id}` : undefined;
         traceInfo(`Starting interactive window with controller ID ${controllerId}`);
         const hasOwningFile = this.owner !== undefined;
-        const { notebookEditor } = ((await this.commandManager.executeCommand(
+        const { notebookEditor } = (await this.commandManager.executeCommand(
             'interactive.open',
             // Keep focus on the owning file if there is one
             { viewColumn: ViewColumn.Beside, preserveFocus: hasOwningFile },
             undefined,
             controllerId,
             this.owner && this.mode === 'perFile' ? getInteractiveWindowTitle(this.owner) : undefined
-        )) as unknown) as INativeInteractiveWindow;
+        ) as unknown) as INativeInteractiveWindow;
         if (!notebookEditor) {
             // This means VS Code failed to create an interactive window.
             // This should never happen.
@@ -305,7 +305,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             controller: controller.controller,
             resourceUri: this.owner
         });
-        this.kernelLoadPromise = kernel?.start({ disableUI: false, document: notebookDocument });
+        this.kernelLoadPromise = kernel.start({ disableUI: false, document: notebookDocument });
         this.kernel = kernel;
         this.notebookController = controller;
     }
--- a/src/client/datascience/interactive-window/interactiveWindow.ts
+++ b/src/client/datascience/interactive-window/interactiveWindow.ts
@@ -173,14 +173,14 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         const controllerId = preferredController ? `${JVSC_EXTENSION_ID}/${preferredController.id}` : undefined;
         traceInfo(`Starting interactive window with controller ID ${controllerId}`);
         const hasOwningFile = this.owner !== undefined;
-        const { notebookEditor } = (await this.commandManager.executeCommand(
+        const { notebookEditor } = ((await this.commandManager.executeCommand(
             'interactive.open',
             // Keep focus on the owning file if there is one
             { viewColumn: ViewColumn.Beside, preserveFocus: hasOwningFile },
             undefined,
             controllerId,
             this.owner && this.mode === 'perFile' ? getInteractiveWindowTitle(this.owner) : undefined
-        ) as unknown) as INativeInteractiveWindow;
+        )) as unknown) as INativeInteractiveWindow;
         if (!notebookEditor) {
             // This means VS Code failed to create an interactive window.
             // This should never happen.
--- a/src/client/datascience/interactive-window/interactiveWindow.ts
+++ b/src/client/datascience/interactive-window/interactiveWindow.ts
@@ -124,8 +124,6 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
     protected fileInKernel: string | undefined;
 
     private isDisposed = false;
-    private kernel: IKernel | undefined;
-    private kernelLoadPromise: Promise<void> | undefined;
     private initialControllerSelected: Deferred<void>;
     private _editorReadyPromise: Promise<NotebookEditor>;
     private notebookDocument: NotebookDocument | undefined;
@@ -171,6 +169,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
     private async createReadyPromise(): Promise<NotebookEditor> {
         const preferredController = await this.notebookControllerManager.getInteractiveController();
         const controllerId = preferredController ? `${JVSC_EXTENSION_ID}/${preferredController.id}` : undefined;
+        this.notebookController = preferredController;
         traceInfo(`Starting interactive window with controller ID ${controllerId}`);
         const hasOwningFile = this.owner !== undefined;
         const { notebookEditor } = ((await this.commandManager.executeCommand(
@@ -187,7 +186,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             throw new Error('Failed to request creation of interactive window from VS Code.');
         }
         this.notebookDocument = notebookEditor.document;
-        this.loadController(notebookEditor.document);
+        this.listenForControllerSelectionChanges(notebookEditor.document);
         this.initializeRendererCommunication();
         return notebookEditor;
     }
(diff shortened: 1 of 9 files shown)


src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -275,34 +267,36 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
+    private async kernelReadyPromise() {
+        const kernel = this.kernelProvider.getOrCreate(this.notebookDocument!, {
+            metadata: this.notebookController!.connection,
+            controller: this.notebookController!.controller,
             resourceUri: this.owner
         });
-        this.kernelLoadPromise = kernel?.start({ disableUI: false, document: notebookDocument });
-        this.kernel = kernel;
-        this.notebookController = controller;
+        await kernel.start({ disableUI: false, document: this.notebookDocument! });
```
rchiodo, line 298: Every time you ask for this you'll be 'starting' the kernel? DOes the kernel handle this case and skip starting more than once?

If it does, can you put a comment here to say that? (so somebody in the future doesn't think this is broken)

joyceerhl: Yeah the IKernel caches the notebookPromise, so calling start again will be a noop. https://github.com/microsoft/vscode-jupyter/blob/2e1c221cdc3ec1128d55886711677f77d699e552/src/client/datascience/jupyter/kernels/kernel.ts#L231 I'll add a comment

IanMatthewHuff: I'm a tiny bit confused on this function as well. With the name I'd expect it to just be returning a promise to await for something. But it's starting up a kernel and also (if I'm reading this right) might be returning different promises if the controller changes.

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -316,8 +310,8 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
     }
 
     public dispose() {
-        if (this.kernel) {
-            this.kernel.dispose().ignoreErrors();
+        if (this.notebookController) {
+            this.notebookController.dispose();
         }
```
DonJayamanne, line 315: Won't this remove the controller from the list of controller.
Not sure we want to dispose this.
It's shared with all notebooks

src/client/datascience/jupyter/kernels/kernelProvider.ts
```diff
@@ -77,7 +77,7 @@ export class KernelProvider implements IKernelProvider {
         this.pendingDisposables.clear();
         await Promise.all(items);
     }
-    public getOrCreate(notebook: NotebookDocument, options: KernelOptions): IKernel | undefined {
+    public getOrCreate(notebook: NotebookDocument, options: KernelOptions): IKernel {
```
DonJayamanne, line 80: Eew, can't remember why I ever returned undefined, so wrong

DonJayamanne: Good fix

DonJayamanne (changes requested):
Nope sure we should be disposing the controller

joyceerhl:
Ended up reverting the change I was attempting to not cache the kernel in the IW. Debugged it locally and it appears to cause execution to slow down significantly for the #%% cell codepath (input box codepath is fine).

rchiodo (approved):

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -261,11 +261,12 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             this.registerKernel(notebookDocument, controller);
             this.initialControllerSelected.resolve();
         }
+        const currentNotebookUri = notebookDocument.uri.toString();
 
         // Ensure we hear about any controller changes so we can update our cache accordingly
         this.notebookControllerManager.onNotebookControllerSelected(
             (e: { notebook: NotebookDocument; controller: VSCodeNotebookController }) => {
-                if (e.notebook !== notebookDocument) {
+                if (e.notebook.uri.toString() !== currentNotebookUri) {
```
DonJayamanne, line 269: This change isn't necessary

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -275,10 +276,14 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
                 ).controller.onDidChangeSelectedNotebooks(
                     (selectedEvent: { notebook: NotebookDocument; selected: boolean }) => {
                         // Controller was deselected for this InteractiveWindow's NotebookDocument
-                        if (selectedEvent.selected === false && selectedEvent.notebook === notebookDocument) {
+                        if (
+                            selectedEvent.selected === false &&
+                            selectedEvent.notebook.uri.toString() === currentNotebookUri
```
DonJayamanne, line 281: This change isn't necessary (i.e. we should be able to compare notebook objects)

DonJayamanne (changes requested):
Please can we revert comparing toString of uri to comparing the object refs

src/client/datascience/interactive-window/interactiveWindow.ts
joyceerhl: When debugging this locally I found that we were failing the comparison check even when there was only one NotebookDocument (the embedded one in the IW). I don't want to rely on the NotebookDocument not changing over time for this code to work correctly

rchiodo: Given that was the reason for the string compare, can you put a comment to that effect? So that the next person looking at this doesn't go - oh we don't need to compare strings.

rchiodo: Or maybe a new function like ```doesNotebookMatch``` that has that comment in it and use that in the two places where you do the string compare?

DonJayamanne: > When debugging this locally I found that we were failing the comparison check even when there was only one NotebookDocument (the embedded one in the IW). I don't want to rely on the NotebookDocument not changin

This is a bug in VS Code. it should be the same (everywhere in our extension we use notebook documents as keys in dictionaries).
@rebornix Any idea what's going on?

DonJayamanne: Because if we cannot rely on notebook document being the same, then we should fix this in vscode or we should change the rest of our code

rebornix: It seems like a bug in VS Code, we should be able to compare notebook document as they are fat objects and always one document object per resource in one extension host.

DonJayamanne: Just check with @rebornix it should be the same, & we should be using object ref comparisons.

IanMatthewHuff (approved):

src/client/datascience/interactive-window/interactiveWindow.ts
DonJayamanne: Updated my comment from `just check` to `just checked`

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -155,11 +142,17 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
         // Wait for a controller to get selected
-        this.initialControllerSelected = createDeferred<void>();
+        this._controllerReadyPromise = createDeferred<VSCodeNotebookController>();
 
-        // Request creation of the interactive window from VS Code
-        this._editorReadyPromise = this.createReadyPromise();
+        // Set up promise for kernel ready
+        this._kernelReadyPromise = this.createKernelReadyPromise();
+
+        // Immediately start trying to connect to kernel
```
rchiodo, line 154: This line shouldn't be necessary AFAIK. Creation of the promise starts the whole thing.

src/client/datascience/jupyter/kernels/kernelDependencyService.ts
```diff
@@ -129,16 +128,10 @@ export class KernelDependencyService implements IKernelDependencyService {
...
-        // hence we get two messages about ipykernel not being installed.
-        // THat's a very poor ux, one could end up with two modal dialog boxes (one after the other for interactive).
-        // hence disabling modal dialog for interactive window for now.
-        // Again to be resolved in https://github.com/microsoft/vscode-jupyter/issues/6135
-        const modal = getResourceType(resource) === 'notebook';
         const selection = this.isCodeSpace
             ? installPrompt
             : await Promise.race([
-                  this.appShell.showErrorMessage(message, { modal }, ...options),
+                  this.appShell.showErrorMessage(message, { modal: true }, ...options),
```
joyceerhl, line 134: Making this modal for now since changing controllers while we're attempting to connect to a kernel does not cancel that connection (kernelProvider still returns the existing kernel promise)

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -94,10 +84,9 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         return this._onDidChangeViewState.event;
     }
     // Promise that resolves when the interactive window is ready to handle code execution.
-    public get readyPromise(): Promise<NotebookEditor> {
-        return this._editorReadyPromise;
+    public get readyPromise(): Promise<void> {
+        return Promise.all([this._editorReadyPromise, this._kernelReadyPromise]).then(noop, noop);
```
joyceerhl, line 88: Previously the kernel promise was part of the editor ready promise, so if the kernel failed to start e.g. due to ipykernel not being installed, subsequent executions would fail while awaiting on the editor ready (if the kernel promise successfully resolved, there wasn't an issue). PR splits out the kernel promise

DonJayamanne (approved):

Issue #7476, opened 2021-09-09
Clicking 'Change Kernel' for interactive window started from script file does nothing


joyceerhl:
1. Have an active interpreter without ipykernel installed
2. Start the interactive window from a script file 
3. Click 'Change Kernel' on the dialog that comes up
4. 🐛 Nothing happens. This is because `notebook.selectKernel` doesn't work if there isn't an active notebook editor

Based on discussion at notebook sync this morning, we can modify `notebook.selectKernel` to take the notebook widget that the kernel picker should be invoked for (`interactive.open` returns the notebook editor widget).

PR #7482, opened 2021-09-10
Fix kernel picker not being shown when changing kernels and IW is not active

joyceerhl:
Fix https://github.com/microsoft/vscode-jupyter/issues/7476, adopting https://github.com/microsoft/vscode/pull/132879

GIF of combined VS Code and Jupyter extension changes:
recording (23)

### DIFF
--- a/news/2 Fixes/7476.md
+++ b/news/2 Fixes/7476.md
@@ -1,0 +1,1 @@
+Clicking 'Change Kernel' for interactive window started from script file when ipykernel is not installed should display the kernel picker.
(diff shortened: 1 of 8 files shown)


rchiodo (approved):

joyceerhl:
Peng is out till Wednesday, merging this now because it's harmless to do so even if https://github.com/microsoft/vscode/pull/132879 hasn't been shipped (tested locally for IW and notebooks; the passed `NotebookEditor` is simply ignored)

Issue #7481, opened 2021-09-10
Magic commands as comments not working


TarjinderSingh:
When using `"jupyter.magicCommandsAsComments": true`, I use to be able to have:

```python
# %%
#!%%time
print('hi')
```

run the Python magic timeit interactively. However, it not longer does that, and just executes the chunk as a normal command.

When I go to the interactive Jupyter interface and put in:

```
%%time
print('hi')
```

I get the standard output of:

```
hi
CPU times: user 37 µs, sys: 13 µs, total: 50 µs
Wall time: 45.1 µs
```

Has something changed?

## Environment data

-   VS Code version: 1.60.0
-   Jupyter Extension version (available under the Extensions sidebar): v2021.8.2041215044
-   Python Extension version (available under the Extensions sidebar): v2021.9.1218897484
-   OS (Windows | Mac | Linux distro) and version: Mac OS 11.5.2
-   Jupyter server running: Local

## Expected behaviour

XXX

## Actual behaviour

XXX

## Steps to reproduce:

[**NOTE**: Self-contained, minimal reproducing code samples are **extremely** helpful and will expedite addressing your issue]

1. XXX

## Logs

<details>

<summary>Output for <code>Jupyter</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Jupyter</code>)
</summary>

<p>

```
XXX
```

</p>
</details>

TarjinderSingh:
I reverted the Python and Jupyter extensions (Python: v2021.6.944021595; and Jupyter: v2021.8.1013163132), and it works again. Something appears to be changed between those versions and the current versions for both.

joyceerhl:
Thanks for reporting a bug @TarjinderSingh. Fix should be in tomorrow's Insiders build and the upcoming Stable release.

PR #7485, opened 2021-09-12
Handle `jupyter.magicCommandsAsComments` setting

joyceerhl:
Fix https://github.com/microsoft/vscode-jupyter/issues/7481


rchiodo (approved):

src/test/datascience/interactiveWindow.vscode.test.ts
```diff
@@ -264,6 +264,18 @@ ${actualCode}
         assert.equal(actualCellText, actualCode);
     });
 
+    test('jupyter.magicCommandsAsComments', async () => {
```
DonJayamanne, line 267: Can we add a test for the scenario when the settign is turned `off`.

DonJayamanne (approved):
Would like to see an additional test when the setting is disabled.
I.e. right now we don't know whether the setting is adhered or not, all we know is magic commands are supported (not whether it can be controlled via a setting)

Issue #7487, opened 2021-09-13
With two notebooks side by side, enabling restart and interrupt buttons in one notebook also enables them in the other


joyceerhl:
1. Open two notebooks side by side. Restart and interrupt buttons should be initially disabled in both
2. Run code in one notebook
3. 🐛 Restart and interrupt buttons are enabled in both. They should only be enabled in the one that is actually running

![image](https://user-images.githubusercontent.com/30305945/133009428-2b4de884-29af-42e0-bc38-17852300f285.png)

greazer:
it's possible this could cause buttons to be disabled when they need to be enabled.

DonJayamanne:
Note: We might be able to get this working with the new when clauses added by @roblourens to vscode core.

roblourens:
Definitely. You can follow the same pattern from debug/run-by-line

DonJayamanne:
This should be fixed, no idea when it wasx fixed, but its working now as expected

PR #7488, opened 2021-09-13
Interrupt kernel button on interactive window toolbar should be disabled when kernel is not busy

### LINKED ISSUE #7269: Interrupt button is enabled even when there's no code running in interactive window
**In notebooks,** 
* the interrupt icon is only enabled when we have some code running.
* Similarly variable icon is only enabled when variable view is not visible

In interactive window:
* The interrupt is always enabled even when no code is running.
* Ideally it should be disabled (like in notebooks)..

joyceerhl:
Fix #7269

### DIFF
--- a/src/client/datascience/commands/activeEditorContext.ts
+++ b/src/client/datascience/commands/activeEditorContext.ts
@@ -14,7 +14,9 @@ import { traceError } from '../../common/logger';
 import { IDisposable, IDisposableRegistry } from '../../common/types';
 import { isNotebookCell } from '../../common/utils/misc';
 import { EditorContexts } from '../constants';
-import { getNotebookMetadata, isJupyterNotebook, isPythonNotebook } from '../notebook/helpers/helpers';
+import { getActiveInteractiveWindow } from '../interactive-window/helpers';
+import { JupyterNotebookView } from '../notebook/constants';
+import { getNotebookMetadata, isPythonNotebook } from '../notebook/helpers/helpers';
 import { IInteractiveWindow, IInteractiveWindowProvider, INotebook, INotebookProvider } from '../types';
 
 @injectable()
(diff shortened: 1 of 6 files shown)


src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -100,24 +115,28 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
     private onDidChangeActiveInteractiveWindow(e?: IInteractiveWindow) {
         this.interactiveContext.set(!!e).ignoreErrors();
         this.updateMergedContexts();
+        this.updateContextOfActiveInteractiveWindowKernel();
     }
```
DonJayamanne, line 119: You can use vscode built in context variable `notebookCellExecuting`
That'll be easier

joyceerhl: Should we adopt that for the notebook editor as well then?

joyceerhl: I just tried it out, that context key only works when you set focus to the notebook cell that is executing. I don't think that will work for our use case.

DonJayamanne: Try this `notebookHasRunningCell`

rchiodo (approved):

src/client/datascience/commands/activeEditorContext.ts
joyceerhl: That wouldn't work well too, e.g. if a notebook and IW are open side by side, running a cell in a notebook will cause the interrupt button in the IW toolbar to light up, and vice versa. Also in the IW case, since the IW may not actually be active (focus may be in the Python file), that would cause issues as well. The current implementation still suffers from this limitation https://github.com/microsoft/vscode-jupyter/issues/7487 for SxS IWs, but at least allows us to ensure that the IW toolbar is enabled correctly based on the active #%% file or active IW, and to ensure that IWs' and notebooks' run state doesn't affect the toolbar button enablement for each other.

DonJayamanne: yuck, agreed.

DonJayamanne (approved):

Issue #7366, opened 2021-08-31
Debugging cells in the interactive window will show the 'temp hashed file' if you restart the kernel


rchiodo:
Latest stable

Repro steps:
1. Open py file with a bunch of cell delimiters
2. Send a cell to the interactive window
3. Restart the kernel
4. Debug a cell 

This will come up:

![image](https://user-images.githubusercontent.com/19672699/131585068-017fea1a-39d0-4ba8-b1f2-a73c948b9000.png)

joyceerhl:
@rchiodo I can't repro this in Insiders now, can you?

rchiodo:
Yeah still repros for me.

rchiodo:
Steps I followed:

1. Run a cell with ipykernel 5 or so
2. Debug same cell
3. Debug same cell again
4. Restart kernel
5. Debug same cell 

Error happens

joyceerhl:
Thanks that worked. Problem was I had ipykernel 6.

joyceerhl:
Problem AFAIK is the `onKernelRestarted` method wasn't hooked up.

Issue #7463, opened 2021-09-08
Re-work restart kernel to just kill IKernel & start a new IKernel


DonJayamanne:
Remove all of the code around restarting kernel & just treat it as:
* Stop kernel
* Start a new kernel

Very similar to user switching to a new kernel.

@joyceerhl not sure if this is a dup issue, but this is something you indicated you'd work on last sprint, I'm assuming you'd still want to tackle this during the next debt week.

I'd like to see this done sooner as we have a few issues with kernel restarts (flaky tests) & feel it would be resolved if we just nixed the kernel process completely.

joyceerhl:
Note, make sure to propagate restart to CellHashProviders. Presumably IKernelProvider will emit the restart event instead. Maybe we can fix up the CellHashProvider to handle the IKernelProvider restart event and hookup the relevant listener there?

DonJayamanne:
>  handle the IKernelProvider restart 

If kernel restart is just a kill kernel & start a new kernel, will we need the event `restart` at all?
Else it gets confusing,

DonJayamanne:
closing as i don't think we need this anymore.
lot of the extension is getting refactored

PR #7497, opened 2021-09-13
Clear `CellHashProvider` internal state on `IKernel` restart

joyceerhl:
Fix https://github.com/microsoft/vscode-jupyter/issues/7366

### DIFF
--- a/news/2 Fixes/7366.md
+++ b/news/2 Fixes/7366.md
@@ -1,0 +1,1 @@
+Fix Debug Cell codelens opening the wrong source file after restarting the kernel in the interactive window.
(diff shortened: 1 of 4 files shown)


IanMatthewHuff (approved):
This looks legit for now, wasn't there some discussion that IKernel restart might go away and we just kill the IKernel and create a new one on restart?

rchiodo (approved):

joyceerhl:
Yeah https://github.com/microsoft/vscode-jupyter/issues/7463 that might not work so well then.

IanMatthewHuff:
> Yeah #7463 that might not work so well then.

I think it's fine now as that's the way the code is. Just more of a note that we might need a tweak here later. If we remove the restarted event on the kernel we can just change this then.

PR #7514, opened 2021-09-14
Better determine whether users install ipykernel

DonJayamanne:


src/client/telemetry/index.ts
```diff
@@ -811,6 +811,7 @@ export interface IEventNamePropertyMapping {
             | 'failed' // Installation disabled (this is what python extension returns).
             | 'install' // User chose install from prompt.
             | 'donotinstall' // User chose not to install from prompt.
+            | 'differentKernel' // User chose not to select a different interpreter/kernel.
```
rchiodo, line 814: I believe this text is backwards?
// User chose to select a different kernel

rchiodo (approved):

Issue #7494, opened 2021-09-13
Continue running the rest of the cell during RBL


roblourens:
We should have a way to allow ending the RBL session and running the rest of the cell. Right now pressing F5 will do this by triggering the debug "continue" command.

Ideally this would be a button in the cell toolbar but we don't want to shift the position of the RBL button when you click it to start a session, this is annoying.

DavidKutu:
This is possible with ctrl+enter. But I agree, a cell button would be more visible.

We could go back to what we had. The interrupt button on the left stops run by line and interrupts execution, and another button on the cell toolbar stops RBL but continues executing.

PR #7521, opened 2021-09-14
add button to stop RBL and continue executing

DavidKutu:
For #7494

This is what it looks like:
RBLContinue


rchiodo (approved):

roblourens (approved):
This will shift the toolbar when it is right-aligned, so I'm not sure whether we want this. But, we can try it out

roblourens:
Does the button actually show up as blue? I don't think we want that

DavidKutu:
@roblourens those are both right, I keep forgetting the cell toolbar can change alignment. Is there a way to change the group:
image
so that it always goes to the side where it doesn't move stuff?

Also, the button does turn blue. I didn't do that, that is a default VS Code icon. Is that a bug?

DonJayamanne (approved):

roblourens:
FYI I added the context key `notebookCellToolbarLocation` if you want to adopt that

Issue #7517, opened 2021-09-14
When prompted to install ipykernel we don't know what went wrong


DonJayamanne:
https://github.com/microsoft/vscode-jupyter/pull/7514

DonJayamanne:
Fixed in #7514

PR #7539, opened 2021-09-15
Better determining if user ran cell or started nb

DonJayamanne:
For https://github.com/microsoft/vscode-jupyter/issues/7517

### DIFF
--- a/src/client/datascience/activation.ts
+++ b/src/client/datascience/activation.ts
@@ -51,7 +51,7 @@ export class Activation implements IExtensionSingleActivationService {
         this.PreWarmDaemonPool().ignoreErrors();
         sendTelemetryEvent(Telemetry.OpenNotebookAll);
 
-        if (!this.rawSupported.supported() && this.extensionChecker.isPythonExtensionInstalled) {
+        if (!this.rawSupported.isSupported && this.extensionChecker.isPythonExtensionInstalled) {
             // Warm up our selected interpreter for the extension
             this.jupyterInterpreterService.setInitialInterpreter().ignoreErrors();
         }
--- a/src/client/datascience/commands/activeEditorContext.ts
+++ b/src/client/datascience/commands/activeEditorContext.ts
@@ -141,7 +141,6 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
                 .getOrCreateNotebook({
                     identity: activeEditor.document.uri,
                     resource: activeEditor.document.uri,
-                    disableUI: true,
                     getOnly: true
                 })
                 .then(async (nb) => {
@@ -167,7 +166,6 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
                 .getOrCreateNotebook({
                     identity: interactiveWindow.notebookUri,
                     resource: interactiveWindow.owner,
-                    disableUI: true,
                     getOnly: true
                 })
                 .then(async (nb) => {
--- a/src/client/datascience/commands/activeEditorContext.ts
+++ b/src/client/datascience/commands/activeEditorContext.ts
@@ -141,6 +141,7 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
                 .getOrCreateNotebook({
                     identity: activeEditor.document.uri,
                     resource: activeEditor.document.uri,
+                    disableUI: true,
                     getOnly: true
                 })
                 .then(async (nb) => {
@@ -166,6 +167,7 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
                 .getOrCreateNotebook({
                     identity: interactiveWindow.notebookUri,
                     resource: interactiveWindow.owner,
+                    disableUI: true,
                     getOnly: true
                 })
                 .then(async (nb) => {
--- a/src/client/datascience/constants.ts
+++ b/src/client/datascience/constants.ts
@@ -379,7 +379,6 @@ export enum Telemetry {
     DebugpyInstallFailed = 'DATASCIENCE.DEBUGPY_INSTALL_FAILED',
     DebugpyInstallCancelled = 'DATASCIENCE.DEBUGPY_INSTALL_CANCELLED',
     ScrolledToCell = 'DATASCIENCE.SCROLLED_TO_CELL',
-    ExecuteNativeCell = 'DATASCIENCE.NATIVE.EXECUTE_NATIVE_CELL',
     CreateNewNotebook = 'DATASCIENCE.NATIVE.CREATE_NEW_NOTEBOOK',
     DebugStepOver = 'DATASCIENCE.DEBUG_STEP_OVER',
     DebugContinue = 'DATASCIENCE.DEBUG_CONTINUE',
--- a/src/client/datascience/editor-integration/codeLensFactory.ts
+++ b/src/client/datascience/editor-integration/codeLensFactory.ts
@@ -65,7 +65,7 @@ export class CodeLensFactory implements ICodeLensFactory {
         this.configService.getSettings(undefined).onDidChange(this.onChangedSettings, this, disposables);
         notebook.onDidChangeNotebookCellExecutionState(this.onDidChangeNotebookCellExecutionState, this, disposables);
         kernelProvider.onDidDisposeKernel(
-            (kernel) => this.notebookData.delete(kernel.notebookUri.toString()),
+            (kernel) => this.notebookData.delete(kernel.notebookDocument.uri.toString()),
             this,
             disposables
         );
--- a/src/client/datascience/interactive-common/notebookProvider.ts
+++ b/src/client/datascience/interactive-common/notebookProvider.ts
@@ -28,7 +28,7 @@ export class NotebookProvider implements INotebookProvider {
     private readonly notebooks = new Map<string, Promise<INotebook>>();
     private _notebookCreated = new EventEmitter<{ identity: Uri; notebook: INotebook }>();
     private readonly _onSessionStatusChanged = new EventEmitter<{ status: ServerStatus; notebook: INotebook }>();
-    private _type: 'jupyter' | 'raw' = 'jupyter';
+    private readonly _type: 'jupyter' | 'raw' = 'jupyter';
     public get activeNotebooks() {
         return [...this.notebooks.values()];
     }
@@ -43,10 +43,7 @@ export class NotebookProvider implements INotebookProvider {
         @inject(IPythonExtensionChecker) private readonly extensionChecker: IPythonExtensionChecker,
         @inject(IConfigurationService) private readonly configService: IConfigurationService
     ) {
-        this.rawNotebookProvider
-            .supported()
-            .then((b) => (this._type = b ? 'raw' : 'jupyter'))
-            .ignoreErrors();
+        this._type = this.rawNotebookProvider.isSupported ? 'raw' : 'jupyter';
     }
     public get onNotebookCreated() {
         return this._notebookCreated.event;
@@ -62,7 +59,7 @@ export class NotebookProvider implements INotebookProvider {
         const serverType: string | undefined = settings.jupyterServerType;
 
         // Connect to either a jupyter server or a stubbed out raw notebook "connection"
-        if (await this.rawNotebookProvider.supported()) {
+        if (this.rawNotebookProvider.isSupported) {
             return this.rawNotebookProvider.connect({
                 ...options
             });
@@ -88,7 +85,7 @@ export class NotebookProvider implements INotebookProvider {
             .catch((ex) => traceWarning('Failed to dispose notebook in disposeAssociatedNotebook', ex));
     }
     public async getOrCreateNotebook(options: GetNotebookOptions): Promise<INotebook | undefined> {
-        const rawKernel = await this.rawNotebookProvider.supported();
+        const rawKernel = this.rawNotebookProvider.isSupported;
 
         // Check our own promise cache
         if (this.notebooks.get(options.identity.toString())) {
--- a/src/client/datascience/interactive-window/interactiveWindow.ts
+++ b/src/client/datascience/interactive-window/interactiveWindow.ts
@@ -64,6 +64,7 @@ import { chainWithPendingUpdates } from '../notebook/helpers/notebookUpdater';
 import { LineQueryRegex, linkCommandAllowList } from '../interactive-common/linkProvider';
 import { INativeInteractiveWindow } from './types';
 import { generateInteractiveCode } from '../../../datascience-ui/common/cellFactory';
+import { initializeNotebookTelemetryBasedOnUserAction } from '../telemetry/telemetry';
 
 type InteractiveCellMetadata = {
     inputCollapsed: boolean;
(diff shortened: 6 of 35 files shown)


src/client/datascience/activation.ts
```diff
@@ -51,7 +51,7 @@ export class Activation implements IExtensionSingleActivationService {
         this.PreWarmDaemonPool().ignoreErrors();
         sendTelemetryEvent(Telemetry.OpenNotebookAll);
 
-        if (!this.rawSupported.supported() && this.extensionChecker.isPythonExtensionInstalled) {
+        if (!this.rawSupported.isSupported && this.extensionChecker.isPythonExtensionInstalled) {
```
DonJayamanne, line 54: Renamed the method to a property (found that we were still awaiting on this in some places, when it was a sync method).

src/client/datascience/commands/activeEditorContext.ts
```diff
@@ -166,6 +167,7 @@ export class ActiveEditorContextService implements IExtensionSingleActivationSer
                 .getOrCreateNotebook({
                     identity: interactiveWindow.notebookUri,
                     resource: interactiveWindow.owner,
+                    disableUI: true,
```
DonJayamanne, line 170: Found we were not passing anything even when a UI was required.
I made this mandatory in some places (there are way too many files to fix, but this makes it more obivous about the intent).

When we remove all of the older classes this will get cleaned up, i found we have a large number of classes that pass this around and it was not possible to fix (assuming we that incorrectly pass `false` as the dfault value in other places) all of them.

src/client/datascience/constants.ts
```diff
@@ -379,7 +379,6 @@ export enum Telemetry {
     DebugpyInstallFailed = 'DATASCIENCE.DEBUGPY_INSTALL_FAILED',
     DebugpyInstallCancelled = 'DATASCIENCE.DEBUGPY_INSTALL_CANCELLED',
     ScrolledToCell = 'DATASCIENCE.SCROLLED_TO_CELL',
-    ExecuteNativeCell = 'DATASCIENCE.NATIVE.EXECUTE_NATIVE_CELL',
```
DonJayamanne, line 382: This is wrong, interactive also uses this event, hence the data is already kaput.
We have a better way of determining whether its executing via Native notebooks

src/client/datascience/editor-integration/codeLensFactory.ts
```diff
@@ -65,7 +65,7 @@ export class CodeLensFactory implements ICodeLensFactory {
         this.configService.getSettings(undefined).onDidChange(this.onChangedSettings, this, disposables);
         notebook.onDidChangeNotebookCellExecutionState(this.onDidChangeNotebookCellExecutionState, this, disposables);
         kernelProvider.onDidDisposeKernel(
-            (kernel) => this.notebookData.delete(kernel.notebookUri.toString()),
+            (kernel) => this.notebookData.delete(kernel.notebookDocument.uri.toString()),
```
DonJayamanne, line 68: Added NotebookDocument as a property on the kernel we don't need the Uri, we can pass the Document directly (we didn't pass Document as it wasn't possible in interacive, now it is)

src/client/datascience/interactive-common/notebookProvider.ts
```diff
@@ -43,10 +43,7 @@ export class NotebookProvider implements INotebookProvider {
         @inject(IPythonExtensionChecker) private readonly extensionChecker: IPythonExtensionChecker,
         @inject(IConfigurationService) private readonly configService: IConfigurationService
     ) {
-        this.rawNotebookProvider
-            .supported()
-            .then((b) => (this._type = b ? 'raw' : 'jupyter'))
-            .ignoreErrors();
+        this._type = this.rawNotebookProvider.isSupported ? 'raw' : 'jupyter';
```
DonJayamanne, line 46: supported was sync. made a property and now obvious & cleaner

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -171,12 +172,13 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
         const editor = await this._editorReadyPromise;
         const controller = await this._controllerReadyPromise.promise;
+        initializeNotebookTelemetryBasedOnUserAction(editor.document.uri, controller!.connection);
         const kernel = this.kernelProvider.getOrCreate(editor.document, {
             metadata: controller!.connection,
             controller: controller!.controller,
             resourceUri: this.owner
         });
-        await kernel.start({ disableUI: false, document: editor.document });
+        await kernel.start({ disableUI: false });
```
DonJayamanne, line 181: the kernel is tied to a notebook (1-1 relationship), hence moot point to pass the notebook all over again to the kernel

src/client/datascience/interactive-window/interactiveWindowCommandListener.ts
```diff
@@ -341,6 +341,7 @@ export class NativeInteractiveWindowCommandListener implements IDataScienceComma
             // Create a new notebook
             notebook = await this.notebookProvider.getOrCreateNotebook({
                 identity: createExportInteractiveIdentity(),
+                disableUI: false,
```
DonJayamanne, line 344: If we're exporting, then notebooks are started, the UI should be displayed. Its not a hidden action, the user chose to do this.

src/client/datascience/ipywidgets/ipyWidgetMessageDispatcher.ts
```diff
@@ -395,6 +395,7 @@ export class IPyWidgetMessageDispatcher implements IIPyWidgetMessageDispatcher {
             this.notebook = await this.notebookProvider.getOrCreateNotebook({
                 identity: this.notebookIdentity,
                 resource: this.notebookIdentity,
+                disableUI: true,
                 getOnly: true
```
DonJayamanne, line 399: The API is not great, if its `getOnly`, then the value of `disableUI` should not matter at all.
Ideally we need two seprate methods, one to `get` (this will get if found, else return undefined) and the other `getOrCreate` (this will get if found, else create)
This is the pattern we follow elsewhere.

Makes the `getOnly` argument unnecessary and avoids the need for `disableUI` in `getOnly` scenarios.

src/client/datascience/jupyter/jupyterExecution.ts
```diff
@@ -118,15 +118,15 @@ export class JupyterExecutionBase implements IJupyterExecution {
 
     /* eslint-disable complexity,  */
     public connectToNotebookServer(
-        options?: INotebookServerOptions,
```
DonJayamanne, line 121: Found this was never optional, we always pass values.
Found a large number of places where this was the case, but couldn't be bothered fixing that.
Came here when looking at `displayUI`

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -37,12 +35,11 @@ export class CellExecutionQueue {
         private readonly notebookPromise: Promise<INotebook>,
         private readonly executionFactory: CellExecutionFactory,
         readonly metadata: Readonly<KernelConnectionMetadata>
-    ) {}
+    ) { }
     /**
      * Queue the cell for execution & start processing it immediately.
      */
     public queueCell(cell: NotebookCell): void {
-        sendKernelTelemetryEvent(cell.notebook.uri, Telemetry.ExecuteCell);
```
DonJayamanne, line 45: Moved into the `Kernel.ts` everything uses `Kernel.ts`

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -784,7 +783,7 @@ export async function sendTelemetryForPythonKernelExecutable(
         });
         const execOutput = await execService.exec(['-c', 'import sys;print(sys.executable)'], { throwOnStdErr: false });
         if (execOutput.stdout.trim().length > 0) {
-            const match = execOutput.stdout.trim().toLowerCase() === sysExecutable;
+            const match = areInterpreterPathsSame(execOutput.stdout.trim().toLowerCase(), sysExecutable);
```
DonJayamanne, line 786: Better way to check if interpreters are the same, we already use this same function a few lines above (line 764)

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -135,11 +137,15 @@ export class Kernel implements IKernel {
             outputTracker,
             cellHashProviderFactory
         );
+        trackKernelResourceInformation(notebookDocument.uri, {
+            kernelConnection: kernelConnectionMetadata,
+        });
     }
     private perceivedJupyterStartupTelemetryCaptured?: boolean;
     public async executeCell(cell: NotebookCell): Promise<NotebookCellRunState> {
+        sendKernelTelemetryEvent(cell.notebook.uri, Telemetry.ExecuteCell);
```
DonJayamanne, line 146: This is used by interactive and native notebooks.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -148,26 +154,26 @@ export class Kernel implements IKernel {
         await promise;
         return promise;
     }
-    public async executeHidden(code: string, document: NotebookDocument) {
+    public async executeHidden(code: string) {
```
DonJayamanne, line 157: No need to pass document in, we already have this.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -224,7 +230,11 @@ export class Kernel implements IKernel {
             );
         }
     }
-    private async startNotebook(options: { disableUI?: boolean; document: NotebookDocument }): Promise<INotebook> {
+    private async startNotebook(options: { disableUI: boolean }): Promise<INotebook> {
+        if (!options.disableUI) {
+            // This means the user is actually running something against the kernel (deliberately).
+            initializeNotebookTelemetryBasedOnUserAction(this.notebookDocument.uri, this.kernelConnectionMetadata);
```
DonJayamanne, line 236: If we're displaying a UI, then we know the user actually started the notebook deliberately.
Meaning the user ran a cell.

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -167,6 +171,8 @@ export class VSCodeNotebookController implements Disposable {
         if (!this.workspace.isTrusted) {
             return;
         }
+        initializeNotebookTelemetryBasedOnUserAction(notebook.uri, this.connection);
+        sendKernelTelemetryEvent(notebook.uri, Telemetry.ExecuteCell);
```
DonJayamanne, line 175: User ran a cell, this is called by VS Code when user runs cells via  VS Code built in commands.
Doing this earlier guarantees the fact that we'll properly track this even as a deliberate action by user to start a notebook/run a cell.

DonJayamanne: Other parts of the code can get executed via export/interactive window, etc.

src/client/datascience/telemetry/telemetry.ts
```diff
@@ -27,13 +27,6 @@ import { createDeferred } from '../../common/utils/async';
  */
 type ContextualTelemetryProps = {
     kernelConnection: KernelConnectionMetadata;
-    /**
-     * Used by WebViews & Interactive window.
```
DonJayamanne, line 31: Removed because of the comments, this was not used anywhere anymore.

src/client/datascience/telemetry/telemetry.ts
```diff
@@ -195,9 +198,6 @@ export function trackKernelResourceInformation(resource: Resource, information:
         ) {
             currentData.switchKernelCount = (currentData.switchKernelCount || 0) + 1;
         }
-        if (information.kernelConnectionChanged) {
-            currentData.switchKernelCount = (currentData.switchKernelCount || 0) + 1;
-        }
```
DonJayamanne, line 200: This only applied to webviews. that code doesn't exist, hence no need

src/client/datascience/variablesView/notebookWatcher.ts
```diff
@@ -56,7 +56,6 @@ export class NotebookWatcher implements INotebookWatcher {
 
     private readonly _onDidExecuteActiveNotebook = new EventEmitter<{ executionCount: number }>();
     private readonly _onDidChangeActiveNotebook = new EventEmitter<{
-        notebook?: INotebook;
```
DonJayamanne, line 59: This was not used.

src/client/datascience/variablesView/types.ts
```diff
@@ -33,7 +33,6 @@ export class IVariableViewPanelMapping {
 }
 
 export interface IActiveNotebookChangedEvent {
-    notebook?: INotebook;
```
DonJayamanne, line 36: This was not used.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -105,8 +104,11 @@ export class Kernel implements IKernel {
     private readonly kernelExecution: KernelExecution;
     private startCancellation = new CancellationTokenSource();
     private _workingDirectory?: string;
+    public get notebookUri() {
+        return this.notebookDocument.uri;
+    }
     constructor(
-        public readonly notebookUri: Uri,
+        public readonly notebookDocument: NotebookDocument,
```
DonJayamanne, line 111: Tie the NotebookDocument and not the Uri.

src/client/datascience/jupyter/kernels/kernelCommandListener.ts
```diff
@@ -194,8 +194,9 @@ export class KernelCommandListener implements IDataScienceCommandListener {
             if (exc instanceof JupyterKernelPromiseFailedError && kernel) {
                 // Old approach (INotebook is not exposed in IKernel, and INotebook will eventually go away).
                 const notebook = await this.notebookProvider.getOrCreateNotebook({
-                    resource: notebookDocument.uri,
-                    identity: notebookDocument.uri,
+                    resource: kernel.notebookDocument.uri,
+                    // TODO: This would be wrong for interactive window;
```
DonJayamanne, line 198: @joyceerhl /cc
I'll create an issue for this.
I think we're passing the wrong `resource` at this point for interactive window.

src/client/datascience/jupyter/kernels/kernelExecution.ts
```diff
@@ -54,12 +54,10 @@ export class KernelExecution implements IDisposable {
         );
     }
 
-    @captureTelemetry(Telemetry.ExecuteNativeCell, undefined, true)
     public async executeCell(notebookPromise: Promise<INotebook>, cell: NotebookCell): Promise<NotebookCellRunState> {
         if (cell.kind == NotebookCellKind.Markup) {
             return NotebookCellRunState.Success;
         }
-        sendKernelTelemetryEvent(cell.notebook.uri, Telemetry.ExecuteNativeCell);
```
DonJayamanne, line 62: This is redundant, as this code gets called by interactive as well.
We have an event `ExecuteCell` which gets called higher up, for interactive & native.

rchiodo (approved):

Issue #6290, opened 2021-06-14
Allow user to pick the 'default' kernel when starting


cassiebreviu:
## Environment data
- VS Code Version: Version: 1.57.0-insider (user setup)
                            Commit: b4c1bd0
                            Date: 2021-06-09T11:49:36.076Z
                            Electron: 12.0.9
                            Chrome: 89.0.4389.128
                            Node.js: 14.16.0
                            V8: 8.9.255.25-electron.0
                            OS: Windows_NT x64 10.0.19042
OS Version: Windows 10 20H2 OS Build 19042.1052
-   Extension version (available under the Extensions sidebar): v2021.7.922878407-dev
-   Python version (& distribution if applicable, e.g. Anaconda): Anaconda
-   Type of virtual environment used: conda

When using conda envs with vs code the default selected environment on opening vs code doesn't select the expected environment from either the settings.json or the cmd line. This is surfaced in a couple ways:

1. A common way to select the conda env is thru the cli. However if I activate my conda env in the cli and then open vs code with `code .` it doesn't respect the conda env I had active. This makes using the `jupyter notebook` command and going into classic more appealing. It will open Jupyter with my env in my directory with my notebooks ready.

    Steps to Reproduce:
    
    1. Open command
    2. Type command `conda activate myenv`
    3. Type command `code .`
    4. vs code will open with the current directory but the env selected will be whatever env was active from the last notebook in vs code not the one that was activated thru the cli or set for this specific notebook the last time it was opened in vs code.
    

2. If I open an existing Jupyter notebook in vs code, and select the default conda env. The next time I go into that notebook, it doesn't remember the default env.
    Steps to Reproduce:
    
    1. Open a notebook in vs code and select the environment
    2. Open a different notebook and select a different environment
    3. Now open the first notebook again, notice it has the environment selected from step 2 (the second notebook) and not the selected environment for the last time that specific notebook was open
<details>

<summary>Output for <code>Python</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Python</code>)
</summary>

<p>

```
XXX
```

</p>
</details>

IanMatthewHuff:
@cassieview Thanks for the feedback here. Several parts of this sound like a bug that I'm not seeing right now, so apologies, but I might ask a clarifying question or two.

> not the one that was activated thru the cli 

[...]

PR #7540, opened 2021-09-15
Draft - Current Env Kernel

IanMatthewHuff:
For #

### DIFF
--- a/src/client/datascience/jupyter/kernels/types.ts
+++ b/src/client/datascience/jupyter/kernels/types.ts
@@ -62,6 +62,7 @@ export type KernelSpecConnectionMetadata = Readonly<{
     interpreter?: PythonEnvironment;
     kind: 'startUsingKernelSpec';
     id: string;
+    useProcessEnv?: boolean;
 }>;
 /**
  * Connection metadata for Kernels started using default kernel.
--- a/src/client/datascience/kernel-launcher/kernelProcess.ts
+++ b/src/client/datascience/kernel-launcher/kernelProcess.ts
@@ -360,12 +360,19 @@ export class KernelProcess implements IKernelProcess {
             // First part of argument is always the executable.
             const executable = this.launchKernelSpec.argv[0];
             traceInfo(`Launching Raw Kernel & not daemon ${this.launchKernelSpec.display_name} # ${executable}`);
+
+            const useProcessEnv =
+                this._kernelConnectionMetadata.kind === 'startUsingKernelSpec' &&
+                this._kernelConnectionMetadata.useProcessEnv;
+
             const [executionService, env] = await Promise.all([
                 this.processExecutionFactory.create(this.resource),
                 // Pass undefined for the interpreter here as we are not explicitly launching with a Python Environment
                 // Note that there might still be python env vars to merge from the kernel spec in the case of something like
                 // a Java kernel registered in a conda environment
-                this.kernelEnvVarsService.getEnvironmentVariables(this.resource, undefined, this.launchKernelSpec)
+                !useProcessEnv
+                    ? this.kernelEnvVarsService.getEnvironmentVariables(this.resource, undefined, this.launchKernelSpec)
+                    : process.env
             ]);
 
             // Add quotations to arguments if they have a blank space in them.
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -29,10 +29,15 @@ import {
     isLocalLaunch,
     isPythonKernelConnection
 } from '../jupyter/kernels/helpers';
-import { IKernelProvider, KernelConnectionMetadata, PythonKernelConnectionMetadata } from '../jupyter/kernels/types';
+import {
+    IKernelProvider,
+    KernelConnectionMetadata,
+    KernelSpecConnectionMetadata,
+    PythonKernelConnectionMetadata
+} from '../jupyter/kernels/types';
 import { ILocalKernelFinder, IRemoteKernelFinder } from '../kernel-launcher/types';
 import { PreferredRemoteKernelIdProvider } from '../notebookStorage/preferredRemoteKernelIdProvider';
-import { INotebookProvider } from '../types';
+import { IJupyterKernelSpec, INotebookProvider } from '../types';
 import { getNotebookMetadata, isPythonNotebook } from './helpers/helpers';
 import { VSCodeNotebookController } from './vscodeNotebookController';
 import { INotebookControllerManager } from './types';
@@ -130,6 +135,9 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
     public async loadNotebookControllers(): Promise<void> {
         if (!this.controllersPromise) {
             const stopWatch = new StopWatch();
+
+            // First thing see if we should load up our current environment as an available kernel
+            this.loadCurrentPythonEnvironmentKernel();
 
             // Fetch the list of kernels ignoring the cache.
             Promise.all([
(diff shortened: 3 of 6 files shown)


IanMatthewHuff:
Putting this up for draft discussion, there are a number of things that I'd want to hash out before finalizing. 

Goals:
No Python extension required
Controller is there right away
Should not need to change kernels if you launch code from a different environment

Basics:

1. Turn on the setting Current Python Environment Kernel (default false)
2. Reload VS Code (new kernel is only added on initial Controller load)
3. There is now a kernel available like this
4. 
image
5. This kernel is available immediately and does not need the Python extension
6. This kernel does a kernelspec launch using a fake default python kernelspec
7. The python argument that we use for that launch defaults to 'python' but I added a setting to configure so this could use 'python3' or a full path to python
8. Now if you launch from a Conda env (note the custom env var) you can select Current Python Environment
image
image
9. The python used is from functionalTestEnv and the custom Conda env var is picked up
10. Close VS Code now and launch from a non-Conda venv. The Controller is still the selected controller for the document.
image
image
11. It launches with the new python from the .venv

Open Questions:

- [ ] What degree of error checking would this need? Right now I'm checking nothing, just launch it. Didn't want to have to wait for the python extension to check for IPykernel. This requires that you are ready to rock. It just fails in a non-clean way.
- [ ] In lots of scenarios this could be pointing at system python (on MacOS) which is not what we want to point customers at in general.
- [ ] This puts a bit of a goofy kernelspec into the file, might not be good for roundtripping.

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -360,12 +360,19 @@ export class KernelProcess implements IKernelProcess {
...
+                this._kernelConnectionMetadata.kind === 'startUsingKernelSpec' &&
+                this._kernelConnectionMetadata.useProcessEnv;
+
             const [executionService, env] = await Promise.all([
                 this.processExecutionFactory.create(this.resource),
                 // Pass undefined for the interpreter here as we are not explicitly launching with a Python Environment
                 // Note that there might still be python env vars to merge from the kernel spec in the case of something like
                 // a Java kernel registered in a conda environment
-                this.kernelEnvVarsService.getEnvironmentVariables(this.resource, undefined, this.launchKernelSpec)
+                !useProcessEnv
```
IanMatthewHuff, line 373: Looking at this now...I think I could possibly just pack the process.env into the fake kernelspec. Not fully sure if I want the merge though.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -549,4 +557,32 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+            const currentEnvKernelSpec: IJupyterKernelSpec = {
+                name: 'vscodeCurrentPythonEnvironmentKernel',
+                path: '/vscodeCurrentPythonEnvironmentKernel',
+                display_name: 'Current Python Environment',
+                argv: [pythonPath || 'python', '-m', 'ipykernel_launcher', '-f', '{connection_file}']
+            };
+
+            const connectionMetadata: KernelSpecConnectionMetadata = {
+                kernelSpec: currentEnvKernelSpec,
+                kind: 'startUsingKernelSpec',
```
IanMatthewHuff, line 580: Make sure to use startUsingKernelSpec to avoid the Python launch paths.

IanMatthewHuff: Also need loc on the names.

rchiodo:
I would think if the active interpreter was set in python extension, it would use that. That's what I'd think my 'current python environment' would be (not the one I started VS code with).

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -549,4 +557,32 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+    private loadCurrentPythonEnvironmentKernel() {
+        const config = this.configuration.getSettings();
+
+        if (config.currentPythonEnvironmentKernel) {
+            traceInfo('Load current environment kernel');
+
+            const pythonPath = config.currentPythonEnvironmentKernelPath;
+
+            // Create our KernelSpec and connection metadata that we will use to launch with
+            const currentEnvKernelSpec: IJupyterKernelSpec = {
```
rchiodo, line 571: Yeah I think I wouldn't expect this if I picked 'current environment'. I'd expect it to use the active interpreter.

rchiodo: This would be a good fallback if the python extension isn't installed (or active interpreter isn't set)

IanMatthewHuff:
> I would think if the active interpreter was set in python extension, it would use that. That's what I'd think my 'current python environment' would be (not the one I started VS code with).

@rchiodo Yeah, that (active python interpreter kernel) has come up in these discussions some as well. I believe that it's a bit of an orthogonal issue? This proposal was specific to users who are launching from things like activated conda terminals where they expect to set their environment in command prompt / terminal then expect to use that: https://github.com/microsoft/vscode-jupyter/issues/6290. I don't believe that those users want or care about the current python active interpreter. They would then still have to use the python extension to swap interpreters. These customers are using the command line as the kernel picker, they don't want to change anything (kernel or active python interpreter) in VS Code.

The name is crappy though, it does sound much more like the concept you are describing (current python extension active interpreter) versus what I was intending (current environment VS Code was launched in).

rchiodo:
> > I would think if the active interpreter was set in python extension, it would use that. That's what I'd think my 'current python environment' would be (not the one I started VS code with).
> 
> @rchiodo Yeah, that (active python interpreter kernel) has come up in these discussions some as well. I believe that it's a bit of an orthogonal issue? This proposal was specific to users who are launching from things like activated conda terminals where they expect to set their environment in command prompt / terminal then expect to use that: #6290. I don't believe that those users want or care about the current python active interpreter. They would then still have to use the python extension to swap interpreters. These customers are using the command line as the kernel picker, they don't want to change anything (kernel or active python interpreter) in VS Code.
> 
> The name is crappy though, it does sound much more like the concept you are describing (current python extension active interpreter) versus what I was intending (current environment VS Code was launched in).

I thought we (or python team) had discussed having the 'active interpreter' automatically match the current environment if it's never been set before? Then both would work. You start VS code from conda, active interpreter (which I believe you'd want) and this kernel would all work.

src/client/datascience/jupyter/kernels/types.ts
```diff
@@ -62,6 +62,7 @@ export type KernelSpecConnectionMetadata = Readonly<{
     interpreter?: PythonEnvironment;
     kind: 'startUsingKernelSpec';
     id: string;
+    useProcessEnv?: boolean;
```
DonJayamanne, line 65: Would be nice to have some comments around this

src/client/datascience/kernel-launcher/kernelProcess.ts
DonJayamanne: Can we flip this instead of looking for `!useProcessEnv`, why not look for `useProcessEnv ?...` Else its a mental switch

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -549,4 +557,32 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             controller.dispose();
         });
     }
+
+    // Load up a python kernel to just launch in the current vscode environment
+    private loadCurrentPythonEnvironmentKernel() {
+        const config = this.configuration.getSettings();
+
+        if (config.currentPythonEnvironmentKernel) {
```
DonJayamanne, line 565: Can we put this behind an experiment?
And optionally enable this for all new users (VS Code providse a way to identify new users)

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -549,4 +557,32 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             controller.dispose();
         });
     }
+
+    // Load up a python kernel to just launch in the current vscode environment
+    private loadCurrentPythonEnvironmentKernel() {
+        const config = this.configuration.getSettings();
```
DonJayamanne, line 563: Do we want this kernel to be the preferred kernel for all new vs code users?
Or should that be in an experiment.

I feel that even if we merged this in, very very few would actually use this
& I think the new users are the ones to benefit the most from this (but only via an experiment, else its possible).

src/client/datascience/kernel-launcher/kernelProcess.ts
DonJayamanne: Will this approach work for Conda environments? I guess the hypothesis is that users will have actiavted their conda environments isn't it?
My concern is users won't have done that and things will fall over.

Can we instead check if this is a conda environment & see if we have the conda env variable setup in process.env (then we'd know that the user hasn't activated the conda environment before launching VS Code).
One simple hack is to look for the name `conda` or the like in the python executable path (this way we are still fast and don't have a dependency on python extension)

Issue #7470, opened 2021-09-08
White background fills entire output cell for Matplotlib plots


ianholst:
When I make a simple matplotlib plot, there is a white background covering the entire output cell area, which is pretty jarring in dark themes. I would expect it to only be white in the area immediately behind the figure. Also, I'm not sure why there is a small dark cutout in the lower right corner of the white background.

![vscodejupyter](https://user-images.githubusercontent.com/2507974/132594559-4067104f-2304-4c2e-aecc-533a11e4a68f.png)

I'm guessing that it's built this way because by default, the figure background is transparent. However, even if the plot image was partially transparent, I would not expect that to affect any parts of the screen beside the image itself, so this might be undesirable behavior.

## Steps to reproduce:

1. Create new blank notebook
2. Enter code in cell:
```python
import matplotlib.pyplot as plt
plt.figure()
plt.plot([1,2], [1,2])
plt.show()
```
3. Run cell

## Environment data

-   VS Code version: 1.61.0-insider
-   Jupyter Extension version: v2021.9.1001208778
-   Python Extension version: v2021.10.1210725981-dev
-   OS: Windows 10
-   Python 3.9.7
-   Type of virtual environment used: none
-   Jupyter server running: Local
-   No other extensions running besides Python + Jupyter

DonJayamanne:
THanks for filing this issue.
I think we need to fix this, have had similar reports in the past as well https://github.com/microsoft/vscode-jupyter/issues/7066

greazer:
The problem here is that to fix this we would likely need to parse the matplotlib html to tweak what the output looks like. Then on top of that we have VS Code themes to worry about. We're not likely going to make a change here in the near future.

ianholst:
Perhaps an option to turn it off then, with it on by default?

DonJayamanne:
Re-opening to discuss this:
We can apply the background to the image element instead of the container DIV, that would solve the issue

ianholst:
I think that would be an ideal solution. Thank you!

IanMatthewHuff:
Verified, but buttons are floating off on the right now, I'll open a new issue for that seems a bit off to me:
![image](https://user-images.githubusercontent.com/812783/135359738-440f0604-dd4d-4269-9f56-6eeff53de951.png)

charlieh108:
Hi, 
I'm sorry if this isn't the right place, but I seem to have almost the exact opposite error! 

[...]

Issue #1909, opened 2020-05-06
Split out renderers into separate bundle (extension)


DonJayamanne:
### Why?
* Faster builds (CI, tests, etc)
    * No need to build this
* Smaller extension size
    * Renderer extension does not need to ship everytime
    * Once users have this, we don't need to download again
* Easier to manage 
    * Can update out of band

### User experience
- [x] microsoft/vscode#132009
- [ ] #6404

DonJayamanne:
* Testing:
* Jupyter renderer extension will be installed automatically when installing the Jupyter extesnsion (https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter-renderers)
* FYI - Jupyter insiders extension (available in marketplace for VS Code insiders) comes bundled with this renderer extension

PR #7546, opened 2021-09-15
Split Notebook renderers into a separate extension

DonJayamanne:
For #1909 , #7470
Corresponding PR is here https://github.com/microsoft/vscode-notebook-renderers/pull/59

We need to pull in the image renderers as they have comms, and comms can only be done with renderers when both are in the same extension (requirement of VS Code API).


build/webpack/webpack.datascience-ui.config.builder.js
```diff
@@ -40,6 +40,10 @@ function getEntry(bundle) {
             return {
                 errorRenderer: [`./src/datascience-ui/error-renderer/index.ts`]
             };
+        case 'imageRenderer':
+            return {
+                imageRenderer: ['./src/datascience-ui/image-renderer/index.ts']
```
DonJayamanne, line 45: This renderer adds the save/plot icon for images.

src/datascience-ui/image-renderer/index.ts
```diff
@@ -0,0 +1,363 @@
...
+            container.appendChild(div);
+        }
+    } catch (ex) {
+        console.error(`Failed to render mime type ${mimeString}`, ex);
+    }
+}
+
+function createSaveAsButton(outputItem: OutputItem, ctx: RendererContext<unknown>) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const btn = document.createElement('button') as any;
```
DonJayamanne, line 106: Using pure JS,  no need of React.
I couldn't use react as we use webpack 4 and we cannot bundle react projects into ESM targets.
It was either upgrade to webpack 5 (complicated & mess) or simplers to just ditch react

package.json
```diff
@@ -9,7 +9,8 @@
         "name": "Microsoft Corporation"
     },
     "extensionPack": [
-        "ms-toolsai.jupyter-keymap"
+        "ms-toolsai.jupyter-keymap",
+        "ms-toolsai.notebook-renderers"
```
DonJayamanne, line 13: The renderer extension will contain the renderers for vega, plotly, etc

package.json
```diff
@@ -1828,30 +1829,15 @@
...
             },
             {
-                "id": "jupyter-notebook-renderer",
-                "entrypoint": "./out/client_renderer/renderers.js",
-                "displayName": "Jupyter Notebook Renderer",
-                "requiresMessaging": "optional",
+                "id": "jupyter-notebook-image-renderer",
+                "entrypoint": "./out/datascience-ui/imageRenderer/imageRenderer.js",
+                "displayName": "Jupyter Image Renderer",
+                "requiresMessaging": "always",
```
DonJayamanne, line 1835: Jupyter extension ships with rendere for images so we can have save as / open plotviewer icons

src/datascience-ui/image-renderer/index.ts
IanMatthewHuff: Does the esm-webpack-plugin work here? As I recall that's actually what you pointed out to me when I was doing work with the renderers needed to support module loading instead of script loading. We're already using it in our extension for that purpose.

IanMatthewHuff: Actually nevermind, I'd prefer not to use react for something this simple anyways :).

rchiodo (approved):

src/datascience-ui/image-renderer/index.ts
```diff
@@ -0,0 +1,200 @@
...
+    btn.style.top = '5px';
+    btn.style.left = '45px';
+    btn.className = 'plotIcon hidden';
+    btn.ariaPressed = 'false';
+    btn.title = 'Save As';
+    btn.ariaLabel = 'Save As';
+    btn.onclick = () => {
+        if (ctx.postMessage) {
+            ctx.postMessage({
+                type: 'openPlot',
```
IanMatthewHuff, line 158: One small tweak here that I think we should make. This is just a string to string comparison. Would be nice to hardcode it now that it's in the same project so we don't accidentally change one side and not the other. Also for SaveAs.

src/datascience-ui/image-renderer/index.ts
DonJayamanne: > oes the esm-webpack-plugin work here

No it doesn't, i think react has somethign special that makes it not work.

> d, I'd prefer not to use react for something this simple anyways :).

I agree, react is the new jquery, everyone seems to use that for anythign and everything to do with HTML...
The code isn't clean, but i think its readable enough and makes it smaller without react.

IanMatthewHuff (changes requested):
Just one small tweak, would like 'openPlot' and 'saveAs' to be in one location now and not two separate strings.

IanMatthewHuff (approved):
Just changing to approved as it's just a small tweak I was suggesting.

src/datascience-ui/image-renderer/index.ts
DonJayamanne: Done

DavidKutu (approved):

.vscode/settings.json
```diff
@@ -1,7 +1,7 @@
 // Place your settings in this file to overwrite default and user settings.
 {
     "files.exclude": {
-        "out": true, // set this to true to hide the "out" folder with the compiled JS files
+        "out": false, // set this to true to hide the "out" folder with the compiled JS files
```
joyceerhl, line 4: Was this change intentional?

news/1 Enhancements/1909.md
```diff
@@ -0,0 +1 @@
+Split Notebook renderers into a seprate extension.
```
joyceerhl, line 1:
```suggestion
Split Notebook renderers into a separate extension.
```

.vscode/settings.json
DonJayamanne: oops

Issue #7028, opened 2021-08-09
Re-enable Web View Tests we have been skipping


DonJayamanne:
* [ ] Notebook Tests
	* [ ] Remote Self Certificate
	* [ ] Remote No auth (Connect to a server that doesn't have a token or password, customers use this and we regressed it once)
	* [ ] Remote Deny insecure (For a connection to a remote machine that is not secure deny the connection and we should not connect)
	* [ ] Remote Password
	* [x] Export/Import
	* [ ] Non default config fails (should not be able to connect to notebook server with bad config)
	* [x] Custom command line (@donjayamanne)
	* [x] Invalid kernelspec works (@donjayamanne) - no longer applicable after refactoring
	* [ ] Kernelspec generates an env section for a conda env: https://github.com/microsoft/vscode-jupyter/issues/7186  (@donjayamanne)
	* [ ] Server cache working (@donjayamanne)
	* [ ] Execution logging (we only need tests for pre-execute, post execute should be removed & we should use native VS Code API)
	* [ ] Notebook launch failure
	* [ ] Startup commands (@donjayamanne)
	* [ ] Current directory (@donjayamanne)
* [ ] Native Editor functional tests @rchiodo
	* [ ] Export with no python
	* [ ] Invalid session still runs
	* [ ] Invalid kernel still runs
	* [ ] Invalid kernel can be switched
	* [ ] Select Jupyter Server
	* [ ] Server already loaded
	* [ ] Server load skipped
	* [ ] Convert to python
	* [ ] Toggle visibility of output (tests for custom commands), e.g. 
	* [ ] Stop on Error Off (with setting off, I think this is required for Interactive Window)
* [ ] Jupyter Provider Registration
	* [x] #8573
* [ ] Interactive Window
	* [x] Export
	* [x] ~Restart with session failure (I don't think we should do this, when using Native Notebook controller code, this will be unnecessary. Else we'll have to write test for everything we have in Native notebooks, & that seems wrong)~
	* [ ] Copy back to source
* [ ] :runner: Intellisense Functional tests @rchiodo
	* [ ] :runner: Filtered Jupyter autocomplete, verify magic commands appear
	* [ ] :runner: Filtered Jupyter autocomplete, verify magic commands are filtered
	* [ ] :runner: Hover on interactive
* [x] Raw kernel (not required, as we have sufficient native notebook tests that cover this)
* [x] Variable Explorer tests
* [x] Plot Viewer tests
[...]

PR #7565, opened 2021-09-16
Submit tests for intellisense (replacing functional tests we had)

rchiodo:
For #7028 

I worked on this a week ago and was waiting on the python extension to submit changes I made. They should be there in the python insiders bits.

However I also want these tests in place for the work I'm doing to move intellisense to the jupyter extension. Once these are in, I'll make changes/add to them to test the work I'm doing to create a language server per interpreter.


IanMatthewHuff (approved):

src/client/datascience/notebook/intellisense/completionProvider.ts
```diff
@@ -76,7 +77,10 @@ export class NotebookCompletionProvider implements CompletionItemProvider {
             })
         ]);
         if (!result) {
+            traceInfoIf(isCI, `Notebook completions not found.`);
             return [];
+        } else {
+            traceInfoIf(isCI, `Completions found, filtering the list: ${JSON.stringify(result)}.`);
```
DonJayamanne, line 83: I think we should just turn this into `traceIfCI`, lol, feels kinda redundant to always pass in `isCI`. What do you think.
We might have a few places where we'd have some other conditions being passed in, but i think traceIfCI` or `traceOnCI` is simpler.
Absolutely no need to make that change, just a thought

src/test/datascience/notebook/intellisense/diagnostics.vscode.test.ts
```diff
@@ -0,0 +1,76 @@
...
+        const cell = vscodeNotebook.activeNotebookEditor?.document.cellAt(0)!;
+
+        traceInfo('Get diagnostics in test');
+        // Ask for the list of diagnostics
+        const diagnostics = await waitForDiagnostics(cell.document.uri);
+        assert.isOk(diagnostics.length);
+        assert.ok(diagnostics.find((item) => item.message.includes('system')));
+    });
+    test('Markdown cells dont get errors', async () => {
+        // Create two cells, one markdown, make sure not diagnostics in markdown
```
DonJayamanne, line 68: Nice.

src/client/datascience/notebook/intellisense/completionProvider.ts
rchiodo: Yeah sounds like a good idea. I'll add a debt issue

DonJayamanne (approved):

rchiodo:
@IanMatthewHuff and @DonJayamanne can you take a second look? 

I reworked how we select a kernel in the tests. It now loops on the notebook.selectKernel function. But the default it picks is lot more complicated now. Actually found a bug with this change: https://github.com/microsoft/vscode-jupyter/issues/7610

I also reworked how variables are waited for in tests. Instead of guessing the messages to be sent, I switched it to just use ```waitForCondition``` which should work regardless of how many messages get passed around.

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -153,10 +157,10 @@ export class VSCodeNotebookController implements Disposable {
     public async updateNotebookAffinity(notebook: NotebookDocument, affinity: NotebookControllerAffinity) {
         traceInfo(`Setting controller affinity for ${notebook.uri.toString()} ${this.id}`);
         this.controller.updateNotebookAffinity(notebook, affinity);
-        // Only when running tests should we force the selection of the kernel.
-        // Else the general VS Code behavior is for the user to select a kernel (here we make it look as though use selected it).
-        if (this.context.extensionMode === ExtensionMode.Test) {
-            await this.setAsActiveControllerForTests(notebook);
+
+        // Keep track of if this controller is preferred for this notebook
+        if (affinity === NotebookControllerAffinity.Preferred) {
```
IanMatthewHuff, line 162: updateNotebookAffinity is called on each document open if you do something like this:
Open foo.ipynb (A kernel in metadata) foo.ipynb added to A perferredWith
Change to kernel B in foo.ipynb
Execute
Close foo.ipynb
Open foo.ipynb (B kernel in metadata) foo.ipynb added to B preferredWith

Seems like that situation would be wrong then if I'm following right.

rchiodo: Not following the mistake? Because A would still be preferred? Yeah seems like map should be the other way. Map between notebook (key) and controller (value)

src/test/datascience/variableView/variableView.vscode.test.ts
```diff
@@ -102,17 +92,12 @@ suite('DataScience - VariableView', function () {
...
-
-        const htmlResult = await variableView?.getHTMLById('variable-view-main-panel');
-
         // Parse the HTML for our expected variables
         const expectedVariables = [
             { name: 'test', type: 'str', length: '11', value: ' MYTESTVALUE' },
             { name: 'test2', type: 'str', length: '12', value: ' MYTESTVALUE2' }
         ];
-        verifyViewVariables(expectedVariables, htmlResult);
+        await waitForVariablesToMatch(expectedVariables, variableView);
```
IanMatthewHuff, line 100: Nice, much better.

src/client/datascience/notebook/vscodeNotebookController.ts
IanMatthewHuff: Yeah, trying to make sure that I'm reading it right. As I see it now PreferredWith is all the notebook documents that this controller is the preferred controller for. Given that the preferred controller could change if you execute and reopen a document you could have the same document in the preferred list for two different controllers. As documents only have one preferred (at least from our extensions perspective) that situation wouldn't be right.

.github/workflows/build-test.yml
```diff
@@ -43,6 +43,8 @@ env:
   IPYWIDGET_SCREENSHOT_PATH: '*-screenshot.png'
   DISABLE_INSIDERS_EXTENSION: 1 # Disable prompts to install Insiders in tests (else it blocks activation of extension).
   VSC_JUPYTER_INSTRUMENT_CODE_FOR_COVERAGE: true
+  VSIX_NAME_PYTHON: 'ms-python-insiders.vsix'
+  VSC_JUPTYER_PYTHON_EXTENSION_VERSION: 'stable'
```
DonJayamanne, line 47: Does this mean, if this is stable, then we use Stable VS Code & stable Python extension
& if its insider we use vscode insiders & python insiders?

rchiodo: No it's just a flag for now. It makes it easy to switch this later.

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -629,7 +625,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             'interactive.open',
             { preserveFocus: true },
             notebookDocument.uri,
-            this.notebookController?.id,
+            this.notebookControllerManager.getSelectedNotebookController(notebookDocument)?.id,
```
DonJayamanne, line 628: I'm assuming these changes can be revereted now?
Would actually prefer to go back to the older API (if we can confirm the issues were with the Python extension)

rchiodo: The older api was saving the controller. This queries for it at run time. Seems more robust this way to me.

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -408,31 +404,4 @@ export class VSCodeNotebookController implements Disposable {
             await newKernel.start({ disableUI: true }).catch(noop);
         }
     }
-    /**
-     * In our tests, preferred controllers are setup as the active controller.
-     *
-     * This method is called on when running tests, else in the real world,
-     * users need to select a kernel (preferred is on top of the list).
-     */
-    private async setAsActiveControllerForTests(notebook: NotebookDocument) {
```
DonJayamanne, line 417: Not sure why thi was deleted, was this to do with getting tets to pass?

rchiodo: Because it doesn't work (notebook.selectKernel has to be called in a loop) and it masks other problems (it always associates the controller with the notebook even when it doesn't work).

I moved all of this same logic into the test helper.ts which does select in a loop.

src/client/datascience/interactive-window/interactiveWindow.ts
DonJayamanne: Aaah yes, users can switch the kernels, thats right.

src/client/datascience/notebook/vscodeNotebookController.ts
rchiodo: The new way seems more likely to find issues as it forces the UI selection change before the controller actually changes.

DonJayamanne: Dont' think tests will pass without this, as without this users will be prompted to select a kernel (this ensures the kernel picker isnt displayed in tests.)

src/client/datascience/notebook/types.ts
```diff
@@ -13,6 +13,7 @@ export interface INotebookControllerManager {
     // Marked test only, just for tests to access registered controllers
     registeredNotebookControllers(): VSCodeNotebookController[];
     getInteractiveController(): Promise<VSCodeNotebookController | undefined>;
+    getPreferredNotebookController(document: NotebookDocument): VSCodeNotebookController | undefined;
```
DonJayamanne, line 16: I don't think this is being used.

src/client/datascience/notebook/vscodeNotebookController.ts
rchiodo: No it does the same thing it used to do. It sends the command 'notebook.selectKernel' in a loop. This always happens in all tests before runnign a cell. The old code used to call this setAsActiveControllerForTests instead. Now it just waits for the UI to work.

src/client/datascience/notebook/types.ts
rchiodo: It's used in the helper.ts file to find the closest match for a kernel.

rchiodo: Basically the the tests that used to call waitForKernelToGetAutoSelected are now instead waiting for the preferred kernel to be auto selected (which effectively is the same as what happened before)

src/client/datascience/notebook/vscodeNotebookController.ts
rchiodo: Meaning the selection to change (not for the select kernel picker to come up). 'notebook.selectKernel' does not put up the picker if you pass an id.

src/test/datascience/notebook/helper.ts
```diff
@@ -287,89 +305,88 @@ export async function waitForKernelToChange(criteria: { labelOrId?: string; inte
...
-    await sleep(500);
+    if (!isRightKernel()) {
+        let tryCount = 0;
+        await waitForCondition(
+            async () => {
+                // Double check not the right kernel (don't select again if already found to be correct)
+                if (!isRightKernel()) {
+                    traceInfoIf(isCI, `Notebook select.kernel command switching to kernel id ${id}: Try ${tryCount}`);
+                    // Send a select kernel on the active notebook editor. Keep sending it if it fails.
+                    await commands.executeCommand('notebook.selectKernel', { id, extension: JVSC_EXTENSION_ID });
```
DonJayamanne, line 316: I dont think this needs to be ezecuted multiple times, i think once is sufficient, and after that we just need to keep waiting until the right kernel is selected.
Else we keep firing this command multkiple times & i'm concerned we could have other side effects of filing this unnecessarily

src/client/datascience/notebook/types.ts
DonJayamanne: Yes found that.

src/client/datascience/notebook/vscodeNotebookController.ts
DonJayamanne: Yup, found it. All good

src/test/datascience/notebook/helper.ts
rchiodo: That's not what I found. There was a potential race between 'createNotebookController (and results getting to VS code proper)' and the 'notebook.selectKernel' command coming in. This race went away when I did this in a loop.

rchiodo: Previously we didn't even care if it worked (hence the comment in the ```setAsActiveControllerForTests```). And we'd set and forget it. But sometimes it didn't work.

With this loop it seems to always work.

DonJayamanne (approved):

DonJayamanne (approved):
Looks good

DonJayamanne (approved):

IanMatthewHuff (approved):

Issue #7530, opened 2021-09-15
Cannot read property 'source' of undefined


roblourens:
When I run the RBL tests, I see this error

```
rejected promise not handled within 1 second: TypeError: Cannot read property 'source' of undefined
/Applications/Visual Studio Code - Insiders.app/Contents/Resources/app/out/bootstrap-fork.js:5
stack trace: TypeError: Cannot read property 'source' of undefined
    at callScopes (/Users/roblou/code/vscode-jupyter/src/client/datascience/jupyter/debuggerVariables.ts:253:32)
/Applications/Visual Studio Code - Insiders.app/Contents/Resources/app/out/bootstrap-fork.js:5
```

The DebuggerVariables code is throwing for a stackTrace response that has an empty stackFrames array

```
{"seq":27,"type":"response","request_seq":20,"success":true,"command":"stackTrace","body":{"stackFrames":[],"totalFrames":1}}
```

PR #7569, opened 2021-09-16
fix flakey tests

DavidKutu:
For #7530
checks if stack frame is undefined

### DIFF
--- a/src/client/debugger/jupyter/debugControllers.ts
+++ b/src/client/debugger/jupyter/debugControllers.ts
@@ -107,8 +107,12 @@ export class RunByLineController implements IDebuggingDelegate {
         // start the process of updating the variables view.
         const stResponse = await this.debugAdapter.stackTrace({ threadId, startFrame: 0, levels: 1 });
 
-        const sf = stResponse.stackFrames[0];
-        return !!sf.source && sf.source.path !== this.debugCell.document.uri.toString();
+        if (stResponse && stResponse.stackFrames[0]) {
+            const sf = stResponse.stackFrames[0];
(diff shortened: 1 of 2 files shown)


src/client/debugger/jupyter/debugControllers.ts
```diff
@@ -107,8 +107,12 @@ export class RunByLineController implements IDebuggingDelegate {
...
         const stResponse = await this.debugAdapter.stackTrace({ threadId, startFrame: 0, levels: 1 });
 
-        const sf = stResponse.stackFrames[0];
-        return !!sf.source && sf.source.path !== this.debugCell.document.uri.toString();
+        if (stResponse && stResponse.stackFrames[0]) {
+            const sf = stResponse.stackFrames[0];
+            return !!sf.source && sf.source.path !== this.debugCell.document.uri.toString();
+        }
+
+        return false;
```
DavidKutu, line 115: @roblourens my paranoid side thinks this logic might be backwards, and we might start getting random situations where RBL stops in other files.
If we say true on this cases, we stepIn on stop, and might randomly step 2 lines.

IDK which one is better,

roblourens (approved):
I think this should be fixed by https://github.com/microsoft/vscode-jupyter/pull/7527 (sorry I guess I didn't realize that when filing the issue). Do you still see the error? Previously we were listening to any stackTrace request, which might be requesting the bottom of the stack and getting an empty response. Now that DebuggerVariables sends the stackTrace request, we know that we are requesting 1 frame starting at 0 when we are paused, so there should always be a frame. It's probably still good to merge this. For example the test code could continue before this has a chance to run.

PR #7571, opened 2021-09-16
Remaining code cleanup 

DavidKutu:
-use a 'disconneced' flag to only send the message and event once
-only dump cells that executed succesfully

For #7346

### DIFF
--- a/src/client/debugger/jupyter/kernelDebugAdapter.ts
+++ b/src/client/debugger/jupyter/kernelDebugAdapter.ts
@@ -110,7 +110,8 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter, ID
                     // If a cell has moved to idle, stop the debug session
                     if (
                         this.configuration.__cellIndex === cellStateChange.cell.index &&
-                        cellStateChange.state === NotebookCellExecutionState.Idle
+                        cellStateChange.state === NotebookCellExecutionState.Idle &&
+                        !this.disconected
                     ) {
                         sendTelemetryEvent(DebuggingTelemetry.endedSession, undefined, { reason: 'normally' });
                         this.disconnect();
@@ -168,11 +169,9 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter, ID
     }
 
     public disconnect() {
-        if (!this.disconected) {
-            void this.session.customRequest('disconnect', { restart: false });
-            this.endSession.fire(this.session);
-            this.disconected = true;
-        }
+        void this.session.customRequest('disconnect', { restart: false });
+        this.endSession.fire(this.session);
+        this.disconected = true;
     }
 
     dispose() {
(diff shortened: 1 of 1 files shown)


src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -170,8 +171,11 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter, ID
     }
 
     public disconnect() {
-        void this.session.customRequest('disconnect', { restart: false });
-        this.endSession.fire(this.session);
+        if (!this.disconected) {
```
DavidKutu, line 174: helps calling disconnect only once. The risk is that the message might fail.

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -199,7 +203,7 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter, ID
 
     private dumpAllCells() {
         this.notebookDocument.getCells().forEach((cell) => {
-            if (cell.kind === NotebookCellKind.Code) {
+            if (cell.kind === NotebookCellKind.Code && cell.executionSummary?.success) {
```
DavidKutu, line 206: this avoids the errors in dumpCell responses.

roblourens: What does that have to do with dumpCell?

src/client/debugger/jupyter/kernelDebugAdapter.ts
roblourens: This doesn't help us understand why it is sent more than once or what the real correct time to send it is. We could check this in but I think it would deserve more investigation

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: I thought only the executed cells were successful, but no, all these requests are failing. I'll remove this.

src/client/debugger/jupyter/kernelDebugAdapter.ts
DavidKutu: I digged, and the reason we sometimes call it twice is that the cell completing always calls disconnect, even after a restart or interrupt. Here's a summary of all the cases:

-finishing, only the cell status calls disconnect
-clicking stop, the controller disconnects + the cell status
-interrupt, the interrupt + the status change
-restart, the restart + the status change

If I move this check to the status change, we ensure to only call disconnect once.

DonJayamanne (approved):

src/client/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -60,8 +60,6 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter, ID
         private fs: IFileSystem,
         private readonly kernel: IKernel | undefined
     ) {
-        void this.dumpAllCells();
```
roblourens, line 63: So in full debugging will I not be able to step into other cells?

DavidKutu: I tired it and it didn't make a difference, it always failed. Lets talk on Monday

DavidKutu: This PR fixes that:
https://github.com/microsoft/vscode-jupyter/pull/7591

Issue #7589, opened 2021-09-19
Running a markdown cell causes interactive window debugging to start opening temp files


greazer:
### Applies To

- [ ] Notebooks (.ipynb files)
- [X] Interactive Window and\/or Cell Scripts (.py files with \#%% markers)

### What happened?

- Open any py file with a ```#%% [markdown]``` cell and a regular code cell.
- Run the markdown cell normally
- Debug the code cell. 

A temporary file representing the code cell will be opened to debug through.

May be a dupe of #7544, but that repro is different, so it may be a separate issue. 

### VS Code Version

1.61-insiders

### Jupyter Extension Version

v2021.9.1001244851

PR #7593, opened 2021-09-20
Fix interactive window debugging sourcemap resolution after running a markdown cell

joyceerhl:
Fix https://github.com/microsoft/vscode-jupyter/issues/7589

### DIFF
--- a/src/client/datascience/editor-integration/cellhashprovider.ts
+++ b/src/client/datascience/editor-integration/cellhashprovider.ts
@@ -9,6 +9,7 @@ import {
     Event,
     EventEmitter,
     NotebookCell,
+    NotebookCellKind,
     Position,
     Range,
     TextDocumentChangeEvent,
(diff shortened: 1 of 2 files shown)


src/client/datascience/editor-integration/cellhashprovider.ts
```diff
@@ -106,6 +107,10 @@ export class CellHashProvider implements ICellHashProvider {
     }
 
     public async addCellHash(cell: NotebookCell) {
+        // Skip markdown cells as they are never actually executed
```
rchiodo, line 110: Should this skip raw cells too? As in make it only work for 'code' cells? I'm not entirely sure if we allow raw execution so maybe not.

rchiodo (approved):

src/client/datascience/editor-integration/cellhashprovider.ts
joyceerhl: We don't support raw cells in the interactive window afaik so this should be fine.

PR #7608, opened 2021-09-21
Remove Kernel selector

DonJayamanne:
This is old redundant code, it doesn't work corectly (very small number of users hit this scenario).


src/client/datascience/jupyter/jupyterExecution.ts
```diff
@@ -216,20 +216,9 @@ export class JupyterExecutionBase implements IJupyterExecution {
                             traceError('Failed to connect to server', ex);
                             if (ex instanceof JupyterSessionStartError && isLocalConnection && allowUI) {
                                 sendTelemetryEvent(Telemetry.AskUserForNewJupyterKernel);
-
-                                // Keep retrying, until it works or user cancels.
-                                const kernelInterpreter = await this.kernelSelector.askForLocalKernel(
-                                    options.resource,
-                                    connection,
-                                    launchInfo.kernelConnectionMetadata
+                                void this.kernelSelector.askForLocalKernel(
```
rchiodo, line 219: This is going to throw even though you asked for another kernel. Shouldn't it break?

src/client/datascience/jupyter/kernels/kernelSelector.ts
```diff
@@ -28,19 +21,16 @@ import { Uri } from 'vscode';
...
 export class KernelSelector {
-    private readonly kernelSelectionsByResource = new Map<Uri, Promise<KernelConnectionMetadata | undefined>>();
     constructor(
-        @inject(KernelSelectionProvider) private readonly selectionProvider: KernelSelectionProvider,
         @inject(IApplicationShell) private readonly applicationShell: IApplicationShell,
-        @inject(IConfigurationService) private configService: IConfigurationService,
-        @inject(InterpreterPackages) private readonly interpreterPackages: InterpreterPackages
-    ) {}
+        @inject(ICommandManager) private readonly commandManager: ICommandManager,
+        @inject(IServiceContainer) protected serviceContainer: IServiceContainer // @inject(IInteractiveWindowProvider) private readonly interactiveWindowProvider: IInteractiveWindowProvider
```
IanMatthewHuff, line 27: This line looks off here.

src/client/datascience/jupyter/jupyterExecution.ts
DonJayamanne: I think its ok for the message to be displayed the startup failed. Else the rest of the code might not work correctly

src/client/datascience/jupyter/kernels/kernelSelector.ts
DonJayamanne: Looks like a cyclic reference,
when we refactor and remove these layers this should all go away, my Pr is merely a hacky solution

rchiodo (approved):

Issue #7586, opened 2021-09-18
RBL starts on second line or doesn't start at all


roblourens:
I can't repro this in a simpler cell, only in the golden scenario notebook:

https://github.com/claudiaregio/data-science/

- Execute all the cells up to the one that starts `#Summarize history of accuracy`
- Start RBL on this cell
- Run all the way through it. At the end, immediately click to start another RBL session
- It doesn't start RBL or show the debug highlight
- Click again
- It starts on the second line

I think it did start in the background but didn't get all the wires connected

Haven't looked through this yet
[log.txt](https://github.com/microsoft/vscode-jupyter/files/7188946/log.txt)

A different thing that happens sometimes is that the debug session never ends, clicking RBL starts a second one but we never actually start running and show the debug line. Burke hit this in the golden scenario run. 

![image](https://user-images.githubusercontent.com/323878/133866941-5bce2f9a-2b6d-407f-b129-49a331bf9590.png)
[log2.txt](https://github.com/microsoft/vscode-jupyter/files/7188951/log2.txt)

Issue #7584, opened 2021-09-18
Breaking RBL by clicking quickly


roblourens:
Do RBL on this cell:

![image](https://user-images.githubusercontent.com/323878/133866223-77f23a57-1649-47a2-9903-e82655aa51ba.png)

Click through it quickly, and get this error
![image](https://user-images.githubusercontent.com/323878/133866197-374491ee-b22f-4303-b563-9ee3666f5038.png)

Get some errors:

![image](https://user-images.githubusercontent.com/323878/133866249-8eac3ca6-35e6-4c88-88de-b14e51a85450.png)
![image](https://user-images.githubusercontent.com/323878/133866257-29fbc6da-f4ef-41aa-ac87-e71f3e6dfebe.png)

roblourens:
I still get this error:

![image](https://user-images.githubusercontent.com/323878/134602385-e7676373-d991-4017-bc43-4464598b67b2.png)

I don't see any real user impact, but we should check it out to make sure it's not a sign of something going wrong

DonJayamanne:
Unable to repro anymore.

PR #7612, opened 2021-09-21
prevent debugpy from getting stuck

DavidKutu:
-prevent debugpy from getting stuck by running code
-if we can't continue in RBL, disconnect

For #7586, #7584 and parts 1 and 7 of #7218

### DIFF
--- a/package.json
+++ b/package.json
@@ -228,8 +228,7 @@
                 "command": "jupyter.runAndDebugCell",
                 "title": "%jupyter.command.jupyter.debugCell.title%",
                 "icon": "$(debug-alt-small)",
-                "category": "Jupyter",
-                "enablement": "!jupyter.notebookeditor.runByLineInProgress"
+                "category": "Jupyter"
             },
             {
                 "command": "jupyter.runByLineNext",
@@ -866,17 +865,12 @@
                     "command": "jupyter.runByLineStop",
                     "when": "notebookType == jupyter-notebook && jupyter.ispythonnotebook && notebookCellType == code && isWorkspaceTrusted && jupyter.notebookeditor.runByLineInProgress && notebookCellToolbarLocation == right",
                     "group": "inline/cell@0"
-                },
-                {
-                    "command": "jupyter.runAndDebugCell",
-                    "when": "notebookType == jupyter-notebook && jupyter.ispythonnotebook && notebookCellType == code && isWorkspaceTrusted && !config.notebook.consolidatedRunButton",
-                    "group": "inline/cell@1"
                 }
             ],
             "notebook/cell/execute": [
                 {
                     "command": "jupyter.runAndDebugCell",
-                    "when": "notebookType == jupyter-notebook && jupyter.ispythonnotebook && notebookCellType == code && isWorkspaceTrusted && config.notebook.consolidatedRunButton"
+                    "when": "notebookType == jupyter-notebook && jupyter.ispythonnotebook && notebookCellType == code && isWorkspaceTrusted && !jupyter.notebookeditor.runByLineInProgress"
                 }
             ],
             "interactive/toolbar": [
--- a/package.json
+++ b/package.json
@@ -228,7 +228,8 @@
                 "command": "jupyter.runAndDebugCell",
                 "title": "%jupyter.command.jupyter.debugCell.title%",
                 "icon": "$(debug-alt-small)",
-                "category": "Jupyter"
+                "category": "Jupyter",
+                "enablement": "!jupyter.notebookeditor.runByLineInProgress"
             },
             {
                 "command": "jupyter.runByLineNext",
@@ -865,12 +866,17 @@
                     "command": "jupyter.runByLineStop",
                     "when": "notebookType == jupyter-notebook && jupyter.ispythonnotebook && notebookCellType == code && isWorkspaceTrusted && jupyter.notebookeditor.runByLineInProgress && notebookCellToolbarLocation == right",
                     "group": "inline/cell@0"
+                },
+                {
+                    "command": "jupyter.runAndDebugCell",
+                    "when": "notebookType == jupyter-notebook && jupyter.ispythonnotebook && notebookCellType == code && isWorkspaceTrusted && !config.notebook.consolidatedRunButton",
+                    "group": "inline/cell@1"
                 }
             ],
             "notebook/cell/execute": [
                 {
                     "command": "jupyter.runAndDebugCell",
-                    "when": "notebookType == jupyter-notebook && jupyter.ispythonnotebook && notebookCellType == code && isWorkspaceTrusted && !jupyter.notebookeditor.runByLineInProgress"
+                    "when": "notebookType == jupyter-notebook && jupyter.ispythonnotebook && notebookCellType == code && isWorkspaceTrusted && config.notebook.consolidatedRunButton"
                 }
             ],
             "interactive/toolbar": [
--- a/src/client/debugger/jupyter/debugControllers.ts
+++ b/src/client/debugger/jupyter/debugControllers.ts
@@ -57,6 +57,7 @@ export class RunByLineController implements IDebuggingDelegate {
     public continue(): void {
         if (typeof this.lastPausedThreadId !== 'number') {
             traceVerbose(`No paused thread, can't do RBL`);
+            this.stop();
             return;
         }
 
@@ -64,6 +65,7 @@ export class RunByLineController implements IDebuggingDelegate {
     }
 
     public stop(): void {
+        void executeDebugPyFix(this.kernel);
         this.debugAdapter.disconnect();
     }
 
@@ -171,12 +173,15 @@ async function cellDebugSetup(
     debugAdapter: IKernelDebugAdapter,
     debugCell: NotebookCell
 ): Promise<void> {
+    await executeDebugPyFix(kernel);
+    await debugAdapter.dumpCell(debugCell.index);
+}
+
+async function executeDebugPyFix(kernel: IKernel) {
     // remove this if when https://github.com/microsoft/debugpy/issues/706 is fixed and ipykernel ships it
     // executing this code restarts debugpy and fixes https://github.com/microsoft/vscode-jupyter/issues/7251
     if (kernel) {
         const code = 'import debugpy\ndebugpy.debug_this_thread()';
         await kernel.executeHidden(code);
     }
-
-    await debugAdapter.dumpCell(debugCell.index);
 }
--- a/src/client/debugger/jupyter/debugControllers.ts
+++ b/src/client/debugger/jupyter/debugControllers.ts
@@ -65,7 +65,8 @@ export class RunByLineController implements IDebuggingDelegate {
     }
 
     public stop(): void {
-        void executeDebugPyFix(this.kernel);
+        // When debugpy gets stuck, running a cell fixes it and allows us to start another debugging session
+        void this.kernel.executeHidden('pass');
         this.debugAdapter.disconnect();
     }
 
@@ -173,15 +174,10 @@ async function cellDebugSetup(
     debugAdapter: IKernelDebugAdapter,
     debugCell: NotebookCell
 ): Promise<void> {
-    await executeDebugPyFix(kernel);
-    await debugAdapter.dumpCell(debugCell.index);
-}
-
-async function executeDebugPyFix(kernel: IKernel) {
     // remove this if when https://github.com/microsoft/debugpy/issues/706 is fixed and ipykernel ships it
     // executing this code restarts debugpy and fixes https://github.com/microsoft/vscode-jupyter/issues/7251
-    if (kernel) {
-        const code = 'import debugpy\ndebugpy.debug_this_thread()';
-        await kernel.executeHidden(code);
-    }
+    const code = 'import debugpy\ndebugpy.debug_this_thread()';
+    await kernel.executeHidden(code);
+
+    await debugAdapter.dumpCell(debugCell.index);
 }
--- a/src/client/debugger/jupyter/debuggingManager.ts
+++ b/src/client/debugger/jupyter/debuggingManager.ts
@@ -40,7 +40,6 @@ import { PythonEnvironment } from '../../pythonEnvironments/info';
 import { DebugCellController, RunByLineController } from './debugControllers';
 import { assertIsDebugConfig } from './helper';
 import { Debugger } from './debugger';
-import { throttle } from 'lodash';
 
 /**
  * The DebuggingManager maintains the mapping between notebook documents and debug sessions.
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -865,12 +866,17 @@
...
+                    "command": "jupyter.runAndDebugCell",
+                    "when": "notebookType == jupyter-notebook && jupyter.ispythonnotebook && notebookCellType == code && isWorkspaceTrusted && !config.notebook.consolidatedRunButton",
+                    "group": "inline/cell@1"
                 }
             ],
             "notebook/cell/execute": [
                 {
                     "command": "jupyter.runAndDebugCell",
-                    "when": "notebookType == jupyter-notebook && jupyter.ispythonnotebook && notebookCellType == code && isWorkspaceTrusted && !jupyter.notebookeditor.runByLineInProgress"
+                    "when": "notebookType == jupyter-notebook && jupyter.ispythonnotebook && notebookCellType == code && isWorkspaceTrusted && config.notebook.consolidatedRunButton"
```
roblourens, line 879: Why this change? It always goes in that menu regardless of the setting by design. Probably will rename the setting or rethink this but I don't want it on the cell toolbar.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -92,102 +93,119 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
...
-                    const editor = this.vscNotebook.activeNotebookEditor;
-                    const range = editor?.selections[0];
-                    if (range) {
-                        cell = editor?.document.cellAt(range.start);
+                }, 1000)
+            ),
+
+            this.commandManager.registerCommand(
+                DSCommands.RunByLineNext,
+                throttle((cell: NotebookCell | undefined) => {
```
roblourens, line 145: Why throttle this? When I am doing normal javascript debugging, I can step as fast as I want. I can even hold down F10 to step really fast. The same should work here.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -92,102 +93,119 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
...
-            this.commandManager.registerCommand(DSCommands.RunAndDebugCell, async (cell: NotebookCell | undefined) => {
-                sendTelemetryEvent(DebuggingTelemetry.clickedRunAndDebugCell);
-                const editor = this.vscNotebook.activeNotebookEditor;
-                if (!cell) {
-                    const range = editor?.selections[0];
-                    if (range) {
-                        cell = editor?.document.cellAt(range.start);
+            this.commandManager.registerCommand(
+                DSCommands.RunAndDebugCell,
+                throttle(async (cell: NotebookCell | undefined) => {
```
roblourens, line 183: I worry it will feel broken if this command sometimes doesn't work if stuff happens too quickly

rchiodo: Yeah throttle of 1 second seems like way too much. Maybe like 100ms

package.json
DavidKutu: why not? 

I moved it because some people will probably want the setting to get rid of the drop down on the run button

roblourens: I'm not going to add another button to every cell toolbar unless masses of angry users kick down my door

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: Because that seems to be the issue, clicking fast enough causes debugpy to freeze.

The changes to the controller are what gets it out of the situation (finishes that frozen session and un freezes debugpy).
The throttling is meant to prevent the situation.

Its removable, but I'd rather have at least some.

roblourens: It's hard to believe that debugpy just can't step fast. I can debug a normal python script and step over 50 lines in a couple seconds

DonJayamanne: I agree, I think we need to identify the root cause, this feels like a bandaid

src/client/debugger/jupyter/debugControllers.ts
```diff
@@ -57,13 +57,15 @@ export class RunByLineController implements IDebuggingDelegate {
...
             traceVerbose(`No paused thread, can't do RBL`);
+            this.stop();
             return;
         }
 
         void this.debugAdapter.stepIn(this.lastPausedThreadId);
     }
 
     public stop(): void {
+        void executeDebugPyFix(this.kernel);
```
DonJayamanne, line 68: Since comments would be good here, didn't make sense why this is required

src/client/debugger/jupyter/debugControllers.ts
```diff
@@ -171,12 +173,15 @@ async function cellDebugSetup(
...
     debugCell: NotebookCell
 ): Promise<void> {
+    await executeDebugPyFix(kernel);
+    await debugAdapter.dumpCell(debugCell.index);
+}
+
+async function executeDebugPyFix(kernel: IKernel) {
     // remove this if when https://github.com/microsoft/debugpy/issues/706 is fixed and ipykernel ships it
     // executing this code restarts debugpy and fixes https://github.com/microsoft/vscode-jupyter/issues/7251
     if (kernel) {
```
DonJayamanne, line 183: This is always not undefined

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -92,102 +93,119 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
...
-            this.commandManager.registerCommand(DSCommands.DebugNotebook, async () => {
-                const editor = this.vscNotebook.activeNotebookEditor;
-                if (editor) {
-                    if (await this.checkForIpykernel6(editor.document)) {
-                        this.updateToolbar(true);
-                        void this.startDebugging(editor.document);
+            this.commandManager.registerCommand(
+                DSCommands.DebugNotebook,
+                throttle(async () => {
+                    const editor = this.vscNotebook.activeNotebookEditor;
```
DonJayamanne, line 99: I think the solution is, if we're already debugging or have started then return from here.
I'd if in progress then get out, mashed three throttle unnecessary.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -92,102 +93,119 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
...
+                    if (!cell) {
+                        return;
+                    }
 
-                if (editor) {
-                    if (await this.checkForIpykernel6(editor.document, DataScience.startingRunByLine())) {
-                        this.updateToolbar(true);
-                        this.updateCellToolbar(true);
-                        await this.startDebuggingCell(editor.document, KernelDebugMode.RunByLine, cell);
+                    if (editor) {
```
DonJayamanne, line 129: Same comment as earlier

src/client/debugger/jupyter/debuggingManager.ts
DonJayamanne: I think they problem is not because it happens fast, it's because we start multiple debug sessions instead of checkingbif we have already started one

DonJayamanne (changes requested):
We don't need the throttle

src/client/debugger/jupyter/debugControllers.ts
roblourens: I'm also curious about this, we execute this code when we start, why do we have to execute it when we stop?

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -95,6 +95,9 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
             this.commandManager.registerCommand(DSCommands.DebugNotebook, async () => {
                 const editor = this.vscNotebook.activeNotebookEditor;
                 if (editor) {
+                    if (this.notebookToDebugger.has(editor.document)) {
```
DonJayamanne, line 98: Not sure this is going to be enough.
If user clicks this button while `checkforIpykernel6` is still running, then yo'ull end up with duplicate sessions.

DonJayamanne: I think you can move this entire if condition from line 98 to 109 into a seprate fucntion, this has been repeated in 3 places, and then you can easily check if that method has already been called once before if `yes`, then return.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -232,6 +212,50 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
         this.runByLineInProgress.set(runningByLine).ignoreErrors();
     }
 
+    private async tryToStartDebugging(mode: KernelDebugMode, editor?: NotebookEditor, cell?: NotebookCell) {
+        if (editor) {
+            if (this.notebookInProgress.has(editor.document)) {
```
roblourens, line 217: This is checking the case where debugging is starting, but doesn't it still need to check `this.notebookToDebugger.has`?

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -243,6 +208,50 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
...
+                return;
```
DonJayamanne, line 218: I think returning early such as 

```
if (!editor) {
	return;
}
```

& a few of thse will improve readabiliyty of this method, right now its nested way too much.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -243,6 +208,50 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
...
+                            if (cell) {
+                                this.updateToolbar(true);
+                                void this.startDebuggingCell(editor.document, KernelDebugMode.Cell, cell);
+                            }
+                            break;
+                        case KernelDebugMode.RunByLine:
+                            if (cell) {
+                                this.updateToolbar(true);
+                                this.updateCellToolbar(true);
+                                await this.startDebuggingCell(editor.document, KernelDebugMode.RunByLine, cell);
```
DonJayamanne, line 244: in line 233 we're not awaiting, but here we're awaiting.
Confused why that's the case, comments would help.

DavidKutu: It'll be good to await them all

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -232,6 +212,51 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
...
+            ) {
+                switch (mode) {
+                    case KernelDebugMode.Everything:
+                        this.updateToolbar(true);
+                        await this.startDebugging(editor.document);
+                        break;
+                    case KernelDebugMode.Cell:
+                        if (cell) {
+                            this.updateToolbar(true);
+                            await this.startDebuggingCell(editor.document, KernelDebugMode.Cell, cell);
```
DonJayamanne, line 241: I'm sorry, but I'm not convinced we need `awaits` here
The old code didn't have awaits, feels like there's something we're not understanding here 
I.e. why do we need an await or what happens if there is no await?

Would like to have some comments why we must await, e.g. tomorrow what would stop working if we removed it.

roblourens:
It took longer (this is after many attempts) but I can still break it

Sep-22-2021 18-37-12

DavidKutu:
@roblourens, the latest commit should catch your scenario

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -218,7 +218,15 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
             return;
         }
 
-        if (this.notebookInProgress.has(editor.document) || this.isDebugging(editor.document)) {
+        if (this.notebookInProgress.has(editor.document)) {
+            return;
+        }
+
+        if (this.isDebugging(editor.document)) {
+            this.updateToolbar(true);
```
roblourens, line 226: Why do you have to do this here? If you are already debugging, shouldn't it already be in this state?

DavidKutu: it should, but that's the buggy state you get trapped in.

this would correct the context keys

DonJayamanne: Not sure i like this, feels like we're re-executing some code & that seems to fix a problem (but we don't know why).
I would like us to fix the root cause, why is this required in the first place

DavidKutu: The root cause is out of our hands, its debugpy freezing inside ipykernel.

And I disagree, we should be trying to fix root causes, but since the root cause is not on our code, we should also try to fix it if it happens.

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: it helps with keeping the context keys aligned with the status of the debugger. My most recent commit puts them after the awaits.

src/client/debugger/jupyter/debuggingManager.ts
DonJayamanne: > its debugpy freezing inside ipykernel.

If that's the cause, how can updating the toolbar fix anything in the kernel (i.e. updating a UI will not address anything in ipykernel freezing).

roblourens: It seems like the root cause is us having the context keys out of sync with the RBL lifecycle. Your latest commit is a good one towards fixing that, I think previously we reset the context keys too early, before debugging actually stopped.

Fundamentally, at this point in the code, two things are out of sync. It's hard to point at one and say this one is always right. It should be possible to write the code so that they are always in sync.

DonJayamanne (commented):
.

DonJayamanne (approved):
Removing the unblocking request.

DonJayamanne (commented):
.

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -241,19 +241,25 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
...
-                        this.updateToolbar(true);
+                        if (this.isDebugging(editor.document)) {
+                            this.updateToolbar(true);
+                        }
                         break;
                     case KernelDebugMode.Cell:
                         if (cell) {
                             await this.startDebuggingCell(editor.document, KernelDebugMode.Cell, cell);
-                            this.updateToolbar(true);
+                            if (this.isDebugging(editor.document)) {
```
DonJayamanne, line 251: Under what circustances will this `if` condition be `false`, i don't think this can ever be `false`, its aleways `true`, it should be true, else the previous await should throw an error.
Having this `if` makes it look as though the previous code can run to completion without errors & not start the debugger (which i think is impossible)

src/client/debugger/jupyter/debuggingManager.ts
```diff
@@ -241,19 +241,25 @@ export class DebuggingManager implements IExtensionSingleActivationService, IDeb
...
+                                this.updateToolbar(true);
+                            }
                         }
                         break;
                     case KernelDebugMode.RunByLine:
                         if (cell) {
                             await this.startDebuggingCell(editor.document, KernelDebugMode.RunByLine, cell);
-                            this.updateToolbar(true);
-                            this.updateCellToolbar(true);
+                            if (this.isDebugging(editor.document)) {
```
DonJayamanne, line 259: As earlier, i don't think this if condition is required.

src/client/debugger/jupyter/debuggingManager.ts
DavidKutu: I guess that's true, it just feels safer.
I'll remove them

src/client/debugger/jupyter/debugControllers.ts
```diff
@@ -57,13 +57,16 @@ export class RunByLineController implements IDebuggingDelegate {
...
+            this.stop();
             return;
         }
 
         void this.debugAdapter.stepIn(this.lastPausedThreadId);
     }
 
     public stop(): void {
+        // When debugpy gets stuck, running a cell fixes it and allows us to start another debugging session
+        void this.kernel.executeHidden('pass');
```
DonJayamanne, line 69: Isn't this a hack, i think we should get to the bottom of this.

DavidKutu: I'll file a bug in ipykernel, and mention it here, but I don't see what else we can do

src/client/debugger/jupyter/debugControllers.ts
```diff
@@ -57,13 +57,16 @@ export class RunByLineController implements IDebuggingDelegate {
     public continue(): void {
         if (typeof this.lastPausedThreadId !== 'number') {
             traceVerbose(`No paused thread, can't do RBL`);
+            this.stop();
```
roblourens, line 60: With this, if I click the RBL button and then click it again before the first pause, it disconnects. That's not that bad of a problem but could be annoying, especially since it will then run the whole cell, even if I didn't want/expect it to. 

I guess this is working around a case where we attach to debugpy and it doesn't respond, but if we have fixed the issue on our end of trying to attach twice at a time, then I don't understand how else we get into that case

src/client/debugger/jupyter/debugControllers.ts
roblourens: > When debugpy gets stuck

It would be good to explain how that happens

roblourens:
Played with it for awhile and was not able to totally break it, nice!

Issue #6389, opened 2021-06-28
All python kernelspecs installed to global locations are hidden when python extension is installed


IanMatthewHuff:
Seems to have been from the kernel search refactor changes. You can install kernelspecs all in the same interpreter. They could have different launch args or different environments. Syncing with the recent changes these kernelspecs are all hidden and unavailable in the kernel picker.

Example install:
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/7c3556c1-a36b-48ed-8960-ff4e4fcf6412)

Listed by Jupyter:
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/74906214-f4a9-4b0f-9bb3-2b0c2bf457c4)

Not listed in VS Code:
![image.png](https://images.zenhubusercontent.com/5bdcd40a0faa800c749bac98/041aeebf-292b-412e-b13e-de2c24fb620f)

Note: I was just testing this before pulling the changes so I'm sure they used to be listed.

IanMatthewHuff:
We chatted a bit about this on teams quick. Feel like there is more here than we would rather deal with in release.

Issues for discussion:
1. Currently global python kernelspecs are hidden if you have the python extension installed. This is not right as you could have a globally installed python kernelspec that specifically customizes the environment. This previously was available, but is now hidden. Perhaps not super likely, but I have one of these that I use.
2. The current code calls out not returning global python kernelspecs as they are ambiguous as to the python environment to run them. 
3. However if you create a named kernelspec from --user (https://ipython.readthedocs.io/en/stable/install/kernel_install.html#kernels-for-different-environments) it will install globally but have the full interpreter path in the argv. Our current code can match this up with existing interpreters so the environment is known. 
4. Correct thing here I believe would be not to even care about if it's a default kernel spec or not, but instead to return all global python kernelspecs that can be identified to a python environment that we have.
5. Second point to check (and a concern when pulling in global pythons) it looks like when we find a kernelspec we hide any interpreters that match the path. If you have a global kernelspec with a custom environment you would not want to hide the interpreter as it's distinct from the interpreter: https://github.com/microsoft/vscode-jupyter/blob/617fccc68807d9055bf42fa26f05b3f96636cbb2/src/client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder.ts#L152

greazer:
[...]

PR #7623, opened 2021-09-22
Allow for global Python specs which use IPyKernel to be displayed.

IanMatthewHuff:
For #6389 


src/client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder.ts
```diff
@@ -149,7 +128,7 @@ export class LocalPythonAndRelatedNonPythonKernelSpecFinder extends LocalKernelS
         // Go through the global kernelspecs that use python to launch the kernel but don't use ipykernel.
         await Promise.all(
             globalKernelSpecs
```
IanMatthewHuff, line 130: So this check for ipykernel_launcher on global python spec was what was hiding my kernelspecs. This kernel finding code has gotten a bit twisty, so I want to make sure that I'm not regression something with this change. But I can't see a reason why this check was here before. I can't see a reason to hide these specs from the user.

src/client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder.ts
```diff
@@ -89,9 +89,6 @@ export class LocalPythonAndRelatedNonPythonKernelSpecFinder extends LocalKernelS
         );
     }
     /**
-     * If user has python extension installed, then we'll not list any of the globally registered Python kernels.
```
IanMatthewHuff, line 92: This line here (which already exists) seemed like the important check:
https://github.com/microsoft/vscode-jupyter/blob/3e6f5860a238b2a340efd317649d6d8ec3f0129e/src/client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder.ts#L137

If we can't match up an interpreter then we do hide the global specs, but if we are able to match up the interpreter then the is no reason to hide.

IanMatthewHuff:
Bleh...should probably have a test for this as well. Lemme see if there is a reasonable way to add one.

src/client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder.ts
rchiodo: The reason was global ipykernel specs- one's that match an interpreter should be skipped because the 'interpreter' kernelspec would find them.

rchiodo: We don't want doubles.

IanMatthewHuff: By "interpreter" kernelspec do you mean the kernelspecs that we create for each interpreter? If so that doesn't work, it's very easy to have multiple global kernelspecs with different names and environments pointing at the same interpreter.

rchiodo: They can have different names but what if they're the same. From the user's point of view they can't tell which one is which (as the python path is the same) and they're functionally equivalent (well unless the env is different for each).

IanMatthewHuff: The different envs is exactly it. My repro has global kernelspecs with different envs pointing at the same interpreter.

DonJayamanne: I think a better way to handle this is to display global python kernel specs ONLY if they are not craeted by us and if they have custom env variables.

That way we avoid duplicates & go back to the previous approach (with the exception of dipslaying global kernelspecs with custom env variables)

DonJayamanne (changes requested):
I think just showing kernelspecs with custom env variables (not created by us would be sufficient)

src/client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder.ts
IanMatthewHuff: @DonJayamanne I believe that would work, but checking to make sure now. That would be pretty similar to what the non-global kernels is already doing here:
https://github.com/microsoft/vscode-jupyter/blob/b7241b5b97c9e0ff4ab7d4aa26c0d1d0b9a643e0/src/client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder.ts#L204

DonJayamanne: Aaah yes, that's correct that should be sufficient.

DonJayamanne (approved):

src/client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder.ts
DonJayamanne: Actually, I think we should modify the code to return all python kernels except those with custom env, just like the code you highlighted.
Else i don't think it will work.
Will wait for your confirmation.

DonJayamanne (changes requested):
Actually, I think we should modify the code to return all python kernels except those with custom env, just like the code you highlighted.
Else i don't think it will work.
Will wait for your confirmation (blocking on the confirmation).

src/client/datascience/kernel-launcher/localPythonAndRelatedNonPythonKernelSpecFinder.ts
IanMatthewHuff: I was just playing with the condition now for global python spec inclusion as:
!isKernelRegisteredByUs && (usingNonIpyKernel || customEnv)  // plus it needs to match an interpreter
For the globalKernelSpecs

This works for my scenario, but it does reverse what is happening in the non global case that I pointed out. Basically the question in my head is if it matters if the kernel spec hides the interpreter as kernel or visa-versa (plus I was starting to add a test for this scenario).

DonJayamanne: > Basically the question in my head is if it matters if the kernel spec hides the interpreter as kernel 

I don' tthink it should, 
Also I'd expect the names to be different (i.e. when users create the names, I do'nt expect the names to be exactly the same as the display names we use for interpreters, which contians python name, version, etc)

IanMatthewHuff: I feel like there is a lot of "de-duplication" stuff going on for the kernel spec finding, but it's a bit spread out all over the place. I'm almost tempted to consider making all the finding stuff just return flat lists, then have a final single deduplication step to sort it all out. That would be a pretty big change though, and I don't like the idea of perturbing this now that it's in pretty good shape now.

DonJayamanne: > I'm almost tempted to consider making all the finding stuff just return flat lists, then have a final single deduplication step to sort it all out

Sounds great, right now it does feel like the code could use some cleanup

> nd I don't like the idea of perturbing this now that it's in pretty good shape now.

Meaning, this is complicated code, i think we should try to clean it up in debt or when ever, if we're concerned about making changes, that just re-iterates the fact that its not in the best shape & not easy to maintain.

But totally fine not refactoring it now.

IanMatthewHuff: Our kernel finding has improved a ton recently, just a complicated problem, as you mentioned. I just find myself thinking "I'm preeeeetty sure this won't break anything" whenever I'm making changes in this area. Low confidence about regressions or side effects.

rchiodo: +1 on the filter in one spot. Sounds like a good refactor to me too. Not sure I'd do it now but maybe next iteration.

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -291,7 +291,7 @@ export function isKernelRegisteredByUs(kernelSpec: IJupyterKernelSpec): 'oldVers
         return 'newVersion';
     }
     const guidRegEx = /[a-f0-9]{32}$/;
-    if (kernelSpec.metadata?.interpreter && kernelSpec.name.toLowerCase().search(guidRegEx)) {
+    if (kernelSpec.metadata?.interpreter && kernelSpec.name.toLowerCase().search(guidRegEx) !== -1) {
```
IanMatthewHuff, line 294: This bug that I fixed was messing up the tests, but I believe that it wasn't causing actual product code issues as I think that only our extension puts the interpreter into the metadata. So even though the guid check was always returning truthy (-1 != false) it wasn't flagging anything as registered by us that wasn't registered by us.

rchiodo (approved):

DonJayamanne (approved):

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -79,7 +79,7 @@ export function getKernelId(spec: IJupyterKernelSpec, interpreter?: PythonEnviro
         // eslint-disable-next-line @typescript-eslint/no-use-before-define
         specName = specName.substring(specName.indexOf(autoGeneratedKernelNameIdentifier));
     }
-    // The arguments in the spec can be different for multkiple kernel specs pointing to the same interpreter.
+    // The arguments in the spec can be different for multiple kernel specs pointing to the same interpreter.
```
DonJayamanne, line 82: Removing all evidence of me having written this :(

PR #7640, opened 2021-09-23
Turn insiders back on

rchiodo:
After this fix yesterday:
https://github.com/microsoft/vscode-python/pull/17499

Python Insiders should work again. 

This PR is trying that out (and enabling the intellisense tests again)


IanMatthewHuff (approved):

src/test/common.ts
```diff
@@ -139,10 +139,11 @@ async function setGlobalPathToInterpreter(pythonPath?: string): Promise<void> {
     await pythonConfig.update('defaultInterpreterPath', pythonPath, true);
     await disposePythonSettings();
 }
-export async function disableExperimentsInPythonExtension(): Promise<void> {
+export async function adjustSettingsInPythonExtension(): Promise<void> {
     const vscode = require('vscode') as typeof import('vscode');
     const pythonConfig = vscode.workspace.getConfiguration('python', (null as any) as Uri);
-    await pythonConfig.update('experiments.enabled', false, vscode.ConfigurationTarget.Global).then(noop, noop);
+    await pythonConfig.update('logging.level', 'debug', vscode.ConfigurationTarget.Global);
```
DonJayamanne, line 145: This shoudln't be necessary, we've set an environment variable at the top of the YAML file.

rchiodo: This is forcing it for python, not jupyter.

DonJayamanne: Wait, I added the env variable in our YAML, all you need is `VSC_PYTHON_FORCE_LOGGING` and that enabled debugigng logging, simpler than updating this setting.

DonJayamanne: Hmm, can't find that, i'm certain I added that, anyways

DonJayamanne: I'd suggest adding that along with out setting `VSC_JUPYTER_FORCE_LOGGING: 'true'`

src/test/datascience/helpers.ts
```diff
@@ -21,7 +21,7 @@ export function defaultDataScienceSettings(): IJupyterSettings {
             level: 'off'
         },
         experiments: {
-            enabled: false,
+            enabled: true,
```
DonJayamanne, line 24: This should be off,
on CI, everythign should be predicatable.
We don't want a feature turning on one CI and not the other.

DonJayamanne (changes requested):
We should turn off experiments for jupyter extension on CI

rchiodo:
Uggh python extension is failing to load now.

src/test/standardTest.ts
```diff
@@ -92,6 +92,8 @@ async function createSettings(): Promise<string> {
         'python.insidersChannel': 'off',
         'jupyter.logging.level': 'debug',
         'python.logging.level': 'debug',
+        'python.experiments.enabled': 'true',
+        'python.experiments.optOutFrom': '[]',
```
DonJayamanne, line 96: Isnt' this supposed to be a regular array
```suggestion
        'python.experiments.optOutFrom': [],
```

rchiodo: Maybe that was causing the extension crash. Thanks.

src/client/api/pythonApi.ts
```diff
@@ -352,16 +357,7 @@ export class InterpreterService implements IInterpreterService {
...
-                    const itemToStore = items;
-                    if (
-                        current &&
-                        (itemToStore === current || JSON.stringify(itemToStore) === JSON.stringify(current))
-                    ) {
-                        return;
-                    }
-                    interpreterCacheForCI.set(resource?.toString() || '', itemToStore);
-                    traceInfo(`Interpreter list for ${resource?.toString()} is ${JSON.stringify(items)}`);
+                    traceInfo(`Interpreter list for ${resource?.toString()} is ${items.map((i) => i.path).join('\n')}`);
```
DonJayamanne, line 360: Wait, isn't this a bug now
If the first call returns empty and second call returns non-empty that's a bug
I guess that's what you're trying to figure out right now.

src/test/initialize.ts
```diff
@@ -25,6 +26,11 @@ process.env.VSC_JUPYTER_CI_TEST = '1';
 // Ability to use custom python environments for testing
 export async function initializePython() {
     await setPythonPathInWorkspaceRoot(PYTHON_PATH);
+    // Make sure the python extension can load if this test allows it
+    if (!process.env.VSC_JUPYTER_CI_TEST_DO_NOT_INSTALL_PYTHON_EXT) {
+        const extension = vscode.extensions.getExtension(PythonExtension)!;
+        await extension.activate();
```
DonJayamanne, line 32: We shouldn't do this, we should let python extension get activated normally in code.
I.e. if its installed and we need it, we activate it (in jupytre ext) & if not installed, then we don't activate it.
This new code will force the python extension to activate before jupyter ext launches and we don't want that, as its not a realistic situation (could hide other bugs).

Happy to leave this until we figure out whats going on with the tests though.

rchiodo: Yeah I did this because I wanted to make sure that wasn't causing the problem. Happy to move it back after I get it all passing

src/client/api/pythonApi.ts
rchiodo: Yeah wanted to make sure this wasn't masking stuff. It was weird that it was caching only during CI.

DonJayamanne: Ther'es no caching, if you check this out, its only for logging purposes.
I.e. the cache is only used for logging, not used in run time

rchiodo: Ah right, it's just using the cache to skip logging.

rchiodo: I actually want it to log every time to make sure we log each call. That might indicate the second call fails when the first passes.

DonJayamanne: You might want to gate this logging with `traceInfoIf`
It should be smart to log only the first time (as we need this logging on user machines)
But do'nt want all subsequent calls to `getInterpreters` else we end up logging way too much (i think this gets logged in the output channel, and adds a lot of noise, as its a massive blob of JSON)

rchiodo: It's already under an if (isCI) statement for the logging.

And I changed it to just log the path. That seems sufficient.

DonJayamanne (approved):

src/client/api/pythonApi.ts
```diff
@@ -425,7 +406,6 @@ export class InterpreterService implements IInterpreterService {
                     api.onDidChangeInterpreter(
                         () => {
                             this.didChangeInterpreter.fire();
-                            this.workspaceCachedActiveInterpreter.clear();
```
DonJayamanne, line 428: At one point Python extension was not caching this, and calling this Python extension API was pretty expensive, are they caching this now?
If not it could adversely impact our code (extension).

rchiodo: Not sure, trying to figure out why the active interpeter seems to not match the ```python.pythonPath``` in some of the tests. I can put this back later if I can figure out why.

rchiodo:
Note to self:
- Remove custom insiders download
- Put back active interpreter cache

Issue #7305, opened 2021-08-26
Add table of contents icon to notebook toolbar


claudiaregio:
We have still been seeing consistent feedback in the notebooks survey that table of contents for notebooks is hard to find (people are still asking for it in the survey as a new feature and mentioning it is frustrating not having one).

Since this has been very consistent feedback we should consider adding a table of contents icon to the notebook toolbar. (Icon request reference here: https://github.com/microsoft/vscode-codicons/issues/80)

PR #7651, opened 2021-09-23
Basic TOC button for jupyter notebooks

IanMatthewHuff:
For #7305

### DIFF
--- a/package.json
+++ b/package.json
@@ -692,6 +692,14 @@
                 "enablement": "notebookType == jupyter-notebook  && !jupyter.variableViewVisible && isWorkspaceTrusted || notebookType == interactive && !jupyter.variableViewVisible && isWorkspaceTrusted"
             },
             {
+                "command": "jupyter.openOutlineView",
+                "title": "%jupyter.command.jupyter.openOutlineView.title%",
+                "shortTitle": "%jupyter.command.jupyter.openOutlineView.shorttitle%",
+                "icon": "$(list-unordered)",
+                "category": "Jupyter",
+                "enablement": "notebookType == jupyter-notebook && !view.outline.visible"
+            },
+            {
                 "command": "jupyter.refreshDataViewer",
                 "title": "%DataScience.refreshDataViewerCommandPalette%",
                 "category": "Jupyter",
@@ -813,9 +821,15 @@
                     "when": "notebookType == 'jupyter-notebook' && isWorkspaceTrusted && config.notebook.globalToolbar != true"
                 },
                 {
+                    "command": "jupyter.openOutlineView",
+                    "title": "%jupyter.command.jupyter.openOutlineView.title%",
+                    "group": "navigation@3",
+                    "when": "notebookType == 'jupyter-notebook' && isWorkspaceTrusted && config.notebook.globalToolbar != true"
+                },
+                {
                     "command": "jupyter.notebookeditor.export",
                     "title": "%DataScience.notebookExportAs%",
-                    "group": "navigation@3",
+                    "group": "navigation@4",
                     "when": "notebookType == 'jupyter-notebook' && isWorkspaceTrusted && config.notebook.globalToolbar != true"
                 },
                 {
@@ -839,6 +853,11 @@
                 {
                     "command": "jupyter.openVariableView",
                     "group": "navigation@1",
+                    "when": "notebookType == 'jupyter-notebook' && isWorkspaceTrusted"
+                },
+                {
+                    "command": "jupyter.openOutlineView",
+                    "group": "navigation@2",
                     "when": "notebookType == 'jupyter-notebook' && isWorkspaceTrusted"
                 },
                 {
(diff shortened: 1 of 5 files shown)


rchiodo:
I thought there was going to be a setting to turn it on/off?

DonJayamanne:
Isn't there a limit to the enumber of icons we can have in the global toolbar?
Hope this doesn't result in some of our other icons being hidden (going into overflow)

IanMatthewHuff:
A couple of notes here.
1. I made the button only enabled if the outline view is not visible and expanded. I could change this to just be visible if the notebook is open at all. Felt better to me this way, but I'm open to having it always available. Just feels funny as the button doesn't do much then, just moved the keyboard focus over.
2. I called the short name (shown on the button) 'Outline' as it seemed shorter and more concise than "Table of Contents" long name in the palette is Jupyter: Open Notebook Outline
3. Right now the button is available for any notebooks, but this doesn't mean that they have anything in the TOC, so it's not as useful for notebooks like this (see below):
image
I could engineer this to be disabled in these cases. The bit that seems iffy to me is that this outline content is based on vscode core code, we'd have to figure out a way to check if the outline view actually has content or not in a context key. I think that it requires markdown headers right now, but we'd be relying on that not changing if we were to base visibility on that. So I left that off for now.

DonJayamanne (approved):

IanMatthewHuff:
@rchiodo Totally forgot the setting, I'll update with that added.

IanMatthewHuff:
@claudiaregio and @greazer  for UX visibility on this review.

IanMatthewHuff:
> Isn't there a limit to the enumber of icons we can have in the global toolbar?
> Hope this doesn't result in some of our other icons being hidden (going into overflow)

Yeah, I get ya. On my monitor it's not bad, as it's right next to the overflow button.
image
image

Could be an issue, but I am adding the setting to hide.

greazer:
> 1. I made the button only enabled if the outline view is not visible and expanded. I could change this to just be visible if the notebook is open at all. Felt better to me this way, but I'm open to having it always available. Just feels funny as the button doesn't do much then, just moved the keyboard focus over.

So the behavior you have right now is pretty much identical to that of the Variables Window right? Basically, if either window isn't visible and usable, then the button is enabled. If it's already being shown on the screen in a usable way, then it's disabled. 

> 2. I called the short name (shown on the button) 'Outline' as it seemed shorter and more concise than "Table of Contents" long name in the palette is Jupyter: Open Notebook Outline

Label seems ok to me. The main thing here is to make it more discoverable at all. Might want to consider creating a "wrapper" command called "Jupyter: Show Table of Contents (Outline View)" that is identical to "Explorer: Focus on Outline View"

> 3. Position is always right of Variables icon.

Any particular reason? My quick thinking is that more people have likely asked for TOC than Variables Window. Ensuring that it's more likely to not be relegated to overflow seems like a good idea. BUT I'm not sure it matters much.

> 4. Right now the button is available for any notebooks, but this doesn't mean that they have anything in the TOC, so it's not as useful for notebooks like this (see below):

The user can turn on code to be shown in the outline view. "Notebook: Outline: Show Code Cells", so it's not like this would always be the case for notebooks without markdown. Seems ok to me.

greazer (approved):
See my comments above. Otherwise looks fine to me.

IanMatthewHuff:
@greazer Inline.
> So the behavior you have right now is pretty much identical to that of the Variables Window right? Basically, if either window isn't visible and usable, then the button is enabled. If it's already being shown on the screen in a usable way, then it's disabled. Consistency here is the right way to go.

Yup, if outline is off screen, or onscreen but collapsed the button is enabled. If it's on screen and expanded disable the button. Seemed smart as if it was on screen it was hard to even see that anything had happened when you clicked the button (only moved keyboard focus) made it seem a bit broken.

> Label seems ok to me. The main thing here is to make it more discoverable at all. Might want to consider creating a "wrapper" command called "Jupyter: Show Table of Contents (Outline View)" that is identical to "Explorer: Focus on Outline View"

So the long title does show up in the command palette currently. Currently it's "Jupyter: Open Notebook Outline" but your text does have Table of Contents in it, so that might be a good change. I can change that.

> Any particular reason? My quick thinking is that more people have likely asked for TOC than Variables Window. Ensuring that it's more likely to not be relegated to overflow seems like a good idea. BUT I'm not sure it matters much.

I was seeing it the other way around (variables over TOC) but I don't have data to back that up aside from our old requests for variable views being very upvoted. Mild preference would be to keep it how I have it as it add the "new" icon to the least dominate position so "Variable" will be in the same location for users muscle memory. 

> The user can turn on code to be shown in the outline view. "Notebook: Outline: Show Code Cells", so it's not like this would always be the case for notebooks without markdown. Seems ok to me.

Yeah, it seemed like it would open a bigger can of worms to try to check that.

claudiaregio:
@IanMatthewHuff @greazer @DonJayamanne  Behavior here seems fine to me as well. I'm wondering now though if we think it makes sense now to have the toolbar default be icon only as opposed to with text? This is one of the few places we have to contribute to notebooks and for smaller monitors/laptops we are likely to end up in the overflow menu thus not solving the discoverability issue.

IanMatthewHuff:
@claudiaregio I've always been pro small icons, but if I recall from earlier discussions there are some pros to either side. Might be something to bring up at the next ux meeting maybe? Since it's an all up notebook issue.

Issue #4576, opened 2021-01-30
Always use LF instead of CRLF for *.ipynb files regardless of platform and when sending code to the kernel


claforte:
## Environment data

-   VS Code version: 1.53.0-insider (2021-01-29T05:13:48.533Z)
-   Jupyter Extension version (available under the Extensions sidebar): v2020.12.414227025
-   Python Extension version (available under the Extensions sidebar): 2021.2.518958983-dev
-   OS (Windows | Mac | Linux distro) and version: Windows 10
-   Python and/or Anaconda version: Python 3.8.5 using Anaconda
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): conda
-   Jupyter server running: Local

## Expected behaviour

When editing a notebook, lines ending in "\n" shouldn't be modified to "\r\n"

## Actual behaviour

It appears as soon as I modify the content of a cell, the editor is replacing all "\n" by "\r\n". I don't understand why, and it's causing downstream tools (e.g. nbdev/nbparse) to no longer be able to parse the .ipynb properly.

## Steps to reproduce:

This seems to happen for any .ipynb file on my PC. I did a git clone of nbdev (`https://github.com/fastai/nbdev/blob/master/nbs/00_export.ipynb`) and ran the notebook, but chances are you only need to open a file:

1. download https://github.com/fastai/nbdev/blob/master/nbs/00_export.ipynb
2. open the file in vscode
3. modify a cell
4. do a diff on the cell.
5. notice that the newlines are messed up.
![image](https://user-images.githubusercontent.com/5090329/106348587-5e7f9200-6295-11eb-83b6-cdadb7a11de0.png)

## Logs

<details>

<summary>Output for <code>Jupyter</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Jupyter</code>)
</summary>

```
User belongs to experiment group 'pythonJoinMailingListVar1'
User belongs to experiment group 'jupyterTest'
User belongs to experiment group 'NativeNotebookEditor'
> D:\Users\claforte\Anaconda3\envs\fastai\python.exe c:\Users\claforte\.vscode-insiders\extensions\ms-toolsai.jupyter-2020.12.414227025\pythonFiles\pyvsc-run-isolated.py -c "import jupyter"
[...]

PR #7659, opened 2021-09-24
Use LF instead of CRLF, when run against kernel

DonJayamanne:
For #4576

I need to test this on a windows machine, will start on getting the windows pc up.

### DIFF
--- a/src/client/datascience/jupyter/jupyterNotebook.ts
+++ b/src/client/datascience/jupyter/jupyterNotebook.ts
@@ -498,7 +498,7 @@ export class JupyterNotebookBase implements INotebook {
                 ? this.session.requestExecute(
                       {
                           // Remove the cell marker if we have one.
-                          code: cellMatcher.stripFirstMarker(code.replace(/\\r\\n/g, '\n')),
+                          code: cellMatcher.stripFirstMarker(code.replace(/\r\n/g, '\n')),
                           stop_on_error: false,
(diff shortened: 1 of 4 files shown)


rchiodo (approved):

src/client/datascience/jupyter/jupyterNotebook.ts
```diff
@@ -498,7 +498,7 @@ export class JupyterNotebookBase implements INotebook {
                 ? this.session.requestExecute(
                       {
                           // Remove the cell marker if we have one.
-                          code: cellMatcher.stripFirstMarker(code),
+                          code: cellMatcher.stripFirstMarker(code.replace(/\r\n/g, '\n')),
```
IanMatthewHuff, line 501: These usages look like the only three usages of session.requestExecute. So should this replace instead go into BaseJupyterSession.requestExecute? That would put it in only one place, and prevent someone else from adding a new usage of requestExecute that re-creates this issue.

DonJayamanne: I don't want to refactor the code, ideally we should have just one.
When the old Jupyter layer goes away we'll automatically have just one.
The other two will be in kernel stuff & i'd like to remove the executeSilently to go through kernel as well (but don't want to introduce a refactoring now)

Hopefully we'll get to some of that in debt week.

IanMatthewHuff (approved):

Issue #7648, opened 2021-09-23
"Go to [#]" codelenses don't work for Interactive Window


greazer:
### Applies To

- [ ] Notebooks (.ipynb files)
- [X] Interactive Window and\/or Cell Scripts (.py files with \#%% markers)

### What happened?

1. Start an interactive session by running a #%% cell.
2. Click on the 'Go to [1]' codelens

Nothing happens. It should highlight the cell in the interactive window and give it focus.
![image](https://user-images.githubusercontent.com/11685408/134590438-d68406c7-1dc3-4470-b288-c885591ad81b.png)

### VS Code Version

v1.61-insiders

### Jupyter Extension Version

v2021.9.1001261101

DonJayamanne:
Lol was fixed in the last release as a regression, looks like we've regressed again after the point fix https://github.com/microsoft/vscode-jupyter/issues/7347
This is bad

DonJayamanne:
> Scrolling seems to work. There just isn't a highlight

Aah, I want aware of that

rchiodo:
I think we should rethink this too. It's really hokey the highlight.

rchiodo:
https://github.com/microsoft/vscode-jupyter/issues/7687

Issue #7658, opened 2021-09-24
SyntaxError by sending a function with triple-quoted string/docsting to Interactive Window


firegate22:
## Environment data

-   VS Code version: 1.61.0
-   Jupyter Extension version (available under the Extensions sidebar):  2021.9.1001269173
-   Python Extension version (available under the Extensions sidebar): 2021.9.1246542782
-   OS (Windows | Mac | Linux distro) and version: windows/mac + linux remote
-   Python and/or Anaconda version: 3.7 or 3.9
-   Type of virtual environment used pyenv virtualenv and/or conda 
-   Jupyter server running: Remote

## Expected behaviour

```
def test(input: str = "test"):
    """
    test function
    """
    print("hallo")
```

get's executed correctly

## Actual behaviour

```
def test(input: str = "test"):
    test function
    print("hallo")

test()
```

is the code send to the interactive window, resulting in an error. 

It basically removes """ without anything else in the line I guess. 
e.g. 
"""test
"""
results in
"""test

other example:
```
def test1():
    """TEST
    
    TEST"""

def test2():
    """TEST
    TEST
    TEST"""
```

results in:

```
def test1():
    """TEST
    TEST"""

def test2():
    TEST
    TEST"""
```

## Steps to reproduce:

```
def test(input: str = "test"):
    """
    test function
    """
    print("hallo")
```

in .py file

mark the code, press shift enter (send to interactive console)

this works:
```
def test(input: str = "test"):
    """test function"""
    print("hallo")

test()
```

</p>
</details>

rchiodo:
Should be fixed in next insiders bits.

firegate22:
Thanks, yes - it's fixed!

PR #7685, opened 2021-09-27
Fix two issues with interactive window

rchiodo:
Fixes: #7648, #7658

### DIFF
--- a/src/client/datascience/interactive-window/interactiveWindow.ts
+++ b/src/client/datascience/interactive-window/interactiveWindow.ts
@@ -23,7 +23,8 @@ import {
     ViewColumn,
     NotebookEditor,
     Disposable,
-    window
+    window,
+    ThemeColor
 } from 'vscode';
 import { IPythonExtensionChecker } from '../../api/types';
 import {
@@ -424,7 +425,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         // of the history. The jupyter.alwaysScrollOnNewCell setting overrides this to always scroll
         // to newly-inserted cells.
         if (settings.alwaysScrollOnNewCell || isLastCellVisible) {
-            this.revealCell(notebookCell, notebookEditor);
+            this.revealCell(notebookCell, notebookEditor, false);
         }
 
         const notebook = kernel?.notebook;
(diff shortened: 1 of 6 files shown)


src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -510,15 +511,31 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
             notebookEditor.revealRange(notebookRange, NotebookEditorRevealType.Default);
+
+            // Also add a decoration to make it look highlighted (peek background color)
+            if (decorationType) {
+                notebookEditor.setDecorations(decorationType, notebookRange);
+
+                // Fire another timeout to dispose of the decoration
+                setTimeout(() => {
+                    decorationType.dispose();
+                }, 2000);
```
DonJayamanne, line 537: 1. Hmm, feels hackis, what if i select another cell or run multiple cells.
Now we'd have multiples with the same style.

2. Why are we disposing this after 2s, why don't we just add this to the list of disposables or another list and clear that (dispose it) when we reveal another cell.

DonJayamanne:
Not sold on the cell highlighting, 
will the highlithing go away when users click on the cells or select another cell?

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -510,15 +511,31 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
+            }
         }, 200); // Rendering output is async so the output is not guaranteed to immediately exist
```
DonJayamanne, line 539: I think we should fix this, filed a bug https://github.com/microsoft/vscode-jupyter/issues/7686

src/client/datascience/interactive-window/interactiveWindow.ts
rchiodo: That's the behavior from before. Just replicating it. It flashed for 2 seconds.

rchiodo:
> Not sold on the cell highlighting,
> will the highlithing go away when users click on the cells or select another cell?

Highlighting will go away after 2 seconds.

DonJayamanne:
> Highlighting will go away after 2 seconds.

Why not 5s or 10s or until user runs another cell or directly interacts with interactive window?

DonJayamanne:
> That's the way it used to work. Not changing the old behavior.

Ok, thats bad. From what you're saying we've some how removed some code that did this same thing early this month.
Sorry, i coudln't find any code or css that was releated in ths PR, hence assumed this was new & some other part of the PR was existing behaviour.

rchiodo:
> > That's the way it used to work. Not changing the old behavior.
> 
> Ok, thats bad. From what you're saying we've some how removed some code that did this same thing early this month.
> Sorry, i coudln't find any code or css that was releated in ths PR, hence assumed this was new & some other part of the PR was existing behaviour.

That code was in the webview CSS/implementation.

rchiodo:
Here's a link to the original code. Sorry I was wrong, it was only a second.

https://github.com/microsoft/vscode-jupyter/blob/f5c89293163be4c90341104ae4aa151610027d98/src/datascience-ui/history-react/interactiveCell.tsx#L86

Issue #7740, opened 2021-09-29
Run by line shouldn't be allowed if kernel is busy. And vice-versa.


DonJayamanne:
Run a long running cell or run the entire notebook and then start run by line.
We need to ensure we handle these cases.

Since run by line will send code directly to the kernel and bypass the cell execution queue.
E.g. if you have 10 cells and run all, then for the last cell we hit `RBL`, then the last cell could start debugging immediately after the first cell has completed.

I think we need to add a method `executeHidden` into the KernelExecution class so that it uses the queue & debugging starts after all queued cells have been completed.
Should be an easy enough fix, else we end up in a very weird spot.

roblourens:
This is basically what's behind https://github.com/microsoft/vscode-jupyter/issues/7638, as far as i can tell

greazer:
Dupe of #7638

roblourens:
Seems to be properly queued now

PR #7763, opened 2021-09-30
execute hidden code with the executionQueue

DavidKutu:
For #7740

### DIFF
--- a/src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
+++ b/src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
@@ -7,7 +7,14 @@ import { noop } from '../../../common/utils/misc';
 import { traceCellMessage } from '../../notebook/helpers/helpers';
 import { INotebook } from '../../types';
 import { CellExecution, CellExecutionFactory } from './cellExecution';
+import { CodeExecution, CodeExecutionFactory } from './codeExecution';
 import { KernelConnectionMetadata, NotebookCellRunState } from './types';
+import type { nbformat } from '@jupyterlab/coreutils';
+
+interface Execution {
+    cellExecution?: CellExecution;
+    codeExecution?: CodeExecution;
+}
 
 /**
  * A queue responsible for execution of cells.
@@ -15,14 +22,14 @@ import { KernelConnectionMetadata, NotebookCellRunState } from './types';
  * All cells queued using `queueCell` are added to the queue and processed in order they were added/queued.
  */
 export class CellExecutionQueue {
-    private readonly queueOfCellsToExecute: CellExecution[] = [];
+    private readonly queueToExecute: Execution[] = [];
     private cancelledOrCompletedWithErrors = false;
     private startedRunningCells = false;
     /**
      * Whether all cells have completed processing or cancelled, or some completed & others cancelled.
      */
     public get isEmpty(): boolean {
-        return this.queueOfCellsToExecute.length === 0;
+        return this.queueToExecute.length === 0;
     }
     /**
      * Whether cells have been cancelled (as a result of interrupt or some have failed).
@@ -33,22 +40,31 @@ export class CellExecutionQueue {
     }
     constructor(
         private readonly notebookPromise: Promise<INotebook>,
-        private readonly executionFactory: CellExecutionFactory,
+        private readonly cellExecutionFactory: CellExecutionFactory,
+        private readonly codeExecutionFactory: CodeExecutionFactory,
         readonly metadata: Readonly<KernelConnectionMetadata>
     ) {}
     /**
      * Queue the cell for execution & start processing it immediately.
      */
-    public queueCell(cell: NotebookCell): void {
-        const existingCellExecution = this.queueOfCellsToExecute.find((item) => item.cell === cell);
-        if (existingCellExecution) {
-            traceCellMessage(cell, 'Use existing cell execution');
-            return;
-        }
-        const cellExecution = this.executionFactory.create(cell, this.metadata);
-        this.queueOfCellsToExecute.push(cellExecution);
+    public queueCell(cell?: NotebookCell, code?: string): void {
+        if (cell) {
+            const cellQueue = this.getCellExecutions(this.queueToExecute);
+            const existingCellExecution = cellQueue.find((item) => item.cell === cell);
+            if (existingCellExecution) {
+                traceCellMessage(cell, 'Use existing cell execution');
+                return;
+            }
+            const cellExecution = this.cellExecutionFactory.create(cell, this.metadata);
+            this.queueToExecute.push({ cellExecution });
 
-        traceCellMessage(cell, 'User queued cell for execution');
+            traceCellMessage(cell, 'User queued cell for execution');
+        } else if (code) {
+            const codeExecution = this.codeExecutionFactory.create(code);
+            this.queueToExecute.push({ codeExecution });
+
+            traceInfo('Hidden cell queued for execution', codeExecution.code.substring(0, 50));
+        }
 
         // Start executing the cells.
         this.startExecutingCells();
@@ -64,7 +80,15 @@ export class CellExecutionQueue {
     public async cancel(forced?: boolean): Promise<void> {
         this.cancelledOrCompletedWithErrors = true;
         traceInfo('Cancel pending cells');
-        await Promise.all(this.queueOfCellsToExecute.map((item) => item.cancel(forced)));
+        await Promise.all(
+            this.queueToExecute.map((item) => {
+                if (item.cellExecution) {
+                    item.cellExecution.cancel(forced);
+                } else {
+                    item.codeExecution?.cancel(forced);
+                }
+            })
+        );
     }
     /**
      * Wait for cells to complete (for for the queue of cells to be processed)
@@ -72,12 +96,21 @@ export class CellExecutionQueue {
      * If no cells are provided, then wait on all cells in the current queue.
      */
     public async waitForCompletion(cells?: NotebookCell[]): Promise<NotebookCellRunState[]> {
+        const queue: CellExecution[] = this.getCellExecutions(this.queueToExecute);
         const cellsToCheck =
-            Array.isArray(cells) && cells.length > 0
-                ? this.queueOfCellsToExecute.filter((item) => cells.includes(item.cell))
-                : this.queueOfCellsToExecute;
+            Array.isArray(cells) && cells.length > 0 ? queue.filter((item) => cells.includes(item.cell)) : queue;
 
         return Promise.all(cellsToCheck.map((cell) => cell.result));
+    }
+    public async waitForHiddenOutput(code: string): Promise<nbformat.IOutput[]> {
+        const queue: CodeExecution[] = this.getCodeExecutions(this.queueToExecute);
+        const execution = queue.find((exec) => exec.code === code);
+
+        if (execution) {
+            return Promise.resolve(execution.output);
+        }
+
+        return [];
     }
     private startExecutingCells() {
         if (!this.startedRunningCells) {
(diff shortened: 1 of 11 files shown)


src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -164,10 +164,12 @@ export class Kernel implements IKernel {
         await promise;
         return promise;
     }
-    public async executeHidden(code: string): Promise<nbformat.IOutput[]> {
+    public async executeHidden(code: string, doc: NotebookDocument): Promise<nbformat.IOutput[]> {
         const stopWatch = new StopWatch();
         const notebookPromise = this.startNotebook();
-        const promise = notebookPromise.then((nb) => executeSilently(nb.session, code));
+
+        const promise = this.kernelExecution.executeHidden(notebookPromise, code, doc);
```
DavidKutu, line 171: should we do this for all the uses of `executeSilently`?

DonJayamanne: No we should not, this has the potential to break variable viewer and others.
I think we'd should have two methods.
And we might want to rename the executeCell to queueAndExecuteCell, 
Similarly executeHidden and queueAndExeceuteHidden

src/client/datascience/jupyter/kernels/types.ts
```diff
@@ -147,7 +147,7 @@ export interface IKernel extends IAsyncDisposable {
     interrupt(): Promise<InterruptResult>;
     restart(): Promise<void>;
     executeCell(cell: NotebookCell): Promise<NotebookCellRunState>;
-    executeHidden(code: string): Promise<nbformat.IOutput[]>;
+    executeHidden(code: string, doc: NotebookDocument): Promise<nbformat.IOutput[]>;
```
DonJayamanne, line 150: We don't need to pass the document

DonJayamanne: It's already available in the class

DavidKutu: We need it [here](https://github.com/microsoft/vscode-jupyter/blob/9eaa2cecebf0e9616af795527488b524f089606b/src/client/datascience/jupyter/kernels/kernelExecution.ts#L87) to get the executionQueue

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -33,22 +40,31 @@ export class CellExecutionQueue {
...
      */
-    public queueCell(cell: NotebookCell): void {
-        const existingCellExecution = this.queueOfCellsToExecute.find((item) => item.cell === cell);
-        if (existingCellExecution) {
-            traceCellMessage(cell, 'Use existing cell execution');
-            return;
-        }
-        const cellExecution = this.executionFactory.create(cell, this.metadata);
-        this.queueOfCellsToExecute.push(cellExecution);
+    public queueCell(cell?: NotebookCell, code?: string): void {
```
DonJayamanne, line 50: This api shows me to pass two empty arguments.
It should instead be `code : NotebookCell | string`
This way it's impossible to pass undefined and it's obvious what's expected without having to look at the code

DonJayamanne: I think it's better to return a promise from this method, then we won't need waitForHidden.
Else you have two methods to deal with.

src/client/datascience/jupyter/kernels/types.ts
DavidKutu: I don't see it in the class, there's `documentExecutions` which uses documents as keys, but there's no document in the class

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -7,22 +7,29 @@ import { noop } from '../../../common/utils/misc';
...
 import { INotebook } from '../../types';
 import { CellExecution, CellExecutionFactory } from './cellExecution';
+import { CodeExecution, CodeExecutionFactory } from './codeExecution';
 import { KernelConnectionMetadata, NotebookCellRunState } from './types';
+import type { nbformat } from '@jupyterlab/coreutils';
+
+interface Execution {
+    cellExecution?: CellExecution;
+    codeExecution?: CodeExecution;
+}
```
DonJayamanne, line 17: This means we can have a cake with both properties as empty, which is not correct.
We should change this to `type Execution = Cell execution | CodeExecution`

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -33,22 +40,31 @@ export class CellExecutionQueue {
...
+            this.queueToExecute.push({ cellExecution });
 
-        traceCellMessage(cell, 'User queued cell for execution');
+            traceCellMessage(cell, 'User queued cell for execution');
+        } else if (code) {
+            const codeExecution = this.codeExecutionFactory.create(code);
+            this.queueToExecute.push({ codeExecution });
+
+            traceInfo('Hidden cell queued for execution', codeExecution.code.substring(0, 50));
+        }
```
DonJayamanne, line 68: We have missed an else clause and that results in bugs, the suggestion I made will ensure typescript types will prevent such bugs

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -102,24 +135,42 @@ export class CellExecutionQueue {
...
             // Remove from the queue only after we process it
             // This way we don't accidentally end up queueing the same cell again (we know its in the queue).
-            const cellToExecute = this.queueOfCellsToExecute[0];
-            traceCellMessage(cellToExecute.cell, 'Before Execute individual cell');
+            const toExecute = this.queueToExecute[0];
+            if (toExecute.cellExecution) {
+                traceCellMessage(toExecute.cellExecution.cell, 'Before Execute individual cell');
+            } else if (toExecute.codeExecution) {
+                traceInfo('Before Execute hidden code', toExecute.codeExecution.code.substring(0, 50));
+            }
```
DonJayamanne, line 154: We have missed an else clause and that results in bugs, the suggestion I made will ensure typescript types will prevent such bugs

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: We'd should not because if we have 10 cells and we're thinking all, then variables will never refresh because those hidden requests will get queued and run only after all 10 cells finish. Similarly plenty of other pitfalls.
Never this new method should be banned to make it obvious how the code will be executed, ie the fact that it's queued.

src/client/datascience/jupyter/kernels/types.ts
DonJayamanne: Check line 113 of kernel.ts file

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -64,21 +80,38 @@ export class CellExecutionQueue {
     public async cancel(forced?: boolean): Promise<void> {
         this.cancelledOrCompletedWithErrors = true;
         traceInfo('Cancel pending cells');
-        await Promise.all(this.queueOfCellsToExecute.map((item) => item.cancel(forced)));
+        await Promise.all(
+            this.queueToExecute.map((item) => {
+                if (item.cellExecution) {
+                    void item.cellExecution.cancel(forced);
+                } else {
+                    void item.codeExecution?.cancel(forced);
```
DonJayamanne, line 88: We must await, that's why we have promise.all

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -64,21 +80,38 @@ export class CellExecutionQueue {
...
+        const queue: CellExecution[] = this.getCellExecutions(this.queueToExecute);
         const cellsToCheck =
-            Array.isArray(cells) && cells.length > 0
-                ? this.queueOfCellsToExecute.filter((item) => cells.includes(item.cell))
-                : this.queueOfCellsToExecute;
+            Array.isArray(cells) && cells.length > 0 ? queue.filter((item) => cells.includes(item.cell)) : queue;
 
         return Promise.all(cellsToCheck.map((cell) => cell.result));
     }
+    public async waitForHiddenOutput(code: string): Promise<nbformat.IOutput[]> {
```
DonJayamanne, line 105: This won't work if we have multiple expositions with the same code, e.g. `pass` or the like..
It will return the result from the first matching Execution with the same code

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -64,21 +80,38 @@ export class CellExecutionQueue {
...
+        const queue: CodeExecution[] = this.getCodeExecutions(this.queueToExecute);
+        const execution = queue.find((exec) => exec.code === code);
+
+        if (execution) {
+            return Promise.resolve(execution.output);
```
DonJayamanne, line 110: Promise.resolve is unnecessary

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -33,25 +37,44 @@ export class CellExecutionQueue {
...
      */
-    public queueCell(cell: NotebookCell): void {
-        const existingCellExecution = this.queueOfCellsToExecute.find((item) => item.cell === cell);
-        if (existingCellExecution) {
-            traceCellMessage(cell, 'Use existing cell execution');
-            return;
-        }
-        const cellExecution = this.executionFactory.create(cell, this.metadata);
-        this.queueOfCellsToExecute.push(cellExecution);
+    public queueCell(code: NotebookCell | string): Promise<nbformat.IOutput[]> {
```
DavidKutu, line 47: I couldn't figure out how to return both types (`Promise<NotebookCellRunState[] | nbformat.IOutput[]>`) here and then use `instanceof nbformat.IOutput[]` on `kernelExecution.ts`.

If you do, let me know @DonJayamanne

DonJayamanne: I'd created two methods

src/client/debugger/jupyter/helper.ts
```diff
@@ -7,7 +7,7 @@ import { IKernelDebugAdapterConfig, KernelDebugMode } from '../types';
 
 export async function isUsingIpykernel6OrLater(kernel: IKernel): Promise<boolean> {
     const code = 'import ipykernel\nprint(ipykernel.__version__)';
-    const output = await kernel.executeHidden(code);
+    const output = await kernel.queueAndExeceuteHidden(code);
```
DonJayamanne, line 10: This should not be queued, only the debugger start method should be queued, all other (older) methods that used hidden Execution b should not have to be queued

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
IanMatthewHuff: Yeah, and as is the naming is quite awkward here at least in my view. It's called queueCell, but it takes a parameter called 'code' now not cell. And then the paths below are harder to read and easy to miss whens on. @DonJayamanne 's suggestion is good. If it was one method it would be better with a more explicit name like queueExecution, then casting early to either CodeExecution or CellExecution so that the code after that is easier to read, but two methods sounds like a legit idea.

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -132,4 +171,28 @@ export class CellExecutionQueue {
             }
         }
     }
+
+    private getCellExecutions(executionList: Execution[]): CellExecution[] {
```
IanMatthewHuff, line 175: Not a perf difference I believe, but this could look cleaner just using .filter

src/client/datascience/jupyter/kernels/codeExecution.ts
```diff
@@ -0,0 +1,176 @@
...
+    ) {}
+
+    public create(code: string) {
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
+        return CodeExecution.fromCode(code, this.errorHandler, this.disposables);
+    }
+}
+
+/**
+ * Responsible for execution of an individual cell and manages the state of the cell as it progresses through the execution phases.
```
IanMatthewHuff, line 27: This is picky, but that's kinda how I am 😆 . The comments here look copied over from the other class, they should be updated for the new class (i.e. this is not responsible for execution of an individual cell).

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
DonJayamanne: I agree, personally we don't need these two methods, i'd just move the filtering into the place where we call this method, else we just iterate throught the same set of list twice.

src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
rchiodo: +1 on a separate method

src/client/datascience/jupyter/kernels/codeExecution.ts
```diff
@@ -0,0 +1,176 @@
...
+ * Execution phases include - enqueue for execution (done in ctor), start execution, completed execution with/without errors, cancel execution or dequeue.
+ *
+ * WARNING: Do not dispose `request: Kernel.IShellFuture` object.
+ * Even after request.done & execute_reply is sent we could have more messages coming from iopub.
+ * Further details here https://github.com/microsoft/vscode-jupyter/issues/232 & https://github.com/jupyter/jupyter_client/issues/297
+ *
+ */
+export class CodeExecution implements IDisposable {
```
rchiodo, line 35: It seems weird that we'd need an entirely new class just to execute a string instead of a notebook cell. Could CellExecution not use this instead?

DonJayamanne: Gets way too messy (lots of if/else), i personally think this is cleaner than trying to get a single class to do both.
Also i don't think execute hidden will require execution of code that has widgets & other complexities.

roblourens:
Please add me as a reviewer on PRs like this so I get a notification. I don't quite follow this though, how does it actually queue RBL?

src/client/datascience/jupyter/kernels/codeExecution.ts
rchiodo: No not a single class. The other class would use this one. This one knows only about code. 

The other one uses this class internally to actually do an execute. It just listens to the promise and updates the cell outputs on the cell. 

Otherwise we end up with the same code doing iopub listening and the potential for having to fix things in two spots if something is broken.

DonJayamanne:
> how does it actually queue RBL?

Assume we have 10 cell and we're hit run all, then we hit RBL for the last cell, ideally what should happen (i plan on bringing this up in standup to discuss expected behavior) is we should start RBL after the 10 cells.
In jupyter all executions are queued (if you're already running a cell & attempt to run another cell, all that happens is, it gets put into a queue & it will run after all previous executions).
Thus, RBL will happen after all previous pending execution requests.

The down side of this is, RBL could start 10 minutes after you clicked RBL, (i'll be bringing up this issue & a few others related to RBL in standup)
Its possible this solution of queueing is the right solution, we could even just as well disable RBL when we have pending executions... lets discuss in standup

src/client/datascience/jupyter/kernels/codeExecution.ts
DonJayamanne: >  The other class would use this one. This one knows only about code.

I'm still not convinced, this is simple and small, it just gets messy. We don't need widgets, etc, hence this code is sufficient as is.

> code doing iopub listening and the potential for having to fix things in two

I'd perfer to refactor when that happens, right now this feels much simpler to me, but thats just me.

DavidKutu:
As we discussed on standup, we'll wait for after release for this. And there's a chance we'll drop this and just not allow run by line/debugging when the kernel is running.

src/client/debugger/jupyter/helper.ts
DavidKutu: Not queuing that is the root of the issue

roblourens:
@DonJayamanne I get it, you're describing how queueing should work, I don't see how this PR implements that. It seems like individual executions will be queued but not the operation of starting RBL itself. But anyway, like we discussed we can revisit it  later

Issue #7325, opened 2021-08-28
Notebooks - Variables explorer cleared when data-viewer is invoked from it and given focus.


claudiaregio:
## Environment data

-   VS Code version: 1.60.0
-   Jupyter Extension version: v2021.8.2031173646

## Steps to reproduce: 
1. Open variable explorer 
2. Open dataframe in data viewer

## Expected behavior

Variables are remain

## Actual behavior

Variables are cleared

I know that this happens because the focus changes to the data viewer which is obviously not a notebook and therefore has no variables to show. However, I found this to be a really strange interaction. When wanting to open two dataframes back to back I thought it broke, took me a second to realize I had to go back to get focus on the notebook to go open the variable again.

PR #7838, opened 2021-10-06
Support using the variable viewer when a data viewer window is open

rchiodo:
Fixes: #7325

### DIFF
--- a/.github/actions/create-venv-for-tests/action.yml
+++ b/.github/actions/create-venv-for-tests/action.yml
@@ -27,6 +27,7 @@ runs:
         python -m ipykernel install --user --name .venvkernel --display-name .venvkernel
         python -m pip uninstall jedi --yes
         python -m pip install jedi==0.17.2
+        python -m pip install pandas
 
         python -m venv .venvnokernel
         source .venvnokernel/bin/activate
--- a/src/client/datascience/variablesView/notebookWatcher.ts
+++ b/src/client/datascience/variablesView/notebookWatcher.ts
@@ -14,6 +14,7 @@ import {
 import '../../common/extensions';
 import { IFileSystem } from '../../common/platform/types';
 import { IDisposableRegistry } from '../../common/types';
+import { IDataViewerFactory } from '../data-viewing/types';
 import { getActiveInteractiveWindow } from '../interactive-window/helpers';
 import { IKernelProvider } from '../jupyter/kernels/types';
 import { isJupyterNotebook } from '../notebook/helpers/helpers';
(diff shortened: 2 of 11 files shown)


.github/actions/create-venv-for-tests/action.yml
```diff
@@ -27,6 +27,7 @@ runs:
         python -m ipykernel install --user --name .venvkernel --display-name .venvkernel
         python -m pip uninstall jedi --yes
         python -m pip install jedi==0.17.2
+        python -m pip install pandas
```
rchiodo, line 30: Need this for the test to not have to put up an install dialog

src/client/datascience/variablesView/notebookWatcher.ts
```diff
@@ -40,7 +41,11 @@ export class NotebookWatcher implements INotebookWatcher {
         return this._onDidRestartActiveNotebook.event;
     }
     public get activeNotebook(): INotebook | undefined {
-        return this.notebookEditorProvider.activeEditor?.notebook || this.getActiveInteractiveWindowNotebook();
+        return (
+            this.notebookEditorProvider.activeEditor?.notebook ||
+            this.getActiveInteractiveWindowNotebook() ||
+            this.getActiveDataViewerNotebook()
```
rchiodo, line 47: This line here is what enables the whole thing. The 'active' notebook can be gathered even from the dataviewer that's active.

DonJayamanne: Wouldn't the same apply to plot viewers?
I think we should just show the Title of the notebook in the variable view & never hide the list of variables, then we'd never have to fix such issues one by one.

DonJayamanne: I'm not convinced this is a good UX, i think we should not hide the variables (seems like a never ending list of scenarios we need to cover), lets discuss in standup.

It should be similar to debugger variables, we display debugger varaibles always, unbless you switch to another debug session.

DonJayamanne (approved):

src/client/datascience/variablesView/notebookWatcher.ts
rchiodo: Variables don't work without a notebook though (plotviewer or file not associated with an interactive window would mean it was broken). Prior to this change the variable list would sometimes clear and sometimes just not work (if you scrolled or tried to open a dataframe it would just do nothing).

This is forcing it to refresh every time you switch active editors so that it always reflects the state of the active editor.

Issue #7829, opened 2021-10-05
Cell delimiters on indented code require a space


rchiodo:
Have code like this:

```python
#%%
def main():
  #%%
  print('do stuff')
```

There should be two cells there. There's only one

![image](https://user-images.githubusercontent.com/19672699/136112704-4b671416-8db5-49d7-90ef-dc5dfd58aaa2.png)

If I add a space it works:
```python
# %%
def main():
  # %%
  print('do stuff')
```

![image](https://user-images.githubusercontent.com/19672699/136112753-1ec1d6a9-841d-4285-bfce-ea0b0dd453fb.png)

Andrewnt219:
Hi @rchiodo, I want to help with this issue. Can you tell me which file(s) should I start looking at?

PR #7955, opened 2021-10-16
Fix `cellMatcher` does not match `#%%`

Andrewnt219:
For #7829 

This is my finding so far. The issue is that your second `#%%` is not recognized as a cell. To be exact, it is not a valid code.

https://github.com/microsoft/vscode-jupyter/blob/6e5df5797ec1e261cebbdf48dcad1cc205042bd8/src/client/datascience/cellMatcher.ts#L34-L36

https://github.com/microsoft/vscode-jupyter/blob/6e5df5797ec1e261cebbdf48dcad1cc205042bd8/src/client/datascience/cellMatcher.ts#L42-L44

For your reference, this is the `codeMatchRegEx`:

https://github.com/microsoft/vscode-jupyter/blob/6e5df5797ec1e261cebbdf48dcad1cc205042bd8/src/client/datascience/constants.ts#L206

The regex basically specifies that both `# %%` and `#%%` are valid. However, your second `#%%` comes with leading whitespaces, which makes the regex failed. Seem like the OR should handle that case. So this is the `defaultCellMaker`

https://github.com/microsoft/vscode-jupyter/blob/6e5df5797ec1e261cebbdf48dcad1cc205042bd8/src/client/datascience/cellMatcher.ts#L30

As the code use string comparison, even if I change the fallback of `defaultCellMarker ` to `#%%`, the code will fail when it encounters `# %%` with leading whitespaces. I believe the best solution is trimming the string of `isCode` before doing any matching.

```ts
 public isCode(code: string): boolean { 
     const formattedCode = code.trim();
     return this.codeMatchRegEx.test(formattedCode) || formattedCode === this.defaultCellMarker; 
 } 
```

### But that is not the only issue

The current `defaultCellMarkerExec` cannot handle both `#%%` and `# %%`.

https://github.com/microsoft/vscode-jupyter/blob/6e5df5797ec1e261cebbdf48dcad1cc205042bd8/src/client/datascience/cellMatcher.ts#L30-L31

A solution is turning `defaultCellMaker` into a regex string

```ts
this.defaultCellMarker = settings?.defaultCellMarker ? settings.defaultCellMarker : '#\\s*%%';
```

With the two changes in `defaultCellMaker` and `isCode`, the matcher should work as you want, but I don't know if changing the fallback of `defaultCellMaker` to a regex string is a good idea.

image


rchiodo (approved):

Andrewnt219:
@rchiodo Should I try fixing the failed tests? Seem like a CI/CD problem.

rchiodo:
> @rchiodo Should I try fixing the failed tests? Seem like a CI/CD problem.

No @DonJayamanne is working on fixing them in another branch. We can just wait until main is passing for the same tests and then do a merge.

rchiodo:
Don fixed the failures. @Andrewnt219 can you remerge main and let the tests run again? They should pass now (or at least we should only get failures that might be related to your change). Main is fully passing again.

src/client/datascience/cellMatcher.ts
```diff
@@ -28,7 +28,7 @@ export class CellMatcher {
         this.codeExecRegEx = new RegExp(`${this.codeMatchRegEx.source}(.*)`);
         this.markdownExecRegEx = new RegExp(`${this.markdownMatchRegEx.source}(.*)`);
         this.defaultCellMarker = settings?.defaultCellMarker ? settings.defaultCellMarker : '# %%';
-        this.defaultCellMarkerExec = this.createRegExp(`${this.defaultCellMarker}(.*)`, /# %%(.*)/);
+        this.defaultCellMarkerExec = this.createRegExp(`(${this.defaultCellMarker}|#\\s*%%)(.*)`, /#\s*%%(.*)/);
```
amunger, line 31: `(${this.defaultCellMarker}|#\\s*%%)(.*)` looks like it will introduce a bug - if the cell marker setting was empty, the regex would be `|\s*%%)(.*)`, and match an empty string (the part to the left of the pipe).

amunger (changes requested):

src/client/datascience/cellMatcher.ts
amunger: I actually just removed this line entirely to simplify the CellMatcher and added some unit tests in https://github.com/microsoft/vscode-jupyter/pull/9559. From the unit tests, it seems like what you're aiming for in this PR is mostly already there.

Issue #7966, opened 2021-10-19
IW disable auto new cell on running last cell


janosh:
Add an IW setting to disable creating a new cell when executing the last cell.

### Current behavior

```py
# %%
print("hi")
```

When executing the above cell in IW, a new cell is added automatically.
```py
# %%
print("hi")
# %%

```

Can't find a way to disable this.

### Desired behavior

Input file remains unchanged.

PR #7995, opened 2021-10-21
Add setting jupyter.newCellOnRunLast

janosh:
For #7966.

Went ahead and took a swing at adding a setting to disable appending a new empty cell to IW files when running the last cell. Not sure how successfully tbh. Is `runCurrentCellAndAddBelow()` the right function to modify? If so, should we rename it to `runCurrentCellAndMaybeAddBelow`?

### DIFF
--- a/src/client/common/configSettings.ts
+++ b/src/client/common/configSettings.ts
@@ -94,6 +94,7 @@ export class JupyterSettings implements IWatchableJupyterSettings {
     public disableZMQSupport: boolean = false;
     public verboseLogging: boolean = false;
     public showVariableViewWhenDebugging: boolean = true;
+    public newCellOnRunLast: boolean = true;
 
     public variableTooltipFields: IVariableTooltipFields = {
         python: {


janosh:
Oh, I didn't know about ctrl + enter￼.

janosh:
This setting doesn't actually work, new cells are appended on running the last even when set to `false`. Probably should have added a test as well. Am I fetching the setting the right way? I just copied

```ts
const { newCellOnRunLast } = this.configService.getSettings(this.documentManager.activeTextEditor.document.uri);
```

from another place in the code by analogy.

rchiodo:
Yeah that looks correct to me. 

I debugged it a bit. I think the check is in the wrong spot.

I believe it should be here:
https://github.com/microsoft/vscode-jupyter/blob/796361a98965ecf0e178034bf939e72e3058022e/src/client/datascience/editor-integration/codewatcher.ts#L343

Instead of passing true for the 'advance' parameter, it should check the setting you added.

janosh:
@rchiodo Thanks for taking a look and sorry for getting this wrong. I think the `true` there must stay and instead we need to check in `runMatchingCell` if `newCellOnRunLast` is `true` in which case we don't do anything in the `else` case:

https://github.com/microsoft/vscode-jupyter/blob/796361a98965ecf0e178034bf939e72e3058022e/src/client/datascience/editor-integration/codewatcher.ts#L1042-L1048

rchiodo:
Yeah that seems better. I was thinking we only wanted to apply the setting on the SHIFT+ENTER case, but we should apply it on the 'runcell' code lens too (which I believe is the other case that has advance == true)

PR #8098, opened 2021-10-29
Improvements to logging

DonJayamanne:
Stop displaying user names in logs (I find lots of users strip out the first few parts of the logs as it contains their user names)
* This change ensures we use `~` where possible in our logs (minimizing displaying user names in logs)
	* As a resuult the text is also smaller
* CHanged some of the logging levels to verbose (logging environment variables is unnecessary, i don't think we've used this)

### DIFF
--- a/src/client/datascience/kernel-launcher/kernelDaemon.ts
+++ b/src/client/datascience/kernel-launcher/kernelDaemon.ts
@@ -89,7 +89,7 @@ export class PythonKernelDaemon extends BasePythonDaemon implements IPythonKerne
             // This is why when we run `execModule` in the Kernel daemon, it finishes (comes back) quickly.
             // However in reality it is running in the background.
             // See `m_exec_module_observable` in `kernel_launcher_daemon.py`.
-            traceVerbose(`Starting kernel from scratch with options ${JSON.stringify(options)}`);
+            traceInfo('Starting kernel from scratch');
             await this.execModule(moduleName, args, options);
         }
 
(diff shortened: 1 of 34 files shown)


src/client/logging/formatters.ts
```diff
@@ -32,6 +32,9 @@ function normalizeLevel(name: LogLevelName): string {
             return norm;
         }
     }
+    if ((name as any) === 'silly') {
```
DonJayamanne, line 35: I don't think we should display `Silly` in console logs.

src/client/logging/trace.ts
```diff
@@ -66,7 +66,11 @@ function logResult(loggers: ILogger[], info: LogInfo, traced: TraceInfo, call?:
     const formatted = formatMessages(info, traced, call);
     if (traced.err === undefined) {
         // The call did not fail.
-        if (!info.level || info.level > LogLevel.Error) {
+        if (info.level && info.level === LogLevel.Error) {
```
DonJayamanne, line 69: Bug, verbose & other logs are not logged

src/client/datascience/kernel-launcher/kernelDaemon.ts
```diff
@@ -89,7 +89,7 @@ export class PythonKernelDaemon extends BasePythonDaemon implements IPythonKerne
             // This is why when we run `execModule` in the Kernel daemon, it finishes (comes back) quickly.
             // However in reality it is running in the background.
             // See `m_exec_module_observable` in `kernel_launcher_daemon.py`.
-            traceInfo(`Starting kernel from scratch with options ${JSON.stringify(options)}`);
+            traceVerbose(`Starting kernel from scratch with options ${JSON.stringify(options)}`);
```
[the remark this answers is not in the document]
rchiodo: I think we could just skip it.

rchiodo (approved):

PR #8102, opened 2021-10-29
Merge release changes back into main

rchiodo:
According to the release [plan](https://github.com/microsoft/vscode-jupyter/issues/7909) this is moving changelog items over.

Engine version will change later when we turn insiders back on.

From this point on, we're shipping only release.


IanMatthewHuff:
I can't quite remember the process, but do you want to bump the engine now (1.63-insiders or whatever) so that we don't accidentally release an insiders build that overwrites live stuff when we turn things back on?

rchiodo:
> I can't quite remember the process, but do you want to bump the engine now (1.63-insiders or whatever) so that we don't accidentally release an insiders build that overwrites live stuff when we turn things back on?

No we instead turn off the insiders version. We don't bump until we ship the release bits so that it can continue to run with the insiders we have on our own boxes.

See the entry in the release plan:
https://github.com/microsoft/vscode-jupyter/issues/7909

DonJayamanne (approved):

IanMatthewHuff (approved):

package.json
```diff
@@ -28,7 +28,7 @@
         "theme": "light"
     },
     "engines": {
-        "vscode": "1.62.0-insider"
```
DonJayamanne, line 31: I thought this should be 1.63.0-insider

DonJayamanne: Perhaps not, i guess its ok. Can't remember what i did.
It shouldn't matter as releases are off

PR #8103, opened 2021-10-29
Fix flakiness of debugger tests. They don't need to show an interrupt…

rchiodo:
They don't need to show an interrupt just have to finish debugging

Doing this to get CI green before we release. But test seems more robust this way. Doesn't matter if keyboard exception doesn't come out.


IanMatthewHuff (approved):
Seems fine, presuming that we don't care about the Interrupt showing up in the UI.

Issue #8107, opened 2021-10-29
Unable to interrupt non-python kernels


DonJayamanne:

DonJayamanne:
**Testing**

* Julia (on linux/mac)
	* Create a julia notebook and select the Julia Jupyter kernel
``` julia
function estimate_pi(n)
    s = 1.0
    for i in 1:n
		sleep(1)
        s += (isodd(i) ? -1 : 1) / (2i + 1)
    end
    4s
end

p = estimate_pi(100_000_000)
println("π ≈ $p")
println("Error is $(p - π)")
``` 
	* Run this cell and try to interrupt
	* Be patient sometimes it takes a few tries (i think because Julia process is busy or the like)
	* The same flaky ness can be experienced in Jupyter as well.

rchiodo:
/verified? Doesn't work for me but I see it sending the interrupt requests.

PR #8108, opened 2021-10-29
Support interrupting non-python kernels

DonJayamanne:
For #8107

Basically non-python kernles are not interrupted today as we don't send the sigint signal.

### DIFF
--- a/src/client/common/configSettings.ts
+++ b/src/client/common/configSettings.ts
@@ -92,6 +92,8 @@ export class JupyterSettings implements IWatchableJupyterSettings {
     public interactiveWindowMode: InteractiveWindowMode = 'multiple';
     // Hidden settings not surfaced in package.json
     public disableZMQSupport: boolean = false;
+    // Hidden settings not surfaced in package.json
+    public disablePythonDaemon: boolean = false;
     public verboseLogging: boolean = false;
     public showVariableViewWhenDebugging: boolean = true;
     public newCellOnRunLast: boolean = true;
--- a/src/client/datascience/kernel-launcher/kernelProcess.ts
+++ b/src/client/datascience/kernel-launcher/kernelProcess.ts
@@ -3,6 +3,7 @@
 'use strict';
 
 import { ChildProcess } from 'child_process';
+import { kill } from 'process';
 import * as fs from 'fs-extra';
 import * as tmp from 'tmp';
 import { CancellationToken, Event, EventEmitter } from 'vscode';
@@ -52,12 +53,21 @@ export class KernelProcess implements IKernelProcess {
     private get isPythonKernel(): boolean {
         return isPythonKernelConnection(this.kernelConnectionMetadata);
     }
+    public get canInterrupt() {
+        if (this.pythonDaemon) {
+            return true;
+        }
+        if (this._kernelConnectionMetadata.kernelSpec.interrupt_mode === 'message') {
+            return false;
+        }
+        return true;
+    }
     private _process?: ChildProcess;
     private exitEvent = new EventEmitter<{ exitCode?: number; reason?: string }>();
     private pythonKernelLauncher?: PythonKernelLauncherDaemon;
     private launchedOnce?: boolean;
     private disposed?: boolean;
-    private kernelDaemon?: IPythonKernelDaemon;
+    private pythonDaemon?: IPythonKernelDaemon;
     private connectionFile?: string;
     private _launchKernelSpec?: IJupyterKernelSpec;
     private readonly _kernelConnectionMetadata: Readonly<KernelSpecConnectionMetadata | PythonKernelConnectionMetadata>;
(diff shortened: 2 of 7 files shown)


src/client/common/configSettings.ts
```diff
@@ -92,6 +92,8 @@ export class JupyterSettings implements IWatchableJupyterSettings {
     public interactiveWindowMode: InteractiveWindowMode = 'multiple';
     // Hidden settings not surfaced in package.json
     public disableZMQSupport: boolean = false;
+    // Hidden settings not surfaced in package.json
+    public disablePythonDaemon: boolean = false;
```
DonJayamanne, line 96: This applies to unix systems.
If daemon doesn't work, then python kernel cannot be interrupted (just like any other non-python kernel)

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -52,12 +52,21 @@ export class KernelProcess implements IKernelProcess {
     private get isPythonKernel(): boolean {
         return isPythonKernelConnection(this.kernelConnectionMetadata);
     }
+    public get canInterrupt() {
+        if (this.pythonDaemon) {
+            return true;
+        }
+        if (this._kernelConnectionMetadata.kernelSpec.interrupt_mode === 'message') {
+            return false;
```
DonJayamanne, line 60: the raw kernel process can only interrupt pyton daemon kernels or kernsls that support interrupt messages

src/client/datascience/raw-kernel/rawKernel.ts
```diff
@@ -160,10 +160,28 @@ export class RawKernel implements Kernel.IKernelConnection {
...
+            // traceInfo(`Interrupting kernel with a shell message`);
+            // const jupyterLab = require('@jupyterlab/services') as typeof import('@jupyterlab/services');
+            // const msg = (jupyterLab.KernelMessage.createMessage({
+            //     msgType: 'interrupt_request' as any,
+            //     channel: 'shell',
+            //     username: this.realKernel.username,
+            //     session: this.realKernel.clientId,
+            //     content: {}
+            // }) as any) as KernelMessage.IShellMessage<'inspect_request'>;
+            // await this.realKernel.sendShellMessage<'inspect_request'>(msg, true, true).done;
```
DonJayamanne, line 180: This code seems to be right, however I can't find any kernel that implements this even though the protocol supports this.
Hence not sure whether we should enable this or not.

Worse case nothing happens
Commented this out so we can discuss this.
Personally feel its ok to enable this, if there are kernels that support message interrupts (for windows or even unix)

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -75,8 +84,17 @@ export class KernelProcess implements IKernelProcess {
...
-            await this.kernelDaemon?.interrupt();
+        if (!this.canInterrupt) {
+            throw new Error('Kernel interrupt not supported in KernelProcess.ts');
+        }
+        if (this.pythonDaemon) {
+            traceInfo('Interrupting kernel via Daemon message');
+            await this.pythonDaemon.interrupt();
+        } else if (this._kernelConnectionMetadata.kernelSpec.interrupt_mode !== 'message' && this._process) {
+            traceInfo('Interrupting kernel via Signals');
+            this._process.kill('SIGINT');
```
DonJayamanne, line 95: This only works on unix systems.
Oh wait, I need to ensure we check the Platform (if windows or not).
because windows doesn't supoort signals

rchiodo: Doesn't this kill the process? Can you send a signal without 'kill'?

src/client/datascience/raw-kernel/rawKernel.ts
rchiodo: Yeah seems okay to me.

src/client/datascience/kernel-launcher/kernelProcess.ts
DonJayamanne: On unix yes, on windows no.

Separate topic: I think .NET 6.0 has added support for signals on windows

rchiodo: Oh I guess it just sends the signal. Kill with 'SIGTERM' is how you really kill. What a crappy method name.

DonJayamanne: > oesn't this kill the process? Can you send a signal wi

That's how you send signals in node. weird isn't it.
Will update to make it more clear by using `process.kill(<pid>, signa)`

src/client/datascience/raw-kernel/rawKernel.ts
DonJayamanne: Ok to enable or  keep it commented out.

rchiodo: Ok to enable. It doesn't hurt anything AFAICT

IanMatthewHuff: Code looks like it's for 'inspect_request'? Isn't that for variables?

DonJayamanne: Sorry,that part was for generics to work. Jupyter Lab NPM package doesn't support the interrupt_message request in their TS Types.
I changed it and used `any`.
The actual payload however is for interrupt request.

DonJayamanne:
Closing this PR in favor of https://github.com/microsoft/vscode-jupyter/compare/interruptNonPythonKernels?expand=1
No chages, for some reason this PR would show thirdparty notices file had changes (even after rebasing with main).

PR #8127, opened 2021-11-01
Ignore events from IJupyterSession if its not the active session

DonJayamanne:
Looking at failures in this log https://pipelines.actions.githubusercontent.com/KyBlh4sBTj8eoTzBTvt0CAApvyyj3PXuc4Tbg2yHyMGda4Kk7D/_apis/pipelines/1/runs/23088/signedlogcontent/17?urlExpires=2021-11-01T20%3A49%3A54.9664908Z&urlSigningMethod=HMACV1&urlSignature=r0HBCl88NwLJc6y5J0gUQAkbXxqRlsdKN%2Fn4dIzcBmo%3D

**Change 1 - Fix issue found in CI**
This is whats happening:
* When restarting or switching kernels, the old IJupyterSession gets disposed
* When disposing the old IJupyterSession we trigger disposed events and the like
* However this is the old session and its not longer valid, we're starting a whole new IJupyterSession that will replace the old one.

Solution: Ensure we don't handle those events

**Change 2 - When a kernel dies, stop all of the executions**
* No point running cells.

### DIFF
--- a/src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
+++ b/src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
@@ -141,6 +141,13 @@ export class CellExecutionQueue implements Disposable {
                 await this.cancel();
                 break;
             }
+            // If the kernel is dead, then no point trying the rest.
+            if (kernelConnection.status === 'dead' || kernelConnection.status === 'terminating') {
+                this.cancelledOrCompletedWithErrors = true;
+                traceInfo(`Cancel all remaining cells due to dead kernel`);
+                await this.cancel();
+                break;
+            }
         }
     }
 }
--- a/src/client/datascience/jupyter/kernels/kernel.ts
+++ b/src/client/datascience/jupyter/kernels/kernel.ts
@@ -92,7 +92,7 @@ export class Kernel implements IKernel {
         return this._info;
     }
     get status(): KernelMessage.Status {
-        return this.notebook?.session?.status ?? 'unknown';
+        return this.notebook?.session?.status ?? (this.isKernelDead ? 'dead' : 'unknown');
     }
     get disposed(): boolean {
         return this._disposed === true || this.notebook?.session.disposed === true;
@@ -101,6 +101,13 @@ export class Kernel implements IKernel {
         return this._kernelSocket.asObservable();
     }
     private notebook?: INotebook;
+    /**
+     * If the session died, then ensure the status is set to `dead`.
+     * We need to provide an accurate status.
+     * `unknown` is generally used to indicate jupyter kernel hasn't started.
+     * If a jupyter kernel dies after it has started, then status is set to `dead`.
+     */
+    private isKernelDead?: boolean;
     public get session(): IJupyterSession | undefined {
         return this.notebook?.session;
     }
@@ -235,7 +242,10 @@ export class Kernel implements IKernel {
         traceInfo(`Restart requested ${this.notebookDocument.uri}`);
         this.startCancellation.cancel();
         try {
-            await this.kernelExecution.restart(this._notebookPromise?.then((item) => item.session));
+            // If the notebook died, then start a new notebook.
+            await (this._notebookPromise
+                ? this.kernelExecution.restart(this._notebookPromise?.then((item) => item.session))
+                : this.start({ disableUI: false }));
             traceInfoIfCI(`Restarted ${getDisplayPath(this.notebookDocument.uri)}`);
         } catch (ex) {
             traceError(`Restart failed ${getDisplayPath(this.notebookDocument.uri)}`, ex);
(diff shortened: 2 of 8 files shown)


src/client/datascience/jupyter/kernels/cellExecutionQueue.ts
```diff
@@ -141,6 +141,13 @@ export class CellExecutionQueue implements Disposable {
                 await this.cancel();
                 break;
             }
+            // If the kernel is dead, then no point trying the rest.
+            if (kernelConnection.status === 'dead' || kernelConnection.status === 'terminating') {
```
DonJayamanne, line 145: Cancel all pending executions, no point even trying

DonJayamanne: What could happen is, if kernel dies and we have more cells, things go into a weird start & execution never completes.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -92,7 +92,7 @@ export class Kernel implements IKernel {
         return this._info;
     }
     get status(): KernelMessage.Status {
-        return this.notebook?.session?.status ?? 'unknown';
+        return this.notebook?.session?.status ?? (this.isKernelDead ? 'dead' : 'unknown');
```
DonJayamanne, line 95: If the IJupyterSession is dead, then Kernel.status will not reflect the status correctly.
Previously it would return `unknown` and that's incorrect. In jupyter `unknown` is actually `not started` and they have `dead` for dead kernels.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -417,8 +428,16 @@ export class Kernel implements IKernel {
...
                 if (!this._ignoreNotebookDisposedErrors) {
+                    const isActiveNotebookDead = this.notebook === notebook;
+
                     this._notebookPromise = undefined;
-                    this._onDisposed.fire();
+                    this.notebook = undefined;
+
+                    // If the active notebook died, then kernel is dead.
+                    if (isActiveNotebookDead) {
+                        this.isKernelDead = true;
```
DonJayamanne, line 438: Ensure we don't trigger eevnts unnecessarily (e.g. if somehow create a new INotebook, then we need to take those into account)

src/client/datascience/raw-kernel/rawJupyterSession.ts
```diff
@@ -200,35 +200,44 @@ export class RawJupyterSession extends BaseJupyterSession {
...
         }
-        if (session?.kernelProcess) {
-            // Watch to see if our process exits
-            this.processExitHandler = session.kernelProcess.exited(({ exitCode, reason }) => {
-                sendTelemetryEvent(Telemetry.RawKernelSessionKernelProcessExited, undefined, {
-                    exitCode,
-                    exitReason: getTelemetrySafeErrorMessageFromPythonTraceback(reason)
-                });
-                traceError(`Raw kernel process exited code: ${exitCode}`);
-                this.shutdown().catch((reason) => {
```
DonJayamanne, line 225: This was wrong, at this point `session !== this.session`, hence we end up shutting down this session and then INotebook.onDidDispose gets fired and everyone thinks the kernel is dead.
But what is happening is we're restarting a kernel and the old IJupyterSession is being disposed & replaced by the new IJupyterSession.

Not easy to replicate the issues as things happen fast. but when its slow, thats when we have failures (as on CI)

src/client/datascience/raw-kernel/rawJupyterSession.ts
```diff
@@ -200,35 +200,44 @@ export class RawJupyterSession extends BaseJupyterSession {
...
+                exitReason: getTelemetrySafeErrorMessageFromPythonTraceback(reason)
             });
-            this._disposables.push(this.processExitHandler);
-        }
+            traceError(`Raw kernel process exited code: ${exitCode}`);
+            // If the raw kernel process dies, then send the terminating event, and shutdown the session.
+            // Afer shutting down the session, the status changes to `dead`
+            this.isTerminating = true;
+            this.onStatusChangedEvent.fire('terminating');
+            this.shutdown()
```
DonJayamanne, line 233: Will fix this later in another PR, i dont think the session should be disposing iteself here.
But lets look into that later.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -101,6 +101,13 @@ export class Kernel implements IKernel {
         return this._kernelSocket.asObservable();
     }
     private notebook?: INotebook;
+    /**
+     * If the session died, then ensure the status is set to `dead`.
+     * We need to provide an accurate status.
+     * `unknown` is generally used to indicate jupyter kernel hasn't started.
+     * If a jupyter kernel dies after it has started, then status is set to `dead`.
+     */
+    private isKernelDead?: boolean;
```
DonJayamanne, line 110: I want to remove this, horrible to have more states (basically yet another wrapper around `status`).
But we need this for raw kernels.
Once we clean up Kernels as a whole this should go away..

wait, i think i'll remove this.

rchiodo (approved):

rchiodo (approved):

PR #8133, opened 2021-11-02
Remove kernel finder from server start

DonJayamanne:
**Original issue**: Looked at using the new Python extension API to search for interpreters 
This requires a refactor of how we search for kernels.
Found that kernel finder is used when starting kernel. But that should be unnecessary, we can never start a kernel without ever knowing (user having selected) the kernel from the picker.

Found this was to do with `connect` method that requried kernel specs.

**Fixes**
* We no longer need to pass metadata or kernel spec connection information when starting notebook servers
* The only use case was when connecting to remote, we re-use the session
	* But with little to zero benefit, as we'd need to switch the kernel anyway, and starting kernels is what's slow
* We no longer serach for kernels when starting notebook servers
* As a restul of the above, a lot more code has gone

### DIFF
--- a/src/client/datascience/jupyter/jupyterExecution.ts
+++ b/src/client/datascience/jupyter/jupyterExecution.ts
@@ -19,8 +19,6 @@ import { IServiceContainer } from '../../ioc/types';
 import { PythonEnvironment } from '../../pythonEnvironments/info';
 import { captureTelemetry, sendTelemetryEvent } from '../../telemetry';
 import { Identifiers, Telemetry } from '../constants';
-import { ILocalKernelFinder, IRemoteKernelFinder } from '../kernel-launcher/types';
-import { trackKernelResourceInformation } from '../telemetry/telemetry';
 import {
     IJupyterConnection,
     IJupyterExecution,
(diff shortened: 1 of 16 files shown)


src/client/datascience/jupyter/jupyterExecution.ts
```diff
@@ -137,67 +129,29 @@ export class JupyterExecutionBase implements IJupyterExecution {
...
 
-                    // In a remote non guest situation, figure out a kernel spec too.
-                    if (
-                        (!kernelConnectionMetadata ||
-                            !kernelConnectionMetadataHasKernelSpec(kernelConnectionMetadata)) &&
-                        connection &&
-                        !options.skipSearchingForKernel
-                    ) {
-                        const kernelFinder = this.serviceContainer.get<IRemoteKernelFinder>(IRemoteKernelFinder);
-                        kernelConnectionMetadata = await kernelFinder.findKernel(
```
DonJayamanne, line 178: This is what I wanted to remove.

src/test/datascience/jupyter/jupyterSession.unit.test.ts
```diff
@@ -453,4 +383,41 @@ suite('DataScience - JupyterSession', () => {
...
+                when(sessionManager.startNew(anything(), anything())).thenCall(() => {
+                    return Promise.resolve(instance(remoteSession));
+                });
+
+                const signal = mock<ISignal<ISessionWithSocket, Kernel.Status>>();
+                when(remoteSession.statusChanged).thenReturn(instance(signal));
+
+                await connect('connectToLiveKernel');
+            });
+            test('Restart should restart the new remote kernel', async () => {
```
rchiodo, line 408: Could you add the test for 'switching' is supposed to kill the old kernel? That's the part I'm worried about. The code you removed from hostJupyterNotebook.

rchiodo (approved):

src/test/datascience/jupyter/jupyterSession.unit.test.ts
DonJayamanne: We can no longer switch kernels this way, switching kernels happens at a higher level in VS Code level.
I'll check if we have tests for that, if not i'll add it.

Issue #8119, opened 2021-11-01
Failure to activate conda env results in failure to detect if ipykernel is installed


DonJayamanne:
Initially reported by Rong.
Basically activation of conda environments fails, hence we cannot launch conda properly.

DonJayamanne:
Jim too is running into this.
**Summary:**
* The process to get env variables seems to be timing out, and the process is terminated hence the SIGTERM error
* When using `conda run` we run into the exact same issue

DonJayamanne:
Possible solutions here https://github.com/microsoft/vscode-jupyter/issues/6694

DonJayamanne:
Timeout increased in Python extesnion

rchiodo:
@greazer can you verify this works?

PR #8142, opened 2021-11-02
Lot and check if we fail to get activated env variables

DonJayamanne:
Related to https://github.com/microsoft/vscode-jupyter/issues/8119


rchiodo (changes requested):
Can you run the regenerate telemetry task?

```npm run generateTelemetry```

rchiodo (approved):

Issue #8060, opened 2021-10-26
Is the kernel error being reported correct (i.e. the most useful?)


greazer:
See the following image. The kernel (interpreter) I selected apparently is missing access to win32api.dll somehow. Yet the error reported is a much more nebulous and late-to-the-party type message: "Cannot execute code, session has been exposed". 

I think we have code to try and display to the user a better error message when a kernel can't be started, but are we getting the right one? 

![image](https://user-images.githubusercontent.com/11685408/138959031-bfb05142-f3c9-43a8-8faa-c7dd2ea3d20d.png)

DonJayamanne:
**Testing:**

* Create a notebook with at least 3 cells (1 & 3 containing print statements)
	* Second cell should have the following code
```python
import IPython
app = IPython.Application.instance()
app.kernel.do_shutdown(True)
``` 
	* An error message should also be displayed indicating the kernel died (bottom right)
	* Try to run another cell or the same cell in the notebook
	* You should be presented with a modal message indicating the kernel is dead with the option to restart the kernel

* Create a notebook with the following code:
```python
import os
os.getpid()
``` 
	* Using this PID kill this process from the task manager or terminal
	* An error message should also be displayed indicating the kernel died (bottom right)
	* Try to run another cell or the same cell in the notebook
	* You should be presented with a modal message indicating the kernel is dead with the option to restart the kernel

PR #8156, opened 2021-11-03
Improvements related to handling of dead kernels

DonJayamanne:
For #8060

**Summary:**
* If kernel dies, display an error message notifying user its dead (bottom right)
	* We do this only for raw kernels, as jupyter automatically restarts kernels 
	* Automatic restarts of raw kernels will be done in a separate PR
* If user attempts to run a cell against a dead kernel, they get a modal dialog
	* With options to restart kernel or view output
	* Today - user gets an error message `Session cannot generate requests` error
* When a cell dies amid exrecution of a notebook, we cancel all remaining executions (no point)
	* This way we do'nt end with error mesages (session disposed) in the output   	 
* When kernels die, ensure we display a meaningful error message where possible
	* With ability to view the output logs

**TODO:** 
Will add tests, just want to get the stuff reviewed


rchiodo (approved):

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -329,9 +351,17 @@ export class Kernel implements IKernel {
                             this.notebookDocument,
                             placeholderCellPromise
                         );
+                        // if (!process.env.XYZ1234) {
```
IanMatthewHuff, line 354: ?

package.nls.json
```diff
@@ -433,9 +435,11 @@
     "DataScience.selectDifferentJupyterInterpreter": "Change Interpreter",
     "DataScience.selectJupyterInterpreter": "Select an Interpreter to start Jupyter",
     "DataScience.switchingKernelProgress": "Switching kernel to '{0}'",
-    "DataScience.sessionStartFailedWithKernel": "Failed to start a session for the Kernel '{0}'. \nView Jupyter [log](command:{1}) for further details.",
+    "DataScience.sessionStartFailedWithKernel": "Failed to start the Kernel '{0}'. \nView Jupyter [log](command:{1}) for further details.",
```
IanMatthewHuff, line 438: Looks like for sentences we use kernel not Kernel in other strings in this file.

IanMatthewHuff (approved):

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: I've left that for testing, will remove.
I thought my comments were on github, hmm

DonJayamanne: Fixed

Issue #8160, opened 2021-11-04
INotebook cache not getting cleared


DonJayamanne:
Found in a flaky CI test https://github.com/microsoft/vscode-jupyter/runs/4099490021?check_suite_focus=true

**Here's whats happening:**

* Start kernel A
* Run a cell
* Switch to kernel B
* Kernel A gets disposed, and underlying INotebook will get disposed as well
* When starting INotebook for Kernel B, we end up getting the cached INotebook for Kernel A

Basically the cache is not removed early enough and we get the old disposed INotebook & IJupyterSession

**Solution** - Stop caching INotebook (found this is done in a number of places, basically the `getNotebook` has been removed completely)

Issue #8143, opened 2021-11-02
Add tests to kill existing kernel when switching kernels


DonJayamanne:
**Today we kill the previous kernel, but we need tests for this**

Require test:
* Select kernel
* Run a cell
* Change kernel
* Confirm the previous kernel is disposed

https://github.com/microsoft/vscode-jupyter/blob/a5b704e42e308b70d27e0f4589ae694ac236701d/src/client/datascience/notebook/vscodeNotebookController.ts#L202-L212

PR #8161, opened 2021-11-04
Remove caching of INotebook 

DonJayamanne:
For #8160, #8143

Stop caching INotebook (found this is done in a number of places, basically the `getNotebook` has been removed completely)
Also added tests for #8143


rchiodo:
Additionally the reason for the 'get' call was because in some spots we didn't actually want to create a notebook. Maybe those are irrelevant now though.

DonJayamanne:
> Maybe those are irrelevant now though.

Agreed, couldn't find a single spot where that was used, hence removed.

rchiodo:
> We don't use INotebookProvider directly anymore, we use IKernel.

If this is the case can we remove INotebookProvider altogether? Or it is only used to create the INotebook wrapper around the session and the kernel? (not sure what the two properties are).

Maybe it would make sense to rename INotebook to be something like ISessionContainer or something so we break cleanly from the past.

DonJayamanne:
> n we remove INotebookProvider altogether? Or it is only used to create the INotebook wrapper around the session and the kernel? (not sure what the two properties are).

Yes thats correct.


> ebook to be something like ISessionContainer or something so we break cleanly from the past.

Sure, will do that in a separate PR (CI is super slow)

DonJayamanne:
> ook to be something like ISessionContainer or something so we break cleanly from the past.

Not really happy with that name.
We have 2 classes in Raw case, RawJupyterSession & RawSession & then having ISessionContainer would only add to the confusion, i'd prefer to refactor this rather than adding a new name that we need to get used to again.

src/client/datascience/jupyter/jupyterNotebookProvider.ts
```diff
@@ -30,31 +30,20 @@ export class JupyterNotebookProvider implements IJupyterNotebookProvider {
         return server?.getConnectionInfo();
     }
 
-    public async createNotebook(options: GetNotebookOptions): Promise<INotebook> {
+    public async createNotebook(options: NotebookCreationOptions): Promise<INotebook> {
         // Make sure we have a server
         const server = await this.serverProvider.getOrCreateServer({
-            getOnly: options.getOnly,
+            getOnly: false,
```
IanMatthewHuff, line 36: Do we use getOnly anymore? Can we just remove it from the options type?

IanMatthewHuff:
> Additionally the reason for the 'get' call was because in some spots we didn't actually want to create a notebook. Maybe those are irrelevant now though.

I had the same thought, but I don't see the usage for that anymore. It used to be important for something like autostart.

src/client/datascience/jupyter/jupyterNotebookProvider.ts
DonJayamanne: For Jupyter server it's required, to do with prewaring.

IanMatthewHuff (approved):
Code looks good to me. I think I'm just a bit hesitant since the caching always was a bit of a big ball of wax before. But no reason not to change now, we'll fix up any issues as they come up.

Issue #8175, opened 2021-11-05
Jupyter session names no longer shown in kernel picker


zpincus:
## Environment data

-   VS Code version: 
```
Version: 1.62.0 (Universal)
Commit: b3318bc0524af3d74034b8bb8a64df0ccf35549a
Date: 2021-11-03T15:24:41.297Z
Electron: 13.5.1
Chrome: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Darwin x64 20.6.0
```

-   Jupyter Extension version (available under the Extensions sidebar): v2021.10.1001414422
-   Python Extension version (available under the Extensions sidebar): v2021.11.1422169775
-   OS (Windows | Mac | Linux distro) and version: MacOS 11.6
-   Jupyter server running: Remote

## Expected behaviour

When using a "remote" jupyter server, the "kernel picker" window allows you to choose among all the running sessions on the server to connect a notebook to. In the previous version of the Jupyter extension, the names of these kernels were shown, along with a clear indication of which was currently selected.

## Actual behaviour

In the new version of the extension, the names are no longer shown, requiring a user to guess as to which session to (re)connect to. Moreover, the "currently selected" text showing which session is connected, is often almost or completely obscured. Both are illustrated below. (Only in the second screenshot can the "currently selected" text be seen at all, cut off as "Cu..." on the very right side of the top line...)

<img width="604" alt="image" src="https://user-images.githubusercontent.com/174217/140477565-9c6c6ef5-fb64-40b0-a288-c36a68774690.png">
<img width="603" alt="image" src="https://user-images.githubusercontent.com/174217/140478658-be528b40-1a80-4a97-81fd-b7863e4acc3b.png">

Also, the "last activity" text doesn't seem to update correctly; I just now re-ran a cell in the notebook, but the text remains unchanged for the currently connected session. (This also applies to the "0 existing connections" text; one might expect that the "currently selected" notebook would show up as having "1 existing connection", but no.)

This behavior pairs particularly badly with #7610, but even if that were fixed, what's the point of letting you choose among sessions if you have to guess which is which?

DonJayamanne:
Here's the update, will submit a PR
![image](https://user-images.githubusercontent.com/1948812/140549431-cf133cea-9277-43b7-89a3-a60ab748228f.png)

PR #8181, opened 2021-11-05
Display notebook paths in kernel picker for remote kernels

DonJayamanne:
For #8175

### DIFF
--- a/news/2 Fixes/8175.md
+++ b/news/2 Fixes/8175.md
@@ -1,0 +1,1 @@
+Display notebook paths associated with remote kernels in the kernel picker.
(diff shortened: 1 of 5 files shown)


rchiodo (approved):

rchiodo:
What does it look like?

DonJayamanne:
@rchiodo the screenshot is here https://github.com/microsoft/vscode-jupyter/issues/8175#issuecomment-962061805
Wait will upload here too so we can review & discuss
image

rchiodo:
I think you should switch it. 

The details should be the 'last activity, 1 existing connections' or maybe not there altogether.
The descriptoin? field (thing to right of name) should be the notebook path. It's the most important piece of information.

IanMatthewHuff (approved):
Agree with Rich's comment on order switch, but looks good.

DonJayamanne:
Done 
image

rchiodo (approved):

Issue #7278, opened 2021-08-24
On restart kernel, reconsider having the "Yes, and don't ask again" option...


greazer:
Testing microsoft/vscode#131209

I'm not sure why we added the "Yes and don't ask again" choice when I restart a kernel. This seems like generally a bad idea, since restarting kernels can be a lossy proposition. 

I think it's fine if we keep the setting, but I don't think we should advertise it.

joyceerhl:
The button should navigate the user to the settings UI page but not actually flip the setting value.

DonJayamanne:
Closing as there haven't been any upvotes on this issue

DonJayamanne:
Suggestion: 
Due to https://github.com/microsoft/vscode/issues/163729  we might want to always ask the user & be consistent with Jupyter as well (i.e. remote the option `do not show again`)

DonJayamanne:
Closing this issue, hasnt received any upvotes

PR #8190, opened 2021-11-05
Remove never ask again option for kernel restart

DonJayamanne:
For #7278


rchiodo (changes requested):
You eliminated the option, you didn't do what the bug said:

```
Based on discussion at today's notebook sync, the button should navigate the user to the settings UI page but not actually flip the setting value.
```

DonJayamanne:
yuck, too much work for now

Issue #8075, opened 2021-10-27
Create a concat doc that listens to LSP events to modify itself


rchiodo:
Replace the current notebook concat document with an implementation that can be shared with pylance.

rchiodo:
This is essentially complete. Implementation resides here:
https://github.com/microsoft/vscode-jupyter-lsp-middleware/blob/88a75b1b17118604ba30c30f98fbf113b0f540a9/src/notebookConcatDocument.ts#L36

It will likely change as I fix bugs and such but should be portable now.

rchiodo:
Verification steps:
- Open a notebook
- Verify semantic colorization happens
- Verify hover works
- Edit cells/Move cells/Delete cells
- Verify diagnostics are as expected (as if the contents of all python cells were concatenated together)

PR #8219, opened 2021-11-10
Upgrade to new lsp middleware concat

rchiodo:
For #8075

Need to do more testing to make sure this all works. Found a problem where the notebook controller wasn't being indicated on reopen of a notebook.

Also had to wire up the 'refresh' for the concat document (when a notebook cells move around)

### DIFF
--- a/src/client/datascience/notebook/vscodeNotebookController.ts
+++ b/src/client/datascience/notebook/vscodeNotebookController.ts
@@ -170,6 +170,13 @@ export class VSCodeNotebookController implements Disposable {
     public async updateNotebookAffinity(notebook: NotebookDocument, affinity: NotebookControllerAffinity) {
         traceInfo(`Setting controller affinity for ${getDisplayPath(notebook.uri)} ${this.id}`);
         this.controller.updateNotebookAffinity(notebook, affinity);
+
+        // Stick in our list of associated documents and indicate controller may have changed.
+        if (!this.associatedDocuments.has(notebook)) {
+            this.associatedDocuments.add(notebook);
+            this._onNotebookControllerSelected.fire({ notebook, controller: this });
+            this._onNotebookControllerSelectionChanged.fire();
+        }
     }
 
     // Handle the execution of notebook cell
--- a/src/client/datascience/notebook/vscodeNotebookController.ts
+++ b/src/client/datascience/notebook/vscodeNotebookController.ts
@@ -170,13 +170,6 @@ export class VSCodeNotebookController implements Disposable {
     public async updateNotebookAffinity(notebook: NotebookDocument, affinity: NotebookControllerAffinity) {
         traceInfo(`Setting controller affinity for ${getDisplayPath(notebook.uri)} ${this.id}`);
         this.controller.updateNotebookAffinity(notebook, affinity);
-
-        // Stick in our list of associated documents and indicate controller may have changed.
-        if (!this.associatedDocuments.has(notebook)) {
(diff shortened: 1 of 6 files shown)


src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -170,6 +170,13 @@ export class VSCodeNotebookController implements Disposable {
     public async updateNotebookAffinity(notebook: NotebookDocument, affinity: NotebookControllerAffinity) {
         traceInfo(`Setting controller affinity for ${getDisplayPath(notebook.uri)} ${this.id}`);
         this.controller.updateNotebookAffinity(notebook, affinity);
+
+        // Stick in our list of associated documents and indicate controller may have changed.
+        if (!this.associatedDocuments.has(notebook)) {
+            this.associatedDocuments.add(notebook);
+            this._onNotebookControllerSelected.fire({ notebook, controller: this });
```
DonJayamanne, line 177: Setting affinity is not the same as setting it as selected.
You could have a notebook that's associated with kernel A, but the recommended kernel (affinity) is Kernel B.

DonJayamanne: I'd consider raising a seprate event and or having a seprate method to get the recommended kernel.
I believe you need this for instances where a kernel isn't seleced, but want to provide the right intellisense, if that's the case a seprate property/event would be better.

DonJayamanne (changes requested):

src/client/datascience/notebook/vscodeNotebookController.ts
rchiodo: No this is exactly what I want. 

This is what is happening now.
- User opens a notebook, kernel is associated with document, notebook selection change event fires
- User closes notebook
- User reopens notebook. Nothing fires, controller doesn't even know what the selection is. 

I think what should happen is the same thing as the first one
- User opens a notebook, kernel is associated with document, notebook selection change event fires

rchiodo: I need this so that on second open, the intellisense is using the correct (selected) kernel.

rchiodo: On first open, the reason it works is because the selection change event fires after setting affinity.

rchiodo: This code here:
https://github.com/microsoft/vscode-jupyter/blob/1247ee68b0179fddb828ebb1a104f76ca8807ce5/src/client/datascience/notebook/vscodeNotebookController.ts#L221

Which is triggered by this:
https://github.com/microsoft/vscode-jupyter/blob/1247ee68b0179fddb828ebb1a104f76ca8807ce5/src/client/datascience/notebook/vscodeNotebookController.ts#L144

On a second open that selection changed event doesn't fire.

DonJayamanne: Ok got it, but if you create a brand new notebook, then this event will still get fired even when no kernel is selected.
Today there's a bug in VS Code causing kernels to get auto selecfed, hence this will work.

I'll merge my fix tomorrow into VS Code and then we can re-visit this PR.
I still don't think this is right.

rchiodo: Okay trying a different way. We can just remember it like VS code does. When it gets set, stick it in a map.

PR #8220, opened 2021-11-10
Disable the one dotnet test on CI

DonJayamanne:
CI is busted because .NET interactive doesn't install
Conflict between .NET 5 & 6 (we install 5, but .NET interactive needs 6)


rchiodo:
I actually fixed this by installing a newer .NET in my last merge. Should be working now.

Issue #6982, opened 2021-08-06
Running cells in quick succession doesn't behave as expected


greazer:
Version: 1.60.0-insider (user setup)
Commit: a4eb4d688477fb9f2176bb2ed932b0b1726caed3
Date: 2021-08-05T21:34:13.867Z (2 hrs ago)
Electron: 13.1.7
Chrome: 91.0.4472.124
Node.js: 14.16.0
V8: 9.1.269.36-electron.0
OS: Windows_NT x64 10.0.22000

1. Take a script that looks like this:

```python
#%%
a = 1

#%%
b = a + 1

#%%
a = a + 1
```
2. Click Run Below 3 times in quick succession
3. Inspect the value of a. 

Actual
![image](https://user-images.githubusercontent.com/11685408/128437023-9a9f18b0-bbe4-416e-887b-5be4318eca95.png)

Expected
'a' should be 2, right?

The reason this is the case can be seen in the execution history of the Interactive window:
![image](https://user-images.githubusercontent.com/11685408/128437141-8c018823-a257-421e-b303-0528667510dc.png)

I would have expected the history to show 3 consecutive copies of the 3 cells in the script running. Rather than each cell being run 3 times in a row.

This may seem contrived, but I ran into it by accident and I suspect that real situations could arise where things get goofed up and the user doesn't realize it.

DonJayamanne:
Moving to triage, as i think this is a `important` and `golden` scenario issue. (basically pretty serious bug)

greazer:
Try to repro. May be fixed

rchiodo:
Still repros but only on first open. If the window is already open, then it behaves as expected.

rchiodo:
Also you have to click the link 3 times. If you click it once it works fine.

greazer:
I can repro all the time.

IanMatthewHuff:
Verification is pretty easy here. Just create a .py file with a few cell markers defined. In each cell just do a small timing loop to burn a second or two of time. Then trigger something like Run Below from the top code lens twice. In the old code you would get Cell A, Cell A, Cell B, Cell B, Cell C, Cell C. Now we get each command handled in order, so Cell A, Cell B, Cell C, Cell A, Cell B, Cell C.

PR #8233, opened 2021-11-10
Interactive Window Ordering Fix

IanMatthewHuff:
For #6982

### DIFF
--- a/src/client/datascience/editor-integration/codewatcher.ts
+++ b/src/client/datascience/editor-integration/codewatcher.ts
@@ -66,7 +66,6 @@ export class CodeWatcher implements ICodeWatcher {
     private codeLensUpdatedEvent: EventEmitter<void> = new EventEmitter<void>();
     private updateRequiredDisposable: IDisposable | undefined;
     private closeDocumentDisposable: IDisposable | undefined;
-    // private addCodeQueue: Promise<boolean>[] = [];
     private addCodeQueue: (() => Promise<boolean>)[] = [];
     private executingAddCode: Promise<boolean> | undefined;
 
--- a/src/client/datascience/editor-integration/codewatcher.ts
+++ b/src/client/datascience/editor-integration/codewatcher.ts
@@ -148,6 +148,7 @@ export class CodeWatcher implements ICodeWatcher {
 
         // Run all of our code lenses, they should always be ordered in the file so we can just
         // run them one by one
+        let finished = Promise.resolve(true);
         for (const lens of runCellCommands) {
             // Make sure that we have the correct command (RunCell) lenses
             let range: Range = new Range(
@@ -156,7 +157,6 @@ export class CodeWatcher implements ICodeWatcher {
                 lens.command!.arguments![3],
                 lens.command!.arguments![4]
             );
-            let finished = Promise.resolve(true);
             if (this.document) {
                 // Special case, if this is the first, expand our range to always include the top.
                 if (leftCount === runCellCommands.length) {
@@ -170,9 +170,9 @@ export class CodeWatcher implements ICodeWatcher {
                 // or if we do not we need to start it up as these commands are all expected to start a new history if needed
                 finished = this.addCode(code, this.document.uri, range.start.line, leftCount);
             }
-
-            await finished;
         }
+
+        await finished;
 
         // If there are no codelenses, just run all of the code as a single cell
         if (runCellCommands.length === 0) {
--- a/src/client/datascience/editor-integration/codewatcher.ts
+++ b/src/client/datascience/editor-integration/codewatcher.ts
@@ -3,7 +3,6 @@
 'use strict';
 import type * as nbformat from '@jupyterlab/nbformat';
 import { inject, injectable } from 'inversify';
-import { resolveModuleName } from 'typescript';
 import {
     CodeLens,
     commands,
@@ -67,10 +66,9 @@ export class CodeWatcher implements ICodeWatcher {
     private codeLensUpdatedEvent: EventEmitter<void> = new EventEmitter<void>();
     private updateRequiredDisposable: IDisposable | undefined;
     private closeDocumentDisposable: IDisposable | undefined;
-    private addCodeQueue: Promise<boolean> | undefined;
-    private addCodeQueue2: Promise<boolean>[] = [];
+    // private addCodeQueue: Promise<boolean>[] = [];
+    private addCodeQueue: (() => Promise<boolean>)[] = [];
     private executingAddCode: Promise<boolean> | undefined;
-    // private addCodeQueue: Promise<boolean> = Promise.resolve(true);
 
     constructor(
         @inject(IInteractiveWindowProvider) private interactiveWindowProvider: IInteractiveWindowProvider,
@@ -168,11 +166,12 @@ export class CodeWatcher implements ICodeWatcher {
 
                 // Note: We do a get or create active before all addCode commands to make sure that we either have a history up already
                 // or if we do not we need to start it up as these commands are all expected to start a new history if needed
-                const success = await this.addCode(code, this.document.uri, range.start.line);
-                if (!success) {
-                    await this.addErrorMessage(this.document.uri, leftCount);
-                    break;
-                }
+                // const success = await this.addCode(code, this.document.uri, range.start.line);
+                // if (!success) {
+                // await this.addErrorMessage(this.document.uri, leftCount);
+                // break;
+                // }
+                this.queueAddCode(code, this.document.uri, range.start.line, leftCount);
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/test/datascience/editor-integration/codewatcher.unit.test.ts
```diff
@@ -550,10 +553,14 @@ testing2`;
...
             )
-            .returns(() => Promise.resolve(true))
+            .returns(() => {
+                deferred.resolve();
+                return Promise.resolve(true);
+            })
             .verifiable(TypeMoq.Times.once());
 
         await codeWatcher.runAllCells();
+        await deferred.promise;
```
IanMatthewHuff, line 563: The unit test did change here. The command now returns when the code has all been queued, which I believe makes sense for the command.

src/client/datascience/editor-integration/codewatcher.ts
```diff
@@ -982,6 +972,56 @@ export class CodeWatcher implements ICodeWatcher {
...
+    private queueAddCode(
+        code: string,
+        file: Uri,
+        line: number,
+        leftCount: number,
+        editor?: TextEditor,
+        debug?: boolean
+    ) {
+        // Either start up this addCode, or queue it for later
+        if (this.executingAddCode) {
```
IanMatthewHuff, line 1017: I feel like TS / JS has a bunch of way to do this type of queuing and they all always look a bit janky to me. Speak up if it looks like I'm doing something unreasonable here, but this way felt the cleanest to me. (Had a promise chain thing I was initially working with that felt hard to read / debug).

IanMatthewHuff:
I fixed up the Unit tests. Will take a peek at what we currently have in the .vscode tests. If we don't have coverage for this situation there I'll add a test before submission for queuing multiple sets of cells and making sure they come out in order.

src/client/datascience/editor-integration/codewatcher.ts
```diff
@@ -163,11 +165,7 @@ export class CodeWatcher implements ICodeWatcher {
 
                 // Note: We do a get or create active before all addCode commands to make sure that we either have a history up already
                 // or if we do not we need to start it up as these commands are all expected to start a new history if needed
-                const success = await this.addCode(code, this.document.uri, range.start.line);
-                if (!success) {
-                    await this.addErrorMessage(this.document.uri, leftCount);
-                    break;
-                }
+                this.queueAddCode(code, this.document.uri, range.start.line, leftCount);
```
IanMatthewHuff, line 168: Previous code here just loops and awaits. So if you did run cells [A, B, C] and then run cells [A, B, C] you would get [A, A, B, B, C, C] in the output, not the expected [A, B, C, A, B, C]. Pretty easy to cause other ordering issues as well with the other cell commands.

src/client/datascience/editor-integration/codewatcher.ts
rchiodo: I think this is a bit of an overkill. It could just chain the promise. Like the interactive window does itself. 

Essentially you have a single promise - addCodePromise that you init to a resolved promise.

addCode changes to do something like so:

```
        // Chain execution promises so that cells are executed in the right order
        if (this.addCodePromise) {
            this.addCodePromise= this.addCodePromise.then(() =>
                this.addCodeImpl(code, fileUri, line, isDebug)
            );
        } else {
            this.addCodePromise= this.addCodeImpl(code, fileUri, line, isDebug);
        }
        try {
            return await this.addCodePromise;
        } catch (exc) {
            // Rethrow, but clear promise so we can execute again
            this.addCodePromise= undefined;
            throw exc;
        }
```

Old addCode method becomes addCodeImpl

Actually feels like we could have a method/service to do this cause we do this a lot.

rchiodo: This would also leave the old code the same as the behavior wouldn't have to account for a queue.

rchiodo: Trying to think of what that generic service might be. 

It would look like a function. A function which returns a promise. 

I wonder if we could do it with a decorator. That would be even better. Something like so:

```
@chainable
public addCode(...): Promise 
```

Where the chainable decorator would do the code I pasted above.

src/client/datascience/editor-integration/codewatcher.ts
```diff
@@ -1052,10 +1092,11 @@ export class CodeWatcher implements ICodeWatcher {
             if (this.document) {
                 // Use that to get our code.
                 const code = this.document.getText(currentRunCellLens.range);
-                await this.addCode(
+                this.queueAddCode(
```
DonJayamanne, line 1095: Isn't this the place where we need to chain the promises, the way Rich suggested.
Basically just chaning the promises and if it fails, then clear the promise (from property).
And I think this method should not  return a promise, instead it should be sync that way calling code will never await.

src/client/datascience/editor-integration/codewatcher.ts
DonJayamanne: I agree, and would suggest moving all of that into the existing `addCode` method, unless i'm reading it incorrectly

DonJayamanne: I suggest the addCode method as that the choke point, thats what everything calls today.
Perhaps rename from `addCode` to `queueAddCode` or the like and make it sync

DonJayamanne (commented):
.

src/client/datascience/editor-integration/codewatcher.ts
rchiodo: Then we could use that in both the interractive window and here (and probably ten other places where we chain stuff)

src/client/datascience/editor-integration/codewatcher.ts
rchiodo: I think we need to await because it checks the result? It then does something. Not sure what addError does.

src/client/datascience/editor-integration/codewatcher.ts
IanMatthewHuff: I had initially started with the promise chain starting from a resolved promise. Felt like it was getting icky with the promises resolving to false (instead of throwing) and the error handler function having to get tacked on. But I'm not adverse to going back. Let me restart from there and see if I can get it looking better.

rchiodo: I tried this on the current interactive window 'submitCodeImpl' and it seems to work. Then we can use this to do promise chaining anywhere:

```typescript
// Turn a method that returns a promise into something that uses a chain of promises
export function chainPromise() {
    return function (_target: Object, propertyKey: string, descriptor: TypedPropertyDescriptor<any>) {
        const originalMethod = descriptor.value;
        const chainedKey = `chained${propertyKey}`;

        // eslint-disable-next-line , @typescript-eslint/no-explicit-any
        descriptor.value = function (...args: any[]) {
            // Check for promise in the object.
            let currentValue = (this as any)[chainedKey] as Promise<any>;
            if (currentValue) {
                currentValue = currentValue.then(() => originalMethod.apply(this, args));
            } else {
                currentValue = originalMethod.apply(this, args);
            }

            // Save promise in object
            (this as any)[chainedKey] = currentValue;

            // If promise fails, clear it.
            return currentValue
                .then((r) => r)
                .catch((e) => {
                    (this as any)[chainedKey] = undefined;
                    throw e;
                });
        };

        return descriptor;
    };
}
```

This changes submitCodeImpl (or your addCode) to this instead:

```
@chainPromise()
    private async submitCodeImpl(code: string, fileUri: Uri, line: number, isDebug: boolean) {
    // Internals don't worry about chaining anymore
}
```

src/test/datascience/editor-integration/codewatcher.unit.test.ts
```diff
@@ -560,6 +560,81 @@ testing2`;
         document.verifyAll();
     });
 
+    test('Test two command ordering correct ordering', async () => {
```
IanMatthewHuff, line 563: Usually I like our vscode tests over unit, but one of these tests actually caught an issue in an earlier checkin. And the way they are set up they actually work pretty good for testing this scenario. I put this test in the old bits and it failed with the expected failure (trying to interleave instead of running each in order).

image

src/client/datascience/editor-integration/codewatcher.ts
IanMatthewHuff: The decorator I didn't like in this situation. Just in general I find them hard to debug (prefer them much more for extra stuff like telemetry / timing on a function as opposed to something like this) and the differences from the usage somewhere like submitCodeImpl made it hard to work in both situations.

src/client/datascience/editor-integration/codewatcher.ts
```diff
@@ -1003,6 +1032,15 @@ export class CodeWatcher implements ICodeWatcher {
             await this.dataScienceErrorHandler.handleError(err);
         }
 
+        if (!result) {
+            // If our cell result was a failure (but not an exception) show an error
+            // for the count of cells cancelled
+            await this.addErrorMessage(file, leftCount);
+
+            // Throw to break out of the promise chain
+            throw new InteractiveCellResultError();
```
IanMatthewHuff, line 1041: Slight difference from something like submitCodeImpl as we want to abort the entire chain not just for exceptions (which we want to bubble up) but also for a error in the code, which returns as a false value from .addCode.

DonJayamanne (approved):

src/client/datascience/editor-integration/codewatcher.ts
IanMatthewHuff: Hmm, thought I put a comment here. I didn't like the decorator for this (though your implementation is rad clever) it feels like really core functionality to be putting into a decorator (versus timing / telemetry) and not easy to debug and the mild differences between something like the addCode here and submitCode in the interactive window (things like the false return value handling) didn't lend to sharing one decorator.

rchiodo (approved):

PR #8237, opened 2021-11-10
Better error handling when kernel crashes due to overridden built in modules 

### LINKED ISSUE #5896: Cancel previous kernel connection when switching to another kernel
## Environment data

-   VS Code version: 1.57 Insiders
-   Jupyter Extension version (available under the Extensions sidebar): 2021.6.841665333
-   Python Extension version (available under the Extensions sidebar): N/A
-   OS (Windows | Mac | Linux distro) and version: macOS Big Sur
-   Python and/or Anaconda version: N/A
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A
-   Jupyter server running: Local | Remote | N/A

## Expected behaviour

* Make sure you don't have Python installed
* Select a kernel which requires Python ext
* Run cells
* A notification asks me to install Python, cancel this notification
* Switch to another kernel
* Cell exection should be canceled now

## Actual behaviour

* Cell execution is in pending state
* Notification of connecting to previous kernel is not dismissed (I can't abort it from the notification either)
image

## Steps to reproduce:

1. XXX

## Logs

<details>

<summary>Output for <code>Jupyter</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Jupyter</code>)
</summary>

<p>

```
XXX
```

</p>
</details>

### LINKED ISSUE #6791: Switching kernel shows two "Connecting to kernel" status
Issue Type: <b>Bug</b>

Switching kernel now renders two "Connecting to kernel" sys info.

Extension version: 2021.8.1055544705
VS Code version: Code - Insiders 1.59.0-insider (04485223ebc6a075f39966bf9b3e261359ac7818, 2021-07-23T19:21:12.120Z)
OS version: Windows_NT x64 10.0.19041
Restricted Mode: No
Remote OS version: Linux x64 4.19.128-microsoft-standard

### LINKED ISSUE #8195: Recommend renaming python file if python file it overwrites builtins
We find that a lot of users run into this when creating python files using reserved names like `sys.py/os.py/random.py`.
As a result Python runtime will import the user defined files and it could fall over.

**Solution:**
If kernel startup fails and we detect such a scenario then recommend renaming that file.

DonJayamanne:
Fixes #8195
Fixes #5896
Fixes #6791

* Display message prompting user to rename file if kernel fails to start due to some funky file in workspace
* Better handling of cancellation (if user stops kernel startup, then cancel all of the kernel startup code early)
	* Came across this when testing kernel failures (constantly running & causing cells to crash/fail)
	* Now cells stop running immediately when you decide to switch kernels & things are killed much earlier/quicker (previously we'd wait to start the kernel & then kill it)


pythonFiles/vscode_datascience_helpers/kernel_launcher.py
```diff
@@ -11,9 +11,6 @@
 # See comment at the point of our use of Popen
 from subprocess import Popen, PIPE  # nosec
 
-from ipython_genutils.encoding import getdefaultencoding
-from ipython_genutils.py3compat import cast_bytes_py2
```
DonJayamanne, line 15: This is no longer required, was only required for python 2.7
Also ipython_genutils is no longer shipping in conda

src/client/common/errors/errorUtils.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+import { WorkspaceFolder } from 'vscode';
```
rchiodo, line 4: Did these all move here from somewhere else?

src/client/datascience/errors/errorHandler.ts
```diff
@@ -20,14 +20,66 @@ import { KernelDiedError } from './kernelDiedError';
...
+        await this.handleErrorImplementation(err);
+    }
+
+    public async handleKernelError(
+        err: Error,
+        purpose: 'start' | 'restart' | 'interrupt' | 'execution',
+        kernelConnection: KernelConnectionMetadata
+    ): Promise<void> {
+        await this.handleErrorImplementation(err, purpose, async (error: BaseError, defaultErrorMessage: string) => {
+            const failureInfo = analyseKernelErrors(
```
rchiodo, line 52: Is this another goofy spelling difference? Analyze is with a 'z' afaik.

src/test/datascience/.vscode/settings.json
```diff
@@ -11,7 +11,7 @@
     "python.linting.mypyEnabled": false,
     "python.linting.banditEnabled": false,
     "python.formatting.provider": "yapf",
-    "python.pythonPath": "python",
```
rchiodo, line 14: Probably dont want to submit this. Although I think the test just overwrites it each time.

DonJayamanne:
> Did these all move here from somewhere else?

Moved from `errors.ts`, but some of the code is new

> Is this another goofy spelling difference? Analyze is with a 'z' afaik.

Damn it, sorry, will change.

> Probably dont want to submit this. Although I think the test just overwrites it each time.

Oops.

rchiodo (approved):

Issue #8185, opened 2021-11-05
Remove starting of a session and disposing it when connecting to remote Jupyter servers


DonJayamanne:
When connecting to remote jupyter servers we create sessions:

```typescript
        // Try creating a session just to ensure we're connected. Callers of this function check to make sure jupyter
        // is running and connectable.
        const session = (await this.sessionManager.startNew(
            undefined,
            defaultKernel,
            launchInfo.connectionInfo.rootDirectory,
            cancelToken,
            launchInfo.disableUI
        )) as JupyterSession;
        const idleTimeout = this.configService.getSettings().jupyterLaunchTimeout;
        // The wait for idle should throw if we can't connect.
        await session.waitForIdle(idleTimeout);
        await session.dispose();
```

If we want to check if the jupyter server is running & connectable, why not use an API such as getting a list of the kernels.
This feels unnecessary. 

E.g. if we cannot start the default kernel then we cannot connect at all (& we have a few issues where users connecting to remote servers cannot start the default kernel). I too ran into this a few months ago, admittedly its because my default python kernel was busted. 

I think we can just remove this & test connectivity by simply calling the API to get a list of kernels/sessions or similar API.

misolori:
Ran into this bug today and Don was able to isolate the issue for me and unblock me. In this case there were no notifications or any indication that something was broken. After re-installing conda I was able to see the `Failed to start the kernel` error.

DonJayamanne:
Done.

PR #8238, opened 2021-11-11
Stop starting default kernel when starting jupyter

DonJayamanne:
For #8185
Miguel ran into this today, updated Mac to Monterey, and then things stopped working, then deleted envioronments, and the default kernlespec was pointing to a non-existent Python and there was nothing he could do.

Temporary work around was for him to update kernelspec.json to point to `python` instead of the non-existent conda env.
Btw - this wasn't a kernelspec we created.

### DIFF
--- a/src/client/datascience/jupyter/kernels/types.ts
+++ b/src/client/datascience/jupyter/kernels/types.ts
@@ -58,24 +58,6 @@ export type KernelSpecConnectionMetadata = Readonly<{
     id: string;
 }>;
 /**
- * Connection metadata for Kernels started using default kernel.
- * Here we tell Jupyter to start a session and let it decide what kernel is to be started.
- * (could apply to either local or remote sessions when dealing with Jupyter Servers).
- */
-export type DefaultKernelConnectionMetadata = Readonly<{
-    /**
-     * This will be empty as we do not have a kernel spec.
-     * Left for type compatibility with other types that have kernel spec property.
-     */
-    kernelSpec?: IJupyterKernelSpec;
-    /**
-     * Python interpreter will be used for intellisense & the like.
-     */
-    interpreter?: PythonEnvironment;
-    kind: 'startUsingDefaultKernel';
-    id: string;
-}>;
-/**
  * Connection metadata for Kernels started using Python interpreter.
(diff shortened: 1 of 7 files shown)


src/client/datascience/jupyter/liveshare/hostJupyterServer.ts
```diff
@@ -175,24 +175,6 @@ export class HostJupyterServer implements INotebookServer {
         this.sessionManager = (await this.sessionManagerFactory.create(
             launchInfo.connectionInfo
         )) as JupyterSessionManager;
-
-        const defaultKernel: DefaultKernelConnectionMetadata = {
-            kind: 'startUsingDefaultKernel',
-            id: ''
-        };
-        // Try creating a session just to ensure we're connected. Callers of this function check to make sure jupyter
-        // is running and connectable.
```
rchiodo, line 184: ```Callers of this function check to make sure jupyter is running and connectable```

I assume this isn't true anymore (or is verified by other means)? I think this comment explains why we did this originally.

rchiodo (approved):

src/client/datascience/jupyter/liveshare/hostJupyterServer.ts
DonJayamanne: We'll fail when we try to start a kernel or the like.
At that point we'll have some information.
Today nothing works, basically things fall over silently, this way, even if jupyter server isn't working, we can at least try to run a kernel & we'll get errors at that point in time.

DonJayamanne:
@rchiodo just realized that we don't need `DefaultKernelConnectionMetadata`
I deleted that, hence a larger change in case you want to re-review.

src/client/datascience/jupyter/kernels/types.ts
```diff
@@ -57,24 +57,6 @@ export type KernelSpecConnectionMetadata = Readonly<{
     kind: 'startUsingKernelSpec';
     id: string;
 }>;
-/**
```
rchiodo, line 60: Yay! Love deleted code

rchiodo (approved):

Issue #8080, opened 2021-10-28
Notebooks - Intellisense not helpful if using . to select pandas df column


claudiaregio:
If i use the . to select a column in a pandas dataframe and then use . again to apply some function to it, the intellisense pop up is not helpful as it does not show the function in full 

![image](https://user-images.githubusercontent.com/19276747/139167741-b264108a-e9f8-4626-9e3b-d8aa28ca8dff.png)

Compared to using the [ ] to select a column 
![image](https://user-images.githubusercontent.com/19276747/139168085-9bb2beb2-a97e-4359-83ec-330adc3dd828.png)

rchiodo:
This is the label from Jupyter being super long. Going to change this anyway to not be so long. That change will fix this bug.

rchiodo:
Repro steps:
- Clone https://github.com/claudiaregio/data-science.git
- Create conda environment using the 'golden_scenario_env.yml'
- Open in VS code
- Open titanic.ipynb
- Switch to the golden_scenario kernel
- Run the first two cells
- Create a new cell
- type in ```titanic_df.``` and wait a bit (jupyter provides column names so might have to wait a bit)
- if you get ```Embarked``` in the autocomplete, hit tab and then ```.``` again
- verify the auto complete is only the next word, and doesn't have ```titanic.Embarked.``` embedded in it.

rchiodo:
Should look something like so:

![image](https://user-images.githubusercontent.com/19672699/144643446-788e75e3-1610-4486-b913-365d0c16a613.png)

IanMatthewHuff:
Looks good.
![image](https://user-images.githubusercontent.com/812783/144680974-3de964d8-42a6-436e-8c15-d8021a6fc53d.png)

Issue #7880, opened 2021-10-10
Jupyter notebook autocomplete suggestion "%%!"


jamesoflol:
Issue Type: <b>Bug</b>

Starting docstring (''') or a function definition causes autocomplete to suggest "%%!". It means I need to ESC away from it constantly, or it chucks "%%!" in my code.

Someone else reported this issue here: https://stackoverflow.com/questions/69134016/vscode-jupyter-strange-autocomplete

Extension version: 2021.9.1001312534
VS Code version: Code 1.61.0 (ee8c7def80afc00dd6e593ef12f37756d8f504ea, 2021-10-07T18:13:09.652Z)
OS version: Windows_NT x64 10.0.19041
Restricted Mode: No

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|AMD Ryzen 5 3600 6-Core Processor               (12 x 4229)|
|GPU Status|2d_canvas: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>rasterization: enabled<br>skia_renderer: enabled_on<br>video_decode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|31.93GB (25.78GB free)|
|Process Argv|--crash-reporter-id 7693d6ac-337e-4f4a-a5f3-9b5534313163|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vsreu685:30147344
python383:30185418
pythonvspyt602:30300191
vspor879:30202332
vspor708:30202333
vspor363:30204092
vstes627:30244334
pythonvspyt639:30300192
pythontb:30283811
pythonvspyt551:30345470
pythonptprofiler:30281270
vshan820:30294714
vstes263:30335439
vscorecescf:30358481
pythondataviewer:30285071
pythonvsuse255:30340121
vscod805:30301674
pythonvspyt200:30340761
binariesv615:30325510
vsccppwtct:30378365
pythonvssor306:30344512
bridge0708:30335490
pygetstartedt2:30371810
dockerwalkthru:30377721
bridge0723:30353136
pythonrunftest32:30373476
pythonf5test824:30373475
javagetstartedt:30364666
pythonvspyt187:30373474
pydsgsc2:30361791
vsqsis200:30374795
[...]

Issue #7772, opened 2021-10-01
Jupyter IntelliSense contain error ICONS and duplicate content.


thefatwhale:
## Environment data

-   VS Code version: 1.60.2
-   Jupyter Extension version (available under the Extensions sidebar): v2021.8.2041215044
-   Python Extension version (available under the Extensions sidebar): v2021.9.1246542782
-   OS (Windows | Mac | Linux distro) and version: Windows 7 x64 SP1
-   Python and/or Anaconda version: Python 3.8.7 and 3.8.8
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): venv
-   Jupyter server running: Local

### display IntelliSense correctly.

1. display `lib` once and with a correct icon.
2. display keyword `as` with keyword icon (not function/method icon).
3. display method once and remove the content of the front.

## Actual behaviour

3 screenshot

![shot1](https://user-images.githubusercontent.com/50515732/135592407-75430f8f-f751-448b-a3dd-74985b34b10b.png)

![image](https://user-images.githubusercontent.com/50515732/135594151-bcaf5927-9735-460a-9fbe-c53182bc255e.png)

![image](https://user-images.githubusercontent.com/50515732/135594197-d49d5116-10f2-4e31-9c68-7fad8d180f3c.png)

claudiaregio:
Thanks @thefatwhale, will take a look at our next triage meeting

DonJayamanne:
I believe this is caused by the intellisense we get from Jupyter as well.
We should look into remove these duplicates (in jupyter we can call VSC completions API to get the completions for that position & when we get response from both, remove duplicates, if we get response just from jupyter, return that item so that we don't slow down intellisense).

rchiodo:
@DonJayamanne had an idea on how to remove dupes.

- During jupyter completion call, call for other completions
- Filter out the ones that are already there

cplager:
[Defauolt value should NOT be suggested when entering list](https://github.com/microsoft/vscode-jupyter/issues/7773) was closed as a duplicate to this problem.  Just wanted to make sure that this problem is also on people's radar (as it is neither an issue with duplicates nor error icons).

Issue #7136, opened 2021-08-14
Auto completion is suggesting completely invalid suggestions.


greazer:
Version: 1.60.0-insider (user setup)
Commit: 58454f7bec838f79fe2422ab5914419147d82e40
Date: 2021-08-13T05:15:29.421Z
Electron: 13.1.8
Chrome: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Windows_NT x64 10.0.22000

Jupyter: v2021.8.2027072902
Pylance: v2021.8.2-pre.1

1. Create a new ipynb that has nothing to do with another that exists in the same workspace.
2. In the new notebook type 
```python 
import pandas as pd
```

Notice that ```pd``` gets offered to be autocompleted as one of the files in the workspace. This is totally invalid. This doesn't happen in a py file in the same workspace following the same procedure. 

I'm also not sure whether this bug is a Jupyter or  PyLance problem, though I'm guessing Jupyter at this point.

![recording (6)](https://user-images.githubusercontent.com/11685408/129452889-4ced427e-3ef6-4a94-89c3-dd97a54b74c7.gif)

greazer:
The autocomplete is coming from the Jupyter Kernel. Would be great to find a way to make this a more reasonable experience. 

Perhaps only accept jupyter kernel autocomplete when specific characters are typed... like '.' or "'", etc.

rchiodo:
Autocomplete from jupyter outside of a string that returns file paths seems useless. That would alleviate this problem.

We could filter based on if inside a string or not.

PR #8247, opened 2021-11-11
Improve jupyter kernel autocompletions

rchiodo:
Fixes: #7136 
Fixes: #7772 
Fixes: #7880
Fixes: #8080

Filter completions from a jupyter kernel so that:
- Strings (by default) don't show anything
- Strings can show file names (if ```'``` or ```"```) are trigger characters
- File names are skipped everywhere else
- Magics support autocomplete
- Duplicates from the real completions are skipped
- Length of the completions are shortened to only match the last part of a object path


package.json
```diff
@@ -1865,6 +1865,12 @@
                     "default": false,
                     "description": "Determines if pylance's experimental notebook support is used or not",
                     "scope": "machine"
+                },
+                "jupyter.completionTriggerCharacters": {
+                    "type": "string",
+                    "default": ".%",
```
DonJayamanne, line 1871: Isn't this specific to just python.

src/client/common/configSettings.ts
```diff
@@ -98,6 +98,7 @@ export class JupyterSettings implements IWatchableJupyterSettings {
     public showVariableViewWhenDebugging: boolean = true;
     public newCellOnRunLast: boolean = true;
     public pylanceHandlesNotebooks: boolean = false;
+    public completionTriggerCharacters: string = '.%';
```
DonJayamanne, line 101: We can leave this empty and let the vscode settings apply the defaults. This way defaults are in one place

src/client/datascience/notebook/integration.ts
```diff
@@ -5,9 +5,12 @@ import { inject, injectable } from 'inversify';
 import { languages } from 'vscode';
 import { IExtensionSingleActivationService } from '../../activation/types';
 import { NOTEBOOK_SELECTOR } from '../../common/constants';
-import { IDisposableRegistry } from '../../common/types';
+import { IConfigurationService, IDisposableRegistry } from '../../common/types';
 import { JupyterCompletionProvider } from './intellisense/jupyterCompletionProvider';
 
+// Default set of trigger characters for jupyter
+const DefaultTriggerCharacters = ['.', ',', '%'];
```
DonJayamanne, line 12: Different from the default value in settings

src/client/datascience/notebook/intellisense/jupyterCompletionProvider.ts
```diff
@@ -157,9 +192,43 @@ export class JupyterCompletionProvider implements CompletionItemProvider {
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+function positionInsideString(word: string, position: Position) {
+    const indexDoubleQuote = word.indexOf('"');
+    const indexSingleQuote = word.indexOf("'");
```
DonJayamanne, line 223: We might want to exclude `"""` (docstrings)

src/client/datascience/notebook/intellisense/jupyterCompletionProvider.ts
```diff
@@ -61,22 +76,29 @@ export class JupyterCompletionProvider implements CompletionItemProvider {
         const timeout =
```
DonJayamanne, line 76: Shouldn't we exclude pylance completions if we're dealing with a Julia, or R or other cell.

src/client/datascience/notebook/intellisense/jupyterCompletionProvider.ts
```diff
@@ -157,9 +192,43 @@ export class JupyterCompletionProvider implements CompletionItemProvider {
...
         };
     }
+
+    private async getPylanceCompletions(
+        document: TextDocument,
+        position: Position,
+        context: CompletionContext,
+        cancelToken: CancellationToken
+    ) {
+        if (document.notebook) {
```
DonJayamanne, line 202: Should probably return [] if language is not python.

DonJayamanne (changes requested):
* I'd consider renaming the setting to be specific to Python.
* Also ensure we support completions for non-python without pylance & any special parsing

package.json
rchiodo: Everything is specific to python. We only register a completion provider for python cells.

src/client/common/configSettings.ts
rchiodo: Sure good idea.

src/client/datascience/notebook/intellisense/jupyterCompletionProvider.ts
rchiodo: Skipping this for now. Doc strings are not going to work the same (as they are likely on a different line)

rchiodo: The default is no completions for strings anyway.

src/client/datascience/notebook/intellisense/jupyterCompletionProvider.ts
rchiodo: They wouldn't come back anyway.

rchiodo: The languageClient would skip them as they don't match it's document filter.

rchiodo: And since this completion provider is only registered for python cells, it doesn't get called in that scenario anyway.

src/client/datascience/notebook/intellisense/jupyterCompletionProvider.ts
rchiodo: Document selector already handles this.

src/client/datascience/notebook/integration.ts
rchiodo: Oops

package.json
DonJayamanne: In that case shouldn't this setting be named `jupyter.pythonCOmpletionTriggerCharacters`
As this doesn't apply to Julia or R

DonJayamanne: ignore my typos

DonJayamanne (approved):

src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
```diff
@@ -0,0 +1,322 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
rchiodo, line 1: Blech github is crappy with renames.

src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
```diff
@@ -0,0 +1,322 @@
...
+        position: Position,
+        context: CompletionContext,
+        cancelToken: CancellationToken
+    ) {
+        if (document.notebook) {
+            const client = await this.languageClientProvider.getLanguageClient(document.notebook);
+            if (client) {
+                // Use provider so it gets translated by middleware
+                const feature = client.getFeature(lsp.CompletionRequest.method);
+                const provider = feature.getProvider(document);
```
rchiodo, line 210: This line here would fail if the document selector didn't match. Meaning Julia wouldn't go through here.

PR #8256, opened 2021-11-12
fix newCellOnRunLast

janosh:
Addresses https://github.com/microsoft/vscode-jupyter/pull/7995#issuecomment-967401508.

Should we add a test? Looked at `codewatcher.unit.test.ts` but it's a bit over my head tbh.

### DIFF
--- a/src/client/datascience/editor-integration/codewatcher.ts
+++ b/src/client/datascience/editor-integration/codewatcher.ts
@@ -1035,7 +1035,7 @@ export class CodeWatcher implements ICodeWatcher {
             // Move the next cell if allowed.
             if (advance) {
                 const editor = this.documentManager.activeTextEditor;
-                const { newCellOnRunLast } = this.configService.getSettings(editor.document.uri);
+                const { newCellOnRunLast } = this.configService.getSettings(editor?.document.uri);
                 if (nextRunCellLens) {
                     this.advanceToRange(nextRunCellLens.range);
                 } else if (newCellOnRunLast && editor) {
--- a/src/client/datascience/editor-integration/codewatcher.ts
+++ b/src/client/datascience/editor-integration/codewatcher.ts
@@ -942,10 +942,7 @@ export class CodeWatcher implements ICodeWatcher {
         // ```
         //
         const cellDelineator = this.getDefaultCellMarker(editor.document.uri);
-        let newCell = `${cellDelineator}\n\n`;
-        if (line >= editor.document.lineCount) {
-            newCell = `\n${cellDelineator}\n`;
-        }
+        const newCell = line >= editor.document.lineCount ? `\n${cellDelineator}\n` : `${cellDelineator}\n\n`;
 
         const cellStartPosition = new Position(line, 0);
         const newCursorPosition = new Position(line + 1, 0);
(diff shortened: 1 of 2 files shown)


rchiodo:
> Should we add a test?

There's actually already a test for that command. You just need to have it run with your setting.

Test is this one:

```Test the RunCurrentCell command```

I believe if you add code like this:

```
        jupyterSettings.newCellOnRunLast= false;
        jupyterSettings.fireChangeEvent();

```

At the top of that test, it will fail.

**Comment by rchiodo** [comment]
It looks like there's also a build error:

```
src/client/datascience/editor-integration/codewatcher.ts(1038,77): error TS2532: Object is possibly 'undefined'.
```

**Comment by janosh** [comment]
@rchiodo I saw the `RunCurrentCell` test but couldn't figure out how to assert no new cell is added when calling `await codeWatcher.runCurrentCell();`.

**Comment by rchiodo** [comment]
Sorry @janosh I didn't realize you updated this.

The test I suggested is actually the wrong one. That test doesn't advance.

This one does: ```Test the RunCellAndAdvance command with next cell```

You could copy that test but set your setting first. 

Then on this line here:
https://github.com/microsoft/vscode-jupyter/blob/4ebc068416829dcebf064b5c848fa386bc05a030/src/test/datascience/editor-integration/codewatcher.unit.test.ts#L1036

You'd verify that it didn't get called. Like set a local variable to true and then assert that it's false after calling runCurrentCellAndAdvance.

**Comment by janosh** [comment]
@rchiodo Thanks for the guidance. That definitely helped! I copied and modified the `Test the RunCellAndAdvance command with next cell` test but wasn't sure about a few things. Does the new test require all the same setup and teardown? Like

```ts
// Verify function calls
textEditor.verifyAll();
activeInteractiveWindow.verifyAll();
document.verifyAll();
```

and

```ts
// Set up our expected calls to add code
activeInteractiveWindow
    .setup((h) =>
        h.addCode(
            TypeMoq.It.isValue('#%%\ntesting1'),
            TypeMoq.It.isValue(fileName),
            TypeMoq.It.isValue(0),
            TypeMoq.It.is((ed: TextEditor) => {
                return textEditor.object === ed;
            }),
            TypeMoq.It.isAny()
        )
    )
    .returns(() => Promise.resolve(true))
    .verifiable(TypeMoq.Times.once());
```

janosh:
@rchiodo One remaining CI issue. TypeScript complains:

> Test runCurrentCellAndAdvance command does not advance when newCellOnRunLast is false:1 | 'set' on proxy: trap returned falsish for property 'selection'
> Result: 913 tests run, 25 skipped, 1 failed.

Is it a problem with this code?

```ts
        // For this test we need to set up a document selection point
        const selection = new Selection(0, 0, 0, 0);
        textEditor.setup((te) => te.selection).returns(() => selection);
```

rchiodo:
Hmm, not sure why I'm not getting your updates, but I noticed today. 

I pushed a fix for the problem. There was some quirkiness about our overrides of the textEditor.selection (seems you can't have a typemoq for a getter and a setter). I also changed the code in the test to only have a single cell so it would force a new cell to be added.

rchiodo (approved):

rchiodo:
Thanks so much for following up @janosh. Should be able to push this today.

Issue #7849, opened 2021-10-07
Slow Kernel startup/restart


claudiaregio:
![image](https://user-images.githubusercontent.com/19276747/136306361-ed4ac6cf-3236-4bda-aa18-d6e0c2cb2074.png)

greazer:
This is one of those issues that can be a killer for retention. It falls in line with a theme of improving our kernel startup story in October.

DonJayamanne:
See here https://github.com/microsoft/vscode-jupyter/issues/6702 & https://github.com/microsoft/vscode-jupyter/issues/7855
I've already started looking into this and found plenty of places that can be improved.

rchiodo:
Not really verifiable. Going to have to wait for user feedback.

de-loots1994:
Still an issue, the largest issue that the behaviour is unpredictable. In our dev team for some people it takes seconds others it doesn't start up at all and keeps saying:
![image](https://user-images.githubusercontent.com/115618999/204757906-e16ef0b5-b46b-489e-b6a1-4e94ae99a69d.png)

DonJayamanne:
> Still an issue, the largest issue that the behaviour is unpredictable. In our dev team for some people it takes seconds others it doesn't start up at all and keeps saying:

I'm sorry about this, please could you file an issue with the details requested in the issue template (i.e. provide the logs)

de-loots1994:
@DonJayamanne our issue seems to be resolved it appears that it has to do with the ipykernel not being active, and with pip install it goes wrong but when getting the pop up from vs code e.g:
![image](https://user-images.githubusercontent.com/115618999/204808733-cbf55a84-204f-4963-8ed1-83f611e8e388.png)

It works...

PR #8261, opened 2021-11-13
Use Object instead of using a flag to hide startup progress UI

DonJayamanne:
Part of #7849

Goal of using an object instead of a boolean flag:
* Assume we auto start a kernel & we pass boolean flag to hide the UI
* Now while auto starting, user starts running a cell
* It will use the same kernel (code) that's being auto started & since that is still in progress no UI is dispalyed

Solution:
* Change the flag to an object
* When user runs cell (kernel.start) we change the state of this object to `true` and all subsequent code will now start displaying the UI (I've added an event handler so that we can start display progress where it was skipped)
* When kernel startup fails because IPYkernel is not installed, we'll know whether it was displayed or not and we can then re-display the error message if the user runs cells just after the failure takes place

### DIFF
--- a/src/client/datascience/interactive-common/notebookServerProvider.ts
+++ b/src/client/datascience/interactive-common/notebookServerProvider.ts
@@ -14,6 +14,7 @@ import { noop } from '../../common/utils/misc';
 import { IInterpreterService } from '../../interpreter/contracts';
 import { sendTelemetryEvent } from '../../telemetry';
 import { Settings, Telemetry } from '../constants';
+import { DisplayOptions } from '../displayOptions';
 import { JupyterInstallError } from '../errors/jupyterInstallError';
 import { JupyterSelfCertsError } from '../errors/jupyterSelfCertsError';
 import { JupyterServerSelector } from '../jupyter/serverSelector';
@@ -30,7 +31,7 @@ import {
 @injectable()
 export class NotebookServerProvider implements IJupyterServerProvider {
     private serverPromise: Promise<INotebookServer | undefined> | undefined;
-    private allowingUI = false;
+    private ui = new DisplayOptions(true);
     constructor(
         @inject(ProgressReporter) private readonly progressReporter: ProgressReporter,
         @inject(IConfigurationService) private readonly configuration: IConfigurationService,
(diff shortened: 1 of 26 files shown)


src/client/datascience/jupyter/jupyterNotebookProvider.ts
```diff
@@ -20,21 +21,40 @@ export class JupyterNotebookProvider implements IJupyterNotebookProvider {
...
     }
 
     public async connect(options: ConnectNotebookProviderOptions): Promise<IJupyterConnection | undefined> {
-        const server = await this.serverProvider.getOrCreateServer(options);
-        return server?.getConnectionInfo();
+        const ui = new DisplayOptions(options.disableUI === true);
+        try {
+            const server = await this.serverProvider.getOrCreateServer({
+                getOnly: false,
+                ui: new DisplayOptions(options.disableUI === true),
```
rchiodo, line 43: This seems like a typo. Shouldn't it use the variable?

src/client/datascience/interactive-common/notebookServerProvider.ts
```diff
@@ -61,7 +62,8 @@ export class NotebookServerProvider implements IJupyterServerProvider {
     ): Promise<INotebookServer | undefined> {
         // When we finally try to create a server, update our flag indicating if we're going to allow UI or not. This
         // allows the server to be attempted without a UI, but a future request can come in and use the same startup
-        this.allowingUI = options.disableUI ? this.allowingUI : true;
+        this.ui.disableUI = options.ui.disableUI;
```
rchiodo, line 65: Instead of having the event, could you just set the ```this.ui``` to the ```options.ui```? Then it would change when the options changed. Wouldn't need the dispose then.

rchiodo: Never mind. This could be changed to not use the event, but you might need the event later.

rchiodo (approved):

src/client/datascience/jupyter/jupyterNotebookProvider.ts
DonJayamanne: No, connect is one of those weird methods that can be called to pre-warm. 
There are juts two public methods that can be called `connect` & `kernel.start`.

I'm still not sure what should be done here. When I work on displaying the UI, i'll ensure to test this & make the changes accordingly.

PR #8278, opened 2021-11-16
Display progress indicator when starting kernel due to user action even when auto start is in progress

DonJayamanne:
* Ensure we display the progress UI when user runs a cell & when we have auto start enabled
* Ensure we pass cancellenation token to all code paths that deal with kernel startup
	* This helps a lot disposing objects early on (progress indicators, kernels, etc)
	 
* No tests, not sure how to add tests - this is something that just needs to be tested.
*

### DIFF
--- a/src/client/datascience/baseJupyterSession.ts
+++ b/src/client/datascience/baseJupyterSession.ts
@@ -6,7 +6,7 @@ import type { JSONObject } from '@lumino/coreutils';
 import type { Slot } from '@lumino/signaling';
 import { Observable } from 'rxjs/Observable';
 import { ReplaySubject } from 'rxjs/ReplaySubject';
-import { Event, EventEmitter } from 'vscode';
+import { CancellationTokenSource, Event, EventEmitter } from 'vscode';
 import { WrappedError } from '../common/errors/types';
 import { disposeAllDisposables } from '../common/helpers';
 import { traceInfo, traceInfoIfCI, traceVerbose, traceWarning } from '../common/logger';
@@ -76,7 +76,7 @@ export abstract class BaseJupyterSession implements IJupyterSession {
     protected onStatusChangedEvent = new EventEmitter<KernelMessage.Status>();
     protected statusHandler: Slot<ISessionWithSocket, KernelMessage.Status>;
     protected connected: boolean = false;
-    protected restartSessionPromise: Promise<ISessionWithSocket> | undefined;
+    protected restartSessionPromise?: { token: CancellationTokenSource; promise: Promise<ISessionWithSocket> };
     private _session: ISessionWithSocket | undefined;
     private _kernelSocket = new ReplaySubject<KernelSocketInformation | undefined>();
     private ioPubEventEmitter = new EventEmitter<KernelMessage.IIOPubMessage>();
@@ -110,7 +110,9 @@ export abstract class BaseJupyterSession implements IJupyterSession {
                 await this.shutdownSession(this.session, this.statusHandler, false);
                 traceVerbose('Shutdown session - get restart session');
                 if (this.restartSessionPromise) {
-                    const restartSession = await this.restartSessionPromise;
+                    this.restartSessionPromise.token.cancel();
+                    const restartSession = await this.restartSessionPromise.promise;
+                    this.restartSessionPromise.token.dispose();
                     traceVerbose('Shutdown session - shutdown restart session');
                     await this.shutdownSession(restartSession, undefined, true);
                 }
@@ -182,7 +184,7 @@ export abstract class BaseJupyterSession implements IJupyterSession {
             // keep the old session (user could be restarting for a number of reasons).
 
             // Just switch to the other session. It should already be ready
-            const newSession = await this.restartSessionPromise;
+            const newSession = await this.restartSessionPromise.promise;
             this.setSession(newSession);
 
             if (newSession.kernel) {
@@ -192,6 +194,7 @@ export abstract class BaseJupyterSession implements IJupyterSession {
                 // Rewire our status changed event.
                 newSession.statusChanged.connect(this.statusHandler);
             }
+            this.restartSessionPromise.token.dispose();
             this.restartSessionPromise = undefined;
             traceInfo('Started new restart session');
             if (oldStatusHandler && oldSession) {
--- a/src/client/datascience/errors/errorHandler.ts
+++ b/src/client/datascience/errors/errorHandler.ts
@@ -12,8 +12,13 @@ import { JupyterInstallError } from './jupyterInstallError';
 import { JupyterSelfCertsError } from './jupyterSelfCertsError';
 import { getLanguageInNotebookMetadata } from '../jupyter/kernels/helpers';
 import { isPythonNotebook } from '../notebook/helpers/helpers';
-import { IDataScienceErrorHandler, IJupyterInterpreterDependencyManager } from '../types';
-import { CancellationError as VscCancellationError, ConfigurationTarget } from 'vscode';
+import {
+    IDataScienceErrorHandler,
+    IJupyterInterpreterDependencyManager,
+    IKernelDependencyService,
+    KernelInterpreterDependencyResponse
+} from '../types';
+import { CancellationError as VscCancellationError, CancellationTokenSource, ConfigurationTarget } from 'vscode';
 import { CancellationError } from '../../common/cancellation';
 import { KernelConnectionTimeoutError } from './kernelConnectionTimeoutError';
 import { KernelDiedError } from './kernelDiedError';
@@ -27,9 +32,10 @@ import {
 } from '../../common/errors/errorUtils';
 import { KernelConnectionMetadata } from '../jupyter/kernels/types';
 import { getDisplayPath } from '../../common/platform/fs-paths';
-import { IBrowserService, IConfigurationService } from '../../common/types';
+import { IBrowserService, IConfigurationService, Resource } from '../../common/types';
 import { Telemetry } from '../constants';
 import { sendTelemetryEvent } from '../../telemetry';
+import { DisplayOptions } from '../displayOptions';
 
 @injectable()
 export class DataScienceErrorHandler implements IDataScienceErrorHandler {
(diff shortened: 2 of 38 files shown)


src/client/datascience/baseJupyterSession.ts
```diff
@@ -110,7 +110,9 @@ export abstract class BaseJupyterSession implements IJupyterSession {
                 await this.shutdownSession(this.session, this.statusHandler, false);
                 traceVerbose('Shutdown session - get restart session');
                 if (this.restartSessionPromise) {
-                    const restartSession = await this.restartSessionPromise;
+                    this.restartSessionPromise.token.cancel();
```
DonJayamanne, line 113: Now we have cancellation tokens to kill restart sessions much faster (previously we'd have to wait for that to complete)

DonJayamanne: THis is just a by product, i'm passing cancellation tokens when starting kernels.
Basically kernel.start & notebook.connect method are called in 3 places & all will have a cancellation token & that makes things better

src/client/datascience/errors/errorHandler.ts
```diff
@@ -151,8 +154,24 @@ export class DataScienceErrorHandler implements IDataScienceErrorHandler {
         if (err instanceof JupyterInstallError) {
             await this.dependencyManager.installMissingDependencies(err);
         } else if (err instanceof JupyterSelfCertsError) {
-            // Don't show the message for self cert errors
-            noop();
+            // On a self cert error, warn the user and ask if they want to change the setting
+            const enableOption: string = DataScience.jupyterSelfCertEnable();
```
DonJayamanne, line 158: Moved into error handler from another place (errorHandler is the central location)

src/client/datascience/interactive-common/notebookServerProvider.ts
```diff
@@ -36,38 +33,34 @@ export class NotebookServerProvider implements IJupyterServerProvider {
...
-    ): Promise<INotebookServer | undefined> {
+    private async createServer(options: GetServerOptions): Promise<INotebookServer | undefined> {
         // When we finally try to create a server, update our flag indicating if we're going to allow UI or not. This
         // allows the server to be attempted without a UI, but a future request can come in and use the same startup
-        this.ui.disableUI = options.ui.disableUI;
-        options.ui.onDidChangeDisableUI(() => (this.ui.disableUI = options.ui.disableUI));
+        if (!options.ui.disableUI) {
+            this.ui.disableUI = false;
+        }
+        options.ui.onDidChangeDisableUI(() => (this.ui.disableUI = options.ui.disableUI), this, this.disposables);
```
DonJayamanne, line 59: We keep our own copy of UIDisplayOptions here.
This is because this class is called in another place to pre-warn the jupyter notebook server (non-raw).
Hence its possible we pre-warm first, and later we start a kernel (i.e. this is called from two different palces with two different object instances for UIDisplayOptions).

src/client/datascience/interactive-common/notebookServerProvider.ts
```diff
@@ -90,14 +83,29 @@ export class NotebookServerProvider implements IJupyterServerProvider {
...
                     : this.progressReporter.createProgressIndicator(localize.DataScience.startingJupyter())
                 : undefined;
-
+        if (progressReporter) {
+            disposables.push(progressReporter);
+        } else {
+            this.ui.onDidChangeDisableUI(
+                async () => {
+                    if (!progressReporter && !this.ui.disableUI) {
+                        progressReporter = (await this.jupyterExecution.getServer(serverOptions))
```
DonJayamanne, line 100: This is the magic sauce that re-displays the progress indicator as required.
I.e. if we have auto start, and it takes say 5-30 seconds or longer, and you start running a cell, now you'll immediately get a progress indicator as though you started the kernel from scratch

src/client/datascience/interactive-common/notebookServerProvider.ts
```diff
@@ -130,38 +135,9 @@ export class NotebookServerProvider implements IJupyterServerProvider {
             // the failure there
             await this.jupyterExecution.refreshCommands();
 
-            if (e instanceof JupyterSelfCertsError) {
-                // On a self cert error, warn the user and ask if they want to change the setting
-                const enableOption: string = localize.DataScience.jupyterSelfCertEnable();
```
DonJayamanne, line 135: Moved into central location

src/client/datascience/jupyter/jupyterExecution.ts
```diff
@@ -202,23 +199,7 @@ export class JupyterExecutionBase implements IJupyterExecution {
                         throw err;
                     }
                 }
-            }
-
-            // Note: This is unlikely, so far only 1 telemetry captured for this.
-            // If we're here, then starting jupyter timeout.
-            // Kill any existing connections.
-            connection?.dispose();
```
DonJayamanne, line 210: This has never been hit , we have just 1 telemetry event for this in the past 6 months.
What happens is we just time out and the error is bubbled up.

src/client/datascience/jupyter/jupyterSession.ts
```diff
@@ -239,7 +242,7 @@ export class JupyterSession extends BaseJupyterSession {
     }
 
     private async createSession(options: {
-        token?: CancellationToken;
```
DonJayamanne, line 242: Because token is no longer nullable in `kernel.start` this method & others will never into situations where token === null

src/client/datascience/jupyter/jupyterSession.ts
```diff
@@ -248,13 +251,21 @@ export class JupyterSession extends BaseJupyterSession {
...
+                traceInfoIfCI(`JupyterSession.createSession ${this.kernelConnectionMetadata.id}`);
+                await this.kernelService.ensureKernelIsUsable(
+                    this.resource,
+                    this.kernelConnectionMetadata,
+                    options.ui,
+                    options.token
+                );
+            } catch (ex) {
+                // If we failed to create the kernel, we need to clean up the file.
+                if (this.connInfo && backingFile) {
```
DonJayamanne, line 264: @rchiodo @IanMatthewHuff I think i found one scenario why the temp files are not getting deleted.
If we fail to connect to a kernel, then the files do not get deleted here (as `ensureKernelIsUsable` throws an error & we exit from here)

src/client/datascience/jupyter/liveshare/hostJupyterServer.ts
```diff
@@ -85,31 +91,46 @@ export class HostJupyterServer implements INotebookServer {
...
+                          getDisplayNameOrNameOfKernelConnection(kernelConnection)
+                      )
+                  );
+            cancelToken.onCancellationRequested(() => disposeAllDisposables(disposables), this, disposables);
+            if (progressDisposable) {
+                disposables.push(progressDisposable);
+            } else {
+                ui.onDidChangeDisableUI(
+                    () => {
+                        if (!ui.disableUI && progressDisposable) {
```
rchiodo, line 107: progressDisposable here is not put in the disposable list. 

Might be better to make the dupe code an anonymous function?

src/client/datascience/jupyter/notebookStarter.ts
```diff
@@ -154,8 +160,10 @@ export class NotebookStarter implements Disposable {
             } catch (ex) {
                 traceError(`Parsing failed ${connection.baseUrl}`, ex);
             }
+            disposeAllDisposables(disposables);
             return connection;
         } catch (err) {
+            disposeAllDisposables(disposables);
```
rchiodo, line 166: Why not put this in a finally block?

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
```diff
@@ -127,14 +127,26 @@ export class HostRawNotebookProvider implements IRawNotebookProvider {
             // We need to locate kernelspec and possible interpreter for this launch based on resource and notebook metadata
             const displayName = getDisplayNameOrNameOfKernelConnection(kernelConnection);
 
-            const progressDisposable = !ui.disableUI
+            let progressDisposable = !ui.disableUI
```
rchiodo, line 130: I'm wondering if the 'displayUi' object can just implement the progress indicator? It activates it when UI is true. 

Then you wouldn't need to put it in these places or have the event handler.

rchiodo: Although maybe the event handler is still there, you used it to do something somewhere else besides progress.

rchiodo (changes requested):
What do you think about making the progress indicator in the DisplayOptions class?

src/client/datascience/jupyter/notebookStarter.ts
DonJayamanne: I don't want the catch statements to slow this down, this needs to happen immediately.

src/client/datascience/raw-kernel/liveshare/hostRawNotebookProvider.ts
DonJayamanne: Good idea, let me try that.

DonJayamanne: Nope, can't do that

DonJayamanne: Wait, just thought of a way.

DonJayamanne: @rchiodo I'd like to do this in debt week.
Basically have DisplayOptions have a method such as 
* `showProgress` which will return a disposable. 
    * If `disableUI === true `, then no progress is displayed & vice versa
	* When disposing the return value the progress is hidden (if displayed).
* Other parts of the code can call `showProgress` any time, then all we do is replace the message with the new one. & when that new message is disposed, we swap back to the old message (kinda like a stack)
* & we can have this displayOptions keyed by resource (or optionally add a resource argument to `showProgress`)
	* This way, when you start two notebooks they will have their own progress messages with their own messages
	
Thats the gist of the API, i believe this is pretty much what you too have in mind, however i'll need to refactor some code to make this work well & would prefer to do that in debt week.

rchiodo (approved):

PR #8291, opened 2021-11-17
Don't look for jupyter at startup with raw kernels

DonJayamanne:
Part of #7849

We pre-warm daemons (with jupyter interpreter) and look for Jupyter when we're dealing with raw kernels.
This is totally unnecessary, the only time jupyter interpreter is used is when exporting & the like (its ok for that to be fractionally slower than slowing all of extension & kernel startup).

### DIFF
--- a/src/client/datascience/activation.ts
+++ b/src/client/datascience/activation.ts
@@ -68,6 +68,9 @@ export class Activation implements IExtensionSingleActivationService {
     @debounceAsync(500)
     @swallowExceptions('Failed to pre-warm daemon pool')
     private async PreWarmDaemonPool() {
+        // Note: we're pre-warming the daemon pool for the interpreter we're using to start jupyter.
+        // Thus if we're using raw kernels, then there's no point in pre-warming a daemon that will use
+        // the interpreter for jupyter.
         if (!this.extensionChecker.isPythonExtensionActive || this.rawSupported.isSupported) {
             // Skip prewarm if no python extension
             return;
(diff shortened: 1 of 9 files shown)


src/client/datascience/activation.ts
```diff
@@ -48,27 +48,27 @@ export class Activation implements IExtensionSingleActivationService {
...
+            void this.jupyterInterpreterService.setInitialInterpreter();
+            void this.PreWarmDaemonPool();
         }
     }
 
     @debounceAsync(500)
     @swallowExceptions('Failed to pre-warm daemon pool')
     private async PreWarmDaemonPool() {
-        if (!this.extensionChecker.isPythonExtensionActive) {
+        if (!this.extensionChecker.isPythonExtensionActive || this.rawSupported.isSupported) {
```
rchiodo, line 71: I'm confused by this? Why not prewarm the daemon pool when using raw? We still use daemons for raw.

DonJayamanne: Here we're attempting to pre-warm the daemon pool against the python env that starts jupyter.
When using raw, we pre-warm against the python env that's associated with the kernel.

Its possible that the kernel selected is not the same as the python env used to start jupyter, 
If using non-raw, then it makes sense

rchiodo: I guess the problem then is nobody knows (at that point in the code) why raw isn't allowed. Can you put a comment explaining why?

IanMatthewHuff (approved):

PR #8294, opened 2021-11-17
Add some verbose logging to check perf

DonJayamanne:


src/client/api/pythonApi.ts
```diff
@@ -301,7 +303,7 @@ export class PythonInstaller implements IPythonInstaller {
 @injectable()
 export class EnvironmentActivationService implements IEnvironmentActivationService {
     constructor(@inject(IPythonApiProvider) private readonly apiProvider: IPythonApiProvider) {}
-
+    @traceDecorators.verbose('Getting activated env variables', TraceOptions.BeforeCall)
```
DonJayamanne, line 306: added this to determine the flow of events. E.g. when looking at the logs theres a huge hole & no idea what's going on.
When we log with the decorators we only log after the method call completes (hence if this method takes 30s, then for 30s we don't see anything in the logs, then finally see a log entry for this) - yes we do see that this took 30s, however we don't really know when this started in the log.

IanMatthewHuff: Nice!

IanMatthewHuff (approved):

rchiodo (approved):

Issue #7902, opened 2021-10-12
Better handling of failed ipykernel installations & ipykernel detection


DonJayamanne:
We have plenty of scenarios where users are prompted to isntall ipykernel again & again.
This could happen when:
* pyzmq is not installed or is busted (https://github.com/microsoft/vscode-jupyter/issues/5963)
* ipython_genutils is not installed
* Conda env activation fails
* Or other dependencies are not installed or busted

ipyhton_genutils is a great example of an upstrema bug, (ipykenrle removed that dependency), but was required by other packages in jupyter (which didn't specify that as a requirement).

**Whats going on**
* Jupyter will get if ipykernel is installed (by importing ipykernel)
* Python extension will prompt user to isntall ipykernel
* Installation is a success
* Jupyter will get a message saying it was not installed (as importing ipykernel fails)
* No permissions to install in powershell

**Thus, though ipykernel installatino ran successfully in temrinal, user gets a message that its not installed**

**Suggestion**
* https://github.com/microsoft/vscode-jupyter/issues/8119
	* [x] Add telemetry to see how often this fails for conda users (if failed, then nothing will work)
	* [x] Increase the timeout for conda environments (or retry with a higher timeout?)
	* [x] Possible solutions here https://github.com/microsoft/vscode-jupyter/issues/6694
* [x] Temporarily enable permissions to run powershell scripts in that powershell session 
	* https://github.com/microsoft/vscode-jupyter/issues/421
	* https://github.com/microsoft/vscode-jupyter/issues/9376
* https://github.com/microsoft/vscode-jupyter/issues/1257
	* Fixes https://github.com/microsoft/vscode-jupyter/issues/5963
	* https://github.com/microsoft/vscode-jupyter/issues/1128
* https://github.com/microsoft/vscode-jupyter/issues/1253
* https://github.com/microsoft/vscode-jupyter/issues/421
* [x] Add telemetry in python extension to see if installation is successful and ipykernel cannot be imported (this is an indication of something wrong with the package)
* If this happens we can notify the user and ask them to do somethign (kernel doctor?)
	* [x] Display message with possible solutions & links to wiki https://github.com/microsoft/vscode-jupyter/issues/8354
* If jupyter extension keeps asking users to install ipykernel > n times without running a cell succesfully, 
	* Add telemetry to detect this scenario
	* then display another button in the prompt withy more informaiton to install it them selves
[...]

PR #8304, opened 2021-11-18
Display Kernel startup Errors in cell output for more visibility

DonJayamanne:
For #7902

image

### DIFF
--- a/src/client/datascience/errors/errorHandler.ts
+++ b/src/client/datascience/errors/errorHandler.ts
@@ -18,7 +18,7 @@ import {
     IKernelDependencyService,
     KernelInterpreterDependencyResponse
 } from '../types';
-import { CancellationError as VscCancellationError, CancellationTokenSource, ConfigurationTarget } from 'vscode';
+import { CancellationError as VscCancellationError, CancellationTokenSource, ConfigurationTarget, NotebookCell, NotebookCellOutput, NotebookCellOutputItem } from 'vscode';
 import { CancellationError } from '../../common/cancellation';
 import { KernelConnectionTimeoutError } from './kernelConnectionTimeoutError';
 import { KernelDiedError } from './kernelDiedError';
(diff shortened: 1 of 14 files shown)


src/client/datascience/errors/errorRendererComms.ts
```diff
@@ -0,0 +1,85 @@
...
+export class ErrorRendererCommunicationHandler implements IExtensionSyncActivationService {
+    constructor(
+        @inject(IDisposableRegistry) private readonly disposables: IDisposableRegistry,
+        @inject(IDocumentManager) private readonly documentManager: IDocumentManager,
+        @inject(ICommandManager) private readonly commandManager: ICommandManager,
+        @inject(IFileSystem) private readonly fs: IFileSystem,
+        @inject(IApplicationShell) private readonly applicationShell: IApplicationShell
+    ) {}
+
+    activate(): void {
```
DonJayamanne, line 23: Moved out of interactive window class, this is not specific to just interactive window anymore.

src/client/datascience/errors/errorHandler.ts
```diff
@@ -256,16 +289,62 @@ export class DataScienceErrorHandler implements IDataScienceErrorHandler {
...
+            return;
+        }
+        const regex = /\[(?<name>.*)\]\((?<command>command:\S*)\)/gm;
+        let matches: RegExpExecArray | undefined | null;
+        while ((matches = regex.exec(errorMessage)) !== null) {
+            if (matches.length === 3) {
+                errorMessage = errorMessage.replace(matches[0], `<a href='${matches[2]}'>${matches[1]}</a>`);
+            }
+        }
+        const execution = controller.controller.createNotebookCellExecution(cellToDisplayErrors);
```
IanMatthewHuff, line 335: I know this is how the API works, just feels odd that we have to do all this just to change the output. Though I guess it does ensure that everything runs through the execution control path.

IanMatthewHuff (approved):

src/client/datascience/errors/errorHandler.ts
```diff
@@ -256,16 +289,62 @@ export class DataScienceErrorHandler implements IDataScienceErrorHandler {
...
+            .get<IKernelProvider>(IKernelProvider)
+            .get(cellToDisplayErrors.notebook);
+        if (!associatedkernel) {
+            return;
+        }
+        // Sometimes the cells are still running, wait for 1s for cells to finish & get cleared,
+        // Then display the error in the cell.
+        const stopWatch = new StopWatch();
+        while (stopWatch.elapsedTime <= 1_000 && associatedkernel.hasPendingCells) {
+            await sleep(100);
```
rchiodo, line 317: Would it make sense to interrupt the current execution instead?

DonJayamanne: No, it will stop. The existing code will still the execution.
The problem is this code gets called just when kernel dies, hence the cells are stored after that, within a few Ms.
Hence the max delay of 1s.

I wish the api allowed us to update cell output without all of this api. But these are a few exceptions

Issue #8295, opened 2021-11-17
Unable to start Jupyter on Mac (Silicon) Monterey


DonJayamanne:
OS: Mac Silicon (M1) Monterey

**Problem**
* Unable to run a cell as Jupyter fails to start

**Logs**
```
File "/opt/homebrew/lib/python3.9/site-packages/traitlets/config/loader.py", line 933, in __call__
namespace._flags.append(self.flag)
AttributeError: 'Namespace' object has no attribute '_flags'
```

We believe this could resolved by updating `traitlets` to the latest version as follows `python -m pip install 'traitlets>=5.1.1`. 

More information on this can be found here https://discourse.jupyter.org/t/attributeerror-namespace-object-has-no-attribute-flags-when-running-jupyter-kernelspec-list-json/11690

DonJayamanne:
**Solution:**
* Update `traitlets` to the latest version as follows `python -m pip install 'traitlets>=5.1.1`
* Please ensure you install `traitlets` into the Python enviornment used to start Jupyter
* To determine the Python environment used to start Jupyter, please do the following 
	* Select the command 'Jupyter: Select Interpreter to start Jupyter Server`
	* In the quickpick the currently selected interrpeter will be displayed in the intput box as a placeholder.
	* Please install `traitlets` into this environment

DonJayamanne:
Please test using this test plan https://github.com/microsoft/vscode-jupyter/issues/8357

PR #8307, opened 2021-11-18
Recommend updating traitlets to 5.1.1 if jupyter fails to start

DonJayamanne:
For #8295

image


src/client/common/utils/localize.ts
```diff
@@ -379,6 +379,18 @@ export namespace DataScience {
...
+        'DataScience.failedToStartJupyter',
+        "Failed to start Jupyter in the environment '{0}'. \nView Jupyter [log](command:jupyter.viewOutput) for further details."
+    );
+    export const failedToStartJupyterWithErrorInfo = localize(
+        'DataScience.failedToStartJupyterWithErrorInfo',
+        "Failed to start Jupyter in the environment '{0}'. \n{1} \nView Jupyter [log](command:jupyter.viewOutput) for further details."
+    );
+    export const failedToStartJupyterDueToOutdatedTraitlets = localize(
+        'DataScience.failedToStartJupyterDueToOutdatedTraitlets',
+        "Failed to start Jupyter in the environment '{0}' possibly due to an outdated version of 'traitlets'. \n{1} \nConsider updating the 'traitlets' module to '5.1.1' or later. \nView Jupyter [log](command:jupyter.viewOutput) for further details."
```
DonJayamanne, line 392: 3 variations of errors we display.
If we have an error message from python (e.g. we have a python stack trace & have been able to capture the python error, then we display that, else we display a generic error message).
If we can find a specific error thats generally associated with the traitlets being old, then we have another for that.

src/datascience-ui/error-renderer/index.ts
```diff
@@ -96,6 +97,11 @@ export const activate: ActivationFunction = (_context) => {
                         line = line.replace(matches[0], `<a href='command:${matches[1]}'>${matches[2]}</a>`);
                     }
                 }
+                while ((matches = akaMsLinks.exec(line)) !== null) {
```
DonJayamanne, line 100: Modified to support https links to aka.ms links
This way the `More info` is also displayed in the error output

src/client/datascience/errors/errorHandler.ts
```diff
@@ -219,15 +227,78 @@ export class DataScienceErrorHandler implements IDataScienceErrorHandler {
...
+                  this.workspace.workspaceFolders || []
+              )})`
+            : '';
+        if (
+            jupyterInterpreter &&
+            failureInfoFromMessage?.reason === KernelFailureReason.jupyterStartFailureOutdatedTraitlets
+        ) {
+            void this.showMessageWithMoreInfo(
+                DataScience.failedToStartJupyterDueToOutdatedTraitlets().format(envDisplayName, pythonError || ''),
+                'https://aka.ms/kernelFailuresJupyterTrailtletsOutdated',
```
IanMatthewHuff, line 292: Looks like a misspell here, or is the aka link also misspelled?

IanMatthewHuff (approved):
LGTM with possible spelling fix

src/client/datascience/errors/errorHandler.ts
DonJayamanne: Yuck, i'll leave that spelling, as the link works and from what I know we can't delete those links.
Also no harm in that.

rchiodo (approved):

PR #8317, opened 2021-11-20
Few Perf Improvements  for kernel startup & more logging

DonJayamanne:
Part of #7849

### DIFF
--- a/src/client/common/installer/productInstaller.ts
+++ b/src/client/common/installer/productInstaller.ts
@@ -6,11 +6,12 @@ import { IPythonInstaller } from '../../api/types';
 import '../../common/extensions';
 import { InterpreterPackages } from '../../datascience/telemetry/interpreterPackages';
 import { IServiceContainer } from '../../ioc/types';
+import { logValue } from '../../logging/trace';
 import { PythonEnvironment } from '../../pythonEnvironments/info';
 import { getInterpreterHash } from '../../pythonEnvironments/info/interpreter';
 import { IApplicationShell } from '../application/types';
 import { STANDARD_OUTPUT_CHANNEL } from '../constants';
-import { traceError, traceInfo } from '../logger';
+import { traceDecorators, traceError, traceInfo } from '../logger';
 import { IProcessServiceFactory, IPythonExecutionFactory } from '../process/types';
 import {
     IConfigurationService,
@@ -94,8 +95,11 @@ export abstract class BaseInstaller {
             .get<IPythonInstaller>(IPythonInstaller)
             .install(product, resource, cancel, reInstallAndUpdate, installPipIfRequired);
     }
-
-    public async isInstalled(product: Product, resource?: InterpreterUri): Promise<boolean | undefined> {
+    @traceDecorators.verbose('Checking if product is installed')
+    public async isInstalled(
+        product: Product,
+        @logValue('path') resource?: InterpreterUri
+    ): Promise<boolean | undefined> {
         // User may have customized the module name or provided the fully qualified path.
         const interpreter = isResource(resource) ? undefined : resource;
         const uri = isResource(resource) ? resource : undefined;
--- a/src/client/common/utils/misc.ts
+++ b/src/client/common/utils/misc.ts
@@ -70,14 +70,14 @@ export type TraceInfo =
 export function tracing<T>(log: (t: TraceInfo) => void, run: () => T, logBeforeCall?: boolean): T {
     const timer = new StopWatch();
     try {
+        if (logBeforeCall) {
+            log(undefined);
+        }
         // eslint-disable-next-line no-invalid-this, @typescript-eslint/no-use-before-define,
         const result = run();
 
         // If method being wrapped returns a promise then wait for it.
         if (isPromise(result)) {
-            if (logBeforeCall) {
-                log(undefined);
-            }
             // eslint-disable-next-line
             (result as Promise<void>)
                 .then((data) => {
--- a/src/client/datascience/jupyter/kernels/helpers.ts
+++ b/src/client/datascience/jupyter/kernels/helpers.ts
@@ -439,6 +439,21 @@ export function isLocalLaunch(configuration: IConfigurationService) {
     return false;
 }
 
+export function getInterpreterHashInMetdata(
+    notebookMetadata: nbformat.INotebookMetadata | undefined
+): string | undefined {
+    // If the user has kernelspec in metadata & the interpreter hash is stored in metadata, then its a perfect match.
+    // This is the preferred approach https://github.com/microsoft/vscode-jupyter/issues/5612
+    if (
+        typeof notebookMetadata === 'object' &&
+        'interpreter' in notebookMetadata &&
+        (notebookMetadata as any).interpreter &&
+        typeof (notebookMetadata as any).interpreter === 'object' &&
+        'hash' in (notebookMetadata as any).interpreter
+    ) {
+        return (notebookMetadata as any).interpreter.hash;
+    }
+}
 export function findPreferredKernel(
     kernels: KernelConnectionMetadata[],
     resource: Resource,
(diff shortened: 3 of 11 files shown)


src/client/common/installer/productInstaller.ts
```diff
@@ -94,8 +95,11 @@ export abstract class BaseInstaller {
             .get<IPythonInstaller>(IPythonInstaller)
             .install(product, resource, cancel, reInstallAndUpdate, installPipIfRequired);
     }
-
-    public async isInstalled(product: Product, resource?: InterpreterUri): Promise<boolean | undefined> {
+    @traceDecorators.verbose('Checking if product is installed')
```
DonJayamanne, line 98: Added some logging, these were useful

src/client/common/utils/misc.ts
```diff
@@ -70,14 +70,14 @@ export type TraceInfo =
 export function tracing<T>(log: (t: TraceInfo) => void, run: () => T, logBeforeCall?: boolean): T {
     const timer = new StopWatch();
     try {
+        if (logBeforeCall) {
+            log(undefined);
+        }
```
DonJayamanne, line 75: Bug fix, we were logging just after the method was called (which was incorrect)

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -439,6 +439,21 @@ export function isLocalLaunch(configuration: IConfigurationService) {
...
+    if (
+        typeof notebookMetadata === 'object' &&
+        'interpreter' in notebookMetadata &&
+        (notebookMetadata as any).interpreter &&
+        typeof (notebookMetadata as any).interpreter === 'object' &&
+        'hash' in (notebookMetadata as any).interpreter
+    ) {
+        return (notebookMetadata as any).interpreter.hash;
+    }
+}
```
DonJayamanne, line 456: An unnecessary change, but will be required later.
Had this as part of some other perf fixes that I'm not submitting just yet, basically moved the extraction of interpreterHash from notebook metadata into a function.

src/client/datascience/kernel-launcher/jupyterPaths.ts
```diff
@@ -15,29 +15,57 @@ const winJupyterPath = path.join('AppData', 'Roaming', 'jupyter', 'kernels');
...
     constructor(
         @inject(IPlatformService) private platformService: IPlatformService,
         @inject(IPathUtils) private readonly pathUtils: IPathUtils,
-        @inject(IEnvironmentVariablesProvider) private readonly envVarsProvider: IEnvironmentVariablesProvider
-    ) {}
+        @inject(IEnvironmentVariablesProvider) private readonly envVarsProvider: IEnvironmentVariablesProvider,
+        @inject(IDisposableRegistry) disposables: IDisposableRegistry,
+        @inject(IMemento) @named(GLOBAL_MEMENTO) private readonly globalState: Memento
+    ) {
+        this.envVarsProvider.onDidEnvironmentVariablesChange(
```
DonJayamanne, line 32: Clear the cache if user updates their env variables

src/client/datascience/kernel-launcher/jupyterPaths.ts
```diff
@@ -15,29 +15,57 @@ const winJupyterPath = path.join('AppData', 'Roaming', 'jupyter', 'kernels');
...
+            disposables
+        );
+    }
 
     /**
      * This should return a WRITABLE place that jupyter will look for a kernel as documented
      * here: https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs
      */
+    @traceDecorators.verbose('Getting Jupyter KernelSpec Root Path')
     public async getKernelSpecRootPath(): Promise<string | undefined> {
```
DonJayamanne, line 46: THis logging was very useful, 
On windows this is very slow as `tryGetRealPath` is slow in windows.
Caching takes only 15-20ms and without that it takes around 400ms on windows, every time. this gets called at least 3 times, thus we shaved well over 1s with this simple change.

src/client/datascience/kernel-launcher/jupyterPaths.ts
```diff
@@ -76,27 +104,41 @@ export class JupyterPaths {
      * We need to look at the 'kernels' sub-directory and these paths are supposed to come first in the searching
      * https://jupyter.readthedocs.io/en/latest/projects/jupyter-directories.html#envvar-JUPYTER_PATH
      */
+    @traceDecorators.verbose('Get Jupyter Paths')
     private async getJupyterPathPaths(cancelToken?: CancellationToken): Promise<string[]> {
```
DonJayamanne, line 108: The logging and caching was very useful, see comments earlier.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -391,7 +391,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
 
         // Prep so that we can track the selected controller for this document
         traceInfoIfCI(`Clear controller mapping for ${getDisplayPath(document.uri)}`);
-        const loadControllersPromise = this.loadNotebookControllers();
+        void this.loadNotebookControllers();
```
DonJayamanne, line 394: Its ok if the controllers haven't loaded, 
We can create the controllers as required.
After all, that's what we do, we get controller information and create the controller.
If we call that again its not re-created.

This way, we don't have to wait for all 10-20 or 50 controllers to get loaded. 
I.e. we can create controllers one by one based on the information available.

Result, we can now run button will be available even earlier.

Got more changes that will imporve this scenario & make this change all the more useful (separate PR).

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -470,11 +467,15 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
                 this.preferredControllers.set(document, targetController);
             } else {
-                traceInfoIfCI(
-                    `TargetController not found ID: ${preferredConnection?.id} for document ${getDisplayPath(
-                        document.uri
-                    )}`
-                );
+                // Possible the kernel discovery hasn't completed yet.
+                if (preferredConnection) {
+                    this.createNotebookControllers([preferredConnection]);
```
DonJayamanne, line 472: If not available, then just create it.

src/client/datascience/kernel-launcher/kernelLauncherDaemon.ts
```diff
@@ -34,7 +33,7 @@ export class PythonKernelLauncherDaemon implements IDisposable {
         @inject(KernelEnvironmentVariablesService)
         private readonly kernelEnvVarsService: KernelEnvironmentVariablesService
     ) {}
-    @traceDecorators.verbose('Launching kernel daemon', TraceOptions.BeforeCall)
+    @traceDecorators.verbose('Launching kernel daemon')
```
DonJayamanne, line 36: `TraceOptions.BeforeCall` is the default, hence removed this in a few places

src/client/datascience/kernel-launcher/jupyterPaths.ts
rchiodo: Didn't you have some decorator way to cache results from a function?

src/client/datascience/kernel-launcher/jupyterPaths.ts
rchiodo: Same comment on this one. I swear you had something that would cache results for a function call into a memento.

rchiodo (approved):

src/client/datascience/kernel-launcher/jupyterPaths.ts
DonJayamanne: Oh yes, yo'ure right, let me check

DonJayamanne: The existing decorator will not , because that uses a memory based cache store, i.e. when user reloads VS Code the cache is empty.

I'd need to create a decorator that uses the global memento? Would definitely make the code cleaner & improve readability.
Will do that next week (I.e. will not merge this PR).

src/client/datascience/kernel-launcher/jupyterPaths.ts
DonJayamanne: Yes now I remember, we had code to cache in memento, I think its in the Python extension, & it may have been removed from this repo as it wasn't used.

PR #8318, opened 2021-11-20
Add 'jupyter.runStartupCommands' to top of the pylance concat file

### LINKED ISSUE #7993: Add ability to provide code completions for modules imported in startup commands
At my workplace, we have a specific code file written to provide useful helper functions whenever we choose to run IPython in a terminal:

```
ipython -i -m server.scripts.interactive
```

I want to have an option to run that file by default every time I create a new Jupyter Interactive Window. I've tried running it manually using Jupyter's `Run File in Interactive Window` feature, but it results in the working directory being shifted to the directory containing the file that is run, which is wrong. The working directory needs to be the workspace root or else the script will fail along with everything else I try to do. Plus, it would simply be nice to not have to navigate to that file just to run it every time I want to use an Interactive Window.

rchiodo:
Fixes #7993 

Import the new concat doc I just fixed in https://github.com/microsoft/vscode-jupyter-lsp-middleware/pull/26. This imports the 'runstartupcommands' onto the top of the concat document sent to pylance.

### DIFF
--- a/news/2 Fixes/7993.md
+++ b/news/2 Fixes/7993.md
@@ -1,1 +1,3 @@
-Let code in 'jupyter.runStartupCommands' work with intellisense.
+Let code in 'jupyter.runStartupCommands' work with intellisense. 
+
+For example, if your startup commands were 'import pandas as pd', auto completion would now work automatically for 'pd.' without having to type the import first.
--- a/news/2 Fixes/7993.md
+++ b/news/2 Fixes/7993.md
@@ -1,0 +1,1 @@
+Let code in 'jupyter.runStartupCommands' work with intellisense.
--- a/src/client/datascience/notebook/intellisense/intellisenseProvider.ts
+++ b/src/client/datascience/notebook/intellisense/intellisenseProvider.ts
@@ -170,6 +170,23 @@ export class IntellisenseProvider implements INotebookLanguageClientProvider, IE
         return interpreterId == notebookId;
     }
 
+    private getNotebookHeader(uri: Uri) {
+        const settings = this.configService.getSettings(uri);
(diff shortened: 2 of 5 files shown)


src/client/datascience/notebook/intellisense/intellisenseProvider.ts
```diff
@@ -170,6 +170,23 @@ export class IntellisenseProvider implements INotebookLanguageClientProvider, IE
...
+        let setting = settings.runStartupCommands || settings.runMagicCommands;
+
+        // Convert to string in case we get an array of startup commands.
+        if (Array.isArray(setting)) {
+            setting = setting.join(`\n`);
+        }
+
+        if (setting) {
+            // Cleanup the line feeds. User may have typed them into the settings UI so they will have an extra \\ on the front.
+            return setting.replace(/\\n/g, '\n');
```
DonJayamanne, line 185: Should we look for `CRLF` as well?

news/2 Fixes/7993.md
```diff
@@ -0,0 +1 @@
+Let code in 'jupyter.runStartupCommands' work with intellisense.
```
DonJayamanne, line 1: A big win for our power users.

DonJayamanne: I'd put a few more comments here about what this means, so that its more obvious in the read me.
E.g. 
`import pandas` in startup command will import pandas & when you go into a cell and type `pandas.` then you immediately get code completions & the pandas is imported.

Would be a great way to re-advertise this feature for users.
I.e. now it works in runtime as well as intellisense.

DonJayamanne (approved):

PR #8324, opened 2021-11-22
Add API to consume new Python API (but no used)

DonJayamanne:
Part of #7849

### DIFF
--- a/src/client/api/types.ts
+++ b/src/client/api/types.ts
@@ -7,6 +7,8 @@ import { InterpreterUri } from '../common/installer/types';
 import { InstallerResponse, Product, Resource } from '../common/types';
 import { IInterpreterQuickPickItem } from '../interpreter/configuration/types';
 import { PythonEnvironment } from '../pythonEnvironments/info';
+import type { SemVer } from 'semver';
+
 export type ILanguageServerConnection = Pick<
     lsp.ProtocolConnection,
     'sendRequest' | 'sendNotification' | 'onProgress' | 'sendProgress' | 'onNotification' | 'onRequest'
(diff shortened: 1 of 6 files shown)


src/client/api/types.ts
```diff
@@ -111,6 +113,15 @@ export type PythonApi = {
      * Registers a visibility filter for the interpreter status bar.
      */
     registerInterpreterStatusFilter(filter: IInterpreterStatusbarVisibilityFilter): void;
+    getCondaVersion?(): Promise<SemVer | undefined>;
+    /**
+     * Returns the conda executable.
+     */
+    getCondaFile?(): Promise<string | undefined>;
+    getEnvironmentActivationShellCommands?(
```
DonJayamanne, line 121: New API exposed by python extension. Possible users don't have the latest python extenison hence we need to take that into account.

src/client/common/process/condaService.ts
```diff
@@ -0,0 +1,89 @@
...
+import { Memento } from 'vscode';
+import { IPythonApiProvider } from '../../api/types';
+import { IFileSystem } from '../platform/types';
+import { GLOBAL_MEMENTO, IMemento } from '../types';
+import { createDeferredFromPromise } from '../utils/async';
+
+const CACHEKEY_FOR_CONDA_INFO = 'CONDA_INFORMATION_CACHE';
+
+@injectable()
+export class CondaService {
```
DonJayamanne, line 15: Exposes the CondaVersion, CondaFile behind a class with caching & wraps the Python API call (possible its not available, etc)

src/client/common/process/condaService.ts
```diff
@@ -0,0 +1,89 @@
...
+            file: this._file,
+            fileHash
+        });
+    }
+    /**
+     * If the last modified date of the conda file is the same as when we last checked,
+     * then we can assume the version is the same.
+     * Even if not, we'll update this with the latest information.
+     */
+    private async getCachedInformation(): Promise<{ version: SemVer; file: string } | undefined> {
```
DonJayamanne, line 73: Comments added how & why we cache

PR #8337, opened 2021-11-23
Change arg to use Interpreter instead of path:string

DonJayamanne:
Required for perf improvements.
We're using `python.path` in a number of places when we have the full interpreter information.
I.e. there are no instances when we'll not have interpreter information, hence use that instead of using the executable path.
Else in the lower decks we end up trying to resolve the python env information from the path.


src/client/api/pythonApi.ts
```diff
@@ -208,15 +207,6 @@ export class LanguageServerProvider implements ILanguageServerProvider {
     }
 }
 
-@injectable()
-export class WindowsStoreInterpreter implements IWindowsStoreInterpreter {
-    constructor(@inject(IPythonApiProvider) private readonly apiProvider: IPythonApiProvider) {}
-
-    public isWindowsStoreInterpreter(pythonPath: string): Promise<boolean> {
-        return this.apiProvider.getApi().then((api) => api.isWindowsStoreInterpreter(pythonPath));
```
DonJayamanne, line 216: Since we're no longer passing just `path:string`, we have the full env information, thus we don't need this.

src/client/common/installer/productInstaller.ts
```diff
@@ -102,21 +102,24 @@ export abstract class BaseInstaller {
     @traceDecorators.verbose('Checking if product is installed')
     public async isInstalled(
         product: Product,
-        @logValue('path') resource?: InterpreterUri
+        @logValue('path') interpreter: PythonEnvironment
```
DonJayamanne, line 105: This is mandatory

DonJayamanne: & it must be an interpreter

src/client/common/process/baseDaemon.ts
```diff
@@ -52,7 +53,7 @@ export abstract class BasePythonDaemon {
     constructor(
         protected readonly pythonExecutionService: IPythonExecutionService,
         protected readonly platformService: IPlatformService,
-        protected readonly pythonPath: string,
+        protected readonly interpreter: PythonEnvironment,
```
DonJayamanne, line 56: Ensure we pass env information around not path

src/client/common/installer/productInstaller.ts
```diff
@@ -277,8 +280,8 @@ export class ProductInstaller implements IInstaller {
     ): Promise<InstallerResponse> {
         return this.createInstaller().install(product, resource, cancel, reInstallAndUpdate, installPipIfRequired);
     }
-    public async isInstalled(product: Product, resource?: InterpreterUri): Promise<boolean | undefined> {
-        return this.createInstaller().isInstalled(product, resource);
+    public async isInstalled(product: Product, interpreter: PythonEnvironment): Promise<boolean | undefined> {
```
DonJayamanne, line 283: This is now cleaner, we were always passing `PyhtonEnvironment`, but was left as `InterpreterUri` as this was a carry over from from Python extension

src/client/common/process/pythonExecutionFactory.ts
```diff
@@ -46,9 +46,7 @@ export class PythonExecutionFactory implements IPythonExecutionFactory {
         @inject(IEnvironmentActivationService) private readonly activationHelper: IEnvironmentActivationService,
         @inject(IProcessServiceFactory) private readonly processServiceFactory: IProcessServiceFactory,
         @inject(IBufferDecoder) private readonly decoder: IBufferDecoder,
-        @inject(IWindowsStoreInterpreter) private readonly windowsStoreInterpreter: IWindowsStoreInterpreter,
         @inject(IPlatformService) private readonly platformService: IPlatformService,
-        @inject(IInterpreterService) private readonly interpreterService: IInterpreterService,
```
DonJayamanne, line 51: Fewer dependencies, fewer calls, should be slightly faster

src/client/datascience/activation.ts
```diff
@@ -82,7 +82,7 @@ export class Activation implements IExtensionSingleActivationService {
         // Warm the daemon pool just for the interpreter used to start Jupyter.
         await this.factory.createDaemon<IPythonDaemonExecutionService>({
             daemonModule: JupyterDaemonModule,
-            pythonPath: interpreter.path
+            interpreter: interpreter
```
DonJayamanne, line 85: We always had interpreter information,

PR #8342, opened 2021-11-23
Faster activation of Python environments such as Conda

DonJayamanne:
For #7849
* Get activated environment variable activation ourselves (& not using Python extension)
* Note: We still rely on python extension to generate the commands used to activate an environment

Here are some stats:
* Non-Conda - Our code takes 300-400ms and Python extension takes 1300-1500ms
* Conda - Our code takes 5s with conda and Python extension takes 10s 

So we're definitely saving a few seconds

### DIFF
--- a/src/client/api/pythonApi.ts
+++ b/src/client/api/pythonApi.ts
@@ -18,7 +18,7 @@ import { isCI } from '../common/constants';
 import { trackPackageInstalledIntoInterpreter } from '../common/installer/productInstaller';
 import { ProductNames } from '../common/installer/productNames';
 import { InterpreterUri } from '../common/installer/types';
-import { traceInfo, traceInfoIfCI } from '../common/logger';
+import { traceDecorators, traceInfo, traceInfoIfCI } from '../common/logger';
 import { getDisplayPath } from '../common/platform/fs-paths';
 import {
     GLOBAL_MEMENTO,
@@ -36,6 +36,7 @@ import { PythonExtension, Telemetry } from '../datascience/constants';
 import { InterpreterPackages } from '../datascience/telemetry/interpreterPackages';
 import { IInterpreterQuickPickItem, IInterpreterSelector } from '../interpreter/configuration/types';
 import { IInterpreterService } from '../interpreter/contracts';
+import { TraceOptions } from '../logging/trace';
 import { PythonEnvironment } from '../pythonEnvironments/info';
 import { areInterpreterPathsSame } from '../pythonEnvironments/info/interpreter';
 import { captureTelemetry, sendTelemetryEvent } from '../telemetry';
@@ -235,6 +236,7 @@ export class PythonInstaller implements IPythonInstaller {
         @inject(IMemento) @named(GLOBAL_MEMENTO) private readonly memento: Memento
     ) {}
 
+    @traceDecorators.verbose('Installing Product', TraceOptions.Arguments | TraceOptions.BeforeCall)
     public async install(
         product: Product,
         resource?: InterpreterUri,
@@ -333,6 +335,7 @@ export class InterpreterService implements IInterpreterService {
     }
 
     @captureTelemetry(Telemetry.InterpreterListingPerf)
+    @traceDecorators.verbose('Get Interpreters', TraceOptions.Arguments | TraceOptions.BeforeCall)
     public getInterpreters(resource?: Uri): Promise<PythonEnvironment[]> {
         this.hookupOnDidChangeInterpreterEvent();
         // Cache result as it only changes when the interpreter list changes or we add more workspace folders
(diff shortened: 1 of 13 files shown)


src/client/interpreter/activation/types.ts
```diff
@@ -10,7 +10,7 @@ export const IEnvironmentActivationService = Symbol('IEnvironmentActivationServi
 export interface IEnvironmentActivationService {
     getActivatedEnvironmentVariables(
         resource: Resource,
-        interpreter?: PythonEnvironment,
+        interpreter: PythonEnvironment,
```
DonJayamanne, line 13: Its never undefined, hence (all of these optionals make the code lot more messy).
Fortunately most of them aren't optional anymore (simply because we were always passing the values).

src/client/api/pythonApi.ts
```diff
@@ -239,6 +236,7 @@ export class PythonInstaller implements IPythonInstaller {
         @inject(IMemento) @named(GLOBAL_MEMENTO) private readonly memento: Memento
     ) {}
 
+    @traceDecorators.verbose('Installing Product', TraceOptions.Arguments | TraceOptions.BeforeCall)
```
DonJayamanne, line 239: More logging

src/client/api/types.ts
```diff
@@ -78,7 +78,7 @@ export type PythonApi = {
      */
     getActivatedEnvironmentVariables(
         resource: Resource,
-        interpreter?: PythonEnvironment,
+        interpreter: PythonEnvironment,
```
DonJayamanne, line 81: This is never undefined, we always pass this around, hence changed this to not be nullable.
Makes the code cleaner

src/client/common/installer/productInstaller.ts
```diff
@@ -158,20 +158,13 @@ export class DataScienceInstaller extends BaseInstaller {
...
     public async install(
         product: Product,
-        interpreterUri?: InterpreterUri,
+        interpreter: PythonEnvironment,
         cancel?: CancellationToken,
         reInstallAndUpdate?: boolean,
         installPipIfRequired?: boolean
     ): Promise<InstallerResponse> {
-        // Precondition
-        if (isResource(interpreterUri)) {
```
DonJayamanne, line 167: This would never happen with our code path, hence removed this, typings ensure this never happens.
I think this was a carry on from when we moved code from Python repo

src/client/common/process/interpreterActivation.ts
```diff
@@ -0,0 +1,379 @@
...
+                    sendTelemetryEvent(Telemetry.GetActivatedEnvironmentVariables, stopWatch.elapsedTime, {
+                        envType,
+                        pythonEnvType: envType,
+                        source: 'jupyter',
+                        failed: true,
+                        reason: 'noActivationCommands'
+                    });
+                    return;
+                }
+                traceVerbose(`Activation Commands received ${activationCommands} for shell ${shellInfo.shell}`);
```
DonJayamanne, line 212: Borrowed this logic from Python extension

src/client/common/process/interpreterActivation.ts
```diff
@@ -0,0 +1,379 @@
...
+                    this.getActivationCommands(resource, interpreter),
+                    this.processServiceFactory.create(resource)
+                ]);
+                if (!activationCommands || activationCommands.length === 0) {
+                    sendTelemetryEvent(Telemetry.GetActivatedEnvironmentVariables, stopWatch.elapsedTime, {
+                        envType,
+                        pythonEnvType: envType,
+                        source: 'jupyter',
+                        failed: true,
+                        reason: 'noActivationCommands'
```
DonJayamanne, line 208: Better telemetry, this way when things fall over we know what could be causing issues with kernel startups.
E.g. if we cannot activate a conda enviornment, then we know for certain that this is the cause for kernel startup to fail.

src/client/common/process/interpreterActivation.ts
```diff
@@ -0,0 +1,379 @@
...
+ * Solution:
+ * 1. Get the commands from Python extensoin to activate a Python environment.
+ * 2. Activate & generate the env variables ourselves.
+ * 3. In parallel get the activated env variables from Python extension.
+ * 4. Return the results from which ever completes first.
+ *
+ * We've found that doing this in jupyter yields much better results.// Copyright (c) Microsoft Corporation. All rights reserved.
+ * Stats: In Jupyter activation takes 800ms & the same in Python would take 2.6s, or with a complex Conda (5s vs 9s).
+ * Note: We cache the activate commands, as this is not something that changes day to day. Its almost a constant.
+ * Either way, we always fetch the latest from Python extension & update the cache.
```
DonJayamanne, line 72: Comments added,

PR #8343, opened 2021-11-23
Pre-warm env vars and faster env file parsing

DonJayamanne:
For #7849

### DIFF
--- a/src/client/common/variables/environment.ts
+++ b/src/client/common/variables/environment.ts
@@ -5,6 +5,8 @@ import { inject, injectable } from 'inversify';
 import * as path from 'path';
 import { sendTelemetryEvent } from '../../telemetry';
 import { EventName } from '../../telemetry/constants';
+import { traceError } from '../logger';
+import { isFileNotFoundError } from '../platform/errors';
 import { IFileSystem } from '../platform/types';
 import { EnvironmentVariables, IEnvironmentVariablesService } from './types';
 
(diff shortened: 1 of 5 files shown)


src/client/common/variables/environment.ts
```diff
@@ -16,10 +18,16 @@ export class EnvironmentVariablesService implements IEnvironmentVariablesService
         filePath?: string,
         baseVars?: EnvironmentVariables
     ): Promise<EnvironmentVariables | undefined> {
-        if (!filePath || !(await this.fs.localFileExists(filePath))) {
```
DonJayamanne, line 19: Remove just this line of code saved almost 500-700ms.
We have two IO operations unnecessarily.

src/client/common/variables/environmentVariablesProvider.ts
```diff
@@ -58,12 +58,15 @@ export class EnvironmentVariablesProvider implements IEnvironmentVariablesProvid
...
         const workspaceFolderUri = this.getWorkspaceFolderUri(resource);
+        if (!workspaceFolderUri) {
+            return;
+        }
         this.trackedWorkspaceFolders.add(workspaceFolderUri ? workspaceFolderUri.fsPath : '');
 
         // eslint-disable-next-line
         // TODO: This should be added to the python API (or this entire service should move there)
         // https://github.com/microsoft/vscode-jupyter/issues/51
-        const envFile = workspaceFolderUri?.fsPath ? path.join(workspaceFolderUri.fsPath, '.env') : '.env';
```
DonJayamanne, line 66: This was bogus, we cannot have a file named `.env` that is not a valid file (unless it exists in the users root directory, doesn't work on windows thought).
I.e its an invalid case, here we're expecting fully qualified file paths.

PR #8344, opened 2021-11-24
Send proper times when executing a cell

DonJayamanne:


src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -441,9 +442,9 @@ export class CellExecution implements IDisposable {
         const props = { notebook: true };
         if (!CellExecution.sentExecuteCellTelemetry) {
             CellExecution.sentExecuteCellTelemetry = true;
-            sendTelemetryEvent(Telemetry.ExecuteCellPerceivedCold, this.stopWatch.elapsedTime, props);
```
DonJayamanne, line 444: Using the `stopWatch` property is wrong.
This value gets reset just when the cell is sent to Jupyter. Thus this isn't the perceived time
E.g. when usre opens a notebook an hits run, and kernel takes 10s to start, afer it starts we send this cell to Jupyter and re-set the stopwatch, but now we're sending <10 for execution times.

Hence using a new property to capture the perceived times.

vibhavagarwal5:
Is this change published, I see that the cell still starts from 40s or something and once it completes then only shows the correct time..

vibhavagarwal5:
@DonJayamanne

rchiodo:
@vibhavagarwal5 this change has nothing to do with the display of the cell execution time. It's for telemetry. Please log an issue if you are not seeing the expected time in a cell. Thanks.

PR #8349, opened 2021-11-24
Prevent unnecessary activation of Python ext

DonJayamanne:


src/client/datascience/preWarmVariables.ts
```diff
@@ -38,14 +41,17 @@ export class PreWarmActivatedJupyterEnvironmentVariables implements IExtensionSi
...
-        const workspaceFolderCount = this.workspace.workspaceFolders?.length ?? 0;
-        if (workspaceFolderCount <= 5) {
-            void this.envVarsProvider.getEnvironmentVariables(undefined);
-            (this.workspace.workspaceFolders || []).forEach((folder) => {
-                void this.envVarsProvider.getEnvironmentVariables(folder.uri);
-            });
+        if (this.pythonChecker.isPythonExtensionInstalled) {
+            // Don't try to pre-warm variables if user has too many workspace folders opened.
+            const workspaceFolderCount = this.workspace.workspaceFolders?.length ?? 0;
+            if (workspaceFolderCount <= 5) {
```
IanMatthewHuff, line 47: I think I'm fine with this, as it seems like a big number of folders already. But any reason in particular for the magic 5 number?

IanMatthewHuff (approved):

src/client/datascience/preWarmVariables.ts
IanMatthewHuff: Sounds good. I can see it solving a problem and seems pretty unlikely to hit. Just not sure what a customer would think when they add in one more folder (5=>6) and lose prewarms. But that seems really edgy of edge cases.

IanMatthewHuff: Actually maybe we should do a traceInfo in the case where we turn this off? That way if a customer reported slowdown we could see from the log if they bumped to 6 folders? Not sure if this is needed, but could help in the unlikely case someone runs into this.

Issue #8378, opened 2021-11-29
Interactive Widgets fail on Apple Silicon, Remote Jupyter & when using non-Raw kernels


evgenyneu:
## Environment data

-   VS Code version: 1.62.3
-   Jupyter Extension version: v2021.10.1101450599
-   Python Extension version: v2021.11.1422169775
-   OS and version: macOS Monterey 12.0.1 running on AppleSilicon Mac
-   Python version: 3.9.5
-   Type of virtual environment used: venv
-   Jupyter server running: Local 

## Expected behaviour

As you move a slider the function is called multiple times.

## Actual behaviour

The function is called once.

## Steps to reproduce:

* Create a Jupiter notebook with this code, which contains a slider.

```
from ipywidgets import interact

def f(x):
  print(x)
    
interact(f, x=10);
```

* Open it in VSCode, move the slider.

## Expected behaviour

The widget calls the function `f` multiple times as you move the slider.

## Actual behaviour

The function `f` is only called once at the beginning:

![demo_m1](https://user-images.githubusercontent.com/880411/143863722-3f7ae9d8-fe11-4659-b490-62e8c2aa2bf3.gif)

## Notes

* I could only see the bug on Apple Silicon Mac and not on my Intel Mac. 

* The widget works as expected on Apple Silicon Mac if opened in jupuyter-lab instead of VSCode.

## Repository that reproduces the bug

https://github.com/evgenyneu/ipywidgets_vscode_bug

DonJayamanne:
Please could you try the latest VS Code insiders version.
Its possible this has been fixed in there. I'm unable to repro this in VS Code insiders.
* Install VSCode insiders from here https://code.visualstudio.com/insiders/
* Install Jupyter extesnsion (in VS Code Insiders you'll automatically get the latest dev version of Jupyter extension built daily from our main branch)

DonJayamanne:
Weird, works for me with the same python packages.
Please could you try another Python environment, possible something hasn't been installed correctly.
Also, please could you install jupyter & notebook into the same environment (`python -m pip install jupyter notebook`) and run Jupyter from there `python -m jupyter notebook` and run the same code to see whether it works.

evgenyneu:
Thanks, @DonJayamanne 

Tried both conda and venv environments (see commands below), still have this problem in both.  Tried python installed by conda and asdf, same problem. Tried different Python versions (3.9.5, 3.10.0, 3.8.11), no difference.


### Conda

```
[...]

PR #8398, opened 2021-11-30
Interactive widget should work with remote Jupyter

DonJayamanne:
For #8378

Basically interactive widgets has never worked with non-raw kernels.
@rchiodo (try the simple sample here https://github.com/microsoft/vscode-jupyter/issues/8378 and you'll see the `print` output doesn't update when using non-raw or remote).

### DIFF
--- a/src/client/datascience/jupyter/jupyterSession.ts
+++ b/src/client/datascience/jupyter/jupyterSession.ts
@@ -20,7 +20,7 @@ import * as localize from '../../common/utils/localize';
 import { DataScience } from '../../common/utils/localize';
 import { captureTelemetry } from '../../telemetry';
 import { BaseJupyterSession, JupyterSessionStartError } from '../baseJupyterSession';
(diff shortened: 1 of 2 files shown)


src/client/datascience/jupyter/jupyterSession.ts
```diff
@@ -133,6 +134,9 @@ export class JupyterSession extends BaseJupyterSession {
 
             // Make sure it is idle before we return
             await this.waitForIdleOnSession(newSession, this.idleTimeout);
+
+            // So that we don't have problems with ipywidgets, always register the default ipywidgets comm target.
+            newSession.kernel?.registerCommTarget(Identifiers.DefaultCommTarget, noop);
```
DonJayamanne, line 139: This is what we have in raw kernels as well, we need to register comm target for `jupyter.widget`.

rchiodo (approved):

PR #8447, opened 2021-12-03
Tests to ensure we do not prompt for python ext when not required

DonJayamanne:
E.g. for scenarios when user doesn't have python ext and opens a .net notebook.
Users shouldn't get prompted to isntall python extension in these cases.

### DIFF
--- a/src/client/api/pythonApi.ts
+++ b/src/client/api/pythonApi.ts
@@ -137,6 +137,10 @@ export class PythonApiProvider implements IPythonApiProvider {
 export class PythonExtensionChecker implements IPythonExtensionChecker {
     private extensionChangeHandler: Disposable | undefined;
     private waitingOnInstallPrompt?: Promise<void>;
+    /**
+     * Used only for testsing
+     */
+    public static promptDispalyed?: boolean;
     constructor(
         @inject(IExtensions) private readonly extensions: IExtensions,
         @inject(IApplicationShell) private readonly appShell: IApplicationShell,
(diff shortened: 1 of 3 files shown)


src/client/api/pythonApi.ts
```diff
@@ -164,6 +168,7 @@ export class PythonExtensionChecker implements IPythonExtensionChecker {
         if (this.waitingOnInstallPrompt) {
             return this.waitingOnInstallPrompt;
         }
+        PythonExtensionChecker.promptDispalyed = true;
```
DonJayamanne, line 171: THis is the easiest way.
The problem is Jupyter extesnion actiavtes before the tests run, and as part of jupyter activation we might have pre-warming variables and the like which could cause python ext prompt to be dispalyed.
Hence there no way to stub or do anything, except capture a flag to indicate whether the prompt was displayed.

src/client/api/pythonApi.ts
```diff
@@ -137,6 +137,10 @@ export class PythonApiProvider implements IPythonApiProvider {
 export class PythonExtensionChecker implements IPythonExtensionChecker {
     private extensionChangeHandler: Disposable | undefined;
     private waitingOnInstallPrompt?: Promise<void>;
+    /**
+     * Used only for testsing
+     */
+    public static promptDispalyed?: boolean;
```
rchiodo, line 143: Spelling: 'promptDispalyed' should be 'promptDisplayed'

rchiodo (approved):

PR #8449, opened 2021-12-03
Add a warning for %matplotlib

### LINKED ISSUE #8365: Document how to use Matplot lib widgets
Recommend usage of `%matplotlib widget` over others.
Update Wiki
Create a diagnostic message as we do for `!pip install`, users are regularly running into this.

rchiodo:
Fixes #8365

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -229,7 +229,7 @@
                 "--ui=tdd",
                 "--recursive",
                 "--colors",
-                //"--grep", "",
+                "--grep", "Test the RunCurrentCell command",
                 "--timeout=300000"
             ],
             "outFiles": ["${workspaceFolder}/out/**/*.js", "!${workspaceFolder}/**/node_modules**/*"],
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -229,7 +229,7 @@
                 "--ui=tdd",
                 "--recursive",
                 "--colors",
-                "--grep", "Test the RunCurrentCell command",
+                //"--grep", "",
                 "--timeout=300000"
             ],
             "outFiles": ["${workspaceFolder}/out/**/*.js", "!${workspaceFolder}/**/node_modules**/*"],
--- a/news/1 Enhancements/8365.md
+++ b/news/1 Enhancements/8365.md
@@ -1,0 +1,1 @@
+Recommend using %matplotlib widget when any other %matplotlib choice is set.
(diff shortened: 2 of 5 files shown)


.vscode/launch.json
```diff
@@ -229,7 +229,7 @@
                 "--ui=tdd",
                 "--recursive",
                 "--colors",
-                //"--grep", "",
```
rchiodo, line 232: dang it

rchiodo:
Looks like this
image

IanMatthewHuff:
Code looks fine, just a bit unsure on the overall change here. Is widget recommended over inline? Should we change our default?

rchiodo:
> Is widget recommended over inline? Should we change our default?

Not sure what you mean by inline? If somebody is using ```%matplotlib``` (the magic) they are picking a different renderer.

DonJayamanne (approved):

rchiodo:
@IanMatthewHuff were you thinking this comes up anytime the user types 'matplotlib'? 

It's only for the magic, which is a specific command for changing how matplotlib renders. If you're doing this, you're not using inline.

IanMatthewHuff:
> @IanMatthewHuff were you thinking this comes up anytime the user types 'matplotlib'?
> 
> It's only for the magic, which is a specific command for changing how matplotlib renders. If you're doing this, you're not using inline.

It's picking the renderer, and [inline] is a renderer option. We don't want a squiggly for that do we?

rchiodo:
> It's picking the renderer, and [inline] is a renderer option. We don't want a squiggly for that do we?

I didn't even notice inline was an option. A redundant option but nevertheless we should skip the warning on that one.

IanMatthewHuff:
> > It's picking the renderer, and [inline] is a renderer option. We don't want a squiggly for that do we?
> 
> I didn't even notice inline was an option. A redundant option but nevertheless we should skip the warning on that one.

Sure that make sense. The initial change looked like a partial attempt to replace one with the other. If you exclude inline as well then I'm fine with this.

IanMatthewHuff (approved):
Approved if 'inline' is excluded from squiggle warning.

rchiodo:
> Approved if 'inline' is excluded from squiggle warning.

Actually I just tried 'inline'. It crashes the kernel. Seems it's not the default. Not sure what it is.

rchiodo:
Okay never mind, that kernel is just hosed. Another kernel, inline is the default and should work.

PR #8451, opened 2021-12-03
Unified progress messages for kernels

DonJayamanne:
Debt item, something that came up in a Code review.
@rchiodo /cc
Also fixes https://github.com/microsoft/vscode-jupyter/issues/8452

**Solution**
* We have a progress reporter specifically for kernels
* the flow is:
	* Create a progress reporter
	* Report progress anywhere in the code for a give resource (resource is unique per kernel)
* If we report progress and progress reporter is not avaialble, we keep track of the messages to be dispalyed
	* E.g. if auto start is enabled, and we start the kernel, but no progress is displayed
	* Then subsequently users hits the run button,   	
	* Now the progress reporter is created and we display the right progress messsage as we have a list of latest progress messages to be displayed
 

Now:
* If you start multiple notebooks, yo'ull get seperate progress indicators
* Also, this fixes (https://github.com/microsoft/vscode-jupyter/issues/8452) a bug with old approach, if you use Julia and run a cell, progress indicator will appear for a fraction of a few sconds, but after that nothing. The problem is we spend a few seconds waiting for idle, but no progress bar, and no spinner in cells either. hence its as though nothing is happening, now we have a progress bar until we get the spinner.

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -243,7 +243,7 @@
     "DataScienceSurveyBanner.bannerLabelYes": "Yes, take survey now",
     "DataScienceSurveyBanner.bannerLabelNo": "No, thanks",
     "InteractiveShiftEnterBanner.bannerMessage": "Would you like to run code in the 'Interactive' window (an IPython console) for 'shift-enter'? Select 'No' to continue to run code in the Python Terminal. This can be changed later in settings.",
-    "DataScience.restartingKernelStatus": "Restarting Jupyter Kernel",
+    "DataScience.restartingKernelStatus": "Restarting Kernel {0}",
     "DataScience.executingCode": "Executing Cell",
     "DataScience.collapseAll": "Collapse all cell inputs",
     "DataScience.expandAll": "Expand all cell inputs",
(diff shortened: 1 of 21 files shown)


package.nls.json
```diff
@@ -243,7 +243,7 @@
     "DataScienceSurveyBanner.bannerLabelYes": "Yes, take survey now",
     "DataScienceSurveyBanner.bannerLabelNo": "No, thanks",
     "InteractiveShiftEnterBanner.bannerMessage": "Would you like to run code in the 'Interactive' window (an IPython console) for 'shift-enter'? Select 'No' to continue to run code in the Python Terminal. This can be changed later in settings.",
-    "DataScience.restartingKernelStatus": "Restarting Jupyter Kernel",
+    "DataScience.restartingKernelStatus": "Restarting Kernel {0}",
```
DonJayamanne, line 246: Removed the word `Jupyter`, we don't use jupyter when `connecting to kernel`
Also added the `{0}`, so we can display the name of the kernel when we're restarting it.

src/client/datascience/progress/kernelProgressReporter.ts
```diff
@@ -0,0 +1,182 @@
...
+                        progressInfo.progressList.splice(index);
+                    }
+                    // If we have previous messages, display the last item.
+                    if (progressInfo.progressList.length > 0) {
+                        progressInfo.reporter.report({
+                            message: progressInfo.progressList[progressInfo.progressList.length - 1]
+                        });
+                    }
+                    // If we have no more messages, then remove the reporter.
+                    if (progressInfo.progressList.length === 0) {
```
rchiodo, line 134: could this be an else?

rchiodo (approved):

PR #8458, opened 2021-12-04
Clear separation between local and remote kernelspecs

DonJayamanne:
Debt
We have LocalKernelSpecs vs Remote kernel specs.
In the code we don't make a distinction, however we do in places by determing whehter we're using a local or remote connection & then change accordingly.

The plan is to disconnect the notion of having local & remote seprately. The kernel connection metadata will tell us whether its local vs remote.

This is a first stage towards not having to reload vS Code when chaning remote kernel Url.

Either way, IMHO is a good change to clearly distinguish between local vs remote.
E.g. if you look at the kernel Launcher, and raw kernel stuff we only deal with `LocalKernelSpecConnection` and not the type `RemoteKernelConnection`
Better typing as well.

### DIFF
--- a/src/client/datascience/jupyter/kernels/helpers.ts
+++ b/src/client/datascience/jupyter/kernels/helpers.ts
@@ -18,7 +18,7 @@ import { EnvironmentType, PythonEnvironment } from '../../../pythonEnvironments/
 import {
     IKernel,
     KernelConnectionMetadata,
-    KernelSpecConnectionMetadata,
+    LocalKernelSpecConnectionMetadata,
     LiveKernelConnectionMetadata,
     PythonKernelConnectionMetadata
 } from './types';
@@ -55,7 +55,7 @@ export function findIndexOfConnectionFile(kernelSpec: Readonly<IJupyterKernelSpe
     return kernelSpec.argv.findIndex((arg) => arg.includes(connectionFilePlaceholder));
 }
 
-type ConnectionWithKernelSpec = KernelSpecConnectionMetadata | PythonKernelConnectionMetadata;
+type ConnectionWithKernelSpec = LocalKernelSpecConnectionMetadata | PythonKernelConnectionMetadata;
 export function kernelConnectionMetadataHasKernelSpec(
     connectionMetadata: KernelConnectionMetadata
 ): connectionMetadata is ConnectionWithKernelSpec {
@@ -130,7 +130,8 @@ export function getDisplayNameOrNameOfKernelConnection(kernelConnection: KernelC
         case 'connectToLiveKernel': {
             return oldDisplayName;
         }
-        case 'startUsingKernelSpec': {
+        case 'startUsingRemoteKernelSpec':
+        case 'startUsingLocalKernelSpec': {
             if (
                 kernelConnection.interpreter?.envType &&
                 kernelConnection.interpreter.envType !== EnvironmentType.Global
@@ -209,7 +210,9 @@ export function getKernelPathFromKernelConnection(kernelConnection?: KernelConne
         : undefined;
     if (
         kernelConnection.kind === 'startUsingPythonInterpreter' ||
-        (kernelConnection.kind === 'startUsingKernelSpec' && kernelConnection.kernelSpec.language === PYTHON_LANGUAGE)
+        ((kernelConnection.kind === 'startUsingRemoteKernelSpec' ||
+            kernelConnection.kind === 'startUsingLocalKernelSpec') &&
+            kernelConnection.kernelSpec.language === PYTHON_LANGUAGE)
     ) {
         return kernelSpec?.metadata?.interpreter?.path || kernelSpec?.interpreterPath || kernelSpec?.path;
     } else {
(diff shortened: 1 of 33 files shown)


src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -785,7 +790,9 @@ export function findPreferredKernel(
                 subScore === 5 && // This is a bit flakey. Number isn't consistent. Should probably just make the order of kernelspecs have the preferred one first
                 score === 5 &&
                 (metadata.kind === 'startUsingPythonInterpreter' ||
-                    (metadata.kind === 'startUsingKernelSpec' && metadata.kernelSpec.language === PYTHON_LANGUAGE)) &&
+                    metadata.kind === 'startUsingRemoteKernelSpec' ||
```
rchiodo, line 793: Why would the logic be different for remote non python kernel specs?

src/client/datascience/jupyter/kernels/types.ts
```diff
@@ -53,7 +53,19 @@ export type KernelSpecConnectionMetadata = Readonly<{
...
+    id: string;
+}>;
+/**
+ * Connection metadata for Remote Kernels started using kernelspec (JSON).
+ * This could be a raw kernel (spec might have path to executable for .NET or the like).
+ * If the executable is not defined in kernelspec json, & it is a Python kernel, then we'll use the provided python interpreter.
+ */
+export type RemoteKernelSpecConnectionMetadata = Readonly<{
+    kernelModel?: undefined;
+    interpreter?: undefined;
```
rchiodo, line 66: Remote kernels shouldn't have interpreters, right? There's no interpreter on the remote machine.

src/client/datascience/jupyter/kernels/helpers.ts
DonJayamanne: Not sure, this is exisit code, let me see.

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -476,7 +480,7 @@ function getKernelConnectionCategory(kernelConnection: KernelConnectionMetadata)
     switch (kernelConnection.kind) {
         case 'connectToLiveKernel':
             return DataScience.kernelCategoryForJupyterSession();
-        case 'startUsingKernelSpec':
+        case 'startUsingLocalKernelSpec':
```
rchiodo, line 483: Shouldn't the remote kernelspec be listed here too?

src/client/datascience/jupyter/kernels/types.ts
DonJayamanne: Yes, but i've had to add this property because we try to check the property `interpreter` in the unified type.
If i don't add this then code won't compile after all, not all kernelspecconnection types will have `interpertetr`.
Hence I've added a dummy propety that will always be either undefied or non-existent.

src/client/datascience/notebook/vscodeNotebookController.ts
DonJayamanne: Oh, I thought I added it, must have remvoed it, I ermember testing this damn part

src/client/datascience/jupyter/kernels/types.ts
DonJayamanne: Fixed

src/client/datascience/notebook/vscodeNotebookController.ts
DonJayamanne: Fixed

src/client/datascience/jupyter/kernels/types.ts
```diff
@@ -44,7 +44,7 @@ export type LiveKernelConnectionMetadata = Readonly<{
  * This could be a raw kernel (spec might have path to executable for .NET or the like).
  * If the executable is not defined in kernelspec json, & it is a Python kernel, then we'll use the provided python interpreter.
  */
-export type KernelSpecConnectionMetadata = Readonly<{
+export type LocalKernelSpecConnectionMetadata = Readonly<{
```
rchiodo, line 47: Do you think we could get rid of the 'startUsingPythonInterpreter' type and just use the existence of the interpreter in the LocalKernelSpecConnectionMetadata instead?

rchiodo (approved):

src/client/datascience/jupyter/kernels/types.ts
DonJayamanne: I persoanlly find the current approach much cleaner, i.e. looking for the name is better than checkinf if `interpreeter` exists or not.
Also that doesn't work for non pytohn kernels. 
E.g. if you have a Java kernel, the interpreter could exist, this means we need to actiavte the interpreter and then run the java kernel.
E.g. conda that registers a java kernel.
Sure we could check the language and make that determination, however I personally find the explicit name to be better (i.e. explicit), instead of checkinf multiple flags.

If we removed this then all of the code where we have `type === startUsingPythonInterpreter` would change to `interpreer !== undefined && kernelspec.langauge === python`
Which feels clunky, miss one thing on some place, you're stuffed.

rchiodo: Maybe a different way to do this would be change ```LocalKernelSpecConnectionMetadata``` to ```NonPythonKernelSpecConnectionMetadata```?

rchiodo: I find it confusing we have 'start with interpreter' and 'kernel spec with interpreter maybe'.

Issue #1367, opened 2020-10-19
Can't change remote Jupyter server without restarting VS Code


clayms:
Issue Type: <b>Bug</b>

After one of the recent updates, I can no longer change the remote Jupyter server without restarting VS Code.  I used to be able to simply comment out the remote and local Jupyter  `python.dataScience.jupyterServerURI`s that I was not using and un-comment the one that I wanted to use and after closing the VS Code Python notebook pane everything would switch just fine.  Now after changing the `python.dataScience.jupyterServerURI` it is only recognized after I restart VS Code.

Is there a way to get back the old behavior?  It was much more convenient. 

```
	"settings":  {
		// local
		// "python.dataScience.jupyterServerURI":  "local"
		// Remote # 1
		// "python.dataScience.jupyterServerURI": "http://100.100.100.001:8888/?token=*******************"
		// Remote # 2
		"python.dataScience.jupyterServerURI": "http://100.100.100.002:8888/?token=*******************"
	 }
```

Extension version: 2020.9.114305
VS Code version: Code 1.50.1 (d2e414d9e4239a252d1ab117bd7067f125afd80a, 2020-10-13T15:06:15.712Z)
OS version: Windows_NT x64 10.0.18363

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-6300U CPU @ 2.40GHz (4 x 2496)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: disabled_off<br>opengl: enabled_on<br>protected_video_decode: unavailable_off<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|15.88GB (5.36GB free)|
|Process Argv|--crash-reporter-id f652528b-5678-4b12-9ca0-3de6b1e1135f|
|Screen Reader|no|
|VM|0%|
</details>

StarryFire:
This is a problem since I am trying to connect VSCode which is running inside a container...and in the same container my instance of jupyter server is also running and hence it never ends up connecting since the token changes on every restart.

sanzgiri:
[...]

PR #8468, opened 2021-12-07
Reloading VS Code not required when switching to remote or vice versa

DonJayamanne:
For #https://github.com/microsoft/vscode-jupyter/issues/1367


src/client/datascience/types.ts
```diff
@@ -844,7 +845,7 @@ export type WebViewViewChangeEventArgs = { current: WebViewViewState; previous:
 export type GetServerOptions = {
     getOnly?: boolean;
     ui: IDisplayOptions;
-    localOnly?: boolean;
+    localOnly: boolean;
```
DonJayamanne, line 848: Being explicy is better, we pass this around into `ConnectNotebookProviderOptions`, 
I found a lot of these and removed most, this is another one where I think its not great, as we have no idea where these are set at all, if you search through the code the property is never initialized becase we take one type & cast into another ..

src/client/datascience/types.ts
```diff
@@ -148,7 +148,7 @@ export interface INotebook {
 export type ConnectNotebookProviderOptions = {
     getOnly?: boolean;
     ui: IDisplayOptions;
-    localOnly?: boolean;
+    localOnly: boolean;
```
DonJayamanne, line 151: Planning on renaming this, basically to mean `start local server` (i.e. when we start jupyter for local purposes).
I thought that's what this mean, but that's not true.

DonJayamanne: Will need a good meaningful name

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -481,6 +479,7 @@ function getKernelConnectionCategory(kernelConnection: KernelConnectionMetadata)
         case 'connectToLiveKernel':
             return DataScience.kernelCategoryForJupyterSession();
         case 'startUsingRemoteKernelSpec':
+            return `Remote ${DataScience.kernelCategoryForJupyterKernel()}`;
```
DonJayamanne, line 482: Temporary

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -118,7 +118,6 @@ export class VSCodeNotebookController implements Disposable {
         disposableRegistry: IDisposableRegistry,
         private readonly languageService: NotebookCellLanguageService,
         private readonly workspace: IWorkspaceService,
-        private readonly localOrRemoteKernel: 'local' | 'remote',
```
DonJayamanne, line 121: Yay, no need, whether it remote or local, is derived from the kernel connection metadata

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -705,48 +757,3 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         });
     }
 }
-
-export function getControllerDisplayName(kernelConnection: KernelConnectionMetadata, currentDisplayName: string) {
```
DonJayamanne, line 709: Unused

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -634,28 +656,58 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+            this.getRemoteKernelConnectionMetadata(token)
+        ]);
+        return local.concat(remote);
+    }
 
-                return this.remoteKernelFinder.listKernels(undefined, connection, token);
-            } finally {
-                ui.dispose();
-            }
+    private async getLocalKernelConnectionMetadata(
```
DonJayamanne, line 677: To be discussed:
* Currently you can use local & remote together
* I.e. if you have a remote Uri, we will list both (some users have asked for this)

Not sure whether this is a good idea, perhaps we can turn this into a flag/setting when users defines the remote Uri, we can ask them whether they'd like to see the local kernel as well (instead of a setting, else they'll never discover that).

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -575,7 +598,6 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
                         this.disposables,
                         this.languageService,
                         this.workspace,
-                        this.isLocalLaunch ? 'local' : 'remote',
```
DonJayamanne, line 578: Yay, no need, whether it remote or local, is derived from the kernel connection metadata

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -84,7 +83,9 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
     }
     private preferredControllers = new Map<NotebookDocument, VSCodeNotebookController>();
 
-    private readonly isLocalLaunch: boolean;
+    private get isLocalLaunch(): boolean {
+        return isLocalLaunch(this.configuration);
```
DonJayamanne, line 87: Will need a better flag, 
If `isLocalLaunch` === true, then tshi means we have a remote uri,
however in this PR, we support local & remote together, ie. even when we're laucning local kernels we can be working against remote kernels.

src/client/datascience/interactive-common/notebookServerProvider.ts
```diff
@@ -56,23 +68,27 @@ export class NotebookServerProvider implements IJupyterServerProvider {
             this.ui.disableUI = false;
         }
         options.ui.onDidChangeDisableUI(() => (this.ui.disableUI = options.ui.disableUI), this, this.disposables);
-
-        if (!this.serverPromise) {
+        const property = options.localOnly ? 'local' : 'remote';
+        if (!this.serverPromise[property]) {
```
DonJayamanne, line 72: Caching local & remote seprately (this is so that you can start jupyter locally - non-raw as well as connect to remote jupyter).

DonJayamanne:
@rchiodo what do you think of the approach taken, that would allow us to pre-vent reloading.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -377,6 +378,24 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         }
     }
 
+    private handlerAddedForRemoteKernels?: boolean;
+    private reloadControllersAfterChangingRemote() {
+        if (this.handlerAddedForRemoteKernels) {
+            return;
+        }
+        this.handlerAddedForRemoteKernels = true;
+        this.serverUriStorage.onDidChangeUri(async () => {
```
DonJayamanne, line 387: When remote URI changes refresh the kernels.
Will need to add more smarts to refresh this when we start a start/stop kernel as well (so that we get teh live kernels showing up)

rchiodo:
I think we'd need to discuss this with the larger team at standup. This is a change in behavior for the UI.

src/test/datascience/interactive-common/notebookServerProvider.unit.test.ts
```diff
@@ -104,10 +104,10 @@ suite('DataScience - NotebookServerProvider', () => {
 
         // Disable UI just lets us skip mocking the progress reporter
         const server = await serverProvider.getOrCreateServer({
-            getOnly: false,
```
DonJayamanne, line 107: This was never set

src/client/datascience/types.ts
```diff
@@ -842,9 +847,11 @@ type WebViewViewState = {
 export type WebViewViewChangeEventArgs = { current: WebViewViewState; previous: WebViewViewState };
 
 export type GetServerOptions = {
-    getOnly?: boolean;
```
DonJayamanne, line 845: This was never used (i.e. never set)

src/client/datascience/types.ts
```diff
@@ -146,9 +145,11 @@ export interface INotebook {
 
 // Options for connecting to a notebook provider
 export type ConnectNotebookProviderOptions = {
-    getOnly?: boolean;
```
DonJayamanne, line 149: This was never used (i.e. never set)

src/client/datascience/interactive-common/notebookProvider.ts
```diff
@@ -43,26 +44,27 @@ export class NotebookProvider implements INotebookProvider {
             }
         });
         // Connect to either a jupyter server or a stubbed out raw notebook "connection"
-        if (this.rawNotebookProvider.isSupported) {
+        if (this.rawNotebookProvider.isSupported && !options.local) {
```
rchiodo, line 47: This seems backwards? Raw is always local?

rchiodo (approved):
Looks good except for the comment I made in notebookProvider.ts

src/client/datascience/interactive-common/notebookProvider.ts
DonJayamanne: Fixed the property, it was poorly named, renamed to `localJupyter` that propery would be true ONLY when starting jupyter for local

PR #8478, opened 2021-12-08
Upgrade to latest lsp-middleware and add a bunch of tests

### LINKED ISSUE #8435: Test needed for semantic token colorization
Multiple failures with this in the last couple of weeks. It's a bit hard to test, but might be able to stick a middleware piece in during the test. (Modify the 'languageServer' class to allow this as it already generates the middleware from the npm package)

rchiodo:
Fixes #8435

### DIFF
--- a/src/test/datascience/notebook/intellisense/semanticTokens.vscode.test.ts
+++ b/src/test/datascience/notebook/intellisense/semanticTokens.vscode.test.ts
@@ -4,7 +4,7 @@
 /* eslint-disable @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires */
 import { assert } from 'chai';
 import * as sinon from 'sinon';
-import { commands } from 'vscode';
+import { commands, Position, window } from 'vscode';
 import { IVSCodeNotebook } from '../../../../client/common/application/types';
 import { traceInfo } from '../../../../client/common/logger';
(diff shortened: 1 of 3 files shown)


rchiodo:
@DonJayamanne can you take a look?

src/test/datascience/notebook/intellisense/semanticTokens.vscode.test.ts
```diff
@@ -91,4 +89,78 @@ suite('DataScience - VSCode semantic token tests', function () {
...
+            true
+        );
+
+        // Then get tokens on second cell. They should start on line 1. If not this
+        // means there's a bug
+        const tokens = (await commands.executeCommand(
+            'vscode.provideDocumentSemanticTokens',
+            cell2.document.uri
+        )) as any;
+        assert.ok(tokens, 'No tokens found on second cell');
```
amunger, line 123: assertions are the same as the previous test, but we made an edit. Can we check that the edit was performed?

rchiodo: Possibly. I could verify the the tokens are not the same as the previous tests tokens (because of the edit). Or hardcode the tokens here and see that they match

DonJayamanne (approved):

Issue #8485, opened 2021-12-08
Remove `jupyter.runMagicCommands` setting


DonJayamanne:
This has been removed for over a year now, if users are using this, they'd be getting errors (warnings about this being an unrecognized setting).
The new setting is `runStartupCommands`.

PR #8486, opened 2021-12-08
Use runStartupCommand instead of runMagicCommand

DonJayamanne:
For #8485

### DIFF
--- a/news/3 Code Health/8485.md
+++ b/news/3 Code Health/8485.md
@@ -1,0 +1,1 @@
+Remove `jupyter.runMagicCommands` (which has been deprecared for over a year) in favor of `runStartupCommands`.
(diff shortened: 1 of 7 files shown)


news/3 Code Health/8485.md
```diff
@@ -0,0 +1 @@
+Remove `jupyter.runMagicCommands` (which has been deprecared for over a year) in favor of `runStartupCommands`.
```
rchiodo, line 1: Shouldn't we remove it from the package.json? I thought VS code did that with deprecated stuff eventually?

rchiodo (approved):

news/3 Code Health/8485.md
DonJayamanne: It never existed in the package.json, hence the PR, it hasn't been there for a long time.

IanMatthewHuff (approved):

Issue #8489, opened 2021-12-08
Display local and remote kernels together


DonJayamanne:

DonJayamanne:
Test plan item https://github.com/microsoft/vscode-jupyter/issues/8778

PR #8490, opened 2021-12-08
Display local & remote kernels together

DonJayamanne:
For #8489

Need to add some tests
* [x] Run a cell against a local kernel, connect to remote Jupyter and verify the local kernel state is not lost, 
* [x] Run cell in same notebook with local kernel and remote kernel
* [x] Ensure local and remote kernels are listed in controllers when connecting to remote
* [x] Ensure only local kernels are listed in controllers when switching to local
* [x] Keep track of the controllers (base url) and verify it is different when connecting to a different remote Jupyter (with a different port)


src/client/common/utils/multiStepInput.ts
```diff
@@ -106,6 +117,19 @@ export class MultiStepInput<S> implements IMultiStepInput<S> {
                 input.placeholder = placeholder;
                 input.ignoreFocusOut = true;
                 input.items = items;
+                if (onDidChangeItems) {
```
DonJayamanne, line 120: As per demo, allow one to remove individual items

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -95,9 +95,10 @@ export function getKernelId(spec: IJupyterKernelSpec, interpreter?: PythonEnviro
         // Lets not assume that non-python kernels cannot have such issues
         argsForGenerationOfId = spec.argv.join('#').toLowerCase();
     }
-    return `${spec.id || ''}.${specName}.${getNormalizedInterpreterPath(spec.interpreterPath || spec.path)}.${
-        getNormalizedInterpreterPath(interpreter?.path) || ''
-    }.${argsForGenerationOfId}`;
+    const prefixForRemoteKernels = remoteBaseUrl ? `${remoteBaseUrl}.` : '';
+    return `${prefixForRemoteKernels}${spec.id || ''}.${specName}.${getNormalizedInterpreterPath(
```
DonJayamanne, line 99: Give a unique id for remote kernels (include the base url)
Else if you start jupyter locally, they could have the same ids (as they point to the same kernelspec)

An added benefit is the fact that this now allows us to support multiple remotes at a time.

rchiodo: This might be rather long? Seems a bit much? Like from azure it can be ```azure.westcoastus.en.mysillyloginname/jupyter/server/332```. I guess we can wait for people complaining though.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -207,10 +209,13 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
                     );
                 });
 
+            // Fetch kernel the fastest possible way (local kernels from cache but remote fetch latest).
             // Fetch the list of kernels from the cache (note: if there's nothing in the case, it will fallback to searching).
+            // Fetching remote kernels cannot be done from cache.
             this.controllersPromise = Promise.all([
-                this.loadNotebookControllersImpl(true, 'useCache'),
-                this.loadNotebookControllersImpl(false, 'useCache')
+                this.loadNotebookControllersImpl(true, false, 'useCache'),
```
rchiodo, line 216: Options instead of args would be nice to have here (so I know what 'true', 'false' is actually setting.)

rchiodo (approved):

src/client/datascience/notebook/notebookControllerManager.ts
DonJayamanne: Agreed

src/client/datascience/jupyter/kernels/helpers.ts
IanMatthewHuff: Does long matter? I don't think that the kernel ID is user visible as all unless I'm off on this.

IanMatthewHuff (approved):

src/client/datascience/notebook/notebookControllerManager.ts
DonJayamanne: I'll do that seprately, found a few other things I'd like to clean up

src/client/datascience/jupyter/kernels/helpers.ts
rchiodo: Yeah it doesn't matter. I was thinking this was in the display. For the id it shouldn't matter.

src/client/datascience/notebook/notebookControllerManager.ts
DonJayamanne: Never mind, done.

PR #8510, opened 2021-12-09
Fix reopening a remote kernel

### LINKED ISSUE #7610: Opening a notebook on remote, closing it, and reopening it does not reuse the same kernel
Switch to remote jupyter.
Make some cells like so:
```
a = 1
```
```
print(a)
```
Run all cells
Close notebook
Reopen.
Run JUST second cell.

It should print A = 1.
It doesn't open the same kernel.

There's a test we had that was supposed to verify this but it was artificially forcing the right kernel to be selected.

rchiodo:
Fixes #7610

Made a new event on the controller indicating when remote kernels are refreshed. This allows me to listen for that and update a memento setting for the kernel id.

Initially I had attempted to just do that for all kernel starts, but the notebookcontrollermanager doesn't have the kernel at that point so there's no way to tell which notebook it's for. I needed an event that fired AFTER the controller has updated because of kernel start.


rchiodo:
Oh and I also changed the UI a little bit.

Old UI shows this for remote kernels in the selector:

image

I changed it so that the notebook path was part of the real name:

image

The reason for this is because there's no way to tell you're connected to an already running kernel in the notebook without it.

All of the kernels look the same:

image


Whereas now the existing ones will look like this:

image

DonJayamanne (approved):

src/client/common/extensions.ts
```diff
@@ -108,6 +112,17 @@ Promise.prototype.ignoreErrors = function <T>(this: Promise<T>) {
     this.catch(() => {});
 };
 
+/**
+ * Logs an error for any promise.
+ */
+Promise.prototype.logErrors = function <T>(this: Promise<T>) {
+    // eslint-disable-next-line no-empty, @typescript-eslint/no-empty-function
+    this.catch((e) => {
```
DonJayamanne, line 120: I would avoid this, creating this means you MUST import `extesnsions.ts` everywhere, and the moment you miss it, things could fall over.

IanMatthewHuff (approved):

PR #8512, opened 2021-12-10
Don't keep deleting kernelspecs generated by the ext for new users

DonJayamanne:
I think we've had this for 2 releases and that should be good enough.

### DIFF
--- a/src/client/datascience/kernel-launcher/localKnownPathKernelSpecFinder.ts
+++ b/src/client/datascience/kernel-launcher/localKnownPathKernelSpecFinder.ts
@@ -6,19 +6,16 @@ import { inject, injectable, named } from 'inversify';
 import { CancellationToken, Memento } from 'vscode';
 import { IWorkspaceService } from '../../common/application/types';
 import { IFileSystem } from '../../common/platform/types';
-import * as path from 'path';
-import * as fs from 'fs-extra';
-import { getKernelId, isKernelRegisteredByUs } from '../jupyter/kernels/helpers';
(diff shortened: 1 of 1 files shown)


rchiodo (approved):

src/client/datascience/kernel-launcher/localKnownPathKernelSpecFinder.ts
```diff
@@ -28,22 +25,10 @@ import { traceInfo } from '../../common/logger';
...
-    private _oldKernelSpecsDeleted = false;
     private _oldKernelSpecsFolder?: string;
-    private get oldKernelSpecsDeleted() {
-        return this._oldKernelSpecsDeleted || this.memento.get<boolean>('OLD_KERNEL_SPECS_DELETED__', false);
-    }
-    private set oldKernelSpecsDeleted(value: boolean) {
-        this._oldKernelSpecsDeleted = value;
-        void this.memento.update('OLD_KERNEL_SPECS_DELETED__', value);
-    }
     private get oldKernelSpecsFolder() {
```
IanMatthewHuff, line 29: Could this (and associated) go as well? Looks like it's just the trace info in the constructor using it.

IanMatthewHuff (approved):
One more thing could possibly go?

src/client/datascience/kernel-launcher/localKnownPathKernelSpecFinder.ts
DonJayamanne: I'm leaving the logging in for now, lets remove that later.

PR #8515, opened 2021-12-10
Remove obsolete expand/collapse t est

### LINKED ISSUE #8492: Test for collapse expand is failing
Code here:
https://github.com/microsoft/vscode-jupyter/blob/792872e14ae1711ed343f5d7feba22ac84f45d22/src/test/datascience/interactiveWindow.vscode.test.ts#L215

I believe @roblourens commit here changed how we track this. We need some other way to do so.

I'm disabling the test for now.

roblourens:
Fix #8492

### DIFF
--- a/src/client/datascience/interactive-window/interactiveWindow.ts
+++ b/src/client/datascience/interactive-window/interactiveWindow.ts
@@ -50,7 +50,6 @@ import { initializeInteractiveOrNotebookTelemetryBasedOnUserAction } from '../te
 import { InteractiveWindowView } from '../notebook/constants';
 
 type InteractiveCellMetadata = {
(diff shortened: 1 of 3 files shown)


rchiodo:
The test is already disabled. I think we'd rather have a test for collapse expand if possible? Are you saying there's no way to detect expand collapse state anymore?

DonJayamanne:
> test is already disabled. I think we'd rather have a test for collapse expand if possible?

I think we need this.
It's a very crucial feature in IW

roblourens:
There is no extension API to read the expand/collapse state

DonJayamanne (approved):

PR #8524, opened 2021-12-10
Add support for Python 3.10 on CI

DonJayamanne:
Found 1 bug as a result of this PR

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -219,14 +219,14 @@ jobs:
           file-pattern: 'telemetry.md'
           skip-label: 'skip telemetry'
           failure-message: 'TELEMETRY.md should be updated when adding or removing telemetry. Please run "npm run generateTelemetry"'
-    
+
       - name: 'Check for news entry'
         uses: brettcannon/check-for-changed-files@v1.1.0
         with:
           file-pattern: 'news/*/*.md'
           skip-label: 'skip news'
           failure-message: 'News entry file missing; see news/README.md for instructions (the ${skip-label} label can be used to pass this check)'
-          
+
   ts_tests:
     needs: pick_environment
     name: Type Script Tests
@@ -339,6 +339,9 @@ jobs:
           - jupyter: raw
             python: noPython
             os: ubuntu-latest
+        exclude:
+          - python: conda  # Conda does't seem to support 3.10 yet.
+            pythonVersion: '3.10'
         #   - pythonVersion: '3.9'
         #     python: nonConda
         #     jupyter: raw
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test.yml
```diff
@@ -573,15 +488,6 @@ jobs:
           beakerx_kernel_java install
           conda install pytorch cpuonly -c pytorch
 
-      # Install PyTorch with pip
-      - name: Install PyTorch on Windows and Linux
-        if: matrix.python != 'conda' && matrix.python != 'noPython' && (matrix.os == 'windows-latest' || matrix.os == 'ubuntu-latest')
-        run: pip install torch==1.7.1+cpu torchvision==0.8.2+cpu torchaudio===0.7.2 -f https://download.pytorch.org/whl/torch_stable.html
```
DonJayamanne, line 579: None of these are used,
We have these for some old tests that relate to fancy tooltips for tensor varaibles (and that functionality is not in the extension at all).

.github/workflows/build-test.yml
```diff
@@ -317,20 +317,15 @@ jobs:
...
         # macOS runners are expensive, and we assume that Ubuntu is enough to cover the UNIX case.
         os: [ubuntu-latest]
         # More details on includes/excludes can be found here https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations
         # Basically with exclude, you can exclude any of the combinations from the result matrix.
         # & with include, you can include additional items to the result matrix.
-        exclude:
-          - os: ubuntu-latest # We don't care how users start Remote Jupyter as long as it works.
-            pythonVersion: '3.9'
-            python: conda
-            jupyter: remote
```
DonJayamanne, line 331: THis isn't required, as we're already including this seprately (was meant to be removed in the previous PR)

.github/workflows/build-test.yml
```diff
@@ -317,20 +317,15 @@ jobs:
       matrix:
         jupyter: [raw, local]
         python: [nonConda, conda]
-        pythonVersion: [3.9]
+        pythonVersion: ['3.9', '3.10']
```
rchiodo, line 320: I believe these are going to fail? Or maybe Ian hasn't submitted his new test yet.

rchiodo (approved):

.github/workflows/build-test.yml
IanMatthewHuff: I would expect it to fail. At least it did for me on my local machine. Didn't try on CI yet though.

DonJayamanne: I have filed an issue and yes, it will fail for Python 3.10.
I'll skip the test if using Python 3.10

IanMatthewHuff: Test validated :). 
image
image

IanMatthewHuff (approved):

PR #8538, opened 2021-12-13
Update completion filtering logic and make completions happen faster

### LINKED ISSUE #8536: Jupyter dataframe column names is inconsistent
1. Get a csv with a couple of columns
2. Create a dataframe from them
3. Run the cells
4. Run ```df.``` and see if the column names appear.

Expected results:
They always do

Actual result:
They sometimes do.

rchiodo:
Fixes #8536

Fixing a couple of issues:
- The first completion from jupyter can take 600ms on my machine. Subsequent ones are much faster, so do a dummy request on startup.
- The sleep timeout is always firing the sleep resolver. Use waitForPromise instead which kills the sleep timer.
- Typing autocomplete on anything other than a '.' doesn't work because the sub word filtering was incorrect. Added a test for this case too.

### DIFF
--- a/src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
+++ b/src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
@@ -18,8 +18,9 @@ import { createPromiseFromCancellation } from '../../../common/cancellation';
 import { traceError, traceInfoIfCI, traceVerbose } from '../../../common/logger';
 import { getDisplayPath } from '../../../common/platform/fs-paths';
 import { IConfigurationService } from '../../../common/types';
-import { sleep } from '../../../common/utils/async';
+import { waitForPromise } from '../../../common/utils/async';
 import { isNotebookCell } from '../../../common/utils/misc';
+import { StopWatch } from '../../../common/utils/stopWatch';
 import { Settings } from '../../constants';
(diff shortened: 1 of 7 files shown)


src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
```diff
@@ -74,22 +75,15 @@ export class PythonKernelCompletionProvider implements CompletionItemProvider {
...
         traceInfoIfCI(`Notebook completion request for ${document.getText()}, ${document.offsetAt(position)}`);
         const [result, pylanceResults] = await Promise.all([
-            Promise.race([
+            waitForPromise(
                 this.getJupyterCompletion(kernel.session, document.getText(), document.offsetAt(position), token),
-                sleep(timeout).then(() => {
-                    if (token.isCancellationRequested) {
-                        return;
-                    }
-                    traceInfoIfCI(`Notebook completions request timed out for Cell ${getDisplayPath(document.uri)}`);
```
rchiodo, line 89: This was firing even after we already returned results. Tracing this messed me up. 

so using 'waitForPromise' which kills the timer.

rchiodo:
Can't repro the test failure locally of course. I don't think it's related to my change though.

rchiodo:
@DonJayamanne or @amunger can you guys take a look?

DonJayamanne (approved):

Issue #5532, opened 2021-04-15
Expose kernels to 3rd party extensions


DonJayamanne:
This is a hackathon project, and the some of the original requirements can be found here https://github.com/microsoft/vscode-jupyter/issues/3998

**Goal: Expose kernel to 3rd party extension authors so they can build on top of Jupyter extension**

**Possible extensions (prototyped):**
* [x] Other extensions can now execute code against kernels
	* User will be asked to allow such exensions to execute on behalf of them
	* 3rd party extensions can install missing packages, & the like (prototyped)
	* 3rd party extensions can inject kernel exensions (prototyped)
	* 3rd party extensions can intercept kernel execution requests
* [ ] SOS Jupyter Kernel (partially working)
	* We need cell pre-execute
* [x] Support functionality found in Jupyter Sidecar widget
	* Allow user to open widgets outside the notebook.
* [x] Custom variable explorer
	* Support for other languages (currently working, Python, R, Scala, C#, F#)
	* Support for variable visualizers (currently working for Python)
	* Other extensions can build on top and support more visualizers for each language
	* TOOD: 
		* Provide extensibility for other extensions to provide details of each variable so that the tree view can be augmented (e.g. drilling down into details of each variable)
		*  Ability to drill down into dict, dataframes in the variable viewer (partially protototyped)
* [x] Better plot viewer
	* Custom visualizer for variable, to open a plot in a seprate view (use ipympl)
	* User doesn't need to isntall ipympl (source shipped with 3rd party extension)
* [x] Custom data frameviewer
	* Basically similar to rest, display dataframe variable with the qgrid widget
	* Ship qgrid widget with 3rd party extension
	* Partially working
* [ ] Open altair (vega) plots with vega viewers
	* Convert vega into JSON, use 3rd party extensions to view the vega json data
	* I.e. more custom visualizers, just built on top of this new functionality

![Screen Shot 2021-04-15 at 11 29 23](https://user-images.githubusercontent.com/1948812/114951115-4ab1e980-9e09-11eb-9131-0db2f1a799a6.png)
![Screen Shot 2021-04-15 at 13 24 11](https://user-images.githubusercontent.com/1948812/114951110-47b6f900-9e09-11eb-9835-c21e3d233639.png)
![Screen Shot 2021-04-15 at 12 51 03](https://user-images.githubusercontent.com/1948812/114951111-484f8f80-9e09-11eb-8584-5aa912a08494.png)

[...]

PR #8545, opened 2021-12-14
Expose kernel  API

DonJayamanne:
For #5532

* [x] Allow anyone to use this in debug mode
* [x] Add telemetry for usage


src/client/api/extension.d.ts
```diff
@@ -0,0 +1,256 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
DonJayamanne, line 1: Later we can publish this as npm or github package, else users have to hand craft this.
For now lets get users to manually copy this over.

src/client/api/extension.d.ts
```diff
@@ -0,0 +1,256 @@
...
+     */
+    interpreterPath?: string;
+    readonly interrupt_mode?: 'message' | 'signal';
+}
+/**
+ * Connection metadata for Kernels started using kernelspec (JSON).
+ * This could be a raw kernel (spec might have path to executable for .NET or the like).
+ * If the executable is not defined in kernelspec json, & it is a Python kernel, then we'll use the provided python interpreter.
+ */
+export type LocalKernelSpecConnectionMetadata = Readonly<{
```
rchiodo, line 102: I'm confused, this is exported here and in types.ts? Which one do we want to expose to people?

rchiodo:
Looks good, just had some questions about where the KernelConnectionMetadata types are defined (seems like they're in more than one spot)

src/client/api/extension.d.ts
DonJayamanne: Users of this API will need the *.d.ts files (or the types).
This file contains everything in one place, this way they can just get it from here.
Ideally we should push this as an npm package or the like, but we can do that later

src/client/api/apiAccessService.ts
```diff
@@ -0,0 +1,86 @@
...
+    extensionId: string;
+    allowed: 'yes' | 'no';
+}[];
+
+const API_ACCESS_GLOBAL_KEY = 'JUPYTER_API_ACCESS_INFORMATION';
+// Some publishers like our own `ms-tolsai` will always be allowed to access the API.
+const TrustedExtensionPublishers = new Set([JVSC_EXTENSION_ID.split('.')[0]]);
+// We dont want to expose this API to everyone, else we'll keep track of who has access to this.
+// However, the user will still be prompted to grant these extensions access to the kernels..
+const PublishersAllowedWithPrompts = new Set([JVSC_EXTENSION_ID.split('.')[0]]);
```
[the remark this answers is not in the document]
DonJayamanne: lets discuss in standup, when Ian is also back

src/client/api/extension.d.ts
rchiodo: But doesn't that mean the other location isn't necessary? Can't we just use this definition ourselves?

src/client/api/apiAccessService.ts
```diff
@@ -29,11 +29,20 @@ export class ApiAccessService {
...
+        @inject(IApplicationEnvironment) private appEnv: IApplicationEnvironment,
+        @inject(IExtensionContext) private context: IExtensionContext
     ) {}
     public async getAccessInformation(): Promise<{ extensionId: string; accessAllowed: boolean }> {
         const info = await this.extensions.determineExtensionFromCallStack();
         const publisherId = info.extensionId.split('.')[0];
+        if (this.context.extensionMode === ExtensionMode.Test) {
+            if (!TrustedExtensionPublishers.has(publisherId) || PublishersAllowedWithPrompts.has(publisherId)) {
+                void this.appShell.showInformationMessage(
+                    'Thanks for trying the Jupyter API. Please file an issue on our repo to use this API in production. This would prevent us from breaking your extension when updating the API (as it is still a work in progress).'
```
rchiodo, line 41: This needs to be localized.

src/client/api/extension.d.ts
DonJayamanne: Not really, I don't want to expose all of the properties, I don't want to expose everythign in Python interpreter or some other properties, else we end up in advertently having to support all of the stuff we expose.
This way we have full control over what's exposed.

DonJayamanne: E.g. I don't want to expose everything in the Python interpretre API

src/client/api/apiAccessService.ts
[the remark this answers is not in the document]
DonJayamanne: My concern is anyone can edit the package.json, i prefer the stuff hardcoded
Though I do like the package.json approach as well, feels a lot cleaner.
But something about that keeps bothering me, feels like we're just dumping stuff into that file, then again, there's no reason we can't put anything we want to...
As you can tell i have no preference for either one

rchiodo (approved):

PR #8546, opened 2021-12-14
Implement memory test

### LINKED ISSUE #1391: Create a memory test to verify we don't keep using more memory
Given the work done on this PR here:
https://github.com/microsoft/vscode-python/pull/11835

We can potentially track memory inside our tests. 
We could use this to write a test that verifies memory isn't growing over time.

> Rich Chiodo

I had originally added code that would track memory usage and output statistics. I was thinking more of just outputting the statistics on every build and we could manually check the growth. Then you could determine which changes caused memory growth spikes. This is similar to how when we were all on the Visual Studio team (well okay not all of us, but Ian, Jim and I) did RPS tests. RPS tests also flagged an error if the growth was off by some percentage, but I think we can skip that part.

You'd track statistics say for each functional test run, or maybe vscode tests run (as those are more realistic)

rchiodo:
Fixes #1391 

Basic memory test. Checks for memory leaks and size growth. Not sure how stable the size changes are, but going to try on CI to see what happens.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -139,7 +139,7 @@
                 "--extensionTestsPath=${workspaceFolder}/out/test"
             ],
             "env": {
-                "VSC_JUPYTER_CI_TEST_GREP": "Memory Test", // Leave as `VSCode Notebook` to run only Notebook tests.
+                "VSC_JUPYTER_CI_TEST_GREP": "", // Leave as `VSCode Notebook` to run only Notebook tests.
                 "VSC_JUPYTER_CI_TEST_INVERT_GREP": "", // Initialize this to invert the grep (exclude tests with value defined in grep).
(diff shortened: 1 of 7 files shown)


.vscode/launch.json
```diff
@@ -139,7 +139,7 @@
                 "--extensionTestsPath=${workspaceFolder}/out/test"
             ],
             "env": {
-                "VSC_JUPYTER_CI_TEST_GREP": "", // Leave as `VSCode Notebook` to run only Notebook tests.
+                "VSC_JUPYTER_CI_TEST_GREP": "Memory Test", // Leave as `VSCode Notebook` to run only Notebook tests.
```
rchiodo, line 142: Whoops. My bad.

rchiodo:
@DonJayamanne @amunger can you guys take a look?

DonJayamanne (approved):

PR #8556, opened 2021-12-15
Non ZMQ scenarios failing when using conda env

### LINKED ISSUE #8553: Packages are not imported from the selected Python environment
- Create new conda environment
- Install jupyter into it
- set ```"jupyter.disableZMQSupport": "true"```
- Try running interactive window with this environment

Kernel never starts.

rchiodo:
Fixes #8553 

Forcing PYTHONPATH to match the sys.path for an environment works around what seems like jupyter injecting its own python path into the sys.path on kernel launch (which they probably do to get kernels to start in the first place). Jupyter's injection is of course wrong if we start jupyter from a different interpreter than the user is using.

I think this should be fine and might even fix other startup problems as we're ensuring the sys.path is the same as the environment.


DonJayamanne (approved):

DonJayamanne:
I've also added tests for this.

PR #8574, opened 2021-12-17
Simpler Kernel API

DonJayamanne:
Reduces the npm packages required by extension authors.
ws & rxjs packages are no longer required to use the kernel API
Also makes it easier to consume the Socket events (users won't have to bind/unbind/bind again, it will be handled internally when kernels restarts)

### DIFF
--- a/src/client/api/extension.d.ts
+++ b/src/client/api/extension.d.ts
@@ -5,8 +5,11 @@
 import { CancellationToken, Event, NotebookDocument } from 'vscode';
 import type { Kernel } from '@jupyterlab/services/lib/kernel';
 import type { Session } from '@jupyterlab/services';
-import { Observable } from 'rxjs/Observable';
-import type { Data as WebSocketData } from 'ws';
+
+/**
+ * Data represents the message payload received over the WebSocket.
+ */
+export type WebSocketData = string | Buffer | ArrayBuffer | Buffer[];
 
 export interface JupyterAPI {
     /**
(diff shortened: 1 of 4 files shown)


src/client/api/kernelApi.ts
```diff
@@ -184,3 +194,72 @@ class JupyterKernelService implements IExportedKernelService {
...
     }
 }
+
+/**
+ * Helper class to wrap the IKernelSocket.
+ * This way users of the API will not need to unbind/rebind the hooks and
+ * listen to changes in the observable.
+ * Also prevents the need for users of the API to depend on rxjs.
+ */
+class KernelSocketWrapper implements IKernelSocket {
```
DonJayamanne, line 204: See PR comments

DonJayamanne: Also class comments makes it easier to understand the purpose

src/client/api/extension.d.ts
```diff
@@ -5,8 +5,11 @@
 import { CancellationToken, Event, NotebookDocument } from 'vscode';
 import type { Kernel } from '@jupyterlab/services/lib/kernel';
 import type { Session } from '@jupyterlab/services';
-import { Observable } from 'rxjs/Observable';
-import type { Data as WebSocketData } from 'ws';
```
DonJayamanne, line 9: Users of this API will not need ws & rxjs anymore.

rchiodo (approved):

rchiodo:
Perf test failing can be ignored. You just need to merge from main.

Issue #7603, opened 2021-09-21
Data Viewer fails to load dataframes containing columns named the same.


greazer:
## Environment data

-   VS Code version: 1.61-insiders
-   Jupyter Extension version (available under the Extensions sidebar): v2021.9.1001256920

## Expected behaviour

Be able to view a simple dataframe

## Actual behaviour

Can't view a simple dataframe

## Steps to reproduce:

1. Create a notebook or IW file
2. Run this code (StateAbbrev.csv is attached)
```python
import pandas as pd

df = pd.read_csv('StateAbbrev.csv')
df.index = df.abbrev
```
3. Try to view the df dataframe in the dataviewer. 

It never loads

The log indicates this:
```ValueError: Overlapping names between the index and columns
    at f.extractJupyterResultText (c:\Users\jimg\.vscode-insiders\extensions\ms-toolsai.jupyter-2021.9.1001256920\out\client\extension.js:90:260708)
    at f.deserializeJupyterResult (c:\Users\jimg\.vscode-insiders\extensions\ms-toolsai.jupyter-2021.9.1001256920\out\client\extension.js:90:260826)
    at f.getDataFrameRows (c:\Users\jimg\.vscode-insiders\extensions\ms-toolsai.jupyter-2021.9.1001256920\out\client\extension.js:90:257939)
    at r.getRows (c:\Users\jimg\.vscode-insiders\extensions\ms-toolsai.jupyter-2021.9.1001256920\out\client\extension.js:52:866224)
    at c:\Users\jimg\.vscode-insiders\extensions\ms-toolsai.jupyter-2021.9.1001256920\out\client\extension.js:52:859192
    at C.wrapRequest (c:\Users\jimg\.vscode-insiders\extensions\ms-toolsai.jupyter-2021.9.1001256920\out\client\extension.js:52:859421)
```

While it may not be common for this to happen when reading a csv file and immediately viewing it, I do think it could be common when doing dataprep.

</details
[StateAbbrev.csv](https://github.com/microsoft/vscode-jupyter/files/7205941/StateAbbrev.csv)
>

rebornix:
We are probably seeing 5000 similar failures on around 500 machines.

rebornix:
The log we saw is from CI tests though (as @DonJayamanne pointed out) but it might share the same root cause.
```
Error: Failed to fetch variable info from the Jupyter server. 
[...]

PR #8597, opened 2021-12-21
Support viewing dataframe with columns used as index

DonJayamanne:
For #7603

### DIFF
--- a/pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
+++ b/pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
@@ -153,7 +153,7 @@ def _VSCODE_getDataFrameRows(df, start, end):
         )
     except:
         pass
-    return _VSCODE_pd_json.to_json(None, df, orient="table", date_format="iso")
+    return _VSCODE_pd_json.to_json(None, df, orient="split", date_format="iso")
 
 
 # Function to get info on the passed in data frame
--- a/src/client/datascience/data-viewing/types.ts
+++ b/src/client/datascience/data-viewing/types.ts
@@ -97,7 +97,7 @@ export enum ColumnType {
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type IRowsResponse = any[];
+export type IRowsResponse = Record<string, unknown>[];
 
 export const IDataViewerFactory = Symbol('IDataViewerFactory');
 export interface IDataViewerFactory {
--- a/src/client/datascience/data-viewing/types.ts
+++ b/src/client/datascience/data-viewing/types.ts
@@ -96,7 +96,6 @@ export enum ColumnType {
     Bool = 'bool'
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type IRowsResponse = Record<string, unknown>[];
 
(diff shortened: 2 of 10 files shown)


pythonFiles/vscode_datascience_helpers/dataframes/vscodeDataFrame.py
```diff
@@ -153,7 +153,7 @@ def _VSCODE_getDataFrameRows(df, start, end):
         )
     except:
         pass
-    return _VSCODE_pd_json.to_json(None, df, orient="table", date_format="iso")
+    return _VSCODE_pd_json.to_json(None, df, orient="split", date_format="iso")
```
DonJayamanne, line 156: `split` allows us to send the data in a JSON format, `table` crashes

DonJayamanne: Note: The schema is already sent via another method,

src/client/datascience/data-viewing/types.ts
```diff
@@ -96,8 +96,7 @@ export enum ColumnType {
     Bool = 'bool'
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type IRowsResponse = any[];
+export type IRowsResponse = Record<string, unknown>[];
```
DonJayamanne, line 99: In our code, we know exactly what we're expecting (or what its supposed to look like), hence provided better typing

DonJayamanne: In the WebView we're expecting an array of objects

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -182,7 +183,7 @@ export class DebuggerVariables extends DebugLocationTracker
         end: number,
         kernel?: IKernel,
         sliceExpression?: string
-    ): Promise<{}> {
+    ): Promise<{ data: Record<string, unknown>[] }> {
```
DonJayamanne, line 186: In our code, we know exactly what we're expecting (or what its supposed to look like), hence provided better typing
Else its difficult to tell what's going where

src/client/datascience/jupyter/debuggerVariables.ts
```diff
@@ -191,7 +192,7 @@ export class DebuggerVariables extends DebugLocationTracker
         // Run the get dataframe rows script
         if (!this.debugService.activeDebugSession || targetVariable.columns === undefined) {
             // No active server just return no rows
-            return {};
+            return { data: [] };
```
DonJayamanne, line 195: No its cleaner, we don't need to test the types in the code, as ts compiler will tell us exactly what we'll get.

src/client/datascience/jupyter/pythonVariableRequester.ts
```diff
@@ -14,6 +13,28 @@ import { JupyterDataRateLimitError } from '../errors/jupyterDataRateLimitError';
...
+        const rowData: Record<string, unknown> = {
+            index: rowIndexValues[index]
+        };
+        columns.forEach((column, columnIndex) => {
+            rowData[column] = row[columnIndex];
+        });
+        return rowData;
+    });
+    return { data };
+}
```
DonJayamanne, line 36: This is the magic, this is where we convert the `split` (series) data format into a table (the same structure we used to get when we called `to_json with table`

Issue #5250, opened 2021-03-23
Migrate IPyWidget tests to the way Ian wrote tests for WebView Variables (IPYWidget tests for Notebooks)


DonJayamanne:
Get rid of puppeteer and other stuff.
Use a single way, and test real bits with real editor.

This will make the tests easier to maintain going forward.
Right now, the tests are flaky as well.

joyceerhl:
Can we / do we want to be able to simulate interaction with the DOM? My understanding was that the puppeteer test infra lets us do that, but maybe that's not a problem.

DonJayamanne:
We can send messages to the WebView, and let that interact with the DOM.

DonJayamanne:
Closing this as done

PR #8611, opened 2021-12-23
Test for Slider IPyWidget

DonJayamanne:
Part of #5250

**Architecture: (borrow from how @IanMatthewHuff enabled UI tests for the other webviews all inside VS Code)**
* Include a new controller script (js script loaded when ipynb is opened)
* When rendering a widget, we will setup hooks exposed in the above script
* The tests will set comms with the renderer (via the new custom script)
* After we run a cell, we use the comms to send a message to renderer and check if an HTML element has been rendered by quering the DOM
* Return value is sent from renderer using the comms (Notebook Messaging API).

This is still a first attempt, not sure I'll get this working, had a few issues.
**Contains just one test to first get the framework setup.**

Blocked by this issue https://github.com/microsoft/vscode-jupyter/issues/8617
Lot of the code was added for debugging (figure out why things weren't working) then came across #8617


package.json
```diff
@@ -1901,7 +1901,8 @@
                 "displayName": "Jupyter IPyWidget Renderer",
                 "mimeTypes": [
                     "application/vnd.jupyter.widget-view+json"
-                ]
+                ],
+                "requiresMessaging": "optional"
```
DonJayamanne, line 1905: Optional, as we only need comms for tests, in production its not required.

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -321,12 +322,23 @@ export class VSCodeNotebookController implements Disposable {
...
-            join(this.context.extensionPath, 'out', 'datascience-ui', 'ipywidgetsKernel', 'ipywidgetsKernel.js'),
-            join(this.context.extensionPath, 'out', 'fontAwesome', 'fontAwesomeLoader.js')
-        ].map((uri) => new NotebookRendererScript(Uri.file(uri)));
+        const scripts: string[] = [];
+        // Only used in tests & while debugging.
+        if (
+            this.context.extensionMode === ExtensionMode.Development ||
+            this.context.extensionMode === ExtensionMode.Test
+        ) {
+            scripts.push(join(this.context.extensionPath, 'out', 'datascience-ui', 'widgetTester', 'widgetTester.js'));
```
DonJayamanne, line 331: Register the script used to query the DOM in tests, the script will check if widgets have been renderered or not & perform some basic operations.
Only included in dev & test mode.

src/datascience-ui/ipywidgets/renderer/index.ts
```diff
@@ -27,3 +32,20 @@ export const activate: ActivationFunction = (_context) => {
         }
     };
 };
+
+function hookupTestScripts(context: RendererContext<unknown>) {
```
DonJayamanne, line 36: If the global entry point exists, then we setup the hooks, else we don't care.
This global object will be available only in tests

rchiodo (approved):

build/webpack/webpack.datascience-ui.config.builder.js
```diff
@@ -139,23 +146,39 @@ function buildConfiguration(bundle) {
         );
     }
     let outputProps =
-        bundle !== 'ipywidgetsRenderer' && bundle !== 'errorRenderer'
+        bundle !== 'ipywidgetsRenderer' && bundle !== 'errorRenderer' && bundle !== 'ipywidgetsKernel'
             ? {}
             : {
-                  library: 'LIB',
+                  library: `LIB${bundle.toUpperCase()}`,
```
DonJayamanne, line 152: Ensure all JS packages have unique global vars.

src/datascience-ui/ipywidgets/kernel/index.ts
```diff
@@ -218,14 +222,18 @@ function convertVSCodeOutputToExecuteResultOrDisplayData(
...
         logMessage('IPyWidget kernel initializing...');
-        initialize();
+        initialize(capturedContext);
     } else {
         setTimeout(attemptInitialize, 100);
     }
 }
-attemptInitialize();
+
+export const activate = attemptInitialize;
```
DonJayamanne, line 239: This is new, the way VS Code works is weird.
If the text `export ... activate` exists anywhere in the JS code, then VS Code will load the JS file as a ES module & call the activate (at that point the class `acquireVSCodeAPI` doesn't exist in the global context).

Note: We weren't expoting the active method, but somewhere in the code that text ends up in the JS file and VS Code tries to load this JS as an ES module and fails.
Solution: Ensure we always setup this script file to be loaded as an es module.
This also changes how the messaging api is handled (that's now an argument in the activate method).

DonJayamanne: Will discuss in standup

rchiodo (approved):

PR #8639, opened 2022-01-04
ensure that an IW has an associated kernel when retrieved

amunger:
For #

-   [ x ] Pull request represents a single change (i.e. not fixing disparate/unrelated things in a single PR).
-   [ x ] Title summarizes what is changing.
-   [ x ] Has a [news entry](https://github.com/Microsoft/vscode-jupyter/tree/main/news) file (remember to thank yourself!).
-   [ x ] Appropriate comments and documentation strings in the code.

### DIFF
--- a/src/client/datascience/interactive-window/interactiveWindow.ts
+++ b/src/client/datascience/interactive-window/interactiveWindow.ts
@@ -251,6 +251,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
                 this._controllerReadyPromise.resolve(e.controller);
 
                 // Recreate the kernel ready promise now that we have a new controller
+                this._kernelReadyPromise = undefined;
                 this.ensureKernelReadyPromise();
             },
             this,
--- a/src/client/datascience/interactive-window/interactiveWindow.ts
+++ b/src/client/datascience/interactive-window/interactiveWindow.ts
@@ -177,7 +177,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
     private ensureKernelReadyPromise() {
         if (!this._kernelReadyPromise) {
             this._kernelReadyPromise = this.createKernelReadyPromise();
-            this._kernelReadyPromise.catch(() => this._kernelReadyPromise = undefined);
+            this._kernelReadyPromise.catch(() => (this._kernelReadyPromise = undefined));
         }
     }
 
--- a/src/client/datascience/interactive-window/interactiveWindow.ts
+++ b/src/client/datascience/interactive-window/interactiveWindow.ts
@@ -256,7 +256,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
                 this._controllerReadyPromise.resolve(e.controller);
 
                 // Recreate the kernel ready promise now that we have a new controller
-                this._kernelReadyPromise = this.createKernelReadyPromise();
+                void this.ensureKernelReady();
             },
             this,
             this.internalDisposables
--- a/src/client/datascience/interactive-window/interactiveWindow.ts
+++ b/src/client/datascience/interactive-window/interactiveWindow.ts
@@ -167,23 +167,26 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             this.internalDisposables
         );
         this.internalDisposables.push(kernel);
-        await kernel.start();
+
+        try{
+            await kernel.start();
+        }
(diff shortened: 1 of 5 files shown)


src/client/datascience/interactive-window/interactiveWindowProvider.ts
```diff
@@ -95,6 +95,9 @@ export class InteractiveWindowProvider implements IInteractiveWindowProvider, IA
             result = this.create(resource, mode);
         }
 
+        // make sure the existing IW has a kernel in case it failed to start previously
+        result.ensureKernelReadyPromise();
+
         await result.readyPromise;
```
rchiodo, line 101: Could you put the 'ensureKernelReadyPromise' call into the getter for 'readyPromise'? Then this code wouldn't have to change.

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -167,12 +167,28 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             this.internalDisposables
         );
         this.internalDisposables.push(kernel);
-        await kernel.start();
+
```
rchiodo, line 170: This function is rather confusing. Internal state is assigned to its result but it also updates that internal state in itself. 

What if you did this instead (and not assign the _kernelReadyPromise as the result of this function)

```
this._kernelReadyPromise = () => {
    try {
        await kernel.start();
        this.fileInKernel = undefined;
        await this.runIntialization(kernel, this.owner);
    } catch {
        // Do nothing
    }
};
```

rchiodo: The exception could be thrown too if we want to make sure it comes out. Maybe like this instead:

```
let exception: Error | undefined;
const readyPromise = () => {
    try {
        await kernel.start();
        this.fileInKernel = undefined;
        await this.runIntialization(kernel, this.owner);
    } catch (e) {
        throw e;
        this._kernelReadyPromise = undefined
    }
};
this._kernelReadyPromise = readyPromise;
await readyPromise;
```

rchiodo: Or maybe a simpler way, turn your 'ensureKernelReadyPromise' into a wrapper around 'createKernelReadyPromise' and have it behave like so:
```
public ensureKernelReadyPromise() {
    if (!this._kernelReadyPromise) {
        this._kernelReadyPromise = this.createKernelReadyPromise().catch(() => this._kernelReadyPromise = undefined);
    }
}

Comment (rchiodo) [reply src/client/datascience/interactive-window/interactiveWindow.ts:170]: And then replace all calls to 'createKernelReadyPromise' with the 'ensureKernelReadyPromise'. The error will still be thrown if somebody waits on the _kernelReadyPromise, but it won't prevent the next call from recreating it.

File: src/client/datascience/types.ts [hunk 3]
```diff
@@ -433,6 +433,7 @@ export interface IInteractiveWindow extends IInteractiveBase {
     readonly readyPromise: Promise<void>;
     readonly kernelPromise: Promise<IKernel | undefined>;
     closed: Event<void>;
+    ensureKernelReadyPromise(): Promise<IKernel>;
```
rchiodo, line 436: I don't think this is necessary as you can do this in the 'readyPromise' getter.

rchiodo (changes requested):
What do you think of my suggestion?

src/client/datascience/interactive-window/interactiveWindowProvider.ts
amunger: that makes sense. I didn't realize readyPromise was only called from the provider, that makes it easy to test

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -243,7 +251,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
                 this._controllerReadyPromise.resolve(e.controller);
 
                 // Recreate the kernel ready promise now that we have a new controller
-                this._kernelReadyPromise = this.createKernelReadyPromise();
+                this.ensureKernelReadyPromise();
             },
```
rchiodo, line 255: This won't 'recreate' it if it's already set. You probably need to clear it here first.

rchiodo (approved):

IanMatthewHuff (approved):

PR #8642, opened 2022-01-05
Allow queueing of multiple cells in IW

### LINKED ISSUE #8022: When executing multiple cells queued ones don't show up anymore
Issue Type: <b>Bug</b>

When executing multiple cells queued ones don't show up anymore; it only shows the most recent one being currently executed and not the ones that are queued for execution in Interactive Python.

VS Code version: Code 1.59.0 (379476f0e13988d90fab105c5c19e7abc8b1dea8, 2021-08-04T23:14:40.191Z)
OS version: Darwin x64 20.6.0
Restricted Mode: No

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz (4 x 2300)|
|GPU Status|2d_canvas: enabled<br>gpu_compositing: enabled<br>metal: disabled_off<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|3, 4, 4|
|Memory (System)|8.00GB (0.15GB free)|
|Process Argv|--crash-reporter-id d07ed298-647e-451d-99de-de1722f8bc11|
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (33)</summary>

Extension|Author (truncated)|Version
---|---|---
codesnap|adp|1.3.4
Bookmarks|ale|13.1.0
python-inline-query-highlighting|ard|1.0.2
path-intellisense|chr|2.4.0
vscode-sql-beautify|cla|0.2.5
gitignore|cod|0.7.0
bracket-pair-colorizer-2|Coe|0.2.1
githistory|don|0.6.18
gitlens|eam|11.6.0
autoconf|mae|0.1.0
markdown-converter|man|3.1.2
git-graph|mhu|1.30.0
python|ms-|2021.8.1102490946
vscode-pylance|ms-|2021.8.0
jupyter|ms-|2021.8.1195043623
remote-containers|ms-|0.191.0
remote-ssh|ms-|0.65.7
remote-ssh-edit|ms-|0.65.7
remote-wsl|ms-|0.58.2
vscode-remote-extensionpack|ms-|0.21.0
azure-account|ms-|0.9.8
cmake-tools|ms-|1.7.3
autodocstring|njp|0.5.4
material-icon-theme|PKi|4.9.0
kuskus-kusto-language-server|ros|1.0.1
kuskus-kusto-syntax-highlighting|ros|1.1.8
markdown-preview-enhanced|shd|0.5.22
rewrap|stk|1.14.0
code-spell-checker|str|1.10.2
cmake|twx|0.0.17
vscodeintellicode|Vis|1.2.14
clang-format|xav|1.9.0
vscode-scope|yiw|0.0.24

(1 theme extensions excluded)

</details><details>
<summary>A/B Experiments</summary>

```
vsliv368cf:30146710
vsreu685:30147344
python383cf:30185419
pythonvspyt602:30300191
vspor879:30202332
vspor708:30202333
vspor363:30204092
pythonvspyt639:30300192
pythontb:30283811
pythonptprofiler:30281270
vshan820:30294714
vstes263:30335439
vscoreces:30322571
pythondataviewer:30285071
pythonvsuse255:30340121
vscod805:30301674
pythonvspyt200:30340761
vscextlangct:30333562
binariesv615:30325510
vsccppwt:30329788
pythonvssor306:30344512
bridge0708:30335490
vstre464:30346472

```

</details>

### LINKED ISSUE #8615: Cannot run lines in multiple interactive windows at the same time
## Environment data

-   VS Code version: 1.63.2
-   Jupyter Extension version: 2021.11.1001550889
-   Python Extension version: 2021.12.1559732655
-   OS and version: Windows 10 2004
-   Python and/or Anaconda version: 3.9
-   Type of virtual environment used: N/A

## Expected behaviour

Run two time-consuming codes in multiple interactive windows simultaneously, like in the 2021.10 version Jupyter.

## Actual behaviour

The first window successfully executes the code, while the second window does not execute until the first window finishes.

## Steps to reproduce:

Open an interactive window, and use "Run Selection/Line in Interactive Window" to send the following code.
```python
while True:
   pass
```
image

Open a second interactive window, and use "Run Selection/Line in Interactive Window" to send the same code to the second window. The second window cannot execute the code unless we turn off the first window.
image

rchiodo:
Fixes #8022
Fixes #8615

### DIFF
--- a/src/client/common/utils/decorators.ts
+++ b/src/client/common/utils/decorators.ts
@@ -266,3 +266,36 @@ export function testOnlyMethod() {
         return descriptor;
     };
 }
+
+// Mark a method that returns a promise to chain it
+export function chainable() {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return function (_target: Object, propertyKey: string, descriptor: TypedPropertyDescriptor<PromiseFunction>) {
+        const originalMethod = descriptor.value!;
+        const chainedKey = `chainedPromiseFor_${propertyKey}`;
+
+        // eslint-disable-next-line , @typescript-eslint/no-explicit-any
+        descriptor.value = async function (...args: any[]) {
+            // Check for promise in the object.
+            let currentValue = (this as any)[chainedKey] as Promise<any>;
+            if (currentValue) {
+                currentValue = currentValue.then(() => originalMethod.apply(this, args));
+            } else {
+                currentValue = originalMethod.apply(this, args);
+            }
+
+            // Save promise in object
+            (this as any)[chainedKey] = currentValue;
+
+            // If promise fails, clear it.
+            return currentValue
+                .then((r) => r)
+                .catch((e) => {
+                    (this as any)[chainedKey] = undefined;
+                    throw e;
+                });
+        };
+
+        return descriptor;
+    };
+}
--- a/src/client/datascience/editor-integration/codewatcher.ts
+++ b/src/client/datascience/editor-integration/codewatcher.ts
@@ -21,6 +21,7 @@ import { IDocumentManager } from '../../common/application/types';
 import { IFileSystem } from '../../common/platform/types';
 
 import { IConfigurationService, IDisposable, IJupyterSettings, Resource } from '../../common/types';
+import { chainable } from '../../common/utils/decorators';
 import * as localize from '../../common/utils/localize';
 import { isUri } from '../../common/utils/misc';
 import { StopWatch } from '../../common/utils/stopWatch';
@@ -33,6 +34,7 @@ import {
     ICodeLensFactory,
     ICodeWatcher,
     IDataScienceErrorHandler,
+    IInteractiveWindow,
     IInteractiveWindowProvider
 } from '../types';
 
@@ -69,7 +71,6 @@ export class CodeWatcher implements ICodeWatcher {
     private codeLensUpdatedEvent: EventEmitter<void> = new EventEmitter<void>();
     private updateRequiredDisposable: IDisposable | undefined;
     private closeDocumentDisposable: IDisposable | undefined;
-    private addCodePromise: Promise<boolean> = Promise.resolve(true);
 
     constructor(
         @inject(IInteractiveWindowProvider) private interactiveWindowProvider: IInteractiveWindowProvider,
@@ -137,6 +138,7 @@ export class CodeWatcher implements ICodeWatcher {
     }
     @captureTelemetry(Telemetry.RunAllCells)
     public async runAllCells() {
+        const iw = await this.getActiveInteractiveWindow();
         const runCellCommands = this.codeLenses.filter(
             (c) =>
                 c.command &&
@@ -145,7 +147,6 @@ export class CodeWatcher implements ICodeWatcher {
                 c.command.arguments.length >= 5
         );
         let leftCount = runCellCommands.length;
-
         // Run all of our code lenses, they should always be ordered in the file so we can just
         // run them one by one
         let finished = Promise.resolve(true);
@@ -168,7 +169,7 @@ export class CodeWatcher implements ICodeWatcher {
 
                 // Note: We do a get or create active before all addCode commands to make sure that we either have a history up already
                 // or if we do not we need to start it up as these commands are all expected to start a new history if needed
-                finished = this.addCode(code, this.document.uri, range.start.line, leftCount);
+                finished = this.addCode(iw, code, this.document.uri, range.start.line, leftCount);
             }
         }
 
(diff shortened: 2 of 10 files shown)


src/client/datascience/editor-integration/codewatcher.ts
```diff
@@ -137,6 +138,7 @@ export class CodeWatcher implements ICodeWatcher {
     }
     @captureTelemetry(Telemetry.RunAllCells)
     public async runAllCells() {
+        const iw = await this.getActiveInteractiveWindow();
```
rchiodo, line 141: Moved the IW fetch to outside of addCode so that multiple calls to addcode wouldn't block on it.

src/client/datascience/editor-integration/codewatcher.ts
```diff
@@ -977,34 +990,14 @@ export class CodeWatcher implements ICodeWatcher {
...
 
-    private async addCode(
-        code: string,
-        file: Uri,
-        line: number,
-        leftCount: number,
-        editor?: TextEditor,
-        debug?: boolean
-    ): Promise<boolean> {
-        this.addCodePromise = this.addCodePromise.then((_previousResult) => {
```
rchiodo, line 988: The promise chain wasn't necessary here anymore because the IW handles queueing itself now.

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -587,7 +593,19 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
                 [notebookCellData]
             );
         });
-        return notebookDocument.cellAt(notebookDocument.cellCount - 1);
+        const cell = notebookDocument.cellAt(notebookDocument.cellCount - 1);
+
+        // All cells added are immediately executed. Indicate that now
```
rchiodo, line 598: This comment is incorrect. I wanted to mark cells as 'busy' but the Notebook API doesn't make that  easy. I would actually have to make them part of a CellExecution, but since we queue those up, I can't.

src/client/datascience/editor-integration/codewatcher.ts
```diff
@@ -1029,7 +1021,7 @@ export class CodeWatcher implements ICodeWatcher {
         if (!result) {
             // If our cell result was a failure (but not an exception) show an error
             // for the count of cells cancelled
-            await this.addErrorMessage(file, leftCount);
+            await this.addErrorMessage(interactiveWindow, leftCount);
 
             // Throw to break out of the promise chain
             throw new InteractiveCellResultError();
```
rchiodo, line 1027: This isn't actually working. Need to move it into the IW for it to break the chain. More updates coming ...

src/client/datascience/editor-integration/codewatcher.ts
IanMatthewHuff: Per this comment why is chainable needed here then?

src/client/common/utils/decorators.ts
```diff
@@ -266,3 +266,36 @@ export function testOnlyMethod() {
         return descriptor;
     };
 }
+
+// Mark a method that returns a promise to chain it
+export function chainable() {
```
IanMatthewHuff, line 271: I know that you had looked at this before, so I don't want to rehash a debate on this. But I'm not keen on something that changes the way the function works this much being "hidden" in a decorator. I like decorators for stuff like logging or error handling. Something like this feels like a big functionality difference which I wouldn't want changed by a decorator. They can also be a pain to debug at times.

I guess if this PR looks good and the chaining cases are all handed then I won't hold things up. But if we run into issue in the future with this decorator hiding real issues or making debugging hard I'd argue against it.

rchiodo: Debugging it isn't hard, you just set breakpoints in the decorator. The reason for it is so we don't have duplicate code for chaining a promise. Not sure how else to implement common code for promise chaining though. A wrapper method instead?

Something like:
```
public doStuffWithPromise() {
   return chained(async () => {
       await doingstuff();
   })
}
```

Does that grok better? It's essentially the same thing as the decorator but more explicit.

IanMatthewHuff: Probably just anti decorator bias on my part. I read decorators like this:

```
@thisStuffDoesn'tReallyMatter
@youCanReadTheFunctionJustFineWithoutThis
@probablyJustLoggingOrSomethingLikeThat
function doStuff() {}
```
This feels more like:
```
@thisIsReallyKindaABigChangeToHowThisFunctionWorks
function doStuff() {}
```

But there is no reason that a decorator can't do something like you are doing here. So I vote we go for it. I was thinking it might cause issues with the case where there is a python error in the chain (where we return a false and show a message) but it looks like you added that in a later commit.

IanMatthewHuff (approved):

PR #8649, opened 2022-01-06
Actually have the preRelease option

rchiodo:
publish

More work for https://github.com/microsoft/vscode/issues/15756


rchiodo:
This is what it looks like now:

image

This is on stable. On insiders it will auto update to the prerelease version.

rchiodo:
The other benefit is that it looks like this version won't end up in the marketplace. It only shows the non prerelease packages:

image

Issue #8322, opened 2021-11-22
%matplotlib QT5 doesn't work 


gonghenghai:
## Environment data

-   VS Code version: 1.62.3
-   Jupyter Extension version (available under the Extensions sidebar): after v2021.10.1001409655
-   Python Extension version (available under the Extensions sidebar): v2021.11.1422169775
-   OS (Windows | Mac | Linux distro) and version: windows
-   Python and/or Anaconda version: 3.9.5
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): venv
-   Jupyter server running: Local

## Expected behaviour

Execute the cells

## Actual behaviour

Just waiting

![Screenshot (109)_LI](https://user-images.githubusercontent.com/21306413/142830813-1ab0d443-3969-470c-a9fd-324e9adee58b.jpg)

DonJayamanne:
Thanks for filing this issue, & I"m sorry yo'ure running into this.
We've been working on some perf related fixes and the like in our current development branch.

* Do you experience similar perf issues if you do not run the first cell `%matplotlib`
* I.e. can you confirm that when using notebooks without that cell magic then cells run fast.

**Please could you**
* Install VS Code insiders from here https://code.visualstudio.com/insiders/
* Install the Jupyter extension (it will automatically install the latest development version, starts with `2021.11.`
* Enable verbose logging
	* Go into your settings file (use the command 'Preferences: Open Settings UI`)
	* Search for `Jupyter logging` 	
	* Change the logging level for Jupyter extension to `verbose`

See here 
![image](https://user-images.githubusercontent.com/1948812/142954133-5c8abe63-f5a7-410f-8e55-3b729cb622df.png)


	* Reload VS Code
* Replicate the above problem
* Go into your `Jupyter` output panel (use the command `Jupyter: View Output`)
	* Copy everything from there, and please paste the contents into this issue

faultdiagnosistoolbox:
I will try to contribute with the information, I asked the original question at https://stackoverflow.com/questions/70056478/vscode-python-interactive-window-freezes-when-using-matplotlib

First I can confirm that when not running %matplotlib (or even running %matplotlib inline) all works as expected. It is only when I have the external window that I see the problem.

I've installed the code-insider edition and the python (including the jupyter) extension version Jupyter v2021.11.1001498571.

Then I
1. Activated verbose jupyter logging
2. Restarted vscode insider
3. Created jupyter interactive window
4. %matplotlib
[...]

PR #8650, opened 2022-01-07
Fix QT5 to work as a backend for matplotlib

rchiodo:
Using ```%matplotlib <backend>``` is actually handled by IPython itself. What this does is changes the event loop in use to handle messages.

The QT event loop doesn't seem to queue messages so if we send more than one requestExecute at a time, it will hang.

Ideally we'd modify IJupyterSession to force all requests to go through a queue, but it doesn't return a promise. We'd have to implement our own custom future that enforced waiting for the previous future before making a request. 

This change seems to have fixed it for me and is much simpler. Going to try this first.

For #8322


src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -325,7 +325,14 @@ export class VSCodeNotebookController implements Disposable {
             join(this.context.extensionPath, 'out', 'datascience-ui', 'ipywidgetsKernel', 'require.js'),
             join(this.context.extensionPath, 'out', 'ipywidgets', 'dist', 'ipywidgets.js'),
             join(this.context.extensionPath, 'out', 'datascience-ui', 'ipywidgetsKernel', 'ipywidgetsKernel.js'),
-            join(this.context.extensionPath, 'out', 'fontAwesome', 'fontAwesomeLoader.js')
+            join(this.context.extensionPath, 'out', 'fontAwesome', 'fontAwesomeLoader.js'),
+            join(
```
IanMatthewHuff, line 329: What's this script addition here?

rchiodo: Oh crap. Forgot about this. I was trying to debug something before. Thanks. I'll remove it.

src/client/datascience/ipywidgets/ipyWidgetMessageDispatcher.ts
```diff
@@ -104,8 +104,8 @@ export class IPyWidgetMessageDispatcher implements IIPyWidgetMessageDispatcher {
     }
 
     public receiveMessage(message: IPyWidgetMessage): void {
-        if (process.env.VSC_JUPYTER_LOG_IPYWIDGETS && message.message.includes('IPyWidgets_')) {
```
rchiodo, line 107: I should remove this one too.

src/client/datascience/notebook/vscodeNotebookController.ts
IanMatthewHuff: Yeah, looked like a debug something. Rest looks good.

IanMatthewHuff (approved):

PR #8651, opened 2022-01-07
More robust solution to execute queueing

rchiodo:
For #8322

This implements the 'delayed' future so that it's impossible to send more than one requestExecute at a time.

Tested locally and it works but doing this separately so that it can be reverted if it causes something to break.


rchiodo:
A restart test seems to be failing on all versions. Not sure why. I'm investigating now (hopefully it repros locally)

src/client/datascience/baseJupyterSession.ts
```diff
@@ -217,7 +219,8 @@ export abstract class BaseJupyterSession implements IJupyterSession {
         if (!this.session?.kernel) {
             throw new SessionDisposedError();
         }
-        return this.session.kernel.requestExecute(content, disposeOnDone, metadata);
+        // Make sure to chain the executes
+        return this.chainingExecute.requestExecute(this.session.kernel, content, disposeOnDone, metadata);
```
IanMatthewHuff, line 223: Question (not sure myself). If this QT event loop can't handle the queuing would we also need this chain on the other request functions here? Like requestDebug and requestInspect?

src/client/datascience/delayedFutureExecute.ts
```diff
@@ -0,0 +1,177 @@
...
+    private statusChangedHandler: (_session: Kernel.IKernelConnection, status: KernelMessage.Status) => void;
+    constructor(
+        private kernelConnection: Kernel.IKernelConnection,
+        previousLink: Kernel.IShellFuture<KernelMessage.IExecuteRequestMsg, KernelMessage.IExecuteReplyMsg>,
+        private content: KernelMessage.IExecuteRequestMsg['content'],
+        private disposeOnDone?: boolean,
+        private metadata?: JSONObject
+    ) {
+        // Setup our request based on the previous link finishing
+        if (previousLink) {
```
IanMatthewHuff, line 42: This is minor, but was wondering on the typing here a tiny bit. Aren't we guaranteed a previousLink at this point?

src/client/datascience/delayedFutureExecute.ts
```diff
@@ -0,0 +1,177 @@
...
+            previousLink.done.then(() => this.requestExecute()).catch((e) => this.doneDeferred.reject(e));
+        }
+
+        // If the kernel dies, finish our future
+        this.statusChangedHandler = (_session: Kernel.IKernelConnection, status: KernelMessage.Status) => {
+            if (status === 'unknown' || status === 'restarting' || status === 'dead') {
+                this.doneDeferred.reject(new CancellationError());
```
IanMatthewHuff, line 49: Is this a cancellation error? Also if we are looking at 'restarting' we probably need 'autorestarting' as well since Jupyter has both.

src/client/datascience/delayedFutureExecute.ts
```diff
@@ -0,0 +1,177 @@
...
+
+        // Run the handler now to check
+        this.statusChangedHandler(kernelConnection, kernelConnection.status);
+    }
+    public get msg(): KernelMessage.IExecuteRequestMsg {
+        if (this.requestFuture) {
+            return this.requestFuture.msg;
+        }
+        return {
+            content: this.content,
```
IanMatthewHuff, line 62: Just to check my understanding, this path (where the requestFuture is not define) is basically just a dummy placeholder until the previous execute finished right?

IanMatthewHuff (approved):
Had a couple of questions (just 'cause these futures can get a bit hairy) but I think that it looks good.

src/client/datascience/delayedFutureExecute.ts
rchiodo: Yes that's correct. I don't think any of our code ever reads 'msg' anyway.

src/client/datascience/delayedFutureExecute.ts
rchiodo: Yeah I should add 'autorestarting'. 

CancellationError is what Jupyter throws on the 'done' promise when the kernel is shutdown

src/client/datascience/delayedFutureExecute.ts
rchiodo: Oh yeah that's true. I can skip the if check.

src/client/datascience/baseJupyterSession.ts
rchiodo: Not sure. It might, but we don't send requestDebug/requestInspect in multiple spots. And requestInspect at the same time as requestExecute does work with QT5. At least I couldn't break it variables open.

src/client/datascience/delayedFutureExecute.ts
IanMatthewHuff: Sounds good, the values for it just didn't quite make sense, so just wanted to confirm that it was a dummy value (essentially).

Issue #7970, opened 2021-10-19
Variable Explorer: Some variables are stuck in the loading state


ronglums:
VS Code Version: 1.62.0-insider (user setup)
Commit: 34eeebdd66033633ac22efd5fab928ae37916168
Date: 2021-10-15T05:16:57.817Z
Electron: 13.5.1
Chrome: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Windows_NT x64 10.0.19043
Jupyter extension: v2021.10.1001358390

Steps to repro:
1. open this notebook file https://github.com/claudiaregio/data-science/blob/main/Titanic.ipynb
2. Run all cells. When completed, open the Variable Explorer.
3. Scroll to the top of the Variable explorer.
4. Jump to the "Neural Network" section in the file, "Run all above cells" on the first cell in the section.
5. When run finishes, scroll to the bottom of Variable Explorer

Expected: all variables are shown
![image](https://user-images.githubusercontent.com/26419852/137988819-5f6a01dc-2b50-49c8-912a-bf93d93d2467.png)

Actual: the bottom variables are stuck in the loading state
![image](https://user-images.githubusercontent.com/26419852/137988505-052d5236-6c26-4fcd-a19a-27d304bafdc1.png)

amunger:
I've been able to repro something similar just by creating a few hundred variables and expanding the variable explorer and the ones stuck in loading are actually on the initial page of the explorer. It's pretty tough to get it to repro like this though, and closing and re-opening the explorer makes the issue go away (not sure about the original repro steps)

IanMatthewHuff:
I have a consistent repro on this (well on at least one aspect). Working on the fix, but to make sure that I note it here:

1. Create a notebook that has a bunch of variables (to make sure that we have pages)
2. Run the notebook, scroll down the variable list (all should be ok)
3. Restart the kernel
4. Run the notebook again
5. Scroll down the variable list (all is not well, things show up as loading... off the first page.)

minsa110:
Thanks for fixing this issue! :) I have one minor suggestion to add here. When I open the variable viewer, it says "No variables defined" while it's searching for / loading the variables. Could we replace that string with something like "Loading" / "Working on it..." / somewhere along those lines? And then afterwards when we actually don't have anything to show, replace that message with "No variables defined"?

Expected:
[...]

PR #8653, opened 2022-01-07
Fix variable on restart and remove old webview code path

IanMatthewHuff:
For #7970


src/datascience-ui/interactive-common/redux/reducers/variables.ts
```diff
@@ -238,8 +238,8 @@ function handleRestarted(arg: VariableReducerArg): IVariableState {
                 sortColumn: arg.prevState.sortColumn,
                 sortAscending: arg.prevState.sortAscending,
                 startIndex: 0,
-                pageSize: arg.prevState.pageSize,
-                refreshCount: 0
+                pageSize: 5,
+                refreshCount: arg.prevState.refreshCount
```
IanMatthewHuff, line 242: This here is the actual fix, when we moved from hosting this in each webview view to hosting a single instance in the variable view we changed how we use refreshCount some. If you switch between documents you can sometime have the same execution count as before, in which case we bump up the refresh count to force a refresh. 

However on restart this could lead to a state where we could have the same execution count and the same refresh count. In this case it would not realized that it needed to fetch new data, and you could get odd places where you would get Loading... and the rows would never load. Just keep refreshCount always counting up even after a restart to prevent this.

rchiodo (approved):

src/datascience-ui/interactive-common/variableExplorer.tsx
```diff
@@ -182,31 +176,19 @@ export class VariableExplorer extends React.Component<IVariableExplorerProps, IV
         if (!fastDeepEqual(this.props.variables, nextProps.variables)) {
             return true;
         }
-        if (
-            prevState.containerHeight !== this.state.containerHeight ||
-            (prevState.gridHeight !== this.state.gridHeight && !this.props.standaloneMode)
-        ) {
+        if (prevState.containerHeight !== this.state.containerHeight) {
```
IanMatthewHuff, line 179: The real fix was above, the other changes here (maybe should have done this in a separate PR, but I was cleaning it up while working so I figured why not) are removing the few special cases used for when we hosted this control in a webview document versus the new single webview view. We now only have the webview view case.

PR #8657, opened 2022-01-07
Ask for prerelease switch

rchiodo:
This is a temporary fix to address comments made here:

https://github.com/microsoft/vscode/issues/15756#issuecomment-1007629482

If an insider's user is not on the prerelease bits, we'll ask them to switch.

Message looks like so:

image


rchiodo:
Uggh this requires this to be fixed first:
https://github.com/microsoft/vscode/issues/139528.

I thought that was already working. Turns out it isn't finished yet.

amunger (approved):

IanMatthewHuff (approved):

rchiodo:
microsoft/vscode#139528 was fixed yesterday. Reopening this.

PR #8671, opened 2022-01-11
Monkey patch the kernel to allow debugging and tracebacks to coexist

### LINKED ISSUE #8568: Interactive window not showing stack trace
## Environment data

-   VS Code version: 1.63.1
-   Jupyter Extension version (available under the Extensions sidebar): 2021.11
-   Python Extension version (available under the Extensions sidebar): 2021.12
-   OS (Windows | Mac | Linux distro) and version: Windows 10
-   Python and/or Anaconda version: 3.9.7
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): venv
-   Jupyter server running: Local | Remote | N/A

## Expected behaviour

Upon a crash, the interactive window should show which line from my code is responsible for the crash. 

## Actual behaviour

image

The information does not show the line where the exception was raised. Howver it does show the stack trace if it's raised from a package. But it does not display the line of my code which prompted the exception.

image
image

rchiodo:
Fixes #8568

The root cause was explained in #8568 but I'll reiterate here.

We have code that does this:

```
import os;os.environ["IPYKERNEL_CELL_NAME"] = '${hash?.runtimeFile}'
```

This environment variable is used by the kernel to 'name' a cell execution so that when we get a stackframe in the debugger, it can be uniquely identified.

The problem was it was never being 'unset'. 

So when an error occurs and jupyter tries to parse all of the cells that lead to an exception, it can't determine which cell ```<ipython-2-hashyvalue>``` is actually referring to.

Example:

1. User executes ```def foo():\n  raise Exception("FOO")\n```
2. We execute ```import os;os.environ["IPYKERNEL_CELL_NAME"] = '<ipython-1-hashy.py>'``` as a silent cell
3. We execute ```def foo():\n  raise Exception("FOO")\n``` 
4. Jupyter associates ```def foo():\n  raise Exception("FOO")\n``` with ```<ipython-1-hashy.py>```
5. User executes ```foo()```
6. We execute ```import os;os.environ["IPYKERNEL_CELL_NAME"] = '<ipython-2-hashy2.py>'``` as a silent cell
7. Jupyter associates ```import os;os.environ["IPYKERNEL_CELL_NAME"] = '<ipython-2-hashy2.py>'``` with ```<ipython-1-hashy.py>``` because IPYKERNEL_CELL_NAME was set
8. We execute ```foo()```
9. Jupyter associates ```foo()``` with ```<ipython-2-hashy2.py>```
5. Exception is raised
6. Traceback is generated that has the following in it
- <ipython-1-hashy.py>
- <ipython-2-hashy.py>
7. ```<ipython-1-hashy.py>``` is mapped to the ```import os;os.environ["IPYKERNEL_CELL_NAME"] = '<ipython-2-hashy2.py>``` from step 7 instead of ```def foo():\n  raise Exception("FOO")\n``` from step 3
8. Traceback is invalid.

The crux of the problem is that we need an environment variable set during a cell executing but then unset for everything that comes after it (or set to something different).

What if we could execute the code we use to generate the hash inside the kernel somehow?

I had 3 ideas on how to solve this (picked the last one)
1. Use the debugger to eval the code prior to running each cell
2. Use the [pre_run_code_hook](https://github.com/ipython/ipython/blob/996c53ed74e204063c12adf4cab6896657c36e60/IPython/core/hooks.py#L55) from IPython to run code before
3. Monkey patch the run_cell in the kernel

I picked the last one because:
1. These entries need to be set regardless of whether or not we're debugging. It's too late if we have to wait for debugging because previous cells wouldn't have been mapped. (Makes stepping into predefined functions impossible)
2. ```pre_run_code_hook``` is deprecated and doesn't seem to actually get the code. The code is necessary in order to compute the hash. Plus we'd have to write an entire kernel extension.

### DIFF
--- a/pythonFiles/vscode_datascience_helpers/kernel/addRunCellHook.py
+++ b/pythonFiles/vscode_datascience_helpers/kernel/addRunCellHook.py
@@ -10,28 +10,31 @@ __VSCODE_wrapped_run_cell = False
 # https://github.com/microsoft/vscode-jupyter/blob/312511f3cbd8b2bb5bc70fa9b771429e22d0c258/src/client/datascience/editor-integration/cellhashprovider.ts#L181
 def __VSCODE_compute_hash(code, number=0):
     hash_digest = _VSCODE_hashlib.sha1(code.encode("utf-8")).hexdigest()
-    return '<ipython-input-{0}-{1}>'.format(number, hash_digest[:12])
+    return "<ipython-input-{0}-{1}>".format(number, hash_digest[:12])
+
 
 # This function will wrap InteractiveShell.run-cell and force its IPYKERNEL_CELL_NAME to match the
-# predicted value used by the Jupyter Extension 
+# predicted value used by the Jupyter Extension
 def __VSCODE_wrap_run_cell(wrapped_func):
-        old_func = wrapped_func.__func__
-        __VSCODE_wrapped_run_cell = True
-        def wrapper(*args, **kwargs):
-            try:
-                store_history = args[2] if len(args) > 2 else kwargs['store_history']
-                if store_history:
-                    predicted_name = __VSCODE_compute_hash(args[1], args[0].execution_count)
-                    os.environ["IPYKERNEL_CELL_NAME"] = predicted_name
-                result = old_func(*args, **kwargs)
-                if store_history:
-                    del os.environ["IPYKERNEL_CELL_NAME"]
-                return result
-            except:
-                return old_func(*args, **kwargs)
-        return _VSCODE_types.MethodType(wrapper, wrapped_func.__self__)
+    old_func = wrapped_func.__func__
+    __VSCODE_wrapped_run_cell = True
+
+    def wrapper(*args, **kwargs):
+        try:
+            store_history = args[2] if len(args) > 2 else kwargs["store_history"]
+            if store_history:
+                predicted_name = __VSCODE_compute_hash(args[1], args[0].execution_count)
+                os.environ["IPYKERNEL_CELL_NAME"] = predicted_name
+            result = old_func(*args, **kwargs)
+            if store_history:
+                del os.environ["IPYKERNEL_CELL_NAME"]
+            return result
+        except:
+            return old_func(*args, **kwargs)
+
+    return _VSCODE_types.MethodType(wrapper, wrapped_func.__self__)
+
 
 # Double check this still exists on IPython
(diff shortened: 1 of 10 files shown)


rchiodo:
@DonJayamanne it would be great if you could look at this one when you get back. I'm a little nervous patching the kernel. Thought there might be some other way to patch it.

rchiodo:
This also fixes errors with stepping into previously defined cells. They had the same problem as the stack trace.

pythonFiles/vscode_datascience_helpers/kernel/addRunCellHook.py
```diff
@@ -0,0 +1,40 @@
+import types as _VSCODE_types
+import os
+import hashlib as _VSCODE_hashlib
+from IPython import get_ipython
+
+# Variable that prevent wrapping from happing more than_once
+__VSCODE_wrapped_run_cell = False
+
+# This function computes the hash for the code. It must follow the same algorithm as in use here:
```
amunger, line 9: Seems like that logic should be in a shared location

amunger: haha nevermind, python vs ts

src/test/datascience/helpers.ts
```diff
@@ -151,20 +153,26 @@ export async function runCurrentFile(
...
-    }
-    assert.ok(codeCell !== undefined, 'No code cell found in interactive window notebook document');
+    await waitForCondition(
+        async () => {
+            const codeCells = notebookDocument?.getCells().filter((c) => c.kind === vscode.NotebookCellKind.Code);
+            codeCell = codeCells && codeCells.length ? codeCells[codeCells.length - 1] : undefined;
+            return codeCell && (numberOfCells === -1 || numberOfCells === codeCells!.length) ? true : false;
+        },
+        defaultNotebookTestTimeout,
+        `No code cell found in interactive window notebook documen`
```
amunger, line 174: typo: document,
also - is there a reason you want to use backticks for a constant string here?

src/test/datascience/interactiveDebugging.vscode.test.ts
```diff
@@ -593,4 +595,97 @@ def foo():
             `Cursor did not move to expected line when hitting stepping into`
         );
     });
+
+    test('Step into a previous cell', async () => {
+        // Need a function and a call to i
```
amunger, line 600: typo: another missing t!

amunger (approved):

PR #8683, opened 2022-01-12
Split markdown cells into code cells if they have code in them

### LINKED ISSUE #8543: markdown cells not being split for code
## Environment data

-   VS Code version: 1.63.0
-   Jupyter Extension version (available under the Extensions sidebar): v2021.11.1001550889
-   Python Extension version (available under the Extensions sidebar): v2021.12.1559732655
-   OS (Windows | Mac | Linux distro) and version: Windows
-   Python and/or Anaconda version: 3.7.9

## Expected behaviour

XXX

## Actual behaviour

XXX

## Steps to reproduce:

1. copy paste the following cell:

```
#%% [markdown]
# ## Some Markdown Text
print("hi")
```

The output in IPython 7.30.1 will only be the markdown text, with no code running.

Removing the "[markdown]" from the cell lets the markdown text be treated as a comment and the code executes, since it's not just a code cell. All of my python files have cells that start with markdown and go into code. Do I have to manually separate every cell?

Same problem is here [microsoft/vscode-jupyter#3501](https://github.com/microsoft/vscode-jupyter/issues/3501)

The automatic splitting appears to be a feature according to [microsoft/vscode-jupyter#2285](https://github.com/microsoft/vscode-jupyter/issues/2285)
Seems this is indeed a bug then

rchiodo:
Fixes #8543

### DIFF
--- a/src/test/datascience/interactiveWindow.vscode.test.ts
+++ b/src/test/datascience/interactiveWindow.vscode.test.ts
@@ -428,6 +428,18 @@ ${actualCode}
         assert.ok(hrefs[2].endsWith("line=2'"), 'Wrong third ref line');
     });
 
+    test('Running a cell with markdown and code runs two cells', async () => {
+        const { activeInteractiveWindow } = await runCurrentFile(
+            interactiveWindowProvider,
+            '# %% [markdown]\n# # HEADER\n# **bold**\nprint(1)',
+            disposables
+        );
+        const lastCell = await waitForLastCellToComplete(activeInteractiveWindow, 2, true);
+
+        // Parse the last cell's output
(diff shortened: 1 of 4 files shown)


src/test/datascience/interactiveWindow.vscode.test.ts
```diff
@@ -428,6 +428,18 @@ ${actualCode}
         assert.ok(hrefs[2].endsWith("line=2'"), 'Wrong third ref line');
     });
 
+    test('Running a cell with markdown and code runs two cells', async () => {
+        const { activeInteractiveWindow } = await runCurrentFile(
```
IanMatthewHuff, line 432: I didn't actually know this was allowed. Odd.

PR #8711, opened 2022-01-16
Return same kernel connection instance everytime

DonJayamanne:
Debt item
Instead of returning a new instance we should return the same instance.
The object is basically a singleton and readonly, hence safe to return the same object.
Also makes it easier for users of the API to tie to an instance instead of just tracking the ID.


DonJayamanne:
Required for https://github.com/microsoft/vscode-jupyter/pull/8716

Issue #8710, opened 2022-01-16
Log paths to ipykernel in output panels


DonJayamanne:
We have users installing packages in user site_packages folders.
and these end up getting used by virtual environments.
E.g. if you have Anaconda Python 3.8 as your default and install ipykernel and then use a virtual env 3.6, this will attempt to use the ipykenrel setup in the user site_packages folder and things won't work.

We should log the paths to the locations of these packages to be able to better diagnose why things are failing.

```python
import ipykernel
print(ipykernel.__file__)
```

PR #8713, opened 2022-01-16
Log ipykernel version and path

DonJayamanne:
For https://github.com/microsoft/vscode-jupyter/issues/8710

### DIFF
--- a/src/client/common/experiments/service.ts
+++ b/src/client/common/experiments/service.ts
@@ -10,7 +10,7 @@ import { sendTelemetryEvent } from '../../telemetry';
 import { EventName } from '../../telemetry/constants';
 import { IApplicationEnvironment } from '../application/types';
 import { JVSC_EXTENSION_ID, STANDARD_OUTPUT_CHANNEL } from '../constants';
-import { traceInfo } from '../logger';
+import { traceVerbose } from '../logger';
 import {
     GLOBAL_MEMENTO,
     IConfigurationService,
@@ -82,7 +82,7 @@ export class ExperimentService implements IExperimentService {
             this.globalState
         );
 
-        traceInfo(`Experimentation service retrieved: ${this.experimentationService}`);
+        traceVerbose(`Experimentation service retrieved: ${this.experimentationService}`);
 
         this.logExperiments();
     }
--- a/src/client/common/process/environmentActivationService.ts
+++ b/src/client/common/process/environmentActivationService.ts
@@ -628,19 +628,20 @@ export class EnvironmentActivationService implements IEnvironmentActivationServi
         if (!interpreter?.path) {
             return;
         }
-        traceInfo(`Getting activation commands for ${interpreter.path}`);
+        traceVerbose(`Getting activation commands for ${interpreter.path}`);
         const key = ENVIRONMENT_ACTIVATION_COMMAND_CACHE_KEY_PREFIX.format(interpreter.path);
         const cachedData = this.memento.get<string[]>(key, []);
         if (cachedData && cachedData.length > 0) {
-            traceInfo(`Getting activation commands for ${interpreter.path} are cached.`);
+            traceVerbose(`Getting activation commands for ${interpreter.path} are cached.`);
             return cachedData;
         }
         if (this.envActivationCommands.has(key)) {
-            traceInfo(`Getting activation commands for ${interpreter.path} are cached with a promise.`);
+            traceVerbose(`Getting activation commands for ${interpreter.path} are cached with a promise.`);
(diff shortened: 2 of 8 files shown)


src/client/common/experiments/service.ts
```diff
@@ -82,7 +82,7 @@ export class ExperimentService implements IExperimentService {
             this.globalState
         );
 
-        traceInfo(`Experimentation service retrieved: ${this.experimentationService}`);
+        traceVerbose(`Experimentation service retrieved: ${this.experimentationService}`);
```
DonJayamanne, line 85: Currently not useful as we dump the entire class, hence changing to verbose (also doesn't display well in output panel, we just get `[Object object]`)

src/client/common/process/environmentActivationService.ts
```diff
@@ -656,7 +657,7 @@ export class EnvironmentActivationService implements IEnvironmentActivationServi
                 if (!activationCommands || activationCommands.length === 0) {
                     return;
                 }
-                traceVerbose(`Activation Commands received ${activationCommands} for shell ${shellInfo.shell}`);
+                traceInfo(`Activation Commands received ${activationCommands} for shell ${shellInfo.shell}`);
```
DonJayamanne, line 660: Ensured to display the one message as information and the rest and versbose.
This way anytime we get the command the first we display it, and subsequent access will not be displayed in output panels (else its very noisy), but sill displayed in verbose.

src/client/datascience/ipywidgets/ipyWidgetScriptSource.ts
```diff
@@ -195,7 +195,7 @@ export class IPyWidgetScriptSource implements ILocalResourceUriConverter {
             this.factory
         );
         this.initializeNotebook();
-        traceInfo('IPyWidgetScriptSource.initialize');
+        traceVerbose('IPyWidgetScriptSource.initialize');
```
DonJayamanne, line 198: Not sure this is useful, leaving as verbose

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -905,7 +905,7 @@ export async function sendTelemetryForPythonKernelExecutable(
 
 export async function executeSilently(session: IJupyterSession, code: string): Promise<nbformat.IOutput[]> {
     traceInfo(
-        `Executing (status ${session.status}) silently Code = ${code.substring(0, 100).splitLines().join('\\n')}`
+        `Executing silently Code (${session.status}) = ${code.substring(0, 100).splitLines().join('\\n')}`
```
DonJayamanne, line 908: Changed this as when I was looking at th emessages i saw
`Executing (status idle) silently code: xyz`
`Executing silently code (completed) : xyz`

Both looked off to me, just changed so that the status is in brackets & its easy to identify the two easily and align them visually (i feel its easier to read)

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -904,12 +903,14 @@ export class Kernel implements IKernel {
                 launchingFile
             );
             if (suggestedDir && (await this.fs.localDirectoryExists(suggestedDir))) {
+                traceInfo('UpdateWorkingDirectoryAndPath in Kernel');
```
DonJayamanne, line 906: Moved these closer to when the paths are changed, this way when its logged we know its changing 
& if its not logged, then cwd is not changing

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -924,7 +925,6 @@ export class Kernel implements IKernel {
                 isLocalHostConnection(this.kernelConnectionMetadata)) &&
             isPythonKernelConnection(this.kernelConnectionMetadata)
         ) {
-            traceInfo('changeDirectoryIfPossible');
```
DonJayamanne, line 927: No longer required as the calling code will log when cwd changes

src/client/extensionActivation.ts
```diff
@@ -87,11 +87,12 @@ async function activateLegacy(
     );
     serviceManager.addSingletonInstance<boolean>(IsCodeSpace, env.uiKind == UIKind.Web);
 
-    // Log versions of extensions.
+    // Log env info.
+    standardOutputChannel.appendLine(`${env.appName} (${version}, ${env.remoteName}, ${env.appHost})`);
```
DonJayamanne, line 91: We tend to ask this.
Logging most of this in output panel will avoid the need to ask for this information.

rchiodo (approved):

PR #8722, opened 2022-01-18
Make sure to find active interpreter even with no folder

### LINKED ISSUE #8409: Syntax colorization for Python is incorrect in Notebookcells when no folder or active interpreter
* Latest Insiders
* Pylance 2021.11.2

**See below (latest insiders):** (this used to work in Insiders, see screenshots here https://github.com/microsoft/vscode-jupyter/issues/8374) 

image

**Here's a screenshot from the stable version (hence a regression):**
image

rchiodo:
Fixes #8409

Active interpreter is possible without a folder. Make sure to support this case.

### DIFF
--- a/src/client/datascience/notebook/intellisense/intellisenseProvider.ts
+++ b/src/client/datascience/notebook/intellisense/intellisenseProvider.ts
@@ -16,6 +16,8 @@ import { INotebookControllerManager, INotebookLanguageClientProvider } from '../
 import { VSCodeNotebookController } from '../vscodeNotebookController';
 import { LanguageServer } from './languageServer';
 
+const EmptyWorkspaceKey = 'EMPTY_WORKSPACE_KEY';
+
 /**
  * This class sets up the concatenated intellisense for every notebook as it changes its kernel.
  */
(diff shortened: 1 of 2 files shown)


src/client/datascience/notebook/intellisense/intellisenseProvider.ts
```diff
@@ -82,15 +84,16 @@ export class IntellisenseProvider implements INotebookLanguageClientProvider, IE
         const folder =
             this.workspaceService.getWorkspaceFolder(fsPath ? Uri.file(fsPath) : undefined)?.uri ||
             (this.workspaceService.rootPath ? Uri.file(this.workspaceService.rootPath) : undefined);
-        if (folder && !this.activeInterpreterCache.has(folder.fsPath)) {
+        const key = folder ? folder.fsPath : EmptyWorkspaceKey;
```
DonJayamanne, line 87: Can't this be an empty string? I'd feel more comfortable with an empty string than using some other hardcoded values.

DonJayamanne (approved):

src/client/datascience/notebook/intellisense/intellisenseProvider.ts
rchiodo: Sure. i don't think it matters.

PR #8733, opened 2022-01-19
Getting started walkthrough for data scientists

minsa110:
For https://github.com/microsoft/vscode-jupyter-internal/issues/243


package.json
```diff
@@ -80,6 +81,58 @@
         }
     },
     "contributes": {
+        "walkthroughs": [
+            {
+                "id": "jupyterWelcome",
+                "title": "Get started with Jupyter Notebooks",
+                "description": "Your first steps to set up a Jupyter project with all the powerful tools and features that the Jupyter extension has to offer!",
+                "when": "false",
```
rchiodo, line 89: when: false? Does this mean it doesn't show up?

package.json
```diff
@@ -2239,7 +2292,7 @@
         "tslint-microsoft-contrib": "^5.0.3",
         "tslint-plugin-prettier": "^2.1.0",
         "typemoq": "^2.1.0",
-        "typescript": "^4.4.3",
+        "typescript": "^4.5.4",
```
rchiodo, line 2295: THis seems like an odd addition? Did you need a new typescript compiler?

rchiodo:
Could you screenshot what it actually looks like and paste it here? It's hard to get a feel for the walkthrough just by looking at the package.json.

rchiodo:
I was also wondering how the walkthrough entries get translated to other languages? Is there support for loading the strings in them from an nls.json file?

package.json
minsa110: Yep, by default it's turned off. The experiment turns it on for folks:

vscode.gettingStarted.overrideCategory.ms-toolsai.jupyter.jupyterWelcome.when:workspacePlatform != webworker

package.json
minsa110: No need for a new TS compiler. I think it just updated it with the one I used.

rchiodo: You should back out this change then. This should also remove the change to package-lock.json

package.json
```diff
@@ -80,6 +81,58 @@
...
+                "title": "Get started with Jupyter Notebooks",
+                "description": "Your first steps to set up a Jupyter project with all the powerful tools and features that the Jupyter extension has to offer!",
+                "when": "false",
+                "steps": [
+                    {
+                        "id": "jupyter.createNewNotebook",
+                        "title": "Create or open a Jupyter Notebook",
+                        "description": "Right click in the file explorer and create a new file with an .ipynb extension. Or, open the [Command Palette](command:workbench.action.showCommands) and run the command \n``Jupyter: Create New Blank Notebook``.\n[Create new Jupyter Notebook](command:toSide:jupyter.createnewnotebook)\n If you have an existing project, you can also [open a folder](command:workbench.action.files.openFolder) and/or clone a project from GitHub: [clone a Git repository](command:git.clone).",
+                        "media": {
+                            "image": "images/OpenOrCreateNotebook.gif",
```
DonJayamanne, line 96: I think i remember VS Code adding support for themable images (svgs) for welcome pages.
Else someone on a light them (default) would get dark images.
@luabud might have more information about this.

DonJayamanne: Also, do we need gifs for everything?
Python extension has SVGs and I personally prefer that, rather than just having dark images

DonJayamanne (changes requested):
Suggest we use SVGs where possible like Python extension to cater all users with different themes

package.json
IanMatthewHuff: I'm a bit torn myself. The SVGs look much better, but at the same time the gifs feel like they explain things better.

DonJayamanne: Agreed, but for some like opening a notebook, I don't think we need a gif, 
also thats the very first walkthrough, i'd like that to be themeble via svg

IanMatthewHuff (approved):

package.json
minsa110: Yea, I think Python walkthrough pages use them. I was thinking that once we start seeing positive numbers from the experiment, we'd ask Miguel to help us make them for us as well.

minsa110:
Hey @rchiodo, sorry I missed your other reply. It is currently not being localized and we plan on running the experiment for users who have English set as their default language. I don't think this has been addressed by other extensions either with exception to [this](https://github.com/microsoft/vscode-cpptools/blob/main/Extension/walkthrough/debugconfig/run-and-debug-project-linux.md) one that C++ team uses

package-lock.json
```diff
@@ -21647,9 +21647,9 @@
             }
         },
         "typescript": {
-            "version": "4.4.3",
```
rchiodo, line 21650: This change needs to be reverted too. Otherwise it doesn't match the package.json.

package.json
IanMatthewHuff: I'm fine with that personally, you can track it how you want, but might be good to create an issue for that change of adding SVGs, just so it doesn't get forgotten in day to day shuffle..

DonJayamanne: Agreed with Ian, approving

DonJayamanne (approved):

package.json
luabud: FWIW there are SVGs that could be reused for this first tile, for example these: https://github.com/microsoft/vscode-python/blob/main/resources/walkthrough/create-notebook.svg or https://github.com/microsoft/vscode-python/blob/main/resources/walkthrough/data-science.svg

rchiodo (approved):

PR #8740, opened 2022-01-19
Switch to reading package json from disk

### LINKED ISSUE #8724: Being prompted to switch to Jupyter pre-release. I'm already on pre-release.
On my vscode insiders I'm seeing this:
image

Yet my extension is this:
image

rchiodo:
Fixes #8724 

Asking VS code core if there's a better way to do this but this will fix the problem.

### DIFF
--- a/src/client/datascience/prereleaseChecker.ts
+++ b/src/client/datascience/prereleaseChecker.ts
@@ -20,12 +20,12 @@ export class PreReleaseChecker implements IExtensionSingleActivationService {
         @inject(IApplicationEnvironment) private readonly appEnv: IApplicationEnvironment,
         @inject(IApplicationShell) private readonly appShell: IApplicationShell,
         @inject(IMemento) @named(GLOBAL_MEMENTO) private globalState: vscode.Memento,
-        @inject(IsPreRelease) private isPreRelease: boolean
+        @inject(IsPreRelease) private isPreRelease: Promise<boolean>
     ) {}
     public async activate(): Promise<void> {
         // Ask user if the version is not prerelease
         const dontAsk = this.globalState.get(PRERELEASE_DONT_ASK_FLAG, false);
-        if (!this.isPreRelease && this.appEnv.channel === 'insiders' && !dontAsk) {
+        if (!(await this.isPreRelease) && this.appEnv.channel === 'insiders' && !dontAsk) {
             const yes = localize.DataScience.usingNonPrereleaseYes();
             const no = localize.DataScience.usingNonPrereleaseNo();
             const dontAskAgain = localize.DataScience.usingNonPrereleaseNoAndDontAskAgain();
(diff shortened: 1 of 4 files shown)


rchiodo:
Seems there isn't a better way to do this. I guess we could write to the package.json during the build process (prior to building). Then it would be in the imported package.json.

src/client/extensionActivation.ts
```diff
@@ -71,7 +73,11 @@ async function activateLegacy(
         (context.extensionMode === ExtensionMode.Development ||
             workspace.getConfiguration('jupyter').get<boolean>('development', false));
     serviceManager.addSingletonInstance<boolean>(IsDevMode, isDevMode);
-    const isPreRelease = isDevMode || context.extension.packageJSON?.__metadata?.preRelease;
+    const packageJSONLive = JSON.parse(
+        fsExtra.readFileSync(path.join(context.extensionPath, 'package.json'), { encoding: 'utf-8' })
```
DonJayamanne, line 77: Can we make this async, this actually slows loading this extension and slows other extensions with this sync operation.

rchiodo: Then it would have to be a separate service because this flag needs to be set before other items in the container are created.

I guess it could be an async function? Instead of setting a boolean, set a promise?

DonJayamanne: > set a promise

Yes, we can just give the promise i.e. turn isPreRelease into a `Promise<boolean>`

IanMatthewHuff (approved):

src/client/datascience/prereleaseChecker.ts
```diff
@@ -20,12 +20,12 @@ export class PreReleaseChecker implements IExtensionSingleActivationService {
...
         @inject(IApplicationShell) private readonly appShell: IApplicationShell,
         @inject(IMemento) @named(GLOBAL_MEMENTO) private globalState: vscode.Memento,
-        @inject(IsPreRelease) private isPreRelease: boolean
+        @inject(IsPreRelease) private isPreRelease: Promise<boolean>
     ) {}
     public async activate(): Promise<void> {
         // Ask user if the version is not prerelease
         const dontAsk = this.globalState.get(PRERELEASE_DONT_ASK_FLAG, false);
-        if (!this.isPreRelease && this.appEnv.channel === 'insiders' && !dontAsk) {
+        if (!(await this.isPreRelease) && this.appEnv.channel === 'insiders' && !dontAsk) {
```
DonJayamanne, line 28: We probably should not await here in `activate`, as this slows loading the entire Jupyter extension.
This could happen in the background.
E.g. 
```typescript
public asycn activate() {
	void this.initialize();
}
private async initialize() {
	await this.isPreRelease
}
```

This way, the jupyter extension will activate without having to awit for the FS operation to complete.
Else its blocked & slows down, I know its not much, but death by a thousand cuts

rchiodo: Sure good idea.

DonJayamanne (approved):

Issue #8431, opened 2021-12-02
Interactive Widgets fail when restarting kernels on Remote Jupyter & non-Raw kernels 


DonJayamanne:
Follow up on https://github.com/microsoft/vscode-jupyter/issues/8378

DonJayamanne:
Wait, this works

DonJayamanne:
Tested and doesn't work, looks like a regression of https://github.com/microsoft/vscode-jupyter/issues/8378

PR #8744, opened 2022-01-19
More widget tests

DonJayamanne:


src/test/datascience/widgets/standard.vscode.test.ts
```diff
@@ -73,32 +76,111 @@ suite('Standard IPyWidget (Execution) (slow) (WIDGET_TEST)', function () {
...
+                const innerHTML = await comms.queryHtml('.widget-checkbox', cell.outputs[0].id);
+                assert.include(innerHTML, 'Check me');
+                assert.include(innerHTML, '<input type="checkbox');
+                return true;
+            },
+            WidgetRenderingTimeoutForTests,
+            'Checkbox not rendered'
+        );
+    });
+    test.skip('Widget renders after executing a notebook which was saved after previous execution', async () => {
```
DonJayamanne, line 134: Doesn't work, will need to fix this (possibly something wrong in our code)
Looks like a regression.
Works only if you wait for some time

IanMatthewHuff: Does the test need to be skipped and an issue filed then? Or were you looking at this fix as part of this PR?

IanMatthewHuff: Whoops, totally missed the skip that was already there.

IanMatthewHuff (approved):

src/test/datascience/widgets/standard.vscode.test.ts
rchiodo: Seems like a regression? Didn't we have this bug before?
https://github.com/microsoft/vscode-jupyter/issues/8431

rchiodo (approved):

src/test/datascience/widgets/standard.vscode.test.ts
DonJayamanne: Aah, yes its a duplicate then.
This wasn't fixed, its actually flaky, it works locally (seems to be 100% of the time).
Must be some race condition (i think thats obvious).

Will re-open #8431
File https://github.com/microsoft/vscode-jupyter/issues/8749 & https://github.com/microsoft/vscode-jupyter/issues/8749

I.e. we have 3 issues related to widgets, but they all seem to work locally (at least for me)
Possible its just flaky

DonJayamanne: Damn it, looks like restart tests are also flaky

amunger (approved):

Issue #8720, opened 2022-01-18
I wish the Interactive Window wouldn't steal editor group while a script is running.


greazer:
1. Open a python script with cells that will take a while to run.
2. Start the script with "Run Below" on the first cell.
3. Open an MD file and then open the preview. The preview should open on top of the Interactive Window.

As the IW gets updated, it will pop to the top of the editor group that it shares with the MD preview window.

Note that this doesn't seem to happen with stdout output. But it does happen with Plotly. Not sure about other plotting packages.

rchiodo:
We use a vscode API to 'scroll' to the latest cell on output. So this would require changes to the VS code API to support not 'stealing' the focus.

rchiodo:
When we implement this, we need to make sure 'goto cell' does still set the focus. We can probably do this with another command though.

PR #8745, opened 2022-01-20
Avoid opening iw editor for adding cells

rebornix:
For #8720

Opening the iw is not necessary for inserting new cells.


DonJayamanne (approved):

rchiodo (approved):

rebornix:
I didn't see any reason why the change failed one of the test env sometimes, merging now but will keep an eye on it.

PR #8747, opened 2022-01-20
Remove redundant API method

DonJayamanne:
While attempting to document the code, found an API call that was unnecessary.
* `listNonPythonKernels` is not required
* We list non-python kernels when fetching all kernels (including Python) in the `listKernels` method

I think we used this code path in the old code and is now redundant.

As Peng mentioned, writing the documentation helps identify issues & I think its already proved beneficial here (I think).

### DIFF
--- a/src/client/datascience/kernel-launcher/localKernelFinder.ts
+++ b/src/client/datascience/kernel-launcher/localKernelFinder.ts
@@ -125,10 +125,6 @@ export class LocalKernelFinder implements ILocalKernelFinder {
         }
     }
 
-    public async listNonPythonKernels(cancelToken?: CancellationToken): Promise<LocalKernelConnectionMetadata[]> {
-        return this.filterKernels(await this.nonPythonKernelFinder.listKernelSpecs(false, cancelToken));
-    }
-
     /**
      * Search all our local file system locations for installed kernel specs and return them
      */
--- a/src/client/datascience/notebook/notebookControllerManager.ts
+++ b/src/client/datascience/notebook/notebookControllerManager.ts
@@ -182,16 +182,9 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             const stopWatch = new StopWatch();
 
             // Fetch the list of kernels ignoring the cache.
-            Promise.all([
-                this.loadNotebookControllersImpl({
-                    listLocalNonPythonKernels: true,
-                    useCache: 'ignoreCache'
-                }),
-                this.loadNotebookControllersImpl({
-                    listLocalNonPythonKernels: false,
-                    useCache: 'ignoreCache'
-                })
-            ])
+            this.loadNotebookControllersImpl({
+                useCache: 'ignoreCache'
+            })
                 .catch((ex) => console.error('Failed to fetch controllers without cache', ex))
                 .finally(() => {
                     this._controllersLoaded = true;
@@ -205,7 +198,6 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
                             timer = setTimeout(
                                 () =>
                                     this.loadNotebookControllersImpl({
-                                        listLocalNonPythonKernels: false,
                                         useCache: 'ignoreCache'
                                     }).catch((ex) =>
                                         console.error(
@@ -228,11 +220,6 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             // Fetching remote kernels cannot be done from cache.
             const promises = [
                 this.loadNotebookControllersImpl({
-                    listLocalNonPythonKernels: true,
-                    useCache: 'useCache'
-                }),
-                this.loadNotebookControllersImpl({
-                    listLocalNonPythonKernels: false,
                     useCache: 'useCache'
                 })
             ];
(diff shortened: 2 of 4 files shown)


src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -125,10 +125,6 @@ export class LocalKernelFinder implements ILocalKernelFinder {
         }
     }
 
-    public async listNonPythonKernels(cancelToken?: CancellationToken): Promise<LocalKernelConnectionMetadata[]> {
-        return this.filterKernels(await this.nonPythonKernelFinder.listKernelSpecs(false, cancelToken));
```
DonJayamanne, line 129: this same method is called inside `this.listKernlesWithoutCache` which is inturn called in `listKernels`

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -182,16 +182,9 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
 
             // Fetch the list of kernels ignoring the cache.
-            Promise.all([
-                this.loadNotebookControllersImpl({
-                    listLocalNonPythonKernels: true,
-                    useCache: 'ignoreCache'
-                }),
-                this.loadNotebookControllersImpl({
-                    listLocalNonPythonKernels: false,
-                    useCache: 'ignoreCache'
```
DonJayamanne, line 192: We always retrieve both python & non-python kernel specs (hence no need to retrieve non-python kernel specs separately)

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -228,11 +220,6 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             // Fetching remote kernels cannot be done from cache.
             const promises = [
                 this.loadNotebookControllersImpl({
-                    listLocalNonPythonKernels: true,
-                    useCache: 'useCache'
-                }),
-                this.loadNotebookControllersImpl({
```
DonJayamanne, line 234: Again, (same as earlier) we always retrieve both python & non-python kernel specs (hence no need to retrieve non-python kernel specs separately)

rchiodo (approved):

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -362,7 +346,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         this.createNotebookControllers(connections);
 
         // If we're listing Python kernels & there aren't any, then add a placeholder for `Python` which will prompt users to install python
```
amunger, line 348: This comment is out of date. (It also seems to be backwards before the change)

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -362,7 +346,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
-        if ('listLocalNonPythonKernels' in options && !options.listLocalNonPythonKernels) {
+        if ('useCache' in options) {
```
amunger, line 349: why would we only care about this when using the cached kernels? It seems like it would be more necessary without the cache since the values are more likely to change.

amunger: ah, it looks like 'useCache' is not actually what I would expect

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -336,21 +323,18 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
      * Turn all our kernelConnections that we know about into registered NotebookControllers
      */
     private async loadNotebookControllersImpl(
-        options:
-            | { listLocalNonPythonKernels: boolean; useCache: 'useCache' | 'ignoreCache' }
-            | { listRemoteKernels: boolean }
+        options: { useCache: 'useCache' | 'ignoreCache' } | { listRemoteKernels: boolean }
```
amunger, line 326: listRemoteKernels is not used?

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -336,21 +323,18 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
         let connections: KernelConnectionMetadata[] = [];
-        if ('listLocalNonPythonKernels' in options) {
-            connections = await (options.listLocalNonPythonKernels
-                ? this.localKernelFinder.listNonPythonKernels(cancelToken.token, options.useCache)
-                : this.localKernelFinder.listKernels(undefined, cancelToken.token, options.useCache)
-            ).catch((ex) => {
-                traceError('Failed to get local kernel connections', ex);
-                return [] as KernelConnectionMetadata[];
-            });
+        if ('useCache' in options) {
```
amunger, line 331: you could switch this if statement to make it more clear what's going on: if (listRemoteKernels)

src/client/datascience/notebook/notebookControllerManager.ts
amunger: I get it now, I missed the | in the parameter list. 'useCache' existing in the options implies we're looking for local kernels.
Seems like this function could be split into findLocal and findRemote with how little code is shared between the two

Issue #4419, opened 2021-01-18
Export the ipynb to a directory outside of the current workspace, there wasn't any python code prompt at the start to change directory to the previous location of the .ipynb file.


zhouwangyang:
## Environment data
Version: 1.53.0-insider (user setup)
Commit: eaa959d34b863dec932077d94daece22504a396b
Date: 2021-01-18T05:13:25.733Z
Electron: 11.2.0
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Windows_NT x64 10.0.17134
anaconda: python 3.8.5(64 bit)
Build of python-insider: 2020.12.414227025
Build of Jupyter-insider: 2021.1.481581316-rc

## Expected behaviour
1.Open the file in the browser,  should get an initial cell added to change directory back to your workspace directory.
2.Open the file in the editor. There should be python code at the start to change directory to the previous location of the .ipynb file

## Actual behaviour
1. Open the file in the browser,  there did not  get an initial cell added to change directory back to your workspace directory.
2. Open the file in the editor. There didn't  python code at the start to change directory to the previous location of the .ipynb file

## Steps to reproduce:
1. Run the code manualTestFile.py, and export the .ipynb to a directory outside of the current workspace(not default export path)
2. Open the file in the browser, you should get an initial cell added to change directory back to your workspace directory
3. Follow the previous steps for import, but import an ipynb that is located outside of your current workspace
4.Open the file in the editor. There should be python code at the start to change directory to the previous location of the .ipynb file

IanMatthewHuff:
@zhouwangyang Thanks for filing. Confirmed I'm not seeing the directory change on import.

joyceerhl:
@IanMatthewHuff I was just debugging this. Provided that the python file we're running from is in the current workspace, this is working for me:
> Open the file in the browser, there did not get an initial cell added to change directory back to your workspace directory.

Exporting and then opening as editor, I am also seeing this:

greazer:
Regression perhaps, but we're unclear as to whether this needs to be fixed at all. We should change the test plan.

IanMatthewHuff:
I still repro this. But doesn't seem worthy of Sept consideration.

amunger:
When trying to reproduce this and looking at the code, it seems like all the limitations are by design:
export python script to notebook - only if current file exists on disk
export IW to notebook - no, no backing file
export notebook to python script - only when nbconvert is used as export method
[...]

PR #8751, opened 2022-01-20
marking 'change directory on import/export' as deprecated

amunger:
For #4419

This feature has a few issues: the path format is broken for Windows, the path is incorrect for importing/exporting notebooks, and it's not implemented in non-nbconvert code-paths. Looking at telemetry, this is a fairly low-use feature, so we can probably just remove it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1555,6 +1555,7 @@
                 "jupyter.changeDirOnImportExport": {
                     "type": "boolean",
                     "default": false,
+                    "deprecationMessage": "This setting is deprecated and will be removed in the next release.",
                     "description": "When importing or exporting a Jupyter Notebook add a directory change command to allow relative path loading to work.",
                     "scope": "resource"
                 },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1555,6 +1555,7 @@
                 "jupyter.changeDirOnImportExport": {
                     "type": "boolean",
                     "default": false,
+                    "deprecationMessage": "This setting is deprecated and will be removed in the next release.",
```
rchiodo, line 1558: I would have expected code changes as well? I assume we're still using this somewhere? Or are you thinking we'd do the actual removal next month?

rchiodo (approved):

package.json
amunger: yes, removing next month was the plan. It still works for translating python files into notebooks on non-Windows, so leaving that functionality while marking deprecated seems like the best way to do it

IanMatthewHuff (approved):

PR #8754, opened 2022-01-20
Add a language switcher that auto detects magics

### LINKED ISSUE #4588: Support language features for other languages when using cell magics
I would like to run bash scripts in a cell.
I use cell magics for this.
I would like the langauge of the cell to be automatically set to `bash`

Similarly when using %html, %%svg, etc...
%%script julia
%%html
%%kql

Provides intellisense, syntax highlighting & the like (ie. use VS Code features to improving cell editing experience for non-python & non-markdown cells)

Additionally we should make sure all of the languages supported by magics. This includes things like:
- Q# #10772

rchiodo:
Fixes #4588

Added a new class 'LanguageSwitcher' that watches for cell magics that switch a cell to a different type. 

Basically if you type something like ```%%html``` the cell will auto switch to html.

### DIFF
--- a/src/client/datascience/notebook/languageSwitcher.ts
+++ b/src/client/datascience/notebook/languageSwitcher.ts
@@ -1,0 +1,88 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+'use strict';
+import { injectable, inject } from 'inversify';
+import { TextDocumentChangeEvent, Range, Position, languages, workspace, TextDocument } from 'vscode';
+import { IExtensionSingleActivationService } from '../../activation/types';
+import { IDocumentManager } from '../../common/application/types';
+import { IConfigurationService, IDisposableRegistry } from '../../common/types';
+import { isNotebookCell } from '../../common/utils/misc';
+import { INotebookControllerManager } from './types';
+
+// This list comes from here: https://ipython.readthedocs.io/en/stable/interactive/magics.html#cell-magics
+const LanguageMagics = [
+    ['%%sql', 'sql'],
+    ['%%bash', 'shellscript'],
+    ['%%sh', 'shellscript'],
+    ['%%kql', 'kql'],
+    ['%%javascript', 'javascript'],
+    ['%%js', 'javascript'],
+    ['%%html', 'html'],
+    ['%%ruby', 'ruby'],
+    ['%%perl', 'perl'],
+    ['%%script sql', 'sql'],
+    ['%%script bash', 'shellscript'],
+    ['%%script sh', 'shellscript'],
+    ['%%script kql', 'kql'],
+    ['%%script javascript', 'javascript'],
+    ['%%script js', 'javascript'],
+    ['%%script html', 'html'],
+    ['%%script ruby', 'ruby'],
+    ['%%script perl', 'perl'],
+    ['%%python', 'python'],
+    ['%%python2', 'python'],
+    ['%%pypy', 'python'],
+    ['%%script python', 'python'],
+    ['%%script python2', 'python'],
+    ['%%script pypy', 'python']
+];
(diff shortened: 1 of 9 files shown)


IanMatthewHuff (approved):

src/client/datascience/notebook/languageSwitcher.ts
```diff
@@ -0,0 +1,88 @@
...
+        if (
+            isNotebookCell(e.document.uri) &&
+            this.configService.getSettings().switchCellsOnMagic &&
+            this.hasPythonController(e.document)
+        ) {
+            this.switchOnMatch(e, LanguageMagics);
+        }
+    }
+
+    private hasPythonController(cellDocument: TextDocument): boolean {
```
DonJayamanne, line 63: I think this should go into the core VS Code extension for Jupyter notebooks.
This change will work well for new notebooks and adding cells.
But if you save this notebook and re-open it, the language will not be preserved.
Also if you open an existing notebook, the languages will still be python, when they should have been changed based on the magics.

DonJayamanne (changes requested):
Best for core Jupyter extension in VS Code to load cells in the right language.

src/client/datascience/notebook/languageSwitcher.ts
IanMatthewHuff: Yeah, I was just looking in the core code recently. It's seems like there is a deeper issue here where language selection isn't preserved in the file at all. 

Isn't that a bigger issue though? If I choose some wacky language I'd expect VS Code to save and honor that. For example if I have '%%html' and choose SQL manually in the language selector I'd expect to get SQL when I reopen the file. Moving this into deserialize cell preferred language doesn't fix that.

DonJayamanne: > Moving this into deserialize cell preferred language doesn't fix that.

It will, if we save the language cell in a metadata of the cell.
Thought I'm not sure why anyone would change a cell from `html` to `sql` when one has `%%html` magic. possible yes, but why? Again we can address this but saving this information in the cell metadata. I don't think VS Code should try to save this information elsewhere, else it could be argued that once I share this file with someone else the same cell is no longer `sql` but treated as `html/python`, i.e. the saved info is still not preserved.

IanMatthewHuff: I think that I was thinking of my quick investigation into this issue some:
https://github.com/microsoft/vscode/issues/140673

This customer is using magics so making this change in core would be good for them. But when investigating on my own without the magics in the cell it was pretty surprising to me that it wouldn't honor the manual language selections that I had made on file save and reopen. I guess I don't see the point of the manual language selector if it's not honored in that situation. Would we expect someone to change it to a different language just temporarily?

rchiodo: I didn't realize the core wasn't saving the language in the metadata. Yeah this should probably be done in the core jupyter extension. And it should do this on open.

Although if we did this on open, it wouldn't have to be in the core language extension. It would just happen on every open.

DonJayamanne: >  It would just happen on every open.

Then we'd be marking documents as dirty when ever a document is opened.
Hence my argument for setting it in core, also in core today we alraedy set the cell langauge as `python` (hence setting language as `html` shoudl also be in core - i.e. serializer)

DonJayamanne: Lets discuss in standup

rchiodo: Didn't realize changing cell language makes it dirty. Seems weird if it doesn't actually write anything to metadata. 

Yeah serializer has to do it if we want to save it and not have the file be dirty.

PR #8768, opened 2022-01-22
Fix problem with markdown being passed to pylance

rchiodo:
Fixes: #8769

With pylance handling intellisense, we were also passing non python cells along. This filters those out so pylance never sees them.

Before change:

image

Notice the off by one error on the diagnostics

After change:

image


rchiodo:
No bug yet entered for this, as nobody else is using pylance yet.

DonJayamanne (approved):

PR #8773, opened 2022-01-22
Fix callstack walking and add 'rchiodo' as an allowed publisher

rchiodo:
I'm working on contextual help and the callstack walking didn't work (chain promise forced stack walk across async boundary and lost original caller)

I've also added myself as an allowed publisher.

### DIFF
--- a/src/client/api/kernelApi.ts
+++ b/src/client/api/kernelApi.ts
@@ -5,7 +5,7 @@ import { injectable, inject } from 'inversify';
 import { Disposable, Event, EventEmitter, NotebookDocument } from 'vscode';
 import { disposeAllDisposables } from '../common/helpers';
 import { traceInfo } from '../common/logger';
-import { IDisposable, IDisposableRegistry } from '../common/types';
+import { IDisposable, IDisposableRegistry, IExtensions } from '../common/types';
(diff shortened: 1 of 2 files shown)


src/client/api/kernelApi.ts
```diff
@@ -35,10 +35,12 @@ export class JupyterKernelServiceFactory {
         @inject(IKernelProvider) private readonly kernelProvider: IKernelProvider,
         @inject(IDisposableRegistry) private readonly disposables: IDisposableRegistry,
         @inject(INotebookControllerManager) private readonly notebookControllerManager: INotebookControllerManager,
-        @inject(ApiAccessService) private readonly apiAccess: ApiAccessService
+        @inject(ApiAccessService) private readonly apiAccess: ApiAccessService,
+        @inject(IExtensions) private readonly extensions: IExtensions
     ) {}
     public async getService() {
-        const accessInfo = await this.chainedApiAccess.chainFinally(() => this.apiAccess.getAccessInformation());
+        const info = await this.extensions.determineExtensionFromCallStack();
```
rchiodo, line 42: This has to happen before the 'chainedApiAccess' promise or the callstack doesn't return anything.

DonJayamanne (approved):

PR #8781, opened 2022-01-24
Support ipywidgets when restarting remote kernels

DonJayamanne:
For #8378 #8431

Looks like the previous fix never worked (i can't get it work at all without the changes in this PR).
Basically when we restart the remote kernel, wen end up with a new websocket connection and this is never used.

Hence this PR is necessary to fix this & the old fix never worked.
It was closed incorrectly (by me), user had validated a different issue. 

THis is required for this release as we have a new entry for this, either we take this PR or remove the new entry for #8378

### DIFF
--- a/src/client/datascience/baseJupyterSession.ts
+++ b/src/client/datascience/baseJupyterSession.ts
@@ -171,6 +171,7 @@ export abstract class BaseJupyterSession implements IJupyterSession {
     public async restart(): Promise<void> {
         if (this.session?.isRemoteSession && this.session.kernel) {
             await this.session.kernel.restart();
+            this.setSession(this.session, true);
             return;
         }
 
@@ -342,7 +343,7 @@ export abstract class BaseJupyterSession implements IJupyterSession {
     }
 
     // Changes the current session.
-    protected setSession(session: ISessionWithSocket | undefined) {
+    protected setSession(session: ISessionWithSocket | undefined, forceUpdateKernelSocketInfo: boolean = false) {
         const oldSession = this._session;
         if (oldSession) {
             if (this.ioPubHandler) {
@@ -367,7 +368,7 @@ export abstract class BaseJupyterSession implements IJupyterSession {
                 session.unhandledMessage.connect(this.unhandledMessageHandler);
             }
             // If we have a new session, then emit the new kernel connection information.
-            if (oldSession !== session && session.kernel) {
+            if ((forceUpdateKernelSocketInfo || oldSession !== session) && session.kernel) {
                 this._kernelSocket.next({
                     options: {
                         clientId: session.kernel.clientId,
--- a/src/client/datascience/jupyter/jupyterWebSocket.ts
+++ b/src/client/datascience/jupyter/jupyterWebSocket.ts
@@ -49,15 +49,19 @@ export function createJupyterWebSocket(cookieString?: string, allowUnauthorized?
             if (this.kernelId) {
                 JupyterWebSockets.set(this.kernelId, this);
                 this.on('close', () => {
-                    clearInterval(this.timer as any);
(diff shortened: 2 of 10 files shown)


src/client/datascience/baseJupyterSession.ts
```diff
@@ -171,6 +171,7 @@ export abstract class BaseJupyterSession implements IJupyterSession {
     public async restart(): Promise<void> {
         if (this.session?.isRemoteSession && this.session.kernel) {
             await this.session.kernel.restart();
+            this.setSession(this.session, true);
```
DonJayamanne, line 174: A new websocket is created, we need to ensure this new websocket connection is used (new argument)

src/datascience-ui/ipywidgets/kernel/index.ts
```diff
@@ -159,19 +156,25 @@ function renderIPyWidget(
 let widgetManagerPromise: Promise<WidgetManager> | undefined;
 async function getWidgetManager(): Promise<WidgetManager> {
     if (!widgetManagerPromise) {
-        // eslint-disable-next-line , @typescript-eslint/no-explicit-any
-        widgetManagerPromise = new Promise((resolve) => WidgetManager.instance.subscribe(resolve as any));
-        widgetManagerPromise
-            .then((wm) => {
+        function reInitializeWidgetManager(resolve?: (value: WidgetManager) => void) {
```
DonJayamanne, line 159: We could end up with new widget managers, the subscription should listen to new changes and re-initialize the promise.

src/client/datascience/jupyter/jupyterWebSocket.ts
```diff
@@ -49,15 +49,19 @@ export function createJupyterWebSocket(cookieString?: string, allowUnauthorized?
             if (this.kernelId) {
                 JupyterWebSockets.set(this.kernelId, this);
                 this.on('close', () => {
-                    clearInterval(this.timer as any);
-                    JupyterWebSockets.delete(this.kernelId!);
+                    if (this.timer !== timer){
+                        clearInterval(timer as any);
+                    }
+                    if (JupyterWebSockets.get(this.kernelId!) === this){
+                        JupyterWebSockets.delete(this.kernelId!);
```
DonJayamanne, line 56: Previously we were deleting the wrong object.

src/test/datascience/widgets/standard.vscode.test.ts
```diff
@@ -177,12 +177,12 @@ suite('Standard IPyWidget (Execution) (slow) (WIDGET_TEST)', function () {
         await kernel.interrupt();
         await testSliderWidget(comms);
 
-        // Clear all cells and restart and test again.
-        await kernel.interrupt();
-        const cell = vscodeNotebook.activeNotebookEditor?.document.cellAt(0)!;
-        await commands.executeCommand('notebook.clearAllCellsOutputs');
-        await waitForCondition(async () => cell.outputs.length === 0, 5_000, 'Cell did not get cleared');
+        // // Clear all cells and restart and test again.
```
rchiodo, line 180: Does this part of the test still not work?

DonJayamanne: looks like tests are failing, i will disabled these again

src/datascience-ui/ipywidgets/kernel/index.ts
rchiodo: The function seems weird? At least made it harder for me to parse

Would this be easier to understand?

```
widgetManagerPromise = Promise.resolve((resolve) => WidgetManager.instance.subscribe((wm) => {
   // your current contents here
}
```

rchiodo (approved):

Issue #8678, opened 2022-01-12
Pylance errors in notebook after editing for over an hour.


rchiodo:
@rchiodo I don't know why my Pylance give me a lots of errors, but the cell and whole notebook I can run without any issue (no errors). Here is list of founded errors: 
  * Unindent not expected
  * Unexpected indentation
  * Expected expression
  * "(" was not closed
  * and so on ...
 

Thanks.

_Originally posted by @markub3327 in https://github.com/microsoft/vscode-jupyter/issues/4307#issuecomment-1011474765_

rchiodo:
@markub3327, I opened this based on your comment on #4307.

Can you provide:
- VS code version
- Python Extension version
- Pylance Extension version
- Jupyter Extension version

And a screenshot of the errors?

Thanks.

markub3327:
Hi,
 
 * OS: Darwin arm64 21.2.0 
 * VS code version 1.63.2 (Universal)
 * Python Extension v2021.12.1559732655
 * Pylance Extension v2022.1.0
 * Jupyter Extension v2021.11.1001550889

Sorry, after the restart of VSCode all errors were out. I tried to provide you screenshot as fast as possible. I think the errors are shown for a long time opened VSCode, because I don't close it for a few days and don't restart my macOS too.

markub3327:
Now I get a warnings about variable that not existed and variable that is correctly defined. Both are marked as undefined variable.

Here is the screenshot
<img width="1109" alt="Snímka obrazovky 2022-01-13 o 12 58 41" src="https://user-images.githubusercontent.com/74611856/149326768-8c7bf958-dafb-4d2d-8713-a6ab20b6362e.png">
:

markub3327:
@rchiodo Now, I get what we are talking about. There are three errors, but my notebook can be run successfully (can see check mark under the cell and the output too). This issue appeared after the approximately 1 hour.

Here is the screenshot:
<img width="1115" alt="Snímka obrazovky 2022-01-13 o 13 05 09" src="https://user-images.githubusercontent.com/74611856/149327613-a985ec54-3bfa-4349-be3c-ebbc913e1a8c.png">

rchiodo:
Thanks that helps.

Can you include your entire notebook here so I can try on my machine?

Do the errors appear only after editing a cell?

Does it require you edit as soon as the notebook opens? There might be a timing related problem with the changes.

markub3327:
[...]

PR #8787, opened 2022-01-24
Add tool for replaying pylance logs

rchiodo:
For #8678

The LogReplayService (only available in dev mode) allows you to playback the debug output of pylance intellisense and have it generate the same edits to a notebook.

I used this to investigate #8678.

### DIFF
--- a/package.json
+++ b/package.json
@@ -278,6 +278,7 @@
                 "command": "jupyter.replayPylanceLogStep",
                 "title": "Step Pylance Log",
                 "category": "Jupyter (Dev)",
+                "icon": "$(debug-start)",
                 "enablement": "jupyter.development && notebookType == jupyter-notebook && isWorkspaceTrusted && jupyter.replayLogLoaded"
             },
             {
@@ -1043,11 +1044,6 @@
                     "command": "jupyter.replayPylanceLog",
                     "title": "Replay Pylance Log",
                     "when": "jupyter.development && isWorkspaceTrusted"
-                },
-                {
-                    "command": "jupyter.debugNotebook",
-                    "title": "%jupyter.command.jupyter.debug.title%",
-                    "when": "false"
                 },
                 {
                     "command": "jupyter.debugNotebook",
--- a/package.json
+++ b/package.json
@@ -269,6 +269,18 @@
                 "enablement": "jupyter.development"
             },
             {
+                "command": "jupyter.replayPylanceLog",
+                "title": "Replay Pylance Log",
+                "category": "Jupyter (Dev)",
+                "enablement": "jupyter.development && notebookType == jupyter-notebook && isWorkspaceTrusted"
+            },
+            {
+                "command": "jupyter.replayPylanceLogStep",
+                "title": "Step Pylance Log",
+                "category": "Jupyter (Dev)",
+                "enablement": "jupyter.development && notebookType == jupyter-notebook && isWorkspaceTrusted && jupyter.replayLogLoaded"
+            },
+            {
                 "command": "jupyter.debugNotebook",
                 "title": "%jupyter.command.jupyter.debug.title%",
                 "icon": "$(bug)",
@@ -915,6 +927,11 @@
                     "command": "jupyter.notebookeditor.export",
                     "group": "Jupyter",
                     "when": "notebookType == 'jupyter-notebook' && isWorkspaceTrusted && jupyter.ispythonnotebook"
+                },
+                {
+                    "command": "jupyter.replayPylanceLogStep",
+                    "group": "navigation@1",
+                    "when": "notebookType == 'jupyter-notebook' && isWorkspaceTrusted && jupyter.replayLogLoaded"
                 }
             ],
             "notebook/cell/title": [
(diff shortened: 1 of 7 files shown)


src/client/datascience/notebook/intellisense/logReplayService.ts
```diff
@@ -0,0 +1,287 @@
...
+import { Commands, EditorContexts } from '../../constants';
+import * as lspConcat from '@vscode/lsp-notebook-concat';
+import { NOTEBOOK_SELECTOR, PYTHON_LANGUAGE } from '../../../common/constants';
+import * as protocol from 'vscode-languageserver-protocol';
+import { traceInfo } from '../../../common/logger';
+import { ContextKey } from '../../../common/contextKey';
+import { sleep, waitForCondition } from '../../../common/utils/async';
+
+/**
+ * Class used to replay pylance log output to regenerate a series of edits. SHIFT+F10 will play the steps one at a time
```
rchiodo, line 21: Ah this keybinding isn't there anymore. Need to update this comment.

rchiodo:
You use this by running the 
image

and then you click on the 'Step Pylance Log' button on the toolbar

image

src/client/datascience/serviceRegistry.ts
```diff
@@ -231,6 +232,9 @@ export function registerTypes(serviceManager: IServiceManager, inNotebookApiExpe
     serviceManager.addSingleton<IExtensionSingleActivationService>(IExtensionSingleActivationService, NotebookUsageTracker);
     serviceManager.addSingleton<IExtensionSingleActivationService>(IExtensionSingleActivationService, MigrateJupyterInterpreterStateService);
     serviceManager.addSingleton<IExtensionSingleActivationService>(IExtensionSingleActivationService, VariableViewActivationService);
+    if (isDevMode) {
+        serviceManager.addSingleton<IExtensionSingleActivationService>(IExtensionSingleActivationService, LogReplayService);
```
rchiodo, line 236: You'll notice it won't even be a part of any release bits.

DonJayamanne, line 236: Would it be better to add `IsDevMode` to the DI container as a constant.
Then in `activate` you can check the value and just exit instead of adding the command handlers?
That way we won't have to pass this around in future blocks of code.

src/client/datascience/notebook/intellisense/logReplayService.ts
```diff
@@ -0,0 +1,287 @@
...
+        this.disposableRegistry.push(
+            this.commandService.registerCommand(Commands.ReplayPylanceLogStep, this.step, this)
+        );
+        this.isLogActive = new ContextKey(EditorContexts.ReplayLogLoaded, this.commandService);
+        void this.isLogActive.set(false);
+    }
+
+    private async replayPylanceLog() {
+        if (vscode.window.activeNotebookEditor) {
+            const file = await this.appShell.showOpenDialog({ title: 'Open Pylance Output Log' });
```
DonJayamanne, line 54: Silly question, but I'm assuming this means we can take the raw output from the output panel.
I.e. I can just copy the output and paste into a file, correct?

package.json
```diff
@@ -915,6 +927,11 @@
                     "command": "jupyter.notebookeditor.export",
                     "group": "Jupyter",
                     "when": "notebookType == 'jupyter-notebook' && isWorkspaceTrusted && jupyter.ispythonnotebook"
+                },
+                {
+                    "command": "jupyter.replayPylanceLogStep",
+                    "group": "navigation@1",
```
DonJayamanne, line 933: Don't we need an icon for this?

package.json
```diff
@@ -1022,6 +1039,16 @@
...
             ],
             "commandPalette": [
+                {
+                    "command": "jupyter.replayPylanceLog",
+                    "title": "Replay Pylance Log",
+                    "when": "jupyter.development && isWorkspaceTrusted"
+                },
+                {
+                    "command": "jupyter.debugNotebook",
+                    "title": "%jupyter.command.jupyter.debug.title%",
```
DonJayamanne, line 1049: This is a duplicate (already have this in line 1052)

src/client/datascience/serviceRegistry.ts
rchiodo: I had that originally but thought hey I could skip the entire class being loaded. Then it doesn't slow down activate at all.

src/client/datascience/notebook/intellisense/logReplayService.ts
rchiodo: Yep. That's how it works.

package.json
rchiodo: Nah, it's just for dev. It shows the text. Maybe if it didn't fit?

src/client/datascience/notebook/intellisense/logReplayService.ts
DonJayamanne: Is there something we can do to ensure the format of the logs don't change without updating this code?
I'm concerned that once we have this stuff working for a while, then when we do want to use it, the format will be out of sync.

src/client/datascience/serviceRegistry.ts
DonJayamanne: Not following you, do you mean to say this will not be bundled into the extension?
I thought it would

DonJayamanne:
Great idea.

src/client/datascience/notebook/intellisense/logReplayService.ts
DonJayamanne: Doesn't matter, its just for us (dev tool).

src/client/datascience/serviceRegistry.ts
DonJayamanne: Doesn't matter, dev only.

DonJayamanne (approved):
Please remove the duplicate `jupyter.debugNotebook` command, else all good.

PR #8792, opened 2022-01-25
Fix tabbing through all levels for completions in a path

### LINKED ISSUE #7816: IntelliSense for data files does not work on subfolders
When trying to read data files via `df = pd.read_csv()`, it appears that intellisense does not show subfolders and I have to explicitly trigger intellisense again after there are no matches. It would be nice if I didn't have to explicitly trigger this and it just worked automatically.

https://user-images.githubusercontent.com/35271042/136071529-1149082d-fd54-4e4b-b9ce-58ea65373b09.mp4

rchiodo:
Fixes #7816 

Thanks @misolori. I used the html extension and figured out what it was returning differently. Its completion items include a command to run when they are completed.

### DIFF
--- a/src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
+++ b/src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
@@ -311,6 +311,20 @@ export function filterCompletions(
     // If not inside of a string, filter out file names (things with a '.' in them or end with '/')
     if (!insideString) {
         result = result.filter((r) => !r.itemText.includes('.') && !r.itemText.endsWith('/'));
+    } else {
+        // If inside a string and ending with '/', then add a command to force a suggestion right after
+        result = result.map((r) => {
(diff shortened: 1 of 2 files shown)


src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
```diff
@@ -311,6 +311,20 @@ export function filterCompletions(
     // If not inside of a string, filter out file names (things with a '.' in them or end with '/')
     if (!insideString) {
         result = result.filter((r) => !r.itemText.includes('.') && !r.itemText.endsWith('/'));
+    } else {
+        // If inside a string and ending with '/', then add a command to force a suggestion right after
+        result = result.map((r) => {
+            if (r.itemText.endsWith('/')) {
```
DonJayamanne, line 317: What happens if a path ends with `\`, or is this specific to just `/`

PR #8808, opened 2022-01-25
Fixes to determining calling ext from call stack

DonJayamanne:


package.json
```diff
@@ -1996,6 +1996,13 @@
             }
         ],
         "viewsContainers": {
+            "activitybar": [
+                {
+                    "id": "jupyter",
```
DonJayamanne, line 2001: Always available

src/client/common/application/extensions.ts
```diff
@@ -30,32 +30,33 @@ export class Extensions implements IExtensions {
     public async determineExtensionFromCallStack(): Promise<{ extensionId: string; displayName: string }> {
         const stack = new Error().stack;
         if (stack) {
-            const root = EXTENSION_ROOT_DIR.toLowerCase();
-            const frames = stack.split('\n').map((f) => {
-                const result = /\((.*)\)/.exec(f);
-                if (result) {
-                    return result[1].toLowerCase();
```
DonJayamanne, line 37: This would not work on unix

rchiodo: For some cases. It looks like it won't work with this change either. jupyterExtRoot is still lower case and you're not lowercasing the results.

rchiodo: Wait did I miss something? It looks like it does check lower case on the filter.

rchiodo (approved):

PR #8820, opened 2022-01-26
Update prereleaseChecker.ts

rchiodo:
Fix typo so it actually asks. This is why nobody complained about the prerelease check

### DIFF
--- a/src/client/datascience/prereleaseChecker.ts
+++ b/src/client/datascience/prereleaseChecker.ts
@@ -28,7 +28,7 @@ export class PreReleaseChecker implements IExtensionSingleActivationService {
                 const dontAsk = this.globalState.get(PRERELEASE_DONT_ASK_FLAG, false);
 
                 // Ask user if the version is not prerelease
-                if (!isPreRelease && dontAsk && this.appEnv.channel === 'insiders') {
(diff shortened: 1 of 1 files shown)


src/client/datascience/prereleaseChecker.ts
```diff
@@ -28,7 +28,7 @@ export class PreReleaseChecker implements IExtensionSingleActivationService {
                 const dontAsk = this.globalState.get(PRERELEASE_DONT_ASK_FLAG, false);
 
                 // Ask user if the version is not prerelease
-                if (!isPreRelease && dontAsk && this.appEnv.channel === 'insiders') {
+                if (!isPreRelease && !dontAsk && this.appEnv.channel === 'insiders') {
```
rchiodo, line 31: Default is false. So it would never ask. Doh.

DonJayamanne (approved):

amunger (approved):

src/client/datascience/prereleaseChecker.ts
IanMatthewHuff: The old "negative embedded in the variable name" trick ;)

IanMatthewHuff (approved):

Issue #8839, opened 2022-01-28
Links for errors in the interactive window and notebooks should be handled by VS Code core


rchiodo:
See this bug here:
https://github.com/microsoft/vscode/issues/141757#issue-1117797222

Basically if you click on a 'file' link in the interactive window both our code and VS code core handles the click. 

We could switch to VS code handling it, but they don't do the same thing we did before.

rchiodo:
@rebornix undid the change to VS code that caused this. Leaving open for now so we use this to decide what to do in the longer term.

PR #8840, opened 2022-01-28
Fix link clicks

rchiodo:
Fixes https://github.com/microsoft/vscode-jupyter/issues/8839

VS code was modified to support opening file links from a webview. 

However it doesn't give focus to the same file if it's already open (which we used to do) and doesn't allow for a line number query (which we used to do).

And since we have our own link handler, it was opening twice. This opening twice breaks the IW (because you can't run a cell if the same file is open twice).

This change creates a new scheme that works around this problem - essentially going back to our link click handling.

### DIFF
--- a/news/2 Fixes/8839.md
+++ b/news/2 Fixes/8839.md
@@ -1,0 +1,1 @@
+Fixes clicking links in error tracebacks opening two copies of a file.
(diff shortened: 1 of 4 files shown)


DonJayamanne (approved):

rchiodo:
@rebornix is going to back out the change to VS code. We'll discuss a longer term solution next week.

PR #8878, opened 2022-01-30
Fixes to exposed API

DonJayamanne:


src/client/api/kernelApi.ts
```diff
@@ -103,7 +103,7 @@ class JupyterKernelService implements IExportedKernelService {
         });
         await this.notebookControllerManager.loadNotebookControllers();
         const items = await this.notebookControllerManager.kernelConnections;
-        return items.map((item) => this.translateKernelConnectionMetataToExportedType(item));
+        return items.map((item) => this.translateKernelConnectionMetadataToExportedType(item));
```
DonJayamanne, line 106: Fixed typo

src/client/datascience/baseJupyterSession.ts
```diff
@@ -104,38 +104,14 @@ export abstract class BaseJupyterSession implements IJupyterSession {
...
-    public dispose(): Promise<void> {
-        return this.shutdown();
+    public async dispose(): Promise<void> {
+        await this.shutdownImplementation(false);
     }
     // Abstracts for each Session type to implement
     public abstract waitForIdle(timeout: number): Promise<void>;
 
     public async shutdown(): Promise<void> {
-        this._isDisposed = true;
```
DonJayamanne, line 114: As per Jupyter API shutdown should shutdown,
We aren't shutting things down and this method isn't used in our code, hence changing the code to shutdown the kernel when calling `shutdown` (including remote kernels).
This way 3rd party extensions using this API will get the desired behavior.

FYI - We call the dispose method to kill local kernels & to leave remote kernels running (i.e. we only ever call `dispose` hence our code isn't affected by this change)

src/client/datascience/baseJupyterSession.ts
```diff
@@ -384,7 +360,8 @@ export abstract class BaseJupyterSession implements IJupyterSession {
     protected async shutdownSession(
         session: ISessionWithSocket | undefined,
         statusHandler: Slot<ISessionWithSocket, KernelMessage.Status> | undefined,
-        isRequestToShutDownRestartSession: boolean | undefined
+        isRequestToShutDownRestartSession: boolean | undefined,
+        shutdownEvenIfRemote?: boolean
```
DonJayamanne, line 364: New argument to ensure we shutdown remote sessions

src/client/datascience/baseJupyterSession.ts
```diff
@@ -104,38 +104,14 @@ export abstract class BaseJupyterSession implements IJupyterSession {
...
-            this.setSession(undefined);
-            this.restartSessionPromise = undefined;
-            this.onStatusChangedEvent.fire('dead');
-            this._disposed.fire();
-            this._disposed.dispose();
-            this.onStatusChangedEvent.dispose();
-        }
-        disposeAllDisposables(this.disposables);
-        traceVerbose('Shutdown session -- complete');
+        await this.shutdownImplementation(true);
```
DonJayamanne, line 114: New method with an argument to ensure remote kernels are shutdown when calling shutdown method

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -191,11 +192,13 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         }
     }
 
-    private async createEditorReadyPromise(): Promise<NotebookEditor> {
-        const preferredController = await this.notebookControllerManager.getActiveInterpreterOrDefaultController(
-            InteractiveWindowView,
-            this.owner
-        );
+    private async createEditorReadyPromise(connection?: KernelConnectionMetadata): Promise<NotebookEditor> {
```
DonJayamanne, line 195: Added ability to connect to an existing kernel connection from IW

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -185,10 +185,10 @@ function getOldFormatDisplayNameOrNameOfKernelConnection(kernelConnection: Kerne
             ? kernelConnection.kernelModel.name
             : kernelConnection.kernelSpec?.name;
 
-    const interpeterName =
+    const interpreterName =
         kernelConnection.kind === 'startUsingPythonInterpreter' ? kernelConnection.interpreter.displayName : undefined;
 
-    return displayName || name || interpeterName || '';
+    return displayName || name || interpreterName || '';
```
DonJayamanne, line 191: typo

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -301,10 +301,10 @@ export class Kernel implements IKernel {
             if (this.notebook) {
                 promises.push(this.notebook.session.dispose().catch(noop));
                 this.notebook = undefined;
-                this._disposed = true;
-                this._onDisposed.fire();
-                this._onStatusChanged.fire('dead');
             }
+            this._disposed = true;
+            this._onDisposed.fire();
+            this._onStatusChanged.fire('dead');
```
DonJayamanne, line 307: These events were not getting fired, when they should have been

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
```diff
@@ -44,84 +44,145 @@ export class KernelConnectionWrapper implements Kernel.IKernelConnection {
...
     public readonly anyMessage = new Signal<this, Kernel.IAnyMessageArgs>(this);
     public get serverSettings() {
-        return this.kernelConnection.serverSettings;
+        return (this.kernelConnection || this._previousKernelConnection).serverSettings;
     }
     public readonly disposed = new Signal<this, void>(this);
+    private _kernelConnection!: Kernel.IKernelConnection;
+    private _previousKernelConnection!: Kernel.IKernelConnection;
+    // private _isRestarting?: boolean;
+    private get kernelConnection(): undefined | Kernel.IKernelConnection {
```
DonJayamanne, line 53: When restarting a kernel, the `kernel.IKernelConnection` gets changed, hence we cannot store this in a variable, as it changes.

Also, when a kernel is shutdown, this could be null.

This makes the code work as expected (when used by 3rd party extensions) when restarting kernels or shutting them down.

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
```diff
@@ -44,84 +44,145 @@ export class KernelConnectionWrapper implements Kernel.IKernelConnection {
...
+            throw new Error(
+                `Kernel connection is not available, status = ${this.status}, connection = ${this.connectionStatus}`
+            );
+        }
+        return this.kernelConnection;
+    }
 
-    constructor(
-        readonly kernel: IKernel,
-        private kernelConnection: Kernel.IKernelConnection,
```
DonJayamanne, line 53: We cannot pass `kernelConnection` as it changes when restarting kernels, its not a constant

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
```diff
@@ -44,84 +44,124 @@ export class KernelConnectionWrapper implements Kernel.IKernelConnection {
...
     }
     public readonly disposed = new Signal<this, void>(this);
+    private _kernelConnection!: Kernel.IKernelConnection;
+    private _previousKernelConnection!: Kernel.IKernelConnection;
+    // private _isRestarting?: boolean;
+    private get kernelConnection(): undefined | Kernel.IKernelConnection {
+        if (this.kernel.session?.kernel === this._kernelConnection) {
+            return this._kernelConnection;
+        }
+        this.stopHandlingKernelMessages(this._kernelConnection);
```
DonJayamanne, line 57: If the previous kernel connection isn't the same as the current, then kernel was restarted and we have a new connection.
Hence remove event handlers associated with the previous connections,

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
```diff
@@ -147,49 +187,55 @@ export class KernelConnectionWrapper implements Kernel.IKernelConnection {
...
-        return this.kernelConnection.sendInputReply(content);
+        return this.getKernelConnection().sendInputReply(content);
     }
     createComm(targetName: string, commId?: string): Kernel.IComm {
-        return this.kernelConnection.createComm(targetName, commId);
+        return this.getKernelConnection().createComm(targetName, commId);
     }
     hasComm(commId: string): boolean {
-        return this.kernelConnection.hasComm(commId);
+        return this.getKernelConnection().hasComm(commId);
```
DonJayamanne, line 205: If the kernel has been shutdown we need to handle such cases

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
```diff
@@ -147,49 +187,55 @@ export class KernelConnectionWrapper implements Kernel.IKernelConnection {
...
-    shutdown(): Promise<void> {
-        return this.kernel.dispose();
+        return this.getKernelConnection().removeMessageHook(msgId, hook);
+    }
+    async shutdown(): Promise<void> {
+        if (
+            this.kernel.kernelConnectionMetadata.kind === 'startUsingRemoteKernelSpec' ||
+            this.kernel.kernelConnectionMetadata.kind === 'connectToLiveKernel'
+        ) {
+            await this.kernel.session?.shutdown();
```
DonJayamanne, line 236: When shutting down a kernel, we must explicitly call `session.shutdown` for remote kernels.
As `shutdown` means kill kernel & in our code base `IKernel.dispose` will never shutdown the remote kernels.

DonJayamanne: This change ensures 3rd party extensions get the desired effect as calling shutdown should ideally shutdown the kernel whether its local or remote.

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -79,14 +79,16 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
     // Listing of the controllers that we have registered
     private registeredControllers = new Map<string, VSCodeNotebookController>();
     private selectedControllers = new Map<string, VSCodeNotebookController>();
-    private readonly allKernelConnections = new Set<KernelConnectionMetadata>();
```
DonJayamanne, line 82: Remove `allKernelConnections` this was a duplicate of the values in `registeredControllers`.
and would go out of sync, hence changed to always derive the values from `registeredControllers`

src/client/api/extension.d.ts
```diff
@@ -257,8 +257,9 @@ export interface IExportedKernelService {
      * Gets a list of all kernel specifications that can be used to start a new kernel or to connect to an existing kernel.
      * Local, remote kernels are returned, including Python interpreters that
      * are treated as kernelspecs (as we can start Kernels for Python interpreters without Jupyter).
+     * @param {boolean} [refresh] Whether to force a refresh of the kernel specs (useful when dealing with remote kernels, as that list can change).
      */
-    getKernelSpecifications(): Promise<KernelConnectionMetadata[]>;
+    getKernelSpecifications(refresh?: boolean): Promise<KernelConnectionMetadata[]>;
```
DonJayamanne, line 262: this allows us to refresh and view list of remote kernels.
Else today we only refresh the remote kernel when we start a new notebook.
Useful for external extensions to get latest remote kernels.

src/client/api/kernelApi.ts
```diff
@@ -96,26 +96,28 @@ class JupyterKernelService implements IExportedKernelService {
...
             pemUsed: 'getActiveKernels'
         });
-        return this.kernelProvider.kernels.map((item) => {
-            return {
-                metadata: this.translateKernelConnectionMetataToExportedType(item.kernelConnectionMetadata),
-                notebook: item.notebookDocument
-            };
-        });
+        return this.kernelProvider.kernels
+            .filter((item) => item.startedAtLeastOnce)
```
DonJayamanne, line 114: See comments in the interface

DonJayamanne: Here we should not return kernel objects that have not started
E.g. when opening a notebook we create a kernel and assign it to the notebook, however the kernel is not active, hence no point returning such kernels to the 3rd party extensions.

src/client/api/kernelApi.ts
```diff
@@ -182,13 +184,13 @@ class JupyterKernelService implements IExportedKernelService {
...
 
-        const connection = new KernelConnectionWrapper(kernel, kernel.session!.kernel!, this.disposables);
+        const connection = new KernelConnectionWrapper(kernel, this.disposables);
         // eslint-disable-next-line @typescript-eslint/no-use-before-define
         const info = { connection, kernelSocket: new KernelSocketWrapper(kernel) };
         JupyterKernelService.wrappedKernelConnections.set(kernel, info);
         return info;
     }
-    private translateKernelConnectionMetataToExportedType(
+    private translateKernelConnectionMetadataToExportedType(
```
DonJayamanne, line 193: typo

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -697,7 +717,6 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
                     // We are disposing as documents are closed, but do this as well
                     this.disposables.push(controller);
                     this.registeredControllers.set(controller.id, controller);
-                    this.allKernelConnections.add(kernelConnection);
```
DonJayamanne, line 700: Removed this property, pointless maintaining state in two properties (buggy & unnecessary)

src/client/datascience/interactive-window/interactiveWindowProvider.ts
```diff
@@ -196,18 +201,18 @@ export class InteractiveWindowProvider implements IInteractiveWindowProvider, IA
...
 
         // Otherwise match the owner.
         return this._windows.find((w) => {
-            if (!owner && !w.owner) {
+            if (!owner && !w.owner && !connection) {
                 return true;
             }
             if (owner && w.owner && this.fs.areLocalPathsSame(owner.fsPath, w.owner.fsPath)) {
-                return true;
+                return connection ? w.originalConnection?.id === connection.id : true;
```
amunger, line 215: alternatively:
`!connection || w.originalConnection?.id === connection.id`

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
```diff
@@ -44,84 +44,124 @@ export class KernelConnectionWrapper implements Kernel.IKernelConnection {
     public readonly unhandledMessage = new Signal<this, IMessage<MessageType>>(this);
     public readonly anyMessage = new Signal<this, Kernel.IAnyMessageArgs>(this);
     public get serverSettings() {
-        return this.kernelConnection.serverSettings;
+        return (this.kernelConnection || this._previousKernelConnection).serverSettings;
     }
     public readonly disposed = new Signal<this, void>(this);
+    private _kernelConnection!: Kernel.IKernelConnection;
```
amunger, line 50: A comment for this variable saying its possibly out of date would be useful since it has the same accessiblity as the getter that should be used instead

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
```diff
@@ -223,7 +271,7 @@ export class KernelConnectionWrapper implements Kernel.IKernelConnection {
         this.startHandleKernelMessages(this.kernel.session?.kernel);
     }
     private startHandleKernelMessages(kernelConnection: Kernel.IKernelConnection) {
-        this.kernelConnection = kernelConnection;
+        this._kernelConnection = kernelConnection;
```
amunger, line 274: set _previousKernelConnection first? in case it changes more than once

src/client/datascience/notebook/notebookControllerManager.ts
```diff
@@ -152,6 +158,16 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         }
     }
 
+    public getControllerForConnection(
+        connection: KernelConnectionMetadata,
+        notebookType: typeof JupyterNotebookView | typeof InteractiveWindowView
+    ) {
+        const id =
+            notebookType === 'jupyter-notebook'
```
amunger, line 166: can you just use the JupyterNotebookView constant here?

src/client/datascience/baseJupyterSession.ts
```diff
@@ -432,7 +440,12 @@ export abstract class BaseJupyterSession implements IJupyterSession {
             return true;
         }
         // If we're in notebooks and using Remote Jupyter connections, then never shutdown the sessions.
```
amunger, line 442: out of date comment. But this also feels like a dangerous change if we're suddenly shutting down external jupyter sessions that we weren't previously

amunger: sounds like a pretty explicit call to shutdown the session, so never mind - probably fine to change the behavior

src/client/datascience/interactive-window/interactiveWindowProvider.ts
DonJayamanne: Personally find this more readable when going with `!connection`, i find reading negations a little difficult.

src/client/datascience/notebook/notebookControllerManager.ts
DonJayamanne: Not required, as this s a strongly typed string value, hence there's no way to get a typo here.
i.e. if we make a typo here or change the constant then tsc compiler will fall over.

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
DonJayamanne: Updated the code to ensure `_previousKernelConnection` never undefined and readonly (as its initialized in the ctor)

amunger: but it is _only_ set in the constructor, so if it changes more than once, it will no longer be the previous kernel connection. Should the variable be renamed to _initialKernelConnection?

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
```diff
@@ -44,84 +44,128 @@ export class KernelConnectionWrapper implements Kernel.IKernelConnection {
...
+        if (this.kernel.session?.kernel === this._kernelConnection) {
+            return this._kernelConnection;
+        }
+        this.stopHandlingKernelMessages(this._kernelConnection);
+        if (this.kernel.session?.kernel) {
+            this.startHandleKernelMessages(this.kernel.session.kernel);
+            return this._kernelConnection;
+        }
+    }
+    private getKernelConnection(): Kernel.IKernelConnection {
```
rchiodo, line 67: What's the difference between this and just ```kernelConnection()```. Seems weird to have both?

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
rchiodo: Not sure why on restart we don't just update ```_kernelConnection``` during a restart?

DonJayamanne: Thats exactly what we do today.
This is required, because while restart is taking place the underlying object `kernel.session.kernel` could have already changed. Hence even before restart has completed we could end up with a whole new kernel connection, so any code running here will end up accessing the old item and things could fall over (basically thats what i found while debugging).

Hence the check to see if `kernel.session.kernel !== _kernelConnection`.

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
DonJayamanne: This will throw an exception if kernelConnection is undefined, and when using the accessor `this.kernelConnection` we don't want exceptions to be thrown.

I.e. the property kernelConnection can be undefined. however the return value of getKernelConnection can never be undefined.

rchiodo: Still seems weird. Maybe name them something different? Like ```possibleKernelConnection``` and ```kernelConnection```?

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
rchiodo: Sorry so the use case is:
- User runs a kernel
- 3rd party uses kernel
- User restarts
- During restart 3rd party accesses the IKernelConnection and it's undefined?
- Crash?

Can we not just have a restart event on the kernel? It seems like the 3rd party is using the kernel when it shouldn't be? It's the same scenario if they try using a kernel before one is started.

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
```diff
@@ -128,13 +128,13 @@ export class KernelConnectionWrapper implements Kernel.IKernelConnection {
         return this.kernelConnection?.connectionStatus || 'disconnected';
     }
     public get info(): Promise<IInfoReply> {
-        return (this.kernelConnection || this._previousKernelConnection).info;
+        return (this.kernelConnection || this._initialKernelConnection).info;
```
rchiodo, line 131: It seems weird to me to make sure these don't crash but that they might return incorrect information. The _initialKernelConnection may be long dead when the caller makes this request.

rchiodo: Is there a way we can make the 3rd party api more robust/obvious? Maybe just have it return undefined when the restart is going on until it's back up?

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
DonJayamanne: > During restart 3rd party accesses the IKernelConnection and it's undefined?

At this point it could have already been updated even before the whole restart has completed (i.e. our extension restart has not yet completed, the async code is still running).

> It's the same scenario if they try using a kernel before one is started.

Basically we're hooked into events and the like and all of those event handlers (for getting data from the kernel would be out dated).
3rd party doesn't even know that a kernel is restarting all they're doing is accesing a kernel, they're not necessarily restarting the kernel (the user could be restarting the kernel using the jupyter extension)

> It seems like the 3rd party is using the kernel when it shouldn't be?

Not true, they don't know its restarting.
E.g. i could write code to get kernel events or widget messages, but if in the middle of all of that the kernel is restarted, then tha'ts when things fall over (as the 3rd party extension code is accessing old kernle when its no longer valid).
The problem is we're still exposing the old kernel when its not valid.

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
DonJayamanne: > _initialKernelConnection may be long dead when the caller makes this request.

Yes, but the info never changes, hence I'm retruning a cached info. The info is basically a kernel informaiton (version etc).

> Maybe just have it return undefined when the restart is going on until it's back up?

No, based on the typings it cannot be undefined. Hence the reason to cache it.

rchiodo: I think we should discuss this at standup.

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
DonJayamanne: Will rename to `getPossibleKernelConnection`
The problem is, e.g. when calling `executeRequest` at that point a kernel connection is required, and we cannot call it without a kernel conneciton.
If 3rd part calls it in such a state (disposed state, then we must handle it that case and throw an exception)
Hence the check to see if its undefined & throw an error.

I could optionally rename the method to `getKernelConnectionAndThrowIfNotFound` to make it more obivous.

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
```diff
@@ -44,84 +44,128 @@ export class KernelConnectionWrapper implements Kernel.IKernelConnection {
...
     }
     public readonly disposed = new Signal<this, void>(this);
+    /**
+     * Use `kernelConnection` to access the value as its not a constant (can change over time).
+     * E.g. when restarting kernels or the like.
+     */
+    private _kernelConnection!: Kernel.IKernelConnection;
+    private readonly _previousKernelConnection: Kernel.IKernelConnection;
+    // private _isRestarting?: boolean;
+    private get kernelConnection(): undefined | Kernel.IKernelConnection {
```
rchiodo, line 57: The rename is good, but you renamed the wrong one. ```get kernelConnection()``` can return undefined. It should be called 'getPossibleKernelConnection'. The other one (which is now named getPossibleKernelConnection) would make more sense as just ```get kernelConnection()``` because it always returns a connection

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
rchiodo: > They don't know it's restarting

This seems like the root of the problem. We're attempting to let the other extension keep a hold of this wrapper for the IKernelConnection. I think they need to watch for status change events and stop holding onto it when it dies. 

Otherwise the state of this thing won't make sense. 

And if they need to know when a restart is happening, then we need to provide an event for that.

DonJayamanne: Thats wrong, in the kernel API today they can hold onto the kernel connection. its totally valid.
In the jupyter lab API you can hold onto the object and restart.

The whole point of the wrapper is to provide the same behaviour: 
* the kernel wrapper is a singleton (looks like the same kernel connection object)
* When restaring they still deal with the same object (that's how jupyter works)

So the problem isn't the 3rd party extenxion , but us, we need to ensure we provide the same thing & they don't need to worry about any restarts or the like, tha'ts not how the API works today, we don't do this either, hence 3rd party extensions dont need to do this

rchiodo: So if the goal is to behave like jupyter, then we need to modify the status change events then. Otherwise a restart will look like a death of a kernel.

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
rchiodo: Doesn't shutdown also dispose? Seems possible we don't need the kind check here but can just call shutdown? 

This shutdown is an explicit shutdown from the other extension so seems like calling shutdown explicitly would make sense?

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
DonJayamanne: >  Otherwise a restart will look like a death of a kernel.

No it won't, we don't kill the old kernel before swapping.
I did want to send restart events as well, but its just too messy to fix that, will get that sorted later in another debt week.

I've tested this and it works as expected. Only problem is we don't get `restarting` events. Which isn't too bad for now

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
DonJayamanne: > Doesn't shutdown also dispose?

Yes it does. however dispose doesn't shutdown.

> This shutdown is an explicit shutdown from the other extension so seems like calling shutdown explicitly would make sense?

Yes, thats the reason i'm ensuring shutdown will kill the kernel.
In our code we had shutdown which would NOT kill the kernel & we never used that code path, we always called dispose (which would shutdown the kernel if its a local kernel).

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
DonJayamanne: Aah yes,

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
rchiodo: Yes so couldn't this code just call ```this.kernel.session?.shutdown()``` every time? Instead of the condition?

rchiodo (approved):

src/client/datascience/jupyter/kernels/kernelConnectionWrapper.ts
DonJayamanne: You're making me look like an idiot now, 
yes absolutely right. no idea why i even wrote this code.
lol...

DonJayamanne: Aah yes, now i remember, that's because calling `kernel.dispose` will clean everything properly.
So `kernel.dispose` is absolutely necessary (i.e. cleanup).

DonJayamanne: We could optionally call shutdown but that won't be a clean way, as that would kill the session and our code could see that as an unhandled (premature) termination of the kernel.

Issue #8780, opened 2022-01-23
Rename the first item in the quick pick for `local or remote jupyter server`


DonJayamanne:
This first item doesn't make sense anymore as we always support local & remote now.
Hence the two options are no longer mutually exclusive.

I suggest we just rename it to `None - Do not connect to a remote jupyter server` or the like.

![image](https://user-images.githubusercontent.com/1948812/150701495-133e304b-ac34-4d91-be79-687e8f3011ed.png)

rchiodo:
/verified
![image](https://user-images.githubusercontent.com/19672699/155628593-959fcb84-e3d1-471a-9893-bdfc50d8ac8a.png)

PR #8884, opened 2022-01-31
Change labels from switch local to remove remote

DonJayamanne:
For #8780


package.nls.json
```diff
@@ -430,8 +430,8 @@
     "DataScience.jupyterSelectURIQuickPickTitle": "Pick how to connect to Jupyter",
     "DataScience.jupyterSelectURIQuickPickPlaceholder": "Choose an option",
     "DataScience.jupyterSelectURIQuickPickCurrent": "Current: {0}",
-    "DataScience.jupyterSelectURILocalLabel": "Default",
-    "DataScience.jupyterSelectURILocalDetail": "VS Code will automatically start a server for you on the localhost",
+    "DataScience.jupyterSelectURINoneLabel": "None",
+    "DataScience.jupyterSelectURINoneDetail": "Do not connect to any remote Jupyter server",
```
amunger, line 434: Is this just a no-op? It sounds more like 'cancel', but don't we usually just let the user close the quick-pick for that?

amunger: never mind, I just tried it out and see that it drops the remote connections from the kernel picker - that description makes sense then

amunger (approved):

Issue #4738, opened 2021-02-10
Test 'Can run a CSharp notebook' is randomly failing on CI


IanMatthewHuff:
Disabling and marking as must-fix to investigate. Looks to have failed on ~30-40% of recent runs.

PR #8900, opened 2022-02-01
Reenable csharp test to see what is failing

rchiodo:
For #4738

### DIFF
--- a/src/test/datascience/notebook/nonPythonKernels.vscode.test.ts
+++ b/src/test/datascience/notebook/nonPythonKernels.vscode.test.ts
@@ -193,7 +193,6 @@ suite('DataScience - VSCode Notebook - Kernels (non-python-kernel) (slow)', () =
         ]);
     });
     test('Can run a CSharp notebook', async function () {
-        return this.skip(); // Flakey disabled and tracked by 4738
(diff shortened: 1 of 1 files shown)


rchiodo:
Root cause of failure was the timeout for 'waitForTextOutput' was higher than the total timeout for the test. So the first wait would fail before the second one even got a chance to run.

amunger (approved):

DonJayamanne:
> Root cause of failure was the timeout for 'waitForTextOutput' was higher than the total timeout for the test. So the first wait would fail before the second one even got a chance to run.

But that means the first one wouldn't complete at all, hence the test still fails.
Hence I don't think we've fixed the underlying issue.

rchiodo:
> > Root cause of failure was the timeout for 'waitForTextOutput' was higher than the total timeout for the test. So the first wait would fail before the second one even got a chance to run.
> 
> But that means the first one wouldn't complete at all & the test would fail. Hence I don't think we've fixed the underlying issue.

No the first one is supposed to fail. It's html. It times out in 5 seconds and then the second wait works. 

Prior to the 5 second wait, the first one would wait for 60 seconds (longer than the total test timeout).

Issue #8888, opened 2022-01-31
no autocomplete suggestions for dataframe columns


claudiaregio:
## Environment data

-   VS Code version: 1.63.2
-   Jupyter Extension version (available under the Extensions sidebar): v2021.11.1001550889
-   Python Extension version (available under the Extensions sidebar): v2021.12.1559732655
-   OS (Windows | Mac | Linux distro) and version: Windows 11
-   Python and/or Anaconda version: 3.9.5
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): conda
-   Jupyter server running: Local 

## Expected behavior

Column names to be suggested when typing . after a dataframe when assigning to a variable

## Actual behavior

Suggestions for column names are no longer popping up when assigning to a variable 

![image](https://user-images.githubusercontent.com/19276747/151648281-46e62c97-1e6d-44c3-9673-7352d4fcfb51.png)

![image](https://user-images.githubusercontent.com/19276747/151867308-aa140236-fc33-4475-87d4-77d047c63202.png)

DonJayamanne:
Works for me, please could you provide the sample code used to replicate this issue.

DonJayamanne:
Got a confirmation of the issue:
* Getting completions when using `dfc.`
* But when you have assignments like `test = dfc.`

greazer:
I see. Yeah we need to fix this. Is this a regression to?

rchiodo:
Probably a bug in the filtering code for jupyter completions. It would be a regression.

PR #8904, opened 2022-02-01
Stop removing items from completions list

DonJayamanne:
For #8888

TODO:
* [x] Add news entry
* [x] Add tests
	* [x] Completions for paths in strings with single and double quotes
	* [x] Completions for paths with spaces in file names
	* [x] Completions for `df.`
	* [x] Completions for `xyz = df.`
	* [x] Completions for `df.ColumnName.`
	* [x] Completions for `df.Colu` 
	* [x] Completions for `xyz = df.ColumnName.`
	* [x] Completions for `xyz = df.Colu`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1949,7 +1949,7 @@
                 "jupyter.pylanceHandlesNotebooks": {
                     "type": "boolean",
                     "default": false,
-                    "description": "Determines if pylance's experimental notebook support is used or not",
+                    "description": "Determines if pylance's experimental notebook support is used or not.",
                     "scope": "machine",
                     "tags": [
                         "experimental"
@@ -1958,7 +1958,7 @@
                 "jupyter.pythonCompletionTriggerCharacters": {
                     "type": "string",
                     "default": ".%",
-                    "description": "Characters which trigger auto completion on a python jupyter kernel (requires reload of VS code)",
+                    "description": "Characters which trigger auto completion on a python jupyter kernel.",
                     "scope": "machine"
                 }
             }
--- a/src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
+++ b/src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
@@ -308,10 +308,7 @@ export function filterCompletions(
         return r;
     });
 
-    // If not inside of a string, filter out file names (things with a '.' in them or end with '/')
-    if (!insideString) {
-        result = result.filter((r) => !r.itemText.includes('.') && !r.itemText.endsWith('/'));
-    } else {
+    if (insideString) {
         // If inside a string and ending with '/', then add a command to force a suggestion right after
         result = result.map((r) => {
             if (r.itemText.endsWith('/')) {
--- a/src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
+++ b/src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
@@ -308,7 +308,10 @@ export function filterCompletions(
         return r;
     });
 
-    if (insideString) {
+    // If not inside of a string, filter out file names (things that end with '/')
+    if (!insideString) {
+        result = result.filter((r) => !r.itemText.endsWith('/'));
+    } else {
         // If inside a string and ending with '/', then add a command to force a suggestion right after
         result = result.map((r) => {
             if (r.itemText.endsWith('/')) {
--- a/src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
+++ b/src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
@@ -10,14 +10,15 @@ import {
     CompletionItemProvider,
     Position,
     Range,
-    TextDocument
+    TextDocument,
+    workspace
 } from 'vscode';
 import * as lsp from 'vscode-languageclient';
 import { IVSCodeNotebook } from '../../../common/application/types';
 import { createPromiseFromCancellation } from '../../../common/cancellation';
 import { traceError, traceInfoIfCI, traceVerbose } from '../../../common/logger';
 import { getDisplayPath } from '../../../common/platform/fs-paths';
-import { IConfigurationService } from '../../../common/types';
+import { IConfigurationService, IDisposableRegistry } from '../../../common/types';
 import { waitForPromise } from '../../../common/utils/async';
 import { isNotebookCell } from '../../../common/utils/misc';
 import { StopWatch } from '../../../common/utils/stopWatch';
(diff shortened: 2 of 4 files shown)


src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
```diff
@@ -308,10 +308,7 @@ export function filterCompletions(
         return r;
     });
 
-    // If not inside of a string, filter out file names (things with a '.' in them or end with '/')
-    if (!insideString) {
```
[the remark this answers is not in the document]
rchiodo: Because otherwise you end up with all the file names in the intellisense. There was an issue specifically for this.

rchiodo (changes requested):
I assume this is to fix the problem with ```tf = df.ColumName```.

I think there's a logic bug further above. It's supposed to remove the ```.``` that jupyter returns.

src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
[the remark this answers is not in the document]
DonJayamanne: > names in the intellisense. There was an issue specifically for this.

Isn't this code already flaky enough, feels like just pure guess work here. We don't know whether a string containing a `.` or `/` is a path or not. It could be something valid.

> I think there's a logic bug further above. It's supposed to remove the . that jupyter returns.

Not sure what you mean, but removing this code makes it work as expected.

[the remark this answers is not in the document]
rchiodo: > Not sure what you mean, but removing this code makes it work as expected.

Yes and breaks the original fix. You'll now get paths in all sorts of things.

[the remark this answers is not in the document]
rchiodo: > We don't know whether a string containing a . or / is a path or not. It could be something valid.

Yes hence the reason it doesn't do this filtering inside of a string. Only outside.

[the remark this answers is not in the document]
DonJayamanne: Ok, how about we only filter out stuff that contains `/` as those could be treated (assumed to be) as paths.
That way the items that contains `.` will still exist such as `df.column`.

src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
```diff
@@ -308,9 +308,9 @@ export function filterCompletions(
         return r;
     });
 
-    // If not inside of a string, filter out file names (things with a '.' in them or end with '/')
+    // If not inside of a string, filter out file names (things that end with '/')
     if (!insideString) {
-        result = result.filter((r) => !r.itemText.includes('.') && !r.itemText.endsWith('/'));
+        result = result.filter((r) => !r.itemText.endsWith('/'));
```
DonJayamanne, line 313: Updated to remove check for `.`

rchiodo: That's likely better, but you'll still get filenames.

rchiodo: Like open a cell, run some code, type the first letter of a file in your workspace. It will show up in your completion list

DonJayamanne: Then shouldn't we be excluding items that contain `/` or `\` instead of looking for stuff that just ends with `/`

rchiodo: They always contain '/' (if they're multiple levels). Jupyter always uses unix path indicators.

But first level files don't have either of those. They just have a possible '.'

rchiodo: Alternatively (instead of trying to filter on . and /) we could do our own file list for the current directory of the kernel. I think that would require us to execute some code though. Might be too slow.

DonJayamanne: Weird I'm not getting any completions for any paths

rchiodo: I made your change and this is what I get:

image

rchiodo: You can see two files that begin with the letter 'i'

DonJayamanne: Fixed.

src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
```diff
@@ -274,41 +289,58 @@ export function filterCompletions(
...
-        if (newText && newRange) {
-            return {
-                ...r,
-                sortText: generateSortString(i),
-                label: newText,
-                itemText: newText,
-                range: newRange
-            };
+        if (newLabel && newText && newRange) {
+            r.label = newLabel;
```
DonJayamanne, line 333: Label and item to be inserted are different now.
Basically assume you have typed `df.Pass` and the completion item is `Passenger`, then 
* We display the completion item label as `Passenger`
* We add the text to be inserted as `enger` at the position after `s`

DonJayamanne: Previously it worked as we'd replace the entire string, now we just insert at the location where cursor is.

rchiodo (approved):

package.json
```diff
@@ -1958,7 +1958,7 @@
                 "jupyter.pythonCompletionTriggerCharacters": {
                     "type": "string",
                     "default": ".%",
-                    "description": "Characters which trigger auto completion on a python jupyter kernel (requires reload of VS code)",
```
DonJayamanne, line 1961: Updated the code to ensure users don't need to reload VS Code

src/client/datascience/notebook/intellisense/pythonKernelCompletionProvider.ts
```diff
@@ -45,11 +46,23 @@ export class PythonKernelCompletionProvider implements CompletionItemProvider {
...
-        @inject(IConfigurationService) config: IConfigurationService
+        @inject(IConfigurationService) config: IConfigurationService,
+        @inject(IDisposableRegistry) disposables: IDisposableRegistry
     ) {
         const triggerChars = config.getSettings().pythonCompletionTriggerCharacters;
         this.allowStringFilter =
             triggerChars != undefined && (triggerChars.includes("'") || triggerChars.includes('"'));
+        workspace.onDidChangeConfiguration(
+            (e) => {
+                if (e.affectsConfiguration('jupyter.pythonCompletionTriggerCharacters')) {
```
DonJayamanne, line 57: Updated the code to ensure users don't need to reload VS Code

PR #8908, opened 2022-02-02
Add goto definition test for notebooks

### LINKED ISSUE #5125: Tests for GoTo Definition
We should have tests for GoTo Definition functionality.

In particular we need tests for the following:

1. Jumping between notebook cells. Define foo in cell 1 then use it in cell 2. Make sure F12 works. This tests our concat doc converting.
2. Jumping outside of notebook cells to python code. This is what was broken in #4972.

rchiodo:
Fixes #5125


amunger (approved):

amunger (changes requested):
looks like one of the modified tests is failing

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -167,7 +167,11 @@ export class CellExecution implements IDisposable {
                 // If the cell is deleted, then dispose the request object.
                 // No point keeping it alive, just chewing resources.
                 if (e === this.cell.document) {
-                    this.request?.dispose(); // NOSONAR
+                    try {
+                        this.request?.dispose(); // NOSONAR
+                    } catch (e) {
+                        traceError(`Error during cell execution dispose: ${e}`);
```
DonJayamanne, line 173: I think something in the new kernel furture wrapper is throwing errors, I've been running into those a couple of times as well. (basically getting errors about request being disposed).
Looks like something isn't being handled correctly.

rchiodo:
Debugger tests are passing. I think we should still make these changes though.

DonJayamanne (approved):

Issue #8902, opened 2022-02-01
No way to set Juypter Server URI programmatically


C-SELLERS:
## Environment data

-   VS Code version: 1.63.2
-   Jupyter Extension version (available under the Extensions sidebar): v2021.11.1001550889
-   Python Extension version (available under the Extensions sidebar): v2021.12.1559732655
-   OS (Windows | Mac | Linux distro) and version: Ubuntu 20.04
-   Python and/or Anaconda version: 3.6.8
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A
-   Jupyter server running: Local | Remote | N/A; Remote

## Expected behaviour

There exist an option to set the Jupyter Server URI programmatically from another extension/shared config/etc. 

Previously I believe this was supported via a config option in `settings.json` like : 
- `"python.dataScience.jupyterServerURI": "http://localhost:8000/jupyter/"`

With that option I could modify the workspace settings thus setting the URI without using the selector.

Any option to set the URI without user interaction would be greatly appreciated. Whether it be a command that can be executed and passed args, or a config option that is respected from `setting.json`

## Actual behaviour

As far as I am aware of, nothing like this exists currently.

Even the command `"jupyter.selectjupyteruri"` does not allow the passing of an uri. Rather it pulls up the selector you pass to it. 

## Steps to reproduce:

[**NOTE**: Self-contained, minimal reproducing code samples are **extremely** helpful and will expedite addressing your issue]

Try and find a way to set the URI without the user interface.

In my particular use case, I just need to be able to set the URI to a remote jupyter server from the context of a VSCode extension. In doing so, any notebook opened would already be configured to use the remote server without user interaction.

rchiodo:
Thanks for the suggestion. We stopped exposing the URI as a setting because of the potential for people to accidentally checkin in their server token.

But seems reasonable to allow 'jupyter.selectjupyteruri' to allow a URI to use.

C-SELLERS:
That makes sense.

Yeah either a refactor to `jupyter.selectjupyteruri` to allow a uri arg and check if its there, or another command just for this like `jupyter.setJupyterUri` would be great!

C-SELLERS:
In the meantime I am trying to find a workaround, any ideas?

I've explored the implementation and storage in context.secret, so it seems its unlikely that its accessible elsewhere 🤔

rchiodo:
[...]

PR #8918, opened 2022-02-03
Modify SelectJupyterURI to allow URI parameter

C-SELLERS:
For #8902

Made the change to allow a bool and URI to be passed to command `jupyter.selectjupyteruri`. If source is a `Uri` it sets the serverSelector to the remote uri provided by the user.

I added the bool mostly because it seemed like that was what it was originally setup to do. But the arg was unused because `this.serverSelector.selectJupyterURI(true, source),` had `true` hardcoded. It might not be necessary but it does allow someone to call `jupyter.selectjupyteruri` and have their selector not include local options.

I also added a test for this case in the commands unit test for the serverSelector.

-   [x] [Test plan](https://github.com/Microsoft/vscode-jupyter/blob/main/.github/test_plan.md) is updated as appropriate. : N/A
-   [x] [`package-lock.json`](https://github.com/Microsoft/vscode-jupyter/blob/main/package-lock.json) has been regenerated by running `npm install` (if dependencies have changed). N/A

### DIFF
--- a/src/client/common/application/commands.ts
+++ b/src/client/common/application/commands.ts
@@ -179,7 +179,7 @@ export interface ICommandNameArgumentTypeMapping extends ICommandNameWithoutArgu
     [DSCommands.ShowDataViewer]: [IShowDataViewerFromVariablePanel];
     [DSCommands.RefreshDataViewer]: [];
     [DSCommands.ClearSavedJupyterUris]: [];
-    [DSCommands.SelectJupyterURI]: [undefined, 'toolbar' | 'nativeNotebookStatusBar' | undefined];
+    [DSCommands.SelectJupyterURI]: [boolean | undefined, string | undefined, 'toolbar' | 'nativeNotebookStatusBar' | undefined];
     [DSCommands.SelectNativeJupyterUriFromToolBar]: [];
     [DSCommands.DebugNotebook]: [];
     [DSCommands.RunByLine]: [NotebookCell];
--- a/src/client/common/application/commands.ts
+++ b/src/client/common/application/commands.ts
@@ -179,7 +179,7 @@ export interface ICommandNameArgumentTypeMapping extends ICommandNameWithoutArgu
     [DSCommands.ShowDataViewer]: [IShowDataViewerFromVariablePanel];
     [DSCommands.RefreshDataViewer]: [];
     [DSCommands.ClearSavedJupyterUris]: [];
-    [DSCommands.SelectJupyterURI]: [boolean | undefined, string | undefined, 'toolbar' | 'nativeNotebookStatusBar' | undefined];
+    [DSCommands.SelectJupyterURI]: [boolean | undefined, Uri | 'toolbar' | 'nativeNotebookStatusBar' | undefined];
     [DSCommands.SelectNativeJupyterUriFromToolBar]: [];
     [DSCommands.DebugNotebook]: [];
     [DSCommands.RunByLine]: [NotebookCell];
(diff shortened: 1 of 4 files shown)


src/client/common/application/commands.ts
```diff
@@ -179,7 +179,7 @@ export interface ICommandNameArgumentTypeMapping extends ICommandNameWithoutArgu
     [DSCommands.ShowDataViewer]: [IShowDataViewerFromVariablePanel];
     [DSCommands.RefreshDataViewer]: [];
     [DSCommands.ClearSavedJupyterUris]: [];
-    [DSCommands.SelectJupyterURI]: [undefined, 'toolbar' | 'nativeNotebookStatusBar' | undefined];
+    [DSCommands.SelectJupyterURI]: [boolean | undefined, string | undefined, 'toolbar' | 'nativeNotebookStatusBar' | undefined];
```
rchiodo, line 182: This actually won't work for the regular command I think. Other instances of calling this command must already pass the string as the second argument.

You could do this by making the second parameter a ```URI  | 'toolbar' | 'nativeNotebookStatusBar'``` and then check if the parameter is a URI in your handler.

C-SELLERS: True, it seems to work in the default case because its called without anything, but if there are instances its been used I suppose there is no reason to break that compatibility. Will push up a change with it allowing a `Uri` object and use `if instanceof Uri`,

src/client/datascience/commands/serverSelector.ts
```diff
@@ -21,11 +22,15 @@ export class JupyterServerSelectorCommand implements IDisposable {
...
                 Commands.SelectJupyterURI,
-                (local: boolean = true, userUri: string | undefined, source: 'nativeNotebookStatusBar' | 'commandPalette' | 'toolbar' = 'commandPalette') => {
+                (local: boolean = true, source: Uri | 'nativeNotebookStatusBar' | 'commandPalette' | 'toolbar' = 'commandPalette') => {
 
-                    if (!local && userUri) {
-                        traceInfo(`Setting Jupyter Server URI to remote: ${userUri}`);
-                        this.serverSelector.setJupyterURIToRemote(userUri)
+                    if (source instanceof Uri) {
+                        traceInfo(`Setting Jupyter Server URI to remote: ${source}`);
+                        console.log(`Setting Jupyter Server URI to remote: ${source}`)
```
rchiodo: I assume you didn't mean to keep the console.log messages?

C-SELLERS: Yeah, my bad! I removed them 😁

rchiodo (approved):

src/client/datascience/commands/serverSelector.ts
```diff
@@ -8,20 +8,31 @@ import { ICommandManager } from '../../common/application/types';
...
         this.disposables.push(
             this.commandManager.registerCommand(
                 Commands.SelectJupyterURI,
-                (_, source: 'nativeNotebookStatusBar' | 'commandPalette' | 'toolbar' = 'commandPalette') =>
-                    this.serverSelector.selectJupyterURI(true, source),
+                (local: boolean = true, source: Uri | 'nativeNotebookStatusBar' | 'commandPalette' | 'toolbar' = 'commandPalette') => {
+
+                    if (source instanceof Uri) {
+                        traceInfo(`Setting Jupyter Server URI to remote: ${source}`);
+                        this.serverSelector.setJupyterURIToRemote(source.toString(true))
```
rchiodo, line 29: I think both of these have to be marked with a 'void' on the front so the linter doesn't fail.

rchiodo:
You should fix the linter problems. The other tests failing are a known issue and not caused by your change. You can ignore those.

Issue #8910, opened 2022-02-02
Flaky test `Debug a cell from a python file:1 | Invalid number of code lenses returned`


DonJayamanne:
Debug a cell from a python file:1 | Invalid number of code lenses returned
https://github.com/microsoft/vscode-jupyter/runs/5029542406?check_suite_focus=true

amunger:
fixed in https://github.com/microsoft/vscode-jupyter/pull/10899

PR #8922, opened 2022-02-03
added logging for code lens additions

amunger:
For #8910

troubleshooting flakey code lens test where the code lenses are never added to the document

### DIFF
--- a/src/client/datascience/editor-integration/codelensprovider.ts
+++ b/src/client/datascience/editor-integration/codelensprovider.ts
@@ -8,6 +8,7 @@ import { ICommandManager, IDebugService, IDocumentManager, IWorkspaceService } f
 import { ContextKey } from '../../common/contextKey';
 import { disposeAllDisposables } from '../../common/helpers';
 import { IFileSystem } from '../../common/platform/types';
+import { traceInfoIfCI } from '../../common/logger';
 
 import { IConfigurationService, IDisposable, IDisposableRegistry, IJupyterSettings } from '../../common/types';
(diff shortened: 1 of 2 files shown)


src/client/datascience/editor-integration/codelensprovider.ts
```diff
@@ -161,7 +162,7 @@ export class DataScienceCodeLensProvider implements IDataScienceCodeLensProvider
             return codeWatcher.getCodeLenses();
         }
 
-        // Create a new watcher for this file
+        traceInfoIfCI(`Creating a new watcher for document ${document.uri}`);
         const newCodeWatcher = this.createNewCodeWatcher(document);
```
amunger, line 166: When does this case happen? (the documentManager doesn't contain a document with the matching URI)

rchiodo: Document manager doesn't? That seems odd. 

I might have expected the version numbers to not match but the document manager should have all active documents.

amunger: I'm not saying it does happen, more wondering if this fallback is ever hit.

IanMatthewHuff:
Is this just running logging on CI currently? If so you can always create it as a draft until it's ready for review.

amunger:
I haven't seen the same test fail on this branch, so I'm planning on just checking this in to help troubleshoot if it becomes flakey in the future

rchiodo (approved):

Issue #7827, opened 2021-10-05
Upgrade to webpack 5


DonJayamanne:
We have lots of outdated dependencies and hacks/work arounds

IanMatthewHuff:
@DonJayamanne I actually had this mostly done at one point (same time I converted the renderers) the one thing that was really difficult to work around was the old version of the Monaco editor that we were using in the webviews. To get Webpack 5 we had to use a newer version of Monaco-editor-webpack-plugin but that plugin required that we move to a newer version of Monaco which caused all sort of issues. Now that webviews are gone this could be pretty easy, I still have a OneNote with the other changes that I had to make.

PR #8923, opened 2022-02-04
Upgrade to webpack 5

rchiodo:
Fixes https://github.com/microsoft/vscode-jupyter/issues/7827


build/webpack/webpack.datascience-ui.config.builder.js
```diff
@@ -235,17 +234,16 @@ function buildConfiguration(bundle) {
...
                 {
                     test: /\.css$/,
-                    use: ['cache-loader', 'thread-loader', 'style-loader', 'css-loader']
+                    use: ['thread-loader', 'style-loader', 'css-loader']
                 },
                 {
                     test: /\.js$/,
                     include: /node_modules.*remark.*default.*js/,
                     use: [
-                        'cache-loader',
```
rchiodo, line 248: Cache loader not supported in Webpack 5

src/client/datascience/constants.ts
```diff
@@ -2,557 +2,10 @@
 // Licensed under the MIT License.
 'use strict';
 
-import * as path from 'path';
 import { EXTENSION_ROOT_DIR, JVSC_EXTENSION_ID } from '../common/constants';
+import * as path from 'path';
 
-export const DefaultTheme = 'Default Light+';
```
rchiodo, line 8: Moved all of these to datascience-ui/constants so they can be referenced in the UI parts without referencing node constructs (webpack no longer auto fills all node methods)

src/datascience-ui/interactive-common/redux/store.ts
```diff
@@ -159,6 +156,7 @@ function createMiddleWare(
     const testMiddleware =
         forceOnTestMiddleware || testMode || isUITest ? createTestMiddleware(transformLoad) : undefined;
 
+    /* Fixup this code if you need to debug redux
```
IanMatthewHuff, line 159: I think some of our tests rely on this middleware logging still.

IanMatthewHuff: I guess we'll see when the test runs come in, but I think some of the variable tests wait for variable complete logging from the middleware.

rchiodo: From what I know, the process environment variables were never set to anything, so unless I misread the logic here, the testMiddleware isn't even possible to enable.

IanMatthewHuff: Hmm, I think I was thinking of the forceTestMiddleware part of webview.ts. Does that still work? Pretty sure that was checking ENV.

IanMatthewHuff: Ahh, that's still on the main node side. Nevermind.

package.json
```diff
@@ -2307,11 +2306,13 @@
         "node-gyp-build": "^4.2.3",
         "node-has-native-dependencies": "^1.0.2",
         "node-html-parser": "^1.1.13",
+        "node-loader": "^2.0.0",
```
IanMatthewHuff, line 2309: Is this used? I don't see it added in the webpack files.

IanMatthewHuff (approved):

package.json
rchiodo: Oh I don't need it. Let me remove.

PR #8938, opened 2022-02-06
Turn pre-release back on

DonJayamanne:


package.json
```diff
@@ -27,7 +27,7 @@
         "theme": "light"
     },
     "engines": {
-        "vscode": "^1.63.0"
+        "vscode": "^1.64.0"
```
rchiodo, line 30: I didn't think we were going to update this unless we had to?

PR #8990, opened 2022-02-10
Fix our wait time 

IanMatthewHuff:
For microsoft/vscode-jupyter-internal#258

### DIFF
--- a/src/client/datascience/raw-kernel/rawJupyterSession.ts
+++ b/src/client/datascience/raw-kernel/rawJupyterSession.ts
@@ -312,7 +312,7 @@ export class RawJupyterSession extends BaseJupyterSession {
             try {
                 await Promise.race([
                     Promise.all([result.kernel.requestKernelInfo(), gotIoPubMessage.promise]),
-                    sleep(Math.min(this.launchTimeout, 10)),
+                    sleep(Math.min(this.launchTimeout, 10_000)),
                     createPromiseFromCancellation({ cancelAction: 'reject', token: options.token })
                 ]);
             } catch (ex) {
--- a/news/2 Fixes/8989.md
+++ b/news/2 Fixes/8989.md
@@ -1,0 +1,1 @@
+Fix our waiting for RequestKernelInfo (might have been hurting raw kernel start some).


src/client/datascience/raw-kernel/rawJupyterSession.ts
```diff
@@ -312,7 +312,7 @@ export class RawJupyterSession extends BaseJupyterSession {
             try {
                 await Promise.race([
                     Promise.all([result.kernel.requestKernelInfo(), gotIoPubMessage.promise]),
-                    sleep(Math.min(this.launchTimeout, 10)),
+                    sleep(Math.min(this.launchTimeout, 10_000)),
```
IanMatthewHuff, line 315: Looks like this has been busted for a bit (ran into this investigating another issue). Not sure that it's causing specific issues, but in telemetry over the last week we have ~1,400,000 times we get a response from requestKernelInfo and ~900,000 times we time out due to this. Via the code I think we should be ok regardless of if we get this or not in general.

IanMatthewHuff: Note there was a comment above asking for 10s here, just looks like the units was missed at the time.

rchiodo (approved):

src/client/datascience/raw-kernel/rawJupyterSession.ts
DonJayamanne: Should try to get this into a recovery build, no idea how many users run into this issue and cannot connect.

DonJayamanne (approved):

src/client/datascience/raw-kernel/rawJupyterSession.ts
IanMatthewHuff: @DonJayamanne I was wondering that. Looking at the code it seemed mostly benign.

Basic thing is that very often it's going to drop out of this promise race early before giving requestKernelInfo a real chance to come back. But this continues on and just warns regardless of if we actually get a reply from requestKernelInfo. It doesn't abort the launch.

Actually though, it could be forcing a second attempt at requestKernelInfo when that's not needed which = startup time lost. That's possibly worth the port right there. Will discuss quick on slack.

Issue #5061, opened 2021-03-07
Incorrect PYTHONPATH variable with .env file with relative paths


secure-mail-documentation-project:
Issue Type: <b>Bug</b>

The jupyter extension sets the PYTHONPATH variable incorrectly when a `.env` file contains relative path definitions.

To reproduce:
- On Windows 10
- vscode project base in `C:\Users\My User\Documents\vscode_test_project`
- `.env` file in above direcory with the content `PYTHONPATH=./some-library`
- python file `subdir\test.py` in above directory that is empty except for the line `# %%` opened in editor
- Click on "Run Cell"
- Jupyter opens
- in jupyter notebook execute `import sys; sys.path`

Expected result:
```
['c:\\Users\\My User\\.vscode\\extensions\\ms-toolsai.jupyter-2021.2.603412351\\pythonFiles\\.does-not-exist',
 'C:\\Users\\My User\\Documents\\vscode_test_project\\some-library',
 'c:\\Users\\My User\\.vscode\\extensions\\ms-toolsai.jupyter-2021.2.603412351\\pythonFiles',
 'c:\\Users\\My User\\.vscode\\extensions\\ms-toolsai.jupyter-2021.2.603412351\\pythonFiles\\lib\\python',
 'C:\\ProgramData\\Miniconda3\\envs\\vscode_test_env\\python39.zip',
 'C:\\ProgramData\\Miniconda3\\envs\\vscode_test_env\\DLLs',
 'C:\\ProgramData\\Miniconda3\\envs\\vscode_test_env\\lib',
 'C:\\ProgramData\\Miniconda3\\envs\\vscode_test_env',
 '',
 'C:\\ProgramData\\Miniconda3\\envs\\vscode_test_env\\lib\\site-packages',
 'C:\\ProgramData\\Miniconda3\\envs\\vscode_test_env\\lib\\site-packages\\win32',
 'C:\\ProgramData\\Miniconda3\\envs\\vscode_test_env\\lib\\site-packages\\win32\\lib',
 'C:\\ProgramData\\Miniconda3\\envs\\vscode_test_env\\lib\\site-packages\\Pythonwin',
 'C:\\ProgramData\\Miniconda3\\envs\\vscode_test_env\\lib\\site-packages\\IPython\\extensions',
 'C:\\Users\\My User\\.ipython']
```

Actual result:
```
[...]

PR #9018, opened 2022-02-14
Load env variables from .env files for non-conda

DonJayamanne:
For #5061

### DIFF
--- a/src/client/common/process/environmentActivationService.ts
+++ b/src/client/common/process/environmentActivationService.ts
@@ -194,7 +194,7 @@ export class EnvironmentActivationService implements IEnvironmentActivationServi
         void this.envVarsService.getEnvironmentVariables(resource);
 
         // Check cache.
-        const [env, customEnvVars] = await Promise.all([
+        let [env, customEnvVars] = await Promise.all([
             this.apiProvider.getApi().then((api) => api.getActivatedEnvironmentVariables(resource, interpreter, false)),
             this.envVarsService.getCustomEnvironmentVariables(resource)
         ]);
@@ -224,6 +224,12 @@ export class EnvironmentActivationService implements IEnvironmentActivationServi
         } else if (this.memento.get(key)) {
             // Remove it from cache (if it exists).
             void this.memento.update(key, undefined);
+        }
+        if (env && customEnvVars) {
+            env = {
+                ...env,
+                ...customEnvVars
+            };
         }
         return env;
     }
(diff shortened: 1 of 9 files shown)


src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
```diff
@@ -48,22 +48,23 @@ export class KernelEnvironmentVariablesService {
                 });
         }
 
-        if (interpreter?.envType !== EnvironmentType.Conda) {
```
DonJayamanne, line 51: This was the problem, when dealing with non-conda we'd not load the env variables.

src/client/common/variables/environmentVariablesProvider.ts
```diff
@@ -116,7 +116,8 @@ export class EnvironmentVariablesProvider implements IEnvironmentVariablesProvid
     }
     private getWorkspaceFolderUri(resource?: Uri): Uri | undefined {
         if (!resource) {
-            return;
+            const workspaceFolders = this.workspaceService.workspaceFolders || [];
```
DonJayamanne, line 119: If we have just one workspace folder opened, then use that to load the env variables.

src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
DonJayamanne: Could easily repro on linux as well.

src/client/common/process/environmentActivationService.ts
```diff
@@ -225,6 +225,12 @@ export class EnvironmentActivationService implements IEnvironmentActivationServi
             // Remove it from cache (if it exists).
             void this.memento.update(key, undefined);
         }
+        if (env && customEnvVars) {
```
DonJayamanne, line 228: In debt will refactor this, its a bigger change, but I think the way we create Python execution services is very complicated now and we've regressed in certain areas hence the need to fix this change in a number of places now.

src/test/datascience/notebook/executionService.vscode.test.ts
```diff
@@ -167,6 +167,27 @@ suite('DataScience - VSCode Notebook - (Execution) (slow)', function () {
 
         await Promise.all([runCell(cell), waitForTextOutput(cell, '\tho\n\tho\n\tho\n', 0, true)]);
     });
+    test.only('Verify loading of env variables form .env file', async function () {
+        if (IS_REMOTE_NATIVE_TEST) {
```
DonJayamanne, line 171: This test runs on Local + Raw & Local + Jupyter

rchiodo (approved):

simeld:
I still have this issue when running python code, but not when debugging.
See
https://github.com/simeld/bug-vscode-dotenv
for a repo with reproducible bug.

DonJayamanne:
@simeld  please file an issue against the python extension repo

PR #9031, opened 2022-02-15
Fix build number for prerelease

DonJayamanne:


gulpfile.js
```diff
@@ -201,9 +201,8 @@ async function updateBuildNumber() {
     const versionParts = packageJson.version.split('.');
     // New build is of the form `MMDDHHM` (7 digits, as first three are reserved for `100` or `100` for patches).
```
amunger, line 202: looks like this comment is wrong, we're not using the day here. If a build is released at 9:15 one day and 8:15 the next day, the newer build will have a lower number.

PR #9032, opened 2022-02-15
Do not use daemon for Jupyter servers

DonJayamanne:


DonJayamanne:
> At least one customer wouldn't work with direct jupyter start. Daemon seems to be easier to start than jupyter itself.

Do you have the details of the issue, that doesn't seem possible.

DonJayamanne:
There's no evidence at all that the daemon would have worked
https://github.com/microsoft/vscode-jupyter/issues/8995#issuecomment-1036615915
The user just installed Jupyter into another Virtual environment, hence we don't know that the daemon worked and using CLI didn't.

Actually the CLI worked in the other new virtual enviornment, as the user was able to get this working as you suggested (Creating a new virtual env and installing jupyter into that).
Hence I think removing daemon for this case is still valid.

rchiodo:
This is the comment that makes me think it used to work:

> It failed when it tried to start Jupyter in the environment. I have had no problem using Jupyter Notebook in Visual Studio Code until **today** after I was prompted to restart Visual Studio Code for an update.

DonJayamanne:
> It failed when it tried to start Jupyter in the environment. I have had no problem using Jupyter Notebook in Visual Studio Code until today after I was prompted to restart Visual Studio Code for an update.

Hmm agreed & its very weird, not sure how the user got into this state. 
How then did users get stuff working before we even had the daemon? 
I'll ping the user and see.

DonJayamanne:
> Hmm agreed & its very weird, not sure how the user got into this state.

Hvae to get to the bottom of this, else it means we have no idea what's goin on and we cannot remove daemons (which isn't a good thing, as daemons came up much later)

rchiodo (approved):

DonJayamanne:
I'll merge this in debt week, no need to change this right now.

IanMatthewHuff:
@DonJayamanne I thought there was an interrupt signal issue on windows as well. Or is that just in the raw kernel case?

DonJayamanne:
> n interrupt signal issue on windows as well. Or is that just in the raw kernel case?

Yes, that's only for the case of raw kernels. When using Jupyter we don't need to use a daemon, jupyter will handle interrupts it self.

IanMatthewHuff (approved):

PR #9040, opened 2022-02-15
Better bulid numbers

DonJayamanne:


gulpfile.js
```diff
@@ -199,10 +199,15 @@ async function updateBuildNumber() {
...
 
     const versionParts = packageJson.version.split('.');
-    // New build is of the form `MMDDHHM` (7 digits, as first three are reserved for `100` or `100` for patches).
+    // New build is of the form `DDDHHMM` (day of year, hours, minute) (7 digits, as out of the 10 digits first three are reserved for `100` or `100` for patches).
     // Use date time for build, this way all subsequent builds are incremental and greater than the others before.
-    const parts = [new Date().getMonth() + 1, new Date().getHours() + 1, new Date().getMinutes() + 1];
-    const buildNumberSuffix = `${parts.map((item) => item.toString().padStart(2, '0')).join('')}0`;
+    // Example build for 3Jan 12:45 will be `0031245`, and 16 Feb 8:50 will be `0470845`
+    const today = new Date();
+    const dayCount = [0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334];
```
rchiodo, line 206: Does it matter that this doesn't take leap year into account? Won't that mean on Feb 29th we'll have the same day count as March 1st?

rchiodo: Maybe you could use this?
https://www.npmjs.com/package/day-of-year

DonJayamanne: > Won't that mean on Feb 29th we'll have the same day count as March 1st?

Only for two different years. For the same year it will be different.
We don't care about uniqueness across years, as the first part of the version `2022` will take care of that. and so will the second part `2022.02`.

rchiodo: I don't think that will work. It's assuming on March 1st we udpate to 2022.03. We don't usually do that based on a month boundary, but rather a endgame boundary.

I think you could also fix it by just assuming Feb always has 29 days. Then it would just have a blank spot for most years, but it would work with leap years.

DonJayamanne: Oh shoot yes, damn it you are right.

gulpfile.js
```diff
@@ -199,10 +199,15 @@ async function updateBuildNumber() {
     // If we have more hot fixes, they'll be YYYY.MMM.120<build number>, YYYY.MMM.130<build number>, & so on.
 
     const versionParts = packageJson.version.split('.');
-    // New build is of the form `MMDDHHM` (7 digits, as first three are reserved for `100` or `100` for patches).
+    // New build is of the form `DDDHHMM` (day of year, hours, minute) (7 digits, as out of the 10 digits first three are reserved for `100` or `100` for patches).
```
amunger, line 202: "`100` or `100`" looks like a typo

amunger (approved):

rchiodo (approved):

Issue #8827, opened 2022-01-27
kernel connection file created in `/tmp/` rather than `~/.local/share/jupyter/runtime/`


Fan-iX:
## Environment data

-   VS Code version: 1.63.2
-   Jupyter Extension version (available under the Extensions sidebar): v2021.11.1001550889
-   Python Extension version (available under the Extensions sidebar): v2021.12.1559732655
-   OS (Windows | Mac | Linux distro) and version: Gentoo Linux / CentOS 7 / WSL Ubuntu 20.04
-   Python and/or Anaconda version: Python 3.8.5
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A
-   Jupyter server running: Remote / WSL

## Expected behaviour

runtime kernel connection json file created in `~/.local/share/jupyter/runtime/` so that it can be recognized by `jupyter console --existing`

## Actual behaviour

runtime kernel connection json file created in `/tmp/` and `jupyter console --existing` shows error message
`[ZMQTerminalIPythonApp] CRITICAL | Could not find existing kernel connection file kernel*.json`

this is not a big problem for ipykernel since I can connect a jupyter console to the kernel according to the `%connect_info` message.
but it becomes a problem when I use irkernel in which the `%connect_info` method haven't been implemented.

## Steps to reproduce:

1. create an interactive window in VS Code.
2. run `%connect_info`
3. see `$> jupyter <app> --existing /tmp/tmp-1033rJp0TWSU8vOv.json` in cell output, indicating that the connection json file is created in `/tmp/`
4. run `jupyter console --existing`
5. get error message `[ZMQTerminalIPythonApp] CRITICAL | Could not find existing kernel connection file kernel*.json`

## Logs

<details>

<summary>Output for <code>Jupyter</code> in the <code>Output</code> panel 
</summary>

<p>

```
%connect_info
```

```
{
  "shell_port": 9002,
  "iopub_port": 9004,
  "stdin_port": 9003,
  "control_port": 9001,
  "hb_port": 9000,
  "ip": "127.0.0.1",
  "key": "0b43f78a-4183-4728-b3c6-d734dfa4fb12",
  "transport": "tcp",
  "signature_scheme": "hmac-sha256",
  "kernel_name": ""
}

Paste the above JSON into a file, and connect with:
    $> jupyter <app> --existing <file>
or, if you are local, you can connect with just:
[...]

PR #9041, opened 2022-02-15
Enable verbose logging and kernel level logging

DonJayamanne:
More logging to see what's going on
Enable logging of messages within IPykernel, so we can see what's going on in there.

### DIFF
--- a/src/client/datascience/kernel-launcher/kernelProcess.ts
+++ b/src/client/datascience/kernel-launcher/kernelProcess.ts
@@ -22,7 +22,7 @@ import {
     IPythonExecutionFactory,
     ObservableExecutionResult
 } from '../../common/process/types';
-import { IOutputChannel, Resource } from '../../common/types';
+import { IJupyterSettings, IOutputChannel, Resource } from '../../common/types';
 import { createDeferred } from '../../common/utils/async';
 import * as localize from '../../common/utils/localize';
 import { noop, swallowExceptions } from '../../common/utils/misc';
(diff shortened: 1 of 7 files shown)


src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -437,7 +438,11 @@ export class KernelProcess implements IKernelProcess {
 
             // If we don't have a KernelDaemon here & we're not running a Python module either.
             // The kernelspec argv could be something like [python, main.py, --something, --something-else, -f,{connection_file}]
-            exeObs = executionService.execObservable(this.launchKernelSpec.argv.slice(1), {
+            const args = this.launchKernelSpec.argv.slice(1);
+            if (this.jupyterSettings.enablePythonKernelLogging) {
+                args.push('--debug');
```
DonJayamanne, line 443: Need to see what's going on inside the kernel

rchiodo: LOL. I have this as the default on my own machine.

rchiodo (approved):

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -349,7 +350,6 @@ export class KernelProcess implements IKernelProcess {
             // a different process tries to open it.
             const tempFile = await this.fileSystem.createTemporaryLocalFile('.json');
             this.connectionFile = tempFile.filePath;
-            await tempFile.dispose();
```
DonJayamanne, line 352: This is wrong, we need the file to exist, the next line we attempt to write to the file.
What is happening is sometimes the file is created and sometimes its deleted after we connect.

@rchiodo (found this based on the error messages logged by the kernel after adding `--debug`)
**New changes after approval**

rchiodo: Should the dispose happen later then? Like maybe in the shutdown of the kernel?

DonJayamanne: Will do that as part of https://github.com/microsoft/vscode-jupyter/issues/8827

PR #9064, opened 2022-02-17
Update npm package versions

DonJayamanne:


build/ci/postInstall.js
```diff
@@ -103,6 +103,29 @@ function makeVariableExplorerAlwaysSorted() {
...
+function fixJupyterLabRenderers() {
+    const warnings = [];
+    ['node_modules/@jupyterlab/cells/lib/widget.js', 'node_modules/@jupyterlab/rendermime/lib/renderers.js'].forEach(
+        (file) => {
+            const filePath = path.join(__dirname, '..', '..', file);
+            if (!fs.existsSync(filePath)) {
+                return;
+            }
+            const textToReplace = `import marked from 'marked'`;
+            const textToReplaceWith = `import { marked } from 'marked'`;
```
DonJayamanne, line 115: The way `marked` had to be imported changed in v3.0 (we were using <1 and now we're using v4)
Jupyter lab too has updated their packages, however they haven't published a stable version yet, hence we have to do this.

package.json
```diff
@@ -2341,5 +2345,11 @@
     },
     "optionalDependencies": {
         "canvas": "^2.7.0"
+    },
+    "resolutions": {
```
DonJayamanne, line 2349: Use of `resolutions` npm package to pin child dependencies
(in npm 8 we can do this natively without any other npm packages)

src/test/datascience/notebook/helper.ts
```diff
@@ -736,7 +736,9 @@ export async function waitForTextOutput(
         async () => assertHasTextOutputInVSCode(cell, text, index, isExactMatch),
         timeout,
         () =>
-            `Output does not contain provided text '${text}' for Cell ${cell.index + 1}, it is ${getCellOutputs(cell)}`
+            `A after ${timeout}ms output does not contain provided text '${text}' for Cell ${
```
rchiodo, line 739: Looks like a findreplace mistake?

DonJayamanne: No it was deliberate, i wanted to include the timeout in the error message. Noticed this test was failing on CI

rchiodo (approved):

PR #9066, opened 2022-02-17
Increase timeout of the test

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/9065

### DIFF
--- a/src/test/datascience/jupyter/kernels/installationPrompts.vscode.test.ts
+++ b/src/test/datascience/jupyter/kernels/installationPrompts.vscode.test.ts
@@ -140,7 +140,9 @@ suite('DataScience Install IPyKernel (slow) (install)', function () {
 
     [true, false].forEach((which) => {
         const interpreterPath = which ? venvPythonPath : venvNoRegPath;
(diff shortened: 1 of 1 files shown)


rchiodo (approved):

src/test/datascience/jupyter/kernels/installationPrompts.vscode.test.ts
```diff
@@ -148,7 +150,6 @@ suite('DataScience Install IPyKernel (slow) (install)', function () {
                 disposables
             );
             const installed = createDeferred();
-            const interpreterPath = which ? venvPythonPath : venvNoRegPath;
```
DonJayamanne, line 151: Looks like the Python extension code has changed, we're getting a response even before the installation has completed.
In the past we'd wait for the installation to complete (even when running the installer in the terminal).
Based on the logs we can see that ipykernel is indeed getting installed (files are getting created as part of ipykernel installation).

@rchiodo When you bring in the changes for package installation we can just enable this test (as we'll no longer be using the terminal).

Hence disabling the test for now

IanMatthewHuff (approved):

PR #9092, opened 2022-02-22
Manually delete temporary ipynb

DonJayamanne:
We've seen users with systems where the temporary ipynb files do not get deleted.
This PR just removes them using node.js
Why don't we do this always instead of using the Jupyter lab API

### DIFF
--- a/src/client/datascience/jupyter/jupyterSession.ts
+++ b/src/client/datascience/jupyter/jupyterSession.ts
@@ -14,7 +14,7 @@ import * as uuid from 'uuid/v4';
 import { CancellationToken } from 'vscode-jsonrpc';
 import { Cancellation } from '../../common/cancellation';
 import { BaseError } from '../../common/errors/types';
-import { traceError, traceInfo, traceInfoIfCI, traceVerbose } from '../../common/logger';
+import { traceError, traceInfo, traceInfoIfCI, traceVerbose, traceWarning } from '../../common/logger';
 import { IOutputChannel, Resource } from '../../common/types';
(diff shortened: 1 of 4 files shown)


src/client/datascience/jupyter/jupyterSession.ts
```diff
@@ -275,7 +277,11 @@ export class JupyterSession extends BaseJupyterSession {
             } catch (ex) {
                 // If we failed to create the kernel, we need to clean up the file.
                 if (this.connInfo && backingFile) {
-                    this.contentsManager.delete(backingFile.path).ignoreErrors();
+                    void this.contentsManager
+                        .delete(backingFile.path)
+                        .catch((ex) => traceWarning(`Failed to delete ${backingFile.path} using Jupyter API`, ex))
+                        // Sometimes the files don't get deleted via Jupyter, delete them manually.
+                        .finally(() => this.fs.deleteLocalFile(backingFile.path));
```
DonJayamanne, line 284: Why don't we do this always instead of using the Jupyter lab API
I'm not sure why we need to use the Jupyter API, perhaps it has some better cleanup?
This way seems safe

rchiodo: Jupyter works in the remote scenario.

I believe the checkpoint option needs to be turned off too, otherwise the temporary file will keep coming back.

DonJayamanne: this code aoplies to local, however checkpoints make sense

PR #9101, opened 2022-02-23
Ensure stacktrace is parsed correctly

### LINKED ISSUE #9059: Allow extensions to pass their display names when registering themselves using the registerRemoteServerProvider()
We could pass a display name variable in the JupyterUriProvider which is passed in the registerRemoteServerProvider() api of the Jupyter extension. This can then be shown in the description of the quick picks rather than something like "ms-toolsai.vscode-ai"

DonJayamanne:
Fixes #9059

### DIFF
--- a/src/client/common/application/extensions.ts
+++ b/src/client/common/application/extensions.ts
@@ -10,6 +10,7 @@ import { EXTENSION_ROOT_DIR } from '../constants';
 import { IExtensions } from '../types';
 import { DataScience } from '../utils/localize';
 import { IFileSystem } from '../platform/types';
+import * as stacktrace from 'stack-trace';
 
 @injectable()
 export class Extensions implements IExtensions {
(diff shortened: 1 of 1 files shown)


src/client/common/application/extensions.ts
```diff
@@ -40,6 +41,12 @@ export class Extensions implements IExtensions {
                     }
                 })
                 .filter((item) => item && !item.toLowerCase().startsWith(jupyterExtRoot)) as string[];
+            stacktrace.parse(new Error('Ex')).forEach((item) => {
```
DonJayamanne, line 44: The stack trace can also be `at xyx.js:123:12` and this doesn't match the regex.
I've left the regex in case that cases some other edge cases and added the use of the package which should be more accurate and cover the other cases we haven't handled with our regex. (not going to fix the regex, as the package would be more accurate)

FYI - this is a breaking change for AML extension

IanMatthewHuff (approved):

src/client/common/application/extensions.ts
DonJayamanne: Oops, should remove that `Ex`

DonJayamanne: Not required.

rchiodo (approved):

PR #9125, opened 2022-02-25
re-enable notebook creation test suite

amunger:
These tests were being registered in the wrong scope and therefore not run

### DIFF
--- a/src/test/datascience/notebook/creation/notebookCreation.vscode.test.ts
+++ b/src/test/datascience/notebook/creation/notebookCreation.vscode.test.ts
@@ -74,7 +74,7 @@ suite('DataScience - VSCode Notebook - (Creation Integration)', function () {
         );
     }
     test('With 3rd party integration, display quick pick when selecting create blank notebook command', async function () {
-        await creationOptions.registerNewNotebookContent('julia');
+        await creationOptions.registerNewNotebookContent('javascript');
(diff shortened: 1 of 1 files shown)


rchiodo (approved):

rchiodo:
Looks like at least two of the new tests don't pass. THey might need to be adjusted to waitForSelectedKernel? Haven't looked yet.

src/test/datascience/notebook/creation/notebookCreation.vscode.test.ts
```diff
@@ -59,86 +59,74 @@ suite('DataScience - VSCode Notebook - (Creation Integration)', function () {
...
 
-            // Create a blank notebook & it should just work.
-            await createNotebookAndValidateLanguageOfFirstCell(PYTHON_LANGUAGE.toLowerCase());
-        });
-        test('Create Java & Julia Notebook using API', async function () {
-            await api.createBlankNotebook({ defaultCellLanguage: 'java' });
+        // Create a blank notebook & it should just work.
+        await createNotebookAndValidateLanguageOfFirstCell(PYTHON_LANGUAGE.toLowerCase());
+    });
+    test('Create javascript & powershell Notebook using API', async function () {
```
amunger, line 113: these are the only two languages that seem to behave as expected, so it seems like this scenario is still mostly broken

rchiodo (approved):

PR #9126, opened 2022-02-25
Fix .env file breaking global interpreter based kernels

### LINKED ISSUE #9127: .env file breaks global python kernel startup
Create a '.env' file in the root of workspace
Add anything to it
Try to start a global python kernel (on windows at least)

Failure:
The kernel failed to start due to the missing module 'ipykernel_launcher'. Consider installing this module. 
View Jupyter [log](command:jupyter.viewOutput) for further details.

rchiodo:
Fixes #9127 

If you have a .env file it forces us to figure out environment variables for kernel launch. The additions of ```PYTHONNOUSERSITE``` prevents a kernel from even starting if its a global kernel.

I think this fix will allow #8553 to work plus fix this for global environments.

Yeah I tested #8553 after this change and it still works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Jupyter Extension for Visual Studio Code
+# Jupyter Extension for Visual Studio Code (with a checkin)
 
 A [Visual Studio Code](https://code.visualstudio.com/) [extension](https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter) that provides basic notebook support for [language kernels](https://github.com/jupyter/jupyter/wiki/Jupyter-kernels) that are supported in [Jupyter Notebooks](https://jupyter.org/) today. Many language kernels will work with no modification. To enable advanced features, modifications may be needed in the VS Code language extensions.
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-# Jupyter Extension for Visual Studio Code
+# Jupyter Extension for Visual Studio Code (with a checkin)
```
rchiodo, line 1: Crap didn't mean to include this

src/test/datascience/.env
```diff
@@ -1,2 +1 @@
 ENV_VAR_TESTING_CI=HelloWorldEnvVariable
-PYTHONPATH=./dummyFolderForPythonPath
```
rchiodo, line 2: @DonJayamanne is this expected to work? It looked like in the failures that it was only using this as the path. For a windows store python at least this breaks any test from running.

IanMatthewHuff (approved):

src/test/datascience/.env
rchiodo: Actually it's worse than I thought. Does more than break the test. It breaks everything just by existing. See this bug here:
https://github.com/microsoft/vscode-jupyter/issues/9127

amunger:
This seems to resolve that specific issue for me locally

amunger (approved):

IanMatthewHuff (approved):

PR #9131, opened 2022-02-25
Release 2022.02.100

amunger:
merging release notes changes back into main


IanMatthewHuff (approved):

src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
```diff
@@ -65,6 +65,7 @@ export class KernelEnvironmentVariablesService {
             return kernelEnv;
         }
         // Merge the env variables with that of the kernel env.
+        const hasInterpreterEnv = interpreterEnv != undefined;
```
rchiodo, line 68: Looks like you need to refresh your branch. This shouldn't be different. I already submitted it to main

amunger: I do not understand why this is showing up. The PR is into origin/main, but somehow it doesn't have an up-to-date version of that branch?

IanMatthewHuff: Was your local main up to date? I think it's a PR for merging release into an older main branch that you had, not the current main head.

DonJayamanne (changes requested):
~Any idea why the package-lock.json changing?~
~This shouldn't change.~

~Please ensure you have the right version of node & npm.~
~Node version is defined in .nvmrc file or our yml files. (14.16.0) The npm version should be the default of 6.14.11.~

Found the problem, its just a change in formatting, all good.

DonJayamanne (approved):

src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
amunger: Pretty sure the issue was that Github is using the common ancestor for comparison between the two branches: https://stackoverflow.com/a/35203417/1253265
I went ahead and merged and confirmed that there are no new commits to this file in main.

PR #9161, opened 2022-03-01
Refactor how errors are handled in both notebooks and IW

### LINKED ISSUE #8817: Interactive window cell not executed after selecting a kernel with IPykernel in it
Tested in latest pre-release version of Jupyter ext

* Open a Python file
* Create a cell with `print("Hello World")`
* Ensure active python interpreter doesn't have ipykernel installed
* Run a cell
* Ensure you get a prompt to install ipykernel
* Select a kernel that has ipykernel isntalled

Problems
* Cell is printed before the progress message for the startup of the kernel (ideally the kernel startup info should be before the cell is printed in IW)
* The cell is not executed at all

rchiodo:
Fixes #8817

I refactored some of how error handling works and where errors are presented.

- The DataScienceErrorHandler no longer throws errors. It simply attempts to install if necessary. If the install fails, the calling code can decide what to do with that.
- The DataScienceErrorHandler no longer generates the 'message' and 'more info' links. This is done at the same time the reason is decided (inside analyzeKernelError). Seemed like a good shortcut was to do all of this in the same spot (instead of having two switch statements in two different files that do the same thing).
- A new ```connectToKernel``` method was created that wraps the starting of a kernel, allowing the user to switch or install if necessary, and returning an error if that fails. This method is used both in a notebook controller handleExecution and in the interactive window. These two high level spots can then decide what to do with errors.
- IW was modified to use the the new ```connectToKernel``` and use an errors returned to modify the message banners in the IW. This used to happen in the kernel (writing to the IW) and it seemed weird for it to be there.
- VSCodeNotebookController was also modified to use ```connectToKernel```. It sticks any errors found into the cell(s) that are executing. The weird part here is that 'executeCell' on a controller might end up picking another controller before it finishes the execute. Not super happy about that, but it's the entry point we're given by VS code. Maybe I could instead change this to end the execute and just run all the cells that were already queued up with a command.


src/client/common/process/environmentActivationService.ts
```diff
@@ -143,7 +143,8 @@ export class EnvironmentActivationService implements IEnvironmentActivationServi
         const title = DataScience.activatingPythonEnvironment().format(
             interpreter.displayName || getDisplayPath(interpreter.path)
         );
-        return KernelProgressReporter.wrapAndReportProgress(resource, title, () =>
+        const cancellationTokenSource = new CancellationTokenSource();
```
DonJayamanne, line 146: This needs to be disposed

src/client/datascience/baseJupyterSession.ts
```diff
@@ -377,6 +377,7 @@ export abstract class BaseJupyterSession implements IJupyterSession {
                 if (!this.canShutdownSession(session, isRequestToShutDownRestartSession, shutdownEvenIfRemote)) {
                     traceVerbose(`Session cannot be shutdown ${session.kernelConnectionMetadata?.id}`);
                     session.dispose();
+                    this._disposed.fire();
```
DonJayamanne, line 380: I don't think this is correct.
In the old class `disposed` is fired in only one place thats the `shutdownImplementation`
This method is used only to shutdown a specific session object, not to shutdown (dispose) this entire class.

Looking at some code in other places I can see that's what's expected
I.e. `this._disposed.fire` should only be called when the entire session is killed off (either via a dispose or explicit call to the `shutdown` method)
```
            // Shutdown the session but not this class.
            this.setSession(undefined);
            this.shutdownSession(session, this.statusHandler, false)
```

DonJayamanne: I'd like to remove this and clean up so that we have the disposed called from just one place. So we know when and where the class is disposed.

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -1774,3 +1793,98 @@ export async function executeSilently(session: IJupyterSession, code: string): P
...
+                    memento,
+                    undefined,
+                    controller.connection.interpreter.path
+                ).ignoreErrors();
+            }
+
+            const handleResult = await errorHandler.handleKernelError(error, 'start', controller.connection, resource);
+            switch (handleResult.kind) {
+                case 'Canceled':
+                    throw new CancellationError(
```
DonJayamanne, line 1859: Don't we need to dispose the kernle here as well?
Basically the kernel hasn't been started

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -377,15 +369,31 @@ export class VSCodeNotebookController implements Disposable {
...
+            // If there was a failure connecting or executing the kernel, stick it in this cell
+            return displayErrorsInCell(cell, execution, ex.toString());
+        }
+
+        // Execution should be ended elsewhere
+    }
+
+    @chainable()
+    private async connectToKernel(doc: NotebookDocument) {
+        // Make sure we don't have multiple cells trying to connect at the same time
```
DonJayamanne, line 395: THis comment is confusing, we don't seem to doing anything with cells here, hence not sure about the comment

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -377,15 +369,31 @@ export class VSCodeNotebookController implements Disposable {
...
+        void execution.clearOutput(cell);
+
+        // Connect to a matching kernel if possible (but user may pick a different one)
+        try {
+            const kernel = await this.connectToKernel(doc);
+            this.updateKernelInfoInNotebookWhenAvailable(kernel, doc);
+            return await kernel.executeCell(cell);
+        } catch (ex) {
+            // If there was a failure connecting or executing the kernel, stick it in this cell
+            return displayErrorsInCell(cell, execution, ex.toString());
```
DonJayamanne, line 387: This doesn't look good.
We're just dumpgin a call stack.
I"d like us to analyse the error and provide better information instead of just using a toString

DonJayamanne: I don't think we had any code that dumped the raw errors like this, this is new code and would be a poor experience

src/client/datascience/raw-kernel/rawJupyterSession.ts
```diff
@@ -122,19 +121,6 @@ export class RawJupyterSession extends BaseJupyterSession {
...
                 throw error;
-            } else if (error instanceof IpyKernelNotInstalledError) {
-                sendKernelTelemetryEvent(
-                    this.resource,
-                    Telemetry.RawKernelSessionStart,
-                    stopWatch.elapsedTime,
-                    undefined,
-                    error
-                );
-                sendKernelTelemetryEvent(this.resource, Telemetry.RawKernelSessionStartNoIpykernel, {
```
DonJayamanne, line 133: We need this telemetry, couldn't find it being sent anywhere.

src/client/datascience/raw-kernel/rawJupyterSession.ts
```diff
@@ -262,19 +248,28 @@ export class RawJupyterSession extends BaseJupyterSession {
...
         this.terminatingStatus = undefined;
-        const process = await this.kernelLauncher.launch(
-            this.kernelConnectionMetadata,
-            this.launchTimeout,
+        const process = await KernelProgressReporter.wrapAndReportProgress(
             this.resource,
-            this.workingDirectory,
-            options.ui,
-            options.token
+            localize.DataScience.connectingToKernel().format(
```
DonJayamanne, line 253: Why do we need `connectingToKernel`?
I thought we already displayed a message when we're connecting to a kernel.
THis just seems yet another but a duplicate.

src/client/datascience/types.ts
```diff
@@ -404,13 +404,29 @@ export interface IInteractiveWindowProvider {
...
+};
+type ErrorKernelErrorResult = {
+    kind: 'Error';
+    error: Error;
+};
+type InstalledKernelErrorResult = {
+    kind: 'Installed';
+};
+type SwitchedKernelErrorResult = {
+    kind: 'Switched';
```
DonJayamanne, line 418: Can we make this more verbosed like `SwitchKernel` `InstalledDependencies` `UnhandledError`
I'm assuming kind=`Error` stands for `UnhandledError`, i.e. we do'nt know what to do about that.

src/test/datascience/errorHandler.unit.test.ts
```diff
@@ -330,7 +325,7 @@ ImportError: No module named 'xyz'
 
             const expectedMessage = DataScience.failedToStartKernelDueToImportFailure().format('xyz');
 
-            verifyErrorMessage(expectedMessage, 'https://aka.ms/kernelFailuresModuleImportErr');
+            verifyErrorMessage(expectedMessage, 'https://aka.ms/kernelFailuresModuleImportErrFromFile');
```
DonJayamanne, line 328: This is the wrong link, the test is now wrong.
If importing `xyz` fails, then this means a module named `xyz` is missing.

src/client/datascience/jupyter/kernels/cellExecution.ts
```diff
@@ -316,32 +303,26 @@ export class CellExecution implements IDisposable {
...
+            // Otherwise it's an error from the kernel itself. Put it into the cell
+            const failureInfo = analyzeKernelErrors(
+                workspace.workspaceFolders || [],
+                error,
+                getDisplayNameOrNameOfKernelConnection(this.kernelConnection),
+                this.kernelConnection.interpreter?.sysPrefix
+            );
+            if (failureInfo) {
+                this.execution
+                    ?.appendOutput([translateErrorOutput(createErrorOutputFromFailureInfo(failureInfo))])
```
DonJayamanne, line 319: Previously error handler would deal with the errors, such as what to do where to display.
Now we have this class that displays the errors in the output & then error handler will display the errors as a popup message.

Can we centralize the code to displaying errors (in output & messagebox) in one place. In the past it was in a few different places and we centralized it to ensure we have a consnstent UX when things fail, and I'm afraid we'll start going down that path again.

**Another observation:**
Assume the kernel failure indicates a module could not be imported or something similar, and we need to perform additional analysis on the error (something that's async & possibly requires user feedback).
Once done, we'll have more information about the error & the context. Next we display the new error (after analysis) information in the output & we'd need to do the exact same thing all over again when displaying the popup error message in the error handler.

The idea behind the error handler was if something goes wrong, we handle it it one place. I know its still possible, but displaying it in two places now means we'd need to do some analysis in two places.  


If we can move it into a class or the like, that would be great.

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -1774,3 +1793,98 @@ export async function executeSilently(session: IJupyterSession, code: string): P
...
+        commandManager
+    )) as boolean;
+    if (selected && editor) {
+        controller = await waitForSelection.promise;
+    }
+    disposable.dispose();
+    return controller;
+}
+
+export async function connectToKernel(
```
DonJayamanne, line 1823: Nice

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -188,25 +172,14 @@ export class Kernel implements IKernel {
...
-            trackKernelResourceInformation(resourceUri, {
-                kernelConnection: kernelConnectionMetadata,
-                isPreferredKernel
-            });
-        }
     }
     private perceivedJupyterStartupTelemetryCaptured?: boolean;
 
-    public addEventHook(hook: (event: 'willRestart' | 'willInterrupt') => Promise<void>): void {
+    public addEventHook(hook: (kernel: IKernel, event: 'willRestart' | 'willInterrupt') => Promise<void>): void {
```
DonJayamanne, line 178: Sending the `IKernel` feels redundant, when the calling code can easily capture that object in context and use it.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -239,7 +212,7 @@ export class Kernel implements IKernel {
         await this.startNotebook(options);
     }
     public async interrupt(): Promise<void> {
-        await Promise.all(this.eventHooks.map((h) => h('willInterrupt')));
+        await Promise.all(this.eventHooks.map((h) => h(this, 'willInterrupt')));
```
DonJayamanne, line 215: Can we not send `this`, not a practice in general, as the calling code can easily figure this out

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -425,95 +377,71 @@ export class Kernel implements IKernel {
...
-                        this.resourceUri,
-                        Telemetry.NotebookStart,
-                        stopWatch.elapsedTime,
-                        undefined,
-                        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                        ex as any
+            this._notebookPromise = this.createNotebook(new StopWatch())
+                .then((n) => {
+                    // If notebook session is disposed, make sure to clear the promise
+                    n.session.onDidDispose(
```
DonJayamanne, line 383: This doesn't look right, session is a child component of Notebook.
If anything the Notebook should monitor the session and dispose itself when its child component session gets disposed.
This way, we don't need to monitor components n levels deep

src/client/datascience/jupyter/kernels/kernelCommandListener.ts
```diff
@@ -142,12 +151,47 @@ export class KernelCommandListener implements IDataScienceCommandListener {
...
+                    ex = new CancellationError(
+                        DataScience.canceledKernelHeader().format(
+                            kernel.kernelConnectionMetadata.interpreter?.displayName || ''
+                        )
+                    );
+                    break;
+            }
+            if (currentCell) {
+                const cellExecution = CellExecutionCreator.getOrCreate(currentCell, kernel.controller);
+                displayErrorsInCell(currentCell, cellExecution, ex).ignoreErrors();
```
DonJayamanne, line 194: Looking at the code in `displayErrorsInCell` we could end up dumping the error object as is, as we're expecting a string as the 3rd argument.

src/client/datascience/jupyter/kernels/types.ts
```diff
@@ -182,8 +186,8 @@ export interface IKernel extends IAsyncDisposable {
      * Executes arbitrary code against the kernel without incrementing the execution count.
      */
     executeHidden(code: string): Promise<nbformat.IOutput[]>;
-    addEventHook(hook: (event: 'willRestart' | 'willInterrupt') => Promise<void>): void;
-    removeEventHook(hook: (event: 'willRestart' | 'willInterrupt') => Promise<void>): void;
+    addEventHook(hook: (kernel: IKernel, event: 'willRestart' | 'willInterrupt') => Promise<void>): void;
+    removeEventHook(hook: (kernel: IKernel, event: 'willRestart' | 'willInterrupt') => Promise<void>): void;
```
DonJayamanne, line 190: Can we remove `IKernel` as the sender, feels redundant, we don't do this in event handlers & other places

src/client/datascience/kernel-launcher/kernelLauncher.ts
```diff
@@ -104,23 +102,9 @@ export class KernelLauncher implements IKernelLauncher {
         timeout: number,
         resource: Resource,
         workingDirectory: string,
-        ui: IDisplayOptions,
         cancelToken: CancellationToken
     ): Promise<IKernelProcess> {
         const promise = (async () => {
-            // If this is a python interpreter, make sure it has ipykernel
```
DonJayamanne, line 111: Are we planning on just assuming things are installed and attempt to start the kernel?
I.e. are we planning on figuring out if ipykernel is installed or not only based on kernel failure?

I think that could be a little dangerous. I'm very concerned that analyzing errors to determine whether ipykernel is installed or not  might not be bullet proof.

Also, in the case of non-raw we call `ensureKernelIsUsable` which in turn calls `installMissingDependencies`.
Hence we should have a consistent approach for both.

src/client/datascience/jupyter/kernels/cellExecution.ts
DonJayamanne: Note: This isn't a must, but would be great if we could achieve this some how, 
After all, this is a huge step forward in improving the code.

src/client/datascience/notebook/vscodeNotebookController.ts
DonJayamanne: E.g. if the kernel dies half way, then dumping a call stack in the output is a poor ux, and that's one of the things we tried to fix. I don't think we should be displaying node.js stack trace at all.

DonJayamanne (changes requested):
A few items

src/client/datascience/jupyter/kernels/helpers.ts
rchiodo: Yes we do. Good catch

src/client/datascience/jupyter/kernels/kernel.ts
rchiodo: I like it this way (follows the .NET event way of doing things) but it doesn't really matter. I can remove it too.

src/client/datascience/jupyter/kernels/kernelCommandListener.ts
rchiodo: I believe you're mistaken. ex.toString does not include the stack. Just the message.

src/client/datascience/kernel-launcher/kernelLauncher.ts
rchiodo: Jupyter is different in that we can't let the kernel startup fail as we can't really parse the output of the failure. But for raw, this speeds up kernel startup by like 3 seconds. Failure should be the odd ball case.

Perhaps the failure analyzer could use the kernelDependencyService to detect instead of looking at the message.

src/client/datascience/notebook/vscodeNotebookController.ts
rchiodo: I don't think ```ex.toString()``` includes the callstack.

I wrote the following code:

```javascript
function throwStuff() {
   throw new Error('Here is an error');
}

function doStuff() {
   try {
      throwStuff();
   }
   catch (ex) {
      console.log(ex.toString());
      console.log('Here is the stack');
      console.log(ex.stack);
   }
}

doStuff();
```

It output this:

```
Error: Here is an error
Here is the stack
Error: Here is an error
    at throwStuff (C:\Users\aku91\test.js:2:10)
    at doStuff (C:\Users\aku91\test.js:7:7)
    at Object.<anonymous> (C:\Users\aku91\test.js:16:1)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47
```

So the callstack only comes out when asked for.

src/client/datascience/raw-kernel/rawJupyterSession.ts
rchiodo: Yeah you're right. I think I might have to send this higher up in the 'connectToKernel' function. The intent of this is to indicate somebody got the prompt, I think? At least the reason returns what choice they made.

src/client/common/process/environmentActivationService.ts
amunger: this CancellationTokenSource isn't cancelled anywhere. Wouldn't it be better to not create it in the first place and not force the user to pass a token to wrapAndReportProgress?

src/client/datascience/interactive-window/interactiveWindow.ts
```diff
@@ -150,55 +147,175 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
+            // Also start connecting to our kernel but don't wait for it to finish
+            this.startKernel(editor.document, preferredController).ignoreErrors();
+        }
+
+        return editor;
+    }
+
+    private async startKernel(notebook: NotebookDocument, controller: VSCodeNotebookController): Promise<void> {
+        if (controller.id !== this._kernelConnectionId) {
+            this._kernelConnectionId = controller.id;
```
amunger, line 172: When will this be false? Is it when we're already connected to the kernel? Will there be a "connecting" sys info left behind in that case, and should we ensure that the _kernelPromise is resolved?

rchiodo: This is false on the first startup and if the kernel fails to have ipykernel and the user decides to switch to a new controller. Then it will reenter here and try again.

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: Thanks, sorry, but would prefer to change to standard node style, as this is very much non standard, e.g. even in VS Codes style for events (event emitter) we don't send the `sender`.

src/client/datascience/jupyter/kernels/kernelCommandListener.ts
DonJayamanne: Yes, you are right, thanks

DonJayamanne: Sorry

src/client/datascience/kernel-launcher/kernelLauncher.ts
DonJayamanne: > erhaps the failure analyzer could use the kernelDependencyService to detect instead of looking at the message.

Yes, that's what the error handler did. Having the analyzer do that would also work.

DonJayamanne: This raises a concern I had
* Based on the output we display errors in the cell output using displayErrorInCell
* Later we display the error message in the error handler to display a messagebox

Basically we have two places where we decide what to do with the errors, can we centralize this (that's what the ErrorHandler did).
Would be great if we can merge `displayErrorInCell` with the `handleError` in the errorhandler.
This way, what we do with the errors is handled in a single location.
Whether we display errors in the cell output and/or display a message box is all in one spot - ie. error handling is centralized

rchiodo: That doesn't work because there are two ways to handle errors. 
- Interactive window
- Regular notebook

The caller of the error handler has to know where to put the error.

rchiodo: Yeah it does seem weird to show the message info in the error handler though. Feels more like the error handler should just return the failure info and let the calling code decide what to do with it.

src/client/common/process/environmentActivationService.ts
rchiodo: This cancellationTokenSource would be canceled if the user clicked the 'cancel' button the 'reportProgress'.

src/client/datascience/kernel-launcher/kernelLauncher.ts
rchiodo: I think we should discuss options for 'where' to display the errors. There are two cases at least (interactive and notebook).

I'm not sure the errorHandler is the right spot to show any UI

src/client/datascience/notebook/vscodeNotebookController.ts
rchiodo: executeCell gets called for all cells in a 'Run All' situation. This connectToKernel was chained so that only one of them would actually attempt to get the kernel at a time.

I'll change the comment to say that.

src/client/datascience/raw-kernel/rawJupyterSession.ts
rchiodo: I moved this into the 'connectToKernel' function.

DonJayamanne (approved):

amunger (approved):

IanMatthewHuff (approved):

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -139,7 +127,7 @@ export class Kernel implements IKernel {
     private _disposed?: boolean;
     private _disposing?: boolean;
     private _ignoreNotebookDisposedErrors?: boolean;
-    private readonly _kernelSocket = new Subject<KernelSocketInformation | undefined>();
+    private readonly _kernelSocket = new ReplaySubject<KernelSocketInformation | undefined>();
```
DonJayamanne, line 130: Why is this a ReplaySubject instead of Subject?

src/client/datascience/interactive-common/notebookProvider.ts
```diff
@@ -66,7 +66,7 @@ export class NotebookProvider implements INotebookProvider {
             if (
                 !(await this.jupyterNotebookProvider.connect({
                     resource: options.resource,
-                    token: options.token,
+                    tokenSource: options.tokenSource,
```
DonJayamanne, line 69: I missed this. Why are we passing the source object. The lower layers should not be able to cancel it self (via the token), only the calling code should be able to cancel the lower calls.

DonJayamanne (changes requested):
A few questinos

src/client/datascience/jupyter/kernels/kernel.ts
rchiodo: Because the kernel finishes creation before the kernelsocket is observed. When the controller signs up for it later, it needs the current value. ReplaySocket does that when something first subscribes.

src/client/common/process/environmentActivationService.ts
```diff
@@ -143,9 +143,14 @@ export class EnvironmentActivationService implements IEnvironmentActivationServi
         const title = DataScience.activatingPythonEnvironment().format(
             interpreter.displayName || getDisplayPath(interpreter.path)
         );
-        return KernelProgressReporter.wrapAndReportProgress(resource, title, () =>
-            this.getActivatedEnvironmentVariablesImpl(resource, interpreter)
-        );
+        const cancellationTokenSource = new CancellationTokenSource();
+        try {
+            return await KernelProgressReporter.wrapAndReportProgress(resource, title, cancellationTokenSource, (_t) =>
```
DonJayamanne, line 148: I agree with @amunger If we didn't have to pass the token this code would be simpler, now we need to pass a token everytime even when not required.

src/client/datascience/interactive-common/notebookServerProvider.ts
```diff
@@ -81,7 +81,11 @@ export class NotebookServerProvider implements IJupyterServerProvider {
         const property = options.localJupyter ? 'local' : 'remote';
         if (!this.serverPromise[property]) {
             // Start a server
-            this.serverPromise[property] = this.startServer(options.resource, options.token, options.localJupyter);
+            this.serverPromise[property] = this.startServer(
+                options.resource,
+                options.tokenSource,
```
DonJayamanne, line 86: Same as other places, this doesn't seem right, i.e passing the tokenSource

src/client/datascience/interactive-common/notebookProvider.ts
rchiodo: No that's not true. For window.withProgress to work, it needs to call cancel. It isn't passed a cancel token, it passes out a cancel token to the progress function. We implement this progress function and we need to look at its cancel value. The tokenSources are passed into this progress reporter.

src/client/datascience/interactive-common/notebookServerProvider.ts
rchiodo: See my other comment. We need a tokenSource now. Perhaps in some places we don't care, but I sided on the forcing it to be passed.

src/client/datascience/interactive-common/notebookProvider.ts
rchiodo: THis isn't actually relevant to this refactor but it was mixed in with my change from the install stuff. It was relevant there as the cancel button wasn't doing anything (it didn't actually cancel the install of a module)

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -1774,3 +1807,201 @@ export async function executeSilently(session: IJupyterSession, code: string): P
...
+    let controller: VSCodeNotebookController | undefined;
+    const waitForSelection = createDeferred<VSCodeNotebookController>();
+    const disposable = controllerManager.onNotebookControllerSelected((e) => waitForSelection.resolve(e.controller));
+
+    const selected = (await selectKernel(
+        resource,
+        notebooks,
+        serviceContainer.get(IInteractiveWindowProvider),
+        commandManager
+    )) as boolean;
```
DonJayamanne, line 1829: I'd prefer to avoid the unnecessary casting, i don' tthink its necessary

src/client/datascience/progress/kernelProgressReporter.ts
```diff
@@ -186,28 +202,38 @@ export class KernelProgressReporter implements IExtensionSyncActivationService {
...
+                            info.progressList.push(message);
+                            progress.report({ message });
+                        }
+                    }
+                    await Promise.race([
+                        createPromiseFromCancellation({ token, cancelAction: 'resolve', defaultValue: true }),
+                        deferred.promise
+                    ]);
+                    if (token.isCancellationRequested) {
+                        info.tokenSources.forEach((t) => t.cancel());
```
rchiodo, line 228: This is where the tokenSources are all canceled. This didn't used to do anything.

src/client/datascience/jupyter/kernels/helpers.ts
rchiodo: Yeah not sure why this is here. Seems unnecessary. I'll remove it.

src/client/datascience/interactive-common/notebookProvider.ts
DonJayamanne: Ok, looking at the code, in KernelProgressReporter, we aren't displayhing the cancel button at all.
I.e. the first for `window.withProgress` doesn't contain the value `cancellable`
see here 
```typescript
            void window.withProgress(
                { location: ProgressLocation.Notification, title },
```

DonJayamanne: Ie. this change isn't going to do anything.
Also, do we want a cancel button? I'm not against it, feels ok, but is a UX change and wanted to double check you wanted that.

src/client/datascience/progress/kernelProgressReporter.ts
rchiodo: Yeah I was using this in the install PR to allow canceling of the install of a module. Perhaps I should create a new interface for that instead. 

I'll back out the tokenSource changes for now.

src/client/datascience/raw-kernel/rawJupyterSession.ts
rchiodo: We didn't show this in some cases. It's just blank. And if this is redundant, it should have no effect since these things stack.

src/client/datascience/interactive-common/notebookProvider.ts
rchiodo: Should be removed now. I'll figure out something different when I implement the install change.

DonJayamanne (approved):

Issue #9089, opened 2022-02-21
Shell commands such as !pip !python !where python do not point to the right executable


DonJayamanne:
This has been reported a few times and we've take the route of recommending users use `%pip` and the like, and didn't really fix the underlying issue.

Solution: Put the python executable directory in front of the PATH entries

joyceerhl:
@DonJayamanne Do you have any specific verification steps?

rzhao271:
I get a list of Pythons when I run `!where python` in the IW but I don't think it found the venv one:

![A list of Pythons in the IW](https://user-images.githubusercontent.com/7199958/156078597-ece171c6-2c9a-4303-ae92-1c72eddd803a.PNG)

DonJayamanne:
@rzhao271 can you confirm you have the latest pre-release version of the Jupyter extension?

rzhao271:
It works!

PR #9163, opened 2022-03-01
Get activate env vars for non-conda as well

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/9089

### DIFF
--- a/src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
+++ b/src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
@@ -9,7 +9,7 @@ import { noop } from '../../common/utils/misc';
 import { IEnvironmentVariablesProvider, IEnvironmentVariablesService } from '../../common/variables/types';
 import { IEnvironmentActivationService } from '../../interpreter/activation/types';
 import { IInterpreterService } from '../../interpreter/contracts';
-import { EnvironmentType, PythonEnvironment } from '../../pythonEnvironments/info';
+import { PythonEnvironment } from '../../pythonEnvironments/info';
 import { IJupyterKernelSpec } from '../types';
 
 @injectable()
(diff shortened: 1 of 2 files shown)


src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
```diff
@@ -51,7 +51,7 @@ export class KernelEnvironmentVariablesService {
 
         let [customEditVars, interpreterEnv] = await Promise.all([
             this.customEnvVars.getCustomEnvironmentVariables(resource).catch(noop),
-            interpreter && interpreter.envType == EnvironmentType.Conda
```
DonJayamanne, line 54: This was wrong, we should get activated env variables for Virtual Env, etc, not just Conda environments.

src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
```diff
@@ -62,6 +62,14 @@ export class KernelEnvironmentVariablesService {
         ]);
         if (!interpreterEnv && Object.keys(customEditVars || {}).length === 0) {
             traceInfo('No custom variables nor do we have a conda environment');
+            // Ensure the python env folder is always at the top of the PATH, this way all executables from that env are used.
+            // This way shell commands such as `!pip`, `!python` end up pointing to the right executables.
+            // Also applies to `!java` where java could be an executable in the conda bin directory.
+            if (interpreter) {
+                const env = kernelEnv || process.env;
+                this.envVarsService.prependPath(env, path.dirname(interpreter.path));
+                return env;
```
DonJayamanne, line 71: Fall back, if we do'nt have any env variables, then prefix PATH as we do in other cases.

DonJayamanne: This will work for global interpreters

src/client/datascience/kernel-launcher/kernelEnvVarsService.ts
DonJayamanne: This will work for virtual environments

rchiodo (approved):

PR #9178, opened 2022-03-01
Remove unnecessary extra parameter

### LINKED ISSUE #5144: Do we need second argument in concatMultilineString
here are the places we pass in a second argument.
I know these were removed in all other places as it was incorrect to trim the values (output).
I feel we may have forgotten about others and accidentally left them.

Suggestion is to clean this up, as its not clear when & why this is required, seems more like trial and error (i.e. why do we need this, when do we need this, do we need this in notebooks, etc).

```typescript

        // For un-executed output we might get text or svg output as multiline string arrays
        // we want to concat those so we don't display a bunch of weird commas as we expect
        // Single strings in our output
        if (Array.isArray(data)) {
            data = concatMultilineString(data as nbformat.MultilineString, true);
        }

        // Fixup latex to make sure it has the requisite $$ around it
        if (mimeType === 'text/latex') {
            data = fixMarkdown(concatMultilineString(data as nbformat.MultilineString, true), true);
        }

```

In Debugger code
```typescript
    private extractOutput(cell: ICell): string | undefined {
        if (cell.state === CellState.error || cell.state === CellState.finished) {
            const outputs = cell.data.outputs as nbformat.IOutput[];
            if (outputs.length > 0) {
                const data = outputs[0].data;
                if (data && data.hasOwnProperty('text/plain')) {
                    return concatMultilineString((data as any)['text/plain']);
                }
                if (outputs[0].output_type === 'stream') {
                    const stream = outputs[0] as nbformat.IStream;
                    return concatMultilineString(stream.text, true);
                }
            }
        }
        return undefined;
    }
```

rchiodo:
Fixes #5144

### DIFF
--- a/news/3 Code Health/5144.md
+++ b/news/3 Code Health/5144.md
@@ -1,0 +1,1 @@
+Eliminate unused parameter for concatMultilineString.
--- a/src/client/datascience/jupyter/kernels/kernel.ts
+++ b/src/client/datascience/jupyter/kernels/kernel.ts
@@ -987,7 +987,7 @@ export function getPlainTextOrStreamOutput(outputs: nbformat.IOutput[]) {
         }
         if (outputs[0].output_type === 'stream') {
             const stream = outputs[0] as nbformat.IStream;
-            return concatMultilineString(stream.text, true);
+            return concatMultilineString(stream.text);
         }
     }
     return;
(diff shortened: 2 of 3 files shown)


rchiodo:
This was only used to get the output for connecting to debugpy during IW debugging. IW debugging works without it.

DonJayamanne (approved):

PR #9179, opened 2022-03-02
Update release plan

DonJayamanne:


.github/release_plan.md
```diff
@@ -5,27 +5,24 @@
 
 # Release candidate (Friday of VS Code endgame week, XXX XX)
 
--   [ ] Update [Component Governance](https://dev.azure.com/vscode-python-datascience/vscode-python-datascience/_componentGovernance) (Click on "microsoft/vscode-jupyter" on that page). Notes are in the OneNote under Python VS Code -> Dev Process -> Component Governance.
-    -   [ ] Provide details for any automatically detected npm dependencies
+-   [ ] Review [Component Governance](https://dev.azure.com/monacotools/Monaco/_componentGovernance/191876?_a=alerts&typeId=10819888&alerts-view-option=active) (Click on "microsoft/vscode-jupyter" on that page). Notes are in the [OneNote](https://microsoft.sharepoint.com/teams/python/_layouts/15/Doc.aspx?sourcedoc=%7B30d33826-9f98-4d3e-890e-b7d198bbbcbe%7D&action=edit&wd=target(Python%20VS%20Code%2FDev%20Process.one%7C685f1141-e5eb-45a3-adcf-dd2504c21ca2%2FCG%5C%2FTPN%7Cbb1ca24f-7cd5-451f-b2b4-ef310ec8b903%2F)&share=IgEmONMwmJ8-TYkOt9GYu7y-AeCM6R8r8Myty0Lj8CeOs4E) under Python VS Code -> Dev Process -> Component Governance.
```
amunger, line 8: It looks like Component Governance OneNote page was renambed to "CG/TPN".
But it seems like the important thing is just to resolve all High/Severe issues which you could just state here

.github/release_plan.md
```diff
@@ -5,27 +5,24 @@
...
     -   [ ] Manually add any repository dependencies (if you can't add manually, refer [here](https://docs.opensource.microsoft.com/tools/cg/features/cgmanifest/)). Only add a cgmanifest.json if the components are not NPM or are not dev only.
--   [ ] Create new release branch with format `release-YYYY.MM.100`
+-   [ ] Create new release branch with format `release/release-YYYY.MM.100`
     -   [ ] Mark release branch as not deletable (add it explicitly here, https://github.com/microsoft/vscode-jupyter/settings/branches)
```
amunger, line 11: I also meant to ask about this step - those settings match all release branches, so there's no reason to add it explicitly, but they also currently allow the branch to be deleted. Should the step say to alter those settings to prevent the branch from being deleted?

amunger: There's also a step in here about deleting old branches, which wouldn't be possible if that change was made

DonJayamanne: > hich wouldn't be possible if that change was made

You would need to undo that change and then delete it.

amunger (approved):

Issue #8975, opened 2022-02-10
Refactor test code to go under a single 'test' folder


rchiodo:
As discussed in our architecture review, this task is to move 'test' related code into a 'test' folder (or whatever you decided to name it).

The main reason for this is to ensure we never reference test related code inside of production code. One way to achieve this is to have a completely separate 'test' folder at the root (instead of under src).

This would include stuff like:
- standardTest.ts
- index.ts
- all *.vscode.test.ts files
- test helper code

It would NOT include:
- all unit tests. These would reside next to the code they're testing

DonJayamanne:
> One way to achieve this is to have a completely separate 'test' folder at the root (instead of under src).

Not moving all of the tests, no need we can use linters to achieve the same requirement.

rchiodo:
> Not moving all of the tests, no need we can use linters to achieve the same requirement.

Yeah I tried this and it was a huge pain having two root dirs in the tsconfig. Changed relative paths everywhere.

DonJayamanne:
Done as per discussion

PR #9180, opened 2022-03-02
Prevent importing test files from extension code

DonJayamanne:
Part of #8975


rchiodo (approved):
Woot! Didn't think it would be that easy

amunger (approved):

PR #9185, opened 2022-03-02
Do not specify node versions

DonJayamanne:
The azure pipelines now use the same version as we do which is the same version shipped in VS Code.

### DIFF
--- a/build/azure-pipeline.pre-release.yml
+++ b/build/azure-pipeline.pre-release.yml
@@ -23,10 +23,6 @@ extends:
   template: azure-pipelines/extension/pre-release.yml@templates
   parameters:
     buildSteps:
-      - task: NodeTool@0
-        inputs:
-          versionSpec: '14.16'
-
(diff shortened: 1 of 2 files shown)


build/azure-pipeline.pre-release.yml
```diff
@@ -23,10 +23,6 @@ extends:
   template: azure-pipelines/extension/pre-release.yml@templates
   parameters:
     buildSteps:
-      - task: NodeTool@0
```
rchiodo, line 26: Not sure why? I assume the latest node works too but why not the older one?

rchiodo (approved):

build/azure-pipeline.pre-release.yml
DonJayamanne: Got the azure pipelines (upstream) to use the same version as VS Code, this way we don't have to specify this and everyone uses the right version

Issue #8988, opened 2022-02-10
Try Except all startup code 


rchiodo:
I think the simplest fix here is to just fix the bug where we're using ```'``` to wrap a path in the startup code. However @amunger had another idea. We could wrap each individual statement in a ```try:except:``` block to prevent one part of startup breaking other pieces.

_Originally posted by @rchiodo in https://github.com/microsoft/vscode-jupyter/issues/8957#issuecomment-1035247483_

IanMatthewHuff:
https://github.com/microsoft/vscode-jupyter/pull/9197 resolved here.

Change was not to fully try except everything. Just parts that we expect might fail. However now we log and send telemetry for failures in our startup code, user specified startup code, variable fetching code, and IW debug configuration code.

rchiodo:
/verified

I get this now (if I make an error) but startup still works:
```
Error 13:42:42: Error executing user defined startup code print(x) NameError: name 'x' is not defined 
     [1;31m---------------------------------------------------------------------------[0m
    [1;31mNameError[0m                                 Traceback (most recent call last)
    Input [1;32mIn [1][0m, in [0;36m<cell line: 1>[1;34m()[0m
[1;32m----> 1[0m [38;5;28mprint[39m(x)
```
    [1;31mNameError[0m: name 'x' is not defined

PR #9197, opened 2022-03-03
Append SVG to figure formats

### LINKED ISSUE #9191: Our plot settings should only append SVG if needed. Not mess with existing figure_formats settings
Our SVG configuration code should not replace user defaults. It should just append on SVG if requests.

IanMatthewHuff:
Fixes #9191 #8988


src/datascience-ui/common/constants.ts
```diff
@@ -584,10 +585,8 @@ export namespace CodeSnippets {
...
     export const ImportIPython = '{0}\nfrom IPython import get_ipython\n\n{1}';
-    export const MatplotLibInitSvg = `import matplotlib\n%matplotlib inline\n${Identifiers.MatplotLibDefaultParams} = dict(matplotlib.rcParams)\n%config InlineBackend.figure_formats = ['svg', 'png']`;
-    export const MatplotLibInitPng = `import matplotlib\n%matplotlib inline\n${Identifiers.MatplotLibDefaultParams} = dict(matplotlib.rcParams)\n%config InlineBackend.figure_formats = ['png']`;
-    export const ConfigSvg = `%config InlineBackend.figure_formats = ['svg', 'png']`;
-    export const ConfigPng = `%config InlineBackend.figure_formats = ['png']`;
+    export const MatplotLibInit = `import matplotlib\n%matplotlib inline\n${Identifiers.MatplotLibDefaultParams} = dict(matplotlib.rcParams)\n`;
+    export const AppendSVGFigureFormat = `import matplotlib_inline.backend_inline\n${Identifiers.MatplotLibFigureFormats} = matplotlib_inline.backend_inline.InlineBackend.instance().figure_formats\n${Identifiers.MatplotLibFigureFormats}.add('svg')\nmatplotlib_inline.backend_inline.set_matplotlib_formats(*${Identifiers.MatplotLibFigureFormats})`;
```
IanMatthewHuff, line 589: Found some interesting stuff looking this up. Using %config to try to fetch the backend formats actually doesn't work until you fully initialize %matplotlib inline, which I didn't want to do for this. Also the figure_format specification has changed a few times and can be a bit fussy between python versions about [] or {} syntax as it was moved from a list to a set at one point.

So instead moved to using the backend_inline classes directly.

IanMatthewHuff: Handy working in the open source world since I could find the impls here https://github.com/ipython/matplotlib-inline/blob/541b26da2f49d0403994b91cb062e8c438920c44/matplotlib_inline/backend_inline.py

rchiodo (approved):
Using their code directly makes me nervous. What if the parameters/objects change?

rchiodo:
I'm also thinking this has the potential to make all of the startup code fail. Perhaps we should put try/except around each section of startup code (we have another issue to do that).

IanMatthewHuff:
> Using their code directly makes me nervous. What if the parameters/objects change?

@rchiodo 
I don't disagree with ya. This function moved location once I believe and the figure format in the config file has changed as well. From what I'd seen the function (since it's listed as a public API) actually felt better than using the old %config method. As the config is actually less of a documented promise and more of a file tweaking helper. On github I see a lot of usage of the function versus config magic.

On the plus side this code now only happens when you add on SVG as opposed to all the time. Lemme look at the safety aspect, maybe I can either roll that into this PR or submit another issue with that fixed.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -883,13 +881,14 @@ export class Kernel implements IKernel {
...
-            traceInfoIfCI(
-                `Initialize config for plots for ${(this.resourceUri || this.notebookDocument.uri).toString()}`
-            );
-            results.push(...configInit.splitLines({ trim: false }));
         }
+
+        // Add in SVG to the figure formats if needed
+        if (settings.generateSVGPlots) {
+            results.push(...CodeSnippets.AppendSVGFigureFormat.splitLines({ trim: false }));
+            traceInfo('Add SVG to matplotlib figure formats');
```
DonJayamanne, line 889: I can no see how startup code will fail completely if users don't have matplotlib installed.
Importing packages too would fail as we won't be adding the site packages folder into the current path, i'm going to look into that now.

DonJayamanne (approved):

src/client/datascience/jupyter/kernels/kernel.ts
IanMatthewHuff: @DonJayamanne Just want to make sure we are not looking at the same thing. I was also going to take a peek at hardening the startup code against exceptions as part of this PR (it's why I've not pushed yet). Couldn't quite tell from your comment above if you were as well.

IanMatthewHuff:
@DonJayamanne @rchiodo . Looked at the error cases here. I believe that Rich's suggestion makes the most sense. Each section of startup code should have its own wrapping to handle exceptions. It all gets combined into one cell, so right now an error at any point (ordering matters) can throw off the rest of startup code.

Plan is to make a quick helper function to wrap any set of code in try / except.

rchiodo (approved):

rchiodo:
I feel like this is safer. We should probably do something extra for startup commands (like show a popup or something), but I think we can do that later.

IanMatthewHuff:
> I feel like this is safer. We should probably do something extra for startup commands (like show a popup or something), but I think we can do that later.

Yeah the startup commands are tricky. Not sure how to best handle them as they are not language specific like the rest of the python startup code.

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: Didn't realize you were doing the same thing, FYI we have a seprate issue for hardening startup code here https://github.com/microsoft/vscode-jupyter/issues/8988
I'll close my PR.
I don't think we should wrap all startup code in try catch. I think something like what I've done would be better, so that we log the errors, Else we'd never know what went wrong.
I think should put a try..except only around the matplot lib code (as users might not have matplot lib).
& the rest of code should not fail except for initializing of the `__vsc_ipynb_file__ ` variable

As for user code, I think we shouldn't wrap that either. I think that should be a seprate request.
If there are failures it should be logged (i don't think we shoudl display messageboxes), and don't think we should add try...except for user code, might not be what they're expecting, if its their code, they should be responsible for that.

DonJayamanne: I looked at that issue as I felt that if users don't have matplot lib then debugger code will not get setup, hence debugging will not work, i.e. that's a regression.
Hence picked that.

IanMatthewHuff: @DonJayamanne I wasn't saying to can your PR, I just missed the mail on it and I was working on something similar. I had the same realization about the matplotlib code (seeing that it could error out and mess up the debugger / user code that comes after it) so I didn't want to go forward with this PR without that fixed. I'd rather do the right thing on this PR so it can be submitted in good shape. Since you closed your PR I'd go ahead with this but I'll look at your PR logging changes and look at integrating that into this PR as well.

I think with the try excepts it does get logged though, at least on CI the output stream from the print on the except message is show in the log.

IanMatthewHuff:
Combined with Don's PR and some other review comments

rchiodo (approved):

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -1803,10 +1808,50 @@ export async function executeSilently(session: IJupyterSession, code: string): P
...
+            if (errorOptions.traceErrors) {
+                traceError(fullMessage);
+            } else {
+                traceWarning(fullMessage);
+            }
+
+            // Send telemetry if requested, no traceback for PII
+            if (errorOptions.logTelemetryErrors) {
+                sendTelemetryEvent(Telemetry.KernelStartupCodeFailure, undefined, {
+                    ename: errorOutput.ename,
```
DonJayamanne, line 1848: Would be good to have some context, whether this is user code, startup code, or debugger variable request code, etc

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -748,3 +786,10 @@ export function getPlainTextOrStreamOutput(outputs: nbformat.IOutput[]) {
     }
     return;
 }
+
+// Options for error reporting from kernel silent execution
+export type SilentExecutionErrorOptions = {
+    traceErrors?: boolean;
+    traceErrorsMessage?: string;
+    logTelemetryErrors?: boolean;
```
DonJayamanne, line 794: Should we send the name of the telemetry event here?
This way we have finer grained telemetry, e.g. we can tell from the calling code whether to log telemetry for startup code, user startup code, for variable fetching, etc

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: All good, 

> t does get logged though, at least on CI the output stream from the print on t

Awesome.

DonJayamanne: No harm done, there are 600 more issues :) to work on.

IanMatthewHuff:
@DonJayamanne Since you are out, I'm checking in, but if you have some additional changes I can roll them in next week.

PR #9201, opened 2022-03-03
Send each startup code as individual requests

### LINKED ISSUE #8988: Try Except all startup code

DonJayamanne:
Fixes #8988

### DIFF
--- a/src/client/datascience/jupyter/kernels/helpers.ts
+++ b/src/client/datascience/jupyter/kernels/helpers.ts
@@ -38,7 +38,7 @@ import { DataScience } from '../../../common/utils/localize';
 import { Commands, Settings, Telemetry } from '../../constants';
 import { concatMultilineString } from '../../../../datascience-ui/common';
 import { sendTelemetryEvent } from '../../../telemetry';
-import { traceError, traceInfo, traceInfoIfCI, traceVerbose } from '../../../common/logger';
+import { traceError, traceInfo, traceInfoIfCI, traceVerbose, traceWarning } from '../../../common/logger';
 import {
     areInterpreterPathsSame,
     getInterpreterHash,
@@ -1734,7 +1734,11 @@ export async function sendTelemetryForPythonKernelExecutable(
     traceVerbose('End sendTelemetryForPythonKernelExecutable');
 }
 
-export async function executeSilently(session: IJupyterSession, code: string): Promise<nbformat.IOutput[]> {
+export async function executeSilently(
+    session: IJupyterSession,
+    code: string,
+    errorMessageForLogging: string = ''
+): Promise<nbformat.IOutput[]> {
     traceInfo(`Executing silently Code (${session.status}) = ${code.substring(0, 100).splitLines().join('\\n')}`);
     // eslint-disable-next-line @typescript-eslint/no-require-imports
     const jupyterLab = require('@jupyterlab/services') as typeof import('@jupyterlab/services');
@@ -1750,6 +1754,7 @@ export async function executeSilently(session: IJupyterSession, code: string): P
         true
     );
     const outputs: nbformat.IOutput[] = [];
+    const errorMessages: string[] = [];
     request.onIOPub = (msg) => {
         if (jupyterLab.KernelMessage.isStreamMsg(msg)) {
             traceInfoIfCI(`Got io pub message (stream), ${msg.content.text.substr(0, 100).splitLines().join('\\n')}`);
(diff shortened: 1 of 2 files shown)


src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -675,7 +675,12 @@ export class Kernel implements IKernel {
         if (this.kernelConnectionMetadata.kind !== 'connectToLiveKernel') {
             // Gather all of the startup code at one time and execute as one cell
             const startupCode = await this.gatherStartupCode(notebookDocument);
-            await this.executeSilently(startupCode);
+            // Execute each startup code one at a time, as its possible that some startup code could fail.
+            // If something fails, we don't want the rest to fail, this way
+            // A single request could fail and the rest will continue on.
+            await startupCode.map((code) =>
+                this.executeSilently([code]).catch((ex) => traceVerbose(`Failed to execute startup code`, ex))
```
rchiodo, line 682: This makes the whole thing take a lot longer. I think the code should each be wrapped in a try/except block instead.

rchiodo: It's also not clear you can split it this way. The startup code may need to be grouped together to run as each line is a separate entry.

rchiodo (changes requested):
I think changing getStartupCode to try/except each piece would be better.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -890,7 +889,8 @@ export class Kernel implements IKernel {
             );
             results.push(...configInit.splitLines({ trim: false }));
         }
-        return results;
+        // Wrap in try..except in case user doesn't have matplot lib installed.
+        return ['try:', ...results.map((line) => `    ${line}`), 'except:', '   pass'];
```
DonJayamanne, line 893: The matplot lib code is the only thing that can fail, added try..except around that, 
The rest should never fail, if it does we want things to fail rather than silently failing
I.e. we need to avoid unnecessarily swallowing all exceptions

rchiodo: > The matplot lib code is the only thing that can fail, added try..except around that

There has been other failures in the past. Like the code that sets the '__vscipnb_file_name__'. That failed because of invalid characters in the substitution. 

The same can be said for the startup code. 

And any failure that happens in a executeSilent is never shown to the user. I think we should try/except each one.

rchiodo: Or maybe show them to the user if the fail. That's probably better for the startup commands at least. Maybe that one should be special cased because the user entered it.

IanMatthewHuff: Sorry I thought that I was working on this. I have it added on my PR.

IanMatthewHuff: This is what I was adding. Missed Don's initial PR coming in for some reason. https://github.com/microsoft/vscode-jupyter/pull/9197/commits/a19e1b9682fcfe4bba2a7cba9fc1d6b02b5f2dfb

DonJayamanne: I'm logging the errors instead of displaying them,
If you have startup code today, it will apply to all kernels. anyone who has startup code for Python and runs a non-python kernle will all of a sudden get failure error messages displayed.

We have a seprate issue to have startup code per language, we'd need to address that first, else we coudl end up spamming the user unnecessarily.

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -1770,7 +1780,16 @@ export async function executeSilently(session: IJupyterSession, code: string): P
         }
     };
     await request.done;
-    traceInfo(`Executing silently Code (completed) = ${code.substring(0, 100).splitLines().join('\\n')}`);
 
+    const codeForLogging = `(Code = ${code.substring(0, 100).splitLines().join('\\n')})`;
+    if (errorMessageForLogging && errorMessages.length) {
```
DonJayamanne, line 1785: If we have an error message, then log failure messages as errors, else as warnings.

src/client/datascience/jupyter/kernels/kernel.ts
DonJayamanne: > Sorry I thought that I was working on this. I have it added on my PR.

Sorry, but didn't see the issue has being assigned to anyone hence I started on this.

src/client/datascience/jupyter/kernels/kernel.ts
```diff
@@ -745,21 +750,19 @@ export class Kernel implements IKernel {
...
 
             result.push(...debugCellScripts);
-        }
 
-        // Run any startup commands that we have specified
-        result.push(...this.getStartupCommands());
+            // Set the ipynb file
+            const file = this.resourceUri?.fsPath;
+            if (file) {
+                result.push(`__vsc_ipynb_file__ = "${file.replace(/\\/g, '\\\\')}"`);
```
DonJayamanne, line 763: No try except here, if this fails, we'll end up logging this, better than adding a try except and we'll never ever see any failures.

PR #9249, opened 2022-03-06
Register kernelspecs in private directory

### LINKED ISSUE #9141: When using Jupyter (instead of raw kernels) create Kernelspecs in a custom directory
Today we create the kernelspec json files in the global known Jupyter kernels directory.
**This causes a few issues:**
* Write access https://github.com/microsoft/vscode-jupyter/issues/5494
* Users edit these kernelspecs (but its not meant for users)
* Maintenance issues
* Code is a little complex (as we need to identify what we created vs what we created as a mapping for custom kernels created by users)
* There have been others

**If we create kernelspecs in a custom directory that user isn't aware of:**
* We know that these are created by us 
* These need not EVER be displayed in the kernel picker
* We can blow these away anytime as we don't have to worry about users editing them

**Solution**
* Create these in a folder specific to the extension version (i.e. within the extension directory)

DonJayamanne:
Fixes #9141

### DIFF
--- a/src/client/datascience/jupyter/interpreter/jupyterInterpreterSubCommandExecutionService.ts
+++ b/src/client/datascience/jupyter/interpreter/jupyterInterpreterSubCommandExecutionService.ts
@@ -119,10 +119,10 @@ export class JupyterInterpreterSubCommandExecutionService
         const spawnOptions = { ...options };
         spawnOptions.token = undefined;
         const jupyterDataPaths = (process.env['JUPYTER_PATH'] || '').split(path.delimiter);
-        jupyterDataPaths.push(await this.jupyterPaths.getKernelSpecTempRegistrationFolder());
+        jupyterDataPaths.push(path.dirname(await this.jupyterPaths.getKernelSpecTempRegistrationFolder()));
         spawnOptions.env = {
             ...process.env,
-            JUPYTER_PATH: jupyterDataPaths.join(path.delimiter)
+            JUPYTER_PATH: jupyterDataPaths.length === 1 ? jupyterDataPaths[0] : jupyterDataPaths.join(path.delimiter)
         };
         return executionService.execModuleObservable('jupyter', ['notebook'].concat(notebookArgs), spawnOptions);
     }
--- a/src/client/datascience/jupyter/interpreter/jupyterInterpreterSubCommandExecutionService.ts
+++ b/src/client/datascience/jupyter/interpreter/jupyterInterpreterSubCommandExecutionService.ts
@@ -118,11 +118,13 @@ export class JupyterInterpreterSubCommandExecutionService
         // We don't want the process to die when the token is cancelled.
         const spawnOptions = { ...options };
         spawnOptions.token = undefined;
-        const jupyterDataPaths = (process.env['JUPYTER_PATH'] || '').split(path.delimiter);
+        const jupyterDataPaths = (process.env['JUPYTER_PATH'] || '')
+            .split(path.delimiter)
+            .filter((item) => item.trim().length);
         jupyterDataPaths.push(path.dirname(await this.jupyterPaths.getKernelSpecTempRegistrationFolder()));
         spawnOptions.env = {
             ...process.env,
-            JUPYTER_PATH: jupyterDataPaths.length === 1 ? jupyterDataPaths[0] : jupyterDataPaths.join(path.delimiter)
+            JUPYTER_PATH: jupyterDataPaths.join(path.delimiter)
         };
         return executionService.execModuleObservable('jupyter', ['notebook'].concat(notebookArgs), spawnOptions);
     }
--- a/src/client/datascience/jupyter/interpreter/jupyterInterpreterSubCommandExecutionService.ts
+++ b/src/client/datascience/jupyter/interpreter/jupyterInterpreterSubCommandExecutionService.ts
@@ -32,6 +32,7 @@ import {
 } from './jupyterInterpreterDependencyService';
 import { JupyterInterpreterService } from './jupyterInterpreterService';
 import { JupyterPaths } from '../../kernel-launcher/jupyterPaths';
+import { IEnvironmentActivationService } from '../../../interpreter/activation/types';
 
 /**
  * Responsible for execution of jupyter sub commands using a single/global interpreter set aside for launching jupyter server.
@@ -51,7 +52,8 @@ export class JupyterInterpreterSubCommandExecutionService
         @inject(IPythonExecutionFactory) private readonly pythonExecutionFactory: IPythonExecutionFactory,
         @inject(IOutputChannel) @named(JUPYTER_OUTPUT_CHANNEL) private readonly jupyterOutputChannel: IOutputChannel,
         @inject(IPathUtils) private readonly pathUtils: IPathUtils,
-        @inject(JupyterPaths) private readonly jupyterPaths: JupyterPaths
+        @inject(JupyterPaths) private readonly jupyterPaths: JupyterPaths,
+        @inject(IEnvironmentActivationService) private readonly activationHelper: IEnvironmentActivationService
     ) {}
 
     /**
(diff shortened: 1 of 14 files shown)


src/client/datascience/jupyter/interpreter/jupyterInterpreterSubCommandExecutionService.ts
```diff
@@ -116,6 +120,17 @@ export class JupyterInterpreterSubCommandExecutionService
...
         spawnOptions.token = undefined;
+        const envVars =
+            (await this.activationHelper.getActivatedEnvironmentVariables(undefined, interpreter, true)) || process.env;
+        const jupyterDataPaths = (process.env['JUPYTER_PATH'] || envVars['JUPYTER_PATH'] || '')
+            .split(path.delimiter)
+            .filter((item) => item.trim().length);
+        jupyterDataPaths.push(path.dirname(await this.jupyterPaths.getKernelSpecTempRegistrationFolder()));
+        spawnOptions.env = {
+            ...envVars,
+            JUPYTER_PATH: jupyterDataPaths.join(path.delimiter)
```
DonJayamanne, line 131: Magic sauce, this ensures Jupyter finds our kernelspecs (registered in a temporary directory)

src/client/datascience/jupyter/kernels/jupyterKernelService.ts
```diff
@@ -234,7 +233,7 @@ export class JupyterKernelService {
         cancelToken?: CancellationToken,
         forceWrite?: boolean
     ) {
-        const kernelSpecRootPath = await this.kernelFinder.getKernelSpecRootPath();
+        const kernelSpecRootPath = await this.jupyterPaths.getKernelSpecTempRegistrationFolder();
```
DonJayamanne, line 236: Change directory where we register kernelspecs

src/client/datascience/kernel-launcher/jupyterPaths.ts
```diff
@@ -38,6 +40,15 @@ export class JupyterPaths {
         );
     }
 
+    /**
+     * Contains the name of the directory where the Jupyter extension will temporary register Kernels when using non-raw.
+     * (this way we don't register kernels in global path).
+     */
+    public async getKernelSpecTempRegistrationFolder() {
+        const dir = path.join(this.context.extensionUri.fsPath, 'temp', 'jupyter', 'kernels');
```
DonJayamanne, line 48: All kernelspecs are always in extension folder.
This way when ever we update the extension, we always end up creating these.

I kind of like this, if users delete environments or the like, the kernels will not live around unnecessarily.

Note: If you upgrade to a new version of the extension, the exact same kernel (with the same name is created)
I.e. the algorithm for geneating kernel names hasn't changed.

src/test/datascience/kernel-launcher/localKernelFinder.unit.test.ts
```diff
@@ -32,992 +31,19 @@ import { IPythonExtensionChecker } from '../../../client/api/types';
...
 import { disposeAllDisposables } from '../../../client/common/helpers';
-import { KernelConnectionMetadata } from '../../../client/datascience/jupyter/kernels/types';
+import {
+    KernelConnectionMetadata,
+    LocalKernelConnectionMetadata
+} from '../../../client/datascience/jupyter/kernels/types';
 import { loadKernelSpec } from '../../../client/datascience/kernel-launcher/localKernelSpecFinderBase';
-import { traceInfoIfCI } from '../../../client/common/logger';
-
-[false, true].forEach((isWindows) => {
```
DonJayamanne, line 50: Removed all of the old tests
@rchiodo  You might recall this is something you asked for (& obviously I agreed with that suggestion).
Basically migrate old tests to the new style.

src/test/datascience/kernel-launcher/localKernelFinder.unit.test.ts
```diff
@@ -2059,6 +1281,133 @@ import { traceInfoIfCI } from '../../../client/common/logger';
...
+                        assert.strictEqual(
+                            kernel?.kernelSpec?.language,
+                            'python',
+                            'No python kernel found matching notebook metadata'
+                        );
+                        // Verify the kernel points to the active interpreter.
+                        assert.strictEqual(kernel?.kind, 'startUsingPythonInterpreter');
+                        assert.deepEqual(kernel?.interpreter, activePythonEnv);
+                    });
+                    test('Return active interpreter for interactive window (metadata only has language)', async function () {
```
DonJayamanne, line 1367: Added some of the missing tests (as part of deleting old tests)

src/test/datascience/kernel-launcher/localKernelFinder.unit.test.ts
rchiodo: I'm confused. Why wouldn't most of these tests still be valid?

src/client/datascience/kernel-launcher/localKernelSpecFinderBase.ts
```diff
@@ -126,19 +137,42 @@ export abstract class LocalKernelSpecFinderBase {
...
+                kernelSpec.specFile &&
+                path.dirname(path.dirname(kernelSpec.specFile)) === globalSpecRootPath;
+            if (kernelSpec && !shouldDeleteKernelSpec) {
+                return kernelSpec;
             }
-            // If we failed to get a kernelspec full path from our cache and loaded list
+            if (kernelSpec?.specFile && shouldDeleteKernelSpec) {
+                // If this kernelSpec was registered by us and is in the global kernels folder,
+                // then remove it.
+                this.deleteOldKernelSpec(kernelSpec.specFile).catch(noop);
```
amunger, line 155: This seems like unexpected behavior - the result of asking for a kernelSpec that exists, is for it to be deleted

src/test/datascience/kernel-launcher/localKernelFinder.unit.test.ts
DonJayamanne: They are, they are duplicates and written a different way.
You suggested that I just migrate them to the new way.

Basically I wrote tests in a different way, which I believe was easier, in the test:
* You can define all global kernle specs, all interpreters & kernlespecs that exist in each interpreter
* In the previous test this wasn't straightforward, it was slightly complicated.
* In that PR you suggested I migrate all the old tests to the new style (if the new tests covered the same scenarios)

src/client/datascience/kernel-launcher/localKernelFinder.ts
```diff
@@ -147,16 +152,14 @@ export class LocalKernelFinder implements ILocalKernelFinder {
...
+        // Do not update the cache if we got kernels from the cache.
+        if (!kernelsRetrievedFromCache) {
+            void this.cacheLocalKernelConnections(kernels);
+        }
         return kernels;
     }
 
-    // This should return a WRITABLE place that jupyter will look for a kernel as documented
-    // here: https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs
-    public async getKernelSpecRootPath(): Promise<string | undefined> {
```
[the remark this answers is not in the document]
DonJayamanne: Yes, we have a list of more paths and the root path was one of them.

src/client/datascience/kernel-launcher/localKernelSpecFinderBase.ts
DonJayamanne: Yes, its a side effect, we need to delete the old kernelspecs & this is a one time task.
I don't want to create a whole new method to iterate through kernels all over again (which is disc io & unnecessary complication).
Also I need a place to filter out these old kernlespecs (hence this is the spot).
In the future this code can be removed.

rchiodo (approved):

DonJayamanne:
@rchiodo Please re-approve, I think I hit the `review requested` button just as you approved it.

amunger (approved):

PR #9294, opened 2022-03-08
Check jupyter dependency when throwing jupyter err

### LINKED ISSUE #9271: Attempting to run a cell without Jupyter (non-zmq) will never fail and cells keeps running indefinitely
* I've disabled zmq
* ran a cell,
*  got prompted to select 
* Selected an interpreter from the prompt

Nothing happens, the cell seems to be running for ever (busy)
and there are no error messages displayed, and execution does not stop.

Errors are however logged in the output panel.

DonJayamanne:
Fixes #9271

### DIFF
--- a/src/client/datascience/errors/errorHandler.ts
+++ b/src/client/datascience/errors/errorHandler.ts
@@ -14,10 +14,9 @@ import { isPythonNotebook } from '../notebook/helpers/helpers';
 import {
     IDataScienceErrorHandler,
     IJupyterInterpreterDependencyManager,
-    IKernelDependencyService,
     KernelInterpreterDependencyResponse
 } from '../types';
-import { CancellationError as VscCancellationError, CancellationTokenSource, ConfigurationTarget } from 'vscode';
+import { CancellationError as VscCancellationError, ConfigurationTarget } from 'vscode';
 import { CancellationError } from '../../common/cancellation';
 import { KernelConnectionTimeoutError } from './kernelConnectionTimeoutError';
 import { KernelDiedError } from './kernelDiedError';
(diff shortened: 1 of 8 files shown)


src/client/datascience/errors/errorHandler.ts
```diff
@@ -104,23 +102,11 @@ export class DataScienceErrorHandler implements IDataScienceErrorHandler {
...
-                    resource,
-                    kernelConnection,
-                    new DisplayOptions(false),
-                    tokenSource.token,
-                    true
-                );
-            } finally {
-                tokenSource.dispose();
-            }
+        } else if (err instanceof JupyterInstallError && (purpose === 'start' || purpose === 'restart')) {
```
rchiodo, line 105: Maybe I'm missing something elsewhere, but this will not work for raw. JupyterInstallError is only thrown in the jupyter cases as far as I remember.

rchiodo (changes requested):
Does this work with raw?

src/client/datascience/errors/errorHandler.ts
DonJayamanne: I don't see the problem with that, 
If we're dealing with raw, then this error will not be throw, meaning its not a problem

rchiodo: No it means we won't install dependencies into a raw kernel. It will just fail.

rchiodo: That's why it was an 'or' before. It should install dependencies if it's an install error or the dependency analyzer says it needs the dependencies.

DonJayamanne: > No it means we won't install dependencies into a raw kernel. It will just fail.

Aaah, then we need to throw KernelDependency error, i'll check.
The only way we can throw `JupyterInstallError ` is in the case of non-raw, possible we're missing another condition. I'll check

rchiodo: The code used to check before running a raw kernel but I removed that and had it check here after it failed.

DonJayamanne: Got it, fixed,

rchiodo (approved):

PR #9297, opened 2022-03-08
Display errors in cells when we run into kernel issues

### LINKED ISSUE #9269: Restore error message when not installing IPYKernel or Jupyter Dependencies
The error message displayed when users cancel the prompt to install ipykernel has changed (the new error is as per screenshot).

Previously we'd display a message indicating the fact that the cell could not be run as ipykernle is not installed.
Yes, the user hit cancel here, but they don't know what ipykernle prompt is, and we chose to display the detailed error message in the output as well (asking with instructions to unblock themselves)

I think we should bring back the old error message.

image

DonJayamanne:
Fixes #9269


src/client/common/errors/types.ts
```diff
@@ -10,6 +11,17 @@ export abstract class BaseError extends Error {
     }
 }
 
+export abstract class BaseKernelError extends Error {
```
DonJayamanne, line 14: All kernel errors will have the `kernelConnectionMetadata`
Ideally we should include the `resource`, but don't need that just yet, hence not adding it now.

src/client/datascience/errors/errorHandler.ts
```diff
@@ -135,11 +164,16 @@ export class DataScienceErrorHandler implements IDataScienceErrorHandler {
             );
             if (failureInfo) {
                 void this.showMessageWithMoreInfo(failureInfo?.message, failureInfo?.moreInfoLink);
+            } else if (err instanceof BaseError) {
+                const message = getCombinedErrorMessage(getErrorMessageFromPythonTraceback(err.stdErr) || err.stdErr);
+                void this.showMessageWithMoreInfo(message);
+            } else {
+                void this.showMessageWithMoreInfo(err.message);
```
DonJayamanne, line 171: Always display errors regardless of the error.

src/client/datascience/errors/kernelPortNotUsedTimeoutError.ts
```diff
@@ -1,18 +1,19 @@
...
-    constructor(public readonly kernelConnection: KernelConnectionMetadata) {
+export class KernelPortNotUsedTimeoutError extends BaseKernelError {
+    constructor(kernelConnection: KernelConnectionMetadata) {
         super(
             'timeout',
             DataScience.rawKernelStartFailedDueToTimeout().format(
                 getDisplayNameOrNameOfKernelConnection(kernelConnection)
-            )
+            ),
+            kernelConnection
```
DonJayamanne, line 16: All kernel errors will have `kernelConnection` so we have more context.

src/client/datascience/jupyter/kernels/helpers.ts
```diff
@@ -1950,10 +1949,6 @@ export async function handleKernelError(
 
     switch (handleResult) {
         case KernelInterpreterDependencyResponse.cancel:
-            throw new CancellationError(
```
DonJayamanne, line 1953: rethrow original error

src/client/datascience/kernel-launcher/kernelProcess.ts
```diff
@@ -191,29 +190,7 @@ export class KernelProcess implements IKernelProcess {
                     return;
                 }
                 traceError('Kernel died', error, stderr);
-                if (error instanceof PythonKernelDiedError) {
```
DonJayamanne, line 194: This error is never thrown anywhere, it was only thrown in the case of kernels being launched using daemon (which isn't used anymore)

rchiodo (approved):

src/client/datascience/jupyter/kernels/helpers.ts
rchiodo: I imagine this fixed the main problem.

src/client/datascience/notebook/vscodeNotebookController.ts
```diff
@@ -385,8 +386,11 @@ export class VSCodeNotebookController implements Disposable {
             }
             return await kernel.executeCell(cell);
         } catch (ex) {
+            const errorHandler = this.serviceContainer.get<IDataScienceErrorHandler>(IDataScienceErrorHandler);
+
             // If there was a failure connecting or executing the kernel, stick it in this cell
-            return displayErrorsInCell(cell, execution, ex.toString());
+            displayErrorsInCell(cell, execution, await errorHandler.getErrorMessageForDisplayInCell(ex));
+            return NotebookCellExecutionState.Idle;
```
rchiodo, line 393: Wouldn't this be an error? Idle seems wrong?

DonJayamanne: This is the execution state, not about the state after the execution , i.e. not execution result.
Valid values from VS Code are running, not running or Pending.

src/client/datascience/jupyter/kernels/helpers.ts
DonJayamanne: Yes

PR #9311, opened 2022-03-09
removed Notebook creation from extension since it is moving to built-in

### LINKED ISSUE #9250: Move "New Jupyter File" (ipynb) command and implementation to VS Code core
Because VS Code core is able to read and write ipynb files without the help of the jupyter extension, it only makes sense to move the code/command that creates a new ipynb file to core as well.

Whether the Jupyter extension is recommended on creating the file, executing code, or picking a kernel will need to be resolved as well.

amunger:
Fixes #9250

This should be merged in coordination with https://github.com/microsoft/vscode/pull/144789 to limit the amount of time that the command is unavailable/duplicated.

The create Notebook flow is still used pretty widely in the test framework, so this change does not completely remove it.

This will remove the functionality of registered 3rd party notebooks appearing in the "Create Jupyter Notebook" quick pick menu since that command will simply create a Python Notebook.


DonJayamanne:
>  "Create Jupyter Notebook" quick pick menu since that command will simply create a Python Notebook.

Thats not true, if you open a julia notebook and use that, then the new notebook created using that command defaults to Julia.
Similarly if you are an R user, the new notebook defaults to R & not python.

amunger:
> the new notebook defaults to R & not python

That happens through controller selection for the created notebook, not the notebook creation itself. This change should not affect that behavior, although I'm having a hard time testing that in OSS.

DonJayamanne:
> Thats not true, if you open a julia notebook and use that, then the new notebo

Yeah looking at the old code, it never worked, looks like the code was lost at some point.
Surprised no one is complaining, I guess it works because untitled notebooks automatically have their kernel change to what they used last, and that results in cell languages getting updated.

rchiodo (approved):

IanMatthewHuff (approved):

DonJayamanne (approved):

DonJayamanne (changes requested):
We must change the engine in package.json to ensure this change goes out ONLY with the corresponding 1.65 version of VS Code.

DonJayamanne (approved):

PR #9339, opened 2022-03-12
added GH issues and wiki links

minsa110:


README.md
```diff
@@ -4,6 +4,11 @@ A [Visual Studio Code](https://code.visualstudio.com/) [extension](https://marke
 
 The Jupyter Extension includes the Jupyter Keymaps and the Jupyter Notebook Renderers extensions by default. The Jupyter Keymaps extension provides Jupyter-consistent keymaps and the Jupyter Notebook Renderers extension provides renderers for MIME types such as latex, plotly, vega, and the like. Both of these extensions can be disabled or uninstalled.
 
+| Link | Description |
+|------|-------------|
+| [File an issue](https://github.com/microsoft/vscode-jupyter/issues/new/choose) | For reporting problems and suggesting enhancements |
```
amunger, line 9: `Help `-> `report issue` from VSCode is nice since it auto-populates version info. Would it work to squeeze that info in here? If that feels too noisy than this change is also fine.

amunger (approved):

README.md
```diff
@@ -4,6 +4,11 @@ A [Visual Studio Code](https://code.visualstudio.com/) [extension](https://marke
...
+| [Go to wiki](https://github.com/microsoft/vscode-jupyter/wiki) | For FAQs and learning how to use/debug the extension |
```
DonJayamanne, line 10: Not sure this is required. For debugging the extension we have a contributing guide.
For using the extension we have VS Code docs.
I.e. I don't think we need to explicitly call this page out, if we do we should first call out the VS Code docs and contributing guide (the latter is already linked in this readme).
Also we might want to provide a link to the discussions channel .

DonJayamanne (changes requested):
Not in favor of promoting WIKI, I think we should add links to VS Code docs for users to learn how to use the extension (we don't have any documentaiton on our WIKI with instructions on how to use the extension)

README.md
DonJayamanne: I think the `help-report issue` isn't that great as it misses a lot of the information that we need , python information , whether its debugging, interactive window or notebook or the like.

PR #9353, opened 2022-03-14
Dispose the temporary kernel connection file

### LINKED ISSUE #9298: Don't start kernel daemon for non-python kernels.
Request is to hide the "Jupyter: Debug Cell" command for .NET Nb. 

Just had a user try to use it while using C# in a notebook and it completely broke their setup :( 

We want to make sure users don't think they can use this debugging feature to the point where they are completely broken.

DonJayamanne:
Fixes #9298

### DIFF
--- a/src/kernels/raw/launcher/kernelProcess.ts
+++ b/src/kernels/raw/launcher/kernelProcess.ts
@@ -106,7 +106,8 @@ export class KernelProcess implements IKernelProcess {
         } else if (
             this._kernelConnectionMetadata.kernelSpec.interrupt_mode !== 'message' &&
             this._process &&
-            this._interruptSignalHandle
+            this._interruptSignalHandle &&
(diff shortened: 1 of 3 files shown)


src/kernels/raw/launcher/kernelProcess.ts
```diff
@@ -331,6 +332,8 @@ export class KernelProcess implements IKernelProcess {
             // a different process tries to open it.
             const tempFile = await this.fileSystem.createTemporaryLocalFile('.json');
             this.connectionFile = tempFile.filePath;
+            // Ensure we dispose this, and don't maintain a handle on this file.
+            await tempFile.dispose(); // Do not remove this line.
```
DonJayamanne, line 336: I believe this is the root cause of #9298
I removed this for some reason, and have brought this back now.

DonJayamanne: Tests have been added

PR #9354, opened 2022-03-15
Fix problem with 'install ipykernel' appearing twice if you hit Cancel on IW

### LINKED ISSUE #9267: Installing ipykernel will come up twice if kernel is started quick enough
Refactor of the error handling caused this. The auto start causes the problem.

rchiodo:
Fixes #9267

Root cause was opening an IW causes a selection to occur in the 'Select Kernel' box. This selection will call 'startKernel'. However opening the window also calls 'startKernel'. These two situations are chained together.

However if the first to enter is canceled, the 'kernelConnectionId' wasn't saved and the second would attempt to select a kernel a second time.

Also fixed a build problem I noticed with fontAwesomeLoader.js

### DIFF
--- a/build/webpack/webpack.datascience-ui.config.builder.js
+++ b/build/webpack/webpack.datascience-ui.config.builder.js
@@ -134,9 +134,8 @@ function buildConfiguration(bundle) {
                     from: path.join(
                         constants.ExtensionRootDir,
                         'src',
-                        'client',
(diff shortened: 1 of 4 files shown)


DonJayamanne (approved):
@rchiodo  Ok i'm sorry, I was continously testing with a Notebook and not IW.

PR #9358, opened 2022-03-15
Fixes to flaky Installer Prompt tests

### LINKED ISSUE #9186: Flaky test `Ensure ipykernel install prompt is NOT displayed when auto start is enabled`
This is a new test

DonJayamanne:
Fixes #9186

### DIFF
--- a/src/kernels/helpers.ts
+++ b/src/kernels/helpers.ts
@@ -31,7 +31,7 @@ import { traceError, traceInfo, traceInfoIfCI, traceVerbose, traceWarning } from
 import { getDisplayPath } from '../client/common/platform/fs-paths';
 import { IPythonExecutionFactory } from '../client/common/process/types';
 import { IPathUtils, IConfigurationService, Resource, IMemento, GLOBAL_MEMENTO } from '../client/common/types';
-import { createDeferred } from '../client/common/utils/async';
+import { createDeferred, createDeferredFromPromise, Deferred } from '../client/common/utils/async';
 import { DataScience } from '../client/common/utils/localize';
 import { getResourceType } from '../client/datascience/common';
 import { Settings } from '../client/datascience/constants';
@@ -44,7 +44,8 @@ import {
     IDataScienceErrorHandler,
     IRawNotebookProvider,
     KernelInterpreterDependencyResponse,
-    IJupyterKernelSpec
+    IJupyterKernelSpec,
+    IDisplayOptions
 } from '../client/datascience/types';
 import { IServiceContainer } from '../client/ioc/types';
 import {
@@ -71,6 +72,8 @@ import { isPythonNotebook } from '../notebooks/helpers';
 import { INotebookControllerManager } from '../notebooks/types';
 import { PreferredRemoteKernelIdProvider } from './raw/finder/preferredRemoteKernelIdProvider';
 import { findNotebookEditor, selectKernel } from '../notebooks/controllers/kernelSelector';
+import { DisplayOptions } from '../client/datascience/displayOptions';
+import { KernelDeadError } from '../extension/errors/kernelDeadError';
 
 // Helper functions for dealing with kernels and kernelspecs
 
@@ -1975,10 +1978,10 @@ export async function handleKernelError(
     return resultController;
 }
 
-function convertContextToFunction(context: 'start' | 'interrupt' | 'restart') {
+function convertContextToFunction(context: 'start' | 'interrupt' | 'restart', options?: IDisplayOptions) {
     switch (context) {
         case 'start':
-            return (k: IKernel) => k.start();
+            return (k: IKernel) => k.start(options);
 
         case 'interrupt':
             return (k: IKernel) => k.interrupt();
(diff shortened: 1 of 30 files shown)


DonJayamanne:
Still flaky,

src/kernels/helpers.ts
```diff
@@ -1975,10 +1977,10 @@ export async function handleKernelError(
     return resultController;
 }
 
-function convertContextToFunction(context: 'start' | 'interrupt' | 'restart') {
+function convertContextToFunction(context: 'start' | 'interrupt' | 'restart', options?: IDisplayOptions) {
     switch (context) {
         case 'start':
-            return (k: IKernel) => k.start();
+            return (k: IKernel) => k.start(options);
```
DonJayamanne, line 1983: Keep track of whether UI is to be displayed or not.
When we use auto start we need to go through his same code.

src/kernels/installer/moduleInstaller.ts
```diff
@@ -55,25 +55,33 @@ export abstract class ModuleInstaller implements IModuleInstaller {
         const environmentService = this.serviceContainer.get<IEnvironmentVariablesService>(
             IEnvironmentVariablesService
         );
+        if (cancelTokenSource.token.isCancellationRequested) {
```
DonJayamanne, line 58: You'll see these in a few places, while running the tests found that these were causing issues.
Basically we had stopped cell execution (either closing a document or interrupt), & these code paths would continue running.

src/kernels/installer/moduleInstaller.ts
```diff
@@ -55,25 +55,33 @@ export abstract class ModuleInstaller implements IModuleInstaller {
...
             }>,
             token: CancellationToken
         ) => {
+            // Args can be for a specific exe or for the interpreter. Both need to
+            // use an activated environment though
+            const deferred = createDeferred();
             // When the progress is canceled notify caller
             token.onCancellationRequested(() => {
                 cancelTokenSource.cancel();
+                deferred.resolve();
```
DonJayamanne, line 74: Another fix, if token was cancelled, we should stop the installation

src/kernels/installer/productInstaller.ts
```diff
@@ -117,6 +117,9 @@ abstract class BaseInstaller {
         if (!installer) {
             return InstallerResponse.Ignore;
         }
+        if (cancelTokenSource.token.isCancellationRequested) {
```
DonJayamanne, line 120: You'll see these in a few places, while running the tests found that these were causing issues.
Basically we had stopped cell execution (either closing a document or interrupt), & these code paths would continue running.

src/kernels/jupyter/launcher/jupyterNotebookProvider.ts
```diff
@@ -51,7 +52,7 @@ export class JupyterNotebookProvider implements IJupyterNotebookProvider {
             token: options.token,
             localJupyter: isLocalConnection(options.kernelConnection)
         });
-
+        Cancellation.throwIfCanceled(options.token);
```
DonJayamanne, line 55: Created helper function to check if session is invalid or if token was cancelled.

src/kernels/kernel.ts
```diff
@@ -70,8 +72,14 @@ import {
...
 } from './types';
 import { KernelExecution } from '../notebooks/execution/kernelExecution';
+import { traceCellMessage } from '../notebooks/helpers';
+import { Cancellation } from '../client/common/cancellation';
 
 export class Kernel implements IKernel {
+    /**
+     * Used for debugging purposes, ability to uniquely identify kernels.
+     */
+    public readonly id: string;
```
DonJayamanne, line 82: This was very useful, found we were creating a few kernels, initially I thought they were the same after adding this, found that we were actually creating new instances

src/kernels/kernel.ts
```diff
@@ -248,20 +263,23 @@ export class Kernel implements IKernel {
         const disposeImpl = async () => {
             traceInfo(`Dispose kernel ${(this.resourceUri || this.notebookDocument.uri).toString()}`);
             this.restarting = undefined;
+            const promises: Promise<void>[] = [];
+            promises.push(this.kernelExecution.cancel());
```
DonJayamanne, line 267: Found that when shutting down a kernel, we didn't cancel the pending executions, all we did was dispose that object.
This is kinda debt

src/kernels/kernel.ts
```diff
@@ -340,11 +358,24 @@ export class Kernel implements IKernel {
...
+        }
         this._startedAtLeastOnce = true;
+        traceInfoIfCI(
+            `Start Notebook (options.disableUI=${options.disableUI}) for ${getDisplayPath(this.notebookDocument.uri)}.`
+        );
+        console.error(new Error('(startNotebook). Where am I called from'));
         if (!options.disableUI) {
             this.startupUI.disableUI = false;
         }
+        options.onDidChangeDisableUI(() => {
```
DonJayamanne, line 374: This was missing, if we have auto start, then we don't display, but when running cells, the errors are not displayed as we're using the stale state of hiding the UI.

src/kernels/kernelDependencyService.ts
```diff
@@ -83,14 +89,25 @@ export class KernelDependencyService implements IKernelDependencyService {
         }
 
         // Cache the install run
-        const cancelTokenSource = new CancellationTokenSource();
         let promise = this.installPromises.get(kernelConnection.interpreter.path);
+        let cancelTokenSource: CancellationTokenSource | undefined;
         if (!promise) {
+            const cancelTokenSource = new CancellationTokenSource();
+            const disposable = token.onCancellationRequested(() => {
```
DonJayamanne, line 96: This was a bug, basically we stop execution, but this part continues, as we created a whole new cancellation token and ignored the one that was passed into this.

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -434,13 +462,59 @@ export class VSCodeNotebookController implements Disposable {
...
     }
+    private static readonly connections = new WeakMap<
+        NotebookDocument,
+        { kernel: Deferred<IKernel>; options: IDisplayOptions }
+    >();
+    private async connectToKernel(
+        doc: NotebookDocument,
+        options: IDisplayOptions = new DisplayOptions(false)
+    ): Promise<IKernel> {
+        let currentPromise = VSCodeNotebookController.connections.get(doc);
```
DonJayamanne, line 473: **This is one of the bugs that was fixed, its an easy to repro issue:**
* Open a notebook, select a kernel A without ipykernel
* Run all cells
* In the prompt displayed select the option to change the kernel to B
* We still get error message indicating ipykernel isnt found in A
* Run all cells again, and we'll still get errors about ipykernel not in A

The problem was we had `connectToKernel` function called from different places and they all had their own while loops, and thats what caused issues (causing multiple kernels to get created, one gets created and disposes the other, and even after we change the controller, the previous loop will treat the older controller as active and weird things happend).

Basically we were running `connectToKernel` function and had multiple while loops (instead of one) = now we cache it per doc.

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -584,7 +658,7 @@ export class VSCodeNotebookController implements Disposable {
             isLocalConnection(this.kernelConnection)
         ) {
             // Startup could fail due to missing dependencies or the like.
-            void newKernel.start({ disableUI: true });
+            this.connectToKernel(document, new DisplayOptions(true)).catch(noop);
```
DonJayamanne, line 661: THis was another fix, here we start `newKernel.start` our selves, but `connectToKernel` function will do something else and those two would step on their toes.

src/notebooks/execution/cellExecutionQueue.ts
```diff
@@ -127,8 +128,14 @@ export class CellExecutionQueue implements Disposable {
 
             let executionResult: NotebookCellRunState | undefined;
             try {
-                await cellToExecute.start(kernelConnection);
-                executionResult = await cellToExecute.result;
+                if (cellToExecute.cell.notebook.isClosed) {
```
DonJayamanne, line 131: Found while debugging the tests.
Basically we stopped some tests early (which was not right), while some cells were still getting executed.
This ensures we don't run cells after a document has been closed.
The logs showed cells with index -1 were running.

src/kernels/helpers.ts
```diff
@@ -1988,17 +1991,117 @@ function convertContextToFunction(context: 'start' | 'interrupt' | 'restart') {
...
+    // But do that only if we require a UI, else we can just use the current one.
+    if (
+        !options.disableUI &&
+        currentPromise?.kernel.resolved &&
+        (currentPromise?.kernel.value?.kernel?.disposed || currentPromise?.kernel.value?.kernel?.disposing)
+    ) {
+        connections.delete(notebook);
+        currentPromise = undefined;
+    }
+    const verifyKernelState = async (
```
rchiodo, line 2028: I feel like this would be better as function external to this method. It makes it a bit cluttered to have this as an internal function. It looks like it doesn't modify the internal state of the function, so I believe it should be moveable.

rchiodo (approved):

PR #9359, opened 2022-03-15
Handle scenarios when we do not get prompted to install ipykernel (fix test failure)

DonJayamanne:
In this test run https://github.com/microsoft/vscode-jupyter/runs/5547740604?check_suite_focus=true
we can see that we didn't get a prompt for ipykernel not installed error message.
Instead we just dump an error `Kernel died`

### DIFF
--- a/src/kernels/raw/launcher/kernelProcess.ts
+++ b/src/kernels/raw/launcher/kernelProcess.ts
@@ -23,7 +23,7 @@ import {
     getErrorMessageFromPythonTraceback
 } from '../../../client/../extension/errors/errorUtils';
 import { BaseError } from '../../../client/../extension/errors/types';
-import { traceInfo, traceError, traceVerbose, traceWarning } from '../../../client/common/logger';
+import { traceInfo, traceError, traceVerbose, traceWarning, traceInfoIfCI } from '../../../client/common/logger';
 import { IFileSystem } from '../../../client/common/platform/types';
 import {
     IProcessServiceFactory,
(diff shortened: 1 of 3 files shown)


src/kernels/raw/launcher/kernelProcess.ts
```diff
@@ -240,6 +241,7 @@ export class KernelProcess implements IKernelProcess {
                 const errorMessage =
                     getErrorMessageFromPythonTraceback(stderrProc || stderr) ||
                     (stderrProc || stderr).substring(0, 100);
+                traceInfoIfCI(`KernelDiedError raised`, errorMessage, stderrProc + '\n' + stderr + '\n');
```
DonJayamanne, line 244: Added some logging in case this test fails again

src/test/datascience/jupyter/kernels/installationPrompts.vscode.test.ts
```diff
@@ -209,7 +209,7 @@ suite('DataScience Install IPyKernel (slow) (install)', function () {
                 assertVSCCellIsNotRunning(cell) &&
                 verifyInstallIPyKernelInstructionsInOutput(cell),
             defaultNotebookTestTimeout,
-            'No errors in cell'
+            'No errors in cell (first time)'
         );
```
DonJayamanne, line 213: Changes to error messages to determine at which step the test failed.

rchiodo (approved):

IanMatthewHuff (approved):

PR #9378, opened 2022-03-16
Log the workspace folders in output

DonJayamanne:
Useful, instead of having to ask users.


amunger:
Is the actual location useful, or just whether or not they have a workspace open and what type? I'm just thinking about people accidentally posting logs that contain their username in a path when they didn't want to share that information (unless we already log file/directory paths at some point)

rchiodo (approved):

PR #9391, opened 2022-03-16
Remove forcing npm package versions

DonJayamanne:
We forced (pinned) npm packages to specific versions due to some vulnerabilities, they have since been fixed upstream.


rchiodo:
If they were fixed upstream, don't we have to update the packages that use those?

DonJayamanne:
> don't we have to update the packages that use those?

Nope, those upstream packages didn't change, just their packages their downloaded.
The package-lock.json was re-generated and there was no change, basically the upstream packages also ended up using the same new packages versions that we were pinning to.

rchiodo:
> > don't we have to update the packages that use those?
> 
> Nope, those upstream packages didn't change, just their packages their downloaded. The package-lock.json was re-generated and there was no change, basically the upstream packages also ended up using the same new packages versions that we were pinning to.

Woah, weird.

Does npm audit not show the status correctly? In the node upgrade I still get security warnings.

DonJayamanne:
Yeah weird, I think i was wrong.
Closing this invalid PR

PR #9394, opened 2022-03-16
Dev/aamunger/deprecate create notebook

amunger:
most of the change is to revert the commit that removed the command, but still keep it off of the command palate and add a deprecation info message when the command is invoked through a keyboard shortcut.


src/interactive-window/commands/commandRegistry.ts
```diff
@@ -479,6 +482,15 @@ export class CommandRegistry implements IDisposable {
         }
     }
 
+    private async createNewNotebook(): Promise<void> {
+        this.appShell
+            .showInformationMessage(
+                'This command has been deprecated and will eventually be removed, please use "Create: New Jupyter Notebook" instead'
```
amunger, line 488: trying to see if there's a smart link that I can use to guide them right to the keyboard shortcut

src/notebooks/notebookCreator.ts
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { inject, injectable } from 'inversify';
```
rchiodo, line 4: I believe this is a merge conflict problem. I deleted this file.

src/test/datascience/notebook/creation/notebookCreation.vscode.test.ts
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
rchiodo, line 1: This file too. This should be gone.

src/notebooks/notebookCreator.ts
amunger: This is what the command runs, so this file is necessary

amunger: It was removed in the previous PR https://github.com/microsoft/vscode-jupyter/pull/9311/files

rchiodo: Oh you removed it. HA! My bad.

rchiodo (approved):

package.nls.json
```diff
@@ -122,6 +122,8 @@
     "jupyter.command.jupyter.collapseallcells.shorttitle": "Collapse",
     "jupyter.command.jupyter.addcellbelow.title": "Add Empty Cell to File",
     "jupyter.command.jupyter.scrolltocell.title": "Scroll Cell Into View",
+    "jupyter.command.jupyter.createnewnotebook.title": "Create New Jupyter Notebook",
```
DonJayamanne, line 125: Since this command doesn't show anywhere, I do'nt think we need to bother with the localizable texts.

src/interactive-window/commands/commandRegistry.ts
```diff
@@ -479,6 +482,15 @@ export class CommandRegistry implements IDisposable {
...
+            )
+            .then(noop, noop);
+        await this.nativeNotebookCreator.createNewNotebook();
```
DonJayamanne, line 491: Can't we just call the VS Code command here directly `commands.executeCommand(<new VS Code Command Id>)`
this way we do'nt need nativeNotebookCreator

src/notebooks/notebookCreator.ts
DonJayamanne: Not, sure why we're bringing this back? 
I thought the whole point of the previous PR was to remove this in favor of the VS code command
I

src/interactive-window/commands/commandRegistry.ts
DonJayamanne: I thought based on the standup discussion today, we'd display the message and then run the new VS Code command to create the jupyter notebook

amunger: yeah that's a good idea, then I can drop the rest of that stuff again

src/notebooks/serviceRegistry.ts
```diff
@@ -55,6 +56,7 @@ export function registerTypes(serviceManager: IServiceManager) {
     );
     serviceManager.addSingleton<INotebookLanguageClientProvider>(INotebookLanguageClientProvider, IntellisenseProvider);
     serviceManager.addBinding(INotebookLanguageClientProvider, IExtensionSingleActivationService);
+    serviceManager.addSingleton<CreationOptionService>(CreationOptionService, CreationOptionService);
```
DonJayamanne, line 59: Do we still need this? I think we can remove this class completely

DonJayamanne (approved):

PR #9408, opened 2022-03-17
added one test telemetry point

amunger:


rchiodo (approved):

src/telemetry/index.ts
```diff
@@ -1530,4 +1530,9 @@ export interface IEventNamePropertyMapping {
         ename: string;
         evalue: string;
     };
+    [Telemetry.RunTest]: {
+        testName: string;
+        testResult: string;
+        isCIServer: boolean;
```
DonJayamanne, line 1536: Why do we need this? 
In production we'd never send such telemetry as users will never have the tests and that code path will never get executed.

DonJayamanne (changes requested):
I think we can remove isCIServer, as this will not be necessary for production.
If anything we can modiy the test hook to disable sending the test telemetry if we're not on CI server

src/telemetry/index.ts
amunger: when we execute tests locally it won't be on a CI server, and I definitely want to be able to filter those out (maybe even not send telemetry for local runs) since they are much more likely to fail when used to try out a code change or new feature

src/test/testHooks.ts
```diff
@@ -0,0 +1,17 @@
+import { Context } from 'mocha';
+import { Telemetry } from '../platform/common/constants';
+import { sendTelemetryEvent } from '../telemetry';
+import { IS_CI_SERVER } from './ciConstants.node';
+
+export const rootHooks = {
+    afterEach(this: Context) {
+        let result = this.currentTest?.isFailed() ? 'failed' : this.currentTest?.isPassed() ? 'passed' : 'skipped';
```
DonJayamanne, line 8: Can we add a condition `if (!IS_CI_SERVER) { return ; }` that way we don't send telemetry to server and we don't need to filter it. Else when ever we run any test locally we always end up sending that telemetry locally.

src/telemetry/index.ts
DonJayamanne: Aaah got it, in that case can we add a condition to the hooks to disable sending telemery for locally run tests.
Else currently any test run locally will always send the telemetry

DonJayamanne (approved):

PR #9439, opened 2022-03-21
Create platform directory structure

### LINKED ISSUE #8981: Refactor extension initialization code into an 'Extension' subfolder off of src
Migrate code that could be seen as the entry point to starting and configuring the extension with core.

- Extension activation
- DI service initialization

rchiodo:
Fixes #8981 

Did a bunch of stuff:
- Renamed 'client' directory to 'platform' and moved some stuff out of the 'client/datascience' directory into their respective root locations.
- Move telemetry code into its own folder. This was in multiple spots. Consolidated it to a root folder
- Create a serviceRegistry.ts hierarchy. All root folders (will) have a serviceRegistry.ts. This is where injectable classes are added to the main DI container. 
- Create a single extension.ts at the 'src' folder level. It simply creates the container and registers everything.


.vscode/launch.json
```diff
@@ -140,7 +140,7 @@
             ],
             "env": {
                 "VSC_JUPYTER_FORCE_LOGGING": "1",
-                "VSC_JUPYTER_CI_TEST_GREP": "", // Leave as `VSCode Notebook` to run only Notebook tests.
```
rchiodo, line 143: Whoops. Didn't mean to change this

src/datascience-ui/common/index.ts
```diff
@@ -108,7 +109,7 @@ export function formatStreamText(str: string): string {
 export function appendLineFeed(arr: string[], modifier?: (s: string) => string) {
     return arr.map((s: string, i: number) => {
         const out = modifier ? modifier(s) : s;
-        return i === arr.length - 1 ? `${out}` : `${out}\n`;
+        return i === arr.length - 1 ? `${out}` : `${out}${os.EOL}`;
```
DonJayamanne, line 112: Not sure this is correct. I remember Jupyter used just `\n` (line feeds).
E.g. if we have some multi-line code in a non-python kernel, then it could crash (I know R kernel doesn't support CRLF & we had to ensure we didn't use that else the execution would just crash)

src/platform/ioc/serviceManager.ts
```diff
@@ -8,7 +8,7 @@ type identifier<T> = string | symbol | Newable<T> | Abstract<T>;
 
 @injectable()
 export class ServiceManager implements IServiceManager {
-    constructor(private container: Container) {}
+    constructor(public readonly container: Container) {}
```
DonJayamanne, line 11: Can we create a property with a simple getter that has a decorator to ensure this gets accessed only during tests.
I.e. use the `@testOnlyMethod` decorator that Ian added.

DonJayamanne: Not required, i think its a good practice

DonJayamanne (approved):
Looks good,
didn't review the tests, eyes & figures started hurting,

src/datascience-ui/common/index.ts
rchiodo: I'll make it only used for export.

src/kernels/serviceRegistry.ts
```diff
@@ -0,0 +1,278 @@
...
+export function registerTypes(serviceManager: IServiceManager, _isDevMode: boolean) {
+    serviceManager.addSingleton<IRawNotebookSupportedService>(
+        IRawNotebookSupportedService,
+        RawNotebookSupportedService
+    );
+    const isVSCInsiders = serviceManager.get<IApplicationEnvironment>(IApplicationEnvironment).channel === 'insiders';
+    const packageJson: { engines: { vscode: string } } | undefined = vscode.extensions.getExtension(JVSC_EXTENSION_ID)
+        ?.packageJSON;
+    const isInsiderVersion = packageJson?.engines?.vscode?.toLowerCase()?.endsWith('insider');
+    setSharedProperty('isInsiderExtension', isVSCInsiders && isInsiderVersion ? 'true' : 'false');
```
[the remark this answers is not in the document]
rchiodo: I didn't change this code. It just moved.

src/platform/ioc/serviceManager.ts
rchiodo: Good idea. Didn't realize we had a testOnlyDecorator.

IanMatthewHuff: Yeah, I added it for the WebViewHost I think if you want to see it used. Do wonder about the best was to sometimes use these decorators, like I'm not sure how you would be expected to know that we added on? Util stuff just kinda hide out of the way sometimes.

src/kernels/raw/session/rawJupyterSession.ts
```diff
@@ -68,7 +64,7 @@ export class RawJupyterSession extends BaseJupyterSession {
     }
 
     public async waitForIdle(timeout: number): Promise<void> {
-        // Wait until status says idle.
+        // Wait until status says idle.dd
```
IanMatthewHuff, line 67: ? just a comment, but maybe a typo here?

rchiodo: Whoops, will fix

src/platform/ioc/serviceManager.ts
rchiodo: FYI, the decorator doesn't work on a getter. Throws an error. I had to change this to a function ```getContainer```

src/datascience-ui/common/index.ts
rchiodo: Export tests were failing and the output made them look the same.

rchiodo: ipynb already remove any \r\n in them. However export generates python files which assumes the same EOL as the OS. The tests actually verify this is the case.

IanMatthewHuff (approved):

IanMatthewHuff:
@rchiodo Nice work! I 100% say would not change this now (or ever maybe), but just something that I like is if the ordering lines up for the functions. 

For example there is registerPlatformTypes and processRegisterTypes. I would slightly prefer having them both follow the same ordering like registerPlatformTypes and registerProcessTypes. Or switch to platformRegisterTypes and processRegisterTypes. Just makes them easier to find searching if they follow the same format.

src/datascience-ui/common/index.ts
DonJayamanne: Oh this is export ipynb to py. i thought it was the other way around.

PR #9440, opened 2022-03-22
Get interpreter for workspace and all

DonJayamanne:
Found while testing that we don't get all of the interpreters.
I can see the environment showing up in Pythons Select Interpreter, but not in our kernel picker.
The difference was with the way get the list of interpreters, we provide a Uri.
I added an additional call to get interpreters without a Uri & that magically returned all interpreters.

### DIFF
--- a/src/kernels/helpers.ts
+++ b/src/kernels/helpers.ts
@@ -178,6 +178,7 @@ export function getDisplayNameOrNameOfKernelConnection(kernelConnection: KernelC
                 kernelConnection.interpreter?.envType &&
                 kernelConnection.interpreter.envType !== EnvironmentType.Global
             ) {
+                const envName = getPythonEnvironmentName(kernelConnection.interpreter);
                 if (kernelConnection.kernelSpec.language === PYTHON_LANGUAGE) {
                     const pythonVersion = `Python ${
                         getTelemetrySafeVersion(kernelConnection.interpreter.version?.raw || '') || ''
@@ -187,9 +188,7 @@ export function getDisplayNameOrNameOfKernelConnection(kernelConnection: KernelC
                         : oldDisplayName;
                 } else {
                     // Non-Python kernelspec that launches via python interpreter
-                    return kernelConnection.interpreter.envName
-                        ? `${oldDisplayName} (${kernelConnection.interpreter.envName})`
-                        : oldDisplayName;
+                    return envName ? `${oldDisplayName} (${envName})` : oldDisplayName;
                 }
             } else {
                 return oldDisplayName;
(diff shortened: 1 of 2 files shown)


src/client/api/pythonApi.ts
```diff
@@ -338,7 +338,11 @@ export class InterpreterService implements IInterpreterService {
         // we want all interpreters regardless of workspace folder so call this multiple times
         const folders = this.workspace.workspaceFolders;
         const all = folders
-            ? await Promise.all(folders.map((f) => this.apiProvider.getApi().then((api) => api.getInterpreters(f.uri))))
+            ? await Promise.all(
+                  [...folders, undefined].map((f) =>
```
DonJayamanne, line 342: Basically add another item into the array to fetch `getInterpreters` with an argument of `undefined`.

src/kernels/helpers.ts
```diff
@@ -212,13 +211,21 @@ export function getDisplayNameOrNameOfKernelConnection(kernelConnection: KernelC
...
-                    : pythonDisplayName;
+                const envName = getPythonEnvironmentName(kernelConnection.interpreter);
+                return envName ? `${envName} (${pythonDisplayName})` : pythonDisplayName;
             }
     }
     return oldDisplayName;
 }
+function getPythonEnvironmentName(pythonEnv: PythonEnvironment) {
+    // Sometimes Python extension doesn't detect conda environments correctly (e.g. conda env create without a name).
+    // In such cases the envName is empty, but it has a path.
```
DonJayamanne, line 222: Basically I was ending up with `Python 3.6` as the display name for a lot of my test environments.
This ensures we have a better display name (e.g. `test1 (Python 3.6)` which is the standard we've been using when we have an env name). 
In such cases the envname is not available as you activate using the fully qualified path, however you can also activate using the folder name, sometimes

src/client/api/pythonApi.ts
IanMatthewHuff: Seems good functionally, just a minor question. Is this a big slow down issue? Thought that getInterpreters was pretty slow? Though I would assume that slowness is just on Python extension populating the initial list, so calling with a folder and undefined might not be a big diff.

DonJayamanne: > Is this a big slow down issue?

No it isn't, python extension uses a cache, and they don't search all over again, basically this just filters stuff (if using `undefined` its a superset of workspace, hence doesn't search worksapce again.

DonJayamanne: & even if it were a slow down, its a bug in Python extension, and we shouldn't be focused on that
RIght now this is a bug, I don't see the kernels, but python extension already has them (again, means its cached and Python shouldn't search all over again, but if they do, its their bug - unlikely to happen)

IanMatthewHuff (approved):

rchiodo (approved):

PR #9455, opened 2022-03-22
Display meaningful error messages in IW

### LINKED ISSUE #8384: Remove `Connecting to xxxx` from IW when changing kernels
* Run a cell
* When you get a promtp to select another kernel, select one with ipykernel
* Cell runs

However you still end up with the message `Connecting to xyz`, which is confusing.
Why are we still trying to connect to `xyz` when I've changed the kernel?

### LINKED ISSUE #9444: Creating IW with a bad kernel and hitting cancel on the the ipykernel install allows for execution still
Testing #9435

If you cancel the ipykernel install the IW comes up correctly like so:

image

However you can still send code to it. 

It would be nice if it either
- Prevented sending anymore code
- Brought up the dialog again if you sent it again.

### LINKED ISSUE #9450: Cancel install for IW now looks a bit odd
Testing #9434

If you cancel ipykernel install for an IW launch it now looks funny compared to the notebook message.

IW:
image

Notebook message is much better:
image

### LINKED ISSUE #9454: Interactive window still says connecting to old kernel even after a switch
Testing #9434

- Create a blank Python file in a folder with 2 environments, one with ipykernel, and one without (any type: pip, conda, poetry)
- Add a cell with a `print` statement
- Select the environment without ipykernel as your initial kernel
- Click on `Run Cell`, select `Change kernel` on the prompt, and pick the environment with ipykernel
- Observe that the text at the top of the interactive window still says `Connecting to <old env name>`

https://user-images.githubusercontent.com/51720070/159592726-a18e2c2f-e46b-453f-9483-6882b1e6d07c.mp4

DonJayamanne:
Fixes #9450
Fixes #9444
Fixes #9454
Fixes #8384

### DIFF
--- a/src/interactive-window/editor-integration/codewatcher.ts
+++ b/src/interactive-window/editor-integration/codewatcher.ts
@@ -8,8 +8,6 @@ import {
     commands,
     Event,
     EventEmitter,
-    NotebookCell,
-    NotebookEditor,
     Position,
     Range,
     Selection,
@@ -24,7 +22,6 @@ import { IFileSystem } from '../../platform/common/platform/types';
 
 import { IConfigurationService, IDisposable, Resource } from '../../platform/common/types';
 import { chainable } from '../../platform/common/utils/decorators';
-import * as localize from '../../platform/common/utils/localize';
 import { isUri } from '../../platform/common/utils/misc';
 import { StopWatch } from '../../platform/common/utils/stopWatch';
 import { traceDecorators } from '../../platform/logging';
@@ -169,7 +166,7 @@ export class CodeWatcher implements ICodeWatcher {
 
                 // Note: We do a get or create active before all addCode commands to make sure that we either have a history up already
                 // or if we do not we need to start it up as these commands are all expected to start a new history if needed
-                finished = this.addCode(iw, code, this.document.uri, range.start.line, leftCount);
+                finished = this.addCode(iw, code, this.document.uri, range.start.line);
             }
         }
 
@@ -220,7 +217,7 @@ export class CodeWatcher implements ICodeWatcher {
                 // We have a cell and we are not past or at the stop point
                 leftCount -= 1;
                 const code = this.document.getText(range);
-                finished = this.addCode(iw, code, this.document.uri, lens.range.start.line, leftCount);
+                finished = this.addCode(iw, code, this.document.uri, lens.range.start.line);
             } else {
                 // If we get a cell past or at the stop point stop
                 break;
@@ -248,7 +245,7 @@ export class CodeWatcher implements ICodeWatcher {
                 // We have a cell and we are not past or at the stop point
                 leftCount -= 1;
                 const code = this.document.getText(lens.range);
-                finished = this.addCode(iw, code, this.document.uri, lens.range.start.line, leftCount);
+                finished = this.addCode(iw, code, this.document.uri, lens.range.start.line);
             }
         }
 
@@ -273,7 +270,7 @@ export class CodeWatcher implements ICodeWatcher {
             if (!normalizedCode || normalizedCode.trim().length === 0) {
                 return;
             }
-            await this.addCode(iw, normalizedCode, this.document.uri, activeEditor.selection.start.line, 0);
+            await this.addCode(iw, normalizedCode, this.document.uri, activeEditor.selection.start.line);
         }
     }
 
@@ -287,7 +284,7 @@ export class CodeWatcher implements ICodeWatcher {
             );
 
             if (code && code.trim().length) {
-                await this.addCode(iw, code, this.document.uri, 0, 0);
+                await this.addCode(iw, code, this.document.uri, 0);
             }
         }
     }
@@ -302,7 +299,7 @@ export class CodeWatcher implements ICodeWatcher {
             );
 
             if (code && code.trim().length) {
-                await this.addCode(iw, code, this.document.uri, targetLine, 0);
+                await this.addCode(iw, code, this.document.uri, targetLine);
             }
         }
     }
@@ -988,20 +985,11 @@ export class CodeWatcher implements ICodeWatcher {
         // This should be chained so that a queue forms when getting the interactive window
         return this.interactiveWindowProvider.getOrCreate(this.document?.uri);
     }
-
-    private getMatchingCell(file: Uri, line: number, editor: NotebookEditor) {
-        return editor.document
-            .getCells()
-            .find(
-                (c) => c.metadata?.interactive?.uristring === file.toString() && c.metadata?.interactive?.line === line
-            );
-    }
     private async addCode(
         interactiveWindow: IInteractiveWindow,
         code: string,
         file: Uri,
         line: number,
-        leftCount: number,
         debug?: boolean
     ): Promise<boolean> {
         let result = false;
@@ -1013,21 +1001,8 @@ export class CodeWatcher implements ICodeWatcher {
                 result = await interactiveWindow.addCode(code, file, line);
             }
         } catch (err) {
-            if (err instanceof InteractiveCellResultError) {
-                // Only show an error message if any left
-                if (leftCount > 0) {
-                    const message = localize.DataScience.cellStopOnErrorFormatMessage().format(leftCount.toString());
-
-                    // If our cell result was a failure show an error
-                    // for the count of cells cancelled
-                    await this.addErrorMessage(interactiveWindow, (e) => this.getMatchingCell(file, line, e), message);
-                }
-            } else {
+            if (!(err instanceof InteractiveCellResultError)) {
                 await this.dataScienceErrorHandler.handleError(err);
-                // If our cell result was a failure show an error
-                // for the count of cells cancelled
-                const message = await this.dataScienceErrorHandler.getErrorMessageForDisplayInCell(err, 'execution');
-                await this.addErrorMessage(interactiveWindow, (e) => this.getMatchingCell(file, line, e), message);
             }
         } finally {
             this.sendPerceivedCellExecute(stopWatch);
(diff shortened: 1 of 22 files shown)


src/interactive-window/interactiveWindow.ts
```diff
@@ -310,20 +312,25 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
+        const kernelName = 'info' in kernel ? kernel.kernelConnectionMetadata.interpreter?.displayName : '';
+        const kernelInfo = 'info' in kernel && kernel.info?.status === 'ok' ? kernel.info : undefined;
+        const banner = kernelInfo ? kernelInfo.banner.split('\n').join('  \n') : kernel.toString();
         const message =
             reason == SysInfoReason.Restart ? DataScience.restartedKernelHeader().format(kernelName || '') : banner;
         this.updateSysInfoMessage(message, true, cellPromise);
     }
 
+    private async finishWithFailureMessage(error: Error, cellPromise: Promise<NotebookCell | undefined>) {
+        let message = await this.errorHandler.getErrorMessageForDisplayInCell(error, 'start');
```
DonJayamanne, line 328: Using the same code that displays error messages in Notebooks.

IanMatthewHuff (approved):

DonJayamanne:
Updated the PR to fix two issues, as they are very much related.

src/interactive-window/interactiveWindow.ts
```diff
@@ -186,22 +190,35 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
+                    false,
+                    sysInfoCell
+                );
+                const kernel = await connectToKernel(
+                    controller,
+                    this.serviceContainer,
+                    this.owner,
+                    notebook,
+                    undefined,
+                    updateMessage
```
DonJayamanne, line 213: Fixes https://github.com/microsoft/vscode-jupyter/issues/9454
If we pick another kernel as a result of ipykernel not being foundin current kernel, then we need to update the message that reads `connecting to kernel xyz`, this callback provides the ability to just that.

src/interactive-window/interactiveWindow.ts
```diff
@@ -177,7 +181,6 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         return editor;
     }
 
-    @chainable()
```
DonJayamanne, line 180: Fix from https://github.com/microsoft/vscode-jupyter/pull/9458

Brought it here to have the tests pass properly and so that I can write more tests in this PR

src/interactive-window/interactiveWindow.ts
```diff
@@ -548,14 +617,12 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
-            notebookCellPromise,
-            this._editorReadyPromise
-        ]);
-        if (!kernel) {
-            return false;
-        }
+        // Wait for the cell to get created (required to insert error messages into cell output).
+        // When we wait for kernel to start, its possible it will fail, and
+        // higher up, we insert errors into the cell (but the cell needs to have been inserted).
+        const { cell, wasScrolled } = await notebookCellPromise;
```
DonJayamanne, line 623: Not a fan of this, personally, I think we should just add the error messages into the cell right here.
Its weird that CodeWatcher has a method `addCodeCell`, and that handles the errors and the calls into this same class to update the cell with the errors.

I agree error handling should be done higher up, but the way the code is written right now:
* CodeWatcher doesn't handle errors related to ipykernel not being found (that internal block displays messages & handles those errors)
* CodeWatcher handles errors in cells and then calls back into IW to add error messages

Feels all too weird right now, hence this hacky code, its very very easy for this to break (will need a test to ensure we don't break this), to break this all we need to do is use a `Promise.all` as follows (which is the old code):
```typescript
const [kernel, { cell, wasScrolled }, editor] = await Promise.all([
            this._kernelPromise.promise,
            notebookCellPromise,
            this._editorReadyPromise
        ]);
```

DonJayamanne: There are two options:
* Refactor
* Write a test to ensure we don't break hack

DonJayamanne: I prefer the former
Looking at code watcher, we don't do anything special with errors, all we do is handle the error and display it back in the IW.
Moving that into this part will make all of that simpler (IMHO)

src/interactive-window/interactiveWindow.ts
```diff
@@ -537,7 +606,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
     private disconnectKernel() {
         this.kernelDisposables.forEach((d) => d.dispose());
         this.kernelDisposables = [];
-        if (this._kernelPromise.resolved) {
+        if (this._kernelPromise.completed) {
```
DonJayamanne, line 609: If its rejejected, we need to create a new one

src/kernels/helpers.ts
```diff
@@ -2138,6 +2150,7 @@ export async function wrapKernelMethodImpl(
                     deadKernelAction: restarted ? 'deadKernelWasRestarted' : 'deadKernelWasNoRestarted'
                 };
             } else {
+                onAction(context, kernel);
```
DonJayamanne, line 2153: Fixes https://github.com/microsoft/vscode-jupyter/issues/9454

Not sure whether the name `onAction` is good, perhaps `notifyAction` might be better, not sure.

src/platform/datascience/types.ts
```diff
@@ -464,15 +463,9 @@ export interface IInteractiveWindow extends IInteractiveBase {
...
     readonly ready: Promise<void>;
     closed: Event<void>;
-    addCode(code: string, file: Uri, line: number, editor?: TextEditor, runningStopWatch?: StopWatch): Promise<boolean>;
-    addMessage(message: string, getIndex?: (editor: NotebookEditor) => number): Promise<void>;
-    debugCode(
-        code: string,
-        file: Uri,
-        line: number,
-        editor?: TextEditor,
-        runningStopWatch?: StopWatch
```
DonJayamanne, line 474: THe last 2 arguments were never used

DonJayamanne:
sorry @IanMatthewHuff I ended up using this PR to fix other related items, hence the re-request.

FYI - I still need to write up some tests for the fixes

src/interactive-window/interactiveWindow.ts
rchiodo: Sounds good to me. Don't have the codewatcher handle errors. Doesn't seem like it should as how is it to know how the interactive window should display them?

rchiodo (approved):

IanMatthewHuff:
@DonJayamanne No prob. Rereviewing now.

src/interactive-window/interactiveWindow.ts
DonJayamanne: We have two methods today in `IDataScienceErrorHandler`, `handleError` and `getErrorMessageForDisplayInCell` 
Yes, the code watcher handles errors, and my suggestion is to :
* Let the code watcher handle the errors by calling `IDataScienceErrorHandler.handleError`
* Let the IW display errors in the IW cells by calling `IDataScienceErrorHandler.getErrorMessageForDisplayInCell`  & display that message in the cells.

This way we remove the call backs, and the like (call back to get the cell information to insert errors).
However we now need to pass the `leftCount` all the way into the IW, so that it can change the error message based on whether there are any more pending cells..

src/interactive-window/interactiveWindow.ts
```diff
@@ -309,21 +334,50 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
+        kernel: IKernel,
         cellPromise: Promise<NotebookCell | undefined>,
         reason: SysInfoReason
     ) {
-        const kernelName =
-            'info' in kernelOrError ? kernelOrError.kernelConnectionMetadata.interpreter?.displayName : '';
-        const kernelInfo =
-            'info' in kernelOrError && kernelOrError.info?.status === 'ok' ? kernelOrError.info : undefined;
-        const banner = kernelInfo ? kernelInfo.banner.split('\n').join('  \n') : kernelOrError.toString();
+        const kernelName = 'info' in kernel ? kernel.kernelConnectionMetadata.interpreter?.displayName : '';
```
IanMatthewHuff, line 366: Does the kernelName = 'info' make sense here? It's not looking at the kernel.info like below. Maybe kernelConnectionMetadata?

src/interactive-window/interactiveWindow.ts
rchiodo: I think the IW could tell if there are pending cells because they should be in the list of cells in the IW now that the cells are queued up before execution.

So you might need the leftCount.

src/interactive-window/interactiveWindow.ts
DonJayamanne: I'm going to remove this and have a separate method that deals with errors, feels weird to look for `info` or something else to figure out whether its kernel or an error.

src/interactive-window/interactiveWindow.ts
DonJayamanne: Fixed that issue in this PR https://github.com/microsoft/vscode-jupyter/pull/9483, yes we don't need that anymore.

src/interactive-window/interactiveWindow.ts
DonJayamanne: Fixed this issue in this PR https://github.com/microsoft/vscode-jupyter/pull/9483
Basically split into two distinct methods, no overloads

src/interactive-window/editor-integration/codewatcher.ts
```diff
@@ -995,44 +989,45 @@ export class CodeWatcher implements ICodeWatcher {
...
+                    const message = localize.DataScience.cellStopOnErrorFormatMessage().format(leftCount.toString());
+
+                    // If our cell result was a failure show an error
+                    // for the count of cells cancelled
+                    await this.addErrorMessage(interactiveWindow, (e) => this.getMatchingCell(file, line, e), message);
+                }
             } else {
                 await this.dataScienceErrorHandler.handleError(err);
+                // If our cell result was a failure show an error
+                // for the count of cells cancelled
```
amunger, line 1028: It looks like this comment only applies to the error message above

amunger: but this all goes away in your other PR

PR #9458, opened 2022-03-23
No need to chain promises when starting kernels

### LINKED ISSUE #9445: Switching to another kernel that requires ipykernel will ask for install twice
Testing #9435

1. Create two empty venvs
2. Refresh interpreter list in VS code
3. Open python file
4. Set one of the empty venvs as the active interpreter
5. Run cell
6. Pick switch kernel when asked to install ipykernel
7. Pick the second empty venv
8. Pick cancel when asked the second time

Another dialog comes up after that point (meaning you get asked twice for the second kernel)

DonJayamanne:
Fixes #9445

### DIFF
--- a/src/interactive-window/interactiveWindow.ts
+++ b/src/interactive-window/interactiveWindow.ts
@@ -177,7 +177,6 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         return editor;
     }
 
-    @chainable()
     private async startKernel(notebook: NotebookDocument, controller: VSCodeNotebookController): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/interactive-window/interactiveWindow.ts
```diff
@@ -177,7 +177,6 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         return editor;
     }
 
-    @chainable()
```
DonJayamanne, line 180: This wasn't required & this is what caused the issue

rchiodo (approved):

PR #9459, opened 2022-03-23
Interrupting kernel doesn't always display errors (flaky test fix)

DonJayamanne:


src/notebooks/execution/cellExecution.ts
```diff
@@ -527,17 +527,17 @@ export class CellExecution implements IDisposable {
...
         } catch (ex) {
+            traceError('Error in waiting for cell to complete', ex);
             // @jupyterlab/services throws a `Canceled` error when the kernel is interrupted.
             // Such an error must be ignored.
             if (
                 ex &&
                 ex instanceof Error &&
                 (ex.message.includes('Canceled') || ex.message.includes(localize.Common.canceled()))
             ) {
-                this.completedSuccessfully();
```
DonJayamanne, line 537: This was wrong, if you look at the comments in line 530, if interrupting a kernel, jupyter lab will throw `Canceled` error.
And we handle that, but mark the cell as completed successfully, which is incorrect.
In correct because if we have 5 cells to be executed, and we interrupted while 3 was running, now 4 & 5 will also run.

This also explains why we have a flaky test, when we interrupt sometimes we don't fail with errors, the cells stop running, but there are no errors (i.e. we're in this code path)

rchiodo: What if it gets translated though? I think this is better as it checks both english and whatever language is in use.

rchiodo (approved):

IanMatthewHuff (approved):
Approve with the consideration that .startsWith might be a better check here.

src/notebooks/execution/cellExecution.ts
DonJayamanne: > think this is better as it checks both english and whatever language is in use.

in either case (what ever the error is) if there is an error we complete with errors, all we do is now have special logging for this scenario.
hence if they localized the message in the future, it works still work, as we fall into else and still end with errors

PR #9463, opened 2022-03-23
Fix minor regression with execution state

### LINKED ISSUE #9405: Running all cells marks every cell as running
image

Expectation is that only a single cell would show as running while others are in the "pending/queued" state.

Jupyter Extension: v2022.3.1000781005
Version: 1.66.0-insider (Universal)
Commit: db0525ff4ba814780e901a5b9e399aaf596b9b20
Date: 2022-03-17T05:18:05.656Z
Electron: 17.1.1
Chromium: 98.0.4758.109
Node.js: 16.13.0
V8: 9.8.177.11-electron.0
OS: Darwin arm64 21.4.0

rchiodo:
Fixes #9405

We shouldn't create a new cell execution as soon as a cell is queued up. We should wait until we execute unless it's the first cell in a list.

This didn't seem like a big change and it's a regression caused by this month's changes, so thought maybe we should fix it.

### DIFF
--- a/src/notebooks/controllers/vscodeNotebookController.ts
+++ b/src/notebooks/controllers/vscodeNotebookController.ts
@@ -101,7 +101,7 @@ export class VSCodeNotebookController implements Disposable {
      */
     public static kernelAssociatedWithDocument?: boolean;
     private isDisposed = false;
-    private runningCellExecution: NotebookCellExecution | undefined;
+    private runningCellExecutions = new Map<NotebookDocument, NotebookCellExecution>();
     get id() {
         return this.controller.id;
     }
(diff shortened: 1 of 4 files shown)


src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -100,6 +101,7 @@ export class VSCodeNotebookController implements Disposable {
      */
     public static kernelAssociatedWithDocument?: boolean;
     private isDisposed = false;
+    private runningCellExecution: NotebookCellExecution | undefined;
```
IanMatthewHuff, line 104: Is this mapping correct? I would think it's one runningCellExecution per NotebookDocument. The controller could have multiple cells running.

DonJayamanne (approved):

src/notebooks/controllers/vscodeNotebookController.ts
rchiodo: Very good catch. I didn't even think of that.

IanMatthewHuff (approved):

PR #9478, opened 2022-03-24
Ensure we don't display code lenses for Notebook cells

DonJayamanne:


src/interactive-window/editor-integration/codelensprovider.ts
```diff
@@ -70,6 +71,9 @@ export class DataScienceCodeLensProvider implements IDataScienceCodeLensProvider
     // CodeLensProvider interface
     // Some implementation based on DonJayamanne's jupyter extension work
     public provideCodeLenses(document: vscode.TextDocument, _token: vscode.CancellationToken): vscode.CodeLens[] {
+        if (document.uri.scheme != PYTHON_FILE.scheme && document.uri.scheme !== PYTHON_UNTITLED.scheme) {
```
DonJayamanne, line 74: Dirty solution, looks like there's no way to eclude notebooks and seems the behaviour has now changed with the document filters. All notebook cells are included and there's no way to exclude notebook cell Uris via the DocumentFilter.

rchiodo (approved):

PR #9493, opened 2022-03-25
Update Main branch from release

IanMatthewHuff:
News entries + version bump to 2022.04.100

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,86 @@
 # Changelog
 
+## 2022.3.100 (25 March 2022)
+
+### Enhancements
+
+
+### Fixes
+
+1. Add support for named index in dataframe viewer for DataFrames and Series.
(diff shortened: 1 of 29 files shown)


rchiodo:
Does anybody know why the engine version is set to insiders only? I don't recall deciding to only allow prerelease on insiders?

I was looking because it will eventually change to 1.67 but not until after VS code releases the 1.66 version.

rchiodo (approved):

IanMatthewHuff:
> Does anybody know why the engine version is set to insiders only? I don't recall deciding to only allow prerelease on insiders?
> 
> I was looking because it will eventually change to 1.67 but not until after VS code releases the 1.66 version.

Hmmm, there is actually nothing about changing the engine version in the release steps. Right now the release branch is still set to insiders for the engine. I'll have to update that. 

As far as I know prerelease was still targeted at insiders. Not sure if I quite remember why, maybe since insiders can pick up API changes early and pre-release would need to be in sync with that?

Issue #9423, opened 2022-03-18
March 2022 Release Plan


rchiodo:
# Prerequisites

-   Python 3.7 and higher
-   run `python3 -m pip install --user -r news/requirements.txt`

# Release candidate (Friday of VS Code endgame week, XXX XX)

-   [x] Review [Component Governance](https://dev.azure.com/monacotools/Monaco/_componentGovernance/191876) (Click on "microsoft/vscode-jupyter" on that page) and resolve all High/Severe issues.

    -   [x] Manually add any repository dependencies (if you can't add manually, refer [here](https://docs.opensource.microsoft.com/tools/cg/features/cgmanifest/)). Only add a cgmanifest.json if the components are not NPM or are not dev only.
-   [x] Create new release branch with format `release/release-YYYY.MM.100`
    -   [x] Create a pull request against `release/release-YYYY.MM` for changes
    -   [ ] Run `npm install` to verify `package-lock.json` did not get updated.
    -   [x] Update [`CHANGELOG.md`](https://github.com/Microsoft/vscode-jupyter/blob/main/CHANGELOG.md)
        -   [x] Run [`news`](https://github.com/Microsoft/vscode-jupyter/tree/main/news) (typically `python news --final --update CHANGELOG.md | code-insiders -`)
        -   [x] Copy over the "Thanks" section from the previous release into the "Thanks" section for the new release
        -   [x] Make sure the "Thanks" section is up-to-date (e.g. compare to versions in [`requirements.txt`](https://github.com/microsoft/vscode-jupyter/blob/main/requirements.txt))
        -   [x] Touch up news entries (e.g. add missing periods)
        -   [x] Check the Markdown rendering to make sure everything looks good
    -   [x] Update [`ThirdPartyNotices-Repository.txt`](https://github.com/Microsoft/vscode-jupyter/blob/main/ThirdPartyNotices-Repository.txt) as appropriate. This file is manually edited so you can check with the teams if anything needs to be added here.
    -   [x] Merge pull request into `release/release-YYYY.MM`
-   [x] Update the [`release` branch](https://github.com/microsoft/vscode-jupyter/branches)
    -   [x] If there are `release` branches that are two versions old (e.g. release-2020.[current month - 2]) you can delete them at this time (you would have to un-protect the release branches temporarily to delete them).
[...]

PR #9494, opened 2022-03-25
Update release engine version

IanMatthewHuff:


rchiodo (approved):

amunger (approved):

DonJayamanne:
i don't think we do this anymore, don't recall why, but i recall removing this option.
i think it's because we wanted users to be able to keep using vscode insiders, this action (changing engine) used to be required to force the insider release of our extension.

if still required were might want to update our release plan
https://github.com/microsoft/vscode-jupyter/issues/9423

IanMatthewHuff:
@DonJayamanne Already discussing in slack. I think we need it for this release as Aaron's change to New File required an update to 1.66. I've already added a note to add this as needed to the release plan.

DonJayamanne:
> required an update to 1.66. I've already added a note to add this as needed to the release plan.

leaving as insider will also work

rchiodo:
> leaving as insider will also work

Not as far as I know. If it says -insider, it won't load in stable, regardless of what version stable is at.

IanMatthewHuff:
> > required an update to 1.66. I've already added a note to add this as needed to the release plan.
> 
> leaving as insider will also work

@DonJayamanne Sorry I'm a bit confused on your comment here. This updating the release branch. Just like here:
https://github.com/microsoft/vscode-jupyter/pull/9153#issuecomment-1054627430

DonJayamanne:
aah yes thanks (couldn't remember that at all)

, please can we update the release plan, too many steps and plenty of this that can be missed (by me). im certain i would had forgotten this.

IanMatthewHuff:
@DonJayamanne Yup. Anytime I do a release I always take notes and then update the release plan. I just wait to do it in one batch at the end. Here is the current list.

https://github.com/microsoft/vscode-jupyter/issues/9423#issuecomment-1079221603

PR #9500, opened 2022-03-27
Support Uri instead of Notebook in kernel API

### LINKED ISSUE #9495: External Kernel API shouldn't require a notebookDocument to create a kernel
Right now our API requires a NotebookDocument to be passed in to create a kernel:

https://github.com/microsoft/vscode-jupyter/blob/e6d86dabc25d9e63b850251b7e5e576dfb86aaea/src/platform/api/extension.d.ts#L281

This is rather limiting and forces callers to create dummy notebook documents. 

We really only need a unique key to make sure we don't create the same kernel more than once.

DonJayamanne:
Fixes #9495

### DIFF
--- a/src/kernels/kernel.ts
+++ b/src/kernels/kernel.ts
@@ -15,7 +15,8 @@ import {
     NotebookController,
     NotebookDocument,
     ColorThemeKind,
-    Disposable
+    Disposable,
+    Uri
 } from 'vscode';
 import { IApplicationShell, IWorkspaceService } from '../platform/common/application/types';
 import { WrappedError } from '../platform/errors/types';
@@ -151,7 +152,8 @@ export class Kernel implements IKernel {
     private startCancellation = new CancellationTokenSource();
     private startupUI = new DisplayOptions(true);
     constructor(
-        public readonly notebookDocument: NotebookDocument,
+        public readonly notebookDocument: NotebookDocument | undefined,
+        public readonly owner: Uri,
         public readonly resourceUri: Resource,
         public readonly kernelConnectionMetadata: Readonly<KernelConnectionMetadata>,
         private readonly notebookProvider: INotebookProvider,
@@ -217,12 +219,12 @@ export class Kernel implements IKernel {
         if (this.restarting) {
             traceInfo(
                 `Interrupt requested & currently restarting ${(
-                    this.resourceUri || this.notebookDocument.uri
+                    this.resourceUri || this.owner
                 ).toString()}`
             );
             await this.restarting.promise;
         }
-        traceInfo(`Interrupt requested ${(this.resourceUri || this.notebookDocument.uri).toString()}`);
+        traceInfo(`Interrupt requested ${(this.resourceUri || this.owner).toString()}`);
         this.startCancellation.cancel();
         const interruptResultPromise = this.kernelExecution.interrupt(
             this._notebookPromise?.then((item) => item.session)
@@ -231,7 +233,7 @@ export class Kernel implements IKernel {
         const status = this.statusProvider.set(DataScience.interruptKernelStatus());
         let result: InterruptResult | undefined;
         try {
-            traceInfo(`Interrupt requested & sent for ${getDisplayPath(this.notebookDocument.uri)} in notebookEditor.`);
+            traceInfo(`Interrupt requested & sent for ${getDisplayPath(this.owner)} in notebookEditor.`);
             result = await interruptResultPromise;
             if (result === InterruptResult.TimedOut) {
                 const message = DataScience.restartKernelAfterInterruptMessage();
@@ -247,7 +249,7 @@ export class Kernel implements IKernel {
         }
     }
     public async dispose(): Promise<void> {
-        traceInfoIfCI(`Dispose Kernel ${getDisplayPath(this.notebookDocument.uri)}`);
+        traceInfoIfCI(`Dispose Kernel ${getDisplayPath(this.owner)}`);
         this._disposing = true;
         if (this.disposingPromise) {
             return this.disposingPromise;
@@ -255,7 +257,7 @@ export class Kernel implements IKernel {
         this._ignoreNotebookDisposedErrors = true;
         this.startCancellation.cancel();
         const disposeImpl = async () => {
-            traceInfo(`Dispose kernel ${(this.resourceUri || this.notebookDocument.uri).toString()}`);
+            traceInfo(`Dispose kernel ${(this.resourceUri || this.owner).toString()}`);
             this.restarting = undefined;
             const promises: Promise<void>[] = [];
             promises.push(this.kernelExecution.cancel());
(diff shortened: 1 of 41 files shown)


src/platform/api/kernelApi.ts
```diff
@@ -135,19 +135,36 @@ class JupyterKernelService implements IExportedKernelService {
...
         }
     }
-    async startKernel(spec: KernelConnectionMetadata, notebook: NotebookDocument): Promise<IKernelConnectionInfo> {
+    async startKernel(spec: KernelConnectionMetadata, owner: Uri): Promise<IKernelConnectionInfo> {
         sendTelemetryEvent(Telemetry.JupyterKernelApiUsage, undefined, {
             extensionId: this.callingExtensionId,
             pemUsed: 'startKernel'
         });
-        return this.startOrConnect(spec, notebook);
+        return this.startOrConnect(spec, this.getAssociatedNotebook(owner, true));
```
DonJayamanne, line 143: Today our code uses NotebookDocument as the owner for kernels.
We use notebook documents as the keys for holding references to kernels, if notebook document is closed or goes out of scope, then the kernels also get disposed/cleared out.

If we want to propagate this change all the way to IKernel, it can be done, just a little bit more work (minor refactoring). But don't see the need just yet, as all that does is introduce more nullables and more `if else`, where as adding a dummy Notebook document will be sufficient, as such dummy ones are only created from outside this extension and those callers (3rd party extensions) are responsible for managing the lifetime of the kernels.

rchiodo (approved):

src/platform/api/kernelApi.ts
amunger: It seems a bit dangerous to introduce the DummyNotebook in production code (someone could try to use the stored object as a full notebook if they didn't realize it wasn't a real notebook).

If nothing else, the methods should at least all throw Not Implemented, but I'd rather just see the refactor to only use what we need (just the URI?)

DonJayamanne: > omeone could try to use the stored object as a full notebook if they didn't realize it wasn't a real notebook

I don't see how or why we'd ever try to do that. & that's why I'm going with this simple approach.
perhaps I can do that as a debt item

> should at least all throw Not Implemented

Done

src/kernels/variables/debuggerVariables.ts
```diff
@@ -123,7 +123,7 @@ export class DebuggerVariables extends DebugLocationTracker
         if (this.active) {
             // Note, full variable results isn't necessary for this call. It only really needs the variable value.
             const result = this.lastKnownVariables.find((v) => v.name === name);
-            if (result && kernel?.notebookDocument.uri.fsPath.endsWith('.ipynb')) {
+            if (result && kernel?.notebookDocument?.uri.fsPath.endsWith('.ipynb')) {
```
rchiodo, line 126: Does this need to use notebookDocument? Could it just use the uri?

src/kernels/variables/kernelVariables.ts
```diff
@@ -80,7 +80,7 @@ export class KernelVariables implements IJupyterVariables {
         token?: CancellationToken
     ): Promise<IJupyterVariable | undefined> {
         // See if in the cache
-        const cache = this.notebookState.get(kernel.notebookDocument);
+        const cache = kernel.notebookDocument ? this.notebookState.get(kernel.notebookDocument) : undefined;
```
rchiodo, line 83: Same here. Why does it need a notebookDocument?

src/kernels/kernel.ts
```diff
@@ -151,7 +152,8 @@ export class Kernel implements IKernel {
     private startCancellation = new CancellationTokenSource();
     private startupUI = new DisplayOptions(true);
     constructor(
-        public readonly notebookDocument: NotebookDocument,
+        public readonly notebookDocument: NotebookDocument | undefined,
```
rchiodo, line 155: I was hoping we could make this private? The fact that they're tied together seems like it might be private.

src/kernels/kernelProvider.ts
```diff
@@ -123,25 +135,56 @@ export class KernelProvider implements IKernelProvider {
...
     }
+    public getInternal(
+        owner: NotebookDocument | Uri
+    ):
+        | {
+              options: KernelOptions;
+              kernel: IKernel;
+          }
+        | undefined {
+        const notebook = workspace.notebookDocuments.find((nb) =>
```
rchiodo, line 154: I believe this would skip notebooks in the interactive case?

src/kernels/kernel.ts
DonJayamanne: I guess I could, but then I'd have to modify all of the followig code:
```nb = kernel.notebookdocument```
with
```nb = workspace.notebooks.find(nb => nb.uri.toString() === kernel.owner.uri.toString)```

File: src/kernels/kernelProvider.ts (hunk 5 above)
Comment (DonJayamanne) [reply src/kernels/kernelProvider.ts:154]: Ok, need a better name, lets chat in standup
We have two Uris:
* a. One is to track the original Python file for IW, or the original IPynb for Notebooks
* b. The other is the unique identifier for a Kernel
	* When we start a kernel, then we have a notebook IW or IPYnb notebook, in thie case the Uri = Uri of the Notebook Editor
	* For 3rd party extensions we don't have a notebook document, but we still need a Uri.

  
What do we call `a` & `b`
Today:
`kernel.resourceUri` (its the associated resource, in IW its *.py & in IPynb notebook its *.ipynb & when its an empty IW, then this is undefined)  
`kernel.notebookDocument` assicated with the IW notebook or IPYNB notebook

For 3rd party we need something thats unique to each kernel, & I'm proposing a Uri.
Today we have a 1-1 relationship, each kernel is associated with a notebook.
I'd like 1-1 relationship, each kernel is associated with a Uri

This way, if we want to get an existing kernel, we have a key to get it.

File: src/kernels/kernelProvider.ts [hunk 6]
```diff
@@ -86,18 +95,23 @@ export class KernelProvider implements IKernelProvider {
         this._onDidRestartKernel.dispose();
         this._onKernelStatusChanged.dispose();
     }
-    public getOrCreate(notebook: NotebookDocument, options: KernelOptions): IKernel {
-        const existingKernelInfo = this.kernelsByNotebook.get(notebook);
+    public getOrCreate(uri: Uri, options: KernelOptions): IKernel {
+        const existingKernelInfo = this.getInternal(uri);
+        const notebook = workspace.notebookDocuments.find((nb) =>
```
rchiodo, line 100: I think this needs to use the findNotebook api because workspace.notebookDocuments doesn't include interactive notebooks.

rchiodo: But maybe I'm remembering wrong.

rchiodo (approved):

src/kernels/variables/kernelVariables.ts
DonJayamanne: Fixed

src/kernels/kernelProvider.ts
DonJayamanne: > workspace.notebookDocuments 

Nope, it does include the interactive window notebook as well.

PR #9526, opened 2022-03-29
Display no ipykernel error and not display a duplicate no IPyKernel error prompt

### LINKED ISSUE #9510: Prompt to install IPykernel does not always show
sometimes it just goes straight to having the cell fail with:

> Running cells with 'Python 3.9.10 ('venv39')' requires ipykernel package.
> Run the following command to install 'ipykernel' into the Python environment. 
> Command: 'c:/venvs/venv39/Scripts/python.exe -m pip install ipykernel -U --force-reinstall'

and never shows the prompt to install or switch kernels.

It seems to repro fairly consistently with these steps:
1. try to run a cell without IPykernel
2. cancel when prompted
3. reload window
4. run the cell again

### LINKED ISSUE #9521: with zmq disabled, prompt to install notebook package shows twice when cancelled
1. add the setting`"jupyter.disableZMQSupport" : true` to disable zmq
2. open a .py file with a `#%%` cell and set the active interpreter to an environment without the notebook package installed.
3. run the cell and cancel the install prompt that appears.

Actual:
Prompt reappears immediately (just once), and needs to be cancelled again

DonJayamanne:
Fixes #9521
Fixes #9510

### DIFF
--- a/src/kernels/helpers.node.ts
+++ b/src/kernels/helpers.node.ts
@@ -1718,7 +1718,6 @@ export async function wrapKernelMethod(
 
     let currentPromise = connections.get(notebook);
     if (!options.disableUI && currentPromise?.options.disableUI) {
-        traceVerbose(`${initialContext} the kernel, change options.disableUI`);
         currentPromise.options.disableUI = false;
     }
     // If the current kernel has been disposed or in the middle of being disposed, then create another one.
--- a/src/kernels/helpers.node.ts
+++ b/src/kernels/helpers.node.ts
@@ -1584,7 +1584,7 @@ export async function notifyAndRestartDeadKernel(
 export async function handleKernelError(
     serviceContainer: IServiceContainer,
     error: Error,
-    context: 'start' | 'interrupt' | 'restart' | 'execution',
+    errorContext: 'start' | 'interrupt' | 'restart' | 'execution',
     resource: Resource,
     kernel: IKernel,
     controller: NotebookController,
@@ -1593,7 +1593,7 @@ export async function handleKernelError(
     const memento = serviceContainer.get<Memento>(IMemento, GLOBAL_MEMENTO);
     const errorHandler = serviceContainer.get<IDataScienceErrorHandler>(IDataScienceErrorHandler);
 
-    if (metadata.interpreter && context === 'start') {
+    if (metadata.interpreter && errorContext === 'start') {
         // If we failed to start the kernel, then clear cache used to track
         // whether we have dependencies installed or not.
         // Possible something is missing.
@@ -1605,7 +1605,7 @@ export async function handleKernelError(
     // Send telemetry for handling the error (if raw)
     const isLocal = isLocalConnection(metadata);
     const rawLocalKernel = serviceContainer.get<IRawNotebookProvider>(IRawNotebookProvider).isSupported && isLocal;
-    if (rawLocalKernel && context === 'start') {
+    if (rawLocalKernel && errorContext === 'start') {
         sendKernelTelemetryEvent(resource, Telemetry.RawKernelSessionStartNoIpykernel, {
             reason: handleResult
         });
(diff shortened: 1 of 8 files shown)


src/kernels/helpers.node.ts
```diff
@@ -1663,12 +1663,13 @@ async function verifyKernelState(
     serviceContainer: IServiceContainer,
     resource: Resource,
     notebook: NotebookDocument,
-    options: IDisplayOptions = new DisplayOptions(false),
```
DonJayamanne, line 1666: Removed this nullable type, this was always passed except in one place.

src/kernels/helpers.node.ts
```diff
@@ -1751,11 +1758,15 @@ export async function wrapKernelMethod(
...
-                }
-            });
+            result.kernel.onDisposed(
+                () => {
+                    if (connections.get(notebook)?.kernel === deferred) {
+                        connections.delete(notebook);
+                    }
+                },
+                undefined,
+                disposables
```
DonJayamanne, line 1768: We weren't disposing this.

src/notebooks/controllers/vscodeNotebookController.node.ts
```diff
@@ -486,12 +485,19 @@ export class VSCodeNotebookController implements Disposable {
         // Execution should be ended elsewhere
     }
 
-    @chainable()
```
DonJayamanne, line 489: This wasn't required, we don't need to chain starting the kernel.

src/platform/errors/errorHandler.node.ts
```diff
@@ -53,9 +53,14 @@ export class DataScienceErrorHandler implements IDataScienceErrorHandler {
         @inject(IKernelDependencyService) private readonly kernelDependency: IKernelDependencyService,
         @inject(IWorkspaceService) private readonly workspaceService: IWorkspaceService
     ) {}
+    private handledErrors = new WeakSet<Error>();
     public async handleError(err: Error): Promise<void> {
         traceWarning('DataScience Error', err);
         err = WrappedError.unwrap(err);
+        if (this.handledErrors.has(err)) {
```
DonJayamanne, line 60: We end using this in a few places, this ensures we don't handle the same error over & over.

src/platform/errors/errorHandler.node.ts
```diff
@@ -151,12 +156,22 @@ export class DataScienceErrorHandler implements IDataScienceErrorHandler {
...
     ): Promise<KernelInterpreterDependencyResponse> {
-        traceWarning('Kernel Error', err);
+        traceWarning(`Kernel Error, context = ${context}`, err);
         err = WrappedError.unwrap(err);
 
         // Jupyter kernels, non zmq actually do the dependency install themselves
         if (err instanceof JupyterKernelDependencyError) {
-            return err.reason;
+            traceWarning(`Jupyter Kernel Dependency Error, reason=${err.reason}`, err);
+            if (err.reason === KernelInterpreterDependencyResponse.uiHidden) {
```
DonJayamanne, line 165: If we don't have the dependencies and UI was disabled, now during the error handling display the UI.

src/kernels/helpers.node.ts
```diff
@@ -1584,7 +1584,7 @@ export async function notifyAndRestartDeadKernel(
 export async function handleKernelError(
     serviceContainer: IServiceContainer,
     error: Error,
-    context: 'start' | 'interrupt' | 'restart' | 'execution',
+    errorContext: 'start' | 'interrupt' | 'restart' | 'execution',
```
DonJayamanne, line 1587: I renamed this, as there's already a global variable named `context` from mocha and that causes issues.
Basically i was using `context` in logging and it crashed during runtime, however ts never complained because that code path was passing the compiler due to it using the global `context` variable.

This way we dont run into such issues and TS compiler will still pick issues correctly

src/platform/errors/errorHandler.node.ts
DonJayamanne: This fixes #9521

src/notebooks/controllers/vscodeNotebookController.node.ts
rchiodo: You should remove the comment below then.

rchiodo (approved):

src/kernels/helpers.node.ts
IanMatthewHuff: Wow, good catch. Kinda scary that mocha was messing with type checking in the main product code. It's only for tests, right?

DonJayamanne: >  It's only for tests, right?

Yes mocha is only used in tests, but its type def files declare a global variable that lets TS into thinking that those global vars are valid in the extension source as well. (I don't think its possible to work around, this unless issue)

PR #9545, opened 2022-03-31
Remove winston for logging (not web compatible)

rchiodo:
Winston is nodejs only. 

This changes our logging to match what the Python extension did (much simpler).

It does however maintain the trace decorators that Don added.


rchiodo:
Telemetry and localization also had to be web specific (used by logging) so moved those too.

Most of the changes are in localize.ts and the logging code. Everything else is just name changes.

src/platform/logging/types.ts
```diff
@@ -0,0 +1,47 @@
...
+    Warn = 30,
+    Info = 20,
+    Debug = 10,
+    Trace = 5,
+    Off = 0
+}
+
+export type Arguments = unknown[];
+
+export interface ILogging {
```
amunger, line 19: Seems like it should be called ILogger. Is this just what Python named it?

rchiodo: Yeah I just copied it. I agree. I'll change the name.

src/extension.node.ts
```diff
@@ -210,9 +214,28 @@ async function activateComponents(
     return activateLegacy(context, serviceManager, serviceContainer);
 }
 
+function addConsoleLogger() {
+    if (process.env.VSC_JUPYTER_FORCE_LOGGING) {
+        let label = undefined;
+        // In CI there's no need for the label.
+        const isCI = process.env.TF_BUILD !== undefined || process.env.GITHUB_ACTIONS === 'true';
```
IanMatthewHuff, line 221: Aren't we already importing an "isCI" into this file?

rchiodo: Thanks. Didn't notice that.

IanMatthewHuff (approved):

amunger (approved):

rchiodo:
I love our tests 🥇 . Windows test failed for interrupt because the daemon path couldn't be constructed for the interrupt daemon module. Root cause - path-browserify won't join rooted paths. It just drops the path altogether.

PR #9550, opened 2022-03-31
Reduce content displayed when installing packages

### LINKED ISSUE #9451: Conda install output in the notification is a bit noisy
Testing #9434

Just my opinion here. But there is lots of random info that now pops up in the notification toast when installing.

image

On one hand, it's a long process, so the message changing lets you know that progress is being made. But on the other hand, it has some unhelpful messages like above. I think I'd personally rather just have an installing message.

DonJayamanne:
Fixes #9451

### DIFF
--- a/src/kernels/installer/moduleInstaller.node.ts
+++ b/src/kernels/installer/moduleInstaller.node.ts
@@ -120,7 +120,7 @@ export abstract class ModuleInstaller implements IModuleInstaller {
                                 output.out.length > 100
                                     ? `${output.out.substring(0, 50)}...${output.out.substring(output.out.length - 50)}`
                                     : output.out;
-                            progress?.report({ message });
+                            progress.report({ message });
                         }
                         traceInfo(output.out);
                         if (output.source === 'stderr') {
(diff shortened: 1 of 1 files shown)


rchiodo (approved):

src/kernels/installer/moduleInstaller.node.ts
```diff
@@ -113,7 +115,13 @@ export abstract class ModuleInstaller implements IModuleInstaller {
             if (observable) {
                 observable.out.subscribe({
                     next: (output) => {
-                        progress?.report({ message: output.out });
+                        if (progress) {
+                            const message =
+                                output.out.length > 100
+                                    ? `${output.out.substring(0, 50)}...${output.out.substring(output.out.length - 50)}`
+                                    : output.out;
+                            progress?.report({ message });
```
IanMatthewHuff, line 123: Not sure if this actually changes anything, but looks like you added the if check, so the ? wouldn't be needed here anymore.

IanMatthewHuff (approved):

src/kernels/installer/moduleInstaller.node.ts
DonJayamanne: Fixed, no idea why `progress` was marked as nullable, it isn't. Removed the `if` condition (i added that to reduce unnecessary processing)

PR #9559, opened 2022-04-01
remove unused Title field and simplify CellMatcher usage

amunger:


src/interactive-window/editor-integration/cellFactory.node.ts
```diff
@@ -151,14 +151,10 @@ export function generateCellRangesFromDocument(document: TextDocument, settings?
                 previousCell.range = new Range(previousCell.range.start, document.lineAt(index - 1).range.end);
             }
 
-            const results = matcher.exec(line.text);
-            if (results !== undefined) {
```
amunger, line 155: the matcher.IsCell check above is a tighter check, and we don't need the results from exec - title was unused

rchiodo (approved):

PR #9569, opened 2022-04-04
Use path instead of path-browserify

### LINKED ISSUE #9568: Exclude path-browserify
We seem to have path and path-browserify in our code.
I dont think we should be using path-browserify directly in our code.

Purely injecting them in the webpack config (as is done today) should suffice.
Note: We have test failures related to the use of path-browserify

Bringing into april, as this needs to be resolved, including this into desktop product could cause issues.

DonJayamanne:
Fixes #9568

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -290,7 +290,7 @@ module.exports = {
                 ]
             }
         ],
-        'local-rules/node-imports': ['error', { allow: ['events'] }],
+        'local-rules/node-imports': ['error', { allow: ['events', 'path'] }],
         strict: 'off'
     },
     settings: {
--- a/src/platform/common/platform/fs-paths.ts
+++ b/src/platform/common/platform/fs-paths.ts
@@ -1,5 +1,5 @@
 import { Uri, WorkspaceFolder } from 'vscode';
-import * as path from 'path-browserify';
+import * as path from 'path';
 
 export function getDisplayPath(
     filename?: string | Uri,
--- a/src/platform/common/application/workspace.ts
+++ b/src/platform/common/application/workspace.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 import { injectable } from 'inversify';
-import * as path from 'path-browserify';
+import * as path from 'path';
 import {
     CancellationToken,
     ConfigurationChangeEvent,
--- a/src/webviews/extension-side/variablesView/variableView.node.ts
+++ b/src/webviews/extension-side/variablesView/variableView.node.ts
@@ -4,7 +4,6 @@
 import '../../../platform/common/extensions';
 
 import * as path from 'path';
-import * as pathBrowser from 'path-browserify';
 import { WebviewView as vscodeWebviewView } from 'vscode';
 
 import { captureTelemetry, sendTelemetryEvent } from '../../../telemetry';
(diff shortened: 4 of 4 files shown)


src/platform/common/platform/fs-paths.ts
```diff
@@ -1,5 +1,5 @@
 import { Uri, WorkspaceFolder } from 'vscode';
-import * as path from 'path-browserify';
```
DonJayamanne, line 2: We don't need to force the use of `path-browserify` in webpack we can swtich the `path` to `path-browserify`.
Note: This has already been done, hence this makes it easier to use this file (code) with both node & desktop code.

DonJayamanne: This also fixes the failures in the tests

rchiodo: I disagree. This is making it harder to determine what code works in the browser and which doesn't. I think the better change here is to switch fs-paths.ts to fs-paths.node.ts and remove the 'path' allow from the linter.

rchiodo (changes requested):
I think we should rename fs-paths back to fs-paths.node

.eslintrc.js
```diff
@@ -290,7 +290,7 @@ module.exports = {
                 ]
             }
         ],
-        'local-rules/node-imports': ['error', { allow: ['events'] }],
+        'local-rules/node-imports': ['error', { allow: ['events', 'path'] }],
```
rchiodo, line 293: This is my biggest worry. Removing this here means everywhere path will be allowed and it doesn't work so good in web.

src/platform/common/platform/fs-paths.ts
rchiodo: Oh actually I've already split them in the main extension. What was the test failure then? I think it's likely we need another linter rule to detect usage of uri.fsPath as that should only be allowed in '.node.ts' files.

rchiodo: Ah this is the test failure:
https://github.com/microsoft/vscode-jupyter/runs/5815274707?check_suite_focus=true

I believe this can be fixed by importing from the common fs-paths instead of the fs-paths.node.ts in the test that's failing.

DonJayamanne: I think we should discuss this, having `path` and `path-browserify` complicates things, this is exactly why we have aliases in webpack, so we can replace things with what we need. here we're duplicating everything, and thats a lot of work.

.eslintrc.js
DonJayamanne: Else we ship path-browserify in node code when `path` is provided natively in node and thats better than any replacement for node, i.e. we're not using native `path` module at all now, that to me is not right at all.

src/platform/common/platform/fs-paths.ts
rchiodo: I agree, we should talk about this in a meeting. 

I see a number of options:
- We use webpack to figure out which path we use. This has the downside of you not realizing which one is in use when you write the code (and you make assumptions that it will work in web)
- We write our own path that works like node no matter what (it supports windows still separators)
- We have web/node versions of common things. You have to be explicit about which one you're using. This has the downside of having duplicate code

DonJayamanne:
Closing as other PR was already merged.

PR #9577, opened 2022-04-04
Fix failing unit test

### LINKED ISSUE #9560: built-in module override message has bad path on Windows
The unit test `Unable to import <name> from user overriding module in workspace folder (windows)` is currently failing after re-enabling it and confirmed the message in insiders:
image

rchiodo:
merge problems prevented unit tests from running on last submission. This fixes the path checking to use the same algorithm as the errorUtils.

Fixes #9560

### DIFF
--- a/src/platform/common/platform/fs-paths.ts
+++ b/src/platform/common/platform/fs-paths.ts
@@ -6,9 +6,7 @@ export function getDisplayPath(
     workspaceFolders: readonly WorkspaceFolder[] | WorkspaceFolder[] = []
 ) {
     const relativeToHome = getDisplayPathImpl(filename);
-    const relativeToWorkspaceFolders = workspaceFolders.map((folder) =>
-        getDisplayPathImpl(filename, folder.uri.fsPath)
-    );
+    const relativeToWorkspaceFolders = workspaceFolders.map((folder) => getDisplayPathImpl(filename, folder.uri.path));
     // Pick the shortest path for display purposes.
     // As those are most likely relative to some workspace folder.
     let bestDisplayPath = relativeToHome;
@@ -28,7 +26,7 @@ function getDisplayPathImpl(filename?: string | Uri, cwd?: string): string {
     } else if (!filename) {
         file = '';
     } else if (filename.scheme === 'file') {
-        file = filename.fsPath;
+        file = filename.path;
     } else {
         file = filename.toString();
     }
--- a/src/platform/common/platform/fs-paths.ts
+++ b/src/platform/common/platform/fs-paths.ts
@@ -20,24 +20,29 @@ export function getDisplayPath(
 }
 
 function getDisplayPathImpl(filename?: string | Uri, cwd?: string): string {
+    // Common file separator is unix based '/'. Handle mixing of paths
+    let cwdReplaced = cwd ? cwd.replace(/\\/g, '/') : undefined;
+    if (cwdReplaced?.includes(':') && cwdReplaced.startsWith('/')) {
+        cwdReplaced = cwdReplaced.slice(1);
+    }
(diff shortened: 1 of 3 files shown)


amunger (approved):

amunger:
it's a real bug, not just a test failure https://github.com/microsoft/vscode-jupyter/issues/9560. (If that's the right test)

src/platform/common/platform/fs-paths.ts
```diff
@@ -6,9 +6,7 @@ export function getDisplayPath(
     workspaceFolders: readonly WorkspaceFolder[] | WorkspaceFolder[] = []
 ) {
     const relativeToHome = getDisplayPathImpl(filename);
-    const relativeToWorkspaceFolders = workspaceFolders.map((folder) =>
-        getDisplayPathImpl(filename, folder.uri.fsPath)
```
rchiodo, line 10: fsPath should be illegal in non node files. I'm going to work on adding a linter rule for this.

rchiodo: This is what will happen now:
image

rchiodo: That doesn't seem right though. Making it handle relative paths now.

src/test/datascience/errorHandler.unit.test.ts
```diff
@@ -231,7 +231,7 @@ suite('DataScience Error Handler Unit Tests', () => {
 
             const expectedMessage = DataScience.failedToStartKernelDueToImportFailureFromFile().format(
                 'Random',
-                getDisplayPath('c:\\Development\\samples\\pySamples\\sample1\\kernel_issues\\start\\random.py', [])
```
DonJayamanne, line 234: I think we should discuss this in standup, we shouldn't be using `path-browserify` in node (desktop), why not use the native `path` module instead of a replacement. This is what webpack should do, in web, use `path-browserify` and in desktop use native `path`, instead of always using a custom module instead of natives.

DonJayamanne: Based on the usage for `path-browserify` you don't need to do what you're doing here, their recommendation is to continue using `path` directly
https://www.npmjs.com/package/path-browserify#install

rchiodo: Yeah for the errorUtils case, maybe this is a exception. It could have webpack determine which path to use.

Although that test was wrong as the code it was testing never called getDisplayPath.

DonJayamanne (commented):
Based on their (path-browserfiy) documentation you don't need to directly using `path-browserify`
https://www.npmjs.com/package/path-browserify#install

rchiodo:
> Based on their (path-browserfiy) documentation you don't need to directly using `path-browserify` https://www.npmjs.com/package/path-browserify#install

Yes it's an explicit thing so that the code knows it's using a unix only separator.

src/test/datascience/errorHandler.unit.test.ts
DonJayamanne: Not sure what we're talking about now, i'm referring to usage documentation

DonJayamanne:
> Yes it's an explicit thing so that the code knows it's using a unix only separator.

Yes, I agree, and that should be good enough for us. Right now with your code even in web it will always use unix separator, so it the same. i don't see how your solution circumvents that issue

src/test/datascience/errorHandler.unit.test.ts
rchiodo: It's explicit so that the users of that code realize what's happening. The problem with using path-browserify in the web is that you get unintended side effects

PR #9580, opened 2022-04-04
Preliminary web extension (with a simple command)

rchiodo:
First step in a web extension (I hope). I can run it under the debugger. 

Does nothing but provide the 'Jupyter.ClearCache' command which we only show in development mode. 

This is really just a refactor to make sure stuff will load in the web.

Highlights (other than a lot of renaming):
- New build task for compiling for web (Compile Web Extension)
- New linters to check for usage of process.env
- Package.json added an enablement clause to every command so they only show up in the node version for now
- Extension.web.ts - just the minimum stuff to bring up the extension. Mostly logging and the command registry.
- ExtensionActivationManager was split into two pieces.
   - First piece that activated the other IExtensionSingleActivation parts
   - Second piece that listened to workspace open events and sent activation telemetry
   - Also removed the migrateDataScienceSettings special activation (only activated on workspace open) that we don't really need anymore.


src/intellisense/pythonKernelCompletionProvider.node.ts
```diff
@@ -30,6 +30,11 @@ import { IInteractiveWindowProvider } from '../interactive-window/types';
 import { Settings } from '../platform/common/constants';
 import { INotebookCompletion } from './types';
 
+let IntellisenseTimeout = Settings.IntellisenseTimeout;
+export function setIntellisenseTimeout(timeoutMs: number) {
```
rchiodo, line 34: Cant use process.ENV anymore (except for node specific code). Tests were not node specific so used this instead.

src/platform/activation/migrateDataScienceSettingsService.node.ts
```diff
@@ -1,192 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
```
rchiodo, line 2: This is the one class that implements IExtensionActivationService. Instead of having to handle this anymore, just removed it. I think we don't need to migrate people off of the old python extension anymore. It's been more than a year.

rchiodo:
TBD: Create a test that verifies the extension loads in the web. 

I'll probably add a smoke test for that to begin with.

src/platform/common/application/applicationEnvironment.node.ts
```diff
@@ -5,19 +5,20 @@
...
-import * as vscode from 'vscode';
 import { IPlatformService } from '../platform/types';
 import { IExtensionContext, IPathUtils } from '../types';
 import { OSType } from '../utils/platform';
-import { Channel, IApplicationEnvironment } from './types';
+import { BaseApplicationEnvironment } from './applicationEnvironment.base';
 
 @injectable()
-export class ApplicationEnvironment implements IApplicationEnvironment {
+export class ApplicationEnvironment extends BaseApplicationEnvironment {
```
rchiodo, line 14: THis was refactored into two classes.
One web, and one node.

The node one supports returning a userSettingsFile and a userCustomKeybindingsFile. The web class will return undefined for these two items (for now)

src/platform/common/configuration/service.base.ts
```diff
@@ -0,0 +1,90 @@
...
+// Licensed under the MIT License.
+
+import { ConfigurationTarget, Uri, WorkspaceConfiguration } from 'vscode';
+import { IServiceContainer } from '../../ioc/types';
+import { IWorkspaceService } from '../application/types';
+import { JupyterSettings } from '../configSettings';
+import { isTestExecution, isUnitTestExecution } from '../constants';
+import { IConfigurationService, IWatchableJupyterSettings } from '../types';
+
+export abstract class BaseConfigurationService implements IConfigurationService {
```
rchiodo, line 11: ConfigurationService was also refactored into a base class and a node/web variant.

This is because SystemVariables uses node only classes so it was split as well into a node/web variant.

src/platform/common/variables/systemVariables.ts
```diff
@@ -0,0 +1,90 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+'use strict';
+import * as Types from '../utils/sysTypes';
+import { IStringDictionary, ISystemVariables } from './types';
+
+/* eslint-disable , @typescript-eslint/no-explicit-any, , jsdoc/check-alignment,jsdoc/check-indentation,jsdoc/newline-after-description, no-restricted-syntax, prefer-const,  */
+
+export abstract class AbstractSystemVariables implements ISystemVariables {
```
rchiodo, line 11: AbstractSystemVariables was common to both node and web implementations.

src/platform/common/variables/systemVariables.node.ts
```diff
@@ -1,97 +1,13 @@
...
-    private __resolveArray(value: string[]): string[] {
-        return value.map((s) => this.__resolveString(s));
-    }
-
-    private __resolveAnyArray<T>(value: T[]): T[];
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    private __resolveAnyArray(value: any[]): any[] {
-        return value.map((s) => this.resolveAny(s));
-    }
-}
```
rchiodo, line 94: Node version has relativePath operations. Not included yet in the web version

src/platform/common/constants.ts
```diff
@@ -43,22 +43,6 @@ export const DEFAULT_INTERPRETER_SETTING = 'python';
 
 export const STANDARD_OUTPUT_CHANNEL = 'STANDARD_OUTPUT_CHANNEL';
 
-export const isCI = process.env.TF_BUILD !== undefined || process.env.GITHUB_ACTIONS === 'true';
```
rchiodo, line 46: Cant use process in web. Well not without predefining every variable we might want and that's not really the intent. 

Most of these I changed to global state that is set by other locations.

src/platform/logging/index.ts
```diff
@@ -12,7 +12,7 @@ import { CallInfo, trace as traceDecorator } from '../common/utils/decorators';
 import { TraceInfo, tracing as _tracing } from '../common/utils/misc';
 import { argsToLogString, returnValueToLogString } from './util';
 import { LoggingLevelSettingType } from '../common/types';
-const homeAsLowerCase = (require('untildify')('~') || '').toLowerCase(); // TODO: This will have to be conditional for node only
```
rchiodo, line 15: Untildify uses node specific stuff, so this has to be set outside of logging.

src/platform/serviceRegistry.node.ts
```diff
@@ -5,9 +5,10 @@
 import { IServiceManager } from '../platform/ioc/types';
 import { CommandRegistry } from './commands/commandRegistry.node';
 import { registerTypes as registerApiTypes } from './api/serviceRegistry.node';
-import { registerTypes as commonRegisterTypes } from './common/serviceRegistry.node';
-import { registerTypes as commonRegisterTerminalTypes } from './terminals/serviceRegistry.node';
-import { registerTypes as activationRegisterTypes } from './activation/serviceRegistry.node';
+import { registerTypes as registerCommonTypes } from './common/serviceRegistry.node';
```
rchiodo, line 8: This was review feedback that @IanMatthewHuff gave. I agree, names were inconsistent.

build/webpack/webpack.extension.web.config.js
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+'use strict';
+
+const path = require('path');
+const tsconfig_paths_webpack_plugin = require('tsconfig-paths-webpack-plugin');
+const webpack = require('webpack');
+const constants = require('../constants');
+const CleanTerminalPlugin = require('clean-terminal-webpack-plugin');
```
rchiodo, line 9: This is the webpack config for the web output of the extension.

It generates:
- extension.web.bundle.js
- test/vscode.test/index.bundle.js
- test/smoke.test/index.bundle.js

The test bundles shouldn't be included in the extension, but are necessary for running tests.

build/webpack/webpack.extension.web.config.js
```diff
@@ -0,0 +1,133 @@
...
+    externals: ['vscode', 'commonjs', 'electron'], // Don't bundle these
+    plugins: [
+        new webpack.ProvidePlugin({
+            process: 'process/browser' // provide a shim for the global `process` variable
+        }),
+        new webpack.DefinePlugin({
+            // Definitions...
+            BROWSER: JSON.stringify(true),
+            process: {
+                platform: 'web'
```
rchiodo, line 104: This lets getOSType() return 'web' when asked.

.vscode/tasks.json
```diff
@@ -27,7 +27,18 @@
                 "kind": "build",
                 "isDefault": true
             },
-            "problemMatcher": ["$tsc-watch", "$ts-checker-webpack-watch"]
+            "problemMatcher": ["$ts-webpack-watch"]
+        },
+        {
+            "label": "Compile Web Extension",
+            "type": "npm",
+            "script": "compile-web-watch",
```
rchiodo, line 35: If you want to test/run any of the web version of stuff, you can run this task now.

It's more finicky than normal ts though. Sometimes errors it generates aren't shown in the problems window.

rchiodo: Note how this is different than building the webview-ui parts. Those are separate bundles. For now. I might have to include them when I get any of the webview UI building

src/platform/activation/workspaceActivation.node.ts
```diff
@@ -1,40 +1,24 @@
...
-    IExtensionActivationService,
-    IExtensionSingleActivationService,
-    IExtensionSyncActivationService
-} from './types';
 import { traceDecoratorError } from '../logging';
+import { IExtensionSingleActivationService } from './types';
 
 @injectable()
-export class ExtensionActivationManager implements IExtensionActivationManager {
+export class WorkspaceActivation implements IExtensionSingleActivationService {
```
rchiodo, line 15: This class split into two parts. 
- WorkspaceActivation - initializes getting interpreters and sending workspace telemetry
- ExtensionActivationManager - initializes the IExtensionSingleActivationServices (both sync and async)

src/platform/api/types.ts
```diff
@@ -152,3 +153,8 @@ export const ILanguageServerProvider = Symbol('ILanguageServerProvider');
 export interface ILanguageServerProvider {
     getLanguageServer(resource?: InterpreterUri): Promise<ILanguageServer | undefined>;
 }
+
+export const IExportedKernelServiceFactory = Symbol('IExportedKernelServiceFactory');
+export interface IExportedKernelServiceFactory {
+    getService(): Promise<IExportedKernelService | undefined>;
```
rchiodo, line 159: To make this work for web, I split this out into a service factory. This way we can have a web and a node version of the JupyterKernelServiceFactory.

src/platform/common/application/workspace.ts
```diff
@@ -25,9 +25,14 @@ export class WorkspaceService implements IWorkspaceService {
...
-        return Array.isArray(workspace.workspaceFolders) && workspace.workspaceFolders.length > 0
-            ? workspace.workspaceFolders[0].uri.fsPath
-            : undefined;
+        const firstWorkspace =
+            Array.isArray(workspace.workspaceFolders) && workspace.workspaceFolders.length > 0
+                ? workspace.workspaceFolders[0]
+                : undefined;
+        if (firstWorkspace) {
+            // eslint-disable-next-line local-rules/dont-use-fspath
+            return getOSType() === OSType.Web ? firstWorkspace.uri.path : firstWorkspace.uri.fsPath;
```
rchiodo, line 34: Special cased this linter rule here. fsPath is no longer allowed in anything except .node.ts files but for this 'rootPath', I figured it made sense to still use it.

src/platform/errors/errorUtils.ts
```diff
@@ -551,6 +551,7 @@ function isBuiltInModuleOverwritten(
         return;
     }
 
+    // eslint-disable-next-line local-rules/dont-use-fspath
```
rchiodo, line 554: This actually doesn't work without using fsPath

src/platform/api.ts
```diff
@@ -87,20 +87,29 @@ export function buildApi(
...
-        process.env.VSC_JUPYTER_EXPOSE_SVC ||
-        context.extensionMode === ExtensionMode.Development
+        context.extensionMode === ExtensionMode.Development ||
+        context.extensionMode === ExtensionMode.Test
     ) {
         /* eslint-disable @typescript-eslint/no-explicit-any */
         (api as any).serviceContainer = serviceContainer;
         (api as any).serviceManager = serviceManager;
+        (api as any).getSymbol = (symbol: String) => {
+            // Needed for web tests. Eval for the extension happens in another web worker so
```
rchiodo, line 106: When running the web tests, the `Symbol(IApplicationShell)` was not the same value as found in the container. So I added this extra export to allow a different bundle to get the correct value of symbols.

I believe what's happening is that the extension bundle is eval'd and then in a separate worker, the test bundle is eval'd. Global data is different between them, so I needed a way to export it from the extension bundle.

This function will need to be updated as we write more tests for the web extension.

src/test/web/smoke.test/smoke.web.test.ts
```diff
@@ -0,0 +1,26 @@
...
+        const extension = vscode.extensions.getExtension<IExtensionTestApi>(JVSC_EXTENSION_ID)!;
+        const api = await extension.activate();
+        await api.ready;
+        extensionApi = api;
+    });
+
+    test('Verify containers', () => {
+        const appShellSymbol = extensionApi.getSymbol<IApplicationShell>('IApplicationShell');
+        assert.ok(appShellSymbol, `Cannot get the symbol for IApplicationShell`);
+        const appShell = extensionApi.serviceManager?.get<IApplicationShell>(appShellSymbol!);
```
rchiodo, line 23: First test just verifies we can talk to the service container.

rchiodo: Meaning our extension can load.

src/test/web/vscode.test/extension.web.test.ts
```diff
@@ -0,0 +1,24 @@
...
+        registerLogger(new ConsoleLogger('WEB TEST'));
+    }
+
+    suiteSetup(() => {
+        addConsoleLogger();
+    });
+    void vscode.window.showInformationMessage('Start all tests.');
+
+    test('Sample test', () => {
+        assert.strictEqual(-1, [1, 2, 3].indexOf(5));
```
rchiodo, line 21: This test isn't running yet. I plan on adding more tests here later.

src/webviews/webview-side/common/main.ts
```diff
@@ -1,7 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-/* eslint-disable @typescript-eslint/no-unused-vars */
+export {};
```
rchiodo, line 3: I believe this lets webpack build with the `declare __webpack_public_path` there.

src/intellisense/pythonKernelCompletionProvider.node.ts
DonJayamanne: Curious why are we preventing the use of `process.env`?
If we allowed it, then we wouldn't need this change

src/extension.web.ts
```diff
@@ -0,0 +1,324 @@
...
+//===============================================
+// loading ends here
+
+// These persist between activations:
+let activatedServiceContainer: IServiceContainer | undefined;
+
+/////////////////////////////
+// public functions
+
+export async function activate(context: IExtensionContext): Promise<IExtensionApi> {
```
rchiodo, line 87: This is the main entry point for the web version of the extension.

package.json
```diff
@@ -276,58 +277,61 @@
                 "command": "jupyter.replayPylanceLog",
                 "title": "Replay Pylance Log",
                 "category": "Jupyter (Dev)",
-                "enablement": "jupyter.development && notebookType == jupyter-notebook && isWorkspaceTrusted"
+                "enablement": "jupyter.development && notebookType == jupyter-notebook && isWorkspaceTrusted && !jupyter.webExtension"
```
rchiodo, line 280: This is how I disabled all of the commands in web. By adding a new contextKey : jupyter.webExtension.

src/intellisense/pythonKernelCompletionProvider.node.ts
DonJayamanne: Not entirely sure, but if we want we can always inject env variables into the web bundle using the `Define` section in webpack. Again, not sure why we're disabling the use of `process.env`

package.json
rchiodo: clearCache is the only command available (which means none in the shipping code)

package.json
```diff
@@ -2075,24 +2115,27 @@
...
         "lint": "eslint -c .eslintrc.js --ext .ts src",
         "prettier-fix": "prettier 'src/**/*.ts*' --write && prettier 'build/**/*.js' --write",
         "clean": "gulp clean",
         "updateBuildNumber": "gulp updateBuildNumber",
         "webpack": "webpack",
         "download-api": "vscode-dts dev",
         "postdownload-api": "vscode-dts main",
-        "generateTelemetry": "gulp generateTelemetryMd"
+        "generateTelemetry": "gulp generateTelemetryMd",
+        "open-in-browser": "vscode-test-web --extensionDevelopmentPath=. .",
```
rchiodo, line 2137: 'open-in-browser' is how you'd verify the web extension without the debugger.

.vscode/launch.json
```diff
@@ -31,6 +31,19 @@
                 "order": 1
             }
         },
+        {
+            "name": "Extension (web)",
```
rchiodo, line 35: This is how you debug the web extension. It requires you run the 'Compile Web Extension' task first (which I talked about in the tasks.json)

src/platform/api.ts
DonJayamanne, line 106: Not following this.
I thought the tests will run in the same context as the webworker where the extension is running.
After all, that's how we're able to access static variables and the like.
Also here we're running the same value `IApplicationShell`

I think what's happening is we might be ending up with two bundles, which results in two copies of IApplicationShell.
Optionally you can use `export const IApplicationShell = Symbol.for("IApplicationShell")` that will work everywhere.

src/platform/common/application/workspace.ts
DonJayamanne: I think it would be good to add a debt item to ensure we use Uris instead of strings.
If `rootPath` were to reutn the Uri, then we woudln't have to do any of this, I believe that's what @roblourens also suggested, & I can see that saving us from writing this if in a number of places.

src/platform/api.ts
rchiodo: There are two bundles. This is how I export the value to be read in the other one.

rchiodo: > After all, that's how we're able to access static variables and the like.

Not sure this is going to work based on what happened with the IApplicationShell.

src/platform/common/constants.ts
DonJayamanne: > Well not without predefining every variable we might want and that's not really the intent.

Aah got it, I think you've answered the question i had.

src/platform/api.ts
rchiodo: What's the difference between `Symbol('IApplicationShell') and Symbol.for('IApplicationShell')`

src/platform/common/process/environmentActivationService.node.ts
```diff
@@ -60,7 +60,8 @@ export const defaultShells = {
     [OSType.Windows]: { shell: 'cmd', shellType: TerminalShellType.commandPrompt },
     [OSType.OSX]: { shell: 'bash', shellType: TerminalShellType.bash },
     [OSType.Linux]: { shell: 'bash', shellType: TerminalShellType.bash },
-    [OSType.Unknown]: undefined
+    [OSType.Unknown]: undefined,
+    [OSType.Web]: undefined
```
DonJayamanne, line 64: Hmm, I think we might want OSWindowsWeb, OSLinuxWeb, OSMacWeb.
As OSWeb seems too ambiguous.
Then again, I guess we could always add that later.

DonJayamanne: Because saying an OS is web feel very wrong, if anything is the UI/deployment/runtime kind

DonJayamanne: I think we should discuss this, the more i think of this, I feel it could end up being one of those things that will cause confusions.

rchiodo: It doesn't for me. When you're in a web extension you can't tell what the OS is. The only thing you can tell is that you're running in a browser.

src/platform/errors/errorUtils.ts
DonJayamanne: Why not normalize the separators and always use `/` then we can use an includes or the like?
Then again, woudl'nt this code path crash in web, as `fsPath` is undefined!

src/platform/logging/index.ts
```diff
@@ -40,6 +40,10 @@ export function setLoggingLevel(level?: LoggingLevelSettingType): void {
     globalLoggingLevel = logLevelMap.get(level) ?? LogLevel.Error;
 }
 
+export function setHomeDirectory(homeDir: string) {
+    homeAsLowerCase = homeDir.toLocaleLowerCase();
```
DonJayamanne, line 44: DId you mean `toLowerCase`

DonJayamanne (changes requested):
All looks ok, except for:
* OSWeb - concerned with the confusion this could cause
* Symbols - trying to access symbols. I think we might have two bundles generated which is causing IApplicationShell to get inserted into two files hence tests pull from one location & extension is pulling from yet another place (that's would explain why its different)

src/platform/api.ts
DonJayamanne: With `Symbol.for` the symbol is unique across the app domain.

src/platform/errors/errorUtils.ts
rchiodo: Oh yeah it would probably crash. Not used yet obviously. 

It will never include 'path' though because of the stupid '/' on the front. I guess I could account for that. 

I'd rather worry about this later when we actually get kernels working in the web extension.

src/platform/api.ts
DonJayamanne: >  This is how I export the value to be read in the other one.

That explains the reason for the issue.

src/platform/errors/errorUtils.ts
rchiodo: Cause before that we're going to switch to using the common vscode path stuff and try to use URIs everywhere.

src/platform/api.ts
DonJayamanne: > This is how I export the value to be read in the other one.

Why do we have two bundles? Can't the test code reference the JS files from the worker?
I would have expected they could, after all its all running in the same execution context.

src/platform/logging/index.ts
rchiodo: I actually thought toLocaleLowerCase would be more appropriate, but not comparing it that way later. I'll switch it back.

src/platform/api.ts
rchiodo: Look here for more information:
https://github.com/microsoft/vscode-test-web

That's how the test code works. It loads the test bundle in a server and loads the extension in webworker.

build/webpack/webpack.extension.web.config.js
DonJayamanne: Out of curiousity, why do we need a bundle for the web tests? Just compiling them using typescript should suffice.

That would resolve the Symbol issue with IApplicationShell & the like.

DonJayamanne: Same for smoke tests, 
i.e. creating a bundle is useful for publishing else i don't think we need to create a bundle.

rchiodo: We need a bundle so the test server can load it. They get loaded into a web context too.

DonJayamanne: Or is it because the vscode runtime cannot load a non-bundled (split bundle) JS file?

If thats the case then this makes sense, we could run into other issues as well.
As we'd get types & objects imported incorrectly.
E.g. we coudl be referencing a class in test/index.bundle.js, but that would actually be referencing a class in the test bundle and not the actually extension bundle, & that would cause a lot of issues with the tests.

DonJayamanne: E.g. things like ts-mockito and a lot of those test concepts will not work

DonJayamanne (approved):

src/platform/common/application/workspace.ts
rchiodo: Will do

src/platform/common/process/environmentActivationService.node.ts
rchiodo: I removed the OS.Web type but used process.platform to determine if running in the web or not still.
Added another function to determine if 'isWeb' instead.

PR #9587, opened 2022-04-04
Display kernel failure msg in last executed cell

### LINKED ISSUE #9375: Can we deal with user code (packages) causing kernel failures
* We know tensorflow incorrectly installed could cause the kernel to crash
Such as importing tensorflow See https://github.com/microsoft/vscode-jupyter/issues/7600 , https://github.com/microsoft/vscode-jupyter/issues/4645
* Similarly other packages
	* https://github.com/microsoft/vscode-jupyter/issues/9368

Misc:
* https://github.com/microsoft/vscode-jupyter/issues/9283

**Suggestions**
* Improve error message dispalyed in the popup
	* When kernel crashes provide a more meaningful message about what happened, users might not know what a kernel is, hence might ignore the error message or wouldn't be able to connect execution of cell to an error message about `kernel failure`
* Include error message in cell output
	* This way they know what cell caused the failure
* Should we pay specific focus to packages such as tensorflow, numpy, fastparquet (and provide a link to a Wiki)
	* Note: We do this today for kernel startup failures (e.g. if we have a DLL exception, we provide error information and a link to a WIKI)	    
	* Provide instructions in the error output specific to some packages (with links to WIKI)

DonJayamanne:
Fixes #9375


src/kernels/jupyter/launcher/notebookStarter.node.ts
```diff
@@ -198,11 +199,13 @@ export class NotebookStarter implements Disposable {
         const promisedArgs: Promise<string>[] = [];
         promisedArgs.push(Promise.resolve('--no-browser'));
         promisedArgs.push(Promise.resolve(this.getNotebookDirArgument(workingDirectory)));
-        // When kernels fail to start, Jupyter will attempt to restart 5 times,
-        // & this is very slow (we dont want users to wait because of kernel failures).
-        // Also when kernels die, we don't restart automatically with raw kernels,
-        // We should'nt do the same with jupyter (else startup code will not run).
-        promisedArgs.push(Promise.resolve('--KernelManager.autorestart=False'));
+        if (this.context.extensionMode === ExtensionMode.Test) {
```
DonJayamanne, line 202: Reverting this change, else this breaks jupyter.
I.e. if a kernel dies, then we never get a response back for anything, we don't know whether the died or not.
With auto start enabled, jupyter will at a minimum re-start the kernel and we know that it was auto-restarted as the kernel died. Note: Jupyter auto restarts when kernels die.

Hence this change is required (& it was wrong to add)

src/kernels/jupyter/launcher/notebookStarter.node.ts
```diff
@@ -198,11 +198,6 @@ export class NotebookStarter implements Disposable {
         const promisedArgs: Promise<string>[] = [];
         promisedArgs.push(Promise.resolve('--no-browser'));
         promisedArgs.push(Promise.resolve(this.getNotebookDirArgument(workingDirectory)));
-        // When kernels fail to start, Jupyter will attempt to restart 5 times,
-        // & this is very slow (we dont want users to wait because of kernel failures).
-        // Also when kernels die, we don't restart automatically with raw kernels,
-        // We should'nt do the same with jupyter (else startup code will not run).
-        promisedArgs.push(Promise.resolve('--KernelManager.autorestart=False'));
```
DonJayamanne, line 205: Reverting this change, else this breaks jupyter.
I.e. if a kernel dies, then we never get a response back for anything, we don't know whether the died or not.
With auto start enabled, jupyter will at a minimum re-start the kernel and we know that it was auto-restarted as the kernel died. Note: Jupyter auto restarts when kernels die.

Hence this change is required (& it was wrong to add)

src/kernels/crashAnalysis/kernelCrashMonitor.node.ts
```diff
@@ -0,0 +1,85 @@
...
+'use strict';
+import type { KernelMessage } from '@jupyterlab/services';
+import { inject, injectable } from 'inversify';
+import { NotebookCell } from 'vscode';
+import { IExtensionSyncActivationService } from '../../platform/activation/types';
+import { IApplicationShell } from '../../platform/common/application/types';
+import { IDisposableRegistry } from '../../platform/common/types';
+import { DataScience } from '../../platform/common/utils/localize';
+import { endCellAndDisplayErrorsInCell } from '../../platform/errors/errorUtils';
+import { getDisplayNameOrNameOfKernelConnection } from '../helpers.node';
```
DonJayamanne, line 12: This file does not have to be specific to Node, it would work perfectly fine in Web as well, in fact it should.
However `getDisplayNameOrNameOfKernelConnection` is specific to node. We'd definitely need a web version of this as well (as this is used to generate the display names for controllers).

There are number of ways to fix this, 
* Introduce an interface for this so that we can inject two different classes during runtime, one that uses web specific & other for node.
* Or get this same file to work in all cases (that might be more work)


@rchiodo which way should we go?
Ideally we can go with the latter, as all we need to get the display names are types (e.g. Python interpreter types) & reference to the `path` module. Basically it gets down to `path` module and the non-web types.
Personally I think its ok to expose the Python interpreter types into web as well, even though they are noops.

Basically types don't do anything in the webcase, i.e. we cannot have kernels with Python interpreter information, but that doesn't mean we shoudl not expose the types in web case.  I think its easier to just expose them (again a noop).

src/kernels/kernelAutoRestartMonitor.node.ts
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import type { KernelMessage } from '@jupyterlab/services';
+import { inject, injectable } from 'inversify';
+import { IExtensionSyncActivationService } from '../platform/activation/types';
+import { IDisposable, IDisposableRegistry } from '../platform/common/types';
+import { DataScience } from '../platform/common/utils/localize';
+import { IStatusProvider } from '../platform/progress/types';
+import { getDisplayNameOrNameOfKernelConnection } from './helpers.node';
```
DonJayamanne, line 9: Same as the previous, this file should work even in web, however `getDisplayNameOrNameOfKernelConnection` is only available in node.

src/kernels/kernelCommandListener.node.ts
```diff
@@ -235,42 +209,4 @@ export class KernelCommandListener implements IDataScienceCommandListener {
                 .ignoreErrors();
         }
     }
-    private onDidStartKernel(kernel: IKernel) {
-        this.kernelsStartedSuccessfully.add(kernel);
-    }
-    private onKernelStatusChanged({ kernel }: { status: KernelMessage.Status; kernel: IKernel }) {
```
DonJayamanne, line 241: Remvoed code form here as:
* progress for restarting kernels has nothing to do with command handler
* DIsplaying warnings about dead kernels has nothing to do with command handler

As both of the above are two distinct features, i've moved into 2 separate files.

src/kernels/crashAnalysis/kernelCrashMonitor.node.ts
rchiodo: Yeah once we have the generic vscode 'path' I think we can have this be a generic version of this file.

PR #9619, opened 2022-04-07
Use a common displayPath function

DonJayamanne:
Found that getDisplayNameOfKernel is still node specific because of a few things & one of the biggest blocker was the usage of `path`, But thats no longer the case, hense removed the node specific `getDIsplayPath`

Found while working on PR https://github.com/microsoft/vscode-jupyter/pull/9617


rchiodo (approved):

rchiodo:
Merging because I'm changing this same stuff with the switch to using URIs so didn't want to have a bigger merge conflict later

PR #9631, opened 2022-04-07
Add support for returning a preferred controller from the 'jupyter.selectjupyteruri' command

rchiodo:
Based on what @rebornix was talking about, this should make it possible to get the 'preferred' controller after setting a remote server URI

### DIFF
--- a/src/platform/common/application/commands.ts
+++ b/src/platform/common/application/commands.ts
@@ -180,7 +180,11 @@ export interface ICommandNameArgumentTypeMapping extends ICommandNameWithoutArgu
     [DSCommands.ShowDataViewer]: [IShowDataViewerFromVariablePanel];
     [DSCommands.RefreshDataViewer]: [];
     [DSCommands.ClearSavedJupyterUris]: [];
-    [DSCommands.SelectJupyterURI]: [boolean | undefined, Uri | 'toolbar' | 'nativeNotebookStatusBar' | undefined];
+    [DSCommands.SelectJupyterURI]: [
+        boolean | undefined,
+        Uri | 'toolbar' | 'nativeNotebookStatusBar' | undefined,
+        NotebookDocument | undefined
+    ];
     [DSCommands.SelectNativeJupyterUriFromToolBar]: [];
     [DSCommands.DebugNotebook]: [];
     [DSCommands.RunByLine]: [NotebookCell];
--- a/src/test/datascience/notebook/helper.ts
+++ b/src/test/datascience/notebook/helper.ts
@@ -40,7 +40,7 @@ import { GLOBAL_MEMENTO, IDisposable, IMemento } from '../../../platform/common/
 import { createDeferred } from '../../../platform/common/utils/async';
 import { swallowExceptions } from '../../../platform/common/utils/misc';
 import { IKernelProvider } from '../../../platform/../kernels/types';
-import { IExtensionTestApi, sleep, waitForCondition } from '../../common.node';
+import { sleep, waitForCondition } from '../../common.node';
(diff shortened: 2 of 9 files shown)


src/test/datascience/notebook/helper.ts
```diff
@@ -383,14 +382,12 @@ export async function waitForKernelToGetAutoSelected(expectedLanguage?: string,
...
-export async function startJupyterServer(api?: IExtensionTestApi) {
-    const { serviceContainer } = api ? { serviceContainer: api.serviceContainer } : await getServices();
+export async function startJupyterServer(notebook?: NotebookDocument) {
     if (IS_REMOTE_NATIVE_TEST) {
-        const selector = serviceContainer.get<JupyterServerSelector>(JupyterServerSelector);
         const uri = await JupyterServer.instance.startJupyterWithToken();
         const uriString = decodeURIComponent(uri.toString());
         traceInfo(`Jupyter started and listening at ${uriString}`);
-        await selector.setJupyterURIToRemote(uriString);
+        return commands.executeCommand('jupyter.selectjupyteruri', false, uri, notebook);
```
rchiodo, line 390: @rebornix this is how you'd call the command. 

Arguments are:
- arg[1] = boolean indicating local or remote. Set to false
- arg[2] = URI of the jupyter server
- arg[3] = NotebookDocument of what you want a preferred kernel for

src/platform/common/application/commands.ts
```diff
@@ -180,7 +180,11 @@ export interface ICommandNameArgumentTypeMapping extends ICommandNameWithoutArgu
     [DSCommands.ShowDataViewer]: [IShowDataViewerFromVariablePanel];
     [DSCommands.RefreshDataViewer]: [];
     [DSCommands.ClearSavedJupyterUris]: [];
-    [DSCommands.SelectJupyterURI]: [boolean | undefined, Uri | 'toolbar' | 'nativeNotebookStatusBar' | undefined];
+    [DSCommands.SelectJupyterURI]: [
+        boolean | undefined,
+        Uri | 'toolbar' | 'nativeNotebookStatusBar' | undefined,
+        NotebookDocument | undefined
```
DonJayamanne, line 186: @rchiodo should we change this to execCommand('selectJupyter...', {uri: Uri, document: NotebookDocument})`
I kinda feel thats a bit better than the current overloads.
E.g. the first argument `local= true` is not used an having everyone to pass a bogus value of `false` makes it unnecessary & feels like a poorly constructed API (I understand its historical)
& feel going with an object is better long term.

DonJayamanne: I.e. we can keep backwards compatibility & also have the new Object syntax for new code going forward.

rchiodo: The problem is there is a 3rd party using this command. If we change the arguments it will break them.

DonJayamanne: > The problem is there is a 3rd party using this command. If we change the arguments it will break them.

We could make it backwards compatible.
I.e. if we get an object as the first argument, then we know its the new style. Because the first argument today is a `boolean|undefined`.

DonJayamanne: Yes, its another overload, but going forward I kind of like the object approach, makes it very flexible.

DonJayamanne: approving PR as this is definitely not a blocker, but think its worth exploring

DonJayamanne (approved):

src/platform/common/application/commands.ts
rchiodo: Seems weird to me

PR #9641, opened 2022-04-09
Switch to using URIs as much as possible

### LINKED ISSUE #9599: Switch to using URIs everywhere possible
This makes the transition to web/node easier.

rchiodo:
Fixes #9599


src/kernels/jupyter/interpreter/jupyterInterpreterOldCacheStateStore.node.ts
```diff
@@ -26,15 +27,15 @@ export class JupyterInterpreterOldCacheStateStore {
...
+        const globalState = persistentStateFactory.createGlobalPersistentState<Uri>('DS-VSC-JupyterInterpreter-2');
         this.workspaceJupyterInterpreter = { state: workspaceState };
         this.globalJupyterInterpreter = { state: globalState };
     }
     private get cacheStore(): CacheInfo {
         return this.workspace.hasWorkspaceFolders ? this.workspaceJupyterInterpreter : this.globalJupyterInterpreter;
     }
-    public getCachedInterpreterPath(): string | undefined {
+    public getCachedInterpreterPath(): Uri | undefined {
         return this.cacheStore.state.value;
```
DonJayamanne, line 39: I don't think we can cache (serlalize Uris) in this manner. Basically it would only get serialized as a JSON, hence all of the methods would not be availabel upon deserializing and things could fall over. I think we should store the string value and use `Uri.file` when reading back from state.value

src/kernels/raw/finder/jupyterPaths.node.ts
```diff
@@ -125,9 +128,9 @@ export class JupyterPaths {
      * https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs
      */
     @traceDecoratorVerbose('Get Kernelspec root path')
-    public async getKernelSpecRootPaths(cancelToken?: CancellationToken): Promise<string[]> {
+    public async getKernelSpecRootPaths(cancelToken?: CancellationToken): Promise<Uri[]> {
         // Paths specified in JUPYTER_PATH are supposed to come first in searching
-        const paths = new Set<string>(await this.getJupyterPathPaths(cancelToken));
+        const paths = new Set<Uri>(await this.getJupyterPathPaths(cancelToken));
```
DonJayamanne, line 133: I don't think this would work, Uri.file('a') !== Uri.file('a')
Hence the uniqueness isn't going to work with the Set class

src/kernels/jupyter/interpreter/jupyterInterpreterOldCacheStateStore.node.ts
rchiodo: Yes the JSON only works if something had actually called 'fsPath' before. I'm actually working on this part right now.

src/kernels/raw/finder/jupyterPaths.node.ts
rchiodo: Thanks forgot about this one. I need to use the comparison key.

src/kernels/raw/finder/jupyterPaths.node.ts
```diff
@@ -56,64 +53,69 @@ export class JupyterPaths {
...
             (async () => {
-                if (this.platformService.isWindows) {
-                    // On windows the path is not correct if we combine those variables.
-                    // It won't point to a path that you can actually read from.
-                    return tryGetRealPath(path.join(this.pathUtils.home, winJupyterPath));
-                } else if (this.platformService.isMac) {
-                    return path.join(this.pathUtils.home, macJupyterPath);
-                } else {
-                    return path.join(this.pathUtils.home, linuxJupyterPath);
+                const userHomeDir = getUserHomeDir();
```
rchiodo, line 69: Should be 1 location to get the user home dir now. We had like 3 before.

src/platform/common/utils/platform.node.ts
```diff
@@ -16,9 +22,12 @@ export function getPathEnvironmentVariable(): string | undefined {
     return getEnvironmentVariable('Path') || getEnvironmentVariable('PATH');
 }
 
-export function getUserHomeDir(): string | undefined {
+export function getUserHomeDir(): Uri | undefined {
     if (getOSType() === OSType.Windows) {
-        return getEnvironmentVariable('USERPROFILE');
+        return fsPathToUri(getEnvironmentVariable('USERPROFILE') || homePath);
```
rchiodo, line 27: Without the homePath stuff tests will fail on windows (for linux based tests)

src/kernels/types.ts
```diff
@@ -319,7 +320,7 @@ export interface IJupyterKernelSpec {
     id?: string;
     name: string;
     language?: string;
-    path: string;
+    path: Uri;
```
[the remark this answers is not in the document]
rchiodo: Maybe as 'uri'?

src/platform/api/extension.d.ts
```diff
@@ -57,12 +57,12 @@ export type PythonVersion = {
 };
 export type PythonEnvironment = {
     displayName?: string;
-    path: string;
+    path: Uri;
```
IanMatthewHuff, line 60: Another possible path=>Uri name for the member.

IanMatthewHuff (approved):
There were a couple path variables that are now Uris that we could consider renaming. Looks good though.

src/kernels/types.ts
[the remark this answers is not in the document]
IanMatthewHuff: I think that's more what I would expect. But a variable called path also being a Uri wouldn't really surprise me, so feel free to change or not based on what you think. Not a big diff for me.

src/kernels/helpers.node.ts
```diff
@@ -4,6 +4,7 @@
 'use strict';
 
 import * as path from '../platform/vscode-path/path';
```
DonJayamanne, line 6: Do we have linters to ensure nodejs module `path` never gets used.
Or would we want to just replace `path` with `vscode-path` in webpack?

I'd think one of the above would be necessary to ensure we don't use path in web files.

rchiodo: We do have a linter that ensures no more usage of 'path' from nodejs anywhere except for '.node' files.

src/kernels/helpers.node.ts
```diff
@@ -1205,7 +1218,7 @@ function compareAgainstKernelDisplayNameInNotebookMetadata(
         argv: [],
         display_name: notebookMetadata.kernelspec.display_name,
         name: notebookMetadata.kernelspec.name,
-        path: ''
+        uri: Uri.file('')
```
DonJayamanne, line 1221: This feels very risky. I think we should mark `path` as a nullable property
I know we have code paths where we check if `path` is empty e.g. `if (path)`, 
Such a condition would evaluate to false as the string is empty, now these would never show up in compilations and they'd always be  `true` because its always an object and never an empty string.

rchiodo: Those were all handled with the rename. They wouldn't compile.

rchiodo: Cause before it used to have stuff like this:
interpreter || { path: string }. 

Now when path is accessed it says interpreter doesn't have it. 

The only risk here is if we had forced things to 'any'.

rchiodo: And well our integration tests and unit tests would likely have caught it.

rchiodo: InterpreterOrUri was another example of where it was checking for '.path'. After the change of .path to a URI those still worked. After this rename, these turned into a compile time error and were actually correct. So this rename was necessary to catch that usage.

PR #9643, opened 2022-04-11
Determine how many are using OCaml and the like

DonJayamanne:


rchiodo (approved):

src/kernels/raw/finder/remoteKernelFinder.node.ts
```diff
@@ -119,6 +120,7 @@ export class RemoteKernelFinder implements IRemoteKernelFinder {
                 // Turn them both into a combined list
                 const mappedSpecs = await Promise.all(
                     specs.map(async (s) => {
+                        sendKernelSpecTelemetry(s, 'remote');
```
amunger, line 123: maybe a bit excessive to send an event for every kernel we find, every time it's listed

DonJayamanne: > maybe a bit excessive to send an event for every kernel we find, every time it's listed

No it isn't, if you have 10 kernels, we only send 10 events. For the whole lifetime of the session.

amunger: oh, I thought we searched for kernels more often. never mind

amunger (approved):

Issue #9650, opened 2022-04-11
Log can be difficult to read or interpret 


greazer:
### Applies To

- [X] Notebooks (.ipynb files)
- [X] Interactive Window and\/or Cell Scripts (.py files with \#%% markers)

### What happened?

When the extension encounters a problem, it will sometimes suggest to the user to take a look at the Jupyter log to see what may have gone wrong. Unfortunately, this log can be extremely difficult to read and understand if you're not an engineer working on it. :)

Here's a few suggestions to consider as to how to better handle this:
* Default the ```Jupyter  > Logging: Level``` setting to something less wordy than "debug". One that is likely to more likely to contain just enough information for our customers to help solve common problems.
* Create two separate logging panels for Jupyter. One called "Jupyter" and the other called "Jupyter Detailed Log" or something like that. They would contain identical information, but the latter would contain many more details that may only be useful when entering a new issue. 
* Continue to make changes to the extension that can reduce or avoid simple "please consult the log" notifications and instead direct the user to how to fix any detected problems. 
* Provide a "one-click" method of entering an issue that automatically provides detailed logs in the issue itself.

### VS Code Version

Version: 1.67.0-insider (user setup) Commit: 6ea335e334b90a9aca753d558dec1d14741eff1e Date: 2022-04-08T07:13:56.078Z Electron: 17.4.0 Chromium: 98.0.4758.141 Node.js: 16.13.0 V8: 9.8.177.13-electron.0 OS: Windows_NT x64 10.0.22593

### Jupyter Extension Version

v2022.4.1001001004

DonJayamanne:
> Default the Jupyter > Logging: Level setting to something less wordy than "debug". One that is likely to more likely to contain just enough information for our customers to help solve common problems.

I think we still log a lot of unnecessary stuff. Should reduce the logging of some of the stuff.

DonJayamanne:
Found that the telemetry logging happened recently and that's what makes the logging more noisy. 
We should review the logs after telemetry logging is enabled only for verbose logging.

PR #9657, opened 2022-04-11
Reduce logging

DonJayamanne:
For #9650

### DIFF
--- a/src/platform/logging/index.ts
+++ b/src/platform/logging/index.ts
@@ -67,7 +67,7 @@ export function traceInfo(message: string, ...args: Arguments): void {
 }
 
 export function traceVerbose(message: string, ...args: Arguments): void {
-    if (globalLoggingLevel >= LogLevel.Debug) {
+    if (globalLoggingLevel <= LogLevel.Debug) {
         loggers.forEach((l) => l.traceVerbose(message, ...args));
     }
 }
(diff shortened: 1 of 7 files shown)


src/platform/logging/index.ts
```diff
@@ -67,7 +67,7 @@ export function traceInfo(message: string, ...args: Arguments): void {
 }
 
 export function traceVerbose(message: string, ...args: Arguments): void {
-    if (globalLoggingLevel >= LogLevel.Debug) {
+    if (globalLoggingLevel <= LogLevel.Debug) {
```
DonJayamanne, line 70: Submitting a PR now, as this increases the logging even in insiders, & when logging the telemetry its very noisy.
I think this was meant to be the other way around.
Basically if we want to log just errors today we will still log this information.

src/platform/logging/index.ts
```diff
@@ -80,7 +80,7 @@ export function traceInfoIfCI(message: string, ...args: Arguments): void {
 /** Logging Decorators go here */
 
 export function traceDecoratorVerbose(message: string, opts: TraceOptions = DEFAULT_OPTS): TraceDecoratorType {
-    return createTracingDecorator({ message, opts, level: LogLevel.Debug });
+    return createTracingDecorator({ message, opts, level: LogLevel.Trace });
```
DonJayamanne, line 83: Also found that verbose logging is used in a number of places, verbose is not the same as debug.
This terminology is confusing - carry on from Python extension ( i think we should just make it consistent instead of having verbose = debug & info = trace, & the like).
But thats a debt item

src/platform/logging/index.ts
rchiodo: The problem is it's not clear what the level means. Is higher stricter? It seems that way given this change, but maybe that's not clear everywhere else.

PR #9694, opened 2022-04-16
Remove the knowledge of ILocalKernelFinder and IRemoteKernelFinder from Notebook Controller Manager

### LINKED ISSUE #9689: Abstract the concept of 'local' and 'remote' from the NotebookControllerManager
NotebookControllerManager has explicit knowledge of this distinction. 

Seems to me like it should treat all kernel spec sources the same.

This work item is to try and abstract out the source of the kernelspec from the controller manager (so the logic won't have any node specific stuff in it - which the local kernels do)

rchiodo:
Fixes #9689 

@DonJayamanne wanted to know your opinion on this. 

The genesis for this is that the notebook controller manager has explicit knowledge of local kernels. For this to work in web we need to remove this knowledge.

This is my attempt to do that by:
- Creating an IKernelFinder that doesn't explicitly say where kernels come from
- Changing NotebookControllerManager to use this new finder
- Moving caching into this IKernelFinder
- Moving some of the telemetry in the NotebookControllerManager into the IKernelFinder
- findKernel could actually be shared between remote and local (uses the same algorithm)
- ILocalKernelFinder no longer needs to cache. It's handled by the caller
- IRemoteKernelFinder gets caching for free

This is just an idea, wanted to know what you thought of it. 

I think right now I have validation of the cached kernels in the wrong spot. This could probably move to the ILocalKernelFinder and the IRemoteKernelFinder.


rchiodo:
Dang it, how do I turn it back into a draft?

src/kernels/kernelFinder.base.ts
```diff
@@ -0,0 +1,297 @@
...
+        private readonly notebookProvider: INotebookProvider,
+        private readonly localKernelFinder: ILocalKernelFinder | undefined,
+        private readonly remoteKernelFinder: IRemoteKernelFinder | undefined,
+        private readonly globalState: Memento
+    ) {}
+
+    // Finding a kernel is the same no matter what the source
+    @traceDecoratorVerbose('Find kernel spec', TraceOptions.BeforeCall | TraceOptions.Arguments)
+    @captureTelemetry(Telemetry.KernelFinderPerf)
+    public async findKernel(
```
rchiodo, line 55: This is common to local/remote

rchiodo: And I think it probably didn't work correctly before? It did them separately.

src/kernels/kernelFinder.base.ts
```diff
@@ -0,0 +1,297 @@
...
+                this.remoteKernelFinder
+                    ? this.remoteKernelFinder.listKernels(resource, connInfo, cancelToken)
+                    : Promise.resolve([]),
+            cancelToken,
+            'remote',
+            useCache
+        ).then((l) => this.finishListingKernels(l, useCache, 'remote'));
+    }
+
+    private async listKernelsUsingFinder(
```
rchiodo, line 183: Caching happens in this function. Should be the same logic you had before.

src/kernels/kernelFinder.base.ts
```diff
@@ -0,0 +1,297 @@
...
+    private finishListingKernels(
+        list: KernelConnectionMetadata[],
+        useCache: 'ignoreCache' | 'useCache',
+        kind: 'local' | 'remote'
+    ) {
+        // Send the telemetry once for each type of search
+        const key = `${kind}:${useCache}`;
+        if (this.startTimeForFetching && !this.fetchingTelemetrySent.has(key)) {
+            this.fetchingTelemetrySent.add(key);
+            sendTelemetryEvent(Telemetry.FetchControllers, this.startTimeForFetching.elapsedTime, {
```
rchiodo, line 234: This telemetry used to be the in the NotebookControllerManager

src/kernels/raw/finder/localKernelFinder.node.ts
```diff
@@ -3,309 +3,44 @@
...
         @inject(LocalKnownPathKernelSpecFinder) private readonly nonPythonKernelFinder: LocalKnownPathKernelSpecFinder,
         @inject(LocalPythonAndRelatedNonPythonKernelSpecFinder)
-        private readonly pythonKernelFinder: LocalPythonAndRelatedNonPythonKernelSpecFinder,
-        @inject(IMemento) @named(GLOBAL_MEMENTO) private readonly globalState: Memento,
-        @inject(IFileSystem) private readonly fs: IFileSystem
+        private readonly pythonKernelFinder: LocalPythonAndRelatedNonPythonKernelSpecFinder
     ) {}
-    @traceDecoratorVerbose('Find kernel spec', TraceOptions.BeforeCall | TraceOptions.Arguments)
-    @captureTelemetry(Telemetry.KernelFinderPerf)
-    public async findKernel(
```
rchiodo, line 85: Find and caching were moved up a level.

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -528,33 +428,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             return;
         }
 
-        const updatePromise = this.computePreferredNotebookController(document);
-        if (!this.isLocalLaunch) {
-            void updatePromise.finally(() => {
-                if (this.isLocalLaunch) {
-                    return;
-                }
-                if (this.failedToFetchRemoteKernels) {
```
rchiodo, line 537: I removed this altogether for now. I assume you're doing this differently anyway.

src/telemetry/index.ts
```diff
@@ -262,8 +263,8 @@ function sendTelemetryEventInternal<P extends IEventNamePropertyMapping, E exten
 
         reporter.sendTelemetryEvent(eventNameSent, customProperties, measures);
     }
-
-    traceVerbose(
+    traceWithLevel(
```
rchiodo, line 266: Oh moved this to an even higher trace level. Telemetry spew was bothering me too.

src/kernels/helpers.ts
```diff
@@ -486,6 +486,11 @@ export function compareKernels(
             return 1;
         } else if (b === activeInterpreterConnection) {
             return -1;
+            // Interpreter should trump kernelspec
+        } else if (a.kind === 'startUsingPythonInterpreter' && b.kind !== 'startUsingPythonInterpreter') {
```
IanMatthewHuff, line 490: Out of interest, why would this be the case?

rchiodo: Well the tests were guaranteeing that so I assume that was the wanted behavior. I think before it was just luck that it was happening.

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -207,102 +225,17 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
-                .finally(() => {
-                    if (!this.isLocalLaunch) {
-                        const cancellation = new CancellationTokenSource();
-                        this.updateRemoteConnections(cancellation.token)
-                            .catch(noop)
-                            .finally(() => cancellation.dispose());
-                    }
-
-                    // Send telemetry related to fetching the kernel connections
+                    // Send telemetry related to fetching the kernel connections. Do it here
```
IanMatthewHuff, line 237: This part feels a lot better with your change.

src/kernels/helpers.ts
rchiodo: In the test that was failing it would get to this point and the two kernelspecs would be equivalent (start from kernelspec and start from interpreter). There was nothing that would distinguish them.

rchiodo:
> Not sure if you are still in Draft. But seems pretty peachy to me. NotebookControllerManager feels a good bit cleaner as the big plus.

Sorry there is one small downside that I forgot to mention.

Previously we used to only validate the preferred kernel when searching through the cache. Now we validate the entire cache before returning the preferred kernel.

So imagine you have 20 kernels in your cache, We will now make sure the file for all 20 intepreters exists before returning the preferred one. We used to only check for file existence on the first one. 

I don't think a STAT call on 20 files is going to take very long, but it is a consequence of moving the caching around.

rchiodo:
I think I could make it so the cache is not validated during the preferred search though. It's all internal to the kernel finder anyway.

rchiodo:
I think the algorithm would have to be tweaked though. Like pass the validation into the findPreferredKernel function and instead of it returning the top of the list, it validates the top until it finds one that's valid.

rchiodo:
But it seems like premature optimization to me. 

I wrote some code to test that hypothesis:

```javascript
var glob = require("glob");
var fsExtra = require("fs-extra");

async function run() {
// options is optional
var files = await new Promise((resolve, reject) => {
        glob('./src/**/*.ts', (ex, res) => {
            if (ex) {
                reject(ex);
            } else {
                resolve(res);
            }
        });
    });
    let start = new Date().getTime();
    await Promise.all(files.map(async (f) => {
      await fsExtra.pathExists(f);
    }));
    let end = new Date().getTime();
    console.log(`Total time for ${files.length} files is ${end-start}`);
}

run();
```

That prints out this:

```
Total time for 771 files is 15
```
So 15 ms for 771 files. So even if the user has a ridiculous number of interpreters, it's still super fast to check for existence.

DonJayamanne (approved):

PR #9715, opened 2022-04-19
use our own app insights client to record test results

### LINKED ISSUE #9218: Enable queries on test run history
I'd like to be able to easily check the run history of a test or general trends in test runs/pass rate.

Some cases I can think of:
- I can easily check if a "flaky test" issue is still relevant. There are plenty of old ones that could probably just be closed.
- If a test fails in CI, I can go check if that test has already been failing before I made my change.
- When troubleshooting a flaky test, I can get a better idea of when it started failing to track down relevant changes

Notes:
- The CI system doesn't send any telemetry, so we currently can't just emit our normal telemetry events
- The "publish test results" step in the github action sounds promising, but is currently failing for auth reasons

amunger:
Fixes #9218

Pulled in a lot of the npm package from https://github.com/microsoft/vscode-extension-telemetry to be able to send telemetry during CI tests.
I did not pull in the web-based version, since I'm not sure where/how that would be used at this point.
I had to add a `sleep` call to let the reporter flush, even after awaiting the dispose which is supposed to do that. If I ran a single test without the `sleep` call, the event would never show up in Kusto.


src/test/utils/ciTelemetry/ciTelemetryReporter.node.ts
```diff
@@ -0,0 +1,32 @@
...
+import * as os from 'os';
+import { BaseTelemetryReporter } from './baseCiTelemetryReporter';
+import { TelemetryAppender } from './TelemetryAppender';
+import { appInsightsClientFactory } from './appInsightsClientFactory.node';
+
+/**
+ * An isolated wrapper for an Application Insights client that is specifically for sending telemetry during CI jobs.
+ * This won't run on a users machine, so there is no need to check for opt-in status.
+ */
+export class CiTelemetryReporter extends BaseTelemetryReporter {
```
rchiodo, line 14: Instead of pulling in copies of the code vscode-extension-telemetry, could you just extend the TelemetryReporter in the extension-telemetry npm module?

amunger: not the way it's currently written - getTelemetryLevel is a utility function which can't be overridden, and is used as a guard on multiple levels

rchiodo: Not sure I'm following. Sorry. 

getTelemetryLevel isn't overridden here? 

Did you change it in the BaseTelemetryReporter?

amunger: I removed getTelemetryLevel entirely since this will only collect telemetry in CI, but in the npm package, it's a global utility function that checks the vscode setting.

rchiodo: Could you monkey patch it in the npm code? Just remove it at run time?

I guess I'm not too sure about making a copy of the extension-telemetry code (with modifications).

rchiodo: Monkey patching in the test only code sounds safer to me than copying the code over.

lramos15:
This is a surprising amount of copied code. In theory since it's just perf objects you wouldn't need most of the code and could just call `trackEvent` on app insights directly. I don't think there's a need to copy over the nicely structured reporters or anything like that.

DonJayamanne:
> This is a surprising amount of copied code. In theory since it's just perf objects you wouldn't need most of the code and could just call trackEvent on app insights directly. I don't think there's a need to copy over the nicely structured reporters or anything like that.

I believe the solution (as discussed yesterday) is to enable telemetry on our CI server and just send the telemetry events specific to the tests on CI servers.

amunger:
> enable telemetry on our CI server

vscode internally disables any telemetry during testing, so we can't just set a configuration to enable it. Monkey patching the function that checks if it's enabled might work, but that also might send additional telemetry from vscode as well

lramos15:
I will look into a module level solution for this and report back with my findings.

rchiodo (approved):

amunger:
still waiting on Logan's thoughts, I just wanted to push up the work that I had done earlier

lramos15:
Just an update that we have added this native functionality to the module so it should be much easier to implement this and involve no code copying.

Issue #9629, opened 2022-04-07
Migrate off of the vscode.proposed.notebookDeprecated API


mjbvz:
From https://github.com/microsoft/vscode/issues/146831

Opening to track migrating off of VS Code's `notebook.deprecate` api proposal. This proposal is currently not on track for finalization

The suggested fix in this case is to move the output id into the custom metadata: https://github.com/microsoft/vscode/issues/146831#issuecomment-1090265265

DonJayamanne:
Proposal:
* Add transient metadata into outputs (upstream vscode changes)
* Add the output id into this transient property
* This output id will now be available on the client and the webview

sadasant:
@DonJayamanne @mjbvz 

Hello hello!

While I was investigating how to do this change, and thanks to @DonJayamanne 's guidance, I made it up to the following PRs:

- https://github.com/microsoft/vscode-jupyter/pull/10468 (though keep in mind that later we opted for naming this property `__id` to avoid the possibility of a future conflict with VS Code's public API).
- https://github.com/microsoft/vscode-notebook-renderers/pull/114
- https://github.com/microsoft/vscode/pull/152361

By the time I presented those PRs, I was somewhat aware that one missing piece was to ensure that the new `__id` property would not be saved, and that it needed to be ignored from the output changes, since otherwise the `.ipynb` would appear as dirty.

While asking for feedback and possible blindspots, @rebornix pointed out that VS Code did not currently have support for `transientCellOutputMetadata` in the `NotebookDocumentContentOptions`. Which means that, to follow the approach I was taking, I would need to ask VS Code to add support for something like `transientCellOutputMetadata`, trading one deprecated API for a new API change.

@rebornix and I reflected on the root cause of the issue and why `output.id` needed to be deprecated, as well as other possible alternatives.
- I didn't see why `output.id` needed to be deprecated, since it seems to be the most straightforward solution.
  - However, @rebornix pointed out that the Jupyter extension is the only use case we currently have for `output.id`.
- We discussed, for example, whether to use the output index rather than an Id.
- In any case, @rebornix quickly identified that **no matter the approach, we would need to always add something to the VS Code API to fulfill the vscode-jupyter needs**.

#### So, why do we currently need output.id in vscode-jupyter?

`vscode-jupyter` has UI interactions that fire events that later need to recover contextual information from the specific cell output that generated the UI. For example, a `Save Plot` button would need to retrieve the data included in the output. We can generalize this into a more generic statement:

[...]

PR #9721, opened 2022-04-20
Add widget tests

DonJayamanne:
Fixes #


.github/workflows/build-test.yml
```diff
@@ -666,15 +666,15 @@ jobs:
         uses: actions/upload-artifact@v2
         if: always()
         with:
-          name: Screenshots-${{matrix.os}}-${{matrix.pythonVersion}}-${{matrix.python}}-${{matrix.jupyter}}
+          name: Screenshots-${{matrix.jupyter}}-${{matrix.python}}-${{matrix.pythonVersion}}-${{matrix.os}}
```
DonJayamanne, line 669: Wanted the file names to line up with the test matrix name (easier to identify)

.gitignore
```diff
@@ -32,6 +32,8 @@ obj/**
 .pytest_cache
 temp/**
 tmp/**
+src/test/datascience/tmp/**
+src/test/datascience/temp/**
```
DonJayamanne, line 36: Moved notebooks created into a directory within the test workspace folder.
Python extension doesn't seem to use workspace active interpreter as active interpreter for files outside the workspace. Seems to be a recent change (our test open the workspace folder `src/test/datascience`).

src/notebooks/execution/cellExecution.node.ts
```diff
@@ -608,6 +624,16 @@ export class CellExecution implements IDisposable {
...
     ) {
+        if (
+            (isCI || this.context.extensionMode === ExtensionMode.Test) &&
+            process.env.IS_WIDGET_TEST &&
+            output.data &&
+            typeof output.data === 'object' &&
+            WIDGET_MIMETYPE in output.data
+        ) {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            (output.data[WIDGET_MIMETYPE] as any)['_vsc_test_cellIndex'] = this.cell.index;
```
DonJayamanne, line 635: Will discuss in standup, Will need this and or https://github.com/microsoft/vscode-jupyter/issues/9629
For now going with something to get the tests working (this only applies to CI scenario).

Basically we need a way to map outputs in webview to outputs of a cell in extension host.

src/test/datascience/widgets/commUtils.ts
```diff
@@ -41,26 +41,77 @@ export class Utils {
     public dispose() {
         disposeAllDisposables(this.disposables);
     }
-    public async queryHtml(selector: string, outputId: string) {
+    public async queryHtml(cell: NotebookCell, selector?: string) {
         // Verify the slider widget is created.
         const request = {
             requestId: Date.now().toString(),
-            id: outputId,
```
DonJayamanne, line 48: `outputId` is deprecated, and we need a way to hack around this. See notes earlier

src/notebooks/execution/cellExecution.node.ts
DonJayamanne: Was discussed in standup, going with temp solution until #9629 is implemented. Completing #9629 requires a bit more work and I'd rather do that in a separate PR (as that will require upstream changes as well)

src/test/common.node.ts
```diff
@@ -144,7 +144,6 @@ async function setPythonPathInWorkspace(
         console.log(`Updating Interpreter path to ${pythonPath} in workspace`);
         await settings.update('pythonPath', pythonPath, config).then(noop, noop);
         await settings.update('defaultInterpreterPath', pythonPath, config).then(noop, noop);
-        await settings.update('defaultInterpreterPath', pythonPath, config).then(noop, noop);
```
DonJayamanne, line 147: Duplicate

build/conda-nonconda-test-requirements.txt
```diff
@@ -13,3 +13,5 @@ bqplot
 K3D
 ipyleaflet
 jinja2==3.0.3 # https://github.com/microsoft/vscode-jupyter/issues/9468#issuecomment-1078468039
+matplotlib
```
DonJayamanne, line 16: Added tests for matplotlib, I believe this is a must as this could be one of the most popular widgets.

src/webviews/webview-side/ipywidgets/kernel/index.ts
```diff
@@ -125,6 +132,9 @@ function renderIPyWidget(
     }
     const output = document.createElement('div');
     output.className = 'cell-output cell-output';
+    if (typeof model._vsc_test_cellIndex === 'number') {
```
DonJayamanne, line 135: Since this attribute will only appear in tests, having this code in prodcution should be fine.
Note: This will be removed once we implement #9629

src/test/datascience/widgets/standard.vscode.test.ts
```diff
@@ -58,6 +50,8 @@ suite('Standard IPyWidget (Execution) (slow) (WIDGET_TEST)', function () {
         vscodeNotebook = api.serviceContainer.get<IVSCodeNotebook>(IVSCodeNotebook);
         kernelProvider = api.serviceContainer.get<IKernelProvider>(IKernelProvider);
         traceInfo('Suite Setup (completed)');
+        // eslint-disable-next-line local-rules/dont-use-process
+        process.env.IS_WIDGET_TEST = 'true';
```
DonJayamanne, line 54: This ensures we add the hacky property into the metadata into the cell outputs (only added in tests)

DonJayamanne: Don't want this env variable setup in CI for all tests, that would mean we have some garbage property in outputs that could stuff up the ipynb JSON saved & break other tests.
Again, this is temporary untill #9629

src/notebooks/helpers.ts
```diff
@@ -585,6 +585,16 @@ function convertOutputMimeToJupyterOutput(mime: string, value: Uint8Array) {
                     })
                 );
             }
+        } else if (mime.toLowerCase() === 'application/vnd.holoviews_load.v0+json') {
```
sadasant, line 588: could this mime type change at some point in the future? The `v0` tells me there might be a `v1` eventually, perhaps 🤔

src/notebooks/execution/cellExecution.node.ts
rchiodo: Why not just put this there all the time? Seems like it doesn't matter much if this is always there?

That way there won't be all of this extra checking here (which is going to have to change for web support)

rchiodo (approved):

src/notebooks/execution/cellExecution.node.ts
DonJayamanne: this will get stored in the json(ipynb), which isn't correct

src/notebooks/helpers.ts
DonJayamanne: good point

src/notebooks/execution/cellExecution.node.ts
rchiodo: Isn't correct but shouldn't hurt anything. Like who cares?

IanMatthewHuff (approved):

src/notebooks/execution/cellExecution.node.ts
DonJayamanne: Agreed, however I think since this is not absolutely necessary, we can leave this just for tests.
I.e. adding for production isn't going to buy us anything.

rchiodo: Except this will have to be refactored for web in order to determine if running under tests or not. More complexity when the added data doesn't hurt anything.

DonJayamanne: ill remove `process.env.IS_WIDGET_TEST` and see if the tests works, that way we don't need anything special, except the `context` object.
then you won't need any special cashing for web.
basically i don't want users to end up with a lit of unnecessary scm changes which they have complained about in the past.
agreed, let's not make it too complicated, hence will remove the process and the like and only use `context` object.

amunger (approved):

raghavthind2005 (approved):

PR #9729, opened 2022-04-20
Fix failures for remote

### LINKED ISSUE #9728: Dead  remote server prevents any kernels from showing up

rchiodo:
Fixes #9728 

The kernel refactor meant that if any kernel fetching failed, no kernels are returned.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2078,7 +2078,6 @@
     "enabledApiProposals": [
         "notebookControllerKind",
         "notebookDebugOptions",
-        "notebookDocumentEvents",
         "notebookDeprecated",
         "notebookEditor",
         "notebookEditorDecorationType",
--- a/src/kernels/kernelFinder.base.ts
+++ b/src/kernels/kernelFinder.base.ts
@@ -5,7 +5,7 @@ import { CancellationToken, Memento } from 'vscode';
 import { isPythonNotebook } from '../notebooks/helpers';
 import { IPythonExtensionChecker } from '../platform/api/types';
 import { createPromiseFromCancellation } from '../platform/common/cancellation';
-import { PYTHON_LANGUAGE, Telemetry } from '../platform/common/constants';
+import { PYTHON_LANGUAGE, Settings, Telemetry } from '../platform/common/constants';
 import { IConfigurationService, Resource } from '../platform/common/types';
 import { getResourceType } from '../platform/common/utils';
 import { noop } from '../platform/common/utils/misc';
@@ -25,6 +25,7 @@ import {
     deserializeKernelConnection,
     serializeKernelConnection
 } from './helpers';
+import { IJupyterServerUriStorage } from './jupyter/types';
 import { PreferredRemoteKernelIdProvider } from './raw/finder/preferredRemoteKernelIdProvider';
 import { ILocalKernelFinder, IRemoteKernelFinder } from './raw/types';
 import { IKernelFinder, INotebookProvider, INotebookProviderConnection, KernelConnectionMetadata } from './types';
@@ -46,7 +47,8 @@ export abstract class BaseKernelFinder implements IKernelFinder {
         private readonly notebookProvider: INotebookProvider,
         private readonly localKernelFinder: ILocalKernelFinder | undefined,
         private readonly remoteKernelFinder: IRemoteKernelFinder | undefined,
-        private readonly globalState: Memento
+        private readonly globalState: Memento,
+        protected readonly serverUriStorage: IJupyterServerUriStorage
     ) {}
 
     // Finding a kernel is the same no matter what the source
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -2078,7 +2078,6 @@
     "enabledApiProposals": [
         "notebookControllerKind",
         "notebookDebugOptions",
-        "notebookDocumentEvents",
```
rchiodo, line 2081: We don't need this anymore as it's published now. It was creating a compile error for me.

src/kernels/kernelFinder.base.ts
```diff
@@ -128,19 +130,22 @@ export abstract class BaseKernelFinder implements IKernelFinder {
...
+        // Get both local and remote kernels.
+        const [localKernels, remoteKernels] = await Promise.all([
+            this.listLocalKernels(resource, cancelToken, useCache).catch((ex) => {
+                traceError('Failed to get local kernels', ex);
+                return [];
+            }),
+            this.listRemoteKernels(resource, cancelToken, useCache).catch((ex) => {
+                traceError('Failed to get remote kernels', ex);
+                // When remote kernels fail, turn off remote
+                void this.serverUriStorage.setUri(Settings.JupyterServerLocalLaunch);
```
rchiodo, line 142: This could be changed in the future to just remove the URI for the failed kernels. Since we only support one still, I force local if any remote failures occur.

Might be better here to check for a ECONNREFUSED? The way this is right now I think if there's any failure it will remove the URI from remote.

rchiodo: But I figured Don was refactoring this anyway? Not sure if that's going to happen this month or not.

rchiodo: I changed my mind. Made it check for 'econn' in the message.

src/kernels/kernelFinder.base.ts
```diff
@@ -128,19 +130,24 @@ export abstract class BaseKernelFinder implements IKernelFinder {
...
-            ]);
+        // Get both local and remote kernels.
+        const [localKernels, remoteKernels] = await Promise.all([
+            this.listLocalKernels(resource, cancelToken, useCache).catch((ex) => {
+                traceError('Failed to get local kernels', ex);
+                return [];
+            }),
+            this.listRemoteKernels(resource, cancelToken, useCache).catch((ex) => {
+                traceError('Failed to get remote kernels', ex);
+                // When remote kernels fail, turn off remote if we get a ECONNREFUSED error
```
DonJayamanne, line 141: Not sure this is the right thing.
Shouldn't this be an action performed by the user, what if there's no net work connection.
I think removing somethingn the user added is best performed by the user. I.e. we can still try and fail, but removing should be left to the user.

amunger (approved):

src/test/datascience/kernel-launcher/kernelFinder.vscode.test.node.ts
```diff
@@ -118,11 +118,11 @@ suite('DataScience - Kernels Finder', () => {
         ) as LocalKernelConnectionMetadata;
         assert.ok(juliaKernelSpec);
 
-        const kernelSpec = await kernelFinder.findKernel(resourceToUse, {
+        const kernelSpec = (await kernelFinder.findKernel(resourceToUse, {
             kernelspec: juliaKernelSpec?.kernelSpec as any,
             orig_nbformat: 4
-        });
-        assert.ok(kernelSpec);
-        assert.deepEqual(kernelSpec, juliaKernelSpec);
```
DonJayamanne, line 126: Curious, why we're not comparing the kernel specs anymore.
Do you know what's different? Because previously we were looking for an exact match and now the match has been relaxed.

DonJayamanne: I'd prefer to have an exact match, after all, both should return the same object. Finding the kernel and listing should return the same thing

src/kernels/kernelFinder.base.ts
DonJayamanne: This feels like a change in behaviour today, if you're fixing a regressino that was introduced in the refactor thats fine, 
however I doubt we removed these items from the list in the past, & that feels like a change thats not necessary to fix this bug.

src/test/datascience/kernel-launcher/kernelFinder.vscode.test.node.ts
rchiodo: Before we only searched local kernels. Now we search both. So either one can come out.

src/kernels/kernelFinder.base.ts
rchiodo: Without this change if the remote fails, you can never start a kernel. Because it keeps picking the remote URI as the 'active' interpreter for the interactive window.

See this code here:
https://github.com/microsoft/vscode-jupyter/blob/a912533e456111d267f6d3838936e610387bbc03/src/notebooks/controllers/notebookControllerManager.ts#L183

URI isn't removed, it's just not used anymore. So if you bring up the picker again, it's in the MRU list

src/test/datascience/kernel-launcher/kernelFinder.vscode.test.node.ts
rchiodo: The test code above was wrong. It 'finds' the expected julia kernel just based on the language. The kernelFinder.findKernel is more specific. It's searching through all kernels to find one and when we're doing remote it finds the remote one first.

DonJayamanne (approved):

PR #9730, opened 2022-04-20
Correctly update notebook metatdata when changing between envs with the same python version number and fix saved display_name

### LINKED ISSUE #9727: Kernel metadata in file is not getting updated when executing with a new kernel
I started out with executing a ipynb with the local python env .debugEnv. After execution and save this is the metadata:

```
 "metadata": {
  "interpreter": {
   "hash": "2e5fd6c28aced2b333922ede1ea4949b12041d4fbe9c4e40c9e7ce883e7249f4"
  },
  "kernelspec": {
   "display_name": "Python 3.7.7 ('.debugEnv': venv)",
   "language": "python",
   "name": "python3"
  },
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 3
   },
```

I then selected the global Conda env functionalTestEnv and executed. I made sure to check sys.executable to make sure that I executed with the new kernel. After execution and save the meta data looked the exact same. Sill with a display name of .debugEnv, it was not updated to the new env.

IanMatthewHuff:
Fixes #9727


src/notebooks/helpers.ts
```diff
@@ -936,7 +936,16 @@ export function updateNotebookMetadata(
                 break;
         }
 
-        if (metadata.kernelspec?.name !== name) {
+        // Pull our old and new interpreter hash, if these don't match we also want to update metadata
+        // since name might be python3 in both scenarios and they might have the same python version so the
+        // check above with language info would not see them as changed.
```
IanMatthewHuff, line 941: https://github.com/microsoft/vscode-jupyter/blob/a912533e456111d267f6d3838936e610387bbc03/src/notebooks/helpers.ts#L898

You can see in this line linked above that "changed" would get set to true already if python language versions are different. If they were not different you could have a comparison of 'python3' to 'python3' for the name values when the hashes were different. This is why the issue showed best if you try to swap between two python envs with the same version number.

src/notebooks/helpers.ts
```diff
@@ -912,7 +912,7 @@ export function updateNotebookMetadata(
     if (kernelConnection?.kind === 'startUsingPythonInterpreter') {
         // Store interpreter name, we expect the kernel finder will find the corresponding interpreter based on this name.
         const kernelSpec = kernelConnection.kernelSpec;
-        let displayName = metadata.kernelspec?.display_name || kernelConnection.interpreter.displayName || '';
+        let displayName = kernelSpec.display_name || kernelConnection.interpreter.displayName || '';
```
IanMatthewHuff, line 915: A different semi-related bug. After making the initial fix the display_name was not getting updated correctly. Here it's pulling from the old saved display name instead of the new display name that we are switching to.

IanMatthewHuff:
@DonJayamanne I won't submit here without a test on this. I'll make sure to add one in this PR. But was a bit unsure where that test code would live best. I'd assume that we already have a suite checking metadata updates?

rchiodo (approved):

DonJayamanne (approved):
Would it be possible to have a test for this. at least test the function `updateNotebookMetadata`.
Its just that we seem to run into this similar issues pretty often and would feel better if we had a unit test for this function (i guess writing a fully integration test would be a lot more work).

IanMatthewHuff:
> Would it be possible to have a test for this. at least test the function `updateNotebookMetadata`. Its just that we seem to run into this similar issues pretty often and would feel better if we had a unit test for this function (i guess writing a fully integration test would be a lot more work).

@DonJayamanne Not sure if you saw my initial comment above. But I'm going to make sure to add a test here, but I was unsure about existing metadata change tests. Do we have some already? I was assuming that we could but I wasn't seeing them off hand.

DonJayamanne:
I don't think we have any existing tests (couldn't find it)

PR #9751, opened 2022-04-22
Use npm module 'is-local-host' to determine if a URL is a local host or not

### LINKED ISSUE #9748: Generalize how we detect localhost
Do we need something more general here? Like so for edge cases? https://www.npmjs.com/package/is-localhost-ip

_Originally posted by @IanMatthewHuff in https://github.com/microsoft/vscode-jupyter/pull/9745#discussion_r856365596_

rchiodo:
Fixes #9748


package.json
```diff
@@ -2170,6 +2170,7 @@
         "iconv-lite": "^0.4.21",
         "image-size": "^1.0.0",
         "inversify": "^5.0.1",
+        "is-localhost-ip": "^1.4.0",
```
IanMatthewHuff, line 2173: The npmjs entry actually said GPL, but from the actual repro it looks like it was licensed to MIT years ago. I'd assume that's ok? https://github.com/tinovyatkin/is-localhost-ip/commit/f9172384513ddca3245c460dbf66ef5234e8311c

IanMatthewHuff: Wondering if component governance uses npm entry...

rchiodo: I guess we'll find out :)

rchiodo:
Blech I can't find an npm module that doesn't have node dependencies. I'm just going to close this for now.

Issue #9167, opened 2022-03-01
Adjust error notification behavior when unable to connect to the remote jupyter server


DonJayamanne:
* Connect to remote Jupyter server `A`
* Kill the jupyter server
* Reload VS Code
* Open a notebook and run a cell against the local kernel. Everything seems fine.
* At some point you're interrupted with a modal dialog with the message `Failed to connect to the remote jupyter server. View Jupyter log for further details`

**Suggestions**:
* Don't asynchronously just show a modal dialog box. We need to find a better way to notify the user when this happens. I'd say they don't need to know until they try to choose a kernel.
* `Failed to fetch kernels from the remote jupyter server <name>. View Jupyter log for further details`
Where `name` is the Uri or the name provided for the connection (in the case of AML or the like).
* Should we also include a button `Disconnect from server`?

DonJayamanne:
TODO: Discuss at standup about using `Lazy Kernel`
Assume remote Uri is `http://xyz.com:1234/jupyter/uri/`
* Add an entry for lazy kernel named `Remote Jupyter Server (http://xyz.com:1234)`
	* If user selects this, then this will wait for Jupyter extension to load remote kernels for the above remote Uri
	* If not selected, & when jupyter extension has completed loading of the remote kernels, then remove this placeholder entry.
	* If kernels cannot be loaded, then do nothing, but as soon as user selects this lazy kernel, then display an error indicating the fact that the Remote Server could not be reached or the like.

**This way: User is notified about errors with the remote kernel only when they attempt to use it**

DonJayamanne:
**Proposed Solution:**
Display the outdated kernel in the kernel picker, upon selecting this, notify the user this connection is no longer valid & remove this kernel from the list (and all other kernels associated with this remote server).


**Technical details**
* Cache the remote kernels & list the cached remote kernels
* Upon loading VS Code, if the remote connection is down, then list cached kernels & when one of these controllers are selected, then notify the user and remove this outdated kernels (controllers)
* Always list cached kernels until the user removes this server from the list of controllers or selects this kernel from list of cached kernels
* Upon removing a remote Uri, the corresponding controllers will have to be removed as well.

DonJayamanne:
Solution: Display the kernels associated with the remote jupyter servers that are no longer valid. Once the user attempts to run a cell against one of these kernels, notify the user about the fact that the server is no longer recheable.

PR #9766, opened 2022-04-25
Remove unnecessary nullables

DonJayamanne:
Part of #9167

Basically we have a lot of unnecessary nullables, thereby simplifying the code
This PR finally fixes this comment
 
image

### DIFF
--- a/src/kernels/jupyter/launcher/jupyterExecution.ts
+++ b/src/kernels/jupyter/launcher/jupyterExecution.ts
@@ -101,7 +101,7 @@ export class JupyterExecutionBase implements IJupyterExecution {
     public connectToNotebookServer(
         options: INotebookServerOptions,
         cancelToken: CancellationToken
-    ): Promise<INotebookServer | undefined> {
+    ): Promise<INotebookServer> {
         // Return nothing if we cancel
         // eslint-disable-next-line
         return Cancellation.race(async () => {
@@ -152,7 +152,7 @@ export class JupyterExecutionBase implements IJupyterExecution {
                     } else if (connection) {
                         // If this is occurring during shutdown, don't worry about it.
                         if (this.disposed) {
-                            return undefined;
+                            throw err;
                         }
 
                         // Something else went wrong
@@ -182,6 +182,7 @@ export class JupyterExecutionBase implements IJupyterExecution {
                 }
                 throw lastTryError;
             }
+            throw new Error('Max number of attempts reached');
         }, cancelToken);
     }
 
--- a/src/kernels/jupyter/launcher/jupyterExecution.ts
+++ b/src/kernels/jupyter/launcher/jupyterExecution.ts
@@ -25,10 +25,10 @@ import {
     INotebookServerOptions,
     INotebookServer,
     JupyterServerUriHandle,
-    INotebookStarter
+    INotebookStarter,
+    INotebookServerFactory
 } from '../types';
 import { IJupyterSubCommandExecutionService } from '../types.node';
-import { IServiceContainer } from '../../../platform/ioc/types';
 
 const LocalHosts = ['localhost', '127.0.0.1', '::1'];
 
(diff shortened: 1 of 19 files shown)


src/kernels/jupyter/launcher/jupyterExecution.ts
```diff
@@ -182,6 +182,7 @@ export class JupyterExecutionBase implements IJupyterExecution {
                 }
                 throw lastTryError;
             }
+            throw new Error('Max number of attempts reached');
```
DonJayamanne, line 185: This function was the root cause, we have an explicit `return undefined` and an implicit at the end of the while loop.
That causes return value to have `undefined` and bubbles all the way the to the top.
This is an unlikely scenario (cancellation is handled higher up, hence we don't need to swallow cancellations and disposing classes here)

src/kernels/jupyter/launcher/liveshare/hostJupyterServer.ts
```diff
@@ -254,9 +254,9 @@ export class HostJupyterServer implements INotebookServer {
     }
 
     // Return a copy of the connection information that this server used to connect with
-    public getConnectionInfo(): IJupyterConnection | undefined {
+    public getConnectionInfo(): IJupyterConnection {
         if (!this.connection) {
-            return undefined;
+            throw new Error('Not connected');
```
DonJayamanne, line 259: THis is impossible, when the class is contructed, the first things that called is the `connect` method that initializes the property `this.connection`. Ideally we shoudl remove such code that initializes the class and then calls methods.
I've found this in a number of places and replaced them with factories, after that we dont end up with nullables with complicates the code.

DonJayamanne: Will change to a factory in another PR

src/kernels/jupyter/launcher/liveshare/hostJupyterServer.ts
```diff
@@ -34,27 +34,35 @@ import { JupyterNotebook } from '../jupyterNotebook';
...
     private disposed = false;
     constructor(
         @inject(IAsyncDisposableRegistry) private readonly asyncRegistry: IAsyncDisposableRegistry,
-        @inject(IJupyterSessionManagerFactory) private readonly sessionManagerFactory: IJupyterSessionManagerFactory,
         @inject(IWorkspaceService) private readonly workspaceService: IWorkspaceService,
         @inject(IOutputChannel) @named(STANDARD_OUTPUT_CHANNEL) private readonly jupyterOutputChannel: IOutputChannel,
-        @inject(IDisposableRegistry) private readonly disposables: IDisposableRegistry
+        @inject(IDisposableRegistry) private readonly disposables: IDisposableRegistry,
+        public connection: IJupyterConnection,
+        private sessionManager: JupyterSessionManager
```
DonJayamanne, line 53: The connection property is accessed using `getConnection` and we have some code that waits for this to be initialized.
Basically as soon as this class is contructed, the first method invoked is `connect` and its impossible for `connection` to be undefined, however due to the way the class was constructed it can be undefined in types & we have code all over the place that treats it as such.

Refactored this class to be created via a factory so that it can never be nullable, as a result the code paths are much simpler now.
No more nullables.

src/kernels/jupyter/launcher/liveshare/hostJupyterServer.ts
```diff
@@ -235,28 +208,22 @@ export class HostJupyterServer implements INotebookServer {
...
                 this.connection.dispose(); // This should kill the process that's running
-                this.connection = undefined;
             }
         } catch (e) {
             traceError(`Error during shutdown: `, e);
         }
     }
 
-    private waitForConnect(): Promise<IJupyterConnection | undefined> {
-        return this.connectPromise.promise;
```
DonJayamanne, line 253: Funky code now a thing of the past due to nullables being removed.

src/kernels/jupyter/launcher/liveshare/hostJupyterServer.ts
```diff
@@ -34,27 +34,35 @@ import { JupyterNotebook } from '../jupyterNotebook';
 import { noop } from '../../../../platform/common/utils/misc';
 import { Cancellation } from '../../../../platform/common/cancellation';
 import { getDisplayPath } from '../../../../platform/common/platform/fs-paths';
-import { INotebookServer, IJupyterSessionManagerFactory } from '../../types';
+import { INotebookServer } from '../../types';
 import { Uri } from 'vscode';
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
 @injectable()
```
rchiodo, line 41: You should probably remove this now then. Otherwise I think the tests will fail saying it's not in the container.

rchiodo (approved):

src/kernels/jupyter/launcher/jupyterExecution.ts
```diff
@@ -152,7 +151,7 @@ export class JupyterExecutionBase implements IJupyterExecution {
                     } else if (connection) {
                         // If this is occurring during shutdown, don't worry about it.
```
amunger, line 152: update comment?

PR #9776, opened 2022-04-25
Add server identifier to all remote connections

DonJayamanne:
Part of #9167

### DIFF
--- a/src/kernels/jupyter/remoteKernelFinder.ts
+++ b/src/kernels/jupyter/remoteKernelFinder.ts
@@ -22,6 +22,7 @@ import { IJupyterSessionManagerFactory, IJupyterSessionManager } from './types';
 import { sendKernelSpecTelemetry } from '../raw/finder/helper';
 import { traceError } from '../../platform/logging';
 import { IPythonExtensionChecker } from '../../platform/api/types';
+import { computeUriHash } from './jupyterUtils';
 
 // This class searches for a kernel that matches the given kernel name.
 // First it searches on a global persistent state, then on the installed python interpreters,
@@ -78,7 +79,8 @@ export class RemoteKernelFinder implements IRemoteKernelFinder {
                             interpreter: await this.getInterpreter(s, connInfo.baseUrl),
                             kernelSpec: s,
                             id: getKernelId(s, undefined, connInfo.baseUrl),
-                            baseUrl: connInfo.baseUrl
+                            baseUrl: connInfo.baseUrl,
+                            serverHash: computeUriHash(connInfo.url)
                         };
                         return kernel;
                     })
@@ -108,7 +110,8 @@ export class RemoteKernelFinder implements IRemoteKernelFinder {
                             model: s
                         },
                         baseUrl: connInfo.baseUrl,
-                        id: s.kernel?.id || ''
+                        id: s.kernel?.id || '',
+                        serverHash: computeUriHash(connInfo.url)
                     };
                     return kernel;
                 });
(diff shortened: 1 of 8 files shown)


src/kernels/jupyter/remoteKernelFinder.ts
```diff
@@ -77,7 +78,8 @@ export class RemoteKernelFinder implements IRemoteKernelFinder {
                             interpreter: await this.getInterpreter(s, connInfo.baseUrl),
                             kernelSpec: s,
                             id: getKernelId(s, undefined, connInfo.baseUrl),
-                            baseUrl: connInfo.baseUrl
+                            baseUrl: connInfo.baseUrl,
+                            serverHash: computeUriHash(connInfo.url)
```
DonJayamanne, line 82: Could call this `serverId` as well, after all its used to uniquely identify a jupyter server. However its a hash of the Url (which is unique).
Open to suggestions.

DonJayamanne: @rchiodo hash should be fine, checked with others in core.
To my knowledge hashes are secure enough to store sensitive information (even brute forcing will take a long time - very long time to crack).
Happy to discuss again if required.

rchiodo: I'm not a security expert. We have those at Microsoft, so you should check with them.

I did find this however:
https://tufin.medium.com/cracking-cryptographic-hashes-1e4b4721dbf3

rchiodo: I _think_ this would be the site to ask on:
https://microsoft.sharepoint.com/sites/MSProtect

DonJayamanne: Thanks, hate having to generate a GUID and store that, complicates the code a lot.

DonJayamanne: I'll move onto the guid approch (as discussed earlier today).

rchiodo: Probably take less time than a security review :)

DonJayamanne: Agreed, and it should be a better solution (if not easier)

connor4312: Hashing is fine for this use case, and hashing secrets is used pretty much universally, e.g. in MACs.

IanMatthewHuff (approved):

src/kernels/jupyter/remoteKernelFinder.ts
connor4312: That's fine. For example, HMAC involves hashing a message with a secret key.

rchiodo: In this case I think a rainbow table attack would work. 

Jupyter uses a known algorithm to generate tokens. The beginning part of the URL is already known. A rainbow table could be generated with every possible hash of that url and every possible token that jupyter could generate, then the user just looks up the hash.

There is no secret key in this case. There's no salt to change the hash.

I think in this case we should ask a security team expert or just change the algorithm to not store a hash of the url.

connor4312: Just going off https://jupyter-notebook.readthedocs.io/en/stable/security.html, a connection token looks like it's 12 bytes (hex encoded). A rainbow table for that would be on the order of 10^13 _petabytes_.

Rainbow tables are used when cracking passwords to ease in cracking users where multiple users might have the same password. When generating tokens, which is what it looks like the jupyter server is doing here, they aren't useful.

rchiodo: I think it's actually bigger than that. It's a guid, so 32 hex characters, or 16^32. I'm not even sure there's a word for that big. 

So maybe the hash is fine then.

raghavthind2005 (approved):

Issue #9792, opened 2022-04-26
Remote server connections are continually created during startup


rchiodo:
Not sure what's causing this yet, but this happens continually in the output log if using remote:

```
info 13:2:26.252: Disposing session manager
info 13:2:26.253: SessionManager - dispose contents manager
info 13:2:26.253: ShutdownSessionAndConnection - dispose session manager
info 13:2:26.254: Disposing session manager
info 13:2:26.254: SessionManager - dispose contents manager
info 13:2:26.255: ShutdownSessionAndConnection - dispose session manager
info 13:2:26.255: Disposing session manager
info 13:2:26.256: SessionManager - dispose contents manager
info 13:2:26.256: ShutdownSessionAndConnection - dispose session manager
info 13:2:26.257: Disposing session manager
info 13:2:26.258: SessionManager - dispose contents manager
info 13:2:26.258: ShutdownSessionAndConnection - dispose session manager
info 13:2:26.259: Disposing session manager
info 13:2:26.260: SessionManager - dispose contents manager
info 13:2:26.260: ShutdownSessionAndConnection - dispose session manager
info 13:2:26.261: Disposing session manager
info 13:2:26.262: SessionManager - dispose contents manager
info 13:2:26.262: ShutdownSessionAndConnection - dispose session manager
```

Seems to happen while we search for interpreters? Maybe it's continually refreshing the remote kernel list as interpreters come in?

rchiodo:
Maybe debounce the remote refresh?

DonJayamanne:
> but this happens continually in the output log if using remote:

Does it matter? I.e. is there a perf implication or the like that requires this to change?

rchiodo:
> > but this happens continually in the output log if using remote:
> 
> Does it matter? I.e. is there a perf implication or the like that requires this to change?

Not sure yet. It's not free at least.

greazer:
Per quick discussion in triage, decided this wasn't worth addressing.

Issue #7232, opened 2021-08-23
Reduce trace log levels to just `off|error|info|verbose`


DonJayamanne:
We have all of these log levels = 'off' | 'error' | 'warn' | 'info' | 'debug' | 'verbose';

Do we need `warn|info|debug`, why not just have `verbose`.

@roblourens @rchiodo @DavidKutu @IanMatthewHuff @joyceerhl /cc

**Questions:**
* Can we have just one instead of `info|debug|verbse`
* Do we even need `off`, why would we never display errors?

roblourens:
A couple messages are very noisy, like `Notebook Session status ...`, it would be nice to have a level where I can see the DA traffic but not other super verbose messages that don't seem that useful.

IanMatthewHuff:
Yeah, the levels feel funny to me. I feel like I could see maybe like three levels at maximum? Like a minimum error level, a pretty much everything level, then a level that adds in a few ultra verbose areas.

DonJayamanne:
> ve a level where I can see the DA traffic but not other super verbose messages that don't seem that useful.

But then if someone else wants to just see kernel messages and not DA, then we still end up in the same place (i.e. DA messages are noisy).

As we use log4js, we can achieve this by creating loggers for each area. This way we can turn on/off logging for each area.

I kind of like that, but we'd need to change the loggers from a function to a factory (just like log4js).


E.g. 
```javascript
const logger = createLogger('debugAdapter');
logger.verbose('Helo');
```

I believe this is possibel, after all we're using log4js and its a simple factory function.

**Summary:**
* Keep just `off, error, info, verbose`
	* In release mode this defaults to `info` (messages only logged in output channel)
	* In debug/development mode, this defaults to verbose (we can configure this to what we want. I prefer verbose)
* Create a factory function to return a logger for individual categories (debug adapter, widgets etc)
	* In release & debug mode, all these categories are on by default.
	* If we only want to see `debug adapter` messages, then we go into the settings and only include `debug adapter` category.
* To make this possible, we'd need to change some existing code to slowly move each message into a category, e.g. kernel messages should probably go into `kernel` or the like.
* Finally add a setting `logging.categories:  ['all', 'debug adapter', 'kernel', ...]`
	* This will not be defined in package.json (i don't want users to be able to see this, its only for our use). If we end up using this often enough, we can optionally add it into package.json and this way we get the necessary auto completion we need.
	* This will default to `all`

How doe this sound?

[...]

PR #9796, opened 2022-04-26
Reduce logging level when disposing

DonJayamanne:
Part of https://github.com/microsoft/vscode-jupyter/issues/9792
I too found the logs annoying.
This just ensures we do nothing if already disposed & reduce logging level (so that we log disposing only in verbose mode - don't see that being useful in default logging level).


rchiodo (approved):

rchiodo:
I'd love to have a category system for logging. Then I could turn on super verbose for one particular area. It'd have to be easy to specify though. Maybe we should just generate a bin file that we parse with something else.

DonJayamanne:
> Then I could turn on super verbose for one particular area

Was discussed here https://github.com/microsoft/vscode-jupyter/issues/7232#issuecomment-904136321
We discussed that and didn't feel it was worth the effort, but left the issue open.

Issue #9799, opened 2022-04-26
IHttpClient not implemented in web extension


rchiodo:
Required by CDN remote download

rchiodo:
This is working now.

rchiodo:
This is already being verified by integration tests that run on every PR.

PR #9832, opened 2022-04-28
Make IHttpClient not being there not break stuff

rchiodo:
For #9799 

Also had an error like so show up:

```
extension.web.bundle.js:4 
        
       Uncaught (in promise) TypeError: Right-hand side of 'instanceof' is not an object
    at eval (extension.web.bundle.js:4:1453946)
    at t.populateTelemetryWithErrorInfo (extension.web.bundle.js:4:1454022)
    at f (extension.web.bundle.js:4:1591407)
    at h (extension.web.bundle.js:4:1590568)
    at eval (extension.web.bundle.js:4:1592187)
```

So fixing that too. Don't need to use instanceof.

### DIFF
--- a/src/platform/errors/index.ts
+++ b/src/platform/errors/index.ts
@@ -6,7 +6,7 @@ import * as stackTrace from 'stack-trace';
 import { getTelemetrySafeHashedString } from '../../telemetry/helpers';
 import { getErrorTags } from './errors';
 import { getLastFrameFromPythonTraceback } from './errorUtils';
-import { getErrorCategory, TelemetryErrorProperties, WrappedError } from './types';
+import { getErrorCategory, TelemetryErrorProperties, BaseError, WrappedError } from './types';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function populateTelemetryWithErrorInfo(props: Partial<TelemetryErrorProperties>, error: Error) {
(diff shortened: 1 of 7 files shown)


IanMatthewHuff (approved):

src/platform/errors/index.ts
```diff
@@ -21,7 +21,8 @@ export function populateTelemetryWithErrorInfo(props: Partial<TelemetryErrorProp
         // Helps us determine that we are rejecting with errors in some places, in which case we aren't getting meaningful errors/data.
         props.failureSubCategory = 'errorisstring';
     }
-    const stdErr = (error instanceof BaseError ? error.stdErr : error.stack) || '';
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const stdErr = (error as any).stdErr ? (error as any).stdErr : error.stack || '';
```
DonJayamanne, line 25:
```suggestion
    const stdErr = (error as BaseError ).stdErr ? (error as BaseError ).stdErr : error.stack || '';
```

DonJayamanne: I think its better to cast it to an expected type, else tomorrow if we rename `stdErr` we don't benefit from types to let us know where we forgot.

DonJayamanne (approved):

PR #9833, opened 2022-04-28
Select active env as fallback, disable remote dbg

DonJayamanne:
* If we don't have a default remote controller, then fallback to the active interpreter instead of falling back to a random controller
* When debugging against remote kernel, notify user this isn't possible

This partially fixes the issue where you shutdown your remote jupyter server while VS Code is opened & launch IW.
`partially` because there are some more fixes to come

### DIFF
--- a/src/notebooks/controllers/notebookControllerManager.ts
+++ b/src/notebooks/controllers/notebookControllerManager.ts
@@ -200,7 +200,12 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             return this.createActiveInterpreterController(notebookType, resource);
         } else {
             traceInfoIfCI('CreateDefaultRemoteController');
-            return this.createDefaultRemoteController(notebookType);
+            const controller = await this.createDefaultRemoteController(notebookType);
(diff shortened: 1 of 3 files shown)


src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -423,7 +428,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             }
         });
 
-        return defaultPython3Kernel || defaultPythonKernel || defaultPythonLanguageKernel || controllers[0];
```
DonJayamanne, line 426: We were returning a random controller, that was causing even more issues 
Basically, why aren't we using the active interpreter like all other cases.

IanMatthewHuff: Good catch, I think I ran into this once recently, and it was confusing.

rchiodo (approved):

IanMatthewHuff (approved):

amunger (approved):

PR #9852, opened 2022-05-01
Expose 3rd party kernel API

DonJayamanne:
Expose kernel API to web extensions (for powertoys & 3rd party extensions)


src/kernels/helpers.ts
```diff
@@ -1012,7 +1012,7 @@ export function removeNotebookSuffixAddedByExtension(notebookPath: string) {
                 .substring(notebookPath.lastIndexOf(jvscIdentifier) + jvscIdentifier.length)
                 .search(guidRegEx) !== -1
         ) {
-            return `${notebookPath.substring(0, notebookPath.lastIndexOf(jvscIdentifier))}.ipynb`;
+            return notebookPath.substring(0, notebookPath.lastIndexOf(jvscIdentifier));
```
DonJayamanne, line 1015: We shouldn't be adding an `ipynb` suffix, as it already contains it.

src/platform/api/extension.d.ts
```diff
@@ -259,9 +259,11 @@ export interface IExportedKernelService {
      */
     getKernelSpecifications(refresh?: boolean): Promise<KernelConnectionMetadata[]>;
     /**
-     * Gets a list of all active kernel connections associated with a resource.
+     * Gets a list of all active kernel connections.
+     * If `uri` is undefined, then the kernel is not associated with any resource. I.e its currently not associated with any notebook in Jupyter extension.
+     * If `uri` is undefined, then the kernel is associated with the resource identified by the Uri.
      */
-    getActiveKernels(): Promise<{ metadata: KernelConnectionMetadata; uri: Uri }[]>;
+    getActiveKernels(): Promise<{ metadata: KernelConnectionMetadata; uri: Uri | undefined }[]>;
```
DonJayamanne, line 266: When we added the `Uri` into this as part of a previous PR we lost the ability to manage existing jupyter sessions. This brings it back.

src/platform/common/application/extensions.web.ts
```diff
@@ -28,6 +28,7 @@ export class Extensions implements IExtensions {
     public async determineExtensionFromCallStack(): Promise<{ extensionId: string; displayName: string }> {
         const stack = new Error().stack;
         if (stack) {
+            const jupyterExtRoot = this.getExtension(JVSC_EXTENSION_ID)!.extensionUri.toString();
```
rchiodo, line 31: Doesn't this need to be lower cased?

rchiodo (approved):

src/platform/common/application/extensions.web.ts
DonJayamanne: why is that?

rchiodo: Because you compare it to lower case values below.

DonJayamanne: Fixed

PR #9854, opened 2022-05-01
Add support for local live kernels

DonJayamanne:
Part of powertoys


src/intellisense/emptyNotebookCellLanguageService.ts
```diff
@@ -68,6 +68,7 @@ export class EmptyNotebookCellLanguageService implements IExtensionSingleActivat
                 break;
             }
             case 'startUsingRemoteKernelSpec':
+            case 'connectToLiveLocalKernel':
```
DonJayamanne, line 71: New type

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -268,6 +297,32 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         return this.controllersPromise;
     }
 
+    private async addLocalLiveKernelController(kernel: IKernel) {
+        // For now connecting to live local kernels is limited to Insiders or the pre-release channel.
+        if (this.appEnv.channel === 'insiders' || (await this.isPreRelease)) {
```
DonJayamanne, line 302: Feature available only in pre-release or vscode insiders

src/platform/common/utils/localize.ts
```diff
@@ -1035,6 +1035,10 @@ export namespace DataScience {
         'jupyter.kernel.category.jupyterRemoteKernel',
         '(Remote) Jupyter Kernel'
     );
+    export const kernelCategoryForLocalLiveJupyterKernel = localize(
+        'jupyter.kernel.category.jupyterLocalLiveKernel',
+        '(Local Live) Jupyter Kernel'
```
DonJayamanne, line 1040: New group in kernel list

src/kernels/helpers.ts
```diff
@@ -1574,3 +1576,29 @@ export function deserializeKernelConnection(kernelConnection: any): KernelConnec
...
+        };
+    } else if (kernel.kernelConnectionMetadata.kind === 'startUsingPythonInterpreter') {
+        return {
+            id: `liveLocal:${kernel.kernelConnectionMetadata.id}`,
+            kernelId: kernel.id,
+            kernelSpec: kernel.kernelConnectionMetadata.kernelSpec,
+            kernelConnectionMetadataKind: kernel.kernelConnectionMetadata.kind,
+            kind: 'connectToLiveLocalKernel',
+            interpreter: kernel.kernelConnectionMetadata.interpreter,
+            kernelModel: undefined
```
rchiodo, line 1601: Are live local kernels not the same as remote? Wondering why they don't have a kernel model.

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -313,6 +311,8 @@ export class VSCodeNotebookController implements Disposable {
                 void kernel.dispose();
             }
             this.associatedDocuments.delete(event.notebook);
+            // Possible we have a live local kernel associated with this.
+            // this.kernelProvider.kernels.forEach((item) => item.connectedResourceUris.delete(event.notebook.uri.toString()));
```
rchiodo, line 315: What's the connectedResourceURIs for? Looks like code you had for something that's not being used at the moment?

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -472,18 +472,31 @@ export class VSCodeNotebookController implements Disposable {
...
-            kernelStarted = true;
-            // If the controller changed, then ensure to create a new cell execution object.
-            if (kernel && kernel.controller.id !== controller.id) {
-                controller = kernel.controller;
-                exec = this.createCellExecutionIfNecessary(cell, kernel.controller);
-            }
-            currentContext = 'execution';
-            if (kernel.controller.id === this.id) {
+            // If we're dealing with live kernel connections, then get the kernel.
+            if (this.kernelConnection.kind === 'connectToLiveLocalKernel') {
```
rchiodo, line 476: Seems weird that this would be special cased here? Why doesn't it behave just like remote live kernels?

src/kernels/helpers.ts
DonJayamanne: No live local kernels are not the same. With live local kernels, we just communicate with the IKernel, there's no kernel model (which only exists when using Jupyter REST api to get information about active sessions).

I have thought about adding a kernelModel for live local kernels as well, as we could have the information such as # of connections, last activity and the like (which we use when dealing with remote kernels).
However that's soemthign that can always be added later on.

src/notebooks/controllers/vscodeNotebookController.ts
DonJayamanne: Yes, I wanted to keep track of the number of connections associated with a local live kernel. So we can popular the numberOfConnections property in the Kernel Model, but felt that was somethign we could add later with a little bit more thought into that implementation.

E.g. assume we connect a notebook to the same live kernel used by interative window, then ideally it would be good to know that there are 2 connections associated with that kernel. However that's not necessarily part of this PR and I felt i could add that later.
Will remove that commented block of code.

src/notebooks/controllers/vscodeNotebookController.ts
DonJayamanne: Basically we need a IKernel connection.
With remote kernels the controller has access to the sesion connection information, & in lower layers of the code we start a connection & execute against that.
With live kernels, the kernel connection information has the connection information as well, however it only needs the id of the kernel to be used. And when running code, all we need to do is fetch the real kernel & use that & that's whats happening here. This is the top most layer of the code & all we do here is fetch the kernel and rest of the code works as expected.

src/kernels/kernelProvider.node.ts
```diff
@@ -42,8 +43,18 @@ export class KernelProvider extends BaseKernelProvider {
     }
 
     public getOrCreate(uri: Uri, options: KernelOptions): IKernel {
-        const existingKernelInfo = this.getInternal(uri);
         const notebook = workspace.notebookDocuments.find((nb) => nb.uri.toString() === uri.toString());
+        // If we're connecting to a live local kernel, then Uri will point to the Uri of the original kernel.
+        if (options.metadata.kind === 'connectToLiveLocalKernel') {
```
rchiodo, line 48: I feel like we've pushed the ownership of handling 'live' kernels up to the kernelProvider and the controller. With remote this is all handled by the kernel finder. 

Seems like we could do it the same way, we just need something in the raw kernel stuff to keep track of running kernels.

rchiodo: Otherwise the controller and the kernelprovider have to special case live local kernels.

rchiodo: ILocalKernelFinder could return running kernels too. In fact jupyter puts them in a known spot. We just need a way to store them for raw.

src/kernels/kernel.base.ts
```diff
@@ -191,12 +190,12 @@ export abstract class BaseKernel implements IKernel {
     public get pendingCells(): readonly NotebookCell[] {
         return this.kernelExecution.queue;
     }
-    public async executeCell(cell: NotebookCell): Promise<NotebookCellRunState> {
+    public async executeCell(cell: NotebookCell, controller: NotebookController): Promise<NotebookCellRunState> {
```
rchiodo, line 193: Not 100% sure but I don't think I found where this is called without the 'this.controller'? Why did you need this extra parameter?

src/kernels/kernelProvider.node.ts
DonJayamanne: > remote this is all handled by the kernel finder.

Don't think kernel finder should own lifetime of kernels, the name itself indicates it is responsible for finding.
KernelProvider IMHO is the right place as that starts kernels and tracks them and dispose them as well.

> ILocalKernelFinder could return running kernels too

Yes, that's possible.

> In fact jupyter puts them in a known spot. We just need a way to store them for raw.

Yes, i thought about that as well, but that requires bit more work, to track the connection json and exposing that and then starting a new kernel with the kernel information provided and that messes up a lot of other things. E.g. we have 2 processes but connected to the same underlying kernel and could complicate how we handle intellisense, widgets, etc. This way we just use the same IKernel and don't really start new sesions.
E.g. even in remote case, we do'nt really start new sessions, we just connect to the exact same controller from multiple notebooks & I'd prefer to do the same here as well.

src/kernels/kernel.base.ts
DonJayamanne: In kernel we automatically assume the controller to be used is the same controller associated with the kernel.
E.g. asume we have a notebook A that starts kernel K1 associated with Controller C1.
Now we have a like kernel and we create conroller C2, and we don't really ever spin up a new kernel IKernel as we have one K1

Next, when run run a cell againts notebook B using controller C2, we need to use kernel K1, however kernel K1 will use its own controller C1 to udpate the cell outputs and the like. However thats wrong. We hould be using controller C2 to update contents/output of notebook B.

Hence, when calling executeCell in kernel K1, we need to pass in the controller.
This way a kernel can be use against multiple controllers

src/kernels/kernelProvider.node.ts
DonJayamanne: > ILocalKernelFinder could return running kernels too. In f

not going to do this, as thi will then need to know about live kernels & IKernel and the like.
I.e. this will need information about what s in KernelProvider.

Also mentioned earlier about issues with using that approach to start kernels. I think that complicates things a lot & could have more issues as well.

rchiodo: I disagree. 

For non zmq, it works just like remote. No difference.
For zmq, we simply need a KernelProcessManager. It returns the list of running processes. Basically mimic how jupyter does it.

rchiodo: In fact we could abstract out the session manager part to be ISessionManagerFactory and ISessionManager and the abstracted part only supports getting the list of running sessions.

rchiodo: IKernelProvider would not be pushed down (as it isn't for IRemoteKernelFinder). Instead the raw code would have to expose a way to get the list of processes it started. Much like the jupyter session manager does.

rchiodo: Actually it's more like the list of RawKernels/RawSessions (as we can't create another RawKernel to connect to the same process), but same idea as the JupyterSessions.

DonJayamanne: > For zmq, we simply need a KernelProcessManager. It returns the list of runnin

Ok, I think I understand what yoo're suggesting, I thought you wanted me to spin up a new process and use the same old `kernel-*.json` connection file (which also works, but more complex).

DonJayamanne: > as we can't create another RawKernel to connect to the same process

Actually we can, & that's what I thought you were suggesting all this time.
All we need to do is start a web socket connection for the same old kernel connection instead of starting a new process..

rchiodo: Hey if that works, probably even better (so that dispose stuff doesn't get confused).

PR #9858, opened 2022-05-01
Store custom metadata under vscode property in nb

### LINKED ISSUE #8789: Add metadata into kernelspec under `vscode-jupyter`  property
Today we store the interpreter information & other stuff in the kernelspec metadata.
Based on the docs here https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs
we should namespace these custom items.

DonJayamanne:
Fixes #8789


rchiodo (approved):

src/test/datascience/notebook/updateNotebookMetadata.unit.test.ts
```diff
@@ -133,18 +133,23 @@ suite('UpdateNotebookMetadata', () => {
             orig_nbformat: 4,
             kernelspec: { display_name: 'Python Default', language: 'python', name: 'python3' },
             language_info: { name: 'python', version: '3.6.0' },
-            interpreter: {
```
amunger, line 136: it would good to have a test for the old metadata too since that will be something we see a lot for a while. Should the metadata be updated? Or will there be stale information left behind?

DonJayamanne: Good point

PR #9866, opened 2022-05-02
Rework how we track remote kernels

### LINKED ISSUE #9826: State not persisted on close and reopen notebook
Testing #9768

Worked well until the last step, where I defined `a` in a cell, closed the notebook, reopened it, then tried to print `a` and got a `NameError`

Recording 2022-04-28 at 15 09 08

rchiodo:
Fixes #9826

Make only one way to store remote kernel ids and also fix problem where the 'notebook.selectKernel' doesn't work.


src/notebooks/controllers/liveKernelSwitcher.ts
```diff
@@ -23,56 +22,54 @@ const MEMENTO_BASE_KEY = 'jupyter-notebook-remote-session-';
...
-                if (match) {
-                    const key = this.getKey(n.uri);
-                    await this.memento.update(key, match.id);
+    private async switchKernel(n: NotebookDocument, kernel: Readonly<KernelConnectionMetadata>) {
+        traceInfo(`Using notebook.selectKernel to force remote kernel for ${n.uri} to ${kernel.id}`);
+        // Do this in a loop as it may fail
+        const success = await waitForCondition(
+            async () => {
+                if (this.vscNotebook.activeNotebookEditor?.document === n) {
+                    await this.commandManager.executeCommand('notebook.selectKernel', {
```
DonJayamanne, line 61: first time using this command in production.

src/kernels/raw/finder/preferredRemoteKernelIdProvider.ts
```diff
@@ -31,7 +32,7 @@ export class PreferredRemoteKernelIdProvider {
         const list: KernelIdListEntry[] = this.globalMemento.get<KernelIdListEntry[]>(ActiveKernelIdList, []);
         if (list) {
             // Not using a map as we're only going to store the last 40 items.
-            const fileHash = this.crypto.createHash(uri.toString(), 'string');
+            const fileHash = this.crypto.createHash(getComparisonKey(uri), 'string');
             const entry = list.find((l) => l.fileHash === fileHash);
```
DonJayamanne, line 36: Won't this cause issues in desktop now.
To ensure this works in desktop (previous versions) & new versions, I think the change should be to compare against the new hash & the old hash (using `uri.toString` & `getComparisonKey`)

DonJayamanne (changes requested):
I think we should maintain backwards compatibility with the old hashes.

src/notebooks/controllers/liveKernelSwitcher.ts
rchiodo: Not true. KernelSelector.ts has it already. I moved this here to be more explicit about the loop.

src/kernels/raw/finder/preferredRemoteKernelIdProvider.ts
rchiodo: Yeah the old hashes won't be there. But then they just have to pick a kernel again once. The problem with uri.toString() is that it's not case insensitive.

rchiodo: I put this back the way it was before. It was working before, so why switch it.

src/notebooks/controllers/liveKernelSwitcher.ts
DonJayamanne: I thought that was specific to Interactive Window with different arguments.
Doesn't matter anyways.

DonJayamanne: It was merely an observation.

DonJayamanne (approved):

PR #9873, opened 2022-05-03
Add test for switching active interpreter

### LINKED ISSUE #5478: Enable test for interactive window starting with active interpreter
This requires changes to the Python API. For example if we want to test that the interactive window picks up the current active interpreter if it's changed without needing to reload VS Code, we need to add the ability to invoke the Python extension's python.setInterpreter command from the Jupyter extension (i.e. make the command handler accept an optional URI or PythonEnvironment).

rchiodo:
Fixes #5478


DonJayamanne (approved):

src/test/datascience/helpers.node.ts
```diff
@@ -99,6 +102,15 @@ export async function insertIntoInputEditor(source: string) {
     return vscode.window.activeTextEditor;
 }
 
+export async function setActiveInterpreter(
+    apiProvider: IPythonApiProvider,
+    resource: Resource,
+    interpreter: vscode.Uri
+) {
+    const pythonApi = await apiProvider.getApi();
+    return pythonApi.setActiveInterpreter(getFilePath(interpreter), resource);
```
amunger, line 111: Make sure you're also clearing Python's workspace interpreter setting in the teardown, Otherwise the interpreter setting that we use will be ignored

rchiodo: Ah good point.

PR #9875, opened 2022-05-03
Cache kernel discovery by interpreters and few minitor fixes

DonJayamanne:


src/kernels/raw/finder/jupyterPaths.node.ts
```diff
@@ -126,7 +127,8 @@ export class JupyterPaths {
      * This list comes from the docs here:
      * https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs
      */
-    @traceDecoratorVerbose('Get Kernelspec root path')
+    @traceDecoratorVerbose('Get KernelSpec root path')
+    @cache(10_000)
     public async getKernelSpecRootPaths(cancelToken?: CancellationToken): Promise<Uri[]> {
         // Paths specified in JUPYTER_PATH are supposed to come first in searching
         const paths = new ResourceSet(await this.getJupyterPathPaths(cancelToken));
```
DonJayamanne, line 134: We can cache this, 10 minutes is good enough.
Why 10, simple, why not 10. I felt 1 was too small

src/kernels/raw/finder/localKernelSpecFinderBase.node.ts
```diff
@@ -188,31 +193,48 @@ export abstract class LocalKernelSpecFinderBase {
         paths: (Uri | { interpreter: PythonEnvironment; kernelSearchPath: Uri })[],
         cancelToken?: CancellationToken
     ): Promise<KernelSpecFileWithContainingInterpreter[]> {
-        const searchResults = await Promise.all(
-            paths.map(async (searchItem) => {
-                const searchPath = isUri(searchItem) ? searchItem : searchItem.kernelSearchPath;
-                if (await this.fs.localDirectoryExists(searchPath.fsPath)) {
-                    const files = await this.fs.searchLocal(`**/kernel.json`, searchPath.fsPath, true);
```
DonJayamanne, line 195: If you look at the logs on CI, we end up searching for kernels every single time in each interpreter.
& the number of interpreters seems to grow by 1 for each test.
This ensures we cache the promise for each `path` argument, & even if we have duplicates it will be handled as we generate unique keys and cache it by each `path` argument.

src/kernels/raw/finder/localPythonAndRelatedNonPythonKernelSpecFinder.node.ts
```diff
@@ -424,9 +424,9 @@ export class LocalPythonAndRelatedNonPythonKernelSpecFinder extends LocalKernelS
     ): Promise<IJupyterKernelSpec[]> {
         traceInfoIfCI(`Finding kernel specs for interpreters: ${interpreters.map((i) => i.uri).join('\n')}`);
         // Find all the possible places to look for this resource
-        const [interpreterPaths, rootSpecPaths, globalSpecRootPath] = await Promise.all([
-            this.findKernelPathsOfAllInterpreters(interpreters),
-            this.jupyterPaths.getKernelSpecRootPaths(),
+        const interpreterPaths = this.findKernelPathsOfAllInterpreters(interpreters);
```
DonJayamanne, line 427: `findKernelPathsOfAllInterpreters` was never returning a promise.

src/kernels/variables/pythonVariableRequester.node.ts
```diff
@@ -159,6 +159,9 @@ export class PythonVariablesRequester implements IKernelVariableRequester {
                   )
                 : [];
 
+            if (kernel.disposed || kernel.disposing) {
```
DonJayamanne, line 162: On CI found plenty of errors where variable requestor would throw errors.
I suspect thats because the kernel is dead,
adding code here to check if kernel is dead/dying, and if thats the case then just exit with an empty list instead of falling over & polluting the logs with unnecessary errors.

rchiodo (approved):

Issue #8933, opened 2022-02-04
Review architecture and design of Jupyter extension


greazer:
- [ ] Overall refactor of top-level folder structure [#8976]
- [ ] Overview of architecture and components
- [ ] Kernel enumeration, startup and execution
- [ ] IntelliSense
- [ ] Interactive Window and other core component interactions
- [ ] Core components at the same time as well, e.g., how the builtin interactive window is put together.

PR #9888, opened 2022-05-04
Simplify args used to start/connect to jupyter

DonJayamanne:
Part of #8933

### DIFF
--- a/src/kernels/jupyter/launcher/jupyterExecution.ts
+++ b/src/kernels/jupyter/launcher/jupyterExecution.ts
@@ -234,12 +234,12 @@ export class JupyterExecutionBase implements IJupyterExecution {
         if (options.localJupyter) {
             // If that works, then attempt to start the server
             traceInfo(`Launching server`);
-            const useDefaultConfig = !options || options.skipUsingDefaultConfig ? false : true;
             const settings = this.configuration.getSettings(options.resource);
-
+            const useDefaultConfig = this.configuration.getSettings(options.resource).useDefaultConfigForJupyter;
+            const workingDir = await this.workspace.computeWorkingDirectory(options.resource);
             // Expand the working directory. Create a dummy launching file in the root path (so we expand correctly)
             const workingDirectory = expandWorkingDir(
-                options.workingDir,
+                workingDir,
                 this.workspace.rootFolder ? urlPath.joinPath(this.workspace.rootFolder, `${uuid()}.txt`) : undefined,
                 this.workspace,
                 settings
--- a/src/kernels/jupyter/launcher/jupyterExecution.ts
+++ b/src/kernels/jupyter/launcher/jupyterExecution.ts
@@ -235,7 +235,7 @@ export class JupyterExecutionBase implements IJupyterExecution {
             // If that works, then attempt to start the server
             traceInfo(`Launching server`);
             const settings = this.configuration.getSettings(options.resource);
-            const useDefaultConfig = this.configuration.getSettings(options.resource).useDefaultConfigForJupyter;
+            const useDefaultConfig = settings.useDefaultConfigForJupyter;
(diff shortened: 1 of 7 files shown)


src/kernels/jupyter/launcher/jupyterExecution.ts
```diff
@@ -234,12 +234,12 @@ export class JupyterExecutionBase implements IJupyterExecution {
         if (options.localJupyter) {
             // If that works, then attempt to start the server
             traceInfo(`Launching server`);
-            const useDefaultConfig = !options || options.skipUsingDefaultConfig ? false : true;
```
DonJayamanne, line 237: This value is really useless as all we do is read the settings and pass it down.
But when its used at that same place we already access the settings.
Hence removed this so that we can access the settings where this is required (no additional dependencies were injected as part of this refacor)

src/kernels/jupyter/launcher/jupyterExecution.ts
```diff
@@ -234,12 +234,12 @@ export class JupyterExecutionBase implements IJupyterExecution {
...
             const settings = this.configuration.getSettings(options.resource);
-
+            const useDefaultConfig = this.configuration.getSettings(options.resource).useDefaultConfigForJupyter;
+            const workingDir = await this.workspace.computeWorkingDirectory(options.resource);
             // Expand the working directory. Create a dummy launching file in the root path (so we expand correctly)
             const workingDirectory = expandWorkingDir(
-                options.workingDir,
```
DonJayamanne, line 242: This value in the options property was totally useless, as we can see the next line we access the workspace object.
hence why not compute the workspace directory here instead of passing properties around.

This simplifies the type a lot, fewer properties.

This simplifies the type a lot

src/kernels/jupyter/launcher/jupyterNotebookProvider.ts
```diff
@@ -24,8 +24,7 @@ export class JupyterNotebookProvider implements IJupyterNotebookProvider {
     ) {}
 
     public async connect(options: ConnectNotebookProviderOptions): Promise<IJupyterConnection> {
-        const server = await this.serverProvider.getOrCreateServer(options);
-        const connection = await server.connection;
+        const { connection } = await this.serverProvider.getOrCreateServer(options);
```
DonJayamanne, line 27: Unnecessar `await`

src/kernels/jupyter/types.ts
```diff
@@ -81,8 +79,6 @@ export type INotebookServerOptions =
     | {
           uri: string;
           resource: Resource;
-          skipUsingDefaultConfig?: boolean;
-          workingDir?: string;
           ui: IDisplayOptions;
```
DonJayamanne, line 82: Now the options to start a notebook server is much smaller.

src/kernels/jupyter/launcher/jupyterExecution.ts
```diff
@@ -278,8 +269,11 @@ export class JupyterExecutionBase implements IJupyterExecution {
         customCommandLine: string[],
         workingDirectory: Uri,
         cancelToken: CancellationToken
-    ): Promise<IJupyterConnection | undefined> {
-        return this.notebookStarter?.start(
+    ): Promise<IJupyterConnection> {
+        if (!this.notebookStarter) {
```
DonJayamanne, line 273: This isn't possible in desktop. hence throw an error.

DonJayamanne: Doesn't need to be localized, its basically a bug in the code that would cause the entire extensoin to go kaboom.
Wish there was a better way to handle such types, these nullables really do complicate the code.

DonJayamanne: Hopefully this will get fixed with the refactoring I'm working on.

rchiodo (approved):

PR #9904, opened 2022-05-05
port python API typing fix

amunger:


DonJayamanne (approved):

rchiodo (approved):

src/platform/api/types.ts
```diff
@@ -97,7 +107,8 @@ export type PythonApi = {
     /**
      * IWindowsStoreInterpreter
      */
-    getSuggestions(resource: Resource): Promise<IInterpreterQuickPickItem[]>;
+    getSuggestions(resource: Resource): Promise<IInterpreterQuickPickItem_PythonApi[]>;
+    getKnownSuggestions(resource: Resource): IInterpreterQuickPickItem_PythonApi[];
```
DonJayamanne, line 111: Just an observation, this new API feels super flaky, its not returning a Promise.
What if Python extension hasn't completely interprter discovery, woudln't that mean this would not return the correct values, or nothing at all.

I think we're meant to await on their promise to wait on interpreter discovery to finish, we might want to do that to make this code more bullet proof. What do you think @rchiodo

rchiodo: I think it's okay for now, but we should investigate later. It looks to me like this api isn't called until we handle the 'select jupyter interpreter' command. Worst case scenario the list is missing stuff.

amunger: I'll merge for now, since it's an improvement over being broken and we can look into further changes

PR #9905, opened 2022-05-05
Reduce logging and retry widget test

### LINKED ISSUE #9901: IPySheets test seems to fail consistently on raw non conda 3.9
Get this error.

```
Render IPySheets:1 | Widget did not render or Hello, World, 42.000 not in html = <div class="cell-output-ipywidget-background"><div class="cell-output cell-output"></div></div>
```

Haven't looked further. Maybe ipysheets doesn't work without jupyter on 3.9.

DonJayamanne:
Fixes #9901

### DIFF
--- a/src/kernels/kernelProvider.base.ts
+++ b/src/kernels/kernelProvider.base.ts
@@ -112,12 +112,11 @@ export abstract class BaseKernelProvider implements IKernelProvider {
     protected disposeOldKernel(uri: Uri) {
         const notebook = workspace.notebookDocuments.find((nb) => nb.uri.toString() === uri.toString());
         if (notebook) {
-            traceInfoIfCI(
-                `Disposing kernel associated with ${getDisplayPath(notebook.uri)}, isClosed=${notebook.isClosed}`
-            );
-
             const kernelToDispose = this.kernelsByNotebook.get(notebook);
             if (kernelToDispose) {
+                traceInfoIfCI(
+                    `Disposing kernel associated with ${getDisplayPath(notebook.uri)}, isClosed=${notebook.isClosed}`
+                );
                 this.pendingDisposables.add(kernelToDispose.kernel);
                 kernelToDispose.kernel
                     .dispose()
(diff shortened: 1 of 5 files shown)


src/test/datascience/widgets/standard.vscode.test.ts
```diff
@@ -41,6 +41,8 @@ suite('Standard IPyWidget (Execution) (slow) (WIDGET_TEST)', function () {
     this.timeout(120_000);
     let previousWidgetScriptSourcesSettingValue: string[] | undefined = undefined;
     const widgetScriptSourcesValue = ['jsdelivr.com', 'unpkg.com'];
+    // Retry at least once, because ipywidgets can be flaky (network, comms, etc).
+    this.retries(1);
```
DonJayamanne, line 45: Based on the logs there's nothing wrong with the widgets, i think its just flaky because we're downloading widgets from a CDN and perhaps a few other issues cause delays in rendering.
Increasing the retries for this test to see if this fixes the ipysheet flakiness.

src/kernels/kernelProvider.base.ts
```diff
@@ -112,12 +112,11 @@ export abstract class BaseKernelProvider implements IKernelProvider {
...
         if (notebook) {
-            traceInfoIfCI(
-                `Disposing kernel associated with ${getDisplayPath(notebook.uri)}, isClosed=${notebook.isClosed}`
-            );
-
             const kernelToDispose = this.kernelsByNotebook.get(notebook);
             if (kernelToDispose) {
+                traceInfoIfCI(
+                    `Disposing kernel associated with ${getDisplayPath(notebook.uri)}, isClosed=${notebook.isClosed}`
+                );
```
DonJayamanne, line 119: We were getting dispose messages even when things weren't getting disposed.
The logs were too noisy and mis leading.

src/test/datascience/widgets/standard.vscode.test.ts
IanMatthewHuff: If it's just a bit of unavoidable flake, then maybe add a couple of retries (2 or 3 instead of 1) so that we really stack the odds of success in our favor? Or would that possibly blow up the time to much if the tests start failing for a legitimate reason?

IanMatthewHuff (approved):
Maybe up the retry number? Otherwise we are just flakey less often.

PR #9908, opened 2022-05-05
Fix up restart session (remove dead code)

DonJayamanne:
We no longer start restart sessions everytime, in the past we'd start a session for use as the restart session.
Now that we don't have that functionality we don't need to hide those kernels from the kernel filter & the like.

### DIFF
--- a/src/kernels/common/baseJupyterSession.ts
+++ b/src/kernels/common/baseJupyterSession.ts
@@ -118,7 +118,6 @@ export abstract class BaseJupyterSession implements IJupyterSession {
         public readonly kind: 'localRaw' | 'remoteJupyter' | 'localJupyter',
         protected resource: Resource,
         protected readonly kernelConnectionMetadata: KernelConnectionMetadata,
-        private restartSessionUsed: (id: Kernel.IKernelConnection) => void,
         public workingDirectory: Uri,
         private readonly interruptTimeout: number
     ) {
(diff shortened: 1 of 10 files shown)


src/kernels/common/baseJupyterSession.ts
```diff
@@ -175,44 +174,33 @@ export abstract class BaseJupyterSession implements IJupyterSession {
             return;
         }
 
-        // Start the restart session now in case it wasn't started
-        if (!this.restartSessionPromise) {
-            this.startRestartSession(false);
```
DonJayamanne, line 180: The code is better now, previously we'd call `startRestartSesion` and expect `restartSessionPromise` to be initialied and we'd await on that.
Now we just call the method `startRestartSession` and await on the result.

src/kernels/common/baseJupyterSession.ts
```diff
@@ -175,44 +174,33 @@ export abstract class BaseJupyterSession implements IJupyterSession {
...
-            }
-            this.restartSessionPromise.token.dispose();
-            this.restartSessionPromise = undefined;
-            traceInfo('Started new restart session');
-            if (oldStatusHandler && oldSession) {
-                oldSession.statusChanged.disconnect(oldStatusHandler);
-            }
-            this.shutdownSession(oldSession, undefined, false).ignoreErrors();
-        } else {
-            throw new SessionDisposedError();
```
DonJayamanne, line 214: Its impossiblel for this to happen, but because of the types being nullable, we write such code.
Basiscally we initialize a class and then call a method to set it in the right state, a factory class will get rid of this.
We already have a factory class, hopefully it will be easy to move some of the stuff into there.

src/kernels/jupyter/session/jupyterSession.ts
```diff
@@ -180,12 +174,19 @@ export class JupyterSession extends BaseJupyterSession {
     }
 
     protected startRestartSession(disableUI: boolean) {
-        if (!this.restartSessionPromise && this.session && this.contentsManager) {
-            const token = new CancellationTokenSource();
-            const promise = this.createRestartSession(disableUI, this.session, token.token);
-            this.restartSessionPromise = { token, promise };
-            promise.finally(() => token.dispose());
+        if (!this.session) {
+            throw new Error('Session disposed or not initialized');
```
DonJayamanne, line 178: its a shame we still need this. because `this.session` can be null, hopefully it will go away in a subsequent refactor

rchiodo (approved):

PR #9916, opened 2022-05-06
Remove unnecessary caching

DonJayamanne:
The use of cache decorator was unnecessary and was breaking local dev.
Because the return value was a uri and that cannot be JSON serialized. 

Unnecessary as everything used in this method is already cached, hence the caching on top was unnecessary.


DonJayamanne:
Will merge once CI goes green even if not apporved, as this is breaking local dev.

PR #9931, opened 2022-05-06
Add test for https servers

### LINKED ISSUE #9844: Add back selfCerts certificate test
https://github.com/microsoft/vscode-jupyter/issues/9758

Per the issue above. We used to have a test for this scenario, but it was one of our old functional test. It's currently commented out, and has not been replaced. Since we've regressed this we should get it back in.

Old test can be seen if you search for --certfile in our code base. Old self signed cert (jcert.pem, jkey.key) are both still in the code base.

rchiodo:
Fixes #9844


src/kernels/raw/finder/localPythonAndRelatedNonPythonKernelSpecFinder.node.ts
```diff
@@ -76,16 +76,11 @@ export class LocalPythonAndRelatedNonPythonKernelSpecFinder extends LocalKernelS
         resource: Resource,
         cancelToken?: CancellationToken
     ): Promise<(LocalKernelSpecConnectionMetadata | PythonKernelConnectionMetadata)[]> {
-        const activeInterpreterPromise = this.extensionChecker.isPythonExtensionInstalled
-            ? this.interpreterService.getActiveInterpreter(resource)
-            : undefined;
         const interpreters = this.extensionChecker.isPythonExtensionInstalled
             ? await this.interpreterService.getInterpreters(resource)
             : [];
-        const activeInterpreter = activeInterpreterPromise ? await activeInterpreterPromise : undefined;
```
rchiodo, line 85: I noticed this was redundant. getInterpreters already includes the active interpreter.

src/kernels/jupyter/jupyterUtils.ts
```diff
@@ -56,7 +56,30 @@ export async function handleCertsError(
     );
     if (value === enableOption) {
         sendTelemetryEvent(Telemetry.SelfCertsMessageEnabled);
-        void config.updateSetting('allowUnauthorizedRemoteConnection', true, undefined, ConfigurationTarget.Workspace);
+        await config.updateSetting('allowUnauthorizedRemoteConnection', true, undefined, ConfigurationTarget.Workspace);
```
rchiodo, line 59: This missing await was what was causing all of my test failures on CI (but not on my local machine). The test was running so fast that this setting wouldn't update by the time we tried to connect and the connection would randomly fail in other spots.

src/kernels/jupyter/jupyterUtils.ts
```diff
@@ -56,7 +56,30 @@ export async function handleCertsError(
...
+        return true;
+    } else if (value === closeOption) {
+        sendTelemetryEvent(Telemetry.SelfCertsMessageClose);
+    }
+    return false;
+}
+
+export async function handleExpiredCertsError(
```
rchiodo, line 67: The certs we have in the repo are expired. Figured expired certificate is still a valid test case and a user might do this too.

src/kernels/raw/finder/localPythonAndRelatedNonPythonKernelSpecFinder.node.ts
DonJayamanne: yes, was planning on doing that. python extension added that feature (agreed i filled a bug) recently in the very latest release.
we added it to ensure things work as expected.

DonJayamanne (approved):

amunger (approved):

PR #9932, opened 2022-05-06
added pre-commit format check

### LINKED ISSUE #9702: Add a commit check to prevent committing unformatted documents
vscode will block a `git commit` if any document hasn't been formatted with prettier, but this repo will let you get all the way to a PR before realizing that a document needs to be formatted. We should add that pre-commit check to the jupyter repo.

amunger:
Fixes #9702

This will only run if you run npm install/ci since husky needs to install for the hook to kick in, but I included a check for prettier as well anyway.
An alternative option is to add the file .git/hooks/pre-commit, but that wouldn't be stored in the repo

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,17 @@
+#!/bin/sh
+if npm list prettier | grep -q prettier@ ; then
+    echo husky - prettier is installed
+else
+    echo husky - prettier is not installed, skipping format check
+    exit 0
+fi
+
+# run prettier on all .ts files that are in the commit
+changed=$(git diff --cached --name-only | grep .ts$ | tr '\n' ' ')
+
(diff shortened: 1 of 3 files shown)


.husky/pre-commit
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+if npm list prettier | grep -q prettier@ ; then
```
rchiodo, line 2: Is this going to run on windows?

rchiodo: It looks like a shell script? Does husky somehow run shell scripts on windows?

amunger: I'm running it on windows, the vanilla pre-commit script through git is able to run the same way, so it's probably using git bash

rchiodo (approved):

Issue #8043, opened 2021-10-26
Dead remote Jupyter server detection takes too long


joaomoreno:
Testing #8024

1. Follow the remote Jupyter server steps
2. Open a notebook
3. Kill `jupyter notebook`
4. Run a cell

🐛 Code should detect the Jupyter server has died... but instead the cell is left to run forever:

![recording (14)](https://user-images.githubusercontent.com/22350/138883157-593d2d43-7c9d-42f5-b830-a56f2d7ca36a.gif)

DonJayamanne:
Should be fixed, closing for verification.

PR #9941, opened 2022-05-08
Verify remote Jupyter Sever connections upon connecting to remote kernels

### LINKED ISSUE #9167: Adjust error notification behavior when unable to connect to the remote jupyter server

DonJayamanne:
Fixes #9167


src/kernels/jupyter/launcher/jupyterExecution.ts
```diff
@@ -210,9 +205,6 @@ export class JupyterExecutionBase implements IJupyterExecution {
                 cancelToken
             );
         } else {
-            // Prepare our map of server URIs
-            await this.jupyterConnection.updateServerUri(options.uri);
-
             // If we have a URI spec up a connection info for it
             return this.jupyterConnection.createConnectionInfo(options.uri);
```
DonJayamanne, line 209: `cerateConnectionInfo` will ensure the server Uri has been updated, calling code need to know about such internals

src/kernels/jupyter/launcher/liveshare/hostJupyterServer.ts
```diff
@@ -150,6 +151,22 @@ export class HostJupyterServer implements INotebookServer {
...
             throw new SessionDisposedError();
         }
+        if (
+            this.sessionManager &&
+            !this.isDisposed &&
+            (kernelConnection.kind === 'connectToLiveRemoteKernel' ||
+                kernelConnection.kind === 'startUsingRemoteKernelSpec')
+        ) {
+            try {
+                await Promise.all([this.sessionManager.getRunningKernels(), this.sessionManager.getKernelSpecs()]);
```
DonJayamanne, line 161: Ensure we try to fetch the kernels (validate the remote jupyter server connection) before creating a session.
This should also fix https://github.com/microsoft/vscode-jupyter/issues/8043

DonJayamanne: As well as the current issue #9167

src/kernels/jupyter/remoteKernelFinder.ts
```diff
@@ -113,6 +112,7 @@ export class RemoteKernelFinder implements IRemoteKernelFinder {
                 return items;
             } catch (ex) {
                 traceError(`Error fetching remote kernels:`, ex);
+                throw ex;
```
DonJayamanne, line 115: Previously we'd swallow exceptions and return an empty list, now we throw the error and higher up when we have errors we just fallback to the cache (this behaviour is documented in the calling code)

src/kernels/jupyter/serverSelector.ts
```diff
@@ -212,22 +217,39 @@ export class JupyterServerSelector {
...
             return DataScience.jupyterSelectURIInvalidURI();
         }
+
+        // Double check http
+        if (!inputText.toLowerCase().startsWith('http')) {
+            return 'Has to be http(s)';
+        }
+        // Double check this server can be connected to. Might need a password, might need a allowUnauthorized
+        try {
+            await this.jupyterConnection.validateRemoteUri(inputText);
```
DonJayamanne, line 238: We now validate the Uri when its entered, previously its not validated, instead we validate it when we save it into the uri storage and error message is displayed on the bottom right (very disconnected ux)

src/kernels/kernelFinder.base.ts
```diff
@@ -172,35 +177,56 @@ export abstract class BaseKernelFinder implements IKernelFinder {
...
         if (useCache === 'ignoreCache') {
-            kernels = await kernelsWithoutCachePromise;
+            try {
+                kernels = await kernelsWithoutCachePromise;
+            } catch (ex) {
+                // Since fetching the remote kernels failed, we fall back to the cache,
+                // at this point no need to display all of the kernel specs,
+                // Its possible the connection is dead, just display the live kernels we had.
+                // I.e. if user had a notebook connected to a remote kernel, then just display that live kernel.
+                kernels = await this.getFromCache(kind, cancelToken);
```
DonJayamanne, line 191: This is what makes #9167 possible, display the old dead kernels and notify the user and let the user remove it from the list of kernels.

src/kernels/kernelFinder.node.ts
```diff
@@ -42,12 +42,6 @@ export class KernelFinder extends BaseKernelFinder {
 
     protected async isValidCachedKernel(kernel: KernelConnectionMetadata): Promise<boolean> {
         switch (kernel.kind) {
-            case 'startUsingRemoteKernelSpec':
-            case 'connectToLiveRemoteKernel':
-                // If this is a a remote kernel, it's valid if the URI is still active
-                const uri = await this.serverUriStorage.getRemoteUri();
```
DonJayamanne, line 48: Not entirely sure why this was done, i guess its a temporary solution for #9167, but this breaks the code for AML (as AML uris don't work like this) also its very slow when loading the cache (making the cache completely useless).

This also slows down loading of local kernel specs as we load local & remote kernels from cache together, hence nohting appears until we validate the remote kernels, which feels like a poor design right now, but we can look into that in debt week.

src/kernels/raw/finder/preferredRemoteKernelIdProvider.ts
```diff
@@ -38,7 +38,15 @@ export class PreferredRemoteKernelIdProvider {
         }
     }
 
-    public async storePreferredRemoteKernelId(uri: Uri, id: string | undefined): Promise<void> {
+    public async clearPreferredRemoteKernelId(uri: Uri): Promise<void> {
+        await this.storePreferredRemoteKernelIdInternal(uri);
```
DonJayamanne, line 42: Bug fix, added tests:
* Assume you open a notebook and select a remote kernel
* Now run a cell
* Open the same notebook tomorrow, it will connect to the same remote kernel (good)
* Shutdown the remote server
* Open the notebook, you'll get an error about the server not being available, good
* Now switch the kernel to a local kernel
* Close vsc & open the same notebook again & it will try to connect to the same remote server.

We weren't clearing the mapping.

src/kernels/raw/finder/preferredRemoteKernelIdProvider.ts
```diff
@@ -47,20 +55,25 @@ export class PreferredRemoteKernelIdProvider {
...
         // Prune list if too big
         sendTelemetryEvent(Telemetry.NumberOfSavedRemoteKernelIds, undefined, { count: list.length });
         while (list.length > MaximumKernelIdListSize) {
+            requiresUpdate = true;
             list.shift();
         }
         traceInfo(`Storing Preferred remote kernel for ${getDisplayPath(uri)} is ${id}`);
-        await this.globalMemento.update(ActiveKernelIdList, list);
+        if (requiresUpdate) {
+            await this.globalMemento.update(ActiveKernelIdList, list);
```
DonJayamanne, line 76: We were always updating even when nothing had to be updated.

src/platform/errors/errorHandler.ts
```diff
@@ -174,6 +180,42 @@ export class DataScienceErrorHandler implements IDataScienceErrorHandler {
...
+        } else if (err instanceof RemoteJupyterServerConnectionError) {
+            const selection = await this.applicationShell.showErrorMessage(
+                DataScience.remoteJupyterConnectionFailedWithServer().format(err.baseUrl),
+                { detail: err.originalError.message || '', modal: true },
+                DataScience.removeRemoteJupyterConnectionButtonText(),
+                DataScience.changeRemoteJupyterConnectionButtonText(),
+                DataScience.selectDifferentKernel()
+            );
+            const serverId = err.serverId;
+            switch (selection) {
```
DonJayamanne, line 192: This lets users decide what needs to be done when the remote server is no longer valid.
Fixes #9167

src/test/datascience/kernel-launcher/remoteKernelFinder.unit.test.ts
```diff
@@ -262,28 +259,4 @@ suite(`Remote Kernel Finder`, () => {
...
     });
-    test('Invalid kernels not returned', async () => {
-        const validKernel: RemoteKernelSpecConnectionMetadata = {
-            kernelSpec: python3spec,
-            baseUrl: connInfo.baseUrl,
-            kind: 'startUsingRemoteKernelSpec',
-            id: '2',
-            serverId: computeUriHash(connInfo.url)
-        };
-        const invalidKernel: RemoteKernelSpecConnectionMetadata = {
```
DonJayamanne, line 273: This isn't a valid test, we cannot end up with 2 kernels from two different remote jupyter uris in the same cache entry. Today we only ever support one at a time.

Also removed verification of remote jupyter uris, as we want failures so user can remove them.

src/kernels/kernelFinder.node.ts
rchiodo: Without this the cache can be completely invalid. I suppose you're handling this by always returning the cache and letting the user pick the invalid items, but then throwing an error then? Basically move this validation to later in the process.

DonJayamanne: Yes, thats correct.

rchiodo: And not sure why the original code would have been slow. It's not actually going to the remote host, it's just querying the encrypted storage.

src/kernels/raw/finder/preferredRemoteKernelIdProvider.ts
```diff
@@ -38,7 +38,15 @@ export class PreferredRemoteKernelIdProvider {
...
     }
 
-    public async storePreferredRemoteKernelId(uri: Uri, id: string | undefined): Promise<void> {
+    public async clearPreferredRemoteKernelId(uri: Uri): Promise<void> {
+        await this.storePreferredRemoteKernelIdInternal(uri);
+    }
+    public async storePreferredRemoteKernelId(uri: Uri, id: string): Promise<void> {
+        await this.storePreferredRemoteKernelIdInternal(uri, id);
+    }
+    public async storePreferredRemoteKernelIdInternal(uri: Uri, id?: string): Promise<void> {
```
amunger, line 47: I think updatePreferredRemoteKernel would be more clear

src/test/datascience/notebook/remoteNotebookEditor.vscode.common.ts
```diff
@@ -215,3 +163,69 @@ export function sharedRemoteNotebookEditorTests(
...
+    await waitForCondition(
+        async () => !!remoteKernelIdProvider.getPreferredRemoteKernelId(nbEditor.document.uri),
+        5_000,
+        'Remote Kernel id not saved'
+    );
+
+    if (doNotSaveAndCloseNotebook) {
+        return;
+    }
+
```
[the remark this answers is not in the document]
amunger: actually should just be two separate methods (one could call the other) since the part before this does not do what the name of the method says it does.

src/kernels/kernelFinder.node.ts
DonJayamanne: Enable azure AML compute and the cache ends up setting up the azure compute connection, thats very slow and now none of the locally cached kernels load either for a long time.

DonJayamanne: Also with azure AML compute this logic was incorrect, as the uri is not the remote jupyter uri, its something else.

Issue #9684, opened 2022-04-14
Support multiple remote servers


rebornix:
As part of the lazy kernel exploration https://github.com/microsoft/vscode/issues/146942#issuecomment-1097405365 , we want to support multiple remote servers, which would allow each notebook to use their preferred remote server, contributed by remote jupyter server provider.

DonJayamanne:
While refactoring the kernel code I explored this task and at that point the plan was to add the Uri (some of remote connection info) into the kernel connection & when starting/connecting to the remote server we'd use that information instead of assuming the connections was in the single setting. 

This was discussed with the team, however not completed. Can't remember why it wasn't completed, either ran out of time or just lost track of it. One use case was the ability to connect to multiple remotes and display in the kernel manager.

DonJayamanne:
Will not be done as we don't see the need to support multiple remotes for now.

Issue #9683, opened 2022-04-14
Support picking IJupyterServerUriInfo in command jupyter.selectjupyteruri 


rebornix:
As part of the lazy kernel exploration https://github.com/microsoft/vscode/issues/146942#issuecomment-1097405365 , we want to support `IJupyterServerUriInfo ` in command `jupyter.selectjupyteruri`. With this support

* Extensions can register their own remote server provider through `api.registerRemoteServerProvider`
* Extensions then ask Jupyter to switch to their remote server without showing the Remote Server quick pick

DonJayamanne:
This is related to the same work we need for Azure ML compute work. https://github.com/microsoft/vscode-jupyter-internal/issues/270

DonJayamanne:
<details>
<summary>Details 1</summary>

* Assume we run a cell, reload VS Code, now we have an entry in the cache for an active kernel with id `xyz` that points to a IJupyterServerUri with id = `a` and handle `handle1`
* Assume we shutdown VS Code
	* The extension that provides the handle `handle1` has now shutdown the jupyter server, hence we know the active kernel `xyz` will no longer exist.
* Reload VS Code
	* The 3rd party extension will now return `IJupyterServerUri` with id = 'a' and handle `handle2`.
	* Jupyter Extension still lists the old kernel `xyz`

**Questions:**
* Should we remove the entry for old active kernel `xyz`? I don't think we should, we should leave this so user knows that it is no longer available (required as it solves issue #9167)
* If user runs a cell against this, what should the user do?
	* There's no point providing the option to remove the old server as its no longer available either.
	* Should we instead keep the old server in the list as well? I.e. should we have an entry for the server with the handle `handle1` and `handle2`, if yes, then display names MUST be unique. But how do we guarantee this, they are created by 3rd party extensions.
	* Should we instead just let the user know that the server is no longer available and provide an option to just select another kernel (this feels like the most valid option)


</details>

rchiodo:
> Reload VS Code
The 3rd party extension will now return IJupyterServerUri with id = 'a' and handle handle2.
Jupyter Extension still lists the old kernel xyz

AFAIK it's not working like this today. The handle is stored by us in the encrypted storage. We always give it to the 3rd party extension to provide us with a URI. So they can generate a new URI if they want.

rchiodo:
Additionally I believe we use that handle to find the extension to activate in order to ask them for a new URI.

DonJayamanne:
[...]

PR #9957, opened 2022-05-09
Remove uri from INotebookServerOptions

DonJayamanne:
Part of #9684 & #9683


src/kernels/jupyter/launcher/liveshare/serverCache.ts
```diff
@@ -110,7 +110,6 @@ export class ServerCache implements IAsyncDisposable {
         }
         return {
             serverId: options.serverId,
-            uri: options.uri,
```
DonJayamanne, line 113: I think this is good now, we don't really need to pass the Uri around as this also contains sensitive information & we don't accidentally log this anymore.

rchiodo (approved):

PR #9978, opened 2022-05-11
Support returning a preferred remote controller

DonJayamanne:
Part of #9683

Part of this PR:
* Introduce the method `selectJuptyerUri` (but renamed to `getSuggestedController`)
* Ensure extensions won't need to add an entry named `pythonRemoteServerProvider` into their package.json
* Rename `computeUri` to `computeServerId`, that seems more meaningful as that's exactly what it is.
* Throw an error to handle situations where server providers (3rd party extensions) cannot return the IJupyterServerUri information (this way we can display meaning full errors)
* Changes to code to return preferred remote notebook controller (look for kernelspec name, else just fallback to the language)

**Note:**
The api method `selectJuptyerUri` described here https://github.com/microsoft/vscode/issues/146942 has been renamed to  `getSuggestedController`, as the Jupyter API doesn't select any controller, it merely returns the suggested controller for a given remote Jupyter server.
(discussed with Peng), and will update the upstream issue once this work has been completed.


.vscode/settings.json
```diff
@@ -1,7 +1,7 @@
 // Place your settings in this file to overwrite default and user settings.
 {
     "files.exclude": {
-        "out": true, // set this to true to hide the "out" folder with the compiled JS files
+        "out": false, // set this to true to hide the "out" folder with the compiled JS files
```
DonJayamanne, line 4: I think this is ok to have, i keep setting this.

package.nls.json
```diff
@@ -288,8 +288,8 @@
     "DataScience.jupyterSelectPasswordPrompt": "Enter your password",
     "DataScience.removeRemoteJupyterServerEntryInQuickPick": "Remove",
     "DataScience.jupyterNotebookFailure": "Jupyter notebook failed to launch. \r\n{0}",
-    "DataScience.remoteJupyterConnectionFailedWithServer": "Failed to connect to the remote Jupyter Server {0}. Verify the server is running and reachable.",
-    "DataScience.remoteJupyterConnectionFailedWithServerWithError":"Failed to connect to the remote Jupyter Server {0}. Verify the server is running and reachable. ({1}).",
+    "DataScience.remoteJupyterConnectionFailedWithServer": "Failed to connect to the remote Jupyter Server '{0}'. Verify the server is running and reachable.",
+    "DataScience.remoteJupyterConnectionFailedWithServerWithError":"Failed to connect to the remote Jupyter Server '{0}'. Verify the server is running and reachable. ({1}).",
```
DonJayamanne, line 292: Was previously displaying the baseUrl, but in the case of somethign like Azure ML, we can display the displayName (which could contain spaces), hence quoting this string.

src/kernels/jupyter/commands/serverSelector.ts
```diff
@@ -48,7 +48,7 @@ export class JupyterServerSelectorCommand implements IDisposable {
             // Find one that is the default for this remote
             if (notebook) {
                 // Recompute the preferred controller
-                await this.controllerManager.computePreferredNotebookController(notebook);
+                await this.controllerManager.initializePreferredNotebookController(notebook);
```
DonJayamanne, line 51: This was not only computing the preferred controller, it was also setting it using the VS Code api.
I've renamed this to be more specific.
The method `computePreferredNotebookController` is now used to return the preferred controller without changing it.

src/kernels/jupyter/jupyterConnection.ts
```diff
@@ -65,7 +66,7 @@ export class JupyterConnection implements IExtensionSyncActivationService {
     private async getUriFromServerId(serverId: string) {
         // Since there's one server per session, don't use a resource to figure out these settings
         const savedList = await this.serverUriStorage.getSavedUriList();
-        return savedList.find((item) => computeUriHash(item.uri) === serverId)?.uri;
+        return savedList.find((item) => computeServerId(item.uri) === serverId)?.uri;
```
DonJayamanne, line 69: Renamed `computeUriHash` to be `computeServerId`, the implementation should not define the method name, its a server id, hence the name should reflect that.

src/kernels/jupyter/jupyterConnection.ts
```diff
@@ -92,18 +93,24 @@ export class JupyterConnection implements IExtensionSyncActivationService {
...
+                    const timeoutInMS = serverUri.expiration.getTime() - Date.now();
+                    // Week seems long enough (in case the expiration is ridiculous)
+                    if (timeoutInMS > 0 && timeoutInMS < 604800000) {
+                        this.pendingTimeouts.push(
+                            setTimeout(() => this.updateServerUri(uri).ignoreErrors(), timeoutInMS)
+                        );
+                    }
                 }
+            } catch (ex) {
+                throw new RemoteJupyterServerUriProviderError(idAndHandle.id, idAndHandle.handle, ex);
```
DonJayamanne, line 113: Throw a custom error so we can handle this, e.g. possible the jupyter server is down and user can now choose to ignore this connection completely.

src/kernels/jupyter/jupyterUriProviderRegistration.ts
```diff
@@ -53,14 +61,35 @@ export class JupyterUriProviderRegistration implements IJupyterUriProviderRegist
     }
 
     private async loadOtherExtensions(): Promise<void> {
+        const extensionIds = new Set<string>();
+        this.globalMemento
```
DonJayamanne, line 65: In the case of `github extensions`, they can register the lazy proxy controller when a notebook opens and they register the provider with us.
When we reload vscode, we try to get the provider information, however its possible github or other extensions haven't activated yet, for such extensions we don'e need an entry in package.json for `pythonRemoteServerProvider`, instead we can just keep track of the extensions that own a specific provider and load and activate those extensions.

src/kernels/jupyter/jupyterUriProviderWrapper.ts
```diff
@@ -14,6 +14,9 @@ export class JupyterUriProviderWrapper implements IJupyterUriProvider {
         return this.provider.id;
     }
     public getQuickPickEntryItems(): vscode.QuickPickItem[] {
+        if (!this.provider.getQuickPickEntryItems) {
```
DonJayamanne, line 17: Extensions such as github or others won't need to implement `getQuickPickEntryItems` (see type definition)

src/kernels/jupyter/serverSelector.ts
```diff
@@ -79,6 +84,32 @@ export class JupyterServerSelector {
         const multiStep = this.multiStepFactory.create<{}>();
         return multiStep.run(this.startSelectingURI.bind(this, allowLocal), {});
     }
+
+    @captureTelemetry(Telemetry.EnterJupyterURI)
+    @traceDecoratorError('Failed to enter Jupyter Uri')
+    public async enterJupyterURI(): Promise<string | undefined> {
```
DonJayamanne, line 90: Wrote this code for server picker for our own lazy controller, don't see any harm keeping this here.
Not used for now, as i didn't want to checkin the layzy proxy controller as that API could change.
Requries a bit more work to remove this hence left it, hope its ok (the plan is to make use of this in the future)

src/kernels/jupyter/session/backingFileCreator.base.ts
```diff
@@ -17,6 +17,13 @@ function getRemoteIPynbSuffix(): string {
     return `${jvscIdentifier}${uuid()}`;
 }
 
+export function generateBackingIPyNbFileName(resource: Resource) {
```
DonJayamanne, line 20: Generic fucntion so that we can have a meaningful name, if we fail to create the backing file. Previously it used to fallback to `<uuid>.ipynb`, and that was ugly.
E.g. point to a jupyter server that doesn't support creation of check points

src/kernels/jupyter/session/jupyterSession.ts
```diff
@@ -234,7 +235,7 @@ export class JupyterSession extends BaseJupyterSession {
 
         // Create our session options using this temporary notebook and our connection info
         const sessionOptions: Session.ISessionOptions = {
-            path: backingFile?.filePath || `${uuid()}.ipynb`, // Name has to be unique
+            path: backingFile?.filePath || generateBackingIPyNbFileName(this.resource), // Name has to be unique
```
DonJayamanne, line 238: Now all paths generate have a predictable name & we can strip out the unnecessary guff and have a meaningful and prettier display name in the kernel picker, previously we'd just display a guid.

src/kernels/jupyter/types.ts
```diff
@@ -202,10 +202,21 @@ export interface IJupyterServerUri {
...
-    readonly id: string; // Should be a unique string (like a guid)
-    getQuickPickEntryItems(): QuickPickItem[];
-    handleQuickPick(item: QuickPickItem, backEnabled: boolean): Promise<JupyterServerUriHandle | 'back' | undefined>;
+    /**
+     * Should be a unique string (like a guid)
+     */
+    readonly id: string;
+    onDidChangeHandlers?: Event<void>;
+    getQuickPickEntryItems?(): QuickPickItem[];
+    handleQuickPick?(item: QuickPickItem, backEnabled: boolean): Promise<JupyterServerUriHandle | 'back' | undefined>;
```
DonJayamanne, line 211: Extensions such as github or others won't need to implement `getQuickPickEntryItems`
Haven't hooked up the code to make use of this API just yet, but added this nevertheless, basically as defined here https://github.com/microsoft/vscode/issues/146942

src/kernels/kernelFinder.base.ts
```diff
@@ -51,22 +52,17 @@ export abstract class BaseKernelFinder implements IKernelFinder {
     public async rankKernels(
         resource: Resource,
         notebookMetadata?: nbformat.INotebookMetadata,
+        preferredInterpreter?: PythonEnvironment,
         cancelToken?: CancellationToken,
-        useCache?: 'useCache' | 'ignoreCache'
+        useCache?: 'useCache' | 'ignoreCache',
+        serverId?: string
```
DonJayamanne, line 58: Added a new argument `serverId` so that we can filter out the kernels that apply to a specific remote jupyter server.
Applies to the lazy controller that spins up jupyter and then asks us to get the prefered controller (surely at that point we need to return the preferred controller specific to that jupyter server)

src/kernels/kernelFinder.base.ts
```diff
@@ -51,22 +52,17 @@ export abstract class BaseKernelFinder implements IKernelFinder {
...
     ): Promise<KernelConnectionMetadata[] | undefined> {
-        const resourceType = getResourceType(resource);
         try {
             // Get list of all of the specs from the cache and without the cache (note, cached items will be validated before being returned)
-            const kernels = await this.listKernels(resource, cancelToken, useCache);
-
-            const isPythonNbOrInteractiveWindow = isPythonNotebook(notebookMetadata) || resourceType === 'interactive';
-
-            // Always include the interpreter in the search if we can
-            const preferredInterpreter =
```
DonJayamanne, line 65: Removed this code from the base method, the exact same code `preferredInterpreter` is calculated in the calling code, hence might as well pass that into this method instead of computing this here, also remove a few dependencies.

src/kernels/jupyter/types.ts
DonJayamanne: The code that uses this has a few more changes, figured i'd submit this PR without those changes as its pretty large as is.

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -423,12 +427,29 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+                : metadata.language_info.name;
+        const kernelName = metadata ? metadata.kernelspec?.name : undefined;
         // Get all remote kernels
         await this.loadNotebookControllers();
-        const controllers = this.getRegisteredNotebookControllers();
+        const controllers = this.getRegisteredNotebookControllers().filter((item) => {
+            // Sort out interactive or non-interactive controllers
+            if (
+                item.connection.kind !== 'startUsingRemoteKernelSpec' ||
+                (notebookType === 'jupyter-notebook' && item.id.includes(InteractiveControllerIdSuffix)) ||
```
rchiodo, line 446: This logic seems backwards? notebooktype==='jupyter-notebook' would never have the interactive controller id suffix?

src/platform/api.ts
```diff
@@ -49,6 +51,14 @@ export interface IExtensionApi {
      * There are a specific set of extensions that are currently allowed to access this API.
      */
     getKernelService(): Promise<IExportedKernelService | undefined>;
+    /**
+     * Returns the suggested controller for a give Jupyter server and notebook.
+     */
+    getSuggestedController(
+        id: string,
```
rchiodo, line 58: What's id in this case? Kernel id?

rchiodo: Oh no it's actually the id of the IJupyterRemoteServerUriProvider. Maybe put a comment here? Or rename the parameter?

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -441,11 +461,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         let defaultPythonLanguageKernel: VSCodeNotebookController | undefined;
         controllers.forEach((item) => {
             // Sort out interactive or non-interactive controllers
-            if (
-                item.connection.kind !== 'startUsingRemoteKernelSpec' ||
-                (notebookType === 'jupyter-notebook' && item.id.includes(InteractiveControllerIdSuffix)) ||
-                (notebookType === 'interactive' && !item.id.includes(InteractiveControllerIdSuffix))
-            ) {
+            if (item.connection.kind !== 'startUsingRemoteKernelSpec') {
```
IanMatthewHuff, line 464: Looks like this line is covered by line 445 above now?

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -457,7 +473,27 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+            let matchingKernelNameController: VSCodeNotebookController | undefined;
+            let matchingKernelLanguageController: VSCodeNotebookController | undefined;
+            controllers.forEach((item) => {
+                // Sort out interactive or non-interactive controllers
+                if (item.connection.kind !== 'startUsingRemoteKernelSpec') {
+                    return;
+                }
+                if (item.connection.kernelSpec.name === kernelName) {
+                    matchingKernelNameController = item;
+                } else if (item.connection.kernelSpec.language === language) {
```
IanMatthewHuff, line 490: When I moved over the exact match code I ran into one existing error case where the kernelspec language didn't match up with the notebook language_info languages due to capitalization, so maybe need a toLower check here? Pretty sure it was for a C# file.

IanMatthewHuff (approved):
Two small comments, but looks good to me.

src/kernels/jupyter/session/backingFileCreator.base.ts
DonJayamanne: Yes agree, if you see the code again, the `suffix` does contain a uuid, hence it is unique.

src/notebooks/controllers/notebookControllerManager.ts
DonJayamanne: Yes, however types a funky, if i don't add this, then the item could be any controller type.
This is because item is a IVSCodeNotebookController and it has a property of type KenelConnectionMetadata, which could be local or remote.
even though we filter it out earlier, the type system doesn't know that

src/notebooks/controllers/notebookControllerManager.ts
DonJayamanne: Oh shoot, I thought I had a translation code, will use that.

DonJayamanne: The function that you can use to translate from kenel to monaco language is `translateKernelLanguageToMonaco`
Here it doesn't apply as we're only dealing with languages of kernelspecs & no language components from VS Code (monaco).

rchiodo (approved):

PR #9985, opened 2022-05-11
Revision to displaying cached remote kernel specs and live kernels

DonJayamanne:
Part of #9167 

The previous implementation was incorrect and as follows:
* We used to cache and always display the cached remove kernel specs and remote live kernels

However the requirement is:
* If the user attempts to connect to a remote kernel, then notify them about the errors. 

To make this possible we decided to only display the kernels they had used.
The implementation of this logic was incorrect:
* We loaded all cached remote kernelspecs & remote live kernels
* Then when we get the latest kernelspecs & latest remove live kernels we'd remove some of them
* & if the user attempted to use any of these outdated items, we'd notify them.

Problems
* Therre's no need to display the cached kernelspecs, we can wait to get the latest kernel specs from the server
* Instead of displayed all caching live remove kernels we only need to display a live kernel that was used by a notebook. If a user connects to a remote jupyter server there could already be a few live remote kernels on that machine, and we don't need to display those, as the user never used them in vscode.

**Revised solution**
* Do not display cached remote kernelspecs (get latest from server)
* Display cached live remote kernels that were used once before (later remove them if the user switches a kernel or removes these remote jupyter servers completely)


src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -314,6 +307,27 @@ export class VSCodeNotebookController implements Disposable {
             }
             this.associatedDocuments.delete(event.notebook);
             this._onNotebookControllerSelectionChanged.fire();
+
```
rchiodo, line 310: You could reverse the dependency here and have live remote kernel tracker listen to notebook selection change events. Then this code wouldn't have to tell the live tracker about its changes.

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -236,11 +234,6 @@ export class VSCodeNotebookController implements Disposable {
         if (cells.length < 1) {
             return;
         }
-        if (this.isConnectionOutdated && !this.outdatedMessageDisplayed) {
-            void this.appShell.showErrorMessage(DataScience.jupyterRemoteConnectFailedModalMessage(), { modal: true });
```
DonJayamanne, line 240: This is no longer required, we have better ways to handle this now.

DonJayamanne: I.e. we let the connection crash and handle the errors in the usual central location `errorHandler.ts`

src/notebooks/controllers/vscodeNotebookController.ts
DonJayamanne: Yeah that is much better.

DonJayamanne: much much

DonJayamanne: Fixed

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -137,8 +140,6 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         @inject(ICommandManager) private readonly commandManager: ICommandManager,
         @inject(IExtensionContext) private readonly context: IExtensionContext,
         @inject(IKernelProvider) private readonly kernelProvider: IKernelProvider,
-        @inject(PreferredRemoteKernelIdProvider)
-        private readonly preferredRemoteKernelIdProvider: PreferredRemoteKernelIdProvider,
```
DonJayamanne, line 141: Removed this as well, as this too could be handled the same way as the new code.

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -632,9 +600,6 @@ export class VSCodeNotebookController implements Disposable {
         // Before we start the notebook, make sure the metadata is set to this new kernel.
         await updateNotebookDocumentMetadata(document, this.documentManager, selectedKernelConnectionMetadata);
 
-        if (isLocalConnection(this.connection)) {
```
rchiodo, line 635: Nice, love the removal of the knowledge of remote.

rchiodo (approved):

src/notebooks/controllers/vscodeNotebookController.ts
DonJayamanne: Agreed, one less dependency less complex

PR #9988, opened 2022-05-11
Fix problem where validation happens on acceptance but failure is ignored

rchiodo:
Started work on ipywidgets and I was getting a cert error for my jupyter server. 

But it wasn't displayed anywhere. 

Recent changes made the validate async, but not sure how the 'change' event doesn't catch the error. Maybe because validate takes longer now. 

This logic seems better to me. Validate on enter, not while typing, especially if validate is actually going to the server.

### DIFF
--- a/src/platform/common/utils/multiStepInput.ts
+++ b/src/platform/common/utils/multiStepInput.ts
@@ -196,7 +196,6 @@ export class MultiStepInput<S> implements IMultiStepInput<S> {
                 input.prompt = prompt;
                 input.ignoreFocusOut = true;
                 input.buttons = [...(this.steps.length > 1 ? [QuickInputButtons.Back] : []), ...(buttons || [])];
-                let validating = validate('');
                 disposables.push(
                     input.onDidTriggerButton((item) => {
                         if (item === QuickInputButtons.Back) {
@@ -209,19 +208,19 @@ export class MultiStepInput<S> implements IMultiStepInput<S> {
                         const inputValue = input.value;
                         input.enabled = false;
                         input.busy = true;
-                        if (!(await validate(inputValue))) {
+                        const validationMessage = await validate(inputValue);
+                        if (!validationMessage) {
+                            input.validationMessage = '';
                             resolve(inputValue);
+                        } else {
+                            input.validationMessage = validationMessage;
                         }
                         input.enabled = true;
                         input.busy = false;
                     }),
-                    input.onDidChangeValue(async (text) => {
-                        const current = validate(text);
-                        validating = current;
-                        const validationMessage = await current;
-                        if (current === validating) {
-                            input.validationMessage = validationMessage;
-                        }
+                    input.onDidChangeValue(async () => {
+                        // Validation happens on acceptance. Just clear as the user types
+                        input.validationMessage = '';
                     }),
                     input.onDidHide(() => {
                         (async () => {
--- a/src/test/datascience/mockInputBox.ts
+++ b/src/test/datascience/mockInputBox.ts
@@ -59,7 +59,9 @@ export class MockInputBox implements InputBox {
             }, 10);
         }, 10);
     }
-    public hide(): void {}
+    public hide(): void {
+        // Do nothing
+    }
     public dispose(): void {
         // Do nothing
     }
--- a/src/test/datascience/mockInputBox.ts
+++ b/src/test/datascience/mockInputBox.ts
@@ -47,13 +47,19 @@ export class MockInputBox implements InputBox {
                     this.didHideEmitter.fire();
                 } else {
                     this.didAcceptEmitter.fire();
+
+                    // Accept may cause validation failure. Check for that
+                    setTimeout(() => {
+                        if (this.validationMessage) {
+                            this.value = this.validationMessage;
+                            this.didHideEmitter.fire();
+                        }
+                    }, 100);
                 }
             }, 10);
         }, 10);
     }
-    public hide(): void {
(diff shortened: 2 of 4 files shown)


src/platform/common/utils/multiStepInput.ts
```diff
@@ -209,19 +208,19 @@ export class MultiStepInput<S> implements IMultiStepInput<S> {
                         const inputValue = input.value;
                         input.enabled = false;
                         input.busy = true;
-                        if (!(await validate(inputValue))) {
+                        const validationMessage = await validate(inputValue);
+                        if (!validationMessage) {
+                            input.validationMessage = '';
                             resolve(inputValue);
+                        } else {
+                            input.validationMessage = validationMessage;
```
rchiodo, line 216: This wasn't being applied so if you hit enter and a failure occurs, nothing would show up.

DonJayamanne (approved):

src/platform/common/utils/multiStepInput.ts
```diff
@@ -209,19 +208,19 @@ export class MultiStepInput<S> implements IMultiStepInput<S> {
...
                         }
                         input.enabled = true;
                         input.busy = false;
                     }),
-                    input.onDidChangeValue(async (text) => {
```
DonJayamanne, line 218: I think we should bring this back, this will impact other places where this is used,
also i don't think it matters if we have this validation as the user types the text

DonJayamanne (changes requested):
I think we should restore validation on type

src/platform/common/utils/multiStepInput.ts
DonJayamanne: > vent doesn't catch the error. Maybe because validate takes longer now.

Even if thats the case, we'd still have the valdiation upon hitting enter.
& even if the values change, then those results would trump the previous error messages

rchiodo: Validation on typing doesn't work because if the validation is the same error, it looks like it didn't do anything. It looks broken.

rchiodo:
Here's an example of why I think we shouldn't have validate on change. 
errorNotChanging

The user can't tell if we're actually checking or not.

rchiodo:
We could ask what others think.

@IanMatthewHuff @amunger @sadasant any thoughts on having validation only on enter?

src/platform/common/utils/multiStepInput.ts
DonJayamanne: Does that matter, if the code works as expected, then the user keeps seeing an error message
I don't see why they'd expect different errors when the underlying problem is still the same - i.e. url is invalid.

DonJayamanne:
@rchiodo can you update the gif to include how it looks when the Uri is correct.
I.e. right now all you're showing is something not working, which is correct. 
What happens when the Uri is valid, please could you update the gif to include that as well.

src/platform/common/utils/multiStepInput.ts
rchiodo: Cause it looks like we didn't check again. They typed in something and got no changes whatsoever (even when the url isn't complete).

They're done when they hit enter. That's when I would expect it to validate.

IanMatthewHuff:
Speaking just from the user perspective I almost always (bar a few situations) prefer validation on commit instead of type. I want to wait until I've put in what I want to put in, then notify me if it's wrong. But I'm not sure if this is a hard and fast rule, or just my personal pref. 

> We could ask what others think.
> 
> @IanMatthewHuff @amunger @sadasant any thoughts on having validation only on enter?

rchiodo:
> @rchiodo can you update the gif to include how it looks when the Uri is correct. I.e. right now all you're showing is something not working, which is correct. What happens when the Uri is valid, please could you update the gif to include that as well.

The red just disappears. 

The problem I have is I can't tell if it's being validated or not because I can type whatever I want and it gives me the same results until I type a valid server.

If say I typed in wrong character I'd not be able to tell if we were checking or not.

IanMatthewHuff:
@rchiodo I expressed my pref for "on commit" but maybe worth asking the UX channel? UX experts might have some guidance here given the input box and the type of input (a long server string).

IanMatthewHuff (approved):

DonJayamanne:
> If say I typed in wrong character I'd not be able to tell if we were checking or not.

Yup makes sense.

DonJayamanne (approved):

rchiodo:
There's also a delay problem so sometimes it doesn't validate as a I type and then it randomly turns red. So I might be halfway through typing the correct URI and it suddenly turns red.

rchiodo:
Asking UX sounds like a good idea to me.

src/test/datascience/mockInputBox.ts
```diff
@@ -47,6 +47,14 @@ export class MockInputBox implements InputBox {
...
                 } else {
                     this.didAcceptEmitter.fire();
+
+                    // Accept may cause validation failure. Check for that
+                    setTimeout(() => {
+                        if (this.validationMessage) {
+                            this.value = this.validationMessage;
+                            this.didHideEmitter.fire();
+                        }
+                    }, 100);
```
sadasant, line 57: This timeout seems pretty arbitrary. Is it possible for validationMessage to be populated after 100ms? would an interval be more reliable?

rchiodo: This is only used in a unit test so the only way this timeout doesn't work is if you're debugging the unit test.

PR #9989, opened 2022-05-11
Handle when Jupyter server handle is not valid

DonJayamanne:
Part of #9683

**Fixes**
* If the `handle` we have is no longer valid, then connecting to Juypyter will no longer work as the 3rd party extension doesn't support the old `handle` that was initially returned.
	* We now display a message if the handle is outdated and allow the user to forget that connection (each connection is unique per `providerId+handle`)
* Added a new method to add a remote jupyter server uri instead of using a command (as per suggestion from Peng)
	* Makes sense as thats a more concrete API with types.

### DIFF
--- a/src/platform/api.ts
+++ b/src/platform/api.ts
@@ -6,6 +6,7 @@
 import { ExtensionMode, NotebookController, NotebookDocument } from 'vscode';
 import { JupyterConnection } from '../kernels/jupyter/jupyterConnection';
 import { computeServerId, generateUriFromRemoteProvider } from '../kernels/jupyter/jupyterUtils';
+import { JupyterServerSelector } from '../kernels/jupyter/serverSelector';
 import { IJupyterUriProvider, IJupyterUriProviderRegistration, JupyterServerUriHandle } from '../kernels/jupyter/types';
 import { INotebookControllerManager, INotebookEditorProvider } from '../notebooks/types';
 import { IDataViewerDataProvider, IDataViewerFactory } from '../webviews/extension-side/dataviewer/types';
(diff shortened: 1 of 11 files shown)


src/platform/api.ts
```diff
@@ -112,6 +118,13 @@ export function buildApi(
             const serverId = computeServerId(uri);
             const { controller } = await controllers.computePreferredNotebookController(notebook, serverId);
             return controller?.controller;
+        },
+        addRemoteJupyterServer: async (id: string, handle: JupyterServerUriHandle) => {
+            const connection = serviceContainer.get<JupyterConnection>(JupyterConnection);
+            const selector = serviceContainer.get<JupyterServerSelector>(JupyterServerSelector);
+            const uri = generateUriFromRemoteProvider(id, handle);
+            await connection.updateServerUri(uri);
+            await selector.setJupyterURIToRemote(uri);
```
DonJayamanne, line 127: Currently when this code gets executed we end up switching to the new Uri, its not necessarily an add. However thats an implementation detail of Jupyter extension (which will be fixed later)

src/platform/api.ts
```diff
@@ -59,6 +60,11 @@ export interface IExtensionApi {
         handle: JupyterServerUriHandle,
         notebook: NotebookDocument
     ): Promise<NotebookController | undefined>;
+    /**
+     * Adds a remote Jupyter Server to the list of Remote Jupyter servers.
+     * This will result in the Jupyter extension listing kernels from this server as items in the kernel picker.
+     */
+    addRemoteJupyterServer(id: string, handle: JupyterServerUriHandle): Promise<void>;
```
rchiodo, line 67: id is 'providerId' here too?

rchiodo (approved):

src/platform/api.ts
DonJayamanne: Damn it, yes

DonJayamanne: Damn it = cursing my self for forgetting this the second time.
Not directed at you :)

PR #9994, opened 2022-05-12
Remove usage of getUri in favor of getRemoteUri

DonJayamanne:
Part of #9167 and #9684

### DIFF
--- a/src/kernels/jupyter/types.ts
+++ b/src/kernels/jupyter/types.ts
@@ -235,9 +235,9 @@ export interface IJupyterServerUriStorage {
     getSavedUriList(): Promise<{ uri: string; time: number; displayName?: string }[]>;
     removeUri(uri: string): Promise<void>;
     clearUriList(): Promise<void>;
-    getUri(): Promise<string>;
     getRemoteUri(): Promise<string | undefined>;
-    setUri(uri: string): Promise<void>;
+    setUriToLocal(): Promise<void>;
+    setUriToRemote(uri: string): Promise<void>;
 }
 
 export const IJupyterBackingFileCreator = Symbol('IJupyterBackingFileCreator');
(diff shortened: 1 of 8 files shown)


src/kernels/jupyter/types.ts
```diff
@@ -235,9 +235,9 @@ export interface IJupyterServerUriStorage {
     getSavedUriList(): Promise<{ uri: string; time: number; displayName?: string }[]>;
     removeUri(uri: string): Promise<void>;
     clearUriList(): Promise<void>;
-    getUri(): Promise<string>;
     getRemoteUri(): Promise<string | undefined>;
```
DonJayamanne, line 238: The plan is to use `getRemoteUri` instead of `getUri`, else we compare whehther its local or not and compare the string value.
This also paves the way to just have a simple function `getRemoteUri(): Promise<string[]>` which will return multiple uris if we have multiple remotes, and if its empty, then the conclusion is simple we don't have any remotes, (without that we need to compare strings, which is more of an internal implementation detail)

src/kernels/jupyter/types.ts
```diff
@@ -235,9 +235,9 @@ export interface IJupyterServerUriStorage {
...
-    setUri(uri: string): Promise<void>;
+    setUriToLocal(): Promise<void>;
```
DonJayamanne, line 239: Added these two methods so that the uri string used to determien whether its local or remote isn't exposed elsewhere.

src/kernels/kernelFinder.base.ts
```diff
@@ -98,10 +98,6 @@ export abstract class BaseKernelFinder implements IKernelFinder {
             }),
             this.listRemoteKernels(resource, cancelToken, useCache).catch((ex) => {
                 traceError('Failed to get remote kernels', ex);
-                // When remote kernels fail, turn off remote if we get a ECONNREFUSED error
-                if (ex.toString().toLowerCase().includes('econn')) {
```
DonJayamanne, line 102: This should be removed for #9167, as the user will decide what to do with old (busted) remotes.
I.e. if the remote is invalid, we log the errors, and if things have been cached, then we'll display them and user will be notified when they attempt to run a cell, at which point they can chose to remove that connection or switch to local or the like.

sadasant (approved):
Looks like a good cleanup to me!

PR #10008, opened 2022-05-12
Removing path-browserify

sadasant:
I noticed that while we were using a cross-environment version of `path`, in a couple of places we were still using `@types/path-browserify`. I asked about it and this seems unintentional. Here's a PR removing this package.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27864,7 +27864,7 @@
                 "minimist": "~1.2.0",
                 "moment": "^2.24.0",
                 "path-browserify": "^1.0.0",
-                "url-parse": "1.5.8"
+                "url-parse": "~1.5.1"
             }
         },
         "@jupyterlab/nbformat": {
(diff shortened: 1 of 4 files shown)


rchiodo:
I'm guessing you're using the wrong version of npm and that's why the package-lock.json is so different.

package-lock.json
```diff
@@ -45486,8 +45473,7 @@
             }
         },
         "url-parse": {
-            "version": "1.5.10",
-            "resolved": "https://registry.npmjs.org/url-parse/-/url-parse-1.5.10.tgz",
+            "version": "https://registry.npmjs.org/url-parse/-/url-parse-1.5.10.tgz",
```
rchiodo, line 45476: This change seems off.

sadasant: checking

IanMatthewHuff: Funky. Don't think I've ever seen that before.

sadasant: I had `npm -v 8.5.5`. I `nvm use v16` rather than installing `16.13.0`. Now I'm using the right version of Node (16.13.0) and NPM (8.1.0) 🙂

rchiodo (approved):

IanMatthewHuff (approved):

Issue #9996, opened 2022-05-12
Support Export in Web


rebornix:
This is part of https://github.com/microsoft/vscode-jupyter/issues/9574 

Currently the Export actions are powered by `nbconvert`. We will run `nbconvert` in a separate process and save the converted content in a temporary file on disk. This is not feasible on Web as Jupyter runs in web worker and there is no `jupyter/nbconvert` in the same environment. To archive the same thing on Web, we would need to ask the remote Jupyter server/service to handle the nbconvert request for us.

There are two ways to run `nbconvert` on the remote Jupyter server:
* The first one is directly calling the nbconvert web service, we can construct an url as `${baseUrl}/nbconvert/${type}/${filePath}?token=${token}` and fetch this resource in the web extension. However it requires the file to **be synced to** the remote Jupyter server 
* The second one is calling `nbconvert` through shell commands
  * We can either run it against files, i.e., `!jupyter nbconvert test.ipynb --to python --stdout`, but it has the same problem as the first approach
  * Or we can pass in the content through `stdin`: `!echo '{"cells": [], "metadata": {},  "nbformat": 4, "nbformat_minor": 2 }' | jupyter nbconvert  --nbformat 4 --stdin --to python --stdout`. This is the only approach that doesn't require file to be synced but if the file is large, it will affect the throughput.
  * The **catch** with running shell commands is it requires us to know the path of the active `jupyter` executable as it might exist in `PATH` and it needs to be platform specific. We might need to run `{sys.executable} -m jupyter nbconvert`

rchiodo:
We've used the {sys.executable} trick in the past to run python on the jupyter server. 

Another idea? Zip the file on the local machine, send zip with jupyter file transfer API, reopen zip in temp folder and run nbconvert on it?

The stdin way sounds plausible but I'm worried about gb size notebooks (that zip up to mbs)

PR #10016, opened 2022-05-13
Export to Python script on Web

rebornix:
Re https://github.com/microsoft/vscode-jupyter/issues/9996

This PR is the initial prep for supporting Export commands on Web. It includes following changes:

* Move `src/interactive-window/commands/exportCommands-node.ts` off node. https://github.com/microsoft/vscode-jupyter/commit/763bf7069bf8abdcbbcd311a04e2e74b0e20e756 Majorly removing dependencies on `fs`
* Move `exportDialog` and `exportFileOpener.ts` off node.
* Common `src/platform/export/exportToPythonPlain.ts` https://github.com/microsoft/vscode-jupyter/commit/8beca33b170a91df23a5450cb8ae8bb9df0baebe. `exportToPythonPlain.node` and `exportToPythonPlain.web` will behave differently on EOL checking
* Lastly, register Export To Python (plain, not nbconvert) to Web service entry

To test the change, run "Extension (web)" and tweak the `when` clause of `jupyter.exportAsPythonScript` command to test on notebooks (as we don't have Interactive Window yet).


src/interactive-window/commands/commandRegistry.web.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    private registerCommandsIfTrusted() {
+        if (!this.workspace.isTrusted) {
+            return;
+        }
+        this.exportCommand?.register();
+    }
+
+    dispose(): void {
```
DonJayamanne, line 35: Not required, the `this.disposables` is actually the global registry, that will get cleared when vscode shutdowns or extension is unloaded.

src/platform/common/platform/fileSystem.ts
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the MIT License.
+
+'use strict';
+
+import * as vscode from 'vscode';
+import { areStringPathsSame } from './fileUtils';
+
+const ENCODING = 'utf8';
+
+export async function writeFile(uri: vscode.Uri, text: string | Buffer): Promise<void> {
```
DonJayamanne, line 11: Can we try to use what we have today, I think we should be able to use the IFileSystem as is.
@rchiodo pinging you to check whether the filesystem code has been web-enabled or not.

src/platform/common/platform/fileUtils.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+'use strict';
+
+import { getOSType, OSType } from '../utils/platform';
+import * as path from '../../../platform/vscode-path/path';
+
+export function normCasePath(filePath: string): string {
+    return getOSType() === OSType.Windows ? path.normalize(filePath).toUpperCase() : path.normalize(filePath);
```
DonJayamanne, line 10: I think we have similar code already in `fileUtils.node.ts`
Instead of copying it here I think we can move it into a common place and have both use the same shared code.

src/platform/export/exportDialog.ts
```diff
@@ -54,7 +59,8 @@ export class ExportDialog implements IExportDialog {
         const targetFileName =
             defaultFileName || !source
                 ? defaultFileName || ''
-                : `${path.basename(source.fsPath, path.extname(source.fsPath))}${extension}`;
+                : // eslint-disable-next-line local-rules/dont-use-fspath
```
DonJayamanne, line 62: Do you need this linter exception, as you aren't using fspath.

src/platform/export/exportToPythonPlain.web.ts
```diff
@@ -0,0 +1,22 @@
...
+import { ExportToPythonPlainBase } from './exportToPythonPlain';
+
+// Handles exporting a NotebookDocument to python
+@injectable()
+export class ExportToPythonPlain extends ExportToPythonPlainBase {
+    public constructor(@inject(IConfigurationService) configuration: IConfigurationService) {
+        super(configuration);
+    }
+
+    override async writeFile(target: Uri, contents: string): Promise<void> {
```
DonJayamanne, line 19: Why do we need to override?

src/platform/export/exportToPythonPlain.ts
```diff
@@ -0,0 +1,97 @@
...
+import { CancellationToken, NotebookCell, NotebookCellKind, NotebookDocument, Uri } from 'vscode';
+import { appendLineFeed } from '../../webviews/webview-side/common';
+import { IConfigurationService } from '../common/types';
+import { IExport } from './types';
+
+// Handles exporting a NotebookDocument to python
+export class ExportToPythonPlainBase implements IExport {
+    public constructor(protected readonly configuration: IConfigurationService) {}
+
+    async writeFile(_target: Uri, _contents: string): Promise<void> {
```
DonJayamanne, line 15: I think marking this as abstract would be better.

src/platform/export/fileConverter.web.ts
```diff
@@ -0,0 +1,94 @@
...
+        token: CancellationToken,
+        _candidateInterpreter?: PythonEnvironment
+    ) {
+        // For web, we perform plain export for Python
+        if (format === ExportFormat.python) {
+            // Unless selected by the setting use plain conversion for python script convert
+            await this.performPlainExport(format, sourceDocument, target, token);
+            await this.exportFileOpener.openFile(format, target, true);
+        } else {
+        }
```
DonJayamanne, line 69: Throw not implemented or display a message not implemented?
Not sure, that way we do'nt forget,
Personally  I think errors are a good reminder about incomplete code.

DonJayamanne (changes requested):
Just some minor issues, but blocking on the code related to fs.
I think we can move into a common location instead of copying it into another file.

src/platform/common/platform/fileSystem.ts
rebornix: `fileSystem.node.ts` loads `fs-extra` and quite a few `*.node` files so it would be a huge change to make `fileSystem` web enabled in this PR.

src/platform/common/platform/fileUtils.ts
rebornix: Nice, will remove.

src/interactive-window/commands/commandRegistry.web.ts
rebornix: We did the same in https://github.com/microsoft/vscode-jupyter/blob/8f7f38e2bc664836af236e8e2ad9f1b85130c3c9/src/interactive-window/commands/commandRegistry.node.ts#L108 . I wonder if we need local `disposables` then.

rebornix:
@DonJayamanne I removed duplicates for `fileUltils`. However for making `filesystem` web enabled, it will be ambitious to have it in this PR as it will require checking the real usage of the `fileSystem.*` apis in each component. My suggestion would be 

* We take another PR to web enable `fileSystem`
* Or we do the "dup" approach, and when we move other components off node, we move away from `fileSystem.node` gradually

src/platform/export/exportDialog.ts
rebornix: The linter doesn't allow us to read `source.fsPath` https://github.com/microsoft/vscode-jupyter/pull/10016/files#diff-905cbd240f7aea1a954bb235c3b83cb741403d43d5133c0cf32ac8da3a22644eR63

src/platform/export/fileConverter.web.ts
rebornix: 👍

rebornix:
With all said above, it seems I do need some sort of `fileSystem.web` as `vscode.workspace.fs.readFile` is not available in unit test environment. I'll give it a try.

DonJayamanne:
> ems I do need some sort of fileSystem.web as vscode.workspace.fs.readFile is not available in unit test environment. I'll give it a try.

Yes,  I think we can implement only some of the methods of the IFileSystem for web, i.e. methods that are specific to local, can throw errors, Ideally those wouldn't be called in web scenario.

src/platform/common/platform/fileSystem.ts
DonJayamanne: Sorry, I meant re-using the same interface fofr IFileSystem, yes fs-extra cannot be used.
I was under the impression we already created a version of IFileSystem for web, sounds like we don't have it.

src/interactive-window/commands/commandRegistry.web.ts
DonJayamanne: There's no need, you can just pass that in here.
We tend to use local variables (to track the list of disposables)  for classes that get created and later disposed (e.g. class created via a factory that later gets disposed, e.g. creating kernels and the like). 
This is a singleton and the lifetime of this class is tied to the lifetime of the vscode session, hence there's no need to dispose anything explicitly here.

rebornix:
Since I'm not going to fix Export on Web and finish all file system related refactoring in one PR, I'll stop here and leave the remaining web/node FileSystem refactoring to next PR. The last few commits since @DonJayamanne 's review are all about introducing a `IFileSystem` instance for Web. As said above, `IFileSystem` was too fat and the interface itself has knowledge about NodeJs, so I made following changes to ensure that we have a good separation between common/web/node and still allow us to use NodeJs related fs when needed:

* `import { IFileSystem } from 'platform/common/platform/types';` is the common file system instance, it doesn't have access to Node. If you don't explicitly call Node `fs`, use this one.
*  Nodejs specific fs API will only exist on `import { FileSystem } from 'platform/common/platform/fileSystem.node`, note that most native fs dependent API are not declared on `IFileSystem`. This is intentional that we know we are using Nodejs API.
* There are a few NodeJs dependent APIs left in `IFileSystem` and current Web fileSystem will throw if they are invoked by mistake. They should be removed from the interface in follow up PRs.

src/platform/export/exportDialog.ts
```diff
@@ -1,5 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
IanMatthewHuff, line 1: Nice catch. Wonder if we need a rule for this. Not sure when it actually legally comes into play, but we do need this.

IanMatthewHuff (approved):
👍 . Looks good. The FileSystem change naming read a bit funny to me (with the base class having the same name as the node class) but from the comments I understand it.

src/kernels/ipywidgets-message-coordination/scriptUriConverter.node.ts
```diff
@@ -3,6 +3,7 @@ import { ILocalResourceUriConverter } from './types';
 import * as path from '../../platform/vscode-path/path';
 import { inject, injectable } from 'inversify';
 import { IFileSystem } from '../../platform/common/platform/types.node';
+import { FileSystem } from '../../platform/common/platform/fileSystem.node';
```
rchiodo, line 6: This change seems unnecessary? Why does it need the concrete type?

src/kernels/jupyter/import-export/jupyterImporter.node.ts
```diff
@@ -41,7 +42,7 @@ export class JupyterImporter implements INotebookImporter {
     private template6Promise?: Promise<string | undefined>;
 
     constructor(
-        @inject(IFileSystem) private fs: IFileSystem,
+        @inject(IFileSystem) private fs: FileSystem,
```
rchiodo, line 45: Same with this. Not sure why you'd want the hard dependency on the real type.

src/platform/export/exportDialog.ts
rchiodo: I'm actually not sure we need this. Other microsoft repos don't always have this either, but we could write an eslint rule for this if we wanted to enforce.

src/webviews/extension-side/plotting/plotViewer.node.ts
```diff
@@ -36,7 +37,7 @@ export class PlotViewer extends WebviewPanelHost<IPlotViewerMapping> implements
         @inject(IThemeFinder) themeFinder: IThemeFinder,
         @inject(IWorkspaceService) workspaceService: IWorkspaceService,
         @inject(IApplicationShell) private applicationShell: IApplicationShell,
-        @inject(IFileSystem) private fs: IFileSystem
+        @inject(IFileSystem) private fs: FileSystem
```
rchiodo, line 40: I think if there's a node specific IFileSystem, there should be a node specific interface. Passing the type here might make it compile, but inversify isn't going to enforce this. If we want it to be enforced in the container, we should have something like IFileSystemNode or inject the concrete type ```@inject(FileSystem)```

rebornix: @rchiodo I used the trick of using node specific `FileSystem` concrete types in *.node.ts` files to avoid the node specific fs APIs bleeding to the interface. 

Not exactly sure how inject work, are these two identical? 

```
@inject(IFileSystem) private fs: FileSystem
```

```
@inject(FileSystem) private fs: FileSystem
```

If they are identical, I will make the change and then introduce `IFileSystemNode` in next PR.

rebornix: Also I'm not sure which approach is the best yet, since we are doing all the refactoring and we will be an unhappy state for a while. I like the idea of `IFileSystemNode` but not sure if making that change would modify 50ish files and quite a few of them will need to inject both `IFileSystemNode` and `IFileSystem`, not a great code smell either.

src/platform/export/exportDialog.ts
rebornix: I think it's necessary, especially that when you copy some files from the repo to another (i.e., copying some helper files from VS Code to a different repo), you would want the copyright to come with it. We enforce this in our eslint rule https://github.com/microsoft/vscode/blob/0f648cd7dbafdbbbefda0f23c61c6c48a8d77619/.eslintrc.json#L94-L103

src/webviews/extension-side/plotting/plotViewer.node.ts
rchiodo: IFileSystemNode could be a superset of IFileSystem. The node files would inject the node specific one (well if they need it). It would be those you modified in this PR

And an object can implement more than one interface in the container. See examples of ```addBinding``` in some of the serviceRegistry files.

rchiodo: If you're injecting FileSystem, then you need to register FileSystem as the thing in the container (not IFileSystem) or there will be binding errors on startup.

I think you could introduce IFileSystemNode without a lot of changes though.

src/kernels/jupyter/jupyterKernelService.node.ts
```diff
@@ -51,11 +51,11 @@ import { IJupyterKernelService } from './types';
 export class JupyterKernelService implements IJupyterKernelService {
     constructor(
         @inject(IKernelDependencyService) private readonly kernelDependencyService: IKernelDependencyService,
-        @inject(IFileSystem) private readonly fs: IFileSystem,
+        @inject(IFileSystemNode) private readonly fs: IFileSystemNode,
         @inject(IEnvironmentActivationService) private readonly activationHelper: IEnvironmentActivationService,
         @inject(IEnvironmentVariablesService) private readonly envVarsService: IEnvironmentVariablesService,
         @inject(JupyterPaths) private readonly jupyterPaths: JupyterPaths
-    ) {}
+    ) { }
```
rchiodo, line 58: I keep seeing this extra space. I wonder if the linting rules for VS code enforce this and ours do the opposite?

src/kernels/raw/finder/localKernelSpecFinderBase.node.ts
```diff
@@ -254,8 +254,7 @@ export async function loadKernelSpec(
     let kernelJson: ReadWrite<IJupyterKernelSpec>;
     try {
         traceVerbose(
-            `Loading kernelspec from ${getDisplayPath(specPath)} for ${
-                interpreter?.uri ? getDisplayPath(interpreter.uri) : ''
+            `Loading kernelspec from ${getDisplayPath(specPath)} for ${interpreter?.uri ? getDisplayPath(interpreter.uri) : ''
```
rchiodo, line 257: Yeah I'm guessing you don't have prettier formatting on save (or it's formatting differently). The linter step in the build is likely going to barf on this one.

src/notebooks/outputs/plotSaveHandler.node.ts
```diff
@@ -20,9 +21,9 @@ const imageExtensionForMimeType: Record<string, string> = {
 export class PlotSaveHandler {
     constructor(
         @inject(IApplicationShell) private readonly shell: IApplicationShell,
-        @inject(IFileSystem) private readonly fs: IFileSystem,
+        @inject(IFileSystem) private readonly fs: FileSystem,
```
rchiodo, line 24: Looks like you missed a spot here.

src/platform/common/process/condaService.node.ts
```diff
@@ -33,7 +34,7 @@ export class CondaService {
     constructor(
         @inject(IPythonApiProvider) private readonly pythonApi: IPythonApiProvider,
         @inject(IMemento) @named(GLOBAL_MEMENTO) private readonly globalState: Memento,
-        @inject(IFileSystem) private readonly fs: IFileSystem,
+        @inject(IFileSystem) private readonly fs: FileSystem,
```
rchiodo, line 37: Another missed spot

src/test/datascience/export/exportToPython.vscode.test.ts
```diff
@@ -22,7 +22,7 @@ suite('DataScience - Export Python', function () {
     teardown(closeActiveWindows);
     suiteTeardown(closeActiveWindows);
     test('Export To Python', async () => {
-        const fileSystem = api.serviceContainer.get<IFileSystem>(IFileSystem);
+        const fileSystem = api.serviceContainer.get<IFileSystemNode>(IFileSystemNode);
```
rchiodo, line 25: Maybe do this as a separate PR, but it would be nice if this test was made to run in web too.

src/webviews/extension-side/plotting/plotViewer.node.ts
rchiodo: Looks like another spot where you missed putting it back

rchiodo (changes requested):
Looks good, just a few missed spots for the conversion.

src/kernels/jupyter/jupyterKernelService.node.ts
rebornix: I was using Prettier extension and trigger it on Save. Let me see if the command line is any different.

src/test/datascience/export/exportToPython.vscode.test.ts
rebornix: Yep, will look into this in next PR>

src/kernels/raw/finder/localKernelSpecFinderBase.node.ts
rebornix: Good catch, I have a setting which enforces ts formatter thus it behaves differently.

src/kernels/jupyter/jupyterKernelService.node.ts
rebornix: Fixed.

rchiodo (approved):

rebornix:
@DonJayamanne I have already FileSystem for Web and separate the common and node FileSystem module, please help take a look.

amunger (approved):

DonJayamanne (approved):

PR #10025, opened 2022-05-13
Dev/aamunger/perf test2

amunger:


amunger:
looks like I broke the actions configuration

.github/workflows/build-test.yml
```diff
@@ -815,6 +815,69 @@ jobs:
           path: .nyc_output
           retention-days: 1
 
+perf-tests:
```
amunger, line 818: copied from smoke-tests

src/test/perfTest.node.ts
```diff
@@ -0,0 +1,62 @@
...
+            glob(`**/*perf.test.js`, (error, files) => {
+                if (error) {
+                    return reject(error);
+                }
+                resolve(files);
+            });
+        });
+
+        // warm up with a basic notebook operation before running the tests
+        await this.launchTest('notebookCellExecution.perf.test', true);
```
IanMatthewHuff, line 22: One thing that I might be missing on this. Per earlier comments and the comments in the tracker it looked like each test is measuring a full test run, including activation. If so what's the warmup needed for?

src/test/standardTest.node.ts
```diff
@@ -16,15 +16,14 @@ const workspacePath = process.env.CODE_TESTS_WORKSPACE
...
-const isRunningSmokeTests = process.env.TEST_FILES_SUFFIX === 'smoke.test*';
 const isRunningVSCodeTests = process.env.TEST_FILES_SUFFIX?.includes('vscode.test');
 setTestExecution(true);
 
 function requiresPythonExtensionToBeInstalled() {
     if (process.env.VSC_JUPYTER_CI_TEST_DO_NOT_INSTALL_PYTHON_EXT) {
         return;
     }
-    return isRunningVSCodeTests || isRunningSmokeTests;
+    return isRunningVSCodeTests || IS_SMOKE_TEST || IS_PERF_TEST;
```
rchiodo, line 26: I believe IS_SMOKE_TEST and IS_PERF_TEST are functions now. I should have changed the case to make that more obvious.

src/test/performance/performanceTracker.node.ts
```diff
@@ -0,0 +1,60 @@
...
+import { StopWatch } from '../../platform/common/utils/stopWatch';
+import { IS_CI_SERVER } from '../ciConstants.node';
+import { sleep } from '../core';
+
+export class PerformanceTracker implements IDisposable {
+    stopWatch = new StopWatch();
+    durations: Record<string, number> = {};
+    checkpointCount = 0;
+    telemetryReporter: TelemetryReporter | undefined;
+    telemetryEnabled = IS_CI_SERVER && !process.env.VSC_JUPYTER_WARMUP;
```
rchiodo, line 14: Same thing here, IS_CI_SERVER should be a function call (it will always return true I think). Unless this is another constant that means the same thing.

The reason it's a function is because in web we can't check process.env, so it has to be set outside.

src/test/index.node.ts
```diff
@@ -187,11 +189,14 @@ export async function run(): Promise<void> {
     // Setup test files that need to be run.
     testFiles.forEach((file) => mocha.addFile(path.join(testsRoot, file)));
 
-    /* eslint-disable no-console */
-    console.time('Time taken to activate the extension');
-    console.log('Starting & waiting for Jupyter extension to activate');
-    await activateExtensionScript();
-    console.timeEnd('Time taken to activate the extension');
+    // for performance tests, extension activation is part of the test run
+    if (!IS_PERF_TEST) {
```
rchiodo, line 193: Same thing here too. IS_PERF_TEST is a function:
https://github.com/microsoft/vscode-jupyter/blob/2b726d6a5b2ab045bcf32e4691baebc820b29fc6/src/test/constants.ts#L30

rchiodo (approved):

src/test/perfTest.node.ts
amunger: I figured opening vscode for the first time after downloading and installing might have perf implications, and adding more tests would mean that whichever test goes first gets the perf hit.

src/test/performance/performanceTracker.node.ts
amunger: looks like IS_CI_SERVER is a const in ciConstants.node.ts

IanMatthewHuff (approved):

Issue #9665, opened 2022-04-11
DataFrame working in web extension


rchiodo:
Only difficulty might be in getting the dataframe scripts to run on the remote jupyter machine.

sadasant:
After meeting with @DonJayamanne and @rchiodo , we now have a path forward regarding this issue.

Our priority is to solve this issue, so first we will make sure to enable the Data Frame viewer on the web:

- By changing the `IDataViewerDependencyService` to receive an `IKernel` besides the `PythonEnvironment` that it receives today.
- By adding a universal version of the `dataViewerDependencyService` that would only work if `IKernel` is provided and throw otherwise an error saying `Installation of "pandas" not supported while debugging. Please ensure you have "pandas" installed.`
- By keeping the current node-specific version of the `dataViewerDependencyService` that would extract the interpreter from the `IKernel` if the `IKernel` is provided.
- Besides ensuring the `dataViewerDependencyService` is registered on the web service registry, and removing the conditional that disables the Data Frame on the web.

This approach is a compromise of a more valuable set of changes, which will come with other PRs. You can see the current plan at: https://github.com/microsoft/vscode-jupyter/issues/10638

PR #10029, opened 2022-05-14
Adding the DebuggingManager to src/platform/serviceRegistry.web.ts

sadasant:
Thanks to the recent PR by @rebornix #10016 and my studies on some branches ([1](https://github.com/microsoft/vscode-jupyter/compare/main...sadasant:fix9665), [2](https://github.com/microsoft/vscode-jupyter/compare/main...sadasant:ian-pairprogramming)), I believe I can start working towards #9665 by ensuring the DebuggingManager runs as part of `serviceRegistry.web.ts`.

The migration is fairly simple, since none of the internals depend on Node.js as far as I could tell.
The changes in this PR can be tested by:
- The web compilation (which passes).
- Debugging via `Extension (web)` and confirming that anything runs on a `.ipynb` file (which does work).
- The existing CI tests.

If this makes sense, this PR will be the first on a series of PRs breaking down #9665 into reasonable chunks of code.


sadasant:
I'm getting:

```
/home/runner/work/vscode-jupyter/vscode-jupyter/src/platform/debugger/jupyter/kernelDebugAdapter.ts
39:29  error  Unexpected path "../../common/platform/types.node" imported in restricted zone. Importing node modules into non node files is not allowed  import/no-restricted-paths
```

However, `kernelDebugAdapter.ts` only depends on `platform/types.node.ts` for `IFileSystem`, which now exists on `platform/types.ts`, so it should be easy to fix.

I'll check eslint before pushing.

sadasant:
I have two more commits here: https://github.com/microsoft/vscode-jupyter/compare/main...sadasant:web-DebuggingManager-extra?expand=1

There I:
1. Add [DebugService on src/platform/common/serviceRegistry.web.ts](https://github.com/microsoft/vscode-jupyter/commit/e4fc5a0714a1afad687f10432c5cb497fcd80054).
2. Add [DataViewerFactory and NotebookWatcher src/platform/serviceRegistry.web.ts](https://github.com/microsoft/vscode-jupyter/commit/21934474d12ddedfcfe8d6cf170c21fc9f0fc9be).

Those commits seem innocuous. Should I add them to this PR?
Please let me know if I'm missing something.

sadasant:
I tried adding `registerInstallerTypes(serviceManager)` to `src/kernels/serviceRegistry.web.ts`, but this change quickly expands in scope because this service tries to install several things in the host machine, which is a problem in itself bigger than this PR.

I also tried adding `PYTHON_VARIABLES_REQUESTER`, `MULTIPLEXING_DEBUGSERVICE`, `RUN_BY_LINE_DEBUGSERVICE`, `JupyterVariableDataProviderFactory`, `ALL_VARIABLES`, `KERNEL_VARIABLES` and `DEBUGGER_VARIABLES` to `src/kernels/serviceRegistry.web.ts` but this quickly goes out of proportion. For starters, it leads to issues with `EXTENSION_ROOT_DIR` in `src/platform/common/constants.node.ts`, which should be fixed by using `context.extensionUri`, but `context` has to be routed through `IExtensionContext`, changing how the related services work significantly with `src/platform/common/constants*`.

I'd rather add these changes as part of a separate PR.

IanMatthewHuff:
@sadasant Ahh, sorry, didn't pick this up before my PR. Looks like you / me / Don were all in the same area. I believe this is now covered by my PR that went in. Do you have a PR up for the variable requestor work that I could take a peek at?

PR #10035, opened 2022-05-15
Exclude package.json in vscode dir

DonJayamanne:
when looking for package.json we could end up picking up vscodes package.json which is wrong (at which point we don't know the calling extension).

### DIFF
--- a/src/platform/common/application/extensions.node.ts
+++ b/src/platform/common/application/extensions.node.ts
@@ -40,7 +40,8 @@ export class Extensions implements IExtensions {
                         return result[1];
                     }
                 })
-                .filter((item) => item && !item.toLowerCase().startsWith(jupyterExtRoot)) as string[];
+                .filter((item) => item && !item.toLowerCase().startsWith(jupyterExtRoot))
+                .filter((item) => this.all.some((ext) => item!.includes(ext.extensionUri.path))) as string[];
             stacktrace.parse(new Error('Ex')).forEach((item) => {
                 const fileName = item.getFileName();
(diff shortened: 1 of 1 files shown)


src/platform/common/application/extensions.node.ts
```diff
@@ -40,7 +40,8 @@ export class Extensions implements IExtensions {
                         return result[1];
                     }
                 })
-                .filter((item) => item && !item.toLowerCase().startsWith(jupyterExtRoot)) as string[];
+                .filter((item) => item && !item.toLowerCase().startsWith(jupyterExtRoot))
+                .filter((item) => this.all.some((ext) => item!.includes(ext.extensionUri.path))) as string[];
```
DonJayamanne, line 44: This ensures we only look for frames that contain extension paths

sadasant (approved):
✅

src/platform/common/application/extensions.node.ts
```diff
@@ -41,7 +41,11 @@ export class Extensions implements IExtensions {
                     }
                 })
                 .filter((item) => item && !item.toLowerCase().startsWith(jupyterExtRoot))
-                .filter((item) => this.all.some((ext) => item!.includes(ext.extensionUri.path))) as string[];
+                .filter((item) =>
+                    this.all.some(
+                        (ext) => item!.includes(ext.extensionUri.path) || item!.includes(ext.extensionUri.fsPath)
```
sadasant, line 46: (as I want to learn)
Oh! Why would one be set and the other wouldn’t?

rchiodo (approved):

PR #10036, opened 2022-05-16
Clean up fileSystem imports

rebornix:
Follow up of the file system refactor. Removing unnecessary `IFileSystem` export from `filesystem.node`

Note that with this change, quite a few files (i.e., `src/interactive-window/interactiveWindow.node.ts`) don't import `*.node.ts` files anymore, but I didn't rename them. This is because they still use `uri.fsPath`, which should be disallowed on Web. To make these modules common/web, we would need to go through their use of `fsPath` and see if they can be replaced by `path` or anything similar.


DonJayamanne (commented):
Looks good, and the comments make perfect sense.

I'd prefer there to be a really simple solution such as using `path` everywhere & prohibiting the use of `fsPath` via an eslint rule, unless its absolutely necessary, which can be dealt on a case by case basis. This way things would work & compile everywhere.

I personally don't like the approach of having `fsPath` in node & something else in web, basically this doesn't explain the why.

Optionally we could use `toString` in place of `fsPath` & `path`. To make it more obvious that we're only interested in the Uri itself (e.g. for use in keys of maps or the like)

rebornix:
@DonJayamanne good point, I prefer using `path` or some sort of `uri.toString()` too, relying on `fsPath` seems like a debt in the long term. Let's discuss in today's sync.

rchiodo (approved):

Issue #10037, opened 2022-05-16
Support debugging of Interactive Window using the Jupyter debugger protocol


DonJayamanne:
Basically implement debugging of IW using the same code used to enable debugging of notebooks.
* [x] https://github.com/microsoft/vscode-jupyter/issues/10037
* [ ] https://github.com/microsoft/vscode-jupyter/issues/10106
* [x] https://github.com/microsoft/vscode-jupyter/issues/10107

roblourens:
I think this old issue was related to this task, maybe it's not relevant anymore: https://github.com/microsoft/vscode-jupyter/issues/7057

DonJayamanne:
Closed #7057 with comments

rchiodo:
This isn't actually working. There's a big problem that I see with the cell to file mapping. See this issue:
https://github.com/microsoft/vscode-jupyter/issues/10258

DonJayamanne:
Curious, how are the tests passing for remote debugging?

rchiodo:
I don't know. Didn't realize we had remote debugging tests. Or maybe they just test the simplest form.

DonJayamanne:
Closing this issue as it has been completed.

PR #10038, opened 2022-05-16
Move notebook debugger into notebook folder

DonJayamanne:
Part of #10037

The plan is to have the following folder structure:
src->platform->debugger->jupyter->notebook (specific to notebook debugging)
src->platform->debugger->jupyter->interactiveWindow (specific to IW debugging)
src->platform->debugger->jupyter->common (Stuff thats common to both will eventually move into this folder)


sadasant (approved):
I love it! Short & well scoped ✅

PR #10040, opened 2022-05-16
Enable Web Debugging

### LINKED ISSUE #9973: Support notebook debugging in the Jupyter web extension
Support notebook debugging in the Jupyter web extension.

IanMatthewHuff:
Fixes #9973


src/platform/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -275,6 +264,35 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter, ID
...
+            // Create our python string of file names
+            const fileListString = fileValues
+                .map((filePath) => {
+                    return '"' + filePath + '"';
+                })
+                .join(',');
+
+            // Insert into our delete snippet
+            const deleteFilesCode = `import os
+_VSCODE_fileList = [${fileListString}]
```
IanMatthewHuff, line 281: This bit works I believe since the file names are coming from jupyter creating them. So I'm not worried about having a " or something like that in the file string.

package.json
```diff
@@ -291,7 +291,7 @@
                 "title": "%jupyter.command.jupyter.debug.title%",
                 "icon": "$(bug)",
                 "category": "Jupyter",
-                "enablement": "notebookKernelCount > 0 && !jupyter.notebookeditor.debuggingInProgress && !jupyter.notebookeditor.runByLineInProgress && !jupyter.webExtension"
+                "enablement": "notebookKernelCount > 0 && !jupyter.notebookeditor.debuggingInProgress && !jupyter.notebookeditor.runByLineInProgress"
```
sadasant, line 294: nice!

src/platform/debugger/jupyter/kernelDebugAdapter.ts
sadasant: Is there a way to log errors just in case they ever happen?

sadasant: Disregard! I just read the `traceErrors: true`

src/platform/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -275,6 +264,35 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter, ID
         return undefined;
     }
 
+    // Use our jupyter session to delete all the cells
+    private async deleteDumpCells() {
+        const fileValues = [...this.cellToFile.values()];
```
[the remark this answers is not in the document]
IanMatthewHuff: cellToFile is a Map so values / keys will be an IterableIterator. Just converted to an array with spread syntax so that I could use .map on it. I think Array.from could also be used, but there isn't a perf difference that I'm aware of.

src/platform/debugger/jupyter/kernelDebugAdapter.ts
IanMatthewHuff: Yeah, the trace should at least put something into our logging (but nothing user visible).

sadasant: Thanks!

IanMatthewHuff:
I think @DonJayamanne just hit some of the same files. Will pull and update (see how git handles two renames / two moves on the same file)

src/platform/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -275,6 +264,35 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter, ID
...
+            const deleteFilesCode = `import os
+_VSCODE_fileList = [${fileListString}]
+for file in _VSCODE_fileList:
+    try:
+        os.remove(file)
+    except:
+        pass
+del _VSCODE_fileList`;
+
+            return executeSilently(this.jupyterSession, deleteFilesCode, {
```
rebornix, line 289: Any idea why the execution is set to `silent: false`, is this intentional https://github.com/microsoft/vscode-jupyter/blob/bf8e4adebaa4da1fd7923eed536e9ce0ffe6f788/src/kernels/helpers.ts#L1452 ?

IanMatthewHuff: Yeah, as I recall the name is bad. Silent also swallows all output and we want to see if there are errors coming in on stdout. It's actually store_history that makes it silent in terms of hiding the execution from execution count.

IanMatthewHuff: Jupyter has doc on the execute params here which detail it a bit more: https://jupyter-client.readthedocs.io/en/latest/messaging.html#execute

rchiodo (approved):

src/platform/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -275,6 +264,35 @@ export class KernelDebugAdapter implements DebugAdapter, IKernelDebugAdapter, ID
         return undefined;
     }
 
+    // Use our jupyter session to delete all the cells
+    private async deleteDumpCells() {
```
rebornix, line 268: I wonder if this is identical to `ContentsManager#delete` considering that the dummy file is created by the Jupyter itself. In Export Web support, I found that I need to create temp files and delete them too so I'm experimenting with `ContentsManager#newUntitled` and `ContentsManager#delete` to see if they work as expected.

src/platform/debugger/jupyter/kernelDebugAdapter.ts
rebornix: @IanMatthewHuff thanks for the detailed explanation, you are right `store_history` is the key here and our `executeSilently` here means skipping history and execution count.

src/platform/debugger/jupyter/kernelDebugAdapter.ts
IanMatthewHuff: Part of the diff here is that this code will be working in both the web and the node version. The ContentsManager is only available when connecting to Jupyter, so that makes sense to use when working with a .web only scenario. I could consider a .web kernelDebugAdapter and use the ContentsManager delete there, but I think that I'd only go there if this cell execution option which covers both scenarios was proving to have issues.

IanMatthewHuff: Might also be a mild diff here in that this file is actually created by ipykernel and not Jupyter:
https://github.com/ipython/ipykernel/blob/51a613d501a86073ea1cdbd8023a168646644c6a/ipykernel/compiler.py#L83
I would assume Jupyter ContentsManager would have access to the same temp location though, so I'm guessing it could be used for the delete.

PR #10045, opened 2022-05-16
Updates to events exposed by kernels and API

DonJayamanne:


src/kernels/raw/session/rawKernel.node.ts
```diff
@@ -22,24 +23,15 @@ input request, translating them, sending them to an IPython kernel over ZMQ, the
 */
 export class RawKernel implements Kernel.IKernelConnection {
     public socket: IKernelSocket & IDisposable;
-    public get statusChanged() {
-        return this.realKernel.statusChanged as any; // NOSONAR
```
DonJayamanne, line 26: These were all incorrect, the first argument of the signal event is the sender, 
With this current code, the sender is `realKernel` instead of `this`.
& that causes issues as the Kernel.IKernelConnection the user expects would point to `this`, but infact it ponts to some other object (whith happens to be `this.realKernel`).

Basically its wrong

src/kernels/raw/session/rawSession.node.ts
```diff
@@ -62,7 +65,10 @@ export class RawSession implements ISessionWithSocket {
         this._kernelChanged = new signaling.Signal<this, Session.ISessionConnection.IKernelChangedArgs>(this);
         this._ioPubMessage = new signaling.Signal<this, KernelMessage.IIOPubMessage>(this);
         this._terminated = new signaling.Signal<this, void>(this);
+        this._anyMessage = new signaling.Signal<this, Kernel.IAnyMessageArgs>(this);
+        this._unhandledMessage = new signaling.Signal<this, KernelMessage.IMessage>(this);
```
DonJayamanne, line 69: We were doing things correctly for some events, but not all, modified to ensure we do the same for all events, as this is required for things to work correctly.

src/platform/api/extension.d.ts
```diff
@@ -263,7 +263,7 @@ export interface IExportedKernelService {
      * If `uri` is undefined, then the kernel is not associated with any resource. I.e its currently not associated with any notebook in Jupyter extension.
      * If `uri` is undefined, then the kernel is associated with the resource identified by the Uri.
      */
-    getActiveKernels(): Promise<{ metadata: KernelConnectionMetadata; uri: Uri | undefined }[]>;
+    getActiveKernels(): { metadata: KernelConnectionMetadata; uri: Uri | undefined }[];
```
DonJayamanne, line 266: This doesn't need to be a Promise at all.

rchiodo (approved):

IanMatthewHuff (approved):

PR #10050, opened 2022-05-16
FetchError is undefined in the web, and capturing errors by .name

sadasant:
Since `FetchError` is undefined in the web, `isErrorType(error, FetchError)` on `populateTelemetryWithErrorInfo` causes an error: `Uncaught (in promise) TypeError: Right-hand side of 'instanceof" is not an object`.

This PR ensures that we're able to catch errors of the type `FetchError` in the web.

It also proposes a change to `isErrorType` to also accept error names in the second parameter.

### DIFF
--- a/src/platform/errors/index.ts
+++ b/src/platform/errors/index.ts
@@ -13,7 +13,8 @@ export function populateTelemetryWithErrorInfo(props: Partial<TelemetryErrorProp
     props.failed = true;
     // Don't blow away what we already have.
     props.failureCategory = props.failureCategory || getErrorCategory(error);
-    if (props.failureCategory === 'unknown' && isErrorType(error, FetchError)) {
+    // FetchError is undefined on Web.
+    if (props.failureCategory === 'unknown' && isErrorType(error, FetchError.name || 'FetchError')) {
         props.failureCategory = 'fetcherror';
     }
     props.stackTrace = serializeStackTrace(error);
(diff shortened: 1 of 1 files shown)


rchiodo (approved):

IanMatthewHuff (approved):

src/platform/errors/index.ts
```diff
@@ -13,9 +13,15 @@ export function populateTelemetryWithErrorInfo(props: Partial<TelemetryErrorProp
     props.failed = true;
     // Don't blow away what we already have.
     props.failureCategory = props.failureCategory || getErrorCategory(error);
-    if (props.failureCategory === 'unknown' && isErrorType(error, FetchError)) {
+
+    // In browsers, FetchError is undefined and the error we receive is a `TypeError` with message `Failed to fetch`.
+    const isBrowserFetchError = !FetchError && error?.name === 'TypeError' && error?.message === 'Failed to fetch';
+    const errorType = isBrowserFetchError ? 'TypeError' : FetchError;
+
+    if (props.failureCategory === 'unknown' && isErrorType(error, errorType)) {
```
DonJayamanne, line 21: I don't think we should override isErrorType to support just one edge case.
I'd prefer to special case this and just check the types/strings in here for node-fetch.
Basically the second argument in `isErrorType` will almost never be used in other places, if we do find other places then we can make it more generic as you have, else I think its just going to complicate it by supporting a string.

DonJayamanne (approved):

src/platform/errors/index.ts
```diff
@@ -81,6 +87,7 @@ function getCallSite(frame: stackTrace.StackFrame) {
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type Constructor<T> = { new (...args: any[]): T };
+
```
sadasant, line 90: @DonJayamanne I saw your comment and I agree! I have undone the changes to isErrorType 🙂

sadasant: On the Azure SDK for JS, we used to only check errors based on `error.name`. I believe there was an issue with instanceof. It might have been when possibly referencing to different copies of the same class, which is something possible in Node.js (specially when two dependencies pull in different versions of an underlying common dependency). I don't think this (whatever might have been the root issue that stuck with me) is a problem in this repository! Also: The fewer changes, the better.

PR #10054, opened 2022-05-17
Add Data Wrangler renderer entrypoint

pwang347:
Fixes #


src/notebooks/outputs/dataWranglerHtmlRendererComms.node.ts
```diff
@@ -0,0 +1,283 @@
...
+    public async activate(): Promise<void> {
+        const dataWranglerExtension =
+            vscode.extensions.getExtension<IDataWranglerExtensionAPI>('ms-toolsai.data-wrangler');
+
+        // do nothing if there is no data wrangler extension
+        if (!dataWranglerExtension) {
+            return;
+        }
+
+        const kernelService = await this.kernelServiceFactory.getServiceInternal();
```
rchiodo, line 93: This seems unnecessary? Is this just to guarantee the data wrangler can talk to the kernel service? 

The kernelProvider is the internal equivalent. It's always there.

src/platform/api/kernelApi.ts
```diff
@@ -64,6 +64,22 @@ export class JupyterKernelServiceFactory implements IExportedKernelServiceFactor
         this.extensionApi.set(accessInfo.extensionId, service);
         return service;
     }
+
+    /**
+     * This allows us to bypass the extension ID check and to call the service from the Jupyter extension.
+     */
+    public async getServiceInternal() {
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
rchiodo, line 72: I think you can skip adding this.

src/telemetry/constants.ts
```diff
@@ -15,7 +15,11 @@ export enum EventName {
 
     OPEN_DATAVIEWER_FROM_VARIABLE_WINDOW_REQUEST = 'OPEN_DATAVIEWER_FROM_VARIABLE_WINDOW_REQUEST_EX',
     OPEN_DATAVIEWER_FROM_VARIABLE_WINDOW_ERROR = 'OPEN_DATAVIEWER_FROM_VARIABLE_WINDOW_ERROR_EX',
-    OPEN_DATAVIEWER_FROM_VARIABLE_WINDOW_SUCCESS = 'OPEN_DATAVIEWER_FROM_VARIABLE_WINDOW_SUCCESS_EX'
+    OPEN_DATAVIEWER_FROM_VARIABLE_WINDOW_SUCCESS = 'OPEN_DATAVIEWER_FROM_VARIABLE_WINDOW_SUCCESS_EX',
+
+    OPEN_DATAWRANGLER_FROM_OUTPUT_REQUEST = 'OPEN_DATAWRANGLER_FROM_OUTPUT_REQUEST_EX',
```
rchiodo, line 20: I don't see an entry for this new telemetry in the index.ts? THat's where you'd put comments about the telemetry and what each event is for. It makes the generated telemetry.md file have more information.

src/webviews/webview-side/data-wrangler-html-renderer/dataWranglerHtmlRenderer.tsx
```diff
@@ -0,0 +1,156 @@
...
+            type: "launchDataWrangler"
+        });
+    };
+
+    override render() {
+        const { getLocalizedStrings } = this.props;
+        const { isDataFrame, lostVariableName, htmlToRender } = this.state;
+        const locStrings = getLocalizedStrings();
+        return (
+            <div>
```
rchiodo, line 139: I'm a little worried about this. Maybe I'm misunderstanding though. After this renderer is registered, won't this div be around ALL html type output?

src/notebooks/outputs/dataWranglerHtmlRendererComms.node.ts
pwang347: I've actually been already using the `kernelProvider` for the variable detection part, I was mainly curious if we needed to wrap the kernel since it's technically being used by a separate extension (sounds like we don't).

src/telemetry/constants.ts
pwang347: good catch, didn't notice that file and tried modifying the MD file directly

src/webviews/webview-side/data-wrangler-html-renderer/dataWranglerHtmlRenderer.tsx
pwang347: that's correct, I can change it to not have the wrapper div around it if we're not rendering our button

typings/dataWrangler.d.ts
```diff
@@ -0,0 +1,260 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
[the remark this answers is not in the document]
pwang347: this file provides the extension API as well, which relies on those types - that said, I think we can probably simplify the interface quite a bit by just grabbing the actual kernel properties we use

typings/dataWrangler.d.ts
```diff
@@ -0,0 +1,260 @@
...
+     * Promise indicating whether all parts of the extension have completed loading or not.
+     * @type {Promise<void>}
+     * @memberof IDataWranglerExtensionAPI
+     */
+    ready: Promise<void>;
+
+    /**
+     * Launches the Data Wrangler using the given variable name.
+     */
+    launchDataWranglerUsingVariable: (
```
DonJayamanne, line 23: If this PR is to allow launching the DW variable explorer, then I think we can just add an API extension in our extension that allows any other extension to create entry points for launching variable viewer for any given variable.

That way we don't tightly couple DW with the Jupyter extension.

src/notebooks/outputs/dataWranglerHtmlRendererComms.node.ts
```diff
@@ -0,0 +1,283 @@
...
+        @inject(IExportedKernelServiceFactory) private readonly kernelServiceFactory: IExportedKernelServiceFactory,
+        @inject(IJupyterVariables)
+        @named(Identifiers.KERNEL_VARIABLES)
+        private readonly jupyterVariables: IJupyterVariables,
+        @inject(IKernelProvider) private readonly kernelProvider: IKernelProvider
+    ) {}
+
+    public async activate(): Promise<void> {
+        const dataWranglerExtension =
+            vscode.extensions.getExtension<IDataWranglerExtensionAPI>('ms-toolsai.data-wrangler');
```
DonJayamanne, line 86: This is a weird dependency structure.
Looks like DW depends on Jupyter and Jupyter depends on DW.
I think we can fix this if we just move DW ito a seprate extension.

rchiodo: DW is a separate extension, but the renderer is here (our renderers have priority in VS code, otherwise the DW renderer is never picked).

I agree this dependency structure is weird. DW should register something with jupyter and this renderer could look for that something. Otherwise we're circular.

rebornix: It's weird that Jupyter contributes an output renderer, which is always available to users, but only functional if Data Wrangler is installed. The contribution for this renderer should only be from the Data Wrangler extension.

> our renderers have priority in VS code

It seems we don't contribute a `text/html` renderer in Jupyter? https://github.com/microsoft/vscode-notebook-renderers/blob/main/package.json

typings/dataWrangler.d.ts
pwang347: discussed offline, this is for the output renderer

src/notebooks/outputs/dataWranglerHtmlRendererComms.node.ts
rchiodo: I think this workaround is fine for now (although I agree with Don it should go in the jupyter renderers extension). We can log an issue on vscode core to allow other extensions to have priority. Not sure why they don't.

pwang347: got it, so maybe for now I can move the renderer UI into the jupyter renderers extension and keep the handling code in DW itself? e.g. expose a message listener, similar to how Jupyter is interacting with some of the renderer webviews today?

rchiodo: That sounds good to me. Just to summarize:
- Renderer side code would move to jupyter renderer for now until we can resolve priority issues in VS code
- Extension side code would move into the DW extension. 
- Dependency on the DW extension would be avoided in the renderer because it would just listen for messages and just fire off another message when user clicked the button. The dependency would be the messages sent. 

@rebornix and @DonJayamanne what do you guys think?

rebornix: > Renderer side code would move to jupyter renderer for now until we can resolve priority issues in VS code

We had good discussions offline. The core would need to allow users to pick their preferred renderer for a mimetype (note that it's not the renderer extension declaring their priority).

rchiodo: I think that all depends upon 'how' the user picks. There's a discoverability problem at the moment. Will there be core changes to address the discoverability problem?

PR #10056, opened 2022-05-18
Delete connection files when shutting down kernel

DonJayamanne:
Found while working on IW debugger.
Basically when re-loading VS Code or shutting down VS Code, the number of connection files was just growing.

Found that the files never get deleted as the code in the dispose is too slow and the VS Codes FS API for deletion doesn't work when shutting down VS Code.


src/kernels/raw/launcher/kernelProcess.node.ts
```diff
@@ -265,7 +265,7 @@ export class KernelProcess implements IKernelProcess {
         }
     }
 
-    public async dispose(): Promise<void> {
+    public dispose(): void {
```
DonJayamanne, line 268: No async code here, hence no need to make this async.
The change results in connection files getting deleted when vscode is shutdown.

This could also mean that other parts of the code have similar issues and cleanup might not work as expected.

src/kernels/raw/launcher/kernelProcess.node.ts
```diff
@@ -277,8 +277,10 @@ export class KernelProcess implements IKernelProcess {
             this.exitEvent.fire({});
         });
         swallowExceptions(async () => {
-            if (this.connectionFile && (await this.fileSystem.localFileExists(this.connectionFile))) {
-                await this.fileSystem.deleteLocalFile(this.connectionFile);
+            if (this.connectionFile) {
+                await this.fileSystem
+                    .deleteLocalFile(this.connectionFile)
```
DonJayamanne, line 282: Checking if the file exists, is even slower, then disposing, we don't care if the file exists or not, just delete and ignore the error (much faster, hence runs to completion when vscode is shutting down), else file never gets deleted

src/kernels/raw/launcher/kernelProcess.node.ts
```diff
@@ -367,7 +369,10 @@ export class KernelProcess implements IKernelProcess {
     }
     private async createConnectionFile() {
         const runtimeDir = await this.jupyterPaths.getRuntimeDir();
-        const tempFile = await this.fileSystem.createTemporaryLocalFile({ fileExtension: '.json', prefix: 'kernel-' });
+        const tempFile = await this.fileSystem.createTemporaryLocalFile({
+            fileExtension: '.json',
+            prefix: 'kernel-v2-'
```
DonJayamanne, line 374: Added `v2` so if users report files aren't getting deleted we know whether its old or new stuff.
Also helps me figure out whether these are before or after this change

src/kernels/raw/session/rawSession.node.ts
```diff
@@ -105,9 +104,11 @@ export class RawSession implements ISessionWithSocket {
         this.isDisposing = true;
         if (!this.isDisposed) {
             this.exitHandler.dispose();
-            await this._kernel.shutdown();
+            await this._kernel
+                .shutdown()
+                .catch((ex) => traceWarning(`Failed to shutdown kernel, ${this.kernelConnectionMetadata.id}`, ex));
             this._kernel.dispose();
-            await this.kernelProcess.dispose().catch(noop);
+            this.kernelProcess.dispose();
```
DonJayamanne, line 111: No longer an async method, hence should be faster.

src/kernels/raw/session/rawSession.node.ts
```diff
@@ -105,9 +104,11 @@ export class RawSession implements ISessionWithSocket {
         this.isDisposing = true;
         if (!this.isDisposed) {
             this.exitHandler.dispose();
-            await this._kernel.shutdown();
+            await this._kernel
+                .shutdown()
+                .catch((ex) => traceWarning(`Failed to shutdown kernel, ${this.kernelConnectionMetadata.id}`, ex));
```
DonJayamanne, line 109: We can ignore the errors.

src/kernels/raw/types.ts
```diff
@@ -37,7 +37,7 @@ export interface IKernelConnection {
     kernel_name?: string;
 }
 
-export interface IKernelProcess extends IAsyncDisposable {
+export interface IKernelProcess extends IDisposable {
```
DonJayamanne, line 40: Sync disposable

src/platform/common/platform/fileSystem.node.ts
```diff
@@ -53,17 +53,7 @@ export class FileSystem extends FileSystemBase implements IFileSystemNode {
...
-        // are empty.  So we have to deal with this ourselves.
-        const files = await this.vscfs.readDirectory(uri);
-        if (files && files.length > 0) {
-            throw createDirNotEmptyError(dirname);
-        }
-        return this.vscfs.delete(uri, {
-            recursive: true,
-            useTrash: false
-        });
+        await new Promise((resolve) => fs.rm(dirname, { force: true, recursive: true }, resolve));
```
DonJayamanne, line 56: Over engineered. We don't need to use VS Code API just to delete a folder.
We can just use native API for this.
VSC API is useful only when dealing with multiple file systems (such as remots or somethign that a user has contributed, i.e. when using Uris, even when we can easily check if uri.scheme=file and then revert to using node API which would be much faster)

src/platform/common/platform/fileSystem.node.ts
```diff
@@ -80,6 +70,9 @@ export class FileSystem extends FileSystemBase implements IFileSystemNode {
     public async localDirectoryExists(dirname: string): Promise<boolean> {
         return this.localPathExists(dirname, FileType.Directory);
     }
+    public override async deleteLocalFile(path: string): Promise<void> {
+        await fs.unlink(path);
```
DonJayamanne, line 74: The base method relies on VS Code API, without this change the files never get deleted when VS Code is shutting down on reloading, meaning the VSC API is slow and as mentioned earlier this is unnecessary as we're dealing with local files, again using native node API is more efficient (proven by the fact that without this change things don't work as expected - when reloading vscode or shutting it down)

src/platform/common/platform/fileSystemUtils.node.ts
```diff
@@ -45,239 +39,6 @@ export function convertStat(old: fs.Stats, filetype: vscode.FileType): vscode.Fi
     };
 }
 
-//==========================================
-// "raw" filesystem
-
-// This is the parts of the vscode.workspace.fs API that we use here.
-// See: https://code.visualstudio.com/api/references/vscode-api#FileSystem
-// Note that we have used all the API functions *except* "rename()".
-interface IVSCodeFileSystemAPI {
```
DonJayamanne, line 54: None of this is used anymore

src/platform/common/platform/types.node.ts
```diff
@@ -9,62 +9,9 @@ import { IFileSystem, TemporaryFile } from './types';
 // filesystem operations
 
 export import FileType = vscode.FileType;
-export import FileStat = vscode.FileStat;
 export type ReadStream = fs.ReadStream;
 export type WriteStream = fs.WriteStream;
 
-// The low-level filesystem operations on which the extension depends.
-export interface IRawFileSystem {
```
DonJayamanne, line 17: None of this is used anymore

rchiodo (approved):

PR #10063, opened 2022-05-19
Create traceback formatter to remove dependency on CellHashProvider

DonJayamanne:
Part of #10037

The fact that cell hash provider knows how to format tracebacks should not be known by the kernel layer.
Thats an implmenetation detail.
I've split the formatter into two, one for IW & one for Notebooks (Ipynb).

This way, we can now have more formatters that will allow us to modify the error messages to include custom links or the like.
E.g. in the past we wanted to add custom messages & links based on different errors, now that we have a formatter we can do this without having to modify the kernel layer.

### DIFF
--- a/src/notebooks/outputs/tracebackFormatter.ts
+++ b/src/notebooks/outputs/tracebackFormatter.ts
@@ -1,0 +1,41 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { injectable } from 'inversify';
+import { NotebookCell } from 'vscode';
+import { ITracebackFormatter } from '../../kernels/types';
+import { getFilePath } from '../../platform/common/platform/fs-paths';
+import { DataScience } from '../../platform/common/utils/localize';
+import { JupyterNotebookView } from '../constants';
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+const LineNumberMatchRegex = /(;32m[ ->]*?)(\d+)(.*)/g;
+
+@injectable()
+export class NotebookTracebackFormatter implements ITracebackFormatter {
+    format(cell: NotebookCell, traceback: string[]): string[] {
+        if (cell.notebook.notebookType !== JupyterNotebookView) {
+            return traceback;
+        }
+        return traceback.map((line) => {
+            const inputMatch = /^Input.*?\[.*32mIn\s+\[(\d+).*?0;36m(.*?)\n.*/.exec(line);
+
+            // We have a match, replace source lines first
+            const afterLineReplace = line.replace(LineNumberMatchRegex, (_s, prefix, num, suffix) => {
+                const n = parseInt(num, 10);
+                return `${prefix}<a href='${cell.document.uri.toString()}?line=${n - 1}'>${n}</a>${suffix}`;
+            });
+
+            return !inputMatch
+                ? afterLineReplace
+                : // Then replace the input line with our uri for this cell
+                  afterLineReplace.replace(
+                      /.*?\n/,
+                      `\u001b[1;32m${DataScience.cellAtFormat().format(
+                          getFilePath(cell.document.uri),
+                          (cell.index + 1).toString()
+                      )}\u001b[0m in \u001b[0;36m${inputMatch[2]}\n`
+                  );
+        });
+    }
+}
--- a/src/interactive-window/editor-integration/cellHashProviderFactory.ts
+++ b/src/interactive-window/editor-integration/cellHashProviderFactory.ts
@@ -9,6 +9,7 @@ import { IKernel, IKernelProvider } from '../../kernels/types';
 import { CellHashProvider } from './cellhashprovider';
 import { ICellHashListener } from './types';
 import { IPlatformService } from '../../platform/common/platform/types';
+import { NotebookDocument } from 'vscode';
 
 @injectable()
 export class CellHashProviderFactory {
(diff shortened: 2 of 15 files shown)


src/notebooks/outputs/tracebackFormatter.ts
```diff
@@ -0,0 +1,41 @@
...
+const LineNumberMatchRegex = /(;32m[ ->]*?)(\d+)(.*)/g;
+
+@injectable()
+export class NotebookTracebackFormatter implements ITracebackFormatter {
+    format(cell: NotebookCell, traceback: string[]): string[] {
+        if (cell.notebook.notebookType !== JupyterNotebookView) {
+            return traceback;
+        }
+        return traceback.map((line) => {
+            const inputMatch = /^Input.*?\[.*32mIn\s+\[(\d+).*?0;36m(.*?)\n.*/.exec(line);
```
rchiodo, line 21: Why would notebook be different than interactive window? Tracebacks should be the same?

rchiodo: As in this 'Input' match should be in both. Although I thought it was in the cell hash provider before.

rchiodo: Oh yeah it is still there, but it seems like the notebook could be skipping some of the other checks for tracebacks. Like for IPython7 tracebacks. They don't have this form.

src/notebooks/outputs/tracebackFormatter.ts
```diff
@@ -0,0 +1,41 @@
...
+
+            // We have a match, replace source lines first
```
amunger, line 23: comment copied out of context, we haven't matched anything yet.

src/notebooks/outputs/tracebackFormatter.ts
rchiodo: I was thinking you could refactor this into a single traceback formatter that depends upon something that gives hashes for executed code (ICellHashProvider) and takes an optional notebookCell as an argument.

rchiodo: That new tracebackformatter would have all of the 'modifyTraceback' code currently in the cell hash provider.

DonJayamanne: Yes we can do that, however for notebook traceback formatting, 90% of the stuff thats in cell has provider.
Its super simple. I can however try move some of that into a seprate class/function and share that between the two. 

But i don't want to use the same class, else Notebook traceback formatter will end up carrying a lot of baggage which isn't required at all. 

Lets think of it this way, its sepration of concerns.
If lets say IW is never going to be supported in Web (hypothetical scenario), we cannot do this as we have a lot of other dependencies in this class, and we make the optional, etc. lots of work arounds. 

Also if you look at the existing code we have a very clear `if else` condition that explicitly handles notebook tracebacks (two very different code paths), and that works very well to have a seprate class than have it in the same class to handle everything.

DonJayamanne: > Like for IPython7 tracebacks. They don't have this form.

I tested the existing code, the extension doesn't handle IPython7 tracebacks when working in notebooks at all. If you test in stable, we don't do anything at all.
I.e. its already broken. Will file a bug for this.

Though if we wanted to fix this, then we'd need the cell hash generators again.

I guess we need a single traceback formatter in that case. Ok, I'll move this back into a single traceback formatter then.

rchiodo: I do think it's good we have a separate class for the traceback formatter (agree with separation of concerns). I've always wanted to refactor this to a separate class. 

Just worried we might be missing something in the super simplified notebook one. 

I think the traceback formatter doesn't need to know about notebooks or kernels though? Just needs the cellhashprovider and a pointer to the actual notebook cell when formatting. Wouldn't that be enough to find the correct hashes? Oh maybe not for pointers to other cells.

IanMatthewHuff (approved):
I know Don and Rich were chatting a bit still (put my two cents in). But the actual difference between the suggestions feels pretty small to me so I'm fine with approving and leaving up to Don.

DonJayamanne:
* CellHashStorageProvider
    - GetOrCreate (indexed by notebook)
    - Get (indexed by notebook)
    - clear/remove (indexed by notebook)

* Cell execution will use the `getInteractiveCellMetadata` to get the code to be executed

* KernelDebug Adapter will depend on CellhashStorageProvider
    - get all of the hashes and send this
    - 

* Rename cell has provider to cellHash generator
    - Uses the cell has storage provider
    - generate (will store in storage)


* Traceformatter will use Cell has storage provider

* Modify IW to dump all of the old cell hashes after we attach

* Add `willExecuteCell` hook to IKernel
    IW will add a hook and dump all of the hashes in `createExecutionPromise`
    The hashes can be retrieved from the storage provider.

* Modifiy `addNotebookCell` to generate the cell hash & store it into the nb cell metatata

DonJayamanne:
**Further revisions**

* CellGeneratedCodeStorageProvider
    - GetOrCreate (indexed by notebook)
    - Get (indexed by notebook)
    - clear/remove (indexed by notebook)

* Cell execution will use the `getInteractiveCellMetadata` to get the code to be executed
    * Revision - no need, just get from the `CellGeneratedCodeStorageProvider`

* KernelDebug Adapter will depend on CellhashStorageProvider
    - get all of the hashes and send this
    - 

* Rename cell hash provider to RunTimeCode generator
    - Uses the cell has storage provider
    - generate (will store in storage)


* Traceformatter will use Cell has storage provider

* Modify IW to dump all of the old cell hashes after we attach

* Modifiy `addNotebookCell` to generate the cell hash & store it into the nb cell metatata
    * No need to stort in metadata

* Add `willExecuteCell` hook to IKernel (WIP)
    IW will add a hook and dump all of the hashes in `createExecutionPromise`
    The hashes can be retrieved from the storage provider.

PR #10069, opened 2022-05-19
Export to html/pdf/python on Web when connected to remote server

### LINKED ISSUE #9996: Support Export in Web

rebornix:
Fixes #9996

This PR covers following changes:

* `src/platform/progress/decorator.ts` and `src/platform/progress/progressReporter.ts` are now common modules (no `node` api usage)
* `exportToHTML`, `exportToPython`, `exportToPdf`, `exportToPythonPlain` are common modules
* `exportBase` has two variations, `exportBase.node` and `exportBase.web`. Their major differences are:
  * how they create files `exportBase.web` doesn't create temp files
  * how they run `nbconvert`
    * `exportBase.node` run `nbconvert` in local process
    * `exportBase.web` run `nbconvert` through silent cell execution. It fetches the content through `contentsManager` and save to workspace fs.
 * `fileConverter` as the common module contains most logics of how we export files, with some exceptions in `node` env, which are overridden in `fileConverter.node`
 * Commands when clauses [updated](https://github.com/microsoft/vscode-jupyter/pull/10069/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L347-L371) to enable export commands when 
   * local env and workspace trusted
   * remote env, and workspace trusted, and document connected to a kernel
 * Tests are updated to adopt the changes


src/platform/export/exportToPDF.ts
```diff
@@ -0,0 +1,24 @@
+import { inject, injectable } from 'inversify';
+import { CancellationToken, NotebookDocument, Uri } from 'vscode';
+import { PythonEnvironment } from '../pythonEnvironments/info';
+import { ExportFormat, IExportBase, INbConvertExport } from './types';
+
+@injectable()
+export class ExportToPDF implements INbConvertExport {
```
rchiodo, line 7: It looks like all of your renames are not be detecting as renames? I usually force this with `git add -A` but not sure if you can do it after the fact.

rchiodo (approved):

src/platform/export/exportToPDF.ts
rebornix: didn't know I could do that, will try with next rename 💪

src/platform/export/exportBase.node.ts
```diff
@@ -1,46 +1,83 @@
...
         protected jupyterService: IJupyterSubCommandExecutionService,
         @inject(IFileSystemNode) protected readonly fs: IFileSystemNode,
-        @inject(INotebookImporter) protected readonly importer: INotebookImporter
+        @inject(IExportDialog) protected readonly filePicker: IExportDialog,
+        @inject(ExportUtil) protected readonly exportUtil: ExportUtil,
+        @inject(INotebookImporter) protected readonly importer: INotebookImporter,
+        @inject(ExportInterpreterFinder) private exportInterpreterFinder: ExportInterpreterFinder
     ) {}
 
     public async export(
```
amunger, line 30: is this method needed in these classes?

src/platform/export/exportToPythonPlain.ts
```diff
@@ -3,34 +3,47 @@
...
-import { IFileSystem } from '../common/platform/types';
+import { IFileSystem, IPlatformService } from '../common/platform/types';
 import { IConfigurationService } from '../common/types';
-import { IExport } from './types';
+import { ExportFormat, IExport, IExportDialog } from './types';
 
 // Handles exporting a NotebookDocument to python
-export class ExportToPythonPlainBase implements IExport {
+@injectable()
+export class ExportToPythonPlain implements IExport {
```
amunger, line 15: I'm confused at why this class is the only Exporter that implements IExport

src/platform/export/exportBase.web.ts
```diff
@@ -0,0 +1,185 @@
...
+import { IFileSystem } from '../common/platform/types';
+import { PythonEnvironment } from '../pythonEnvironments/info';
+import { ExportUtilBase } from './exportUtil';
+import { ExportFormat, IExportBase, IExportDialog, INbConvertExport } from './types';
+import { traceLog } from '../logging';
+import { reportAction } from '../progress/decorator';
+import { ReportableAction } from '../progress/types';
+
+@injectable()
+export class ExportBase implements INbConvertExport, IExportBase {
```
amunger, line 22: ExportHelper is probably a better name now since they aren't being extended anymore

IanMatthewHuff (approved):

src/platform/export/exportToPythonPlain.ts
rebornix: I'm not sure either, it's a special case introduced via https://github.com/microsoft/vscode-jupyter/commit/621acee2bc6e5e00c7d398342e1a03f2f6bec04c

amunger: I think it's just because it's the only export method that doesn't use nbConvert. Doesn't feel like a good reason to split the interface to me, but not something to worry about for this change.

src/platform/export/exportBase.node.ts
rebornix: I still put it there since we register it as `serviceManager.addSingleton<INbConvertExport>(INbConvertExport, ExportBase, 'Export Base');`, not sure why it's needed there. I can send a pr later to have it fixed.

PR #10074, opened 2022-05-19
Refactor cell hash provider

DonJayamanne:


src/interactive-window/editor-integration/cellHashProviderFactory.ts
```diff
@@ -43,13 +45,18 @@ export class CellHashProviderFactory {
         if (existing) {
             return existing;
         }
+        const nb = getAssociatedNotebookDocument(kernel);
+        if (!nb) {
+            throw new Error('No notebook associated with the kernel');
```
DonJayamanne, line 50: This should never happen, we ensure we never call this code for kernels created by anything other than Jupyter extesnsion.

src/notebooks/outputs/tracebackFormatter.ts
```diff
@@ -0,0 +1,17 @@
...
+import { injectable } from 'inversify';
+import { NotebookCell } from 'vscode';
+import { ITracebackFormatter } from '../../kernels/types';
+import { JupyterNotebookView } from '../constants';
+
+@injectable()
+export class NotebookTracebackFormatter implements ITracebackFormatter {
+    public format(cell: NotebookCell, traceback: string[]): string[] {
+        if (cell.notebook.notebookType !== JupyterNotebookView) {
+            return traceback;
```
DonJayamanne, line 13: @rchiodo I found that the code for IW and Notebooks share very little similarities.
The only similarity is the regex stuff, which I plan on moving into a helper function or the like into `src/kernels/outputs/tracebackFormatter.ts` (which would be common to both formatters)

Basically notebook traceback formatters don't need the cell has providers at all.
& I'd like to make that very clear with the code (by having two seprate classes).
Also the cell hash provider is very specific to IW, hence that code only lives in the IW folder.

src/interactive-window/editor-integration/cellHashProviderFactory.ts
```diff
@@ -7,8 +7,9 @@ import { IConfigurationService } from '../../platform/common/types';
 import { IServiceContainer } from '../../platform/ioc/types';
 import { IKernel, IKernelProvider } from '../../kernels/types';
 import { CellHashProvider } from './cellhashprovider';
-import { ICellHashListener } from './types';
+import { ICellHashListener, IGeneratedCodeStorageFactory } from './types';
```
DonJayamanne, line 10: @IanMatthewHuff hope this name is better,

src/interactive-window/generatedCodeStoreManager.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+    private onDidCreateKernel(kernel: IKernel) {
+        const nb = getAssociatedNotebookDocument(kernel);
+        if (kernel.creator !== 'jupyterExtension' || nb?.notebookType !== InteractiveWindowView) {
+            return;
+        }
+        // Possible we changed kernels for the same document.
+        this.storageFactory.get(nb).clear();
+
+        kernel.onRestarted(
```
DonJayamanne, line 39: This class will be responsible for clearing the hashes & the like.
Still need to wire up generation of the hashes.

src/interactive-window/editor-integration/cellHashProviderFactory.ts
```diff
@@ -7,8 +7,9 @@ import { IConfigurationService } from '../../platform/common/types';
 import { IServiceContainer } from '../../platform/ioc/types';
```
DonJayamanne, line 7: @rchiodo @IanMatthewHuff Here's a rough implementation of what was discussed.
Still some work to be done, please review and let me know.

TODO:
* We need to ensure the cell hashes are generated fo IW before we execute (in the `createExecutionPromise` of `InteractiveWindow.node.ts` file).
* Need to get all of the hashes and send them in this same plaec.

@rchiodo I don't think we need the `willExecute` hook, as the cell hashes are very specific to IW and doesn't apply to notebooks.

Anyways, this is still a WIP. But hopefully the architecture is as discussed and looks like.

src/interactive-window/editor-integration/cellhashprovider.ts
```diff
@@ -400,7 +343,7 @@ export class CellHashProvider implements ICellHashProvider {
         return Array.isArray(traceback) ? traceback.map(this.modifyTracebackFrame.bind(this)) : [];
     }
 
-    private findCellOffset(hashes: IRangedCellHash[] | undefined, codeLines: string): number | undefined {
+    private findCellOffset(hashes: ICellHash[] | undefined, codeLines: string): number | undefined {
```
rchiodo, line 346: Can't comment farther down, but I thought the traceback stuff was moving out of this too?

src/notebooks/execution/cellExecution.ts
```diff
@@ -495,7 +511,7 @@ export class CellExecution implements IDisposable {
             // Compute the hash for the cell we're about to execute if on the interactive window
             let hash: ICellHash | undefined = undefined;
             if (this.cell.notebook.notebookType === InteractiveWindowView) {
-                hash = await this.cellHashProvider.addCellHash(this.cell);
+                hash = await this.generatedCodeStore?.getCellHash(this.cell);
```
rchiodo, line 514: Could you skip checking for interactive window? Hash would be empty for notebooks and then this code doesn't need to know about interactive.

src/notebooks/outputs/tracebackFormatter.ts
rchiodo: Sounds good to me. I wasn't sure if that was the case but if it is all the better.

rchiodo:
> Can't comment farther down, but I thought the traceback stuff was moving out of this too? 

Oh looks like you just haven't gotten that far.

rchiodo:
Looks good to me. Thanks.

src/interactive-window/editor-integration/cellhashprovider.ts
DonJayamanne: Yes, it wil.

src/notebooks/execution/cellExecution.ts
DonJayamanne: Agreed, thats possible. Still don't like the fact that we have to inject this thing, yuck..
Anways

rchiodo: I thought you figured out a way to have it just show up in the metadata. No injecting.

If I recall that's why we needed the willExecute hook.

await willExecute hook
get interactive metadata

src/interactive-window/editor-integration/codeGenerator.ts
```diff
@@ -364,15 +297,13 @@ export class CellHashProvider implements ICellHashProvider {
...
     */
     private addHiddenLines(
+        debug: boolean,
         cell: NotebookCell,
         source: string[],
         trueStartLine: number,
         firstNonBlankLineIndex: number
     ): { runtimeLine: number; debuggerStartLine: number } {
-        if (
-            this.debugService?.activeDebugSession &&
```
DonJayamanne, line 373: I think this class needs to be told whether it should generate debugging code or not.
There by removes the debug service and the place where we call the method `generateCode` (previously called `addCellhash`) we already have a variable named `isDebug`).

src/interactive-window/editor-integration/codeGenerator.ts
```diff
@@ -197,32 +161,7 @@ export class CellHashProvider implements ICellHashProvider {
...
-            }
-            if (!inserted) {
-                list.push(hash);
-            }
-            this.hashes.set(uristring, list);
-
+            this.storage.store(cell, hash);
             // Save a regex to find this file later when looking for
             // exceptions in output. Track backs only work on local files.
             if (!this.traceBackRegexes.has(uristring)) {
```
DonJayamanne, line 167: Will be removing these properties, can go into the `tracebackFormatter` we can generate the regex on the fly after some some string lookups.

src/interactive-window/interactiveWindow.node.ts
```diff
@@ -575,22 +578,15 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
                 await this.interactiveWindowDebugger.attach(kernel);
-
-                // Enable has to happen after the hidden code so that we don't hit breakpoints from previous cells
-                // Example:
-                // User has breakpoint on previous cell with name <ipython-2-hashystuff>
-                // We turn on tracing
-                // Hidden cell executes to set next cell to name <ipython-3-hashyotherstuff>
-                // Breakpoint fires in <ipython-2-hashystuff> because hidden cell inherits that value.
-                // So we have to enable tracing after we send the hidden cell.
-                kernelBeginDisposable = kernel.onPreExecute((c) => {
```
DonJayamanne, line 589: All of this complication has been removed now, much simpler flow

src/interactive-window/editor-integration/codeLensFactory.ts
```diff
@@ -156,7 +156,10 @@ export class CodeLensFactory implements ICodeLensFactory {
             cache.cellRanges.length &&
             this.configService.getSettings(document.uri).addGotoCodeLenses
         ) {
-            const hashes = this.getHashes();
+            const storage = this.generatedCodeStorageFactory.get({ fileUri: document.uri });
```
DonJayamanne, line 159: Added the ability to get the hashes for a give file Uri.
Even this class can do this by going through all notebook documents, then finding a storage that has the `document.uri` in the property `all` which contians all of the Uris.

But that feels like the job of the factory to return the right storage.
Either a helper function or just move into storage, i chose to add an overload into the storage.

src/interactive-window/editor-integration/codeGenerator.ts
```diff
@@ -0,0 +1,292 @@
...
+import { traceInfo } from '../../platform/logging';
+import { IConfigurationService } from '../../platform/common/types';
+import { getCellResource, uncommentMagicCommands } from './cellFactory';
+import { CellMatcher } from './cellMatcher';
+import { ICellHash, IInteractiveWindowCodeGenerator, IGeneratedCodeStore } from './types';
+import { getInteractiveCellMetadata } from '../helpers';
+
+// This class provides hashes for debugging jupyter cells. Call getHashes just before starting debugging to compute all of the
+// hashes for cells.
+export class CodeGenerator implements IInteractiveWindowCodeGenerator {
```
DonJayamanne, line 26: This is the old `CellHashProvider`, now responsible for just generating the code and storing it in the storage.

src/interactive-window/editor-integration/cellhashprovider.ts
```diff
@@ -1,559 +0,0 @@
...
-    endOffset: number;
-    deleted: boolean;
-    realCode: string;
-    trimmedRightCode: string;
-    firstNonBlankLineIndex: number; // zero based. First non blank line of the real code.
-}
-
-// This class provides hashes for debugging jupyter cells. Call getHashes just before starting debugging to compute all of the
-// hashes for cells.
-export class CellHashProvider implements ICellHashProvider {
```
DonJayamanne, line 49: Now split into:
* Code generate
* Code storage provider
* Traceback formatter (two classes, one for nb & one for IW)

src/interactive-window/outputs/tracebackFormatter.ts
```diff
@@ -0,0 +1,172 @@
...
+    ) {}
+    public format(cell: NotebookCell, traceback: string[]): string[] {
+        if (cell.notebook.notebookType !== InteractiveWindowView) {
+            return traceback;
+        }
+        const storage = this.storageFactory.get({ notebook: cell.notebook });
+        if (!storage) {
+            return traceback;
+        }
+        return traceback.map((traceFrame) => {
```
DonJayamanne, line 33: Not fully tested, but this is the new code
Will add tests

src/notebooks/execution/cellExecution.ts
DonJayamanne: Can do, but the concern @IanMatthewHuff  had was as follows:
* In one place we get the information out to the metadata
* In the other place we use the storage provider.
I'm torn over this, I would love to remove the storage provider, but agree with Ian, there's a descrepency, and we should stick to just one way to get the stuff, not two. I.e. not store the data in metadata as well as dictionary.

FYI - we need a dictionary as users can always clear all of the cells in the IW & start debugging a cell (at which point all of the cell hash information is lost - as its stored in the cells).

src/notebooks/outputs/tracebackFormatter.ts
DonJayamanne: Updated the code, not yet tested, 
will add tests

src/test/datascience/editor-integration/cellhashprovider.unit.test.ts
```diff
@@ -1,576 +1,566 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 'use strict';
-import { assert } from 'chai';
```
DonJayamanne, line 4: Will enable to fix and add more tests

src/test/datascience/editor-integration/codewatcher.unit.test.ts
```diff
@@ -1,2404 +1,2404 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 'use strict';
```
DonJayamanne, line 3: Will enable to fix and add more tests

src/notebooks/execution/cellExecution.ts
DonJayamanne: Happy to change this, but as mentioned torn between the two approachs (one less dependency vs two places to store the same information)

src/interactive-window/editor-integration/types.ts
```diff
@@ -68,30 +78,35 @@ export interface ICellHash {
...
 
-export const ICellHashListener = Symbol('ICellHashListener');
-export interface ICellHashListener {
-    hashesUpdated(hashes: IFileHashes[]): Promise<void>;
+export const IGeneratedCodeStore = Symbol('IGeneratedCodeStore');
+export interface IGeneratedCodeStore {
+    clear(): void;
+    readonly all: IFileHashes[];
+    getCellHash(cell: NotebookCell): ICellHash | undefined;
+    getFileHashes(fileUri: Uri): ICellHash[];
```
DonJayamanne, line 99: # Question: What do we call this method?

`getCellHash` is wrong again. Should it be `getForCell` and `getForFile` or `getCellInfo` and `getCellInfosByFile`, etc?

I want to get rid of this name `hash` from the method, it can be very confusing.
I'm not too fussed about the type `ICellHash` as that's rarely seen, its merely used to structure the data.

src/notebooks/execution/cellExecution.ts
DonJayamanne: @rchiodo made the changes, I think it like it better now.

src/interactive-window/editor-integration/types.ts
rchiodo: getGeneratedCode?

src/interactive-window/interactiveWindow.node.ts
```diff
@@ -754,13 +746,20 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             strippedCode,
             isMarkdown ? MARKDOWN_LANGUAGE : language
         );
+        const interactive = {
+            uristring: file.toString(), // Has to be simple types
+            line: line,
+            originalSource: code
+        };
+        const id = uuid();
+        const cellInfo = this.codeGeneratorFactory
```
rchiodo, line 755: Woot, This is much nicer :)

rchiodo (approved):
Looks great. Much simpler now.

src/interactive-window/editor-integration/codeGenerator.ts
IanMatthewHuff: Might make sure to tweak the class level comments before final review.

src/interactive-window/editor-integration/codeGeneratorFactory.ts
```diff
@@ -0,0 +1,40 @@
...
+import { inject, injectable } from 'inversify';
+import { IDocumentManager } from '../../platform/common/application/types';
+import { IConfigurationService } from '../../platform/common/types';
+import { CodeGenerator } from './codeGenerator';
+import { IGeneratedCodeStorageFactory, IInteractiveWindowCodeGenerator } from './types';
+import { NotebookDocument } from 'vscode';
+
+@injectable()
+export class CodeGeneratorFactory {
+    private readonly cellHashProvidersIndexedByNotebooks = new WeakMap<
```
IanMatthewHuff, line 13: Probably will all happen in clean up, but do want to make sure that we clean out the old cell hash name as much as possible. Love the new name, so I'd want the names in this file to be codeGenerators versus hash providers for the variable names.

src/interactive-window/editor-integration/codeLensFactory.ts
IanMatthewHuff: Should the storage maybe perform that little .all.find check? I'd think that you pretty much would always use that line.

src/interactive-window/editor-integration/types.ts
IanMatthewHuff: I liked code either for GeneratedCode or RuntimeCode. Seems like Generated code would work well. To me that expresses the intent / usage best.

src/interactive-window/interactiveWindow.node.ts
IanMatthewHuff: 100%

IanMatthewHuff (approved):

src/interactive-window/editor-integration/codeLensFactory.ts
DonJayamanne: can't, as we don't know which storage has the relevant yuri, hence we need to go through all storages, and only the factory had access to all via the internal map .

src/interactive-window/editor-integration/types.ts
DonJayamanne: Done, thanks

src/interactive-window/editor-integration/codeGeneratorFactory.ts
DonJayamanne: > old cell hash name as much as possible

Will do that

DonJayamanne: Done

src/interactive-window/editor-integration/codeGenerator.ts
DonJayamanne: Done

Issue #10060, opened 2022-05-18
Remote widget source provider doesn't work if any security required


rchiodo:
1. Remove the CDN sources (make them blank)
2. Connect to a remote server
3. Run something that needs a widget
4. Widget won't be downloaded

Root cause is the remote cdn provider doesn't do anything except pass the URI up to the webview for downloading. This doesn't work because the webview doesn't have the auth headers (or token) to make a valid request. 

I think the solution here is to refactor the CDN downloading code to support the server as a CDN source, but then it would make the appropriate requests with the correct headers, download from the remote jupyter server, and copy the source files to the same spot it does for CDNs.

Essentially remove this [source provider ](https://github.com/microsoft/vscode-jupyter/blob/9d9ad7d1f4756789ba9374d1a0351a558279ed8b/src/kernels/ipywidgets-message-coordination/remoteWidgetScriptSourceProvider.ts#L14) and have the CDN code also try kernel's baseURL. Meaning the kernel's base URL is always another CDN source.

DonJayamanne:
Upstream issue https://github.com/microsoft/vscode/issues/149868

DonJayamanne:
I believe this is now fixed. marking as closed so it get validated.

rchiodo:
This is not fixed for me at least. Same result as this issue that I just reopened #10510 

I did this:

- Set   ```"jupyter.widgetScriptSources": [],```
- Connected to a remote server
- Ran something with a widget

I get this dialog three times in a row:

![image](https://user-images.githubusercontent.com/19672699/176745641-02a33027-8f52-44dc-8862-eec909c430ab.png)

Then this pops up in the bottom right:

![image](https://user-images.githubusercontent.com/19672699/176745730-456a14f1-54dc-4d84-99f4-8408822d54e6.png)


Restarting the kernel does NOT make the widget appear.

I also thought we weren't supposed to need the CDN on remote? It should be able to pull the js files from the nbextension folder, right?

PR #10076, opened 2022-05-19
Support using a CDN for widgets

### LINKED ISSUE #9984: Support ipywidgets in the web extension
IPyWidgets don't work in the web extension. 

Root cause is downloading files and then making them available to the renderer.

rchiodo:
Fixes #9984 

Most widgets will work after this change but not all. Some require js to load from the jupyter server and that requires some changes in VS code to accomplish #10060


src/kernels/ipywidgets-message-coordination/scriptUriConverter.ts
```diff
@@ -19,11 +19,11 @@ export class ScriptUriConverter implements ILocalResourceUriConverter {
         return this.requestUriEmitter.event;
     }
     public get rootScriptFolder(): Uri {
-        return Uri.file(this._rootScriptFolder);
+        return this._rootScriptFolder;
```
rchiodo, line 22: This file isn't node specific anymore, but doesn't actually work in the web because we can't write to the extensionUri location. It isn't used for web right now but once [#10060](https://github.com/microsoft/vscode-jupyter/issues/10060) is implemented, it can be used to support that scenario. 

That's why I left this being usable for web too.

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -165,7 +167,8 @@ export class VSCodeNotebookController implements Disposable, IVSCodeNotebookCont
             viewType,
             label,
             this.handleExecution.bind(this),
-            this.getRendererScripts()
+            this.getRendererScripts(),
+            [this.scriptConverter.rootScriptFolder]
```
rchiodo, line 171: Whoops. This is the change needed for @mjbvz's new API (which hasn't gone through yet)

I'll just comment this out with a note saying this is how we'd use it.

src/platform/common/types.ts
```diff
@@ -192,29 +192,9 @@ export type DownloadOptions = {
     extension: 'tmp' | string;
 };
 
-export const IFileDownloader = Symbol('IFileDownloader');
```
rchiodo, line 195: IFileDownloader is not used. Leftovers from the python extension.

package.json
```diff
@@ -2183,8 +2183,6 @@
         "redux": "^4.0.4",
         "redux-logger": "^3.0.6",
         "reflect-metadata": "^0.1.12",
-        "request": "^2.87.0",
```
rchiodo, line 2186: Request has been deprecated, switched to 'cross-fetch' instead.

src/platform/common/net/httpClient.ts
```diff
@@ -0,0 +1,64 @@
...
+        const workspaceService = serviceContainer.get<IWorkspaceService>(IWorkspaceService);
+        this.requestOptions = { headers: { proxy: workspaceService.getConfiguration('http').get('proxy', '') } };
+    }
+
+    public async downloadFile(uri: string): Promise<Response> {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        return fetch.fetch(uri, this.requestOptions);
+    }
+
+    public async getJSON<T>(uri: string, strict: boolean = true): Promise<T> {
```
rchiodo, line 27: Going to remove these. Not used anywhere.

rchiodo: I removed the unit test for this class well because it's just a light wrapper around fetch. I don't think it needs a unit test

src/platform/common/net/httpClient.ts
```diff
@@ -0,0 +1,36 @@
...
+import { IHttpClient } from '../types';
+import { IServiceContainer } from '../../ioc/types';
+import { IWorkspaceService } from '../application/types';
+import { traceVerbose } from '../../logging';
+import * as fetch from 'cross-fetch';
+
+@injectable()
+export class HttpClient implements IHttpClient {
+    public readonly requestOptions: RequestInit;
+    constructor(@inject(IServiceContainer) serviceContainer: IServiceContainer) {
```
amunger, line 16: why inject the entire service container instead of the actual dependencies?

rchiodo: It was doing that before, I didn't notice. I'll change it if there's another update.

src/kernels/ipywidgets-message-coordination/cdnWidgetScriptSourceProvider.base.ts
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
amunger, line 1: is '.base' needed in the filename?

rchiodo: No. I just used it to distinguish that this is a base class for the .node and .web files. I've done this in other spots.

src/platform/common/platform/fileSystem.ts
```diff
@@ -101,4 +109,64 @@ export class FileSystem implements IFileSystem {
         const data = typeof text === 'string' ? Buffer.from(text) : text;
         return this.vscfs.writeFile(uri, data);
     }
+
+    async createTemporaryFile(options: { fileExtension?: string; prefix?: string }): Promise<TemporaryFileUri> {
+        // In non node situations, temporary files are created in the globalStorageUri location (extension specific)
```
amunger, line 114: This seems to be in node context as well since it's in a common file. Does globalStorageUri mean different things in those contexts, or will this only be called in web?

amunger (approved):

src/platform/common/platform/fileSystem.ts
rchiodo: No it works for both. It's intended to work for both (using it in the tests now). globalStorageUri is just a location on disk for desktop, and it's an in memory location for web.

I should change the comment though.

IanMatthewHuff (approved):

PR #10078, opened 2022-05-20
File system prep for web variable view

sadasant:
This PR tries to provide a web version of the FileSystem class, and to move the `platform/common/constants.node.ts` to a file I'm calling `namespaces.ts` which is platform agnostic.

I'm thinking I could write this in several ways, so instead of leaning towards a direction, I want your feedback:
Now that you have this PR, how would you approach the changes being referenced? would you do something else?


src/kernels/variables/pythonVariableRequester.ts
```diff
@@ -243,19 +244,29 @@ export class PythonVariablesRequester implements IKernelVariableRequester {
...
         }
     }
 
     // Read in a .py file and execute it silently in the given notebook
     private async runScriptFile(kernel: IKernel, scriptFile: string) {
-        if (await this.fs.localFileExists(scriptFile)) {
-            const fileContents = await this.fs.readLocalFile(scriptFile);
-            return kernel.session ? executeSilently(kernel.session, fileContents) : [];
+        try {
+            const uriPath = vscode.Uri.parse(scriptFile);
```
rchiodo, line 257: I think it would be better to use IFileSystem here to read the script file.

rchiodo: Otherwise you're duping the fileSystem read code here.

src/kernels/variables/types.ts
```diff
@@ -27,6 +27,11 @@ export interface IJupyterVariable {
     fileName?: string;
 }
 
+export const IRootDirectory = Symbol('IRootDirectory');
```
rchiodo, line 30: I don't think this is necessary. We already have IExtensionContext, which has an extensionUri property on it. That would be the root directory.

I also think IRootDirectory is a little ambiguous. Root of what? Could be workspace. Could be vscode. Could be extension.

src/platform/common/namespaces.ts
```diff
@@ -0,0 +1,55 @@
+import * as path from '../../platform/vscode-path/path';
+export * from './constants';
+
+export namespace DataFrameLoading {
+    export function SysPath(rootDir: string) {
```
rchiodo, line 5: This is good but I think all of these should take a URI and return a URI. We shouldn't be doing path manipulation with  path entries. Path entries are always posix, path.join is OS specific.

You'd use the urlPath functions instead to concat stuff together. See 'src/platform/vscode-path/resource'.

src/platform/common/platform/fileSystem.node.ts
```diff
@@ -5,23 +5,36 @@ import * as tmp from 'tmp';
...
 /**
  * File system abstraction which wraps the VS Code API.
  */
 @injectable()
 export class FileSystem extends FileSystemBase implements IFileSystemNode {
     private globFiles: (pat: string, options?: { cwd: string; dot?: boolean }) => Promise<string[]>;
+
     constructor() {
         super();
+        this.rootDirectory = EXTENSION_ROOT_DIR;
```
rchiodo, line 24: I don't think this is necessary. We should just use IExtensionContext as I mentioned before.

src/platform/common/platform/fileSystem.web.ts
```diff
@@ -0,0 +1,113 @@
+import * as vscode from 'vscode';
```
rchiodo, line 1: I don't think this file is necessary. The common FileSystem is identical to this.

src/webviews/extension-side/webviews/webview.node.ts
```diff
@@ -37,7 +37,11 @@ export abstract class Webview implements IWebview {
     ) {
         const webViewOptions: WebviewOptions = {
             enableScripts: true,
-            localResourceRoots: [Uri.file(this.options.rootPath), Uri.file(this.options.cwd), ...additionalRootPaths]
+            localResourceRoots: [
+                fs.normalize(this.options.rootPath),
```
rchiodo, line 41: Why do you need to normalize here? It should load regardless.

rchiodo: Wouldn't normalize only be necessary for comparison?

rchiodo (changes requested):
I like the new 'namespaces' file but I think the RootDirectory and Roots on the FileSystems  are unnecessary. 

You should also use url path joins where possible.

src/platform/common/platform/fileSystem.web.ts
rchiodo: Meaning I don't think you need a web FileSystem. The common FileSystem should already work.

sadasant: One difference is that in Node, `fsPath` and `Uri.file` were used, but they don't work in web. Am I misinterpreting that?

src/webviews/extension-side/webviews/webview.node.ts
sadasant: `file://` URls do not work well in the web. I've had to replace all the code that uses `Uri.file` in order to make things work in the web.

sadasant:
Got some awesome feedback over Slack as well. Sharing here to keep it on record:

Rich:
> Don’t use fsPath or path at all. Just use URIs by themselves and the uri path manipulation functions to add/remove stuff from them. Then you don’t have to worry about separators.
> Or about creating Uri.file items. Because everything starts from some other root URI.

Ian:
> Basically any time you see raw string paths, string path comparisons, manipulations, ect… then you probably need to replace that with Uris. Rich already did a big pass on this previously, it’s big part of the web work.

Thank you both! That really simplifies things

src/webviews/extension-side/webviews/webview.node.ts
rebornix: I think we can follow how VS Code builtin webview extensions handle `localResourceRoots` with `Uri` https://github.com/microsoft/vscode/blob/main/extensions/markdown-language-features/src/preview/preview.ts#L430-L444

src/platform/common/platform/fileSystem.web.ts
rebornix: The common `FileSystem` is meant to use `vscode.workspace.fs` for all the operations, and the `FileSystem.node` override some if `nodejs` fs is required. The methods in this file are safe to move into the common `FileSystem`.

src/platform/common/platform/fileSystem.ts
```diff
@@ -16,15 +16,21 @@ const ENCODING = 'utf8';
...
 export class FileSystem implements IFileSystem {
     protected vscfs: vscode.FileSystem;
+    public rootDirectory: string;
+
     constructor() {
         this.vscfs = vscode.workspace.fs;
+        this.rootDirectory = vscode.Uri.file('/').path;
+    }
+
+    public normalize(path: string): vscode.Uri {
```
rebornix, line 26: Since it's wrapping `vscode.Uri.file`, I would suggest we call it `file`. `Normalize` is a bit ambigious here.

src/kernels/variables/debuggerVariables.ts
```diff
@@ -8,23 +8,25 @@ import { DebugAdapterTracker, Disposable, Event, EventEmitter } from 'vscode';
 import { DebugProtocol } from 'vscode-debugprotocol';
 import { getAssociatedNotebookDocument } from '../../notebooks/controllers/kernelSelector';
 import { IDebugService, IVSCodeNotebook } from '../../platform/common/application/types';
-import { DataFrameLoading, GetVariableInfo } from '../../platform/common/constants.node';
+// TODO: Clean this up. And the SysPath use cases in this file.
```
IanMatthewHuff, line 11: Looks like an old TODO here

src/platform/common/namespaces.ts
IanMatthewHuff: Rich's recent widgets PR has some good examples of this I think to look at for the path => Uri conversion.

rebornix (commented):
Looks good in overall. My suggestions are the same as Rich and Ian. The filesystem.web is not necessary, we should make the common one support Web (as it's already using `vscode.workspace.fs`).

For the `rootDirectory`, I suggest we take a look at how VS Code builtin markdown preview extension handles them. It feels wrong to invent this since VS Code supports multi-workspace and we should not convert Uri to string and then to `Uri`.

src/webviews/extension-side/webviews/webview.node.ts
DonJayamanne: > file:// URls do not work well i

i think the root problem is the  that were still not using uris in pokeberry off places. i think we should be using uris instead of assuming that files paths have the scheme such as `file://`..

ie if we have pieces that hard code Uri.file, then that's a problem, i think it should be uri.join asking with the real paths from either the work space or the extension root using context.extensionUri.

what im getting at is, we shouldn't have to hardcode Uri.file nor should we have to parse using Uri.parse

DonJayamanne: > use url path joins where possible.

that's my point too

src/platform/common/namespaces.ts
DonJayamanne: also id suggest renaming the file. namespace is too ambiguous. namespace are just a collection names. 
eg rename to script utils or something that's a more descriptive. as this no longer contains constants, but contains done utility functions to return psths

sadasant:
Closing this. I'll be able to re-construct it in a cleaner way afterwards.

PR #10086, opened 2022-05-20
Watch for URI removal to remove passwords

IanMatthewHuff:
Fixes #

Followup fix from review comment here:
https://github.com/microsoft/vscode-jupyter/pull/10077#discussion_r878343138

### DIFF
--- a/src/kernels/jupyter/launcher/serverUriStorage.ts
+++ b/src/kernels/jupyter/launcher/serverUriStorage.ts
@@ -24,9 +24,9 @@ export class JupyterServerUriStorage implements IJupyterServerUriStorage {
     public get onDidChangeUri() {
         return this._onDidChangeUri.event;
     }
-    private _onDidRemoveUri = new EventEmitter<string>();
-    public get onDidRemoveUri() {
-        return this._onDidRemoveUri.event;
+    private _onDidRemoveUris = new EventEmitter<string[]>();
+    public get onDidRemoveUris() {
+        return this._onDidRemoveUris.event;
     }
     constructor(
         @inject(IWorkspaceService) private readonly workspaceService: IWorkspaceService,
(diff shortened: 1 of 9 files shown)


src/kernels/jupyter/launcher/serverUriStorage.ts
```diff
@@ -149,6 +150,13 @@ export class JupyterServerUriStorage implements IJupyterServerUriStorage {
             Settings.JupyterServerRemoteLaunchUriListKey,
             undefined
         );
+
+        // Notify out that we've removed the list to clean up controller entries, passwords, ect
+        this._onDidRemoveUris.fire(
```
IanMatthewHuff, line 155: I waffled on just calling removeUri() function per each item in the list and leaving the event as is. But the remove operation felt a bit different to me (such as removeURI() watching the active URI and setting back to local on removal, which we don't want here) so I moved this instead to a bulk event here instead of firing a chain of the other single removal events.

IanMatthewHuff:
Follow up PR per a review comment that I'd missed the removal if using the trash button from the server list, which was a good catch.

IanMatthewHuff:
Actually leaving this in draft for a bit of investigation. Seeing a bit of an odd error from VS Code core on password deletion when the entry from password connect is removed from the list.

Also seeing a secondary issue that if I input a server with a password it actually doesn't get put into the server list and switched in to remote mode the first time that I input it and add the pass word. It has the happen the second time after the password is already saved.

IanMatthewHuff:
Ok, the issues that I found out seen outside of this PR, so going ahead with it. Issues file here:
https://github.com/microsoft/vscode-jupyter/issues/10103
https://github.com/microsoft/vscode-jupyter/issues/10102

rchiodo (approved):

PR #10092, opened 2022-05-22
Proposal: platform/common/constants.node to ScriptPathUtils class

sadasant:
This PR is a continuation from the feedback received in #10078

Here, I'm focusing on asking: What's the best approach to make `platform/common/constants.node` agnostic to Node.js?

As an answer, I'm:
- Taking the feedback that `IExtensionContext`'s `extensionUri` should be enough to refer to the root.
  - Where I'm assuming `context.extensionPath` == `context.extensionUri.path`.
- Changing from namespaces to a class that loads the context using `inversify`'s `inject`.
- Changing from namespace variables to class getters, which keep how the properties are used (they don't look like functions), thus minimizing the impact of the changes on the affected files.

This approach is fairly opinionated in an attempt to minimize the surface of the changes.
How does this look?
Any and all feedback appreciated.


src/platform/common/types.ts
```diff
@@ -350,3 +350,33 @@ export interface ICellRange {
     range: Range;
     cell_type: string;
 }
+
+// Used by platform/common/scriptPathUtils
+export const IScriptPathUtils = Symbol('IConfigurationService');
+export interface IScriptPathUtils {
+    rootDirectory: string;
+    DataFrameLoading: {
```
DonJayamanne, line 358: please change the casing out properties to camel, eg `dataFrameLoading`

src/platform/common/types.ts
```diff
@@ -350,3 +350,33 @@ export interface ICellRange {
     range: Range;
     cell_type: string;
 }
+
+// Used by platform/common/scriptPathUtils
+export const IScriptPathUtils = Symbol('IConfigurationService');
+export interface IScriptPathUtils {
+    rootDirectory: string;
```
DonJayamanne, line 357: i world remove this, and use `context.extensionUri` where ever this is required, else we just end up having more than one way to get the same value

DonJayamanne: i dont think this is used anywhere, hence you can remove this.

src/platform/common/types.ts
```diff
@@ -350,3 +350,33 @@ export interface ICellRange {
...
+    DataFrameLoading: {
+        SysPath: string;
```
DonJayamanne, line 359: id chance all of these to be camel. cased, and i don't think `sysPath` is used outside this class, never your can remove this

src/platform/common/types.ts
sadasant: Just to confirm, `context.extensionPath` is valid in this case. Is that correct?

DonJayamanne: Yes, however if this is going to be used on web as well, then you would want to use `context.extensionUri`

src/platform/common/types.ts
DonJayamanne: > I'm assuming the approach is ok otherwise the casing?

Yes.

src/platform/common/scriptPathUtils.ts
```diff
@@ -0,0 +1,50 @@
+import { inject } from 'inversify';
+import * as path from '../vscode-path/path';
+import { IExtensionContext, IScriptPathUtils } from './types';
+
+export class ScriptPathUtils implements IScriptPathUtils {
+    constructor(@inject(IExtensionContext) private readonly context: IExtensionContext) {}
+
+    get dataFrameLoading() {
```
DonJayamanne, line 8: This file seems to be specific to node.js
As we're using `this.context.extensionPath`. 
However that means this cannot be used on the web and we want to be able to use this on the web.

Hence my suggestion is:
* use `this.context.extensionUri` instead of `context.extensionPath`
This changes the types from strings to Uri, which is a good thing as we want to avoid using strings for paths where ever possible, & use Uris instead.

sadasant: My concern then is that Uris need to convert to strings at some point, right? specially in these cases where we’re concatenating paths.

src/platform/common/types.ts
sadasant: Even then, I need to do `uri.path` at some point to get the string version of the URI 🤔

sadasant: Ohhh `extensionPath` is shorthand of `extensionUri.fsPath`.... I expected it to be shorthand of `extensionUri.path` 🤦‍♂️ fixing it!

src/kernels/kernel.node.ts
```diff
@@ -72,11 +72,11 @@ export class Kernel extends BaseKernel {
         if (getAssociatedNotebookDocument(this)?.notebookType === InteractiveWindowView) {
             // If using ipykernel 6, we need to set the IPYKERNEL_CELL_NAME so that
             // debugging can work. However this code is harmless for IPYKERNEL 5 so just always do it
-            if (await this.fs.localFileExists(AddRunCellHook.ScriptPath)) {
-                const fileContents = await this.fs.readLocalFile(AddRunCellHook.ScriptPath);
+            if (await this.fs.localFileExists(this.scriptPaths.addRunCellHook.scriptPath.fsPath)) {
+                const fileContents = await this.fs.readLocalFile(this.scriptPaths.addRunCellHook.scriptPath.fsPath);
```
sadasant, line 76: To make `scriptPathUtils` less Node specific, I made `scriptPath` into a `Uri`, and here I'm using `fsPath`.

src/platform/common/scriptPathUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+        const sysPath = uriPath.joinPath(
+            this.context.extensionUri,
+            'pythonFiles',
+            'vscode_datascience_helpers',
+            'dataframes'
+        );
+        const dataFrameImport = `__import__('vscodeDataFrame')`;
+        return {
+            dataFrameSysImport: `import sys\nsys.path.append("${sysPath.path.replace(/\\/g, '\\\\')}")`,
```
sadasant, line 17: To make `scriptPathUtils` less Node specific, I made `scriptPath` into a `Uri`. However, In the current design I can't fix `dataFrameSysImport` to either use `fsPath` or `path` depending on the environment. In a future commit, I could change the design to, for example, expose both `dataFrameSysImport` and `dataFrameSysImportFsPath`, as a possible solution. However, I should ask: Would `path` here work both for Node.js and the browser?

sadasant: cc: @DonJayamanne

sadasant:
Main seems to be broken for me. I'm getting like 300 errors when I compile main.

src/kernels/kernel.node.ts
rchiodo: Maybe you could move the reading of scripts into scriptPathUtils? Then it's isolated in one spot.

rchiodo (approved):
Looks good to me.

IanMatthewHuff (approved):
Looks like there are some build errors? But the approach looks good to me. Approved assuming the build errors are cleaned up. From my reading they look like small fixes to make.

src/kernels/kernel.node.ts
sadasant: I'd like to think about that in a separate PR. This PR is pretty scoped. Adding `fs` to the `scriptPathUtils` is something I'd like to think with the greater context of changes I'm doing.

src/platform/common/scriptPathUtils.ts
DonJayamanne: @rchiodo will the use of `path` work here?
Since we're dealing with remote jupyter kernels, the values sent to the kernel should contain actual file paths, hence I'm not sure `path` property would work here. I thought you mentioned that its not a real path and has some funky logic to escape drive letters & network paths.

I.e. I don't think this will work in desktop & remote. Possible this is an exception where we can use fsPath or we need something similar.

DonJayamanne (changes requested):
Got a concern about the use of `Uri.path` instead of `uri.fsPath`, awaiting response from @rchiodo

src/platform/common/scriptPathUtils.ts
rchiodo: Wow, good catch. Yeah this needs to be redesigned entirely for the web. It needs to import the entire script as a text file and run it that way instead of importing it through the file path.

rchiodo (changes requested):
I'm afraid the script stuff isn't going to work in the web

src/platform/common/scriptPathUtils.ts
rchiodo: We could do like I did for the addRunCellHook.

src/kernels/kernel.node.ts
DonJayamanne: I would disagree, merging this PR as is breaks this feature completely (i.e. we're going to knowingly introduce a bug), 
The way i would see this PR is `Ready the script utils for web`
Hence, I think it should be done now, instead of later

src/platform/common/scriptPathUtils.ts
DonJayamanne: > Maybe you could move the reading of scripts into scriptPathUtils? Then it's isolated in one spot.

Feels like we'll have to address the above comment in this PR, moving reading of the script file to this new scriptPathUtils file and then importing the script as mentioned by @rchiodo

DonJayamanne: > Looks like there are some build errors? But the approach looks good to me

I sure do hope the build failures are a result of the use of `path`, if not, I'd like some tests for this (perhaps we can add them in debt week - to ensure we don't introduce such breakages)

PR #10104, opened 2022-05-23
inject web class into services

amunger:
The wrong class was put into the serviceProvider and causing error logs about the base class not having the injectable tag


amunger:
We have a test that prevents this kind of thing for node, we should also implement that for web if possible. I can create an issue for that

rchiodo (approved):

PR #10113, opened 2022-05-24
Non local generic file system

### LINKED ISSUE #10112: InteractiveWindow shouldn't use areLocalPathsSame.
https://github.com/microsoft/vscode-jupyter/blob/760d61e0fd56d4eb0dfb9cfdbc61dad4e38ab14c/src/interactive-window/interactiveWindow.ts#L457

Anything with local paths should be .node specific because it's the only case where you can read from the local file system. 

We should change IFileSystem to not have any 'local' functions on it as it makes it confusing as to when you can use strings and when you can use URIs.

Local paths should only be used in IFileSystemNode, where the node extension can reliably assume it can read the local file system.

sadasant:
Ensures the non-Node-specific FileSystem does not have any `local` functions.

Fixes #10112

### DIFF
--- a/src/interactive-window/interactiveWindow.ts
+++ b/src/interactive-window/interactiveWindow.ts
@@ -454,7 +454,9 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         let saved = true;
         const file = getFilePath(fileUri);
         // Make sure the file is saved before debugging
-        const doc = this.documentManager.textDocuments.find((d) => this.fs.areLocalPathsSame(d.fileName, file));
+        const doc = this.documentManager.textDocuments.find((d) =>
+            this.fs.arePathsSame(Uri.parse(d.fileName), Uri.parse(file))
+        );
         if (doc && doc.isUntitled) {
             // Before we start, get the list of documents
             const beforeSave = [...this.documentManager.textDocuments];
@@ -686,7 +688,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             this.fileInKernel = file;
             await kernel.executeHidden(`__file__ = '${file.replace(/\\/g, '\\\\')}'`);
         } else if (
-            (!this.fileInKernel || !this.fs.areLocalPathsSame(this.fileInKernel, file)) &&
+            (!this.fileInKernel || !this.fs.arePathsSame(Uri.parse(this.fileInKernel), Uri.parse(file))) &&
             this.mode !== 'perFile'
         ) {
             traceInfoIfCI(`Initializing __file__ in setFileInKernel with ${file} for mode ${this.mode}`);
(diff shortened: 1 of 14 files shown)


src/interactive-window/interactiveWindow.ts
```diff
@@ -454,7 +454,9 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         let saved = true;
         const file = getFilePath(fileUri);
         // Make sure the file is saved before debugging
-        const doc = this.documentManager.textDocuments.find((d) => this.fs.areLocalPathsSame(d.fileName, file));
+        const doc = this.documentManager.textDocuments.find((d) =>
+            this.fs.arePathsSame(Uri.parse(d.fileName), Uri.parse(file))
```
rchiodo, line 458: d.fileName should be replaced with d.uri
file should be replaced with fileUri

src/interactive-window/interactiveWindow.ts
```diff
@@ -686,7 +688,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             this.fileInKernel = file;
             await kernel.executeHidden(`__file__ = '${file.replace(/\\/g, '\\\\')}'`);
         } else if (
-            (!this.fileInKernel || !this.fs.areLocalPathsSame(this.fileInKernel, file)) &&
+            (!this.fileInKernel || !this.fs.arePathsSame(Uri.parse(this.fileInKernel), Uri.parse(file))) &&
```
rchiodo, line 691: Same thing here. Don't use the 'file' if you can help it.

src/platform/common/platform/fileSystem.ts
```diff
@@ -30,16 +30,7 @@ export class FileSystem implements IFileSystem {
...
-            // eslint-disable-next-line local-rules/dont-use-fspath
-            return this.areLocalPathsSame(path1.fsPath, path2.fsPath);
-        } else {
-            return path1.toString() === path2.toString();
-        }
-    }
-
-    areLocalPathsSame(path1: string, path2: string): boolean {
-        return arePathsSame(path1, path2);
+        return path1.toString() === path2.toString();
```
rchiodo, line 33: There is a isEqual on URI that should be used for this. It has an ignoreCase option I believe for windows.

rchiodo: If you look up areInterpreterPathsSame it's using that isEqual function.

src/interactive-window/interactiveWindow.ts
sadasant: On it!

src/platform/common/platform/fileSystem.ts
sadasant: do you mean on `vscode.Uri`? I can't find that `isEqual`

We do have a `arePathsSame` in `platform/fileUtils.ts` though

rchiodo (approved):
Thanks. Looks good.

Maybe a comment in FileSystemNode that says why we have to use 'local' only in node?

IanMatthewHuff (approved):

PR #10119, opened 2022-05-24
No more css and themes on the web views

sadasant:
These classes around CSS and Themes are not going to work on the web side, and we've confirmed over the chat that these are no longer needed. Please, let me know if I got that wrong.

How does this look?


rchiodo (approved):

DonJayamanne (changes requested):
I think we still need to know on the UI side whether its light or dark theme, as that controls the icons in webviews.
Hence we cannot sending the value `isDark` to the UI.

src/webviews/extension-side/webviewHost.ts
```diff
@@ -161,10 +159,6 @@ export abstract class WebviewHost<IMapping> implements IDisposable {
                 this.webViewRendered();
                 break;
 
-            case CssMessages.GetCssRequest:
```
DonJayamanne, line 164: I'd rename this to `GetTheme` which would return `isDark`

src/webviews/extension-side/webviewHost.ts
```diff
@@ -296,23 +290,6 @@ export abstract class WebviewHost<IMapping> implements IDisposable {
...
-        this.setTheme(requestIsDark);
-        const isDark = settings.ignoreVscodeTheme
-            ? false
-            : await this.themeFinder.isThemeDark(settings.extraSettings.theme);
-        const resource = this.owningResource;
-        const css = await this.cssGenerator.generateThemeCss(resource, requestIsDark, settings.extraSettings.theme);
-        return this.postMessageInternal(CssMessages.GetCssResponse, {
-            css,
-            theme: settings.extraSettings.theme,
-            knownDark: isDark
```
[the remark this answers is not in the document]
DonJayamanne: Removing the CSS as is will actually break things, we need to ensure we don't set any CSS anywhere, and just use the styles injected by VS Code (removing our styles for font size & family will fix this).

rchiodo:
> I think we still need to know on the UI side whether its light or dark theme, as that controls the icons in webviews. Hence we cannot sending the value `isDark` to the UI.

I thought 'isDark' requires reading the theme json though? Maybe we can determine it some other way.

src/webviews/extension-side/webviewHost.ts
sadasant: Thank you! can do!

src/webviews/extension-side/webviewHost.ts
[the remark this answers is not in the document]
sadasant: Thank you! can do!

DonJayamanne:
> I thought 'isDark' requires reading the theme json though?

Nope

> Maybe we can determine it some other way.

Yes,
we can use the `activeColorTheme` property in `vscode.window`

src/webviews/extension-side/webviewHost.ts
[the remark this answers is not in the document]
DonJayamanne: I think we should also remove the following, as this should already be setup by VS Code for webviews.
`background-color: var(--vscode-editor-background);`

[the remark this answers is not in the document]
DonJayamanne: We should also remove the file `cssCodeGenerator.node.ts` as thats not required (basically if we're removing the loading of themes from files then this file too needs to go)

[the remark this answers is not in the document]
DonJayamanne: Similarly remote `IThemeFinder` and `ICodeCssGenerator`

sadasant:
@DonJayamanne thank you so much!! Your feedback helps me a lot 🙏

rchiodo (approved):

src/webviews/webview-side/interactive-common/common.css
```diff
@@ -10,8 +10,8 @@ html {
 }
 
 #main-panel {
-    background: var(--override-background, var(--vscode-editor-background));
-    color: var(--override-foreground, var(--vscode-editor-foreground));
+    background: var(--vscode-editor-background);
+    color: var(--vscode-editor-foreground);
```
DonJayamanne, line 14: Do we need this? I would have thought this is not necessary.

src/webviews/webview-side/interactive-common/common.css
```diff
@@ -10,8 +10,8 @@ html {
 }
 
 #main-panel {
-    background: var(--override-background, var(--vscode-editor-background));
-    color: var(--override-foreground, var(--vscode-editor-foreground));
+    background: var(--vscode-editor-background);
```
DonJayamanne, line 13: Do we need this? I would have thought this is not necessary.

DonJayamanne: Doesn't matter

src/webviews/webview-side/interactive-common/common.css
DonJayamanne: Doestn' matter

sadasant: Just trying to make the least changes possible, `--vscode-editor-foreground` is in many places 😅

src/webviews/webview-side/interactive-common/common.css
sadasant: Just trying to make the least changes possible, `--vscode-editor-foreground` is in many places 😅

PR #10122, opened 2022-05-24
Hide progress message after export completes

### LINKED ISSUE #9112: Exporting to HTML notification persists
## Environment data
-   VS Code Version: 1.65.0-insider (user setup)
-   Commit: eeeab304b9ac777b4a8dddf55f66ffebd9e4e7f2
-   Jupyter Extension version (available under the Extensions sidebar): v2022.2.1010561006
-   Python Extension version (available under the Extensions sidebar): v2022.1.1854775815-dev
-   OS (Windows | Mac | Linux distro) and version: Windows_NT x64 10.0.22000
-   Python and/or Anaconda version: 4.11.0

## Expected behaviour

1. Export notebook to html
2. While notebook is exporting, toast notification pops up
3. Export completes successfully
4. When asked, "Would you like to open the exported file?", don't click on anything (instead, I opened it directly on my browser)
5. Come back to VS Code, and the "exporting to HTML" notification should disappear

image

## Actual behaviour
The "exporting to HTML" notification persists until I click "cancel" ("would you like to open..." does not persist)

DonJayamanne:
Fixes #9112

### DIFF
--- a/src/platform/export/fileConverter.node.ts
+++ b/src/platform/export/fileConverter.node.ts
@@ -1,14 +1,14 @@
 import { inject, injectable, named } from 'inversify';
-import { CancellationToken, NotebookDocument, Uri } from 'vscode';
+import { Uri } from 'vscode';
 import { IApplicationShell } from '../common/application/types';
 import { IConfigurationService } from '../common/types';
-import { PythonEnvironment } from '../pythonEnvironments/info';
 import { ProgressReporter } from '../progress/progressReporter';
 import { ExportFileOpener } from './exportFileOpener';
 import { ExportFormat, INbConvertExport, IExportDialog, IFileConverter, IExport } from './types';
 import { IFileSystemNode } from '../common/platform/types.node';
 import { FileConverter as FileConverterBase } from './fileConverter';
 import { ExportUtil } from './exportUtil.node';
+import { noop } from '../common/utils/misc';
 
 // Class is responsible for file conversions (ipynb, py, pdf, html) and managing nb convert for some of those conversions
 @injectable()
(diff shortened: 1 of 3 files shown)


src/platform/export/fileConverter.node.ts
```diff
@@ -39,24 +39,7 @@ export class FileConverter extends FileConverterBase implements IFileConverter {
         );
     }
 
-    override async performExport(
```
rchiodo, line 42: Was this not used?

DonJayamanne: We were overriding it just to change one argument in a method call of the base class.
Rather than copying and pasting the code from the base class, I've added a new method that can be overridden to change the behaviour of what happens when opening the file.

src/platform/export/fileConverter.ts
```diff
@@ -107,10 +108,17 @@ export class FileConverter implements IFileConverter {
         }
```
DonJayamanne, line 108: @rchiodo this method was identical in the child class, except for the line 110, we were adding a new argument.
Tomorrow if we have more formats then we need to copy code from the parent class into this child class, which isn't good

src/platform/export/fileConverter.ts
```diff
@@ -107,10 +108,17 @@ export class FileConverter implements IFileConverter {
...
 
         if (target) {
-            await this.exportFileOpener.openFile(format, target, true);
```
DonJayamanne, line 110: Moved this code into a different method so that the child classes can do what they like (change behaviour on how files are opened after its exported)

rchiodo (approved):

PR #10131, opened 2022-05-25
Enable IW in web

### LINKED ISSUE #9717: Support interactive window in web extension
There's a non trivial amount of work to make the interactive window support web. Mostly around the commands using the local file system for import/export.

amunger:
Fixes #9717

One notable command that was not enabled is `execute selection` since it uses an execution service to get the applicable block of code from a selection and I haven't taken the time to convert that yet.

### DIFF
--- a/src/interactive-window/editor-integration/codeLensFactory.ts
+++ b/src/interactive-window/editor-integration/codeLensFactory.ts
@@ -17,7 +17,13 @@ import {
 import { IDocumentManager, IVSCodeNotebook, IWorkspaceService } from '../../platform/common/application/types';
 import { traceWarning, traceInfoIfCI } from '../../platform/logging';
 
-import { ICellRange, IConfigurationService, IDisposableRegistry, Resource } from '../../platform/common/types';
+import {
+    ICellRange,
+    IConfigurationService,
+    IDisposableRegistry,
+    IsWebExtension,
+    Resource
+} from '../../platform/common/types';
 import * as localize from '../../platform/common/utils/localize';
 import { getInteractiveCellMetadata } from '../helpers';
 import { IKernelProvider } from '../../kernels/types';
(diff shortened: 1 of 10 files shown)


rchiodo (approved):

rchiodo:
We should probably convert the interactive tests to also run for the web. I did that for ipywidget tests and remote tests. I think you could likely follow a similar pattern.

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -233,11 +233,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
                     ? this.notebook.notebookDocuments.find((item) => item.notebookType === notebookType)
                     : undefined;
             const controller = await this.createDefaultRemoteController(notebookType, notebook);
-            if (controller) {
-                return controller;
-            }
-            traceVerbose('No default remote controller, hence returning the active interpreter');
-            return this.createActiveInterpreterController(notebookType, resource);
```
DonJayamanne, line 240: Isn't this a breaking change for desktop?
I.e. its possible that we'll never have a controller, when in the past we'd return the active interpreter.

DonJayamanne (changes requested):
Looks like we've introduced a breaking change to not have an active interpreter controller for IW

src/notebooks/controllers/notebookControllerManager.ts
amunger: whoops, I misunderstood what `isLocalLaunch` meant, good catch

src/interactive-window/editor-integration/codeLensFactory.ts
```diff
@@ -256,8 +263,13 @@ export class CodeLensFactory implements ICodeLensFactory {
                 Commands.DebugStop,
                 Commands.RunCellAndAllBelowPalette
             ];
-            fullCommandList = fullCommandList.filter((item) => !commandsToBeDisabledIfNotTrusted.includes(item));
+        } else if (IsWebExtension) {
```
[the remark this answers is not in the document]
amunger: you mean a separate `if`? yeah that would probably make more sense

amunger, line 266: need to inject the boolean IsWebExtension, this is just the symbol, which is always true

DonJayamanne (approved):

PR #10137, opened 2022-05-25
Turn off string filtering for completions

### LINKED ISSUE #8983: No autocomplete for pandas column name in Jupyter notebook using bracket notation ['col']
## Environment data

-   VS Code version: 1.64.00
-   Jupyter Extension version (available under the Extensions sidebar):  v2022.1.1001821375
-   Python Extension version (available under the Extensions sidebar): v2022.0.1814523869
-   OS (Windows | Mac | Linux distro) and version: Windows 10
-   Python and/or Anaconda version: python 3.8.12

## Expected behaviour

Autocomplete works for pandas dataframes in jupyter notebooks.

## Actual behaviour

Autocomplete does not work. Users have to manually type the column names every time. It even doesn't work when I've manually specified the column name several times. I've tried all the language servers, including Jedi, Pylance, and the default option. No autocomplete. 

## Steps to reproduce:

1. Create pandas dataframe 

`df = pd.DataFrame({'test': [1,2], 'test2': [2,3]})`

2. Call dataframe variable and check if autocomplete works for the columns

`df['test']`

autocomplete_fail

## Logs

<details>

<summary>Output for <code>Jupyter</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Jupyter</code>)
</summary>

<p>

```
XXX
```

</p>
</details>

rchiodo:
Fixes #8983 - sort of.

We had added filtering to jupyter completions to get the results to be more relevant. This broke a specific scenario though, user typing in this:

```
df['']
```
And wanting to see column names in the quotes.

This PR 'fixes' this but not filtering column names from quotes. However the completions from jupyter are terrible. It just returns everything - unless you reactivate Jedi. Then they're better. We disable Jedi by default  though because it causes perf problems. 

So I added a setting to turn Jedi back on.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1985,7 +1985,7 @@
                 },
                 "jupyter.pythonCompletionTriggerCharacters": {
                     "type": "string",
-                    "default": ".%",
+                    "default": ".%'\"",
                     "description": "Characters which trigger auto completion on a python jupyter kernel.",
                     "scope": "machine"
                 },
@@ -1999,6 +1999,12 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Add PYTHONNOUSERSITE to kernels before starting. This prevents global/user site-packages from being used in the PYTHONPATH of the kernel.",
+                    "scope": "machine"
+                },
+                "jupyter.enablePreciseKernelCompletions": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Enables the use of Jedi to generate more precise kernel completions. This can impact peformance, such as preventing executions from running. Use with caution.",
                     "scope": "machine"
                 }
             }
--- a/news/2 Fixes/8893.md
+++ b/news/2 Fixes/8893.md
@@ -1,0 +1,1 @@
+Fix to provide autocomplete inside of quoted strings. This fix also enabled a setting to allow the use of Jedi for completions in a kernel, but should be used with caution. Jedi can hang the kernel preventing exeuction from happening.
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -2000,6 +2000,12 @@
                     "default": false,
                     "description": "Add PYTHONNOUSERSITE to kernels before starting. This prevents global/user site-packages from being used in the PYTHONPATH of the kernel.",
                     "scope": "machine"
+                },
+                "jupyter.enablePreciseKernelCompletions": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Enables the use of Jedi to generate more precise kernel completions. This can impact peformance, such as preventing executions from running. Use with caution.",
```
IanMatthewHuff, line 2007: I think this works, but question. Do most users even have a clue what Jedi is?

IanMatthewHuff: Actually this is probably good as is, at least they could look it up.

IanMatthewHuff: Possible alternate name / wording:

enableExtendedKernelCompletions - "Turns on the Jedi language server to provide extended IntelliSense completions for running Jupyter notebooks. This can greatly impact notebook cell execution performance. Use with caution."

Honestly not sure if that's better so take it or leave it :). Kinda liked extended over precise since precise sounded more like numerically getting closer to a target.

IanMatthewHuff (approved):

package.json
rchiodo: I like 'extended' too but didn't like the use of Jedi language server. It's not a jedi language server. It's internal to the kernel.  Maybe I should mention the config setting this changes.

wjandrea:
> We disable Jedi by default though because it causes perf problems.

You're talking about running Jedi on the Jupyter/IPython kernel, right? If I'm running Jedi as my IDE language server, that's separate?

wjandrea:
NVM, just noticed your comment [here](https://github.com/microsoft/vscode-jupyter/pull/10137#discussion_r882176104):

> It's not a jedi language server. It's internal to the kernel.

cheers

PR #10136, opened 2022-05-25
Warn on usage of skip on tests and suites

DonJayamanne:

PR #10138, opened 2022-05-25
await for exported file open on export commands

rebornix:
We didn't wait for the exported file to be opened in the export commands, thus the tests might fail as they are checking `activeTextEditor` for validation.

### DIFF
--- a/src/platform/export/fileConverter.ts
+++ b/src/platform/export/fileConverter.ts
@@ -111,7 +111,7 @@ export class FileConverter implements IFileConverter {
             // As far as this method is concerned the export was successful, whether the user opens the file or not
             // should not have any bearing on the completion (resolving) this method.
             // Hence don't await.
-            this.openExportedFile(format, target).catch(noop);
+            await this.openExportedFile(format, target).catch(noop);
         }
     }
 


src/platform/export/fileConverter.ts
```diff
@@ -111,7 +111,7 @@ export class FileConverter implements IFileConverter {
             // As far as this method is concerned the export was successful, whether the user opens the file or not
             // should not have any bearing on the completion (resolving) this method.
             // Hence don't await.
```
rebornix, line 113: I should update this comment, but I don't understand why the tests don't fail before.

rchiodo: Cause Don had added a 'suite.only' to a different test. So only his test was running. The export test wasn't running at all.

See this PR where Don fixes that:
https://github.com/microsoft/vscode-jupyter/pull/10136#pullrequestreview-985488189

src/platform/export/fileConverter.ts
```diff
@@ -111,7 +111,7 @@ export class FileConverter implements IFileConverter {
             // As far as this method is concerned the export was successful, whether the user opens the file or not
```
rchiodo, line 111: This comment seems wrong now.

src/platform/export/fileConverter.ts
```diff
@@ -111,7 +111,7 @@ export class FileConverter implements IFileConverter {
...
             // should not have any bearing on the completion (resolving) this method.
             // Hence don't await.
-            this.openExportedFile(format, target).catch(noop);
+            await this.openExportedFile(format, target).catch(noop);
```
rchiodo, line 114: Do we want to let this throw if it fails?

src/platform/export/fileConverter.ts
rchiodo: You probably need his change to verify this is working.

DonJayamanne: I don't think we need this, I hvae a fix, didn't realize my PR was meged, I'll submit a PR with the fxies.

DonJayamanne: There are a few other bugs in export (basically it never uses nbconvert today)

DonJayamanne: Here's the PR https://github.com/microsoft/vscode-jupyter/pull/10140

DonJayamanne: > Cause Don had added a 'suite.only' to a different test. So only his test was running. The ex

Whats interesting is i added that skip only yesterday, but these tests have been around before then. weird.

src/platform/export/fileConverter.ts
DonJayamanne: > hus the tests might fail as they are checking activeTextEditor for validation.

Not true, we wait for actext text editor to change, instead of checking if its different, hence even if the file opens 2 seconds later we're still waiting for that (i.e. its an async wait, instead of a sync check for active editor)

rebornix:
Closing in favor of https://github.com/microsoft/vscode-jupyter/pull/10140

PR #10140, opened 2022-05-25
Fixes to nbconvert and tests

DonJayamanne:
Fix export to honour nbconvert and fix tests

### DIFF
--- a/src/platform/common/types.ts
+++ b/src/platform/common/types.ts
@@ -88,7 +88,7 @@ export interface IJupyterSettings {
     readonly stopOnFirstLineWhileDebugging: boolean;
     readonly textOutputLimit: number;
     readonly magicCommandsAsComments: boolean;
-    readonly pythonExportMethod: string;
+    readonly pythonExportMethod: 'direct' | 'commentMagics' | 'nbconvert';
     readonly stopOnError: boolean;
     readonly remoteDebuggerPort: number;
     readonly colorizeInputBox: boolean;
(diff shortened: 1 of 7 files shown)


src/platform/common/types.ts
```diff
@@ -88,7 +88,7 @@ export interface IJupyterSettings {
     readonly stopOnFirstLineWhileDebugging: boolean;
     readonly textOutputLimit: number;
     readonly magicCommandsAsComments: boolean;
-    readonly pythonExportMethod: string;
+    readonly pythonExportMethod: 'direct' | 'commentMagics' | 'nbconvert';
```
DonJayamanne, line 91: Added typesafety

src/platform/export/fileConverter.ts
```diff
@@ -100,7 +102,10 @@ export class FileConverter implements IFileConverter {
         candidateInterpreter?: PythonEnvironment
     ) {
         // For web, we perform plain export for Python
-        if (format === ExportFormat.python) {
+        if (
+            format === ExportFormat.python &&
+            this.configuration.getSettings(sourceDocument.uri).pythonExportMethod !== 'nbconvert'
```
DonJayamanne, line 107: This was incorrect, not sure how the tests were passing, basically we should be using `nbconvert` when the setting is set to nbconvert. we weren't doing that.
& we also have tests for that, which was broken.

src/test/datascience/notebook/exportFull.vscode.test.ts
```diff
@@ -139,12 +174,12 @@ suite('DataScience - VSCode Notebook - (Export) (slow)', function () {
         await commands.executeCommand('jupyter.exportAsPythonScript');
 
         // Wait until our active document changes
-        await deferred;
+        await deferred.promise;
```
DonJayamanne, line 177: Not sure how tests passed, but we were missing this.

Issue #9503, opened 2022-03-28
Output messages sent from widgets not displayed in cell output


AoifeHughes:
Hi, 

My system:

``` bash
Version: 1.65.2 (Universal)
Commit: c722ca6c7eed3d7987c0d5c3df5c45f6b15e77d1
Date: 2022-03-10T14:33:49.188Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Darwin x64 21.1.0
```

``` bash
 ➜  ~ python --version
Python 3.10.0
 ➜  ~ jupyter --version
Selected Jupyter core packages...
IPython          : 8.0.1
ipykernel        : 6.9.0
ipywidgets       : 7.7.0
jupyter_client   : 7.1.2
jupyter_core     : 4.9.1
jupyter_server   : not installed
jupyterlab       : not installed
nbclient         : 0.5.13
nbconvert        : 6.4.4
nbformat         : 5.2.0
notebook         : 6.4.10
qtconsole        : 5.2.2
traitlets        : 5.1.1
```

Similar issues: 
- https://github.com/microsoft/vscode-jupyter/issues/3767
- https://github.com/jupyter-widgets/ipywidgets/issues/2512

Code used: 

``` python
from IPython.display import display
from ipywidgets import widgets, interactive
button = widgets.Button(description="Run!")
def on_button_clicked(b):
    print('Clicked')
display(button)
button.on_click(on_button_clicked)
```

In Jupyter Web interface this works perfectly. In VScode it does nothing. 

However, the example give here https://github.com/microsoft/vscode-jupyter/issues/3767#issuecomment-625555606 works just fine. So it seems that they only work with decorators? 

I tested this using:

``` python
from IPython.display import display
from ipywidgets import widgets, interactive, interact
button = widgets.Button(description="Run!")
@interact(b=[0,1])
def on_button_clicked(b):
    if b ==1:
        print('Clicked')
``` 

This seemed to work fine. A bit of a hacky workaround, particularly as decorators don't seem to have a method for buttons...

DonJayamanne:
Thanks for filing this issue, I can replicate this at my end, Will try to dig into the issue and see whats causing this.

DonJayamanne:
Took a while to figure out the issue, its a bug at our end.
Difficult to explain but we'll need to:
* Track the comm id associated with the widget displayed
[...]

PR #10141, opened 2022-05-26
Shared kernelconnection wrapper

DonJayamanne:
Part of #9503

### DIFF
--- a/src/kernels/common/baseJupyterSession.ts
+++ b/src/kernels/common/baseJupyterSession.ts
@@ -31,6 +31,7 @@ import { ChainingExecuteRequester } from './chainingExecuteRequester';
 import { getResourceType } from '../../platform/common/utils';
 import { KernelProgressReporter } from '../../platform/progress/kernelProgressReporter';
 import { isTestExecution } from '../../platform/common/constants';
+import { KernelConnectionWrapper } from './kernelConnectionWrapper';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function suppressShutdownErrors(realKernel: any) {
@@ -71,6 +72,13 @@ export class JupyterSessionStartError extends WrappedError {
 }
 
 export abstract class BaseJupyterSession implements IJupyterSession {
+    /**
+     * Keep a single instance of KernelConnectionWrapper.
+     * This way when sessions change, we still have a single Kernel.IKernelConnection proxy (wrapper),
+     * which will have all of the event handlers bound to it.
+     * This allows consumers to add event handlers hand not worry about internals & can use the lower level Jupyter API.
+     */
+    private _wrappedKernel?: KernelConnectionWrapper;
     private _isDisposed?: boolean;
     private readonly _disposed = new EventEmitter<void>();
     protected readonly disposables: IDisposable[] = [];
@@ -83,27 +91,29 @@ export abstract class BaseJupyterSession implements IJupyterSession {
     protected get session(): ISessionWithSocket | undefined {
         return this._session;
     }
-    public get kernel(): Kernel.IKernelConnection | undefined {
-        return this._session?.kernel || undefined;
+    public get kernelId(): string {
+        return this.session?.kernel?.id || '';
     }
+    public get kernel(): Kernel.IKernelConnection | undefined {
+        if (this._wrappedKernel) {
+            return this._wrappedKernel;
+        }
+        if (!this._session?.kernel) {
+            return;
+        }
+        this._wrappedKernel = new KernelConnectionWrapper(this._session.kernel, this.disposables);
+        return this._wrappedKernel;
+    }
+
     public get kernelSocket(): Observable<KernelSocketInformation | undefined> {
         return this._kernelSocket;
     }
     public get onSessionStatusChanged(): Event<KernelMessage.Status> {
         return this.onStatusChangedEvent.event;
     }
-    public get onIOPubMessage(): Event<KernelMessage.IIOPubMessage> {
-        if (!this.ioPubEventEmitter) {
-            this.ioPubEventEmitter = new EventEmitter<KernelMessage.IIOPubMessage>();
-        }
-        return this.ioPubEventEmitter.event;
-    }
-
     public get status(): KernelMessage.Status {
         return this.getServerStatus();
     }
-
-    public abstract get kernelId(): string;
 
     public get isConnected(): boolean {
         return this.connected;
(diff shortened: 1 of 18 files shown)


src/kernels/debugger/kernelDebugAdapterBase.ts
```diff
@@ -89,18 +90,9 @@ export abstract class KernelDebugAdapterBase implements DebugAdapter, IKernelDeb
             this.debugCell = notebookDocument.cellAt(configuration.__cellIndex!);
         }
 
+        this.jupyterSession.kernel?.iopubMessage.connect(this.onIOPubMessage, this);
         this.disposables.push(
-            this.jupyterSession.onIOPubMessage(async (msg: KernelMessage.IIOPubMessage) => {
```
DonJayamanne, line 93: This is what I'd like to fix in this PR.
We need a way to hook up to the iopub and anymessage events in the underlyiing Jupyter kernel connection.
This property `onIOPubMessage` is a custom event we wire up. If I were to introduce anyMessage then i'd have to handle kernel restarts etc... As we expose the underlying kernel everywhere it makes sense to just hook to those events.

Hence created a kernel wrapper so that we have a single instance, of the Kernel.KernleConnection on which we can monitor the events.

src/kernels/common/baseJupyterSession.ts
```diff
@@ -83,28 +91,30 @@ export abstract class BaseJupyterSession implements IJupyterSession {
...
+        return this._wrappedKernel;
     }
+
     public get kernelSocket(): Observable<KernelSocketInformation | undefined> {
         return this._kernelSocket;
     }
     public get onSessionStatusChanged(): Event<KernelMessage.Status> {
         return this.onStatusChangedEvent.event;
     }
-    public get onIOPubMessage(): Event<KernelMessage.IIOPubMessage> {
```
DonJayamanne, line 95: We need a way to hook up to the `iopub` and `anymessage` events in the underlyiing Jupyter kernel connection.
This property `onIOPubMessage` is a custom event we wire up. If I were to introduce `anyMessage` event then i'd have to handle kernel restarts all over again, tomorrow if we wanted to handle other events, then all over gain etc... 

Also, as we expose the underlying jupyter kernel (Kernel.IKenrelConnection) everywhere it makes sense to just hook to those events.

Hence created a kernel wrapper so that we have a single instance, of the Kernel.KernleConnection on which we can monitor the events (basically moved onIOPubMessage & similar events into the kernel.IKernelConnection and have that keep track of all the event handlers even when kernels restart)

src/kernels/debugger/kernelDebugAdapterBase.ts
```diff
@@ -89,18 +90,9 @@ export abstract class KernelDebugAdapterBase implements DebugAdapter, IKernelDeb
...
-                const anyMsg = msg as any;
-                traceInfoIfCI(`Debug IO Pub message: ${JSON.stringify(msg)}`);
-                if (anyMsg.header.msg_type === 'debug_event') {
-                    this.trace('event', JSON.stringify(msg));
-                    if (!(await this.delegate?.willSendEvent(anyMsg))) {
-                        this.sendMessage.fire(msg.content);
-                    }
-                }
-            })
+            new Disposable(() => this.jupyterSession.kernel?.iopubMessage.disconnect(this.onIOPubMessage, this))
```
DonJayamanne, line 95: Now we can hook up to the underlying Jupyter API events.

src/platform/api/kernelConnectionWrapper.ts
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import type { Kernel } from '@jupyterlab/services';
+import { IDisposable } from '../common/types';
+import { noop } from '../common/utils/misc';
+import { BaseKernelConnectionWrapper } from '../../kernels/jupyter/baseKernelConnectionWrapper';
+import { IKernel } from '../../kernels/types';
+
+export class KernelConnectionWrapper extends BaseKernelConnectionWrapper {
```
DonJayamanne, line 10: wrapper around IKernel for 3rd party API.

src/kernels/raw/session/rawSocket.node.ts
```diff
@@ -100,7 +101,7 @@ export class RawSocket implements IWebSocketLike, IKernelSocket, IDisposable {
         // If from ipywidgets, this will be serialized already, so turn it back into a message so
         // we can add the special hash to it.
         const message = this.deserialize(data);
-
+        this.onAnyMessage(message as any);
```
DonJayamanne, line 104: When we get messages from the UI kenel, we send that directly to the kernel bypassing the Jupyter lab API
As a result we don't invoke the event `anymessage` (which happens in jupyter code when messages are sent to the kernel)

For https://github.com/microsoft/vscode-jupyter/issues/9503, we need the ability to monitor all messagesa & anymessage event handler will pick up some of these messages which only works in Jupyter & not in raw kernels.

Basically this fixes an issue.

src/kernels/raw/session/rawJupyterSession.node.ts
```diff
@@ -51,9 +51,6 @@ export class RawJupyterSession extends BaseJupyterSession {
         }
         return super.status;
     }
-    public get kernelId(): string {
-        return this.session?.kernel?.id || '';
```
DonJayamanne, line 55: Duplicate code in both child classes, moved to base class

src/kernels/jupyter/session/jupyterSession.ts
```diff
@@ -76,15 +69,6 @@ export class JupyterSession extends BaseJupyterSession implements IJupyterServer
         // Wait for idle on this session
         return this.waitForIdleOnSession(this.session, timeout);
     }
-
-    public override get kernel(): Kernel.IKernelConnection | undefined {
-        return this.session?.kernel || undefined;
```
DonJayamanne, line 81: Duplicate code in both child classes, moved to base class

src/kernels/jupyter/baseKernelConnectionWrapper.ts
```diff
@@ -74,27 +57,8 @@ export class KernelConnectionWrapper implements Kernel.IKernelConnection {
         return this.possibleKernelConnection;
     }
 
-    constructor(readonly kernel: IKernel, disposables: IDisposable[]) {
-        const emiStatusChangeEvents = () => {
-            this.statusChanged.emit(kernel.status);
-            if (kernel.status === 'dead' && !kernel.disposed && !kernel.disposing) {
-                this.connectionStatusChanged.emit('disconnected');
```
rchiodo, line 81: Where did this logic go?

rchiodo: Assuming we still need to track connection status.

rchiodo: Oh never mind. This moved to the new class.

amunger (approved):

src/kernels/raw/session/rawSocket.node.ts
rchiodo: Can you put this statement as a comment? I'm probably going to wonder why we're calling this.onAnyMessage here later on.

PR #10142, opened 2022-05-26
Proposal: Isomorphic pythonVariableRequester and debuggerVariables.ts

sadasant:
This PR is a continuation of #10092

In this PR, rather than making the `constants.node.ts` work with URIs, I'm removing any reference to the root in that file and moving those references to the node file, and I'm also using these on the `pythonVariableRequester` to make non-node-specific.

How far from the goalpost am I in this PR?
Any feedback appreciated.

If this approach is ok, this might be my last PR before the web variable viewer.

---

Update: Thanks to the feedback provided now I have also made debuggerVariables isomorphic!

### DIFF
--- a/src/kernels/variables/debuggerVariables.node.ts
+++ b/src/kernels/variables/debuggerVariables.node.ts
@@ -10,7 +10,7 @@ import { getAssociatedNotebookDocument } from '../../notebooks/controllers/kerne
 import { IDebugService, IVSCodeNotebook } from '../../platform/common/application/types';
 import { DataFrameLoading, GetVariableInfo } from '../../platform/common/scriptConstants';
 import { traceError } from '../../platform/logging';
-import { IConfigurationService, Resource } from '../../platform/common/types';
+import { IConfigurationService, IExtensionContext, Resource } from '../../platform/common/types';
 import { DebugLocationTracker } from '../debugger/debugLocationTracker';
 import { sendTelemetryEvent } from '../../telemetry';
 import { Identifiers, Telemetry } from '../../webviews/webview-side/common/constants';
@@ -24,7 +24,6 @@ import {
     IJupyterVariablesResponse
 } from './types';
 import { convertDebugProtocolVariableToIJupyterVariable, DataViewableTypes } from './helpers';
-import { EXTENSION_ROOT_DIR } from '../../platform/constants.node';
 import { IFileSystemNode } from '../../platform/common/platform/types.node';
 
 const KnownExcludedVariables = new Set<string>(['In', 'Out', 'exit', 'quit']);
@@ -49,7 +48,8 @@ export class DebuggerVariables
         @inject(IDebuggingManager) private readonly debuggingManager: IDebuggingManager,
         @inject(IConfigurationService) private configService: IConfigurationService,
         @inject(IVSCodeNotebook) private readonly vscNotebook: IVSCodeNotebook,
-        @inject(IFileSystemNode) private readonly fs: IFileSystemNode
+        @inject(IFileSystemNode) private readonly fs: IFileSystemNode,
+        @inject(IExtensionContext) private readonly context: IExtensionContext
     ) {
         super(undefined);
         this.debuggingManager.onDoneDebugging(() => this.refreshEventEmitter.fire(), this);
@@ -319,8 +319,8 @@ export class DebuggerVariables
             // Run our dataframe scripts only once per session because they're slow
             const key = this.debugService.activeDebugSession?.id;
             if (key && !this.importedDataFrameScriptsIntoKernel.has(key)) {
-                const scriptPath = path.join(EXTENSION_ROOT_DIR, DataFrameLoading.ScriptPath);
-                const contents = await this.fs.readLocalFile(scriptPath);
+                const scriptPath = DataFrameLoading.getScriptPath(this.context);
+                const contents = await this.fs.readLocalFile(scriptPath.fsPath);
                 await this.evaluate(contents);
                 this.importedDataFrameScriptsIntoKernel.add(key);
             }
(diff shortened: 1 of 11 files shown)


src/kernels/variables/debuggerVariables.node.ts
```diff
@@ -316,7 +319,9 @@ export class DebuggerVariables
             // Run our dataframe scripts only once per session because they're slow
             const key = this.debugService.activeDebugSession?.id;
             if (key && !this.importedDataFrameScriptsIntoKernel.has(key)) {
-                await this.evaluate(DataFrameLoading.DataFrameSysImport);
+                const scriptPath = path.join(EXTENSION_ROOT_DIR, DataFrameLoading.ScriptPath);
```
IanMatthewHuff, line 322: Could debuggerVariables also be made isomorphic at this point? At a quick glance it looked like the same usage of scripts as pythonVariableRequester.

src/test/datascience/widgets/notebooks/bqplot_widgets.ipynb
```diff
@@ -7,13 +7,11 @@
                 "# Prerequisites\n",
                 "\n",
                 "pip install bqplot"
-            ],
-            "execution_count": null,
-            "outputs": []
+            ]
```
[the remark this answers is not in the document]
sadasant: Yes, thank you!

src/kernels/variables/debuggerVariables.node.ts
sadasant: Checking!

src/webviews/extension-side/dataviewer/jupyterVariableDataProvider.ts
```diff
@@ -115,7 +115,7 @@ export class JupyterVariableDataProvider implements IJupyterVariableDataProvider
                 type: variable.type,
                 maximumRowChunkSize: variable.maximumRowChunkSize,
                 name: variable.name,
-                fileName: variable.fileName
+                fileName: variable.fileName?.path
```
rchiodo, line 118: It would be better to use the helper `getFilePath` function here.

src/platform/common/scriptConstants.ts
```diff
@@ -1,11 +1,9 @@
-import { EXTENSION_ROOT_DIR } from '../constants.node';
-import * as path from '../../platform/vscode-path/path';
+import * as path from '../vscode-path/path';
 
 export * from './constants';
 
 export namespace DataFrameLoading {
-    export const SysPath = path.join(EXTENSION_ROOT_DIR, 'pythonFiles', 'vscode_datascience_helpers', 'dataframes');
-    export const DataFrameSysImport = `import sys\nsys.path.append("${SysPath.replace(/\\/g, '\\\\')}")`;
+    export const SysPath = path.join('pythonFiles', 'vscode_datascience_helpers', 'dataframes');
```
rchiodo, line 6: Maybe the name here should change to ScriptSubPath? It's not sys path anymore.

rchiodo: Not sure why you'd even need this? Couldn't you just have ScriptPath? (Which I think would make sense to rename to ScriptSubPath so that callers have some indication they need to combine it with the extensionUri. 

Personally I'd prefer callers not to have to think about that at all. Like maybe this changes from a namespace to a static class and then there's a function `getScriptPath(extensionContext: IExtensionContext): string` that does the combination.

That should make `debuggerVariables.node.ts` to be able to be turned into just `debuggerVariables.ts` because it wouldn't need anything node specific anymore

src/kernels/variables/debuggerVariables.node.ts
```diff
@@ -319,8 +319,8 @@ export class DebuggerVariables
             // Run our dataframe scripts only once per session because they're slow
             const key = this.debugService.activeDebugSession?.id;
             if (key && !this.importedDataFrameScriptsIntoKernel.has(key)) {
-                const scriptPath = path.join(EXTENSION_ROOT_DIR, DataFrameLoading.ScriptPath);
-                const contents = await this.fs.readLocalFile(scriptPath);
+                const scriptPath = DataFrameLoading.getScriptPath(this.context);
```
rchiodo, line 322: YOu should be able to make this entire file 'debuggerVariables.ts'? Or did you want to wait for the next PR?

rchiodo (approved):

src/kernels/variables/debuggerVariables.node.ts
sadasant: I'm on it, I'm on it! Next push 🙂

src/test/datascience/widgets/notebooks/bqplot_widgets.ipynb
```diff
@@ -7,11 +7,13 @@
                 "# Prerequisites\n",
                 "\n",
                 "pip install bqplot"
-            ]
+            ],
```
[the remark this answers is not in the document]
sadasant: I keep using this file to test, then I add my changes carelessly 🤦‍♂️ I'll stop doing this 😆

src/kernels/variables/debuggerVariables.ts
```diff
@@ -116,7 +119,7 @@ export class DebuggerVariables
         if (this.active) {
             // Note, full variable results isn't necessary for this call. It only really needs the variable value.
             const result = this.lastKnownVariables.find((v) => v.name === name);
-            if (result && kernel?.resourceUri?.fsPath.endsWith('.ipynb')) {
+            if (result && kernel?.resourceUri?.path.endsWith('.ipynb')) {
```
IanMatthewHuff, line 122: This was part of the existing code already and I don't think that it's much of an issue. But since we have the vscode-path path already I think that path.extname is a safer way to get this.

sadasant: I like that! Thank you!

IanMatthewHuff (approved):
Approving with one possible area to checkout.

PR #10148, opened 2022-05-26
Fix issue with password input for remote connections not being accepted.

### LINKED ISSUE #10103: Changing to a Jupyter server with a password doesn't "take" until you do it a second time
1. Start connected to local
2. Start up a Jupyter server with a password
3. Connect to that server and input the password
4. 
image
5. Note at this point we are still connected to local:
image
6. Server list doesn't have the new connection
7. Now connect to that server a second time
8. Password will not be asked (as it's saved)
9. Now we are in remote mode and the server is in the list

IanMatthewHuff:
Fixes #10103

### DIFF
--- a/src/kernels/jupyter/launcher/jupyterPasswordConnect.ts
+++ b/src/kernels/jupyter/launcher/jupyterPasswordConnect.ts
@@ -43,10 +43,7 @@ export class JupyterPasswordConnect implements IJupyterPasswordConnect {
         }
 
         // Add on a trailing slash to our URL if it's not there already
-        let newUrl = url;
-        if (newUrl[newUrl.length - 1] !== '/') {
-            newUrl = `${newUrl}/`;
-        }
+        const newUrl = addTrailingSlash(url);
 
         // See if we already have this data. Don't need to ask for a password more than once. (This can happen in remote when listing kernels)
         let result = this.savedConnectInfo.get(newUrl);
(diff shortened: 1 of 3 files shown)


src/kernels/jupyter/launcher/jupyterPasswordConnect.ts
```diff
@@ -482,7 +479,17 @@ export class JupyterPasswordConnect implements IJupyterPasswordConnect {
     // When URIs are removed from the server list also remove them from
     private onDidRemoveUris(uris: string[]) {
         uris.forEach((uri) => {
-            this.savedConnectInfo.delete(uri);
+            const newUrl = addTrailingSlash(uri);
+            this.savedConnectInfo.delete(newUrl);
```
IanMatthewHuff, line 483: Don't usually like to do this, but snuck in a small related fix as the same time. Just something that I ran into fixing the other issue. Fixing that if you type in 'http://localhost:8888' it wouldn't clear the password (due to the trailing slash getting added).

src/platform/common/utils/multiStepInput.ts
```diff
@@ -224,9 +224,15 @@ export class MultiStepInput<S> implements IMultiStepInput<S> {
                     }),
                     input.onDidHide(() => {
                         (async () => {
-                            reject(
-                                shouldResume && (await shouldResume()) ? InputFlowAction.resume : InputFlowAction.cancel
-                            );
+                            // If we are busy we might be validating, which might pop up new UI like the password UI, which triggers a hide here
+                            // In that case don't reject, promise can wait and continue after validation is done
+                            if (!input.busy) {
```
IanMatthewHuff, line 229: So the issue was throwing me a bit at first. Stuck a catch around our input.showInputBox call in serverSelector.ts since it wasn't returning and saw that something was getting thrown, but not an Error, but an InputFlowAction.
image
Recently we changed our validation to happen when the URL was committed. For the password case, this validation can pop up another input and hide gets called on the server selection in progress. Then the InputFlowAction gets chucked as an error since the multistep input box uses that for navigation. Fix here was to check for the busy case (which only happens while validating) then in that case don't reject the promise if we get hidden.

rchiodo (approved):

Issue #10152, opened 2022-05-26
Circular dependencies / Tangles between modules


rebornix:
As we introduced the top/primitive components via https://github.com/microsoft/vscode-jupyter/issues/8976 and moved the moduels/helpers to their logical folders, we had circular dependencies between the top components. Below is a simplified version of component dependnecies generated by Structure101

<img height=500 src="https://user-images.githubusercontent.com/876920/170587008-823b5235-b2bc-44f4-b127-53c28ad26d66.png" />

We want to gradually remove the dependencies in wrong directions, including

* No dependencies on `notebooks`/`iw`/`webview` in `platform`
* No dependencies on `iw` in notebooks
* Move feature code/helpers from `telemetry` to feature components. `notebooks`/`iw`/`webview` depends on `telemetry` but not the other way around
* ...

With above items tackled we could end up with an ideal architecture like below (we still need to revisit if it fits)

<img height=600 src="https://user-images.githubusercontent.com/876920/170587745-1eea3167-ad28-4945-aaed-57049ee26ca3.png" />

rebornix:
While moving towards the ideal infrastructure, we might find our changes/PRs get really large even for fixing a single layer breaking relationship (e.g., remove `webviews` dependencies from `kernels`). To ensure us together contribute to the refactoring without stepping on each other's feet too often, we could take steps, from straightforward ones (`telemetry`, `api`), to layer protection through eslint rules and lastly a set of new top components.

## Separate feature and infra code

- [ ] Telemetry should only depend on `platform` component
  - [ ] Move `InterpreterPackageTracker` and `ImportTracker` out of `telemetry`
  - [ ] Move `sendKernelListTelemetry` and kernel types from `telemetry` to `kernels`
  - [ ] Move `splitMultilineString` to `platform` or helper (currently it's imported from `webviews`)
- [ ] API currently consists of core typings and API exposed to external extensions, these should be split into two components
  - [ ] Core typings in `platform`
  - [ ] Top component as `api`. No dependencies from `notebooks/kernel/webviews/iw` to `api`
- [ ] Logging
  - [ ] Move `GitHubIssueCodeLensProvider` and `GitHubIssueCommandListener` out of `platform/logging`
  - [ ] Move `LogReplayService` out of `intellisense` 

## Layer protections

- [ ] :runner: no dependents on `interactive-window` 
[...]

PR #10153, opened 2022-05-26
No interactive window dependencies in notebook

rebornix:
Part of https://github.com/microsoft/vscode-jupyter/issues/10152

This PR includes two major changes:

* Instead of asking interactive window what's the associated editor/document for a resource Uri, the interactive window module register a provider which can help find editor/document 6d1c30e9f112149b4fcc1efc72ce0e1b49e8d566
* Instead of checking if a document is interactive window and then read `code` from cell's special metadata, the interactive window will pass in source code override when requesting cell execution 692233057f86b469bd5cd9de7d43846c80be68a7

### DIFF
--- a/src/interactive-window/types.ts
+++ b/src/interactive-window/types.ts
@@ -46,6 +46,7 @@ export interface IInteractiveWindowProvider {
      * @param owner The URI of a text document which may be associated with an interactive window.
      */
     get(owner: Uri): IInteractiveWindow | undefined;
+    getActiveInteractiveWindow(): IInteractiveWindow | undefined;
 }
 
 export interface IInteractiveBase extends Disposable {
--- a/src/notebooks/notebookEditorProvider.ts
+++ b/src/notebooks/notebookEditorProvider.ts
@@ -8,10 +8,14 @@ import { Uri, NotebookData, NotebookCellData, NotebookCellKind } from 'vscode';
 import { IVSCodeNotebook } from '../platform/common/application/types';
 import { PYTHON_LANGUAGE } from '../platform/common/constants';
 import '../platform/common/extensions';
+import { Resource } from '../platform/common/types';
+import { getResourceType } from '../platform/common/utils';
+import { getComparisonKey } from '../platform/vscode-path/resources';
 import { captureTelemetry } from '../telemetry';
 import { Telemetry, defaultNotebookFormat } from '../webviews/webview-side/common/constants';
 import { JupyterNotebookView } from './constants';
-import { INotebookEditorProvider } from './types';
+import { IEmbedNotebookEditorProvider, INotebookEditorProvider } from './types';
+import { getOSType, OSType } from '../platform/common/utils/platform';
 
 /**
  * Notebook Editor provider used by other parts of DS code.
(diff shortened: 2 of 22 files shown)


rchiodo (approved):

src/notebooks/notebookEditorProvider.ts
```diff
@@ -46,4 +51,52 @@ export class NotebookEditorProvider implements INotebookEditorProvider {
         const doc = await this.vscodeNotebook.openNotebookDocument(JupyterNotebookView, data);
         await this.vscodeNotebook.showNotebookDocument(doc);
     }
+
+    registerEmbedNotebookProvider(provider: IEmbedNotebookEditorProvider): void {
+        this.providers.add(provider);
+    }
+
+    findNotebookEditor(resource: Resource) {
+        const key = resource ? getComparisonKey(resource, true) : 'false';
```
amunger, line 60: will any actually use the comparison key 'false'?

amunger: or the next several lines just be skipped

amunger: oh this was all just moved, probably just don't worry about it for now then

amunger (approved):

src/interactive-window/types.ts
```diff
@@ -46,6 +46,7 @@ export interface IInteractiveWindowProvider {
      * @param owner The URI of a text document which may be associated with an interactive window.
      */
     get(owner: Uri): IInteractiveWindow | undefined;
+    getActiveInteractiveWindow(): IInteractiveWindow | undefined;
```
DonJayamanne, line 49: Isn't this a duplicate of the `activeWindow` property?

DonJayamanne (changes requested):
Looks like we have introduced a duplicate method, rest looks fine

src/interactive-window/types.ts
DonJayamanne: Looking at the interface there doesn't seem to be any difference , except one is a property & the other (new code) is a method

src/notebooks/notebookEditorProvider.ts
rebornix: I don't know, it was copied from the original implementation though

src/interactive-window/types.ts
rebornix: @DonJayamanne they are different in terms of implementation. `getActiveInteractiveWindow` checks `activeWindow` first, and then check `activeTextEditor`  and try to find the interactive window through resource uri mapping.

The name `active` here is a bit ambiguous but it was not introduced in this PR. I'll suggest we tackle it in another PR.

rebornix:
@DonJayamanne we didn't introduce a duplicated function, it's actually implemented differently and moved from a helper module into `InteractiveWindowProvider`. We can see how to improve the `active`* confusion in following PRs.

src/interactive-window/types.ts
rebornix: My current feeling is `InteractiveWindowProvider.activeWindow` should be renamed to `focusedWindow` (as it's checking `activeNotebookEditor`) but I might need to verify all scenarios to see if that's valid.

DonJayamanne: I think we should rename them to be more clear. Personally I'd rather have just one way to get active.

DonJayamanne (approved):

PR #10155, opened 2022-05-27
Move cell exec result handler into a seprate class (part 2)

DonJayamanne:
To be reviewed after https://github.com/microsoft/vscode-jupyter/pull/10141

Part 2 of https://github.com/microsoft/vscode-jupyter/pull/10141

**Cell Execution class does the following:**
* Executes code, updates the cell status, etc
* Handles the messages received from Jupyter in response to the execution request & then updates the output

The second step has been moved into a separate class.


src/notebooks/execution/cellExecutionMessageHandlerFactory.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Kernel } from '@jupyterlab/services';
+import { NotebookCell, NotebookController } from 'vscode';
+import { ITracebackFormatter } from '../../kernels/types';
+import { IApplicationShell } from '../../platform/common/application/types';
+import { disposeAllDisposables } from '../../platform/common/helpers';
+import { IDisposable, IExtensionContext } from '../../platform/common/types';
+import { CellOutputDisplayIdTracker } from './cellDisplayIdTracker';
+import { CellExecutionMessageHandler } from './cellExecutionMessageHandler';
+
+export class CellExecutionMessageHandlerFactory {
```
DonJayamanne, line 13: We have a 1-1 relationship between the IKernel and this class.
I.e. if we change the kernel, then this class dies and we end up with a whole new factory.

src/platform/common/refBool.ts
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
 export class RefBool {
```
DonJayamanne, line 4: Surprised how the linters didn't catch this, this was used in the cell execution code.

src/notebooks/execution/cellExecutionMessageHandlerFactory.ts
rchiodo: Do you mean the controller? It doesn't seem to point to the kernel.

src/notebooks/execution/cellExecutionMessageHandler.ts
```diff
@@ -0,0 +1,574 @@
...
+     * Or when execution has been cancelled.
+     */
+    public dispose() {
+        traceCellMessage(this.cell, 'Execution disposed');
+        disposeAllDisposables(this.disposables);
+        this.prompts.forEach((item) => item.dispose());
+        this.prompts.clear();
+    }
+    public endCellExecution() {
+        this.prompts.forEach((item) => item.dispose());
```
rchiodo, line 138: Wouldn't dispose call this function?

src/notebooks/execution/cellExecution.ts
```diff
@@ -503,26 +355,22 @@ export class CellExecution implements IDisposable {
...
-        };
-        request.onReply = (msg) => {
-            // Cell has been deleted or the like.
-            if (this.cell.document.isClosed) {
-                request.dispose();
-            }
-            this.handleReply(clearState, msg);
-        };
-        request.onStdin = this.handleInputRequest.bind(this, session);
+        this.cellExecutionHandler = this.factory.create(this.cell, {
```
rchiodo, line 358: I'm wondering if the 'cellExecutionHandler' could just have a single method on it, 'dispose()' that's called when ending. 

Otherwise you could push the onErrorHandlingExecuteRequestIOPubMessage into the factory.create (don't need to sign up for it separately).

Then the factory would be more like: registerListener or something that returns a disposable. Essentially it's listening to the request.

rchiodo: Then all you need to do in 'endCellTask' (the only spot that uses cellExecutionHandler)  is stop listening - dispose the listener.

src/notebooks/execution/cellExecutionMessageHandlerFactory.ts
DonJayamanne: sorry yes

src/notebooks/execution/cellExecutionMessageHandler.ts
DonJayamanne: that too yes, but when cell execution ends we don't want to dispose this straight away. That's more clear in the second PR, shoot, should have moved this to the next PR.

This is what we need:
* When we're executing a cell, this class will handle all of the outptu.
* Once execution completes, we need it to stop handling prompts and the like, hence we call endCellExecution
* But we don't want to dispose it just yet (in PR 3) - we stop disposing it
* Sorry, I shoulld have created the method endCellExecution in PR 3
* In PR 3, what we do is when execution ends, we call endCelLExecution
* However that class lives on and keeps monitoring the  iopub messages
* Dispose is called when the cell is deleted.

src/notebooks/execution/cellExecution.ts
DonJayamanne: Done, 
* Left dipose method so that this is only disposed when the cellExecution dies (called durijng dispose of cellExecution)
* endCellExecution is called internally by this class when `request.done` completes (either rejects or resolves)
* that way calling code (CellExecution) doesn't need to call anything)

DonJayamanne: Basically `endCellExecution` is no longer a public method, & as suggested we only call dispose.

rchiodo: What do you think of changing it to registerListener or maybe watchRequest? Instead of a factory.create? It's more like you're registering something to listen/watch a request? The semantics of 'create' seem strange to me.

DonJayamanne: Fixed

src/notebooks/execution/cellExecutionMessageHandler.ts
```diff
@@ -0,0 +1,574 @@
...
+        this.clearLastUsedStreamOutput();
+        this.execution = undefined;
+    }
+    private startHandlingExecutionMessages(
+        request: Kernel.IShellFuture<KernelMessage.IExecuteRequestMsg, KernelMessage.IExecuteReplyMsg>
+    ) {
+        request.onIOPub = (msg) => {
+            // Cell has been deleted or the like.
+            if (this.cell.document.isClosed) {
+                request.dispose();
```
sadasant, line 148: Should we return here?

src/notebooks/execution/cellExecutionMessageHandler.ts
```diff
@@ -0,0 +1,574 @@
...
+            try {
+                this.handleIOPub(msg);
+            } catch (ex) {
+                this._onErrorHandlingIOPubMessage.fire(ex);
+            }
+        };
+        request.onReply = (msg) => {
+            // Cell has been deleted or the like.
+            if (this.cell.document.isClosed) {
+                request.dispose();
```
sadasant, line 159: Same here? I’m just not sure of the effects of continuing.

src/notebooks/execution/cellExecutionMessageHandler.ts
```diff
@@ -0,0 +1,574 @@
...
+        } else {
+            // Undefined for not success or failure
+            this.temporaryExecution?.end(undefined);
+        }
+        this.previousResultsToRestore = undefined;
+        this.temporaryExecution = undefined;
+    }
+    private handleIOPub(msg: KernelMessage.IIOPubMessage) {
+        // eslint-disable-next-line @typescript-eslint/no-require-imports
+        const jupyterLab = require('@jupyterlab/services') as typeof import('@jupyterlab/services');
```
sadasant, line 218: Any benefit in requiring here vs importing from the top? I’m guessing this loads heavy stuff?

DonJayamanne: Yes, thats correct, and we don't want webpack to inline these imports either.

DonJayamanne:
@rchiodo I've merged the other PR into this, hence you should see the full picture on how the code fits into the bigger picture.

rchiodo (approved):

PR #10161, opened 2022-05-27
Fix link clicking in the interactive window

### LINKED ISSUE #10149: Clicking on error links in the interactive window does not go to the same editor
Run some code like so:

```python
#%% Test exception
def raiser():
    print('stuff')
    raise Exception("Inside a <function>")

# %%
raiser()
```

When clicking on the exception links, it opens a new copy of the python file. 

It should open the same one.

rchiodo:
Fixes #10149

Also fixes an offset issue with the links (recent regression).

### DIFF
--- a/src/notebooks/outputs/errorRendererComms.node.ts
+++ b/src/notebooks/outputs/errorRendererComms.node.ts
@@ -59,7 +59,7 @@ export class ErrorRendererCommunicationHandler implements IExtensionSyncActivati
     }
 
     private async openFile(fileUri: string) {
-        const uri = Uri.parse(fileUri);
+        let uri = Uri.parse(fileUri);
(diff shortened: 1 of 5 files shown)


src/notebooks/outputs/errorRendererComms.node.ts
```diff
@@ -68,13 +68,14 @@ export class ErrorRendererCommunicationHandler implements IExtensionSyncActivati
                 const lineNumber = parseInt(lineMatch[1], 10);
                 selection = new Range(new Position(lineNumber, 0), new Position(lineNumber, 0));
             }
+            uri = uri.with({ query: '' });
```
rchiodo, line 71: This fixes the opening a new file. With the query on the uri, the check to match below fails.

sadasant (approved):

PR #10185, opened 2022-05-30
Ensure we select the right controller

### LINKED ISSUE #10046: Test 'When switching from remote to local, then clear the preferred remote kernel' is flaky
Flaky or completely busted

DonJayamanne:
Fixes #10046

This is actually a bug fix, instead of just a fix of the test.
The test was failing due to an actual bug in the code

### DIFF
--- a/src/kernels/jupyter/launcher/liveshare/hostJupyterServer.ts
+++ b/src/kernels/jupyter/launcher/liveshare/hostJupyterServer.ts
@@ -112,7 +112,7 @@ export class HostJupyterServer implements INotebookServer {
                 actionSource
             );
             this.throwIfDisposedOrCancelled(cancelToken);
-            traceInfo(`Started session for kernel ${kernelConnection.id}`);
+            traceInfo(`Started session for kernel ${kernelConnection.kind}:${kernelConnection.id}`);
             return session;
         };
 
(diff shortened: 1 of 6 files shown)


src/kernels/jupyter/launcher/liveshare/hostJupyterServer.ts
```diff
@@ -112,7 +112,7 @@ export class HostJupyterServer implements INotebookServer {
                 actionSource
             );
             this.throwIfDisposedOrCancelled(cancelToken);
-            traceInfo(`Started session for kernel ${kernelConnection.id}`);
+            traceInfo(`Started session for kernel ${kernelConnection.kind}:${kernelConnection.id}`);
```
DonJayamanne, line 115: Having the kind is useful, added this in a number of places

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -673,7 +673,7 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
                     traceVerbose(`Early registration of controller for Kernel connection ${preferredConnection.id}`);
                     this.createNotebookControllers([preferredConnection]);
                 }
-            } else {
+            } else if (document.notebookType === InteractiveWindowView) {
```
DonJayamanne, line 676: This was the bug fix, its not a flaky test issue, even when we had a preferred controller we'd change it to the active interpreter.

rchiodo (approved):

PR #10251, opened 2022-06-01
Run by line and debug cell not stopping

### LINKED ISSUE #10207: Run by line does not work.
## Environment data
Build of Visual Studio Code - Insiders: 1.68.0-insider(user setup)
Build of python-Pre-Release: v2022.7.11511005
Build of jupyter-Pre-Release: v2022.5.1001532360
Build of pylance: v2022.5.3
Build of jupyter Notebook Renderers: v1.0.8
Build of jupyter Keymap: v1.0.0
image

## Steps to reproduce:
1. Opt into the custom editor experiment (not yet supported in native notebooks)
2. Open a new blank Jupyter notebook.
3. Add the following code to a cell:
```
def bar():
    for i in range(5):
         print(i)

def foo():
    bar()

foo()
print('Done')
```
4. Press F10 to step through the code in the cell line-by-line. 

## Expected behaviour
Verify that it first steps over function definitions for foo() and bar(), then steps into the function definitions after foo() is called, then steps back out when the loop in bar() finishes and prints 'Done'.

## Actual behaviour
No response when press F10

## Jupyter Log
image

rchiodo:
Fixes #10207

Breaktpoints firing on the first line more than once is the debugpy issue with Python 3.10. I just need to update debugpy and IPython.


amunger:
did you check IW debugging as well since they share this base class? I think it was still working previously

amunger:
looks like IW still works

amunger (approved):

rchiodo:
Added some more fixes.
- Debugger variable fetching not working because of changes for web support in the variable explorer
- IW debug tests failing because of new 'debug type' for IW debugging with IPython debugger.

src/platform/common/scriptConstants.ts
```diff
@@ -14,11 +14,6 @@ export namespace DataFrameLoading {
 
     export const DataFrameInfoFunc = '_VSCODE_getDataFrameInfo';
     export const DataFrameRowFunc = '_VSCODE_getDataFrameRows';
-
-    // Constants for the debugger which imports the script files
-    export const DataFrameImport = `__import__('vscodeDataFrame')`;
```
rchiodo, line 19: These don't make sense anymore because we don't actually import a file. We can just reference the function name directly now.

src/kernels/variables/debuggerVariables.ts
```diff
@@ -335,7 +336,7 @@ export class DebuggerVariables
             // Run our variable info scripts only once per session because they're slow
             const key = this.debugService.activeDebugSession?.id;
             if (key && !this.importedGetVariableInfoScriptsIntoKernel.has(key)) {
-                const scriptPath = DataFrameLoading.getScriptPath(this.context);
```
rchiodo, line 338: This was a typo that was preventing any variables from loading while debugging.

src/test/datascience/interactiveDebugging.vscode.test.ts
```diff
@@ -47,6 +48,7 @@ suite('Interactive window debugging', async function () {
         traceInfo(`Start Test ${this.currentTest?.title}`);
         api = await initialize();
         disposables.push(vscode.debug.registerDebugAdapterTrackerFactory('python', tracker));
+        disposables.push(vscode.debug.registerDebugAdapterTrackerFactory(pythonIWKernelDebugAdapter, tracker));
```
rchiodo, line 51: Don's change for IW debugging uses a different debug type than 'python', so we need to register a tracker for his new type here in order to watch the debugger in the tests. Depends upon version of ipython though so I left the other tracker there too.

amunger: The IW tests loop through the two debugger types for the entire suite. Was that supposed to happen in here instead maybe?

PR #10263, opened 2022-06-02
Fixes to IW debugging with breakpoints

### LINKED ISSUE #10258: Interactive debugging does not work with ipykernel debugger
Force ipykernel debugger on:
```json
  "jupyter.forceIPyKernelDebugger": true
```
Have code like so (part of one of our tests):
```
# %%
def foo(): 
   print(4)

# %%%
foo()
```

Set a breakpoint on the print(4)
Debug the second cell.

Continue when debugger comes up.

Expected result:
Stops on breakpoint (old debugger does)

Actual result:
Doesn't stop on breakpoint.

The cellToFile mapping that occurs in notebooks doesn't work for an interactive file. It needs more than cell URI to map as it has the same file for all dumped cells. This means breakpoints won't bind because they won't be rewritten to the mapped file.

DonJayamanne:
Fixes #10258

Will add tests next week.


src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -82,14 +90,121 @@ export class KernelDebugAdapter extends KernelDebugAdapterBase {
...
+                    request.arguments.lines = request.arguments.lines.filter(
+                        (line) => line <= cell.metadata.generatedCode!.endLine
+                    );
+                }
+                if (request.arguments.breakpoints) {
+                    request.arguments.breakpoints = request.arguments.breakpoints.filter(
+                        (bp) => bp.line <= cell.metadata.generatedCode!.endLine
+                    );
+                }
+                if (sortedLines.filter((line) => line > cell.metadata.generatedCode!.endLine).length) {
```
DonJayamanne, line 192: @rchiodo we add breakpoints for each cell by splitting the breakpoints by cells.

src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -82,14 +90,121 @@ export class KernelDebugAdapter extends KernelDebugAdapterBase {
...
+    protected override translateDebuggerFileToRealFile(
+        source: DebugProtocol.Source | undefined,
+        lines?: { line?: number; endLine?: number; lines?: number[] }
+    ) {
+        if (!source || !source.path || !lines || (typeof lines.line !== 'number' && !Array.isArray(lines.lines))) {
+            return;
+        }
+        // Find the cell that matches this line in the IW file.
+        const cell = this.cellToDebugFileSortedInReverseOrderByLineNumber.find(
+            (item) => item.debugFilePath === source.path
```
rchiodo, line 118: Comment seems wrong here. It's not searching by line but rather by the translated debugFilePath. That does map to a specific cell entry.

src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
rchiodo: Is breakpoints the only thing that needs this special translation? What about stack frames?

rchiodo:
Tests are actually already there, you just need to undo the changes I made for this issue. Then they'll run the normal tests.

src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
DonJayamanne: that's fine

rchiodo (approved):

rchiodo:
Oh wait I shouldn't have approved. You should undo the changes I made here:
https://github.com/microsoft/vscode-jupyter/blob/6516ddbd9b17d83395f1c9a7cffbc4a5ae39319c/src/test/datascience/interactiveWindow.vscode.test.ts#L53
and here
https://github.com/microsoft/vscode-jupyter/blob/6516ddbd9b17d83395f1c9a7cffbc4a5ae39319c/src/test/datascience/.vscode/settings.json#L33

PR #10303, opened 2022-06-03
Dataframe viewer is broken

### LINKED ISSUE #10302: Data Viewer is broken
Data viewer always shows up empty.

rchiodo:
Fixes #10302 

Root cause was the removal of the CSS provider (necessary for moving to web). However the dataframe viewer was waiting for the CSS message before rendering.

I also removed all of the references to the CSS provider's font data.


src/webviews/webview-side/data-explorer/mainPanel.tsx
```diff
@@ -98,7 +98,7 @@ export class MainPanel extends React.Component<IMainPanelProps, IMainPanelState>
                 fetchedRowCount: -1,
                 filters: {},
                 indexColumn: data.primaryKeys[0],
-                styleReady: false,
```
rchiodo, line 101: This fixes the rendering not happening. I'll enter another issue to remove the CSS messages entirely.

rchiodo:
The font ends up different.

Stable view:

image

After this change:

image

src/webviews/webview-side/interactive-common/mainState.ts
```diff
@@ -57,8 +57,6 @@ const darkStyle = `
             --code-string-color: #ce9178;
             --code-variable-color: #9CDCFE;
             --code-type-color: #4EC9B0;
-            --code-font-family: Consolas, 'Courier New', monospace;
```
amunger, line 60: why are so many font stylings being removed?

rchiodo: Because they're not supported anymore since we got rid of the CSS generator. This was a fallback but the default fallback of '--vscode-font-family' is more appropriate.

amunger (approved):

Issue #10307, opened 2022-06-03
Reduce web bundle size


rebornix:
Currently the unzipped bundle size for Web bits is 4.67MB. Considering that the web extension doesn't support bundle splitting, a large bundle file means long download and evaluation (activation). 

<img width="779" alt="image" src="https://user-images.githubusercontent.com/876920/171951584-41f6b85d-44e7-408f-b82f-b57b1d86423a.png">

There have been quite some good efforts to reduce the bundle size. We can follow the same patterns and also leverage the eslint rules to ensure that we don't overlook them in the future. From a 10000 feet bird view, we can do at least following to bring down the bundle size noticeably:

- [x] Load lodash function modules only https://github.com/microsoft/vscode-jupyter/pull/10308
- [x] Mark `moment.js` as external as it's not used by us, or `jupyterlab` sub-components we depend on
- [x] `semver` is used for Python intepreter and Python extension versions check, we probably don't need this in Web. If so, make sure we are not referecing `semver` in Web related code paths
- [x] `hash`. We use `sha1`, `sha256`, `sha512` only, no need to reference them all. Also we can validate if we need all three sha algorithms.
- [x] `fnv`. Originally introduced for A/B testing in Python https://github.com/microsoft/vscode-jupyter/commit/db297c4a7833e1d5aab453541aa5b1751abd890b. The code sets the default algorithm to `fnv` and now it's used as default way to generate file hash https://github.com/microsoft/vscode-jupyter/blob/main/src/kernels/jupyter/preferredRemoteKernelIdProvider.ts#L55
  - To move off `fnv`, we would need a migration plan to move from `fnv` to simpler hashes as `sha512`.
- External/Ideas
  - Load sub modules from `jupyterlab/services` instead of the whole bundle
  - Replace `url-parse` with inhouse `vscode.Uri.parse`

rebornix:
<img width="728" alt="image" src="https://user-images.githubusercontent.com/876920/173411760-6972b30a-70e8-4573-b9aa-3fa2ebebc35b.png">

We have already reduced the web bundle size by 25%.

DonJayamanne:
Closing this for now, as I do not think anything else needs to beon done here.
At least not for now, if we do, we can create concrete issues with details

PR #10308, opened 2022-06-03
Avoid loading full lodash

rebornix:
Re #10307

image

Size reduced by 0.52 MB. We also introduced a new eslint rule to prevent this from happening in the future.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -265,6 +265,7 @@ module.exports = {
         ],
         'react/jsx-uses-vars': 'error',
         'react/jsx-uses-react': 'error',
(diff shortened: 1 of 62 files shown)


rchiodo (approved):

amunger (approved):

.eslintrc.js
```diff
@@ -265,6 +265,7 @@ module.exports = {
         ],
         'react/jsx-uses-vars': 'error',
         'react/jsx-uses-react': 'error',
+        'no-restricted-imports': ['error', { "paths": ["lodash"] }],
```
DonJayamanne, line 268: Please add `rxjs` as well, as thats big as well.

PR #10329, opened 2022-06-06
Remove unnecessary quoting of args

DonJayamanne:
Found an issue with a kernel, and the root cause was we were adding double quotes to all arguments even when something could have already had double quotes.

E.g. if we have a kernel spec with argv as follows `['connection_file={connectionFile}', 'x="hello world"]`, 
then we update it as follows:
`['connection_file=<replaced with real connection file>', '"x="hello world""]`

All we need to do is update the `{connectionFile}` and leave everything else as is, i.e. its wrong to modify any other variables except for `connectionFile`.

### DIFF
--- a/src/kernels/raw/launcher/kernelProcess.node.ts
+++ b/src/kernels/raw/launcher/kernelProcess.node.ts
@@ -480,20 +480,8 @@ export class KernelProcess implements IKernelProcess {
                     promiseCancellation as Promise<NodeJS.ProcessEnv | undefined>
                 ])
             ]);
-            // Add quotations to arguments if they have a blank space in them.
-            // This will mainly quote paths so that they can run, other arguments shouldn't be quoted or it may cause errors.
             // The first argument is sliced because it is the executable command.
-            const args = this.launchKernelSpec.argv.slice(1).map((a) => {
-                // Some kernel specs (non-python) can have argv as `--connection-file={connection_file}`
-                // The `connection-file` will be quoted when we update it with the real path.
-                if (a.includes('--connection-file')) {
-                    return a;
-                }
-                if (a.includes(' ')) {
-                    return `"${a}"`;
-                }
-                return a;
-            });
+            const args = this.launchKernelSpec.argv.slice(1);
             exeObs = executionService.execObservable(executable, args, {
                 env,
                 cwd: workingDirectory
--- a/src/kernels/raw/launcher/kernelProcess.node.ts
+++ b/src/kernels/raw/launcher/kernelProcess.node.ts
@@ -354,10 +354,10 @@ export class KernelProcess implements IKernelProcess {
             // Remember, non-python kernels can have argv as `--connection-file={connection_file}`,
             // hence we should not replace the entire entry, but just replace the text `{connection_file}`
             // See https://github.com/microsoft/vscode-jupyter/issues/7203
+            const connectionFile = this.connectionFile.includes(' ')
+                ? `"${this.connectionFile}"` // Quoted for spaces in file paths.
+                : this.connectionFile;
             if (this.launchKernelSpec.argv[indexOfConnectionFile].includes('--connection-file')) {
-                const connectionFile = this.connectionFile.includes(' ')
-                    ? `"${this.connectionFile}"` // Quoted for spaces in file paths.
-                    : this.connectionFile;
                 this.launchKernelSpec.argv[indexOfConnectionFile] = this.launchKernelSpec.argv[
                     indexOfConnectionFile
                 ].replace(connectionFilePlaceholder, connectionFile);
@@ -366,7 +366,7 @@ export class KernelProcess implements IKernelProcess {
                 // E.g. in Python the name of the argument is `-f` and in.
                 this.launchKernelSpec.argv[indexOfConnectionFile] = this.launchKernelSpec.argv[
                     indexOfConnectionFile
-                ].replace(connectionFilePlaceholder, this.connectionFile);
+                ].replace(connectionFilePlaceholder, connectionFile);
             }
         }
     }
@@ -409,7 +409,10 @@ export class KernelProcess implements IKernelProcess {
 
         // We still put in the tmp name to make sure the kernel picks a valid connection file name. It won't read it as
         // we passed in the arguments, but it will use it as the file name so it doesn't clash with other kernels.
-        newConnectionArgs.push(`--f=${this.connectionFile}`);
+        const connectionFile = this.connectionFile!.includes(' ')
+            ? `"${this.connectionFile}"` // Quoted for spaces in file paths.
+            : this.connectionFile;
+        newConnectionArgs.push(`--f=${connectionFile}`);
 
         return newConnectionArgs;
     }
--- a/src/test/datascience/kernelProcess.unit.test.ts
+++ b/src/test/datascience/kernelProcess.unit.test.ts
@@ -347,9 +347,9 @@ suite('kernel Process', () => {
             `--Session.signature_scheme="${connection.signature_scheme}"`,
             `--Session.key=b"${connection.key}"`,
             `--shell=${connection.shell_port}`,
-            `--transport="${connection.transport}"`,
+            `--transport=${connection.transport}`,
             `--iopub=${connection.iopub_port}`,
-            `--f=${tempFile}`,
+            `--f="${tempFile}"`,
             `--debug`
         ];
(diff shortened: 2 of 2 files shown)


src/kernels/raw/launcher/kernelProcess.node.ts
```diff
@@ -366,7 +366,7 @@ export class KernelProcess implements IKernelProcess {
                 // E.g. in Python the name of the argument is `-f` and in.
                 this.launchKernelSpec.argv[indexOfConnectionFile] = this.launchKernelSpec.argv[
                     indexOfConnectionFile
-                ].replace(connectionFilePlaceholder, this.connectionFile);
+                ].replace(connectionFilePlaceholder, connectionFile);
```
DonJayamanne, line 369: We weren't quoting this place.

src/kernels/raw/launcher/kernelProcess.node.ts
```diff
@@ -409,7 +409,10 @@ export class KernelProcess implements IKernelProcess {
 
         // We still put in the tmp name to make sure the kernel picks a valid connection file name. It won't read it as
         // we passed in the arguments, but it will use it as the file name so it doesn't clash with other kernels.
-        newConnectionArgs.push(`--f=${this.connectionFile}`);
+        const connectionFile = this.connectionFile!.includes(' ')
+            ? `"${this.connectionFile}"` // Quoted for spaces in file paths.
+            : this.connectionFile;
+        newConnectionArgs.push(`--f=${connectionFile}`);
```
DonJayamanne, line 415: All we need to do is quote the connection file & nothing else.

src/kernels/raw/launcher/kernelProcess.node.ts
```diff
@@ -480,20 +483,8 @@ export class KernelProcess implements IKernelProcess {
...
-            // This will mainly quote paths so that they can run, other arguments shouldn't be quoted or it may cause errors.
             // The first argument is sliced because it is the executable command.
-            const args = this.launchKernelSpec.argv.slice(1).map((a) => {
-                // Some kernel specs (non-python) can have argv as `--connection-file={connection_file}`
-                // The `connection-file` will be quoted when we update it with the real path.
-                if (a.includes('--connection-file')) {
-                    return a;
-                }
-                if (a.includes(' ')) {
-                    return `"${a}"`;
```
DonJayamanne, line 493: This was wrong, we were quoting anything that had spaces.

rchiodo: Seems correct to me? Why wouldn't we?

rchiodo: We should only skip quoting if there was quotes already in the argument. 

What if it looks like this:

```json
"spec": {
        "argv": [
          "C:\\My file\\At\\Spot \\with spaces\\foo.exe",
          "-m",
          "ipykernel_launcher",
          "-f",
          "{connection_file}"
        ],
```

DonJayamanne: If that's the case, then the author of the kernelspec should be quoting that not us.

DonJayamanne: Also, what if me as the author of the kernelspec adds something like
```
"spec": {
        "argv": [
          "C:\\My file\\At\\Spot \\with spaces\\foo.exe",
          "-m",
          "ipykernel_launcher",
          "-f",
          "{connection_file}"
	'some_other_arg="Hello World"'
        ],
```

Now our code comes in and breaks this by adding
```
"spec": {
        "argv": [
          "C:\\My file\\At\\Spot \\with spaces\\foo.exe",
          "-m",
          "ipykernel_launcher",
          "-f",
          "{connection_file}"
	'"some_other_arg="Hello World""'
        ],
```

rchiodo: That's why we have to only do this if there aren't already quotes in the argument.

DonJayamanne: I think thats wrong, we'll never get it right, because Jupyter doesn't do anything like that.
Also quoting should be done only for `connection_file` for stuff that we muck around with, and as for other stuff it shoudl be left as is.

Because if we ended up trying to quote stuff that only needs to be quoted, then 100% of the time its unnecessary as Jupyter doesn't do it, meaning it already works and we're not doing anything either.
& if we did have to quote something, then that means we're doing something wrong, when it already works without the necessary quotes in Jupyter.

E.g. how would one end up quoting something from here:
Here's a perfect example, this is a valid OCAML kernelspec.json file
```json
  "argv": [
    "/bin/sh",
    "-c",
    "eval $(opam env --switch=4.13.1 --shell=sh) && /Users/donjayamanne/.opam/4.13.1/bin/ocaml-jupyter-kernel \"$@\"",
    "ocaml-jupyter-kernel",
    "-init", "/Users/donjayamanne/.ocamlinit",
    "--merlin", "/Users/donjayamanne/.opam/4.13.1/bin/ocamlmerlin",
    "--verbosity", "app",
    "--connection-file", "{connection_file}"
  ]
```
I don't think we can some up with any logic that can make this work.
Hence my argument, we shoudl never be quoting these things.

DonJayamanne: If you look at the last part of the 3rd argument, the author (code that genrated the kernelspec) has already added the necessary double quotes, but we come in and break this completely.

DonJayamanne: With the above example, assume we had a kernel spec that didn't have the last part `\"$@\"",` in the 3rd argument then we'd still add the double quotes (based on the rule that `"` doesn't exist, hence lets add it) when its completely unnecessary.

rchiodo: So I just created a kernelspec from a venv with spaces in it. By running this:

```
python -m ipykernel install --user --name=VenvWithSpaces
```

Meaning not generating it by hand. 

The kernelspec generated by this did not add quotes:

```
"venvwithspaces": {
      "resource_dir": "C:\\Users\\aku91\\AppData\\Roaming\\jupyter\\kernels\\venvwithspaces",
      "spec": {
        "argv": [
          "D:\\Source\\Testing\\Testing_3\\Venv with Spaces\\Scripts\\python.exe",
          "-m",
          "ipykernel_launcher",
          "-f",
          "{connection_file}"
        ],
        "env": {},
        "display_name": "VenvWithSpaces",
        "language": "python",
        "interrupt_mode": "signal",
        "metadata": {
          "debugger": true
        }
      }
    },
```

Does this kernelspec work with your change? (it works with what we have in stable)

Your ocaml example seems like it would be just fine to me. It would either be this:

```
eval $(opam env --switch=4.13.1 --shell=sh) && /Users/donjayamanne/.opam/4.13.1/bin/ocaml-jupyter-kernel \"$@\"
```

Or if the the user hadn't put quotes anywhere, it would be this:

```
"eval $(opam env --switch=4.13.1 --shell=sh) && /Users/donjayamanne/.opam/4.13.1/bin/ocaml-jupyter-kernel $@"
```

DonJayamanne: If you look at your kernelspec, we never quote the first argument, the first argument is not an argument for the process.
Its the process (command) it self, the rest of the values in `argv` are the arguments passed into the executable `D:\\Source\\Testing\\Testing_3\\Venv with Spaces\\Scripts\\python.exe`
Hence quoting of the first value in `argv` is never required

DonJayamanne: > Your ocaml example seems like it would be just fine to me. It would either be this:

Nope, it was getting changed to this & thus breaks kernel startup.
`"eval $(opam env --switch=4.13.1 --shell=sh) && /Users/donjayamanne/.opam/4.13.1/bin/ocaml-jupyter-kernel \"$@\""`

I.e. we were always adding `"` around any item in `argv` if it contained a space, with the assumption that it needs it.

rchiodo: Yes but I'm arguing the bug is that we didn't check for quotes. We shouldn't have added it to that one.

Your argument is we should never add quotes because the kernelspec is already setup to work. Sounds fine now that I realize we're skipping the first argument. That's the one I assume would have spaces in it most of the time.

rchiodo (approved):

PR #10333, opened 2022-06-06
Handling of unhandled Promise rejections

DonJayamanne:
This PR fixes all of the unhandled promise rejections we have.
Also ensures we have a CI check to ensure we don't end up with new promise rejections.

Basically unhandled promises are bad and could potentially kill the process. In the Python extension we used to keep all of these clean (i.e. ensure we never had any unhandled promise rejections).

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -388,3 +388,11 @@ gulp.task('validateDependencies', async () => {
         }
     }
 });
+gulp.task('verifyUnhandledErrors', async () => {
+    const fileName = path.join(__dirname, 'unhandledErrors.txt');
+    const contents = fs.pathExistsSync(fileName) ? fs.readFileSync(fileName, 'utf8') : '';
+    if (contents.trim().length) {
+        console.error(contents);
+        throw new Error('Unhandled errors detected. Please fix them before merging this PR.', contents);
+    }
+});
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -382,12 +382,15 @@ gulp.task('validateDependencies', async () => {
         console.log(modules);
         if (modules.length > 1 || modules[0] !== '@jupyterlab/coreutils') {
             // we already validate that we are not using moment in @jupyterlab/coreutils
-            const message = `The following modules require moment: ${modules.join(', ')}. Please validate if moment is being used.`;
+            const message = `The following modules require moment: ${modules.join(
+                ', '
+            )}. Please validate if moment is being used.`;
             console.error(message);
             throw new Error(message);
         }
     }
 });
+
 gulp.task('verifyUnhandledErrors', async () => {
     const fileName = path.join(__dirname, 'unhandledErrors.txt');
     const contents = fs.pathExistsSync(fileName) ? fs.readFileSync(fileName, 'utf8') : '';
--- a/src/interactive-window/commands/commandRegistry.ts
+++ b/src/interactive-window/commands/commandRegistry.ts
@@ -120,7 +120,7 @@ export class CommandRegistry implements IDisposable, IExtensionSingleActivationS
         this.registerCommand(Commands.RunCurrentCell, this.runCurrentCell);
         this.registerCommand(Commands.RunCurrentCellAdvance, this.runCurrentCellAndAdvance);
         this.registerCommand(Commands.ExecSelectionInInteractiveWindow, (textOrUri: string | undefined | Uri) => {
-            void this.runSelectionOrLine(textOrUri);
+            this.runSelectionOrLine(textOrUri).catch(noop);
         });
         this.registerCommand(Commands.RunAllCellsAbove, this.runAllCellsAbove);
         this.registerCommand(Commands.RunCellAndAllBelow, this.runCellAndAllBelow);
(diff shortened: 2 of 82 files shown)


src/kernels/raw/launcher/kernelProcess.node.ts
```diff
@@ -213,6 +214,9 @@ export class KernelProcess implements IKernelProcess {
 
         // Don't return until our heartbeat channel is open for connections or the kernel died or we timed out
         try {
+            if (deferred.rejected) {
+                await deferred.promise;
```
DonJayamanne, line 218: No point waiting for ports if kernel has already died

src/kernels/raw/session/rawSocket.node.ts
```diff
@@ -278,6 +278,8 @@ export class RawSocket implements IWebSocketLike, IKernelSocket, IDisposable {
                 this.postToSocket(msg.channel, data);
             });
         }
+        // Ensure we don't have any unhandled exceptions (swallow exceptions as we're not awaiting on this promise).
+        this.sendChain.catch(noop);
```
DonJayamanne, line 282: Swallow more unhandled exceptions for dangling promises

src/kernels/variables/pythonVariableRequester.ts
```diff
@@ -36,6 +37,24 @@ export function parseDataFrame(df: DataFrameSplitFormat) {
     return { data };
 }
 
+export async function safeExecuteSilently(
+    kernel: IKernel,
+    code: string,
+    errorOptions?: SilentExecutionErrorOptions
+): Promise<nbformat.IOutput[]> {
+    if (kernel.disposed || kernel.disposing || !kernel.session || !kernel.session.kernel || kernel.session.disposed) {
```
DonJayamanne, line 45: If the kernel has died/dying, then no point attempting to send requests for variables.

src/test/datascience/notebook/helper.ts
```diff
@@ -166,7 +165,9 @@ async function createTemporaryNotebookFromNotebook(
     await workspace.fs.writeFile(uri, Buffer.from(JSON.stringify(notebook)));
 
     disposables.push({
-        dispose: () => swallowExceptions(() => workspace.fs.delete(uri))
+        dispose: () => {
+            void workspace.fs.delete(uri).then(noop, noop);
```
DonJayamanne, line 169: More dangling promises

gulpfile.js
```diff
@@ -364,3 +364,12 @@ gulp.task('validateTelemetryMD', async () => {
         throw new Error('Telemetry MD is not valid, please re-run `npm run generateTelemetry`');
     }
 });
+
+gulp.task('verifyUnhandledErrors', async () => {
+    const fileName = path.join(__dirname, 'unhandledErrors.txt');
```
DonJayamanne, line 369: Optionally, this will ensure we no longer have unhandled exceptions in our new code.

src/interactive-window/commands/commandRegistry.ts
```diff
@@ -120,7 +120,7 @@ export class CommandRegistry implements IDisposable, IExtensionSingleActivationS
         this.registerCommand(Commands.RunCurrentCell, this.runCurrentCell);
         this.registerCommand(Commands.RunCurrentCellAdvance, this.runCurrentCellAndAdvance);
         this.registerCommand(Commands.ExecSelectionInInteractiveWindow, (textOrUri: string | undefined | Uri) => {
-            void this.runSelectionOrLine(textOrUri);
+            void this.runSelectionOrLine(textOrUri).catch(noop);
```
DonJayamanne, line 123: more dangling promises

src/kernels/kernel.base.ts
```diff
@@ -330,9 +330,11 @@ export abstract class BaseKernel implements IKernel {
...
             this.perceivedJupyterStartupTelemetryCaptured = true;
             sendTelemetryEvent(Telemetry.PerceivedJupyterStartupNotebook, stopWatch.elapsedTime);
-            executionPromise.finally(() =>
-                sendTelemetryEvent(Telemetry.StartExecuteNotebookCellPerceivedCold, stopWatch.elapsedTime)
-            );
+            executionPromise
+                .finally(() =>
+                    sendTelemetryEvent(Telemetry.StartExecuteNotebookCellPerceivedCold, stopWatch.elapsedTime)
+                )
+                .catch(noop);
```
rchiodo, line 337: Why not use 'ignoreErrors'?

rchiodo (approved):

src/kernels/kernel.base.ts
DonJayamanne: For some reason I never liked that.

DonJayamanne: It just feels a little dirty to wrap something so simple and native. Though i agree it saves an import.
& the downside of the extension to native promises is if someone else creates a similar extension that does something different. These aren't show stoppers, but enough to make me go `hmm, eew`

amunger:
I feel like this could end up hiding a lot of issues, and the default should be to log the error, rather than just swallow it.

DonJayamanne:
>  feel like this could end up hiding a lot of issues, and the default should be to log the error, rather than just swallow it.

We were already ignoring these errors deliberately as we diddn't care whether they failed or not.
In other places we do have `catch(ex => traceError`, however there are plenty of cases when this isn't required at all. 
I.e. we just want to fire something & forget about it.

Issue #10202, opened 2022-05-31
Prefer an output channel over console.log 


Tyriar:
Testing #10170

I see a bunch of console.logs when opening a Python notebook, these should go to the Jupyter output channel instead to try keep the main devtools console clean.

<img width="680" alt="Screen Shot 2022-05-31 at 3 49 59 pm" src="https://user-images.githubusercontent.com/2193314/171296679-36dfa5d0-ad2f-45ae-b048-8174295ee925.png">

rchiodo:
Those are actually from an npm module we consume. I suppose we could have it send a message instead.

rchiodo:
Or maybe just remove those and only log errors.

Tyriar:
That makes things trickier if you don't own the code, this is a pretty minor issue that mainly helps vscode devs/extension authors, feel free to close off if it would involve upstream changes that probably won't happen.

rchiodo:
Oh we actually own the npm module. It's this one: 
https://github.com/microsoft/vscode-jupyter-ipywidgets.git

Tyriar:
@rchiodo is that for all 3? There's the SW ones, ipywidgets and the requirejs one (in a file that has `console.warn("Registering custom require.js for Jupyter Kernel")`)

rchiodo:
SW wouldn't be us AFAIK. 

The require.js one is likely another renderer side script we generate.

PR #10341, opened 2022-06-06
Remove console log usage

rchiodo:
Fixes https://github.com/microsoft/vscode-jupyter/issues/10202

Related to the PR I have in the ipywidget npm module, this removes uses of console.log in our renderers. Mostly it was widget stuff.
https://github.com/microsoft/vscode-jupyter-ipywidgets/pull/3

### DIFF
--- a/src/kernels/ipywidgets-message-coordination/ipyWidgetMessageDispatcher.ts
+++ b/src/kernels/ipywidgets-message-coordination/ipyWidgetMessageDispatcher.ts
@@ -7,7 +7,7 @@ import type { KernelMessage } from '@jupyterlab/services';
 import * as uuid from 'uuid/v4';
 import { Event, EventEmitter, NotebookDocument } from 'vscode';
 import type { Data as WebSocketData } from 'ws';
-import { traceInfoIfCI, traceVerbose, traceError } from '../../platform/logging';
+import { traceVerbose, traceError, traceInfo } from '../../platform/logging';
 import { IDisposable } from '../../platform/common/types';
 import { Deferred, createDeferred } from '../../platform/common/utils/async';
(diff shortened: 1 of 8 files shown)


src/kernels/ipywidgets-message-coordination/ipyWidgetMessageDispatcher.ts
```diff
@@ -106,7 +106,7 @@ export class IPyWidgetMessageDispatcher implements IIPyWidgetMessageDispatcher {
     public receiveMessage(message: IPyWidgetMessage): void {
         switch (message.message) {
             case IPyWidgetMessages.IPyWidgets_logMessage:
-                traceInfoIfCI(`Widget Message: ${message.payload}`);
+                traceInfo(`Widget Message: ${message.payload}`);
```
rchiodo, line 109: Made this traceInfo because after these two changes, this is really only stuff that we actually want to show to the user (or us when request logs).

src/webviews/webview-side/ipywidgets/common/kernel.ts
```diff
@@ -393,9 +390,6 @@ class ProxyKernel implements IMessageHandler, Kernel.IKernelConnection {
     }
     private messageHookInterceptor(msg: KernelMessage.IIOPubMessage): boolean | PromiseLike<boolean> {
         try {
-            logMessageOnlyOnCI(
```
rchiodo, line 396: Figured if we needed these later, we can add a new `IPYWidgets_LogMessageIfCI` message. Then on the other side  it would only log these if running on CI.

amunger (approved):

DonJayamanne (approved):

PR #10346, opened 2022-06-06
Check that tests were run

amunger:
Run a custom test output parser as a gulp task.
Checks for the test-result.xml file that test runs should be outputting.
Displays info about the test run and details about any test failures.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -59,70 +59,85 @@ gulp.task('validateTranslationFiles', (done) => {
 
 gulp.task('checkTestResults', async (done) => {
     const core = require('@actions/core');
+    var glob = require("glob")
 
-    const data = await fs.promises.readFile('test-results.xml');
-    const parser = require('xml-js');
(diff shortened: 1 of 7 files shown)


amunger:
Need to look at why the failed checks aren't writing test-results files, and the empty nopython suite is a separate issue

gulpfile.js
```diff
@@ -57,6 +57,74 @@ gulp.task('validateTranslationFiles', (done) => {
...
+    const data = await fs.promises.readFile('test-results.xml');
+    const parser = require('xml-js');
+    const report = JSON.parse(parser.xml2json(data, { compact: true }));
+
+    let testsRun = 0;
+    let failCount = 0;
+    let skipCount = 0;
+    try {
+        if (report.testsuites) {
+            core.info(JSON.stringify(report.testsuites._attributes));
```
rchiodo, line 72: It looked like this might not be there?

rchiodo (approved):

PR #10355, opened 2022-06-07
More tree shaking

### LINKED ISSUE #10307: Reduce web bundle size

rebornix:
Fixes #10307 

### DIFF
--- a/build/webpack/webpack.extension.web.config.js
+++ b/build/webpack/webpack.extension.web.config.js
@@ -87,7 +87,7 @@ const config = {
             }
         ]
     },
-    externals: ['vscode', 'commonjs', 'electron'], // Don't bundle these
+    externals: ['vscode', 'commonjs', 'electron', 'moment'], // Don't bundle these
     plugins: [
         // Work around for Buffer is undefined:
         new webpack.ProvidePlugin({
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -265,7 +265,7 @@ module.exports = {
         ],
         'react/jsx-uses-vars': 'error',
         'react/jsx-uses-react': 'error',
-        'no-restricted-imports': ['error', { paths: ['lodash'] }],
+        'no-restricted-imports': ['error', { paths: ['lodash', 'rxjs'] }],
         'import/no-restricted-paths': [
             'error',
             {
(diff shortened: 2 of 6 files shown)


build/webpack/webpack.extension.web.config.js
```diff
@@ -87,7 +87,7 @@ const config = {
             }
         ]
     },
-    externals: ['vscode', 'commonjs', 'electron'], // Don't bundle these
+    externals: ['vscode', 'commonjs', 'electron', 'moment'], // Don't bundle these
```
DonJayamanne, line 90: Is it possible to do something to prevent we don't end up in a situtation where tomorrow we pull in a 3rd party package or an existing 3rd party package ends up pulling in moment, then things will not work, or it will fail only in certian cases based on where and how the 3rd party package ends up using moment.

E.g. can we write up a linter to look at package-lock.json and ensure that nothing ever pulls in moment except for jupyter. Else I feel this is a risky change (as mentioned if some other package pulls it in, then things could break sometimes)

DonJayamanne (changes requested):
Would like to ensure no other package pulls in moment, if they do, then we'd need to review that separately. Else we end up in a situation where something could break in the future.

rebornix:
Thanks @DonJayamanne for the suggestion, we should validate the dependencies to see if any new package we use load `moment` and fail accidentally, great catch! 

Added a gulp task to validate the lock file and added it to the pre-commit hook.

.husky/pre-commit
```diff
@@ -15,4 +15,5 @@ if [ -z "$changed" ]; then
 fi
 
 npm run generateTelemetry
+npm run validateDependencies
```
rchiodo, line 18: Can we maybe do this in the linter step for the build? I'm wary of adding yet more stuff to pre-commit. Pre-commit already takes like 60 seconds for me.

rebornix: @rchiodo 👍  I dislike slow commit, 60 seconds is too much. I'll move it to the linter since we always do PR.

rchiodo (approved):

PR #10374, opened 2022-06-08
made vscode-jupyter to latest LSP

heejaechang:
vscode-jupyter was using old LSP. this will move them to the lastest.


src/intellisense/languageServer.node.ts
```diff
@@ -79,6 +91,8 @@ class NerfedExecuteCommandFeature implements DynamicFeature<ExecuteCommandRegist
     }
 }
 
+let languageClient: LanguageClient | undefined;
```
rchiodo, line 94: This can't be a global variable. The LanguageServer is not a singleton.

rchiodo (changes requested):
I think the language client still needs to be passed into the 'languageServer' class. There's one per notebook interpreter.

src/intellisense/languageServer.node.ts
heejaechang: ah.. okay.

src/intellisense/languageServer.node.ts
```diff
@@ -166,33 +199,24 @@ export class LanguageServer implements Disposable {
...
+            await client.start();
 
-            // After starting, wait for it to be ready
-            while (languageClient && !languageClient.initializeResult) {
-                await sleep(100);
-            }
-            if (languageClient) {
-                await languageClient.onReady();
-            }
+            // Expose client once it is fully initialized.
```
rchiodo, line 216: I don't think you need this line anymore. Not used.

heejaechang: I think languageClient is still exposed through middleware?

rchiodo: No this variable isn't used. It's just assigned. You use 'client' instead everywhere, so you don't really need this variable.

rchiodo:
Looks like all of our intellisense tests are failing, which is likely caused by this change. 

https://github.com/microsoft/vscode-jupyter/runs/6801323085?check_suite_focus=true

Did you try running a notebook to see if you can get completions still?

heejaechang:
oops. clicked wrong button ..

src/intellisense/languageServer.node.ts
rchiodo: Oh didn't notice that as it wasn't in the changes. Yeah looks like it's needed for the lambda capture.

rchiodo:
Looks like the completion tests are still failing. Do you want any help debugging them?

rchiodo (approved):

PR #10380, opened 2022-06-08
Rest of 'notebooks' removal from 'kernels'

rchiodo:
For #10152 

This is the last of the 'notebooks' references in the 'kernels' folder. 

Adding lint rule to reenforce this.


src/interactive-window/interactiveWindowCommandListener.ts
```diff
@@ -241,7 +242,7 @@ export class InteractiveWindowCommandListener implements IDataScienceCommandList
                             openQuestion1
                         );
                         if (selection === openQuestion1) {
-                            await this.ipynbProvider.open(uri);
+                            await openAndShowNotebook(uri);
```
DonJayamanne, line 245: Should we remove the `open` method from the `InotebookEditorProvider` interface?

DonJayamanne (approved):

PR #10426, opened 2022-06-10
Remove unused code

### LINKED ISSUE #10418: Remove unused code
@jrieken wrote a great [analyzer](https://github.com/jrieken/ts-unused) to scan the code base and find unused code. This can be used in Jupyter to identify code which are no longer used in both desktop and web bundles.

rebornix:
Fixes #10418

The PR removes symbols which are not referenced at all (either on desktop or web).


IanMatthewHuff:
I'm pro most of the removals, but did wonder a bit about the vscode-path removals. Wasn't that a direct directory copy of existing code that @rchiodo added for web? Seems a bit odd to have only a partial implementation of the API.

rebornix:
@IanMatthewHuff you are correct, the `vscode-path` was borrowed for enabling `path` on Web. The catch here is we want a web enabled `path` but that depends on `strings` and a few other base modules from VS Code, though only a few functions are used, we copied them all over. The modules are pretty stable in VS Code (normally no changes to existing functions anymore, only additions), and `strings` contains tons of logic to handle unicode segmentation and emojis, which are not relevant to us.

If we could export `path` module from vscode, it might help so we don't have to manually tree shake the modules. But I'm also not too concerned about removing dead code from those modules, as we are very unlikely to modify them.

IanMatthewHuff:
@rebornix Thanks for explanation, I was worried about having to update it all the time doing "partial" updates of only some functions. Sounds good to remove then.

IanMatthewHuff (approved):

rchiodo:
I'm going to merge this because it's affecting one of the test failures I'm working on. I think we're not parsing semvers correctly and that's messing up conda detection.

PR #10437, opened 2022-06-13
Make telemetry use platform only.

rebornix:
Re #10152 

Attempt to enforce using `platform` modules only in `telemetry`. The PR includes two changes:

* Move kernel and notebook specific telemetry registration into `notebooks/kernel`
* Separate `IEventNamePropertyMapping` with the rest of basic helpers in `telemetry`. Make sure we only import `platform` in `telemetry`, except `IEventNamePropertyMapping

---

The reason `IEventNamePropertyMapping` is an exception is this mapping is used to track all our telemetry points and they also serve type checking

```ts
sendTelemetryEvent(eventName, eventValue)
```

We will check `eventValue` against `IEventNamePropertyMapping[eventName)`. This means current approach will have to import types from all other modules. To move off this we would need to a different approach of type checking, for example 

```
interface IDebuggingTelemetry {
  'ipykernel6Status': {
        status: 'installed' | 'notInstalled';
    }
}

sendTelemetryEvent<IDebuggingTelemetry>('ipykernel6Status', { status })
```

which won't require us to register all the typings in `IEventNamePropertyMapping`. Considering this is a rather big change and would also require changes for our tooling, I'll keep layer breaking imports in this file only.


amunger:
Did telemtry.md get removed completely? I think Soojin still finds that file useful

TELEMETRY.md
```diff
@@ -2,9498 +2,3 @@
 
 Expand each section to see more information about that event.
 
-<details>
```
rchiodo, line 5: Yeah looks like telemetry generation is broken. You likely need to change the telemetry generator too.

src/telemetry/types.ts
```diff
@@ -95,3 +112,1229 @@ export interface IInterpreterPackages {
     getPackageVersion(interpreter: PythonEnvironment, packageName: string): Promise<string | undefined>;
     trackPackages(interpreterUri: InterpreterUri, ignoreCache?: boolean): void;
 }
+
+export interface IEventNamePropertyMapping {
```
rchiodo, line 116: ah the telemetry generator has to look here instead of index.ts for the telemetry information. That's why telemetry.md is empty.

rchiodo: See this file here:
https://github.com/microsoft/vscode-jupyter/blob/0a9f6009cc4aca40d233ffa8f03a5c943e93e810/src/platform/tools/telemetryGenerator.node.ts#L336

TELEMETRY.md
rebornix: It seems the generator fails to work after moving files around, fix it now.

src/telemetry/types.ts
rebornix: 🍺 thanks!

rebornix:
@amunger @rchiodo thanks! The telemetry.md should now be fixed.

amunger (approved):

src/notebooks/telemetry/interpreterPackageTracker.ts
```diff
@@ -3,16 +3,16 @@
...
-import { IInterpreterPackages } from './types';
+import { IExtensionSingleActivationService } from '../../platform/activation/types';
+import { IPythonExtensionChecker, IPythonApiProvider } from '../../platform/api/types';
+import { IExtensions, IDisposableRegistry, InterpreterUri } from '../../platform/common/types';
+import { isResource, noop } from '../../platform/common/utils/misc';
+import { IInterpreterService } from '../../platform/interpreter/contracts';
+import { INotebookControllerManager } from '../types';
+import { IInstaller, Product } from '../../kernels/installer/types';
+import { IVSCodeNotebookController } from '../controllers/types';
+import { trackKernelResourceInformation } from '../../kernels/telemetry/helper';
```
DonJayamanne, line 14: A side effect is, everything that didn't depend on kernels, now pulls in dependency from kernels folder.

DonJayamanne: I guess that's what we're trying to fix here, and this PR is just one of many.

DonJayamanne (approved):

src/notebooks/telemetry/interpreterPackageTracker.ts
rebornix: @DonJayamanne you are right. This particular module `InterpreterPackageTracker` needs to read `INotebookControllerManager`, `./src/kernels`, and `./src/telemetry`. The best place for it is probably out of`./src/notebooks` and placed in a folder where we have all the moduels that glue things together.

PR #10441, opened 2022-06-13
Workaround for vscode-nls for locally generated .VSIXs

sadasant:
Since using `vscode-nls` and `vscode-nls-dev`, when packaging the `vscode-jupyter` extension locally, localization is unavailable and the extension becomes unusable, replacing any translated text with `Failed to load message bundle` messages. This is because `vscode-nls-dev` does not currently support packaging extensions outside of the Azure pipelines. This is related to issue: https://github.com/microsoft/vscode-nls-dev/issues/40 -- this PR is to disable `vscode-nls-dev` from being used on environments without a specific environment variable of our choice, and to allow replacing `vscode-nls` with our previous localization approach in Node.js if localization isn't working.

Feedback appreciated.

### DIFF
--- a/.github/actions/build-vsix/action.yml
+++ b/.github/actions/build-vsix/action.yml
@@ -5,7 +5,7 @@ inputs:
   package_command:
      description: 'Package.json package script'
      required: false
-     default: 'package:nls'
+     default: 'package'
 outputs:
   path:
     description: 'Path to the VSIX'
--- a/build/azure-pipeline.pre-release.yml
+++ b/build/azure-pipeline.pre-release.yml
@@ -43,5 +43,5 @@ extends:
       - script: npm run updateBuildNumber
         displayName: Update build number
 
-      - script: gulp prePublishBundle
+      - script: npm run prePublishBundleNLS
         displayName: Build
--- a/build/azure-pipeline.stable.yml
+++ b/build/azure-pipeline.stable.yml
@@ -39,5 +39,5 @@ extends:
       - script: npm run updateBuildNumber
         displayName: Update build number
 
-      - script: gulp prePublishBundle
+      - script: npm run prePublishBundleNLS
         displayName: Build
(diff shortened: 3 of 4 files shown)


build/webpack/webpack.extension.node.config.js
```diff
@@ -38,12 +38,16 @@ const config = {
             {
                 test: /\.ts$/,
                 use: [
-                    {
-                        loader: 'vscode-nls-dev/lib/webpack-loader',
-                        options: {
-                            base: constants.ExtensionRootDir
-                        }
-                    },
+                    ...(process.env.BUILD_WITH_VSCODE_NLS
```
sadasant, line 41: Is this name ok? What other name would be better? How can we provide this to our CI pipeline?

rchiodo: Well there may be environment variables already active in the azure-pipeline stuff. But if not you can add one into our versions.

sadasant: For now I'm thinking on setting this environment variable in a `package:nls` npm script, and using that instead of `package` as the default package command.

rchiodo (approved):

.github/actions/build-vsix/action.yml
```diff
@@ -5,7 +5,7 @@ inputs:
   package_command:
      description: 'Package.json package script'
      required: false
-     default: 'package'
+     default: 'package:nls'
```
rchiodo, line 8: Sorry this is the wrong version AFAIK. You should be modifying the azure-pipelines.prerelease and azure-pipelines.stable.yml

rchiodo (changes requested):
I think you modified the wrong file.

.github/actions/build-vsix/action.yml
rchiodo: This action.yml is used by our build-test.yml

build/webpack/webpack.extension.node.config.js
sadasant: For now I have changed how our build `.yml` files work. Instead of `gulp prePublishBundle` they will call `npm run prePublishBundleNLS`, which injects the environment variable with `cross-env`. I think it's a very versatile approach with minimal maintenance!

PR #10458, opened 2022-06-14
Remaining kernel webview layer protection

rebornix:
Re #10152 

It includes the remaining work for removing webview dependencies from kernels:

* Move Jupyter Export/Import into `webview/extension-side` for now, as this component depends on `interactive-window` and `notebooks`
* Move Kernel server preload into `webview/extension-side` for now as it depends on `interactive-window`, `notebooks` and `kernel`.
  * It feels like the kernel server preload should be part of `kernels` but currently it has assumptions of how a kernel is loaded (in a notebook or iw). We should take a look later if we should move off `notebook/iw` for this feature.


src/webviews/extension-side/serviceRegistry.node.ts
```diff
@@ -27,4 +32,6 @@ export function registerTypes(serviceManager: IServiceManager, _isDevMode: boole
         IJupyterVariableDataProviderFactory,
         JupyterVariableDataProviderFactory
     );
+    serviceManager.add<INotebookExporter>(INotebookExporter, JupyterExporter);
```
rchiodo, line 35: Would it make more sense to put this into the interactive src tree? It has nothing to do with webviews but at least it depends upon the IW code.

rchiodo (approved):

rebornix:
> Would it make more sense to put this into the interactive src tree? It has nothing to do with webviews but at least it depends upon the IW code.

I'll put this in my todo list for the second round of refactoring.

PR #10468, opened 2022-06-16
Output metadata Id on new output

### LINKED ISSUE #9629: Migrate off of the vscode.proposed.notebookDeprecated API

sadasant:
This PR migrates off of the `output.id` used to view and save plots to the new `output.metadata.id`.

For this PR to work, the following two PRs would need to be merged:

- https://github.com/microsoft/vscode-notebook-renderers/pull/114
- [coming soon]

Fixes #9629

### DIFF
--- a/src/kernels/execution/helpers.ts
+++ b/src/kernels/execution/helpers.ts
@@ -35,6 +35,7 @@ import {
     kernelConnectionMetadataHasKernelModel,
     getKernelRegistrationInfo
 } from '../helpers';
+import * as uuid from 'uuid/v4';
 
 export enum CellOutputMimeTypes {
     error = 'application/vnd.code.notebook.error',
@@ -219,6 +220,7 @@ export function cellOutputToVSCCellOutput(output: nbformat.IOutput): NotebookCel
 function getOutputMetadata(output: nbformat.IOutput): CellOutputMetadata {
     // Add on transient data if we have any. This should be removed by our save functions elsewhere.
     const metadata: CellOutputMetadata = {
+        id: uuid(),
         outputType: output.output_type
     };
     if (output.transient) {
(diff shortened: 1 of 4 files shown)


src/kernels/execution/helpers.ts
```diff
@@ -353,6 +355,10 @@ export type CellOutputMetadata = {
      * Whether to display the open plot icon.
      */
     __displayOpenPlotIcon?: boolean;
+    /**
+     * A unique identifier to track a specific output.
+     */
+    id?: string;
```
rebornix, line 361: Maybe we could call this `__id` which would be similar to `__isJson` and `__displayOpenPlotIcon` so even if in the future the core introduces `CellOutputMetadata.id` we won't have a conflict.

src/webviews/extension-side/plotView/helpers.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { NotebookCellOutput } from 'vscode';
+import { CellOutputMetadata } from '../../../kernels/execution/helpers';
+
+export function getCellOutputMetadata(cell?: NotebookCellOutput): CellOutputMetadata | undefined {
```
rebornix, line 7: We probably don't need a helper function for this, `cell?.metadata as CellOutputMetadata | undefined`

rebornix (approved):
Looks good in overall, a few nitpicks

PR #10471, opened 2022-06-16
Fix error links in web to work

### LINKED ISSUE #10287: Error links in notebooks cannot be clicked in web extension
Open in insiders.vscode.dev
Create an exception in a cell
Try clicking on a link

Nothing happens.

rchiodo:
Fixes #10287

Root cause - comm message handler wasn't registered in web. 

Also fixed some issues with it jumping to the wrong cell if other cells had the same execution count.

Also added a test.


src/notebooks/outputs/errorRendererComms.ts
```diff
@@ -105,27 +118,27 @@ export class ErrorRendererCommunicationHandler implements IExtensionSyncActivati
         const uri = Uri.parse(cellUri);
 
         // Show the matching notebook if there is one
-        let editor = this.notebooks.notebookEditors.find((n) => arePathsSame(n.notebook.uri.fsPath, uri.fsPath));
```
rchiodo, line 108: Paths here don't work on web, so switched to just finding the cell instead.

src/notebooks/outputs/tracebackFormatter.ts
```diff
@@ -48,9 +48,6 @@ export class NotebookTracebackFormatter implements ITracebackFormatter {
 
         const inputMatch = /^Input.*?\[.*32mIn\s+\[(\d+).*?0;36m(.*?)\n.*/.exec(traceFrame);
         if (inputMatch && inputMatch.length > 1) {
-            const executionCount = parseInt(inputMatch[1]);
-            // Don't assume the current cell is where the error is.
```
rchiodo, line 52: This wasn't working if the execution count is the same in a previous cell. Not sure how in a notebook formatter the error would come from a different cell? Removed this as I don't think we actually need it.

src/notebooks/outputs/errorRendererComms.ts
```diff
@@ -29,33 +39,36 @@ export class ErrorRendererCommunicationHandler implements IExtensionSyncActivati
...
-                            await commands.executeCommand(command, params);
-                        }
-                    } else {
-                        this.applicationShell.openUrl(href);
-                    }
+        messageChannel.onDidReceiveMessage(this.onDidReceiveMessage, this, this.disposables);
+    }
+
+    // Public for testing
+    public async onDidReceiveMessage(e: {
```
IanMatthewHuff, line 46: You mentioned for testing, but honestly I like this better just as a general thing. Lambdas always look funny to me when they get really long.

IanMatthewHuff (approved):

PR #10477, opened 2022-06-16
add dispose to mark the client is done and should be never reused.

heejaechang:
mark LanguageClient diposed so that it can never auto restart itself.


heejaechang:
@rchiodo can you take a look at?

debonte (approved):

rchiodo (approved):

PR #10486, opened 2022-06-17
Service registration refactor part 3

rebornix:
Re #10152 

Moving more service registrations which depend on notebook/kernels out of platform.


src/webviews/extension-side/activationManager.ts
```diff
@@ -8,7 +8,7 @@ import {
     IExtensionActivationManager,
```
rebornix, line 8: `ActivationManager` is now moved into `extension-side/activationManager`.

src/webviews/extension-side/serviceRegistry.node.ts
```diff
@@ -105,4 +116,9 @@ export function registerTypes(serviceManager: IServiceManager, _isDevMode: boole
     );
 
     serviceManager.addSingletonInstance<IExtensionSideRenderer>(IExtensionSideRenderer, new ExtensionSideRenderer());
+
+    serviceManager.addSingleton<ISurveyBanner>(ISurveyBanner, DataScienceSurveyBanner);
+    serviceManager.addBinding(ISurveyBanner, IExtensionSingleActivationService);
+    // Activation Manager
+    serviceManager.add<IExtensionActivationManager>(IExtensionActivationManager, ExtensionActivationManager);
```
rebornix, line 123: WorkspaceActivation, Banners, and ActivationManager registered through `extension-side`.

src/webviews/extension-side/survey/dataScienceSurveyBanner.node.ts
```diff
@@ -5,10 +5,14 @@
 
 import { inject, injectable } from 'inversify';
 import { NotebookCellExecutionState, NotebookCellExecutionStateChangeEvent, UIKind } from 'vscode';
```
rebornix, line 7: Survey banner is a standalone feature and doesn't belong to `platform`

src/telemetry/types.ts
```diff
@@ -1262,34 +1262,7 @@ export interface IEventNamePropertyMapping {
...
-     */
-    [Telemetry.TerminalShellIdentification]: {
-        failed: boolean;
-        reason: 'unknownShell' | undefined;
-        terminalProvided: boolean;
-        shellIdentificationSource: 'terminalName' | 'settings' | 'environment' | 'default' | 'vscode';
-        hasCustomShell: undefined | boolean;
-        hasShellInEnv: undefined | boolean;
-    };
+    [Telemetry.TerminalShellIdentification]: ShellIdentificationTelemetry;
```
DonJayamanne, line 1265: This reduces the readability of the telemetry now.
E.g. in the past we'd have to go into this one file and everyting was documented in here.
If one looked at `TerminalShellIdentification` then we can see all of the properties & the documentation here.
With this new change we'll end up jumping around a lot more & that makes it difficult.

TELEMETRY.md
```diff
@@ -9557,40 +9557,13 @@ No properties for event
...
-                             If environment, then this means we identified the type of the shell based on their environment (env variables, etc).
-                                 I.e. their default OS Shell.
-                             If default, then we reverted to OS defaults (cmd on windows, and bash on the rest).
-                                 This is the worst case scenario.
-                                 I.e. we could not identify the shell at all.
- hasCustomShell - If undefined (not set), we didn't check.
-                  If true, user has customzied their shell in VSC Settings.
- hasShellInEnv - If undefined (not set), we didn't check.
-                 If true, user has a shell in their environment.
-                 If false, user does not have a shell in their environment.
```
DonJayamanne, line 9578: As you can see now all of the documentaiton is lost from the MD file as well.
I don' think this is desirable .

DonJayamanne (changes requested):
The documetnation is lost in telemetry.md and looking the telemetry definition now requires jumping across files.
I thought a sugegstion we agreed on was:
* Move all of the types into a top level file named `telemetryIntegration.ts`
* Add a function sendTelemetry or the like inot this and ensure every place calls this to send telemetry
* & this file will in trun calls into telemtry/index.ts & nothing else can reference the telemtry/index.ts & other files except this top level integration file.

rebornix:
> I thought a sugegstion we agreed on was:

> * Move all of the types into a top level file named telemetryIntegration.ts
> * Add a function sendTelemetry or the like inot this and ensure every place calls this to send telemetry
& this file will in trun calls into telemtry/index.ts & nothing else can reference the telemtry/index.ts & other files except this top level integration file.

Yes this is still what we want to do to ensure that we have a single file which have types mapping for telemetry key and value. Didn't try to do this yet as it's not in the scope of this PR.

> This reduces the readability of the telemetry now.
> E.g. in the past we'd have to go into this one file and everyting was documented in here.
If one looked at TerminalShellIdentification then we can see all of the properties & the documentation here.
With this new change we'll end up jumping around a lot more & that makes it difficult.

This is where it falls apart currently. We currently import types from other components (https://github.com/microsoft/vscode-jupyter/blob/main/src/telemetry/types.ts#L8-L24) into this file instead of having them all inlined. The type introduced in this PR is at least the same as others: one Go To Definition and you will see the full definition of this type and documention.

With that said, I agree that having a self-explanatory telemetry typings file is what we should try to archive. Open questions here are:
* If we want to use the typings elsewhere, do we duplicate the typings or we do `type x = IEventNamePropertyMapping[EventName.x]`
* When we use types from other modules, do we also duplicate the jsdoc for the type in `telemetry/types.ts`

We can discuss this in standup, for now I'll revert the change to the `telemetry/types.ts`.

rebornix:
> Move all of the types into a top level file named telemetryIntegration.ts
> Add a function sendTelemetry or the like inot this and ensure every place calls this to send telemetry
& this file will in trun calls into telemtry/index.ts & nothing else can reference the telemtry/index.ts & other files except this top level integration file.

Some more thoughts on this strategy here, currently the `src/telemetry` is now in following structure:

* `telemetry/index.ts` which implements `sendTelemetry`
* `telemetry/types.ts` which declares all typings for telemetry events
* helper functions for sending specific telemetry events, including not limited to
  * `telemetry/startupTelemetry`, which gather startup stats, workspace stats and others, then call `sendTelemetry`
  * `telemetry/envFileTelemetry`, it implements `sendActivationTelemetry`, which calls `sendTelemetry` at the end
  * ...
 
`telemetry/index.ts` here is not the entry file or wrapper file for the `telemetry` component, and it doesn't import any helper functions from other files in `src/telemetry` folder (except `types.ts`). When we move all types and `sendTelemetry` to a top file `telemetryIntegration.ts`, we would find that `telemetryIntegration.ts` doesn't need to import from `src/telemetry` as the module now only contains helper functions which wrap `sendTelemetry`.

Maybe what we could do is having the top `telemetryIntegration.ts` file, which contains `sendTelemetry` and all `typings`. This single file can be imported by anyone, and we store all the helper functions in either `platform` or feature components (i.e. store `sendKernelTelemetry` in `kernels`).

DonJayamanne:
> Maybe what we could do is having the top telemetryIntegration.ts file,

agreed, i think that's what we should do.

amunger (approved):

rchiodo (approved):

DonJayamanne (approved):

PR #10491, opened 2022-06-17
Fix error with rendering widgets offscreen

### LINKED ISSUE #10485: Scrolling cells off screen prevents rendering widgets
Have a notebook with enough cells that cells scroll off screen.

Render something like bqplot graphs off screen.

Scroll up.

Expected:
- Plots show up
Actual:
- There's no output.

rchiodo:
Fixes #10485 

Disposing of widgets prevents them from rerendering when they activate later.


src/platform/tools/telemetryGenerator.node.ts
```diff
@@ -280,7 +280,7 @@ function writeTelemetryEntry(entry: TelemetryEntry) {
             const link = `https://github.com/microsoft/vscode-jupyter/tree/main/${l.file}`;
             writeOutput(`\n[${l.file}](${link})`);
             writeOutput('```typescript');
-            writeOutput(l.code);
+            writeOutput(l.code.replace('\\r\\n', '\n'));
```
rchiodo, line 283: I believe the differences in the telemetry lately has been because of some us being on Windows and others not.

rebornix: Good catch Rich. Neither github.com nor github.dev (monaco diff editor) can render the line break changes ;(

rebornix (approved):

src/platform/tools/telemetryGenerator.node.ts
IanMatthewHuff: Makes sense. I was looking at the markdown and wondering what I was missing...

PR #10494, opened 2022-06-17
Separate python/intepreter api and extension api

rebornix:
Re #10152 

There are two types of APIs

* Python/Intepreter APIs which we build on top of Python extension
* API we expose to other extensions, defined in `extension.d.ts`

This PR attempts to separate these two. The first group API sits in `platform` and can be used by other components, while the extension api is moved out of `platform` and should not be used by any internal/primitive component.


rchiodo (approved):
I'm wondering what platform is anymore? You're moving 'global' things to webviews/extension-side, but isn't it the same as being platform? What does platform represent if not global things?

rebornix:
> I'm wondering what platform is anymore? You're moving 'global' things to webviews/extension-side, but isn't it the same as being platform? What does platform represent if not global things?

Discussed with @rchiodo offline about the wording and definitions, especially around what "global" means in the context.

There are two types of "global" components. The first group includes utilities(`logging`), helper functions (`vscode-path`), or infra code (`ioc`). The second group includes **global standalone classes/components** (e.g., extension api, data viewer), they are not used by other components (i.e., `notebooks`, `kernels`), but are exposed to `extension.node/web.ts`. 

During my refactoring, I tried to use `src/platform` for components in the first group, and use `src/webviews/extension-side` for the second group. Neither of the folder names are intuitive so we can discuss how we are going to call them. One idea I have now is:

* `src/base`, for infra and utilities code.
* `src/platform`, for global standalone classes. 

@microsoft/datascience-vsc-team-reviewers thoughts?

PR #10501, opened 2022-06-18
Change Install Python Extension and Install Python from dummy kernel to kernelSelector commands

### LINKED ISSUE #10513: Change 'Python 3' dummy install kernel into kernel selector commands
Replace the 'Python 3' dummy kernel that we have to help install the Python Extension or Python and replace it with commands that show up in the kernel selection space to perform the same general actions. Also change the install flow to show only a single modal dialog during the Python extension install flow.

IanMatthewHuff:
Fixes #10513


IanMatthewHuff:
Gotta merge with refactoring changes and will look at adding a test or two. Not quite sure yet how to best test this, it's not really unit testable, and it might be hard to engineer the situations on CI. Might look at stubbing out listKernels in the controller somehow to fake not finding python kernels.

src/platform/api/pythonApi.ts
```diff
@@ -199,14 +205,19 @@ export class PythonExtensionChecker implements IPythonExtensionChecker {
...
+    }
+
+    // Notify the user that Python is require, and open up the Extension installation page to the
+    // python extension
     public async showPythonExtensionInstallRequiredPrompt(): Promise<void> {
         // If workspace is not trusted, then don't show prompt
         if (!this.workspace.isTrusted) {
             return;
         }
-        if (this.waitingOnInstallPrompt) {
```
IanMatthewHuff, line 207: Old unused code.

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -378,6 +421,84 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
...
+        } else {
+            sendTelemetryEvent(Telemetry.PythonNotInstalled, undefined, { action: 'dismissed' });
+        }
+    }
+
+    // Called when we select the command to install the python extension via the kernel picker
+    // If new controllers are added before this fully resolves any in progress executions will be
+    // passed on, so we can trigger with the run button, install, get a controller and not have to
+    // click run again
+    private async installPythonExtensionViaKernelPicker(): Promise<void> {
```
DonJayamanne, line 450: The notebookControllerManager is very complex right now.
I think this can go into a seperate class, all we're doing here is just handling the command, 
I think we can move the handler along with the command registration into a separate class.

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -378,6 +421,84 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         }
     }
 
+    // This is called via the "install python" command in the kernel picker in the case where
+    // we have the python extension installed, but 0 valid python kernels / interpreters found
+    // just pop up a dialog box to prompt the user on how to install python
+    // Unlike installing the python extension we don't expect in progress executions to be handled
+    // when this command is installed, user will have to manually install python and rerun the cell
+    private async installPythonViaKernelPicker(): Promise<void> {
```
DonJayamanne, line 429: I think we can move this into a seprate class

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -215,6 +225,29 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
             this,
             this.disposables
         );
+
+        // Register our command that will handle installing the python extension via the kernel picker
+        this.disposables.push(
```
DonJayamanne, line 230: I think we can move this into a separate class.

src/notebooks/controllers/notebookControllerManager.ts
```diff
@@ -324,9 +357,19 @@ export class NotebookControllerManager implements INotebookControllerManager, IE
         // we don't end up with a kernel that asks to install python)
         if (!this.isWeb) {
             if ([...this.registeredControllers.values()].some((item) => isPythonKernelConnection(item.connection))) {
-                this.removeNoPythonControllers();
+                // We have some type of python kernel, turn off both install helper commands
+                await this.showInstallPythonExtensionContext.set(false);
+                await this.showInstallPythonContext.set(false);
             } else {
-                this.registerNoPythonControllers();
+                if (!this.extensionChecker.isPythonExtensionInstalled) {
```
DonJayamanne, line 364: I think this too can be moved into a serpate class.
Today we do expose the controllers via a public property and also have a method to load all of the controllers & wait for it to complete loading.

Basically, I'd like to ensure this complex class doesn't get more complicated (with additional logic & processing)

DonJayamanne (approved):
All good, but would prefer to move the code into a new file instead of placing them in NotebookControllerManager

package.nls.json
```diff
@@ -910,6 +910,7 @@
     },
     "DataScience.pythonExtensionRequired": "The Python Extension is required to perform that task. Click Yes to open Python Extension installation page.",
     "DataScience.pythonExtensionRequiredToRunNotebook": "Python Extension required to run Python notebooks.",
+    "DataScience.installingPythonExtension": "Installing Python extension and locating Python kernels.",
```
DonJayamanne, line 913: I'd consider changing this to `location Kernels`. as we can end up location non-python kernels that are installed in conda environments.

rchiodo (approved):
Looks good. I'd like to do what Don suggested too though. Move the logic to another class that just listens to the notebookControllerManager if at all possibe.

src/notebooks/types.ts
```diff
@@ -26,14 +26,14 @@ export interface INotebookControllerManager {
         controller: IVSCodeNotebookController;
         selected: boolean;
     }>;
+    readonly onNotebookControllersLoaded: Event<Readonly<IVSCodeNotebookController[]>>;
```
IanMatthewHuff, line 29: I debated back and forth a bit on this. I like keeping the commands independent from the notebook controller (at a basic level they could just be a small sub class or even functional) so they have their own service. However to keep the logic for when to show they need to know what controllers have been loaded. Previously we were not looking to expose this, however we're already using 'getRegisteredNotebookControllers' in product code, and it makes sense to me that other services could see a readonly list of what controllers are registered.

rebornix: 👏 how notebook controllers are registered (through detection or an async command/resolver) is irrelevant to externals IMHO so exposing events for `notebookControllers` make better sense than exposing another service/command.

Issue #9807, opened 2022-04-27
Debugging web tests is painful


rchiodo:
Console logger doesn't work - probably because it isn't enabled. Might need to do the same trick done with the URI for the jupyter server - write the isTestExecution() result into the bundled js file.

Debugging doesn't work because creating files using the 'vscode-test-web' scheme doesn't work. Could probably retry with the 'file' scheme if the first scheme fails.

DonJayamanne:
> Debugging doesn't 

Yes, source maps don't work. This is something i found last time i enabled bundling web tests into the extension. Sux.

DonJayamanne:
> do the same trick done with the URI for the jupyter server 

whats that?

DonJayamanne:
@rchiodo is this still an issue.
I think the PR i recently added resolves this issue, as we can now fully debug Web tests from within VS Code.

rchiodo:
Locally, but not when they fail on CI. I still think we need to log output somewhere (not necessarily to the console).

DonJayamanne:
> but not when they fail on CI. I still think we need to log output somewhere (not necessarily to the console).

I think during debt week can do this easily.
Just re-direct all of the console output to a log file.

Note: This PR https://github.com/microsoft/vscode-jupyter/pull/10512 now logs everything to the console window.

DonJayamanne:
@rchiodo I'm sorry but I do'nt quite understand this statement

> Locally, but not when they fail on CI. I still think we need to log output somewhere (not necessarily to the console).

Currently on CI we have enabled verbose logging, hence all of the logging is in the console window on CI, hence you can download the logs & look at what went wrong. We only log to the console based on some CI env variables.

* So we can get the logs today from the CI job (you can download the output of the job)
* Logging to console is only conditional

What more do we need or not need here.
I have a simple PR that logs the output of the test results to a file, however i see very little value in that, as we can get the full logs from the CI job today (without this PR, though you'll see a lot more in the logs files - stuff before the test task starts in the job).

rchiodo:
> I have a simple PR that logs the output of the test results to a file, however i see very little value in that, as we can get the full logs from the CI job today (without this PR, though you'll see a lot more in the logs files - stuff before the test task starts in the job).

When we discussed this in standup, we felt like having 100s of MBS of logs in the output doesn't really help (you can't read them in the github site) and it was potentially a slowdown. 

[...]

PR #10512, opened 2022-06-20
Enable verbose logging for web tests

DonJayamanne:
Partially addresses #9807 

Today we only log the errors in web tests, and this makes it difficult to figure out what was going on when the tests failed.
I.e. a lot of contextual information is missing, information we have in other tests & not in web tests.

* Enables verbose logging for web tests
* Re-directs console log messages fromm webview out to the extension host (e.g. error messages from widget libraries or the like)

### DIFF
--- a/build/launchWebTest.js
+++ b/build/launchWebTest.js
@@ -2,15 +2,31 @@
 // Licensed under the MIT License.
 
 const path = require('path');
+const fs = require('fs-extra');
 const test_web = require('@vscode/test-web');
 const { startJupyter } = require('./preLaunchWebTest');
+const jsonc = require('jsonc-parser');
+const extensionDevelopmentPath = path.resolve(__dirname, '../');
+const packageJsonFile = path.join(extensionDevelopmentPath, 'package.json');
+
 async function go() {
     let exitCode = 0;
     let server;
     try {
         server = (await startJupyter()).server;
-        const extensionDevelopmentPath = path.resolve(__dirname, '../');
         const bundlePath = path.join(extensionDevelopmentPath, 'out', 'extension.web.bundle');
+
+        // Changing the logging level to be read from workspace settings file.
+        // This way we can enable verbose logging and get the logs for web tests.
+        const settingsJson = fs.readFileSync(packageJsonFile).toString();
+        const edits = jsonc.modify(
+            settingsJson,
+            ['contributes', 'configuration', 'properties', 'jupyter.logging.level', 'scope'],
+            'resource',
+            {}
+        );
+        const updatedSettingsJson = jsonc.applyEdits(settingsJson, edits);
+        fs.writeFileSync(packageJsonFile, updatedSettingsJson);
 
         // Now run the test
         await test_web.runTests({
--- a/build/webpack/webpack.extension.web.config.js
+++ b/build/webpack/webpack.extension.web.config.js
@@ -8,9 +8,6 @@ const webpack = require('webpack');
 const constants = require('../constants');
 const CleanTerminalPlugin = require('clean-terminal-webpack-plugin');
 
-const prodEntry = {
-    extension: './src/extension.web.ts'
-};
 const devEntry = {
     extension: './src/extension.web.ts'
 };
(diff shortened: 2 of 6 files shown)


src/test/datascience/widgets/rendererUtils.ts
```diff
@@ -160,3 +161,46 @@ function getOutputCellIndex(outputItem: OutputItem): number | undefined {
...
     }
 }
+
+let consoleLoggersHijacked = false;
+/**
+ * Highjack the console log messages & send them to the extension host.
+ * On CI we cannot see the console log messages of the webview, so we need to hijack them.
+ * Sometimes when widgets fail this is very useful in diagnosing what went wrong on the webview side of things.
+ */
+function hijackLogging() {
```
DonJayamanne, line 171: Note: This is a test file `src/test/...`, hence this only applies to tests & this function is only called in the widget tests.

build/webpack/webpack.extension.web.config.js
```diff
@@ -12,8 +12,7 @@ const prodEntry = {
     extension: './src/extension.web.ts'
 };
 const devEntry = {
-    extension: './src/extension.web.ts',
-    'test/index': './src/test/web/index.ts' // source of the web extension test runner
```
DonJayamanne, line 16: I don't think this entry point is useful anymore.

src/test/datascience/widgets/commUtils.ts
```diff
@@ -22,8 +22,10 @@ export function initializeWidgetComms(serviceContainer: IServiceContainer): Util
     const disposable = messageChannel.onDidReceiveMessage(async ({ editor, message }) => {
         traceInfo(`Received message from Widget renderer ${JSON.stringify(message)}`);
```
DonJayamanne, line 23: Any message sent to extension host gets logged here.
Hence no need to explicitly handle the `log` command for logging purposes.

src/test/datascience/widgets/rendererUtils.ts
```diff
@@ -160,3 +161,46 @@ function getOutputCellIndex(outputItem: OutputItem): number | undefined {
...
+            message: `Handled message in Widget renderer ${message}`
+        });
+    };
+    ['log', 'error', 'warn'].forEach((channel) => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const originalConsoleLogger = (console as any)[channel] as unknown as Function;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (console as any)[channel] = (...args: any[]) => {
+            logMessage('log', args);
+            originalConsoleLogger(...args);
```
DonJayamanne, line 203: For widget tests this is crucial, we can now see why things fail, e.g. due to flaky logic/network or the like.

DonJayamanne: I had to write some of this in my other PRs to peek what was going on in webview tests

build/webpack/webpack.extension.web.config.js
rchiodo: Aren't you writing to this file below? I thought this means that index.ts is included in the bundle too.

rchiodo (approved):

build/webpack/webpack.extension.web.config.js
DonJayamanne: No its not included in the bundle, when we have multiple entry points such as `extension` and `test/index` then we end up with multiple bundles.

amunger:
related https://github.com/microsoft/vscode-jupyter/issues/9807

src/test/web/index.ts
```diff
@@ -54,6 +60,8 @@ export async function activate(context: IExtensionContext): Promise<IExtensionAp
             }
         });
     });
+
+    // Can't tell if on CI or not, so assume not.
```
amunger, line 64: setCI(true) is called at the top of this file - is that not reliable? I feel like either this line or the other isn't what it seems

build/launchWebTest.js
```diff
@@ -2,16 +2,32 @@
...
+        // This way we can enable verbose logging and get the logs for web tests.
+        const settingsJson = fs.readFileSync(packageJsonFile).toString();
+        const edits = jsonc.modify(
+            settingsJson,
+            ['contributes', 'configuration', 'properties', 'jupyter.logging.level', 'scope'],
+            'resource',
+            {}
+        );
+        const updatedSettingsJson = jsonc.applyEdits(settingsJson, edits);
+        fs.writeFileSync(packageJsonFile, updatedSettingsJson);
```
sadasant, line 29: I would assume it should be possible to achieve the same via environment variables. Is that not the case? It's just interesting to me

DonJayamanne: We canot use env variables for web tests as we're launching chrome.

src/test/web/index.ts
DonJayamanne: Removed the old comment

build/launchWebTest.js
sadasant: While that is true, there are ways to pass environment variables to webpack: https://webpack.js.org/guides/environment-variables/

(please bear with me, I'm out of context)

DonJayamanne:
@amunger thank for linking the existing issue, didn't realize there was one.
I found it very painful when trying to figure out what went wrong, submitting this as a seprate PR instead of incluing it with my other pr where i made these changes to see what was going on.

amunger (approved):

IanMatthewHuff (approved):

PR #10518, opened 2022-06-21
More ts unused

rebornix:
Re #10307 #10152 

After a week of refactoring, I did another pass of ts unused to check what components are no longer used in the project. This PR includes deletion of code that are not being used by anyone.

There are still places but I'm not very confident with removing all of them. Will follow up on them with team and also discuss if we could have this ts-unused scanning run regularly.


src/kernels/installer/pyenv.node.ts
```diff
@@ -1,232 +0,0 @@
-import * as path from '../../platform/vscode-path/path';
```
rchiodo, line 1: This seems like a bug. We should support pyenv. Maybe I forgot to port something from the python extension.

rchiodo (approved):

src/kernels/installer/pyenv.node.ts
rebornix: @rchiodo this is the code we moved from Python into Jupyter for Python Installer. The files included unit tests and code checking pyenv but this is not used by us (Jupyter). We have our own pyenv support in https://github.com/microsoft/vscode-jupyter/blob/ecc0aaf4fab122f24375fe0e1afd9c59548dddca/src/kernels/installer/pyenv.node.ts#L1

PR #10417, opened 2022-06-10
updated node version

amunger:
match up with what vscode is using

PR #10530, opened 2022-06-22
Update node version to match with whats in VSCode

DonJayamanne:
VS Code is using 16.13.2, not 16.14.2

@amunger noticed you updated it(https://github.com/microsoft/vscode-jupyter/pull/10417/files)  from 16.13.0 to 16.14.2, however VSC Insiders is using 16.13.2 and Stable is using 16.13.0.
I'm assuming 16.14 was a mistake hence changing to 16.13.2


rchiodo (approved):

amunger:
vscode build throws an error if you are using <14

Issue #10435, opened 2022-06-13
Rework remote support to account for more scenarios


rchiodo:
Remote support is confusing:
- [ ] https://github.com/microsoft/vscode-jupyter/issues/10363
- [x] https://github.com/microsoft/vscode-jupyter/issues/10191

Using this issue to describe the use cases.

## Use Cases

1. (web) User opens a notebook and wants to run a cell so they can see the output of a cell
2. (web) User is in a notebook and wants to run a cell in the browser so they don't have to pay for/setup a remote compute
3. (web) User is in a notebook and wants to run a cell on their local machine so they don't have to pay for/setup a remote compute
4. (web) User is in a notebook and wants to run a cell on a different remote server because this one is more expensive
5. (desktop) User opens a notebook and wants to run a cell without entering anything so they can see the output
6. (desktop) User opens a notebook and wants to run it on their own compute so they can use the GPU/CPU on that machine
7. (desktop) User wants to specify their own compute for all notebooks so they don't have to keep picking a server
8. (desktop) User wants to run some cells on local and some cells on remote (need to check telemetry for this)
9. (codespaces) User wants to run a cell and not have to specify anything except clicking the run button.

rchiodo:
Use case 1 - Web run a cell.

Solution ideas:

- User opens notebook, run button is immediately available
- Command for 'Connect to your own jupyter server' appears.
- It loads UI for picking a URI. User has one step to enter the URI or pick from MRU list
- Once URI is picked, the 'default' kernel from the remote is selected and the cell runs.
- Jupyter:Local/Remote is removed from the UI as it's not necessary.

rchiodo:
What does the user expect to be in the kernel drop down list?

What would their expectations be in these scenarios?

- Web first time
- Web second time
- Desktop first time
- Desktop second time
- Codespaces first time
- Codespaces second time

rchiodo:
Use case 2 - Web run a cell in the browser

Solution 1

- Depend upon pyodide in the web extension
- User gets option to use pyodide kernel automatically?

Solution 2

- Add command for 'Run in browser (with limited support)'
- This installs pyodide extension
- Runs cell with pyodide kernel

rchiodo:
Use case 3 - Web run a cell on local machine

Solution 1
- Command to run on local machine
- It goes through the steps of reopening it in VS code locally? Or maybe installing the VS code server and using that as the entry point? 


Could do this later, but command would be a way to kick it off.

[...]

PR #10537, opened 2022-06-22
Split NotebookControllerManager

rchiodo:
For #10435 

Split NotebookControllerManager into 5 parts as a prereq for redesigning some of this:
- IControllerSelection - selected kernel of a notebook and different events
- IControllerLoader - responsible for loading kernels on startup or during refresh events
- IControllerRegistration - list of all of the controllers ever created
- IControllerPreferredService - computes the preferred kernel for a notebook
- IControllerDefaultService - computes a 'default' controller. This is the active interpreter in local and the 'default' Python 3 kernel in remote.


src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -0,0 +1,190 @@
...
+                    // Update our list kernel connections.
+                    this.registeredConnections.set(id, metadata);
+
+                    // See if we already created this controller or not
+                    const controller = this.registeredControllers.get(id);
+                    if (controller) {
+                        // If we already have this controller, its possible the Python version information has changed.
+                        // E.g. we had a cached kernlespec, and since then the user updated their version of python,
+                        // Now we need to update the display name of the kernelspec.
+                        controller.updateConnection(metadata);
```
rchiodo, line 91: One of the changes made was to always update existing controllers if the metadata id is a match. We were only doing this for specific cases before but I figured why not just always update it.

src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -0,0 +1,190 @@
...
+                        return false;
+                    } else if (this.kernelFilter.isKernelHidden(metadata)) {
```
rchiodo, line 93: All filtering is done in this one spot now.

src/notebooks/controllers/controllerDefaultService.ts
```diff
@@ -0,0 +1,125 @@
...
+        viewType: typeof JupyterNotebookView | typeof InteractiveWindowView
+    ): Promise<IVSCodeNotebookController | undefined> {
+        if (this.isLocalLaunch) {
+            traceInfoIfCI('CreateActiveInterpreterController');
+            return createActiveInterpreterController(viewType, resource, this.interpreters, this.registration);
+        } else {
+            traceInfoIfCI('CreateDefaultRemoteController');
+            const notebook =
+                viewType === JupyterNotebookView
+                    ? this.notebook.notebookDocuments.find((item) => item.notebookType === viewType)
```
DonJayamanne, line 44: This would be wrong if we have multiple notebooks, we should look for the Uri as well.
I think passin g the notebook as a arg would fix this issue.

src/notebooks/controllers/controllerDefaultService.ts
```diff
@@ -0,0 +1,125 @@
...
+        @inject(IControllerLoader) private readonly loader: IControllerLoader,
+        @inject(IInterpreterService) private readonly interpreters: IInterpreterService,
+        @inject(IVSCodeNotebook) private readonly notebook: IVSCodeNotebook,
+        @inject(IDisposableRegistry) readonly disposables: IDisposableRegistry,
+        @inject(ServerConnectionType) private readonly serverConnectionType: ServerConnectionType,
+        @inject(IsWebExtension) private readonly isWeb: boolean
+    ) {}
+    public async computeDefaultController(
+        resource: Resource,
+        viewType: typeof JupyterNotebookView | typeof InteractiveWindowView
```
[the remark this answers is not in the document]
rchiodo: When this is called for the interactive window there is no notebook.

src/notebooks/controllers/controllerDefaultService.ts
rchiodo: This is the exact code as we already had. I just copied it.

src/notebooks/controllers/controllerDefaultService.ts
```diff
@@ -0,0 +1,125 @@
...
+            }
+            traceVerbose('No default remote controller, hence returning the active interpreter');
+            return createActiveInterpreterController(viewType, resource, this.interpreters, this.registration);
+        }
+    }
+
+    @traceDecoratorVerbose('Get default Remote Controller')
+    private async createDefaultRemoteController(
+        notebookType: typeof JupyterNotebookView | typeof InteractiveWindowView,
+        notebook?: NotebookDocument
```
DonJayamanne, line 59: I don't think we need the first argument, the notebook type is a property of NotebookDocument object.

rchiodo: Yeah if there's a notebook document. In the case of the IW, there isn't one yet.

src/notebooks/controllers/controllerLoader.ts
```diff
@@ -0,0 +1,260 @@
...
+                    1_000
+                );
+            },
+            this,
+            this.disposables
+        );
+
+        // Make sure to reload whenever we do something that changes state
+        const forceLoad = () => this.loadControllers(true);
+        this.serverUriStorage.onDidChangeUri(forceLoad, this, this.disposables);
```
DonJayamanne, line 77: Would be best to move all of these to the `activate` method, else seems inconsistent to have some event handlers in the ctro and others in `activate`

src/notebooks/controllers/controllerDefaultService.ts
rchiodo: Here's that same method in the original file:

```typescript
    public async getActiveInterpreterOrDefaultController(
        notebookType: typeof JupyterNotebookView | typeof InteractiveWindowView,
        resource: Resource
    ): Promise<IVSCodeNotebookController | undefined> {
        if (this.isLocalLaunch) {
            traceInfoIfCI('CreateActiveInterpreterController');
            return this.createActiveInterpreterController(notebookType, resource);
        } else {
            traceInfoIfCI('CreateDefaultRemoteController');
            const notebook =
                notebookType === JupyterNotebookView
                    ? this.notebook.notebookDocuments.find((item) => item.notebookType === notebookType)
                    : undefined;
            const controller = await this.createDefaultRemoteController(notebookType, notebook);
            // If we're running on web, there is no active interpreter to fall back to
            if (controller || this.isWeb) {
                return controller;
            }
            traceVerbose('No default remote controller, hence returning the active interpreter');
            return this.createActiveInterpreterController(notebookType, resource);
        }
    }

```

src/notebooks/controllers/controllerLoader.ts
rchiodo: I disagree. At least the initial load has to happen in the ctor. 

Otherwise another IExtensionSingleActivationService could take a dependency on the controllerLoader and have the loadpromise not even start yet.

src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -0,0 +1,318 @@
...
+        serverId?: string | undefined
+    ): Promise<{
+        preferredConnection?: KernelConnectionMetadata | undefined;
+        controller?: IVSCodeNotebookController | undefined;
+    }> {
+        traceInfoIfCI(`Clear controller mapping for ${getDisplayPath(document.uri)}`);
+        const loadControllersPromise = this.loader.loaded;
+        // Keep track of a token per document so that we can cancel the search if the doc is closed
+        const preferredSearchToken = new CancellationTokenSource();
+        const disposable = this.notebook.onDidCloseNotebookDocument(
```
DonJayamanne, line 86: We'll end up creating multiples of these for each `computePreferred` method.
I think we could instead just create a common handler in the `active` method and store the cancellation tokesn in a dictionary.
& then the handler is invoked we look for items in that dictionary (keyed by document) & then cancel that token.

src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+    public async computePreferred(
+        document: NotebookDocument,
+        serverId?: string | undefined
+    ): Promise<{
+        preferredConnection?: KernelConnectionMetadata | undefined;
+        controller?: IVSCodeNotebookController | undefined;
+    }> {
+        traceInfoIfCI(`Clear controller mapping for ${getDisplayPath(document.uri)}`);
+        const loadControllersPromise = this.loader.loaded;
```
DonJayamanne, line 83: Do we need this variable, why not just remove this variable?

src/notebooks/controllers/controllerPreferredService.ts
rchiodo: That does seem better. But I didn't write any of this. This is all existing code. I guess that seems harmless enough, but I was trying to not change any of the logic (or as little of the logic as possible).

src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -0,0 +1,318 @@
...
+                    return {};
+                }
+
+                traceInfo(
+                    `PreferredConnection: ${preferredConnection.id} found for NotebookDocument: ${getDisplayPath(
+                        document.uri
+                    )}`
+                );
+
+                const targetController = this.registration.values.find(
```
DonJayamanne, line 164: Shouldn't we wait for the controllers to be loaded here just like we do in line 177?
Else we could end in a race condition, where we look for some remote controllers but the promise hasn't competed hence the preferred controller will not be setup.

src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -0,0 +1,318 @@
...
+                    'interactive'
+                );
+                preferredConnection = defaultInteractiveController?.connection;
+            }
+
+            // See if the preferred connection is in our registered controllers, add the sufix for the interactive scenario
+            let targetController;
+            if (preferredConnection) {
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                targetController = this.registration.get(preferredConnection, document.notebookType as any);
```
DonJayamanne, line 191: Same here, don't we need to controllers to wait until they have loaded?

src/interactive-window/interactiveWindowProvider.ts
```diff
@@ -178,7 +181,7 @@ export class InteractiveWindowProvider
                 resource,
                 mode,
                 this.serviceContainer.get<IExportDialog>(IExportDialog),
-                this.notebookControllerManager,
+                this.serviceContainer.get<IControllerSelection>(IControllerSelection),
```
DonJayamanne, line 184: Isn't this an indication of some cyclic reference (feels weird for the new code to have cyclic references)

src/notebooks/controllers/controllerRegistration.ts
DonJayamanne: Yeah, i guess it makes it siimpler to read and maintain

src/notebooks/controllers/controllerPreferredService.ts
rchiodo: It's used below.

src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -0,0 +1,318 @@
...
+            let preferredConnection: KernelConnectionMetadata | undefined;
+            // Don't attempt preferred kernel search for interactive window, but do make sure we
+            // load all our controllers for interactive window
+            const notebookMetadata = getNotebookMetadata(document);
+            const resourceType = getResourceType(document.uri);
+            const isPythonNbOrInteractiveWindow = isPythonNotebook(notebookMetadata) || resourceType === 'interactive';
+            if (document.notebookType === JupyterNotebookView && !this.isLocalLaunch && isPythonNbOrInteractiveWindow) {
+                const defaultPythonController = await this.defaultService.computeDefaultController(
+                    document.uri,
+                    document.notebookType
```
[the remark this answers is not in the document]
rchiodo: We are? I think you're referring to the api just needing a document? The document isn't available in the IW case.

src/notebooks/controllers/controllerPreferredService.ts
rchiodo: It doesn't do that now, so it's an existing but but it does seem like it should wait right here.

src/notebooks/controllers/controllerPreferredService.ts
rchiodo: If the two code paths above both wait, then no.

src/notebooks/controllers/controllerDefaultService.ts
DonJayamanne: Might have to change that then, looks like a bug

src/interactive-window/interactiveWindowProvider.ts
rchiodo: No it's not. I was just lazy. THere's already a bunch of serviceContainer calls, I didn't feel like importing another item into the provider.

src/notebooks/controllers/controllerPreferredService.ts
DonJayamanne: Aagreed,

src/notebooks/controllers/controllerPreferredService.ts
DonJayamanne: Else it gives the impression that it could change at a later point hence the need to capture into a variable.

src/notebooks/controllers/controllerPreferredService.ts
DonJayamanne: Yes, but if they don't, e.g. in the case of a remote kernels (first time opening a new notebook) or the like.
E.g. assume a user doesn't have any local kernels, then looking at this code path we don't wait for remote kernels to finish loading. 

Or Do we? Based on the code it doesn't look like we do.

src/notebooks/controllers/controllerPreferredService.ts
rchiodo: Maybe before it was so it was this specific instance of loading? That's what it does. It's only going to wait for loading that was going on when this function started. If a refresh happens before we get to the bottom, that refresh doesn't have to be waited on. But maybe we want to wait on any refreshes too.

src/notebooks/controllers/controllerPreferredService.ts
rchiodo: No actually it's on purpose. See the comment below. If it isn't loaded yet, we just create it instead.

rchiodo:
I think I'd rather leave the algorithm the way it is unless we can prove there's a bug. There's some subtle behavior here and I'd rather not change it if I can.

src/notebooks/controllers/controllerLoader.ts
```diff
@@ -0,0 +1,261 @@
...
+        @inject(IKernelFinder) private readonly kernelFinder: IKernelFinder,
+        @inject(IKernelProvider) private readonly kernelProvider: IKernelProvider,
+        @inject(IPythonExtensionChecker) private readonly extensionChecker: IPythonExtensionChecker,
+        @inject(IInterpreterService) private readonly interpreters: IInterpreterService,
+        @inject(IJupyterServerUriStorage) private readonly serverUriStorage: IJupyterServerUriStorage,
+        @inject(ServerConnectionType) private readonly serverConnectionType: ServerConnectionType,
+        @inject(IControllerRegistration) private readonly registration: IControllerRegistration
+    ) {
+        this.loadControllers(true).ignoreErrors();
+    }
```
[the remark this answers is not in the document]
rchiodo: No this guarantees we start the load promise as soon as this object is created. Activate doesn't guarantee dependency ordering so moving it there means somebody could access the 'loaded' promise before we activate.

src/notebooks/controllers/controllerDefaultService.ts
rchiodo: Yeah sure looks like it especially since it uses the notebook metadata. I'll add an issue.

rchiodo: https://github.com/microsoft/vscode-jupyter/issues/10542

DonJayamanne (approved):

src/notebooks/controllers/controllerPreferredService.ts
DonJayamanne: Make sense, thought I do'nt think we'd need to wait on the refreshes, because that would happen before the call to this method.
I.e. i'd expect the calling code to orchestrate that behavior, e.g. refresh, & then call this method - not call this method & then call refresh.

I'm not fussed either way, marking this comment of mine as resolved.

PR #10538, opened 2022-06-22
top module as docs

rebornix:
Re #10152 

This PR covers the discussion we had in yesterday's standup: top level files for both core utilities and living docs. It introduces:

* `src/commands.ts`, which includes all the typings for command names and types for return result.
* `src/telemetry.ts`, which includes all the typings for telemetry event names and value types.
  * the telemetry helper functions are now moved into `src/platform/telemetry`, they are not allowed to import typings from other top components but only serve as helper utilities

We would use these two files as living docs and type checking for command execution and telemetry. For telemetry, we might want to move towards VS Code's tooling and break it down, and when that happens, we will only keep `src/platform/telemetry`.


src/commands.ts
```diff
@@ -13,12 +13,12 @@ import {
     Uri,
     ViewColumn
 } from 'vscode';
-import { IShowDataViewerFromVariablePanel } from '../../messageTypes';
```
rebornix, line 16: There are 59 file modified but most of them are import path changes and file movement.

src/platform/telemetry/startupTelemetry.ts
```diff
@@ -1,13 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
```
rebornix, line 1: In followup PRs, I would probably move these files to where they should be. For example, `sendStartupTelemetry` is only used by `extension.ts` so it should not sit in `platform` as it's not used by anyone else.

src/telemetry.ts
```diff
@@ -11,17 +11,19 @@ import {
...
+import { EnvironmentType, PythonEnvironment } from './platform/pythonEnvironments/info';
+import { TelemetryErrorProperties, ErrorCategory } from './platform/errors/types';
+import { ExportFormat } from './notebooks/export/types';
+import { InterruptResult, KernelConnectionMetadata, KernelInterpreterDependencyResponse } from './kernels/types';
+import { IExportedKernelService } from './standalone/api/extension';
+import { PreferredKernelExactMatchReason } from './notebooks/controllers/notebookControllerManager';
+import { SelectJupyterUriCommandSource } from './kernels/jupyter/serverSelector';
+import { TerminalShellType } from './platform/terminals/types';
+
+export * from './platform/telemetry/index';
```
rebornix, line 26: We would move some of the functions into this file from `platform/telemetry/index` instead of re-export in following PRs.

IanMatthewHuff (approved):

rchiodo (approved):

PR #10540, opened 2022-06-22
wait for interactive window input code to finish

amunger:
fix flakey export test

### DIFF
--- a/src/test/datascience/helpers.ts
+++ b/src/test/datascience/helpers.ts
@@ -99,9 +99,21 @@ export async function createStandaloneInteractiveWindow(interactiveWindowProvide
     return activeInteractiveWindow;
 }
 
-export async function insertIntoInputEditor(source: string) {
-    // Add code to the input box
-    await vscode.window.activeTextEditor?.edit((editBuilder) => {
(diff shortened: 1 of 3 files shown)


rchiodo (approved):

rebornix (approved):

src/test/datascience/helpers.ts
```diff
@@ -219,6 +219,16 @@ export async function waitForInteractiveWindow(
     return notebookDocument!;
 }
 
+export async function runInteractiveWindowInput(
+    code: string,
+    interactiveWindow: InteractiveWindow,
+    newCellCount: number
+) {
+    await insertIntoInputEditor(code);
+    await vscode.commands.executeCommand('interactive.execute');
```
DonJayamanne, line 228: Might want to port this code into other plaes where we call `interactive.execute`, else we'll end up with the same issues.

DonJayamanne (changes requested):
Might want to port this code into other plaes where we call `interactive.execute`, else we'll end up with the same issues.

DonJayamanne (approved):

PR #10545, opened 2022-06-23
Web version of the DataFrame

### LINKED ISSUE #9665: DataFrame working in web extension

### LINKED ISSUE #10310: De-reference the constructor from the rendering of VariableExplorerButtonCellFormatter
The `VariableExplorerButtonCellFormatter` is currently tying the constructor of the `VariableExplorer` in: https://github.com/microsoft/vscode-jupyter/blob/main/src/webviews/webview-side/interactive-common/variableExplorer.tsx#L125

We should not take the reference from the constructor, for performance reasons.
(though I haven't tested this kind of performance in a while, but seems reasonable to move this state-referencing out of the constructor)

sadasant:
The only thing missing on the web to enable the DataFrame features was to have a web-ready version of the `DataViewerDependencyService`.

After discussing ideas with @DonJayamanne  and @rchiodo , Rich's approach seemed the most straightforward way to add web support to `DataViewerDependencyService`. Rich's approach, as far as I understood, was to:

- `DataViewerDependencyService` to take a dependency on the `IKernel` and have it execute a `%pip install pandas` instead of using `IInstaller`.
  - It would use either `%conda` or `%pip` based on the `IKernel`'s `KernelConnectionMetadata`. `%conda` if it had an interpreter (local case) that was conda based. Otherwise `%pip`.
- For figuring out the version of pandas, use the kernel. Just import pandas and print the version.

This PR follows that approach. Besides that:

- Moves the `parseSemVer` function from `platform/common/utils.node` to `platform/common/utils`.
- Adds `IDataViewerDependencyService` to the extension-side web service registry.
- `removes the `isWeb` conditional from the `variableExplorerButtonCellFormatter.tsx`.

Thus:
Fixes #9665 
Fixes #10310

Feedback always appreciated!

---


src/webviews/extension-side/serviceRegistry.node.ts
```diff
@@ -1,67 +1,67 @@
...
-}
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+'use strict';
+
+import { IExtensionSingleActivationService, IExtensionSyncActivationService } from '../../platform/activation/types';
+import { IServiceManager } from '../../platform/ioc/types';
+import { DataViewer } from './dataviewer/dataViewer';
+import { DataViewerCommandRegistry } from './dataviewer/dataViewerCommandRegistry';
+import { DataViewerDependencyService } from './dataviewer/dataViewerDependencyService';
```
sadasant, line 9: This file looks like it was completely rewritten unless you hide the whitespaces. I'm guessing it has something to do with the line breaks. Anyway, to hide the whitespaces, you can click here: https://github.com/microsoft/vscode-jupyter/pull/10545/files?diff=unified&w=1

rchiodo (approved):
Looks good, although I think you could have made the unit tests much simpler with a sinon.stub of 'executeSilently' instead of having to mock the kernel behavior.

src/webviews/extension-side/serviceRegistry.node.ts
sadasant: Do you mean `npm run prettier-fix`? I am running it and I see no difference.

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
```diff
@@ -35,87 +59,347 @@ suite('DataScience - DataViewerDependencyService', () => {
...
+
+    test('What happens if there are no idle kernels?', async () => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (kernel.session as any) = jupyterSession;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (kernel.status as any) = 'busy';
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (kernelProvider.kernels as any) = [kernel];
+
+        shellFuture = {
```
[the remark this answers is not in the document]
sadasant: I might be wrong but I assume that plain functions are not very easy to mock? One would need to change when they are being loaded.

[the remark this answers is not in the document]
rchiodo: Do a search for sinon.stub. It's actually very easy to mock them.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,115 @@
...
+                sendTelemetryEvent(Telemetry.UserInstalledPandas);
+            }
+        } else {
+            sendTelemetryEvent(Telemetry.UserDidNotInstallPandas);
+            throw new Error(DataScience.pandasRequiredForViewing());
+        }
+    }
+
+    private async getVersionOfPandas(): Promise<SemVer | undefined> {
+        const outputs = await executeSilently(this.kernelSession, `import pandas; print(pandas.__version__)`);
```
IanMatthewHuff, line 105: This is the bit that's not working for the "Show in Data Viewer" scenario. We are just debugging python here. What we can do here is instead of the interpreter pass the activeDebugSession. Then we can execute this same code on the debug session.

debuggerVariables.evaluate has an example of code execution.

IanMatthewHuff (changes requested):
We chatted in teams. Looks like this change will be a forcing factor for checking pandas version in the python debug scenario the right way. Instead of executing the code snippit on the kernel session we'll execute it on the activeDebugSession.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
DonJayamanne: I think we should delerte the pandas module if we didn't import this, else we end up modifying user state by importing modules.
E.g.
```
__vsc_pandas_previously_imported = sys.modules['pandas']
try:
	import pandas as __vsc_pandas
	print(__vsc_pandas.version)
finally:
	del sys.modules['pandas']
```

DonJayamanne: @IanMatthewHuff can we not fix that as a seprate issue, i thought thats an existing issue today.
I.e. we can focus on fixing that issue in a separate PR.

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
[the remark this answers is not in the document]
DonJayamanne: I don't think we can stub them, it works only if the stubbing is applied before the module is imported.
However since the imports are at the top level it might not work (as all modules end up getting loaded & imported before the tests run)

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,115 @@
...
+    return version.compare(minimumSupportedPandaVersion) > 0;
+}
+
+/**
+ * Responsible for managing dependencies of a Data Viewer.
+ */
+@injectable()
+export class DataViewerDependencyService implements IDataViewerDependencyService {
+    private get kernel(): IKernel {
+        const kernel = this.kernelProvider.kernels.find((kernel) => kernel.status === 'idle');
```
DonJayamanne, line 33: This looks wrong, we're just looking for any random kernel, this method could even end up returning an idle `rust` kernel

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
```diff
@@ -35,87 +59,347 @@ suite('DataScience - DataViewerDependencyService', () => {
...
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } as any as IShellFuture<IExecuteRequestMsg, IExecuteReplyMsg>;
 
-        dependencyService = new DataViewerDependencyService(
-            instance(appShell),
-            instance(installer),
-            instance(pythonExecFactory),
-            instance(interpreterService),
-            false
+        const resultPromise = dependencyService.checkAndInstallMissingDependencies(interpreter);
```
DonJayamanne, line 83: This API doesn't look right, i don't think we can determine the kernel from the interpreter, we should instead be passing the notebook document or the kernel, from there on we can then determine the interpreter.

DonJayamanne (changes requested):
API doesn't look right, we're picking random kernels, instead of a the kernel associated with the current execution.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
DonJayamanne: > This change breaks the old code path for that scenario. 

Oh, if it breaks it then we have no choice but fix it here, I take back what i suggested, we cannot break something and then leave it for another PR.

> Would rather not have to do the work not, but the curr

I agree, I missed the fact that this would break existing functionality.
Feels like we are missing a test case for that, else we would have shipped a breaking change had you not caught this.

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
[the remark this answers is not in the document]
sadasant: Yes, but that really depends on how the module is built. Plain functions can’t be stubbed. That doesn’t mean it can’t work in the case of executeSilently, I haven’t taken the time yet.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
rchiodo: Whoops I missed this. You have an interpreter (which by the way isn't going to work in the web version). That could be used to find a matching kernel.

However at the point where this service is called, you could just pass it the kernel instead. Don't send it the interpreter, send it the kernel in use.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,115 @@
...
+        return isConda ? 'conda' : 'pip';
+    }
+
+    constructor(
+        @inject(IApplicationShell) private readonly applicationShell: IApplicationShell,
+        @inject(IKernelProvider) private readonly kernelProvider: IKernelProvider,
+        @inject(IsCodeSpace) private isCodeSpace: boolean
+    ) {}
+
+    public async checkAndInstallMissingDependencies(interpreter: PythonEnvironment): Promise<void> {
```
rchiodo, line 59: You'd pass the IKernel here instead of the environment.

rchiodo: Or maybe even a function that can execute code and return the results. Then it wouldn't need to know about debuggers or kernels.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
DonJayamanne: That was my original suggestion to Daniel, to pass the kernel into this method

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
DonJayamanne: Yes, this is what I suggested to @sadasant when I spoke to him about this, I think I didn't explain it well enough.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: For a while I tried to use the debugger to run the python code to check the pandas version and install it, and while it worked in the sense that I could see the execution in the debugger, I could not get the output to come back in the results.

It also drastically changed the signature of the dependency installer, from just receiving a kernel to variably receiving either a kernel or a debug session, for example.

Given that, as far as I could tell, the "Show in Data Viewer" scenario of the debugger is not implemented in the web, and with the idea of simplifying the implementation as much as possible, I have decided to keep the older dependency installer as the `.node` version of the installer, and to only use `IKernel` in the isomorphic version of the installer.

Note: At some point, Ian did suggest me this approach I ended up with, but I really wanted to see if I could make it work with the debugger. Although not being able to make it work is my fault and not the debugger's the complexity of the implementation of the dependency installer at the end felt gross, so I was also not motivated to move forward with supporting both IKernel and the debugger session in the isomorphic installer.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: I have resolved this! Thank you for pointing this out.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: Now I'm passing IKernel! Thank you for your guidance 🙏

src/webviews/extension-side/serviceRegistry.node.ts
DonJayamanne: I've been using wsl, mac, Linux and windows and never had any such issues.
i believe ian is on a mac and rich has used wsl as well.

not sure we need such a change.

sadasant: so, the underlying problem is that the line endings have changed -- not sure what else to propose

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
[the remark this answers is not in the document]
sadasant: Ok it worked! yay 🙂 had to change imports, but not in a bad way. Here's the commit: https://github.com/microsoft/vscode-jupyter/pull/10545/commits/ae2d66776f9e3984c55e79cd5d032061af7a1c19

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,123 @@
...
+import { Product } from '../../../kernels/installer/types';
+import { IApplicationShell } from '../../../platform/common/application/types';
+import { traceWarning } from '../../../platform/logging';
+import { IsCodeSpace } from '../../../platform/common/types';
+import { parseSemVer } from '../../../platform/common/utils';
+import { DataScience, Common } from '../../../platform/common/utils/localize';
+import { EnvironmentType } from '../../../platform/pythonEnvironments/info';
+import { sendTelemetryEvent, Telemetry } from '../../../telemetry';
+import { IDataViewerDependencyService } from './types';
+import * as helpers from '../../../kernels/helpers';
```
DonJayamanne, line 18: Feels like a poor way to get the tests working.
Tomorrow if someone were to change this style of import, then the tests would fail, 
i.e. i don't think we should force the import style just to get the tests passing.

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
```diff
@@ -4,118 +4,99 @@
...
+
     test('Throw exception if pandas is installed and version is = 0.20', async () => {
-        when(
-            pythonExecService.exec(deepEqual(['-c', 'import pandas;print(pandas.__version__)']), anything())
-        ).thenResolve({ stdout: '0.20.0' });
+        const version = '0.20.0';
 
-        const promise = dependencyService.checkAndInstallMissingDependencies(interpreter);
+        const stub = sinon.stub(helpers, 'executeSilently');
+        stub.returns(Promise.resolve([{ ename: 'stdout', output_type: 'stream', text: version }]));
```
DonJayamanne, line 58: Same comment as previously

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,123 @@
...
+                throw new Error(DataScience.failedToInstallPandas());
+            }
+        } else {
+            sendTelemetryEvent(Telemetry.UserDidNotInstallPandas);
+            throw new Error(DataScience.pandasRequiredForViewing());
+        }
+    }
+
+    private async getVersionOfPandas(kernel: IKernel): Promise<SemVer | undefined> {
+        const command = 'import pandas;print(pandas.__version__)';
```
DonJayamanne, line 101: I think we should change this to:
```python
import pandas as _VSCODE_pandas
print(_VSCODE_pandas.__version__)
del _VSCODE_pandas
```

This way we don't change the imported modules in users code.
I.e. if user didn't import pandas we shoudln't leave this module imported into the global (current) scope

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
```diff
@@ -4,118 +4,99 @@
...
+
     test('All ok, if pandas is installed and version is > 1.20', async () => {
-        when(
-            pythonExecService.exec(deepEqual(['-c', 'import pandas;print(pandas.__version__)']), anything())
-        ).thenResolve({ stdout: '0.30.0' });
-        await dependencyService.checkAndInstallMissingDependencies(interpreter);
+        const version = '3.3.3';
+
+        const stub = sinon.stub(helpers, 'executeSilently');
+        stub.returns(Promise.resolve([{ ename: 'stdout', output_type: 'stream', text: version }]));
```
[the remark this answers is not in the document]
sadasant: I forgot to spy the calls and read what was sent to them -- but, since I agree with you that the imports shouldn't affect the tests, I am just going to revert using Sinon.stubs for the `executeSilently` function.

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
sadasant: I forgot to spy the calls and read what was sent to them -- but, since I agree with you that the imports shouldn't affect the tests, I am just going to revert using Sinon.stubs for the `executeSilently` function.

rchiodo: Wait. You shouldn't have to modify the import. This test here doesn't:

https://github.com/microsoft/vscode-jupyter/blob/4af7b2a2fe854871a2733babfe5bf36d506b6f39/src/test/datascience/jupyter/kernels/installationPrompts.vscode.test.ts#L346

Sinon should be able to stub the module as long as you import the module under test AFTER the stubbed module has been imported into the test module.

sadasant: yes, but how am I going to import the module after I stub it if I have the module imported at the top of the file?

rchiodo: dataViewerDependencyService.ts = moduleUnderTest 
kernel/helpers.ts = module you need to stub

Top of file:

```
import * as helpers from '.../kernels/helpers.ts'
import { DataViewerDependencyService } from '.../dataviewerDependencyService.ts'

..

suitesetup(() => {
    sinon.stub(helpers, 'executeSilently')
}
```
This is the pattern followed in other locations. helpers is imported before dataviewer dependency so the module you have is the one used by both.

rchiodo: Example:
https://sinonjs.org/how-to/stub-dependency/

sadasant: Well, but that is what I have as part of the PR right now (there are a couple of small cleanups that I haven't pushed, but specifically regarding this import), I have: `import * as helpers from '../../../kernels/helpers';` but, as Don said here: https://github.com/microsoft/vscode-jupyter/pull/10545#discussion_r906112779

> Feels like a poor way to get the tests working.
Tomorrow if someone were to change this style of import, then the tests would fail,
i.e. i don't think we should force the import style just to get the tests passing.

Kingpancho:
Cancel subscription to

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
sadasant: By changing the imports I believe Don and I refer to the fact that it was `import { executeSilently }` inside of the `dataviewerDependencyService` before I started using Sino.stub

I believe Don's point is that if someone changes that import (as I currently have in this PR on `dataviewerDependencyService`) for something other than `import * as helpers from '../../../kernels/helpers';`, the test will fail.

Please, let me know if I'm missing something.

sadasant: I personally think `import * as helpers from '../../../kernels/helpers';` is ok.

Kingpancho:
Cancel stop sending emails

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
rchiodo: You shouldn't need to change the import in DataViewerDependencyService.

sadasant: I'll try!

sadasant: Seems like it's working! I believe this didn't use to work before (like, in the before times), but my memory is really bad.

src/webviews/extension-side/serviceRegistry.node.ts
sadasant: I changed my .gitconfigs and made sure to undo these changes 👍 though for the record, I do prefer LF ends of lines, just like I prefer formatting enforced 😆

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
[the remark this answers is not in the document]
sadasant: I pushed a change that now verifies the parameters sent to `executeSilently`.

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
sadasant: Updated!

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: Updated!

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,124 @@
...
+
+function isVersionOfPandaSupported(version: SemVer) {
+    return version.compare(minimumSupportedPandaVersion) > 0;
+}
+
+/**
+ * Responsible for managing dependencies of a Data Viewer.
+ */
+@injectable()
+export class DataViewerDependencyService implements IDataViewerDependencyService {
```
IanMatthewHuff, line 33: Does this need .web file naming since it's the class we inject in the web scenario? I'm guessing you might not have since it doesn't actually have web only dependencies, but in function is seems to be.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,124 @@
...
+    constructor(
+        @inject(IApplicationShell) private readonly applicationShell: IApplicationShell,
+        @inject(IsCodeSpace) private isCodeSpace: boolean
+    ) {}
+
+    private packaging(kernel: IKernel): 'pip' | 'conda' {
```
IanMatthewHuff, line 39: There are a lot of different env types here. Might not be safe to just call everything not conda as pip. As a first pass we could just ignore the other types. Poetry and what not.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,124 @@
...
+
+        const selection = this.isCodeSpace
+            ? Common.install()
+            : await this.applicationShell.showErrorMessage(
+                  DataScience.pandasRequiredForViewing(),
+                  { modal: true },
+                  Common.install()
+              );
+
+        const command = `${this.packaging(kernel)} install pandas`;
```
IanMatthewHuff, line 85: If we are running in the kernel I would think we might want to use the pip / conda commands. Pretty sure they are safer:
https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-pip
Can have issues otherwise:
https://jakevdp.github.io/blog/2017/12/05/installing-python-packages-from-jupyter/

sadasant: Done!

src/webviews/extension-side/dataviewer/dataViewerDependencyService.node.ts
```diff
@@ -39,7 +39,11 @@ export class DataViewerDependencyService implements IDataViewerDependencyService
         @inject(IsCodeSpace) private isCodeSpace: boolean
     ) {}
 
-    public async checkAndInstallMissingDependencies(interpreter: PythonEnvironment): Promise<void> {
+    public async checkAndInstallMissingDependenciesOnKernel(): Promise<void> {
+        throw new Error('Not implemented');
+    }
+
+    public async checkAndInstallMissingDependenciesOnEnvironment(interpreter: PythonEnvironment): Promise<void> {
```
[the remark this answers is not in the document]
DonJayamanne: id still prefer to have the same method name and the same argument interpreter.
you can get the interpreter from the `kernel. connectionMetadata. interpreter`.

ie it is doable, and i feel that's much better than having two methods, one for local and one for remote.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,125 @@
...
+        const isConda = envType === EnvironmentType.Conda;
+        return isConda ? 'conda' : 'pip';
+    }
+
+    public async checkAndInstallMissingDependenciesOnEnvironment(): Promise<void> {
+        throw new Error('Not implemented');
+    }
+
+    public async checkAndInstallMissingDependenciesOnKernel(kernel: IKernel): Promise<void> {
+        const pandasVersion = await this.getVersionOfPandas(kernel);
```
DonJayamanne, line 50: tomorrow, if someone comes along and modifies the code to call this method instead of the ones with the argument, it's not wrong.
but will break things.
i think the problem is there's no way for someone to know which methods to call and when.
hence the suggestion to have a single method that takes the kernel argument.
as mentioned earlier from the kernel we can get the interpreter in one line.

we can get the kernel for any given resource as well, all you need is the nottebook or notebookuri. i

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
DonJayamanne: of you haven't fixed this, why not just stub the kernel.session methods. i think that's much easier.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: Does this still apply with `%pip` and `%conda`?

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
```diff
@@ -4,118 +4,136 @@
...
     });
+
     test('Throw exception if pandas is installed and version is < 0.20', async () => {
-        when(
-            pythonExecService.exec(deepEqual(['-c', 'import pandas;print(pandas.__version__)']), anything())
-        ).thenResolve({ stdout: '0.10.0' });
+        const version = '0.10.0';
 
-        const promise = dependencyService.checkAndInstallMissingDependencies(interpreter);
+        const stub = sinon.stub(helpers, 'executeSilently');
```
DonJayamanne, line 96: not sure if this works, however ifeel it might be easier to stub the `kernel. session` methods.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: I think that's what we're doing 🤔 `isConda ? 'conda' : 'pip'`

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
sadasant: If it isn't working I'm hallucinating 😅 I wouldn't discard that possibility though

src/webviews/extension-side/dataviewer/dataViewerDependencyService.node.ts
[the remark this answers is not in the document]
sadasant: I do like that idea! Thank you! I'll figure it out 👍

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: I do like that idea! Thank you! I'll figure it out 👍

src/webviews/extension-side/dataviewer/types.ts
```diff
@@ -130,5 +130,6 @@ export interface IJupyterVariableDataProviderFactory {
 
 export const IDataViewerDependencyService = Symbol('IDataViewerDependencyService');
 export interface IDataViewerDependencyService {
-    checkAndInstallMissingDependencies(interpreter: PythonEnvironment): Promise<void>;
+    checkAndInstallMissingDependenciesOnEnvironment(environment: PythonEnvironment): Promise<void>;
+    checkAndInstallMissingDependenciesOnKernel(kernel: IKernel): Promise<void>;
 }
```
DonJayamanne, line 135: id prefer to haber a single method, two feels like a work around which introduces debt. 
there's nothing preventing one from using the wrong method.

src/webviews/extension-side/dataviewer/dataViewerCommandRegistry.ts
```diff
@@ -99,7 +99,10 @@ export class DataViewerCommandRegistry implements IExtensionSingleActivationServ
                         this.debugService.activeDebugSession.configuration
                     );
 
-                    pythonEnv && (await this.dataViewerDependencyService.checkAndInstallMissingDependencies(pythonEnv));
+                    pythonEnv &&
+                        (await this.dataViewerDependencyService.checkAndInstallMissingDependenciesOnEnvironment(
+                            pythonEnv
+                        ));
                 }
```
DonJayamanne, line 106: wait, will this work with remote debugging.
i don't think so.
ie if im debugging a remote kernel, then will this fixed path work?
i think we'll need to get that working as well for the web. 
in there loader it didn't work, but now that we're enabling df viewer for web, 100% of web users will use remote kernels, hence this won't work when they're debugging. 

again, i think this is what @IanMatthewHuff mentioned.
we'd need to change the code as we should be sending the request using the debugger instead of interpreter or kernel (whether it's local or remote)

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
DonJayamanne: this test look wrong. 
previously we had tests for the method that took an interpreter, now we have lost that test.
i think we need this test in 4 different permutations (local vs remote, debug vs non debug)

DonJayamanne: Nope there's a new test file.

src/webviews/extension-side/dataviewer/dataViewerCommandRegistry.ts
sadasant: I don't think the command to debug a python file with the Python extension is working on the web. I can't get to make it work.

Besides that, this approach you are pointing out would be broken on remote today, is that the case?

I would like to separate the problem of the debugger into a separate PR, because the debugger in itself is fairly complex and it does not seem to be working working on the web anyway.

sadasant: Also, this file is using the pre-existing `dataViewerDependencyService`, not the new one

src/webviews/extension-side/dataviewer/dataViewerDependencyService.node.ts
```diff
@@ -39,7 +39,11 @@ export class DataViewerDependencyService implements IDataViewerDependencyService
         @inject(IsCodeSpace) private isCodeSpace: boolean
     ) {}
 
-    public async checkAndInstallMissingDependencies(interpreter: PythonEnvironment): Promise<void> {
+    public async checkAndInstallMissingDependenciesOnKernel(): Promise<void> {
```
rchiodo, line 42: Why do we have two versions? Can't they both use the kernel?

sadasant:
After speaking with @rchiodo about the debugger, I will include the debugger changes in this PR. I will close it for now, to make sure we have something cleaner to review with the debugger changes included.

PR #10549, opened 2022-06-23
Don't show modal install prompt when installing via "Install Python Extension" directly

### LINKED ISSUE #10548: Only show the "Do you want to install extension" modal dialog if triggered from the run button
Per feedback from Jim. The modal dialog to confirm extension install should only be shown when activated from the run command. If you directly pick the kernel command in the kernel picker, then don't show it.

IanMatthewHuff:
Fixes #10548


src/notebooks/controllers/installPythonControllerCommands.ts
```diff
@@ -161,4 +176,40 @@ export class InstallPythonControllerCommands implements IExtensionSingleActivati
...
         }
     }
+
+    // We don't always want to show our modal warning for installing the python extension
+    // this function will choose if this should be shown, and return true if the install should
+    // proceed and false otherwise
+    private async shouldInstallExtensionPrompt(): Promise<boolean> {
+        // We want to show the dialog if the active document is running, in this case, the command
+        // was triggered from the run button and we want to warn the user what we are doing
+        if (this.isActiveNotebookDocumentRunning()) {
```
IanMatthewHuff, line 186: Put some hefty comments here, as this is "cuter" than I generally like to get. Per feedback from Jim on the v-team this change means that you only get the model "Sure you want to install extension" dialog when you use the run button to trigger the command. If you just directly click it in the kernel picker you don't get the modal dialog.

src/notebooks/controllers/installPythonControllerCommands.ts
```diff
@@ -22,6 +29,8 @@ import { IVSCodeNotebookController } from './types';
 export class InstallPythonControllerCommands implements IExtensionSingleActivationService {
     private showInstallPythonExtensionContext: ContextKey;
     private showInstallPythonContext: ContextKey;
+    // WeakSet of executing cells, so they get cleaned up on document close without worrying
+    private executingCells: WeakSet<NotebookCell> = new WeakSet<NotebookCell>();
```
IanMatthewHuff, line 33: This feels like more work than it should be to track. But I believe this is currently what we need to do to get ExecutionState as it's not accessible directly from the Cell.

amunger (approved):

src/notebooks/controllers/installPythonControllerCommands.ts
```diff
@@ -161,4 +176,40 @@ export class InstallPythonControllerCommands implements IExtensionSingleActivati
...
+                sendTelemetryEvent(Telemetry.PythonExtensionNotInstalled, undefined, { action: 'download' });
+                return true;
+            } else {
+                // If they don't want to install, just bail out at this point
+                sendTelemetryEvent(Telemetry.PythonExtensionNotInstalled, undefined, { action: 'dismissed' });
+                return false;
+            }
+        }
+
+        // If the active notebook is not running, this command was triggered by direction clicking
```
amunger, line 203: "by direction clicking"?

rchiodo (approved):

src/notebooks/controllers/installPythonControllerCommands.ts
DonJayamanne: Can we not track the notebook instead of tracking the cells?
i.e. `WeakSet<NotebookDocument>`. Then further down we don't need to iterate through all the cells to check if the notebook is running

src/notebooks/controllers/installPythonControllerCommands.ts
```diff
@@ -161,4 +176,40 @@ export class InstallPythonControllerCommands implements IExtensionSingleActivati
...
+        // in this case, they clicked on "Install Python Extension" so no need for a modal
+        return true;
+    }
+    private isActiveNotebookDocumentRunning(): boolean {
+        if (window.activeNotebookEditor) {
+            return window.activeNotebookEditor.notebook.getCells().some((cell) => {
```
DonJayamanne, line 209: Can we not track the notebook instead of tracking the cells?
i.e. `WeakSet<NotebookDocument>`. Then we can get rid of the iterate through all the cells to check if the notebook is running

src/notebooks/controllers/installPythonControllerCommands.ts
DonJayamanne: This feels rather flaky, i'd prefer to pass an argument to be more explicit whether to show the modal or not.
Else if a user is running some notebook and then changes the tab (active noteboo), we'd end up showing this prompt.

src/notebooks/controllers/installPythonControllerCommands.ts
DonJayamanne: Nope won't work

src/notebooks/controllers/installPythonControllerCommands.ts
```diff
@@ -62,6 +74,18 @@ export class InstallPythonControllerCommands implements IExtensionSingleActivati
         );
     }
 
+    // Track if there are any cells currently executing or pending
+    private onDidChangeNotebookCellExecutionState(stateEvent: NotebookCellExecutionStateChangeEvent) {
+        if (
+            stateEvent.state === NotebookCellExecutionState.Pending ||
+            stateEvent.state === NotebookCellExecutionState.Executing
+        ) {
```
DonJayamanne, line 82: Might want to  check the notebook type and ensure we only look at Jupyter & IW notebooks.

DonJayamanne (approved):
Everythign is ok, however I think we should check the notebook type when tracking the execution states. Else we could end up tracking states of some notebook like Github issues notebook or the like.

src/notebooks/controllers/installPythonControllerCommands.ts
IanMatthewHuff: > "by direction clicking"? looks like a half edited comment here

Yeah apparently my brain died mid comment there :D. I'll fix it up.

PR #10550, opened 2022-06-23
standalone/intellisense

rebornix:
Re #10152 

The `intellisense` includes glue/integration code for hooking up lsp and jupyter server auto complete. I found that the component is well isolated and no one else depends on it thus I moved this into `standalone`.

The PR proves that the layering is correct, no files in `platform`/`notebooks`/`kernels`/`interactive-window`/`webviews` are modified after moving `intellisense` folder.


rchiodo:
I'm not sure about this change. All of the root level things could go under standalone? We could move kernels there too. I thought the high level directories where the high level components? Maybe I'm arguing for removing the 'standalone' folder then. As every subdirectory under there could be a root level folder.

rebornix:
> I'm not sure about this change. All of the root level things could go under standalone? We could move kernels there too. I thought the high level directories where the high level components

@rchiodo if we try to move `kernels` into `standalone`, then `notebooks` and `interactive-window` would depend on `standalone`, and at the same time, `standalone` depends on `notebooks`, which is a circular dependency. While where we move `intellisense` wouldn't introduce any layer dependencies change.

If there is another component that depends on `intellisense`, then we should keep it as a top component. Currently it seems we don't have any, this the major difference between `intellisense` with the rest of the top components and it's the Yes/No question I will ask myself where to contribute a new component. Does this make sense to you?

DonJayamanne:
I'm with Rich on this, personally don't see much value for standalone anymore.
Feels rather unnecessary to move yet again into another `standalone` directory, when these components are already in their own root level directories.

rchiodo:
So you're saying standalone components can't depend upon each other. That's what makes them different? So Kernels isn't standalone because other standalone components depend upon it.

Sure that's fine. Maybe we can put that in an eslint rule too somehow.

rebornix:
> I'm with Rich on this, personally don't see much value for standalone anymore.

@DonJayamanne I'm interested in alternatives from your perspective. if we don't have a `standalone`, what's the recommended approach for managing folder structures for features like `api`, `export-command`, etc.

DonJayamanne:
> ur perspective. if we don't have a standalone, what's the

Can we not have an eslint rule for these, feels better than moving folders up a level.

rebornix:
> Can we not have an eslint rule for these

what eslint rule are you referring to here?

rebornix:
As we discussed in today's stand up, we would document the rules of where to contribute

PR #10551, opened 2022-06-23
changed npm down to 8.1.0 in CONTRIBUTING.md

Yoyokrazy:
Fixes #

-   [x] ~Appropriate comments and documentation strings in the code.~
-   [x] ~Has telemetry for feature-requests.~

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@
 ### Prerequisites
 
 1. [Node.js](https://nodejs.org/) 16.13.2
-2. [npm](https://www.npmjs.com/) 8.3
+2. [npm](https://www.npmjs.com/) 8.1.0
 3. [Python](https://www.python.org/) 3.6 or later
 4. Windows, macOS, or Linux
 5. [Visual Studio Code](https://code.visualstudio.com/)


Yoyokrazy:
npm v8.3 when running install would change the package-lock file and was updating something, reverting to 8.1 seemed to keep everything in sync

rchiodo:
The change it creates looks correct to me. It's updating url-parse to 1.5.10 which it needs to be. I think instead you should submit the package-lock change.

DonJayamanne:
@rchiodo i think the change is valid, npm version 8.3 was shipped with node 16.14, but we use node 16.13 (that's what VS Code uses & I updated this recently).
& npm version that ships with node 16.13 (which is what we use) ends up shipping 8.1.
Hence this PR is correct.

DonJayamanne (approved):
@Yoyokrazy thanks & please add a news entry to thank your self for this contribution.

IanMatthewHuff:
The Node / NPM version stuff is a pain at times. AFAIK Node versions can continue to update "supported" versions of npm. So you can continue to update versions of NPM with the same supported Node version. But I guess just going with "ships with" is the easiest thing to specify.

rchiodo:
This is the total of the package lock change:

```
diff --git a/package-lock.json b/package-lock.json
index c511e148c..b14ed6bed 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -26780,7 +26780,7 @@
                 "minimist": "~1.2.0",
                 "moment": "^2.24.0",
                 "path-browserify": "^1.0.0",
-                "url-parse": "1.5.8"
+                "url-parse": "^1.5.10"
             }
         },
```
Not sure why 8.1 doesn't know to make this change, but that's correct. Our minimum version for url-parse is now 1.5.10

PR #10554, opened 2022-06-23
Copy nbextension again if we have new files

DonJayamanne:
Perf improvements to copying nbexensions

### DIFF
--- a/src/kernels/ipywidgets/localIPyWidgetScriptManager.node.ts
+++ b/src/kernels/ipywidgets/localIPyWidgetScriptManager.node.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 import * as path from '../../platform/vscode-path/path';
-import { Uri } from 'vscode';
+import { Memento, Uri } from 'vscode';
 import { IFileSystemNode } from '../../platform/common/platform/types.node';
 import { IExtensionContext } from '../../platform/common/types';
 import { StopWatch } from '../../platform/common/utils/stopWatch';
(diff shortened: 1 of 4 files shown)


src/kernels/ipywidgets/localIPyWidgetScriptManager.node.ts
```diff
@@ -95,4 +67,59 @@ export class LocalIPyWidgetScriptManager extends BaseIPyWidgetScriptManager impl
...
+            const filesPromise = this.fs.searchLocal('**/*.*', nbextensionsPath).catch((ex) => {
+                // Handle errors an ensure we don't crash the rest of the code just because we couldn't search for files.
+                traceError(`Failed to search for files in ${nbextensionsPath}`, ex);
+                return [];
+            });
+            // If we have previously copied the files, and the file count is the same, then don't overwrite them.
+            const previouslyCopiedFileCount = this.globalMemento.get<number>(this.getMementoKey(), 0);
+            if (previouslyCopiedFileCount > 0) {
+                const files = await filesPromise;
+                if (files.length === previouslyCopiedFileCount) {
```
rchiodo, line 90: I guess the assumption is that files aren't updated? Or they're only updated by restarting the kernel? Was thinking you could upgrade a package and have a newer version of some file.

rchiodo (approved):

src/kernels/ipywidgets/localIPyWidgetScriptManager.node.ts
DonJayamanne: yes, it just felt a little too slow to always copy the files.
in hindsight, this change might be risky, with the example you mentioned

DonJayamanne:
Need to find a more efficient way to address this

PR #10557, opened 2022-06-23
Refactor how variables and DF info is fetched

### LINKED ISSUE #10516: NameError: name '_VSCODE_getVariableInfo' is not defined
I was debugging...

When I chose "`view value in data viewer`" for a variable at a checkpoint for the first time, it worked fine as normal. 

Then  I clicked `F5`, and the programme **stopped at the same checkpoint**. 

But when I chose "`view value in data viewer`" for a variable again, it didn't work and the log was as follows

```shell
error 11:44:25.215: Error: Traceback (most recent call last):
  File "<string>", line 1, in <module>
NameError: name '_VSCODE_getVariableInfo' is not defined

    at o.handleErrorResponse (vscode-file://vscode-app/d:/vscode/Microsoft%20VS%20Code/resources/app/out/vs/workbench/workbench.desktop.main.js:1041:45806)
    at vscode-file://vscode-app/d:/vscode/Microsoft%20VS%20Code/resources/app/out/vs/workbench/workbench.desktop.main.js:1041:45245
warn 11:44:25.215: DataScience Error Error: Traceback (most recent call last):
  File "<string>", line 1, in <module>
NameError: name '_VSCODE_getVariableInfo' is not defined

    at o.handleErrorResponse (vscode-file://vscode-app/d:/vscode/Microsoft%20VS%20Code/resources/app/out/vs/workbench/workbench.desktop.main.js:1041:45806)
    at vscode-file://vscode-app/d:/vscode/Microsoft%20VS%20Code/resources/app/out/vs/workbench/workbench.desktop.main.js:1041:45245
```

DonJayamanne:
Fixes #10516

### DIFF
--- a/pythonFiles/vscode_datascience_helpers/getVariableInfo/vscodeGetVariableInfo.py
+++ b/pythonFiles/vscode_datascience_helpers/getVariableInfo/vscodeGetVariableInfo.py
@@ -82,9 +82,8 @@ def _VSCODE_getVariable(what_to_get, *args):
         _VSCODE_getVariableTypes(*args)
 
     try:
-        import sys
-        del sys.modules['_VSCODE_json']
-        del sys.modules['_VSCODE_builtins']
+        del _VSCODE_json
+        del _VSCODE_builtins
     except:
         pass
 
--- a/pythonFiles/vscode_datascience_helpers/getVariableInfo/vscodeGetVariableInfo.py
+++ b/pythonFiles/vscode_datascience_helpers/getVariableInfo/vscodeGetVariableInfo.py
@@ -1,73 +1,90 @@
-# Query Jupyter server for the info about a dataframe
-import json as _VSCODE_json
-import builtins as _VSCODE_builtins
+def _VSCODE_getVariable(what_to_get, *args):
 
-# Function to do our work. It will return the object
-def _VSCODE_getVariableInfo(var):
-    # Start out without the information
-    result = {}
-    result["shape"] = ""
-    result["count"] = 0
-    result["type"] = ""
+    # Query Jupyter server for the info about a dataframe
+    import json as _VSCODE_json
+    import builtins as _VSCODE_builtins
 
-    typeName = None
+    # Function to do our work. It will return the object
+    def _VSCODE_getVariableInfo(var):
+        # Start out without the information
+        result = {}
+        result["shape"] = ""
+        result["count"] = 0
+        result["type"] = ""
+
+        typeName = None
+        try:
+            vartype = type(var)
+            if hasattr(vartype, "__name__"):
+                result["type"] = typeName = vartype.__name__
+        except TypeError:
+            pass
+
+        # Find shape and count if available
+        if hasattr(var, "shape"):
+            try:
+                # Get a bit more restrictive with exactly what we want to count as a shape, since anything can define it
+                if (
+                    isinstance(var.shape, tuple)
+                    or typeName is not None
+                    and typeName == "EagerTensor"
+                ):
+                    _VSCODE_shapeStr = str(var.shape)
+                    if (
+                        len(_VSCODE_shapeStr) >= 3
+                        and _VSCODE_shapeStr[0] == "("
+                        and _VSCODE_shapeStr[-1] == ")"
+                        and "," in _VSCODE_shapeStr
+                    ):
+                        result["shape"] = _VSCODE_shapeStr
+                    elif _VSCODE_shapeStr.startswith("torch.Size(["):
+                        result["shape"] = "(" + _VSCODE_shapeStr[12:-2] + ")"
+                    del _VSCODE_shapeStr
+            except TypeError:
+                pass
+
+        if hasattr(var, "__len__"):
+            try:
+                result["count"] = len(var)
+            except TypeError:
+                pass
+
+        # return our json object as a string
+        _VSCODE_builtins.print(_VSCODE_json.dumps(result))
+
+
+    def _VSCODE_getVariableProperties(var, listOfAttributes):
+        result = {
+            attr: repr(getattr(var, attr))
+            for attr in listOfAttributes
+            if hasattr(var, attr)
+        }
+        _VSCODE_builtins.print(_VSCODE_json.dumps(result))
+
+
+    def _VSCODE_getVariableTypes(varnames):
+        # Map with key: varname and value: vartype
+        result = {}
+        for name in varnames:
+            try:
+                vartype = type(globals()[name])
+                if hasattr(vartype, "__name__"):
+                    result[name] = vartype.__name__
+            except TypeError:
+                pass
+        _VSCODE_builtins.print(_VSCODE_json.dumps(result))
+
+    if what_to_get == "properties":
+        _VSCODE_getVariableProperties(*args)
+    elif what_to_get == "info":
+        _VSCODE_getVariableInfo(*args)
+    else:
+        _VSCODE_getVariableTypes(*args)
+
     try:
-        vartype = type(var)
-        if hasattr(vartype, "__name__"):
-            result["type"] = typeName = vartype.__name__
-    except TypeError:
+        import sys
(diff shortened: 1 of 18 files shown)


src/kernels/variables/debuggerVariables.ts
```diff
@@ -316,44 +321,14 @@ export class DebuggerVariables
         }
         throw Error('Debugger is not active, cannot evaluate.');
     }
-
-    private async importDataFrameScripts(): Promise<void> {
-        try {
-            // Run our dataframe scripts only once per session because they're slow
-            const key = this.debugService.activeDebugSession?.id;
```
DonJayamanne, line 323: This same code was repeated in two places, now one place

src/kernels/variables/debuggerVariables.ts
```diff
@@ -316,44 +321,14 @@ export class DebuggerVariables
...
-        // See if we imported or not into the kernel our special function
-        await this.importGetVariableInfoScripts();
-
         // Then eval calling the variable info function with our target variable
+        const code = await this.varScriptGenerator.generateCodeToGetVariableInfo({
+            isDebugging: false,
+            variableName: variable.name
+        });
         const results = await this.evaluate(
-            `${GetVariableInfo.VariableInfoFunc}(${variable.name})`,
```
DonJayamanne, line 356: The problem here is when debugging we have an await and its possible after we first import the scripts the stack frame has changed, hence the variables will be lost (not in scope).

This fixes that by ensureing that the function call always imports the necessary scripts (previously it was done seprately)

src/kernels/variables/debuggerVariables.ts
```diff
@@ -9,9 +9,13 @@ import { DebugAdapterTracker, Disposable, Event, EventEmitter } from 'vscode';
 import { DebugProtocol } from 'vscode-debugprotocol';
 import { Identifiers } from '../../platform/common/constants';
 import { IDebugService, IVSCodeNotebook } from '../../platform/common/application/types';
-import { DataFrameLoading, GetVariableInfo } from '../../platform/common/scriptConstants';
 import { traceError, traceVerbose } from '../../platform/logging';
-import { IConfigurationService, IExtensionContext, Resource } from '../../platform/common/types';
+import {
```
DonJayamanne, line 13: Note: I haven't been able to repro the issue yet, however I can see where it can fail.
Will try to repro this (having trouble debugging IW right now)

src/kernels/variables/debuggerVariables.ts
rchiodo: THe original code worked because of it loading a module but also had the added benefit of it not needing to be loaded more than once.

Could we change the new code to create a module to accomplish the same thing? So we don't have to load it more than once?

rchiodo: Meaning we wouldn't have to recreate the functions on every stack frame.

The side benefit is we would have less extra cruft in the debugger variable list too.

src/kernels/variables/debuggerVariables.ts
rchiodo: There's a repro in the bug. You don't need to use IW to repro. Just debug a plain python file.

src/kernels/variables/debuggerVariables.ts
DonJayamanne: I do'nt see any harm in re-creating (re-loading this everytime).
Also we cannot go with a module as that won't work in web/remote.

I'll update the code to delete the imported namespaces & the functions.

Baically change to as follows:
```python
import xyz
def __vsc_getDataFrameInfo():
....

print(results)

delete sys.modules[xyz]
del __vsc_getDataFrameInfo
```

This way we clean up all of the imports and also the function definitions we create.

DonJayamanne: I don't see that as being an issue, its not a large chunk of code that could slow things down, only runs when the user has stopped at a stack frame & running that code is super fast

pythonFiles/vscode_datascience_helpers/getVariableInfo/vscodeGetVariableInfo.py
```diff
@@ -1,73 +1,90 @@
-# Query Jupyter server for the info about a dataframe
-import json as _VSCODE_json
-import builtins as _VSCODE_builtins
+def _VSCODE_getVariable(what_to_get, *args):
```
DonJayamanne, line 1: @rchiodo I've modified the code so that everything is under a function now, all we need to do is delete the imported modules & delete this function that we declare (when cleaning up).
If we were to import a module, we'd still need to delete the imported module, 

Now we should not end up with tonnes of function definitinos as everything is local to this one function, hence no need to clean those individual functions (as its all local & goes out of scope)

src/kernels/variables/debuggerVariables.ts
DonJayamanne: aaah got it, thanks

src/kernels/variables/debuggerVariables.ts
DonJayamanne: Fixed

src/test/datascience/variableView/variableView.vscode.test.ts
```diff
@@ -100,6 +103,21 @@ suite('DataScience - VariableView', function () {
...
         ];
         await waitForVariablesToMatch(expectedVariables, variableView);
+
+        // Verify we don't have any new variables apart from test, test2, os & sys
+        const kernel = kernelProvider.get(cell.notebook.uri)!;
+        const outputs = await kernel.executeHidden('%who_ls');
+        // Sample output is `["test", "test2", "os", "sys"]`
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const vars = ((outputs[0].data as any)['text/plain'] as string)
+            .trim()
```
DonJayamanne, line 113: Added a new test to ensure we don't leave any messy variables behind

DonJayamanne: Filed a bug to remove os & sys (stuff that we add when starting kernels & the like)

src/kernels/variables/pythonVariableRequester.ts
```diff
@@ -57,32 +54,26 @@ export async function safeExecuteSilently(
...
-            kernel,
-            `import builtins\nbuiltins.print(${DataFrameLoading.DataFrameInfoFunc}(${expression}))`,
-            {
-                traceErrors: true,
-                traceErrorsMessage: 'Failure in execute_request for getDataFrameInfo',
-                telemetryName: Telemetry.PythonVariableFetchingCodeFailure
-            }
-        );
+        const code = await this.dfScriptGenerator.generateCodeToGetDataFrameInfo({
+            isDebugging: false,
```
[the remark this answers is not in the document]
DonJayamanne: Fixed, 
had to refactor as the debugger tests weren't passing

src/kernels/variables/debuggerVariables.ts
rchiodo: Oh I guess it is possible. Maybe exec wouldn't work in the debugger though

https://stackoverflow.com/questions/5362771/how-to-load-a-module-from-code-in-a-string

rchiodo: This would have the benefit of not really needing the cleanup code too.

rchiodo: Yeah but a module wouldn't need to reload on EVERY variable request. It'd be one and done.

rchiodo: I think we could leave the module there. We did before.

DonJayamanne: > Could have up to 30 requests for this (assuming variable view was showing 30 variables at a time)

Will check.
FYI - we only have this when debugging, e.g. in regular execution, we send everything as a single request.
Will look into the execution times now.

PR #10558, opened 2022-06-24
Change pylanceHandlesNotebooks default to true

debonte:
Fixes https://github.com/microsoft/pylance-release/issues/2963

The pylanceHandlesNotebooks experiment is finished and was successful, so changing the setting to be on by default.

-   [X] Pull request represents a single change (i.e. not fixing disparate/unrelated things in a single PR).
-   [X] Title summarizes what is changing.
-   [X] ~Has a [news entry](https://github.com/Microsoft/vscode-jupyter/tree/main/news) file (remember to thank yourself!).~
-   [X] ~Appropriate comments and documentation strings in the code.~
-   [X] ~Has sufficient logging.~
-   [X] ~Has telemetry for feature-requests.~
-   [X] ~Unit tests & system/integration tests are added/updated.~
-   [X] ~[Test plan](https://github.com/Microsoft/vscode-jupyter/blob/main/.github/test_plan.md) is updated as appropriate.~
-   [X] ~[`package-lock.json`](https://github.com/Microsoft/vscode-jupyter/blob/main/package-lock.json) has been regenerated by running `npm install` (if dependencies have changed).~

### DIFF
--- a/package.json
+++ b/package.json
@@ -1970,12 +1970,9 @@
                 },
                 "jupyter.pylanceHandlesNotebooks": {
                     "type": "boolean",
-                    "default": false,
+                    "default": true,
                     "description": "Determines if pylance's experimental notebook support is used or not.",
-                    "scope": "machine",
-                    "tags": [
-                        "experimental"
-                    ]
+                    "scope": "machine"
                 },
                 "jupyter.pythonCompletionTriggerCharacters": {
                     "type": "string",
--- a/package.json
+++ b/package.json
@@ -1971,7 +1971,7 @@
                 "jupyter.pylanceHandlesNotebooks": {
                     "type": "boolean",
                     "default": true,
-                    "description": "Determines if pylance's experimental notebook support is used or not.",
+                    "description": "Determines if Pylance manages notebook concat doc creation.",
                     "scope": "machine"
                 },
                 "jupyter.pythonCompletionTriggerCharacters": {
--- a/src/platform/common/configSettings.ts
+++ b/src/platform/common/configSettings.ts
@@ -94,7 +94,7 @@ export class JupyterSettings implements IWatchableJupyterSettings {
     public verboseLogging: boolean = false;
     public showVariableViewWhenDebugging: boolean = true;
     public newCellOnRunLast: boolean = true;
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1970,12 +1970,9 @@
                 },
                 "jupyter.pylanceHandlesNotebooks": {
                     "type": "boolean",
-                    "default": false,
+                    "default": true,
                     "description": "Determines if pylance's experimental notebook support is used or not.",
```
DonJayamanne, line 1974: should we change this description?
ie is this feature still experimental ?

debonte: Thanks. I didn't notice the word experimental in the description. @heejaechang, is the new string an accurate description of the setting?

debonte: Can/should the setting be hidden so it doesn't show up in the settings UI?

package.json
```diff
@@ -1970,12 +1970,9 @@
                 },
                 "jupyter.pylanceHandlesNotebooks": {
                     "type": "boolean",
-                    "default": false,
-                    "description": "Determines if pylance's experimental notebook support is used or not.",
-                    "scope": "machine",
-                    "tags": [
-                        "experimental"
-                    ]
+                    "default": true,
```
IanMatthewHuff, line 1973: Just to check (I've not followed this work as much) is the pylanceHandlesNotebooks = false case still valid going forward? I.E. there is still a reason that a user might opt out of this?

debonte: I don't believe so.

Is your question, "Should we remove the pylanceHandlesNotebooks = false code path, rather than changing the default"? That's a question for @heejaechang.

FYI, Pylance has three approaches to notebooks at the moment:
1. Defaults -- Jupyter is in charge of concat doc
2. pylanceHandlesNotebooks = true -- Pylance is in charge of concat doc
3. pylanceLspNotebooksEnabled = true -- No concat doc. Separate document for each cell.

Option 3 is an active experiment. I was imagining that we'd wait until that experiment finished to make a decision and come up with a deprecation/removal plan for the other options.

IanMatthewHuff: Thanks for the explanation. Seems like this is fine for now as long as we handle the overall depreciation plan at some point.

rchiodo (approved):

PR #10562, opened 2022-06-24
CRLF vs LF could be a problem for us

sadasant:
Now that several of us are using OSX, CRLF vs LF could be a problem for us. This PR ensures that LF is always used on Git, by specifying it in the `.gitattributes`, and preemptively replaces a bunch of CRLFs for LFs.

Changes look minimal if one disables whitespaces: https://github.com/microsoft/vscode-jupyter/pull/10562/files?diff=unified&w=1 😁

Feedback always appreciated!


DonJayamanne:
I've been using wsl, mac, Linux and windows and never had any issues. perhaps i have set-up my git settings in a way that it works as expected... 
i believe ian is on a mac and rich has used wsl as well.

basically I'm not convinced we need such a change. 
also i can see a change to a non text file, ttf which i believe happens to be a binary file.
that could be risky. 

ill let @rchiodo @IanMatthewHuff  and others review this PR.
however I'm not convinced we need this.

sadasant:
@DonJayamanne I'm on Mac 🤔

DonJayamanne (changes requested):
> 'm on Mac 🤔

Same here.
I'd suggest you double check your changes.
Ian and I are on Mac as well. I'd suggest you check your settings, as we haven't had to change anything in the repo

DonJayamanne:
I just edited an existing file that you seem to have trouble with in another PR of yours
Here's the PR https://github.com/microsoft/vscode-jupyter/pull/10563 i submitted with a single line change & for some reason I'm not running into the LF issues you  are.

IanMatthewHuff:
@DonJayamanne @sadasant Yeah. I don't have this issue. Researching now. I honestly don't recall making a specific change for this, but it has been a while since I set up this machine. Daniel what do you currently have in `git config -l` ?

sadasant:
@IanMatthewHuff I have `core.eol=lf` in my Git config. Could that be the problem?

sadasant:
I have forfeited, I will update my Git config. 🏳

PR #10572, opened 2022-06-26
Remove duplicate arguments

DonJayamanne:
Found some duplicate & unnecessary arguments.

### DIFF
--- a/src/kernels/execution/kernelExecution.ts
+++ b/src/kernels/execution/kernelExecution.ts
@@ -3,7 +3,7 @@
 
 'use strict';
 
-import { EventEmitter, NotebookCell, NotebookCellKind, NotebookController, NotebookDocument, workspace } from 'vscode';
+import { EventEmitter, NotebookCell, NotebookCellKind, NotebookDocument, workspace } from 'vscode';
 import { CellExecutionFactory } from './cellExecution';
 import { CellExecutionQueue } from './cellExecutionQueue';
 import { KernelMessage } from '@jupyterlab/services';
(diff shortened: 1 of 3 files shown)


src/kernels/execution/kernelExecution.ts
```diff
@@ -44,22 +43,20 @@ export class KernelExecution implements IDisposable {
     constructor(
         private readonly kernel: IKernel,
         appShell: IApplicationShell,
-        readonly metadata: Readonly<KernelConnectionMetadata>,
         private readonly interruptTimeout: number,
-        controller: NotebookController,
```
DonJayamanne, line 49: `controller` can be accessed from `kernel` argument, hence we don't need to pass this as a separate argument.

src/kernels/execution/kernelExecution.ts
```diff
@@ -44,22 +43,20 @@ export class KernelExecution implements IDisposable {
     constructor(
         private readonly kernel: IKernel,
         appShell: IApplicationShell,
-        readonly metadata: Readonly<KernelConnectionMetadata>,
```
DonJayamanne, line 47: `metadata` can be accessed from `kernel` argument, hence we don't need to pass this as a separate argument.

src/notebooks/controllers/notebookIPyWidgetCoordinator.ts
```diff
@@ -36,15 +36,15 @@ class NotebookCommunication implements IWebviewCommunication, IDisposable {
         this.changeController(controller);
     }
     public changeController(controller: IVSCodeNotebookController) {
-        if (this.controller === controller) {
+        if (this.controller.id === controller.id) {
```
DonJayamanne, line 39: we have an id to uniquely identify controllers, hence using that when comparing controllers.

IanMatthewHuff (approved):

PR #10573, opened 2022-06-26
Ensure kernel points to live kernel connection when starting remote kernels

### LINKED ISSUE #9865: Active/Live Remote Kernel instance is not selected after it is started
While testing https://github.com/microsoft/vscode-jupyter/issues/9768, I was quite confused by the kernel picker UI.

1. I select "Python 3" unstarted (the only kernel at the start)
2. I run a cell
3. The kernel picker says the unstarted kernel is the currently selected kernel
4. There is a new entry that is the newly started kernel which is truly the currently selected kernel

Perhaps there is a solution in the proxy/lazy kernel work.

DonJayamanne:
Fixes #9865


IanMatthewHuff:
Looks like there is a new failed test: 3rd Party Kernel Service API Start Kernel

src/kernels/notebookControllerWrapper.ts
```diff
@@ -0,0 +1,152 @@
...
+    Uri
+} from 'vscode';
+import { disposeAllDisposables } from '../platform/common/helpers';
+import { IDisposable } from '../platform/common/types';
+
+/**
+ * Returns a class thats identical to NotebookController.
+ * Basically similar to a `Proxy` class, but with the ability to determine whether the instance of this class is a proxy or not.
+ */
+export class NotebookControllerWrapper implements NotebookController {
```
rchiodo, line 23: I'm not sure what this is used for? Why do we need a NotebookController proxy?

rchiodo: Can you explain that in the comment at the top of class?

src/interactive-window/interactiveWindow.ts
```diff
@@ -197,14 +198,12 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             };
             // When connecting, we need to update the sys info message
             this.updateSysInfoMessage(this.getSysInfoMessage(metadata, SysInfoReason.Start), false, sysInfoCell);
-            const kernel = await KernelConnector.connectToKernel(
-                controller,
-                metadata,
-                this.serviceContainer,
-                { resource: this.owner, notebook: this.notebookDocument },
+            const vscController = Array.from(this.controllerRegistrations.values).find(
```
amunger, line 201: isn't `this.controllerRegistrations.values` already an array? (making `Array.from` unnecessary)

src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -160,6 +175,68 @@ export class ControllerRegistration implements IControllerRegistration {
...
+                    item.connection.kind === 'connectToLiveRemoteKernel' &&
+                    item.connection.id === kernel.session?.kernel?.id
+            );
+            if (liveController) {
+                this.kernelProvider.updateKernel(kernel, liveController.connection, liveController.controller);
+
+                const notebookEditor = this.notebook.notebookEditors.find(
+                    (item) => item.notebook.uri.toString() === kernel.uri.toString()
+                );
+                await this.commandManager.executeCommand('notebook.selectKernel', {
```
amunger, line 226: I'm confused about this - I thought the new proxy classes were being updated so that we wouldn't actually have to switch to a new kernel.

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -443,27 +485,31 @@ export class VSCodeNotebookController implements Disposable, IVSCodeNotebookCont
...
+        let controller: NotebookController = this.controller;
+        let oldController: NotebookController = this.controller;
         let kernelStarted = false;
         try {
-            kernel = await this.connectToKernel(doc, new DisplayOptions(false));
-            kernelStarted = true;
+            kernel = await this.connectToKernel({ notebook }, new DisplayOptions(false));
             // If the controller changed, then ensure to create a new cell execution object.
-            if (kernel && kernel.controller.id !== controller.id) {
+            if (kernel && kernel.controller.id !== oldController.id) {
```
amunger, line 502: it seems like this could just compare to `controller.id` and drop the `oldController` variable

src/interactive-window/interactiveWindow.ts
DonJayamanne: Oh yes, at one point this was pointing to some dictionary, i think i forgot this changed.

src/notebooks/controllers/controllerRegistration.ts
DonJayamanne: We don't have that API yet, the plan is to work on that API with core soon, after we have this PR

src/kernels/ipywidgets/commonMessageCoordinator.ts
```diff
@@ -106,10 +115,45 @@ export class CommonMessageCoordinator {
                 (m) => {
                     traceInfoIfCI(`${ConsoleForegroundColors.Green}Widget Coordinator received ${m.type}`);
                     this.onMessage(m.type, m.payload);
-
+                    const kernel = this.kernelProvider.get(this.document.uri);
                     // Special case the WidgetManager loaded message. It means we're ready
                     // to use a kernel. (IPyWidget Dispatcher uses this too)
                     if (m.type === IPyWidgetMessages.IPyWidgets_Ready) {
+                        if (kernel?.kernelConnectionMetadata.kind === 'startUsingRemoteKernelSpec') {
```
sadasant, line 122: Should `startUsingRemoteKernelSpec` be elsewhere? Like in a `constants` kind of file?

src/kernels/ipywidgets/commonMessageCoordinator.ts
```diff
@@ -106,10 +115,45 @@ export class CommonMessageCoordinator {
...
+                        // We could solve this issue easily by not executing cells, until the webview is ready,
+                        // however that would have significant performance implications.
+                        // Hence for ipywidgets, once the webview has completely been initialized, we can attempt to re-render the widgets.
+                        const nbEditor = this.serviceContainer
+                            .get<IVSCodeNotebook>(IVSCodeNotebook)
+                            .notebookEditors.find((item) => item.notebook === this.document);
+                        if (nbEditor) {
+                            traceInfoIfCI('Re-rendering widgets');
+                            notebooks
+                                .createRendererMessaging('jupyter-ipywidget-renderer')
```
sadasant, line 153: Should `jupyter-ipywidget-renderer` be elsewhere? Like in a `constants` kind of file?

src/kernels/ipywidgets/commonMessageCoordinator.ts
rchiodo: No it's an enum. The compiler will check if you type it in wrong. Here's an example of me typing it in wrong:

image

src/kernels/ipywidgets/commonMessageCoordinator.ts
```diff
@@ -106,10 +115,45 @@ export class CommonMessageCoordinator {
...
+                        // should be able to render them now.
+                        // E.g assume we're dealing with remote kernel specs,
+                        // Then we start a kernel, next we create a controller that points to the live kernel & change the controller.
+                        // What happens now is the, webview is re-loaded (as theres a change in the controllers).
+                        // However if the user were to run a cell, then the output would get displayed and it could end up,
+                        // being displayed in the old webview & when the webview is re-loaded, it would be displayed in the new webview.
+                        // However its possble the kernel is not yet ready at that point in time.
+                        // We could solve this issue easily by not executing cells, until the webview is ready,
+                        // however that would have significant performance implications.
+                        // Hence for ipywidgets, once the webview has completely been initialized, we can attempt to re-render the widgets.
```
sadasant, line 146: I love this!! Thank you 🙏

src/kernels/ipywidgets/commonMessageCoordinator.ts
```diff
@@ -148,6 +192,14 @@ export class CommonMessageCoordinator {
         this.getIPyWidgetScriptSource().onMessage(message, payload);
     }
 
+    private sendPendingWebViewMessages() {
+        if (!this.webview || !this.webview.isReady || !this.readyMessageReceived) {
+            return;
+        }
+        while (this.queuedMessages.length) {
```
sadasant, line 199: `while` will lock the thread. Could there be many queued messages? It might be better to use `queuedMessages.forEach`

src/kernels/kernelProvider.base.ts
```diff
@@ -53,6 +55,27 @@ export abstract class BaseKernelProvider implements IKernelProvider {
...
+        if (NotebookControllerWrapper.isWrapped(controllerWrapper)) {
+            controllerWrapper.update(controller);
+        } else {
+            throw new Error('IKernel instantiated without passing a NotebookControllerWrapper');
+        }
+        const metadataWrapper = kernel.kernelConnectionMetadata;
+        if (KernelConnectionMetadataProxy.isWrapped(metadataWrapper)) {
+            metadataWrapper.update(metadata);
+        } else {
+            throw new Error('IKernel instantiated without passing a KernelConnectionMetadataWrapper');
```
sadasant, line 69: Should this error message be localized?

amunger (approved):

src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -179,6 +197,75 @@ export class ControllerRegistration implements IControllerRegistration {
...
+                        path: path.basename(kernel.uri.path),
+                        type: 'notebook',
+                        kernel: kernel.session.kernel.model
+                    },
+                    notebook: {
+                        path: path.basename(kernel.uri.path)
+                    },
+                    name: kernel.session.kernel.name,
+                    numberOfConnections: 1,
+                    lastActivityTime: new Date(Date.now()),
```
sadasant, line 227:
```suggestion
                    lastActivityTime: new Date(),
```

src/kernels/ipywidgets/commonMessageCoordinator.ts
DonJayamanne: both will lock the thread.
only async code did not block threads in v8.
l running such code in a tight loop is ok, we won't have millions, it's a very very small data set and execution, which will not be cpu intensive

src/kernels/kernelProvider.base.ts
DonJayamanne: no, it's not meant for the user, these are not meant to happen. it's only for logging

src/kernels/kernelProvider.base.ts
```diff
@@ -53,6 +55,27 @@ export abstract class BaseKernelProvider implements IKernelProvider {
         disposables.push(this._onDidStartKernel);
         disposables.push(this._onDidCreateKernel);
     }
+    updateKernel(kernel: IKernel, metadata: KernelConnectionMetadata, controller: NotebookController): void {
+        const controllerWrapper = kernel.controller;
+        if (NotebookControllerWrapper.isWrapped(controllerWrapper)) {
+            controllerWrapper.update(controller);
+        } else {
+            throw new Error('IKernel instantiated without passing a NotebookControllerWrapper');
```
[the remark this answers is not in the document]
DonJayamanne: no, it's not meant for the user, these are not meant to happen. it's only for logging

src/kernels/kernelProvider.web.ts
```diff
@@ -50,12 +52,12 @@ export class KernelProvider extends BaseKernelProvider {
         const kernel = new Kernel(
             uri,
             resourceUri,
-            options.metadata,
+            KernelConnectionMetadataProxy.wrap(options.metadata),
```
rchiodo, line 55: Would it make sense to do this inside of the Kernel class?

src/kernels/types.ts
```diff
@@ -237,6 +237,10 @@ export interface IKernelProvider extends IAsyncDisposable {
      * WARNING: If called with different options for same Notebook, old kernel associated with the Uri will be disposed.
      */
     getOrCreate(uri: Uri, options: KernelOptions): IKernel;
+    /**
```
rchiodo, line 240: Feels like this would make more sense to be on the IKernel itself. Why does the provider even have to know about the wrappers?

src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -179,6 +197,75 @@ export class ControllerRegistration implements IControllerRegistration {
...
+                    id: kernel.session.kernel.id
+                },
+                serverId: kernel.kernelConnectionMetadata.serverId,
+                interpreter: kernel.kernelConnectionMetadata.interpreter
+            };
+            if (!this.registeredControllers.has(this.getControllerId(liveKernelConnection, 'jupyter-notebook'))) {
+                this.add(liveKernelConnection, [JupyterNotebookView, InteractiveWindowView]);
+            }
+            const liveController = Array.from(this.registeredControllers.values()).find(
```
rchiodo, line 236: Array.from shouldn't be necessary here.

src/standalone/api/kernelApi.ts
```diff
@@ -221,13 +221,15 @@ class JupyterKernelService implements IExportedKernelService {
...
             // By not forcing the cast, we ensure the types are compatible.
             // All we're doing is ensuring the readonly version of one type is compatible with the other.
-            // Also, we must return a readonly version of the type (to prevent anyone from stuffing this).
+            // Also, we must return a readonly version of the type (to prevent 3rd party extensions from stuffing this).
             // Else it breaks the Jupyter extension
             // We recast to KernelConnectionMetadata as this has already define its properties as readonly.
-            const translatedConnection = Object.freeze(readWriteConnection) as KernelConnectionMetadata;
+
+            const translatedConnection = Object.freeze(
+                JSON.parse(JSON.stringify(connection))
```
rchiodo, line 231: +1

rchiodo (dismissed):

rchiodo:
I wonder if we could tag all of this code with 'Unnecessary if we could get a new API for creating controllers'? Maybe a comment in some key places?

src/notebooks/controllers/controllerRegistration.ts
DonJayamanne: Yes it is required, i'm iterating through the values in a Map (not an array)

src/kernels/types.ts
DonJayamanne: Keeping it here, as the kernelProvider is responsible for determining whether a new controller needs to be created or not when using `getorCreate` and providing the controller, metedata information & the like.

src/kernels/kernelProvider.web.ts
DonJayamanne: Commented in another place, i think its best left in kernel provider.

src/interactive-window/interactiveWindow.ts
```diff
@@ -173,42 +177,48 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             // This cannot happen, but we need to make typescript happy.
             throw new Error('Controller not selected');
         }
-        if (this.currentKernelInfo.kernel) {
-            return this.currentKernelInfo.kernel.promise;
+        if (this.currentKernelInfo.kernelStarted) {
+            return this.currentKernelInfo.kernelStarted.promise.then(() => this.currentKernelInfo.kernel!);
```
sadasant, line 181:
```suggestion
            await this.currentKernelInfo.kernelStarted.promise
            return this.currentKernelInfo.kernel!;
```

This seems more intuitive to me.

src/interactive-window/interactiveWindow.ts
```diff
@@ -173,42 +177,48 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
+                this.currentKernelInfo.kernel = k;
                 this.updateSysInfoMessage(
                     this.getSysInfoMessage(k.kernelConnectionMetadata, SysInfoReason.Start),
                     false,
                     sysInfoCell
                 );
             };
+            const onKernelStartCompleted = async (action: KernelAction, _: unknown, k: IKernel) => {
+                if (action !== 'start' && action !== 'restart') {
+                    return;
```
sadasant, line 206: Would a log here help debug this in the future?

sadasant (dismissed):
It's a bit heavy to review for me, but I really like the idea of it, and I love the comments! Thank you.

src/interactive-window/interactiveWindow.ts
DonJayamanne: i don't think that's necessary.
this call back is only interested in a specific message, hence logging wouldn't server any purpose.

PR #10574, opened 2022-06-26
Split IJupyterSession into Raw and Jupyter

DonJayamanne:
Minor refactor to split the interface into two, 
* Also `IJupyterSession` is confusing name, as we use the term `JupyterSession` mosly when dealing with non-raw (real jupyter sessions). 
Note: We already had two separate interfaces, but we never really made it clear (i.e. we had a seprate interface for Jupyter sessions that inherited `IJupyterSession`).

* Pre for local live kernels (easier when we have two distinct interfaces)

### DIFF
--- a/src/kernels/types.ts
+++ b/src/kernels/types.ts
@@ -270,7 +270,7 @@ export interface IBaseKernelConnectionSession extends IAsyncDisposable {
     readonly status: KernelMessage.Status;
     readonly kernelId: string;
     readonly kernelSocket: Observable<KernelSocketInformation | undefined>;
-    isServerSession(): this is IJupyterServerSession;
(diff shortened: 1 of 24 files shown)


src/kernels/types.ts
```diff
@@ -303,13 +301,17 @@ export interface IJupyterSession extends IAsyncDisposable {
     shutdown(): Promise<void>;
 }
 
-export interface IJupyterServerSession extends IJupyterSession {
```
DonJayamanne, line 306: We already had a specific interface that pointed to the non-raw version.

IanMatthewHuff:
Good call. I think I was responsible for the old names, and I like your new ones better. Captures the meaning.

IanMatthewHuff (approved):

PR #10577, opened 2022-06-26
Do not show modal error everytime for widget error

DonJayamanne:
Basically if we have a widget that registers 20 scripts for a single cell output, then the webview attempts to load the scripts 20 times & for each failure we send a message to the extension host
& on the extension host we display 20 modal dialogs (for the same kernel session) if CDN isn't configured

This PR ensures we display the message for the kernel only once,
E.g. when using widgets like gmaps one will end up getting around 30 modal dialogs, perhaps more (I gave up and had to kill VS Code, restarting VS Code was easier than hitting cancel 30+ times)

### DIFF
--- a/src/kernels/ipywidgets/commonMessageCoordinator.ts
+++ b/src/kernels/ipywidgets/commonMessageCoordinator.ts
@@ -57,6 +57,7 @@ export class CommonMessageCoordinator {
     private jupyterOutput: IOutputChannel;
     private readonly configService: IConfigurationService;
     private webview: IWebviewCommunication | undefined;
+    private modulesForWhichWeHaveDisplayedWisdgetErrorMessage = new Set<string>();
 
     public constructor(
         private readonly document: NotebookDocument,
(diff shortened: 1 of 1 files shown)


src/kernels/ipywidgets/commonMessageCoordinator.ts
```diff
@@ -57,6 +57,7 @@ export class CommonMessageCoordinator {
     private jupyterOutput: IOutputChannel;
     private readonly configService: IConfigurationService;
     private webview: IWebviewCommunication | undefined;
+    private modulesForWhichWeHaveDisplayedWisdgetErrorMessage = new Set<string>();
```
IanMatthewHuff, line 60: Small spelling fix wisdget to widget.

IanMatthewHuff (approved):
Approved with a spelling fix to make.

src/kernels/ipywidgets/commonMessageCoordinator.ts
DonJayamanne, line 60:
```suggestion
    private modulesForWhichWeHaveDisplayedWidgetErrorMessage = new Set<string>();
```

src/kernels/ipywidgets/commonMessageCoordinator.ts
```diff
@@ -169,13 +170,15 @@ export class CommonMessageCoordinator {
...
+            const key = `${payload.moduleName}:${payload.className}:${payload.moduleVersion}`;
             if (!payload.isOnline) {
                 errorMessage = DataScience.loadClassFailedWithNoInternet().format(
                     payload.moduleName,
                     payload.moduleVersion
                 );
                 this.appShell.showErrorMessage(errorMessage).then(noop, noop);
-            } else if (!cdnsEnabled) {
+            } else if (!cdnsEnabled && !this.modulesForWhichWeHaveDisplayedWisdgetErrorMessage.has(key)) {
+                this.modulesForWhichWeHaveDisplayedWisdgetErrorMessage.add(key);
```
DonJayamanne, line 181:
```suggestion
            } else if (!cdnsEnabled && !this.modulesForWhichWeHaveDisplayedWidgetErrorMessage.has(key)) {
                this.modulesForWhichWeHaveDisplayedWidgetErrorMessage.add(key);
```

PR #10597, opened 2022-06-27
Custom test report writer to load test results

### LINKED ISSUE #10389: Web tests output should match node tests output
Web tests are harder to:
- See which tests failed
- Look at the logs

It would be nice if we could unify all of this to look and behave the same.

DonJayamanne:
Fixes #10389

### DIFF
--- a/src/test/web/customReporter.ts
+++ b/src/test/web/customReporter.ts
@@ -1,0 +1,178 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import type * as mochaTypes from 'mocha';
+import { env, extensions, UIKind, Uri, workspace } from 'vscode';
+import { JVSC_EXTENSION_ID_FOR_TESTS } from '../constants';
+import { noop } from '../core';
+const { inherits } = require('mocha/lib/utils');
+const defaultReporter = require('mocha/lib/reporters/spec');
+
+const constants = {
+    EVENT_RUN_BEGIN: 'start',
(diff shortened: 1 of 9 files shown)


rchiodo (approved):

src/test/web/customReporter.ts
```diff
@@ -0,0 +1,178 @@
...
+        try {
+            error[prop] = JSON.parse(JSON.stringify(err[prop]));
+        } catch {
+            error[prop] = '<..?..>';
+        }
+    });
+    return error as Exception;
+}
+
+function CustomReporter(this: any, runner: mochaTypes.Runner, options: mochaTypes.MochaOptions) {
```
DonJayamanne, line 120: Can use ES6 classes, as we cannot import the mocha reporter writers in web, have to use the good old fashioned `require` to import and then extend (again using mocha dependencies to inherit instead of pulling any node specific modules to inherit stuff - as this needs to work in the browser as well).

src/test/web/index.ts
```diff
@@ -47,7 +50,9 @@ export async function activate(context: IExtensionContext): Promise<IExtensionAp
 
             try {
                 // Run the mocha test
-                mocha.run((failures) => {
+                mocha.run(async (failures) => {
+                    // Wait for a while, till we've managed to send the test results to the test server.
+                    await sleep(5_000);
```
DonJayamanne, line 55: We only need a few ms, but added 5s, basically we cannot kill VS Code web until all the results have been sent through. I didn't want to complicate the code by adding more checks & locks to determine whether all the data has gone through, this should work 99.9% of the time & keeps the code simple & clean.

PR #10602, opened 2022-06-28
Support remote commands in the kernel picker

### LINKED ISSUE #10191: Suggestion: Query clipboard permissions when opening an existing server instead of always prompting
Testing https://github.com/microsoft/vscode-jupyter/issues/10170

I find the setting up Jupyter experience kind of complicated. Have you considered doing a soft permissions query instead of a hard text read when connecting to avoid this prompt? Additionally, it would be better from a user's perspective imo to ask for the clipboard on ctrl/cmd+v, not when clicking some button.

Screen Shot 2022-05-31 at 10 12 11 am

```
navigator.permissions.query({ name: 'clipboard-read' }).then(console.log)
```

### LINKED ISSUE #10363: Choosing/changing remote local servers is buggy and confusing.
Prep
- Start a jupyter server (ala run ```jupyter notebook```)
- Start VS Code Insiders (I used build from 6/7/2022 and corresponding Jupyter extension)
- Create a new notebook
- Select a local kernel and run a simple cell

Bug 1
- Run ```Jupyter: Specify Jupyter server for connections```
- Choose the server you created in the prep.
- Press enter

I'm still connected to my local kernel. I expected to have to choose a new kernel.

Bug 2
- Run ```Notebook: Select notebook kernel```

The suggested kernel will be the local kernel I'm already connected to. I expect to either have no suggestion, or a suggestion from the server I just connected to.
image

Bug 3
- Choose one of the server's kernels, run a cell. All seems well.
- Run ```Notebook: Select notebook kernel``` You'll still see the local kernel suggested. Probably a dupe of Bug 2, but still.
- Select the local kernel and run a cell.

You'll run against the local kernel, yet the status bar says you're connected to a Remote Jupyter server. Highly confusing.

Bug 4
- Choose one of the server kernels again.
- Ctrl+C the Jupyter server started in Prep.
- Run a cell in VS Code (which is still "connected" to the dead server)

Cell "executes" indefinitely and I can't interrupt it.
image

I'm forced to reload.

### LINKED ISSUE #10435: Rework remote support to account for more scenarios

rchiodo:
Fixes #10435 
Fixes #10363 
Fixes #10191

Refactor remote support such that:
- There's a top level command in the kernel picker for switching to remote
- You can only be in 'remote' or 'local' mode
- The remote picker lets you type the URI right into the first stage
- If you're in remote mode already, you can
  - Enter another URI
  - Go back to local mode
- If you're in local mode, you can enter remote mode by entering a URI
- The 'Jupyter specify server connection' command also is the same URI picker too. No more 'None' and no more 'Existing' step.

Telemetry for these new modes is already handled. Commands send telemetry on their own and remote/local mode is already sent in telemetry.


rchiodo:
This can be merged after endgame.

src/kernels/jupyter/launcher/serverConnectionType.ts
```diff
@@ -1,41 +0,0 @@
...
-// Licensed under the MIT License.
-
-import { inject, injectable, named } from 'inversify';
-import { EventEmitter, Memento } from 'vscode';
-import { GLOBAL_MEMENTO, IMemento, IsWebExtension } from '../../../platform/common/types';
-
-export const mementoKeyToIndicateIfConnectingToLocalKernelsOnly = 'connectToLocalKernelsOnly';
-
-@injectable()
-export class ServerConnectionType {
```
rchiodo, line 11: I deleted this class and moved its functionality to the JupyterServerUriStorage class.

The reason for this is that this class was essentially exposing the internal state of the JupyterServerUriStorage class and they had to be in sync. And they weren't always.

To make this impossible, I just removed this class altogether and added a new interface to the JupyterServerUriStorage class. Most of the changes for this is simply using an interface instead of a class.

Good use case for why we should expose interfaces to the DI container and not classes.

src/kernels/jupyter/launcher/serverUriStorage.ts
```diff
@@ -28,16 +31,32 @@ export class JupyterServerUriStorage implements IJupyterServerUriStorage {
...
     constructor(
         @inject(IWorkspaceService) private readonly workspaceService: IWorkspaceService,
         @inject(ICryptoUtils) private readonly crypto: ICryptoUtils,
         @inject(IEncryptedStorage) private readonly encryptedStorage: IEncryptedStorage,
         @inject(IApplicationEnvironment) private readonly appEnv: IApplicationEnvironment,
         @inject(IMemento) @named(GLOBAL_MEMENTO) private readonly globalMemento: Memento,
-        @inject(ServerConnectionType) private readonly serverConnectionType: ServerConnectionType
+        @inject(IsWebExtension) readonly isWebExtension: boolean
     ) {
+        // Remember if local only
```
rchiodo, line 51: THis is the code that was originally in the serverConnectionType.ts file.

src/kernels/jupyter/serverSelector.ts
```diff
@@ -177,9 +146,9 @@ export class JupyterServerSelector {
             placeholder,
             items,
             activeItem,
-            title: allowLocal
-                ? DataScience.jupyterSelectURIQuickPickTitle()
-                : DataScience.jupyterSelectURIQuickPickTitleRemoteOnly(),
+            acceptFilterBoxTextAsSelection: true,
```
rchiodo, line 149: Since URI entering happens on the quick pick, we can't use input box validation. So I added this to allow a modal dialog to appear.

src/notebooks/controllers/controllerLoader.ts
```diff
@@ -183,10 +180,6 @@ export class ControllerLoader implements IControllerLoader, IExtensionSyncActiva
 
     private async loadControllersImpl(cancelToken: vscode.CancellationToken) {
         let cachedConnections = await this.listKernels(cancelToken, 'useCache');
-        // Remove all remove kernels if we're no longer interested in them.
```
rchiodo, line 186: All filtering is done in the controllerRegistration.

src/platform/common/utils/multiStepInput.ts
```diff
@@ -146,7 +163,25 @@ export class MultiStepInput<S> implements IMultiStepInput<S> {
...
+                        const itemLabel = selectedItems.length ? selectedItems[0].label : '';
+                        let resolvable = itemLabel ? true : false;
+                        if (itemLabel && validate) {
+                            input.enabled = false;
+                            input.busy = true;
+                            const message = await validate(itemLabel);
+                            if (message) {
+                                resolvable = false;
+                                // No validation allowed on a quick pick. Have to put up a dialog instead
+                                await this.shell.showErrorMessage(message, { modal: true });
```
rchiodo, line 176: I think I need to change this to support markdown. For remote there's a URL the user can click on and it doesn't seem to be clickable at the moment. Maybe dialogs don't support this.

src/standalone/activation/activationManager.ts
```diff
@@ -28,7 +29,15 @@ export class ExtensionActivationManager implements IExtensionActivationManager {
         this.syncActivationServices.map((item) => item.activate());
     }
     public async activate(): Promise<void> {
-        // Activate all activation services together.
-        await Promise.all([this.singleActivationServices.map((item) => item.activate())]);
+        // Activate all activation services together. Don't fail them all if one fails.
+        await Promise.all([
+            this.singleActivationServices.map(async (item) => {
+                const promise = item.activate();
```
rchiodo, line 35: I had a bug in my activate and it made the extension not start. Figured this was better.

src/test/datascience/jupyter/serverSelector.unit.test.ts
```diff
@@ -302,32 +298,16 @@ suite('DataScience - Jupyter Server URI Selector', () => {
...
-            await selector.selectJupyterURI(true);
+            await selector.selectJupyterURI();
 
             assert.equal(showInputBox.firstCall.args[0].value, expectedDefaultUri);
         }
 
         test('Display default uri', async () => {
             await testDefaultUri(defaultUri);
         });
-        test('Display default uri if clipboard is empty', async () => {
```
rchiodo, line 318: Not supporting the clipboard anymore. Fixes the problem with web where the user may not have clipboard permissions. 

New workflow doesn't really need clipboard as the user can also filter in the same entry so defaulting to the url in the clipboard doesn't really make sense.

package.nls.json
```diff
@@ -955,5 +955,10 @@
     "Installer.dataScienceInstallPrompt": "Data Science library {0} is not installed. Install?",
     "Products.installingModule": "Installing {0}",
     "DataScience.webNotSupported": "Operation not supported in web version of Jupyter Extension.",
-    "DataScience.outputSizeExceedLimit": "Output exceeds the <a href={0}>size limit</a>. Open the full output data <a href={1}>in a text editor</a>"
+    "DataScience.outputSizeExceedLimit": "Output exceeds the <a href={0}>size limit</a>. Open the full output data <a href={1}>in a text editor</a>",
+    "DataScience.installPythonTitle": "Install Python",
```
DonJayamanne, line 959: @IanMatthewHuff looks like this will need to be reviewed closly as you're also working in the similar area (installing python when its missing). I.e. you might want to review the workflow & the changes.
Feels like this stuff related to no python might require its own folder/set of classes & its been worked on for months if not years (i.e. workflow keeps changing, hence better to keep it isolated & easier to manage)

Agai, I haven't reviewed anything yet, but just a thought about this (prompting to install python - getting started) part of the exetnsion

src/kernels/jupyter/launcher/serverUriStorage.ts
```diff
@@ -28,16 +31,32 @@ export class JupyterServerUriStorage implements IJupyterServerUriStorage {
     public get onDidRemoveUris() {
         return this._onDidRemoveUris.event;
     }
+    public get currentUri(): string | undefined {
+        return this._currentUri;
```
DonJayamanne, line 35: We could have issues due to race conditions.
Basically `getUri` is async, hence `currentUri` could be empty when it shouldn't be and could give the wrong impression when its used.

DonJayamanne: Isn't this a duplicate of  the `getUri` method, without the async bit.

src/kernels/jupyter/launcher/serverUriStorage.ts
```diff
@@ -28,16 +31,32 @@ export class JupyterServerUriStorage implements IJupyterServerUriStorage {
...
+    }
+    public get onDidChange(): Event<void> {
+        return this._onDidChangeUri.event;
+    }
+    public get isLocalLaunch(): boolean {
+        return this.currentUri === Settings.JupyterServerLocalLaunch || this.currentUri === undefined;
```
DonJayamanne, line 41: I think we can remove `currentUri` and make a sync version of `isLocalLaunch` available by:
* Returning `true` if we're on web
* Retruning `true` if we have a value in the global memento 
     * I.e. when ever we get/update the encrypted store `await this.encryptedStorage.store(Settings.JupyterServerRemoteLaunchService, key, uri);`, then we know we have a remote, at the same time we can update a memento with a flag.
     * & then use that flag to determine if we have a remote or not & thats a synchronous request, hence we don't need promises & we can still have a non-async version of `isLocalLaunch` without accessing the `this.currentUri`

DonJayamanne: Again, as mentioned earlier, if we access this property too early then we might get th ewrong information, e.g. if `currentUri` hasn't been set we'll get `isLocalLaunch` as true even in web cases (which is incorrect).

DonJayamanne: This race condition was resolved by the `ServerConnectionType` class.

src/kernels/jupyter/launcher/serverUriStorage.ts
rchiodo: Yes. The sync version. getUri just uses this value once it's set anyway.

src/kernels/jupyter/launcher/serverUriStorage.ts
```diff
@@ -28,16 +31,32 @@ export class JupyterServerUriStorage implements IJupyterServerUriStorage {
...
     ) {
+        // Remember if local only
+        this._localOnly = isWebExtension
+            ? false
+            : this.globalMemento.get<boolean>(mementoKeyToIndicateIfConnectingToLocalKernelsOnly, true);
+
         // Cache our current state so we don't keep asking for it from the encrypted storage
-        this.getUri().ignoreErrors();
+        this.getUri()
+            .then((s) => (this._currentUri = s))
```
DonJayamanne, line 58: I don't think we can assume this will happen very quickly, its possible the encrypted storage is very slow (or could present UI dialog boxes to the user and the like) & the rest of the code will run to completion, leaving the properties like `currentUri` & `isLocalLaunch` in a bad (incorect state).

E.g. in the past when ever we accessed encrypted storage, we'd get a dialog prompting the user to allow accessing secret store, this could happen in the future as well, (.e. this promise could get resolved seconds or minutes later)

src/kernels/jupyter/launcher/serverUriStorage.ts
rchiodo: And no it wasn't solved by the ServerConnectionType class. It was merely hidden. ServerConnectionType would say not local while the URI wasn't set yet. So if you tried to get the URI, you'd be off of the actual state.

It's why I removed the ServerConnectionType.

rchiodo: In fact I think we can get rid of the async version. We don't need it. It could be internal only.

DonJayamanne: > say not local while the URI wasn't set yet. So if you tried to get the URI, you'd be off of the actual state.

We can solve this easily by updating the mememto & a private property.
Hence if the private property has been set, then use that else use memento.
When the class is instantiated the first time (ie. loading VS Code), then property will be `undefined` and we fall back to mememto.

& later on everytime we use the property as that will have the latest state.

src/kernels/jupyter/launcher/serverUriStorage.ts
DonJayamanne: along with a private property (so that we have sync & latest access to information - even while we're still busy updating the memeto)

rchiodo: Actually isLocalLaunch should use the _local flag. That was a copy of the logic in the ServerConnectionType. I'll update.

src/kernels/jupyter/launcher/serverUriStorage.ts
rchiodo: The whole problem was this state that is mixed between two classes. That seemed broken to me. The true state was in the SErverUriStorage, not the ServerConnectionType.

src/kernels/jupyter/launcher/serverUriStorage.ts
DonJayamanne: I think we can remove this, we don't need a suync version, we can make this async &
just make other parts of the code async when accessing this.
From what I could tell we use this in adding controllers, and we need to do is turn the `IControllerRegistration.add` method into an async method.
From what I could tell that should be totally fine, as almost every method that calls the `add` method is already async.

src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -161,6 +180,14 @@ export class ControllerRegistration implements IControllerRegistration {
         return this.registeredControllers.get(id);
     }
 
+    private isFiltered(metadata: KernelConnectionMetadata): boolean {
+        const userFiltered = this.kernelFilter.isKernelHidden(metadata);
+        const connectionTypeFiltered = isLocalConnection(metadata) !== this.isLocalLaunch;
+        const urlFiltered =
+            isRemoteConnection(metadata) && !this.serverUriStorage.currentUri?.includes(metadata.baseUrl);
```
DonJayamanne, line 187: I think you can use `serverId` instead of `baseUrl` that would still allow you to keep all of this synchronous

src/kernels/jupyter/launcher/serverUriStorage.ts
DonJayamanne: From what I can tell we only need `currentUri` for filtering:
* We could use the `getUri` async method & turn all others into async
* Or just use the `serverId` (we can save server Id in memto as that's safe & hashed, and its part of the connection metadata, & since thats all synchronous, we can use that via `currentServerId` property & then use that for filtering)

rchiodo: Wait what's serverId? Is that the hash of the connection information.

src/notebooks/controllers/controllerRegistration.ts
rchiodo: serverId is a hash of the url? Not sure about that. It's not cryptographically secure. I mean if that was allowed, we wouldn't need encrypted storage.

src/kernels/jupyter/launcher/serverUriStorage.ts
rchiodo: Hmm. ServerId is a hash of the current uri. It's not secure but it's only in memory. If we wrote it to a memento then there's a risk of leaking somebody's token. I guess there's a risk of leaking the token in memory too, but same with encrypted storage. 

I don't think we can solve this with server id, because we can't write it to the memento.

rchiodo: currentUri being sync is only a problem if controllerRegistration.add is called before it's resolved. 

However if it's resolved and the onDidChangeUri event were to fire, it doesn't matter. It's the same as if it was originally set to remote.

Oh but that still means isLocalLaunch wouldn't match. Hmm. 

So then we'd have to force async.

That ripples through the controllerRegistration add, onDidChangeUri, and on didChangeKernelFilter

- add being async seems possible.
- onDidChangeUri would should work if it actually gets passed the URI involved. 
- onDidChangeKerneFilter could be made async, doesn't really matter

Okay I think I can get rid of the synchronous of the currentUri.

src/notebooks/controllers/controllerRegistration.ts
rchiodo: Going to try doing this with async instead.

src/kernels/jupyter/launcher/serverUriStorage.ts
DonJayamanne: > I don't think we can solve this with server id, because we can't write it to the memento.

Given that we can store it in memento, i think we can solve it

DonJayamanne: > Oh but that still means isLocalLaunch wouldn't match. Hmm.

My suggestion is:
* Create a property to keep track of isloca or remote (initially undefiend)
* When loading the first time, this property is empty and we read the value from memento to get the server id or the like (then we know whether its local or not)
* Subsequently when we update memento, we can update this property as well
* Then next time we can check if this property is not undefined & use the value in that property.
Thats how the serverConnectionType class worked, i.e. it solved this problem by having synchronous access via the `_isLocalLaunch` private property.

src/notebooks/controllers/controllerRegistration.ts
DonJayamanne: > The difficult part is an update after onDidChangeUri. Maybe if I pass the URI there and during add, I can skip the sync 'currentUri'

See comments above, the previous class `ServerConnectionType._isLocalLaunch`

src/kernels/jupyter/launcher/serverUriStorage.ts
rchiodo: hashjs is not cryptographically secure. We'd have to use something else. The security review team specifies this when we said we don't use it to store anything but just to hash stuff that isn't a password. We start using it to store passwords (which is what the token in the url is) we need to use something other than hashjs.

package.nls.json
IanMatthewHuff: @DonJayamanne I think Rich was just adding on the Command localization that I missed in my addition. The only real conflict to consider here is just the fact that we won't have only one command anymore in the new user scenario. Means that we don't get that nice "just hit the run button" launch, but we can tweak that as we move forward. Maybe prioritization for commands in the kernel picker somehow.

src/notebooks/controllers/commands/serverConnectionControllerCommands.ts
```diff
@@ -0,0 +1,307 @@
...
+                label: d.label,
+                detail: undefined,
+                description: d.controller.description,
+                controller: d
+            };
+        });
+
+        // Then group them by kind
+        const kernelsPerCategory = groupBy(picks, (a, b) =>
+            compareIgnoreCase(a.controller.controller.kind || 'z', b.controller.controller.kind || 'z')
```
IanMatthewHuff, line 240: Is the 'z' for alphabetizing? Doesn't seem like that would work for non-en.

src/platform/common/utils/multiStepInput.ts
IanMatthewHuff: I think they don't sadly. Our "Install Python" popup tries to link to Python.org, but it's not clickable. Pretty sure when we talked about it in v-team that's just how it is.

IanMatthewHuff:
Seems good to me. This is going to be a big improvement. Honestly the hardest part of this PR was that I find the MultiStep input control flow really wacky and hard to read. But that's not anything on you or this PR, I think I just need to play with that myself a bit versus looking at the PR.

src/notebooks/controllers/commands/serverConnectionControllerCommands.ts
rchiodo: I copied this from the core, but it does seem wrong. 'z' only works in ascii.

rchiodo:
> Seems good to me. This is going to be a big improvement. Honestly the hardest part of this PR was that I find the MultiStep input control flow really wacky and hard to read. But that's not anything on you or this PR, I think I just need to play with that myself a bit versus looking at the PR.

I agree about the multistep input. It's really a state machine. I think it would be better with an iterator type pattern instead of a promise type pattern. Took me a long time to get the promises to work out correctly.

src/notebooks/controllers/commands/serverConnectionControllerCommands.ts
rchiodo: I'll enter an issue to relook at this in both the core and here.

src/kernels/jupyter/serverSelector.ts
```diff
@@ -188,17 +157,17 @@ export class JupyterServerSelector {
                     );
                     items.splice(items.indexOf(e.item), 1);
                     onDidChangeItems.fire(items.concat([]));
+                } else if (e.button) {
```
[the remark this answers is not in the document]
rchiodo: Yeah there's only two buttons for the server selector. Back and the 'delete' this URI.

DonJayamanne (approved):

IanMatthewHuff (approved):

src/kernels/jupyter/serverSelector.ts
```diff
@@ -54,11 +51,7 @@ export type SelectJupyterUriCommandSource =
     | 'prompt';
 @injectable()
 export class JupyterServerSelector {
-    private readonly localLabel = `$(zap) ${DataScience.jupyterSelectURINoneLabel()}`;
-    private readonly newLabel = `$(server) ${DataScience.jupyterSelectURINewLabel()}`;
-    private readonly remoteLabel = `$(server) ${DataScience.jupyterSelectURIRemoteLabel()}`;
```
sadasant, line 59: Are these not needed anymore?

rchiodo: No they shouldn't be. Maybe. I think we might need the 'None' label.

src/notebooks/controllers/commands/serverConnectionControllerCommands.ts
```diff
@@ -0,0 +1,307 @@
...
+    for (const element of data.slice(0).sort(compare)) {
+        if (!currentGroup || compare(currentGroup[0], element) !== 0) {
+            currentGroup = [element];
+            result.push(currentGroup);
+        } else {
+            currentGroup.push(element);
+        }
+    }
+    return result;
+}
```
sadasant, line 43: Ohh fancy! Would it make sense to have a common place for functional utilities like this?

src/test/datascience/debugger.vscode.test.ts
```diff
@@ -118,7 +118,7 @@ suite('VSCode Notebook - Run By Line', function () {
             );
 
             // Give the files a quick chance to clean up
-            await sleep(1000);
+            await sleep(3000);
```
sadasant, line 121: Is this necessary?

src/test/datascience/interactiveDebugging.vscode.common.ts
```diff
@@ -96,7 +96,9 @@ export function sharedIWDebuggerTests(
                 await closeNotebooksAndCleanUpAfterTests(disposables);
             });
 
-            test('Debug a cell from a python file', async () => {
+            // Flakey test: https://github.com/microsoft/vscode-jupyter/issues/10521
+
+            test.skip('Debug a cell from a python file', async () => {
```
sadasant, line 101: Should we have an issue to re-enable this test?

src/test/datascience/interactiveWindow.vscode.test.ts
```diff
@@ -99,7 +99,8 @@ suite(`Interactive window Execution`, async function () {
         await vscode.commands.executeCommand('python.clearWorkspaceInterpreter');
     }
 
-    test('Export Interactive window to Notebook', async () => {
+    // Flakey test: https://github.com/microsoft/vscode-jupyter/issues/10649
+    test.skip('Export Interactive window to Notebook', async () => {
```
sadasant, line 103: Should we have an issue to re-enable this test?

src/test/datascience/interactiveDebugging.vscode.common.ts
sadasant: I see the issue! Disregard.

src/test/datascience/debugger.vscode.test.ts
rchiodo: Test was failing attempting to see if files were cleaned up. Stopped failing after I did this, so yeah probably.

There's likely another way to accomplish this but I wasn't working on this code so I just wanted the test to pass

src/notebooks/controllers/commands/serverConnectionControllerCommands.ts
rchiodo: Maybe. Not sure where that would be. It's only used in this class at the moment.

src/test/datascience/debugger.vscode.test.ts
sadasant: Ok that makes sense! Thanks for the context

PR #10604, opened 2022-06-28
Web version of the DataFrame

### LINKED ISSUE #9665: DataFrame working in web extension

### LINKED ISSUE #10310: De-reference the constructor from the rendering of VariableExplorerButtonCellFormatter

sadasant:
The only thing missing on the web to enable the DataFrame features was to have a web-ready version of the `DataViewerDependencyService`.

After discussing ideas with the team, our immediate goals are:

- To ensure the web variable viewer can use the data viewer.
- In the case the data viewer is accessed through the variable viewer, it should confirm that it has `pandas` and install it if it does not have it.
  - This is possible because we have a Kernel at hand thanks to the variable viewer.
- In the case theta viewer is accessed through the debuggers:
  - It should work on Node.js as is today.
  - On the web, it should check if it has `pandas` installed. If so, it should work. If it doesn't have `pandas`, it should throw an error, asking users to install `pandas`.

To accomplish the above, this PR:

- Changes the Node.js version of the `DataViewerDependencyService` to accept an `IKernel` besides the interpreter that it accepts today, and then retrieve the interpreter from the kernel to keep the current behavior intact.
- Add a web version of the `DataViewerDependencyService` that would accept an `IKernel` and would also load the `IJupyterVarialbes` on the constructor, to either use the kernel or the debugger to at least get the version of `pandas`.
- On the web, if no kernel is provided and if `IJupyterVarialbes` did load, the debugger is able to read the version of `pandas` if it exists, and the data viewer works as is then. If `pandas` does not exist, it will throw an error asking the user to install `pandas`.
- When an `IKernel` is provided, it would use either `%conda` or `%pip` based on the `IKernel`'s `KernelConnectionMetadata`. `%conda` if it had an interpreter (local case) that was conda based. Otherwise `%pip`.

Besides the above, which are the main changes of this PR, this PR also:

- Moves the `parseSemVer` function from `platform/common/utils.node` to `platform/common/utils`.
- Adds `IDataViewerDependencyService` to the extension-side web service registry.
- `removes the `isWeb` conditional from the `variableExplorerButtonCellFormatter.tsx`.

Thus:
Fixes #9665 
Fixes #10310

Here are some issues that are not fixed with this PR:

- This PR does not add support for an isomorphic IInstaller, which would be accomplished by figuring out how to install dependencies with the debugger #10605
- Once we have an isomorphic `IInstaller` through #10605 , we should update the `DataViewDependencyService` to only use the `IInstaller` #10606

Finally, a comment about the current state of the debugger on the web:
- The Python debugger (from the `vscode-python` extension) does not currently work on the `openInBrowser` approach to test our web features.

Feedback always appreciated!

---


src/kernels/variables/debuggerVariables.ts
```diff
@@ -311,7 +310,7 @@ export class DebuggerVariables
                 return results;
             } else {
                 traceError(`Cannot evaluate ${code}`);
-                return undefined;
+                throw new Error(`Cannot evaluate ${code}`);
```
sadasant, line 313: Do we have code that cannot be evaluated by these methods in the debugger?

src/kernels/variables/debuggerVariables.ts
```diff
@@ -297,8 +297,7 @@ export class DebuggerVariables
         this.importedGetVariableInfoScriptsIntoKernel.delete(key);
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    private async evaluate(code: string, frameId?: number): Promise<any> {
+    async evaluate(code: string, frameId?: number): Promise<DebugProtocol.EvaluateResponse['body']> {
```
sadasant, line 300: I'm exposing this function so I can use it in the debugger path within the data viewer dependency service.

src/test/datascience/data-viewing/dataViewerDependencyServic.node.unit.test.ts
```diff
@@ -0,0 +1,121 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
sadasant, line 1: This is the same file that was in `src/test/datascience/data-viewing/dataViewerDependencyServic.node.test.ts`. Only that `checkAndInstallMissingDependencies` now receives an object.

src/test/datascience/data-viewing/dataViewerDependencyService.unit.test.ts
```diff
@@ -4,118 +4,136 @@
...
-
-        dependencyService = new DataViewerDependencyService(
-            instance(appShell),
-            instance(installer),
-            instance(pythonExecFactory),
-            instance(interpreterService),
-            false
-        );
+        kernel = instance(mock<IKernel>());
+        dependencyService = new DataViewerDependencyService(instance(appShell), false, undefined);
```
sadasant, line 27: These are the kernel-specific tests. This code works in Node and in the browsers.

src/test/datascience/data-viewing/dataViewerDependencyServiceDebugger.unit.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        interpreter = {
+            displayName: '',
+            uri: Uri.file(path.join('users', 'python', 'bin', 'python.exe')),
+            sysPrefix: '',
+            sysVersion: '',
+            version: new SemVer('3.3.3')
+        };
+        appShell = mock(ApplicationShell);
+        variableProvider = instance(mock<IJupyterVariables>());
+        dependencyService = new DataViewerDependencyService(instance(appShell), false, variableProvider);
```
sadasant, line 37: These tests use the debugger.

src/webviews/extension-side/dataviewer/dataViewerCommandRegistry.ts
```diff
@@ -99,7 +99,10 @@ export class DataViewerCommandRegistry implements IExtensionSingleActivationServ
                         this.debugService.activeDebugSession.configuration
                     );
 
-                    pythonEnv && (await this.dataViewerDependencyService.checkAndInstallMissingDependencies(pythonEnv));
+                    pythonEnv &&
+                        (await this.dataViewerDependencyService.checkAndInstallMissingDependencies({
+                            interpreter: pythonEnv
+                        }));
```
sadasant, line 105: The interpreter is only being used in Node.js, in the web the interpreter is ignored and the debugger is used (in this case, because here we don't have a kernel attached).

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,201 @@
...
+// It takes a long time, then it breaks with:
+// error: Command \"clang -Wno-unused-result -Wsign-compare -Wunreachable-code...
+export const debuggerInstallPandas = [
+    'import subprocess as _VSCODE_subprocess',
+    'import sys as _VSCODE_sys',
+    `def _VSCODE_install_pandas():
+  _VSCODE_subprocess.check_call([_VSCODE_sys.executable, "-m", "pip", "install", "pandas"])`,
+    '_VSCODE_install_pandas()',
+    'del _VSCODE_subprocess, _VSCODE_sys, _VSCODE_install_pandas'
+];
```
sadasant, line 48: I'm thinking on leaving this here for future reference.

src/kernels/variables/types.ts
```diff
@@ -56,6 +57,7 @@ export interface IJupyterVariables {
     ): Promise<IJupyterVariable | undefined>;
     // This is currently only defined in kernelVariables.ts
     getVariableProperties?(name: string, kernel?: IKernel, cancelToken?: CancellationToken): Promise<JSONObject>;
+    evaluate(code: string, frameId?: number): Promise<DebugProtocol.EvaluateResponse['body']>;
```
DonJayamanne, line 60: I don't think `IJuptyerVariables1` needs such a generic method.
Basically we're exposing the ability to execute random code in an interface used specifically to get variables.
Also `frameId` is specific to debugger, hence this doesn't belong here.

src/kernels/variables/kernelVariables.ts
```diff
@@ -406,4 +407,8 @@ export class KernelVariables implements IJupyterVariables {
         }
         return this.enhancedTooltipsExperimentPromise;
     }
+
+    evaluate(): Promise<DebugProtocol.EvaluateResponse['body']> {
+        throw new Error('Not implemented.');
```
DonJayamanne, line 412: This method shouldn't be added here

src/webviews/extension-side/dataviewer/types.ts
```diff
@@ -128,7 +128,16 @@ export interface IJupyterVariableDataProviderFactory {
...
+export interface IdataViewerDependencyServiceOptionsWithInterpreter {
+    interpreter: PythonEnvironment;
+}
+export type IDataViewerDependencyServiceOptions =
+    | IDataViewerDependencyServiceOptionsWithKernel
+    | IdataViewerDependencyServiceOptionsWithInterpreter;
 export const IDataViewerDependencyService = Symbol('IDataViewerDependencyService');
 export interface IDataViewerDependencyService {
-    checkAndInstallMissingDependencies(interpreter: PythonEnvironment): Promise<void>;
+    checkAndInstallMissingDependencies(options: IDataViewerDependencyServiceOptions): Promise<void>;
```
DonJayamanne, line 142: This should change to `IKernel`, we don't need an overload to pass an interprereter.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,201 @@
...
+            sendTelemetryEvent(Telemetry.UserDidNotInstallPandas);
+            throw new Error(DataScience.pandasRequiredForViewing());
+        }
+    }
+
+    private async executeSilently(
+        commands: string[],
+        options: IDataViewerDependencyServiceOptions
+    ): Promise<(string | undefined)[]> {
+        const kernel = (options as IDataViewerDependencyServiceOptionsWithKernel).kernel;
```
DonJayamanne, line 178: This doesn't look right.
Dependency service shouldn't take a dependency on a variable provider.
At this point, all we need to do is:
* Check if we have an active debug session
* If we have a debug session, then use the debug sessions `evaluate` method to run some aribtrary code & get the desired output from there.

Today we check in other places whther we have an active Debug session using `IDebugService.activeDebugSession` and if that is not undefined then we know we have a debug session.
Next you can use `IDebugService.activeDebugSesion.customRequest` method to send a custom request.

DonJayamanne (changes requested):
Variable viewer should not expose any method to execute arbitarry code, that breaks the API.
We should only have one way to execute code:
* Debugger session using the `customRequest` method
* & methods on IKernel

Having other classes call into a variable class to evaluate code when it doesn't need anything from variables seems in correct.

I've provided a high level solution on how we could achieve this by passig in the IKernel instead of any other object.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: - To use activeSession.customRequest we need a frame Id, and the frame Id of the variable in the Python debugger used to show the data viewer does not work. It runs the code and shows to succeed even in the logs, but it does not yield any value, therefore that frameId is not useful, and no frameId doesn’t work either.
- The debugger cannot currently be used to install pandas, the installation doesn’t finish and the debugger becomes unusable afterwards.

DonJayamanne: Please see my comments here https://github.com/microsoft/vscode-jupyter/issues/10605#issuecomment-1168183711

Again, I don't think we should expose `evaluate` method onn a vairbla provider class for anyone to use.

DonJayamanne: > The debugger cannot currently be used to install pandas, the installation doesn’t finish and the debugger becomes unusable afterwards.

We can get the path to th eexecutable using `import sys;sys.executable` and use the old code to install using the interpreter information.

src/kernels/variables/debuggerVariables.ts
DonJayamanne: Not a fan of this

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
DonJayamanne: > ctiveSession.customRequest we need a frame Id, and the frame Id of the variable 

Not following your comments here.
Are you saying that frameId is useless or it is required. 
Looking at the code you have exposed `frameId` as an optional argument in some new methods. & if its not used, then I dont see the need to expose such an argument to methods.

>ut it does not yield any value, therefore that frameId is not useful, and no frameId doesn’t work either.

Have you tried a frame id of `0`?

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: @DonJayamanne you can see the approach I took here. Here I used `sys.executable`. This code above does work — except that it crashes after a while and makes the debugger unusable.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: Regarding frameId:

- The frameId of the variable coming from the debugger causes the results to be empty.
- Frame Id 0 doesn’t work either, I can show you on a call.
- One other frameId worked for Rich and I but essentially we could only get hold of it by using the evaluate method from IJupyterVariables (If I understood him correctly. It does work — when getting the version at the very least, then the installation has another challenge but the installation begins to work, just doesn’t finish).

src/webviews/extension-side/dataviewer/types.ts
sadasant: The issue is that when we are in the debugger, we don’t have a kernel.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: I did use `sys.executable` [here](https://github.com/microsoft/vscode-jupyter/pull/10604#discussion_r908347095) and it (later, in the middle of the installation) causes a major issue in the web. I believe that you mean only to extract the interpreter path and use the existing interpreter code then, if that’s correct I need to investigate how to make that code work on the web.

src/kernels/variables/kernelVariables.ts
DonJayamanne: > I can implement it as part of this PR during the next cycle, given that it seems unlikely to get this PR merged soon.

I prefer this option.

src/webviews/extension-side/dataviewer/types.ts
DonJayamanne: In that case, I'd suggest passing either `Ikernel` or `DebugSession` so that its more obvious what the choices are.
From `DebugSession` the method can then determine the `sys.executable` & then get the interpreter & run the same old code as in the past.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
DonJayamanne: >  Python debugging doesn't quite work in the web)

Exactly, hence the only time we're debugging is when we launch the debugger for IW or notebooks & we know exactly what the IKernel is in that case.

IanMatthewHuff:
@DonJayamanne and @sadasant I don't believe there is a rush here for this particular release. At this point we are already in endgame. I don't believe we should be taking this big a change at this point regardless of the review state. It's the plus of the shorter release cycle it should just go on the next train.

I'd agree with Don that I don't want the evaluate access on the variables, that doesn't seem like the right spot for it to me. I'd expect that to be via a more general mechanism of some type, not just arbitrary code execution. That adds a really broad access point into a scoped interface. Regarding the IKernel access my understanding is that when debugging is launched via normal .py debugging (for the open in data viewer scenario) we won't have access to the IKernel so that scenario would need debugger install for the web.

I worry that we're yanking this PR around too much which isn't good for Daniel's time. This is just a proposal from my side, but what about we start with maybe a meeting on the broader architecture here. I feel like it might be more time efficient at a general interface level before the implementation get's fully hashed out.

src/kernels/variables/types.ts
rchiodo: Maybe we could make a more generic service that you can execute a piece of code on. The 'executeSilently' method we use elsewhere seems like it would be a good candidate. It could keep track of the topmost frame id and use the debugger if the kernel was 'frozen' at a breakpoint.

PR #10625, opened 2022-06-28
format traceback for standalone IW in IPython8

### LINKED ISSUE #10614: Error highlight in Interactive window makes text unreadable in dark mode
the highlight for the undefined value in a NameError is too close to the text color:
image

amunger:
Fixes #10614

We don't have anything stored for generated code in a standalone Interactive Window, but we still need to fix up some formatting for IPython8.

We also don't render links for line numbers in this case, and this PR does not fix that.

### DIFF
--- a/src/interactive-window/outputs/tracebackFormatter.ts
+++ b/src/interactive-window/outputs/tracebackFormatter.ts
@@ -27,20 +27,24 @@ export class InteractiveWindowTracebackFormatter implements ITracebackFormatter 
             return traceback;
         }
         const storage = this.storageFactory.get({ notebook: cell.notebook });
-        if (!storage) {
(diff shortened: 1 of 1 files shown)


rchiodo (approved):

src/interactive-window/outputs/tracebackFormatter.ts
```diff
@@ -27,20 +27,24 @@ export class InteractiveWindowTracebackFormatter implements ITracebackFormatter
             return traceback;
         }
         const storage = this.storageFactory.get({ notebook: cell.notebook });
-        if (!storage) {
+        const useIPython8Format = traceback.some((traceFrame) => /^[Input|File].*?\n.*/.test(traceFrame));
```
rebornix, line 30: This looks like something we can write unit test for.

PR #10650, opened 2022-06-30
Add a long timeout to make sure we don't wait forever to hook the python API after install

### LINKED ISSUE #10617: installing python and locating kernels should stop after certain amount of time
Testing #10590

in the output channel for jupyter, i'm seeing this:
`
 Find preferred kernel for /Users/meganrogge/Repos/vscode/src/vs/workbench/contrib/terminal/browser/test.ipynb with metadata {"orig_nbformat":4,"language_info":{"name":"python"}} & preferred interpreter `

and nothing has happened after 1 min

IanMatthewHuff:
Fixes #10617

### DIFF
--- a/src/notebooks/controllers/installPythonControllerCommands.ts
+++ b/src/notebooks/controllers/installPythonControllerCommands.ts
@@ -10,6 +10,7 @@ import {
     notebooks,
     window
 } from 'vscode';
+import { IDataScienceErrorHandler } from '../../kernels/errors/types';
 import { isPythonKernelConnection } from '../../kernels/helpers';
 import { IExtensionSingleActivationService } from '../../platform/activation/types';
 import { IPythonApiProvider, IPythonExtensionChecker } from '../../platform/api/types';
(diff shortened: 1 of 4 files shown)


src/notebooks/controllers/installPythonControllerCommands.ts
```diff
@@ -179,9 +183,10 @@ export class InstallPythonControllerCommands implements IExtensionSingleActivati
                 // Don't move forward until we have hooked the API
                 // Note extensions.installExtension seems to return "mostly" after the install is done, but at that
                 // point we don't see it installed via the checker and don't have the API so wait for it here
-                await this.pythonApi.pythonExtensionHooked;
+                const hookResult = await Promise.race([sleep(60_000), this.pythonApi.pythonExtensionHooked]);
```
IanMatthewHuff, line 186: The timeout is a bit long here, but that was intentional. After VSCode returns from the .installExtension command it doesn't actually show up as installed yet. We need to wait a bit for it to actually get registered and activated. Since python takes a while to activate made a longer wait here. On the plus, it's just activation and not all interpreter discovery.

rchiodo: Is there a way to tell if something is disabled instead of not installed?

rchiodo: Ideally we'd skip installing and instead we'd ask to enable.

PR #10651, opened 2022-06-30
increase execution count for cells added from the input box

### LINKED ISSUE #10631: error stack trace does not have links in IW if code has been run from the input box
As soon as code is executed from the input box, errors will stop rendering links on the line numbers.

Code that is executed through the input box does not go through the `CodeGenerator`, and does not increase the execution count, so we end up with a mismatch and the `TracebackFormatter` cannot find the code to reference

image

amunger:
Fixes #10631

When a cell is added from the input box, we don't have enough information to fully generate the code, but we still need to bump the execution count so that the traceback formatter can find the correct code when adding links to stack traces

### DIFF
--- a/src/interactive-window/editor-integration/types.ts
+++ b/src/interactive-window/editor-integration/types.ts
@@ -110,6 +110,7 @@ export type InteractiveCellMetadata = {
 };
 
 export interface IInteractiveWindowCodeGenerator extends IDisposable {
+    bumpExecutionCount(): void;
     reset(): void;
     generateCode(
         metadata: Pick<InteractiveCellMetadata, 'interactive' | 'id' | 'interactiveWindowCellMarker'>,
--- a/src/interactive-window/interactiveWindow.ts
+++ b/src/interactive-window/interactiveWindow.ts
@@ -160,6 +160,28 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         } else if (this.isWebExtension) {
             this.insertInfoMessage(DataScience.noKernelsSpecifyRemote()).ignoreErrors();
         }
+
+        this.listenForNewNotebookCells();
+    }
+
+    private listenForNewNotebookCells() {
+        workspace.onDidChangeNotebookDocument(
+            (e) => {
+                e.contentChanges.forEach((change) => {
+                    change.addedCells.forEach((cell: NotebookCell) => {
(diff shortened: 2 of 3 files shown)


src/interactive-window/interactiveWindow.ts
```diff
@@ -160,6 +160,28 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
             this.insertInfoMessage(DataScience.noKernelsSpecifyRemote()).ignoreErrors();
         }
+
+        this.listenForNewNotebookCells();
+    }
+
+    private listenForNewNotebookCells() {
+        workspace.onDidChangeNotebookDocument(
+            (e) => {
+                e.contentChanges.forEach((change) => {
```
DonJayamanne, line 170: You'd need to filter this and ensure we only check for IW notebook types.

DonJayamanne: Also, I think you can compare the notebook here, instead of doing that further below.
Else when you update a cell of another IW, we still end up running the for loop unnecessarily.

src/interactive-window/editor-integration/types.ts
```diff
@@ -110,6 +110,7 @@ export type InteractiveCellMetadata = {
 };
 
 export interface IInteractiveWindowCodeGenerator extends IDisposable {
+    bumpExecutionCount(): void;
```
DonJayamanne, line 113: Would be good to add some comments here when this should be used

DonJayamanne (approved):
Few suggestions

src/interactive-window/interactiveWindow.ts
```diff
@@ -160,6 +162,8 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         } else if (this.isWebExtension) {
             this.insertInfoMessage(DataScience.noKernelsSpecifyRemote()).ignoreErrors();
         }
+
+        this.codeGenerator = codeGeneratorFactory.getOrCreate(this.notebookDocument);
```
DonJayamanne, line 166: If we're going to own this object (by storing a ref here), then we should dispose this object in the `dispose` method of this class.

src/interactive-window/editor-integration/types.ts
DonJayamanne: N/A

rchiodo (approved):

PR #10661, opened 2022-07-01
Set baseUrl of Jupyter after initialization

### LINKED ISSUE #10510: Matplotlib widget does not load when CDN is disabled and when connected to a Remote Jupyter Server
* First render a cell with a 3rd party widget & enable CDNs
* Then go into settings and update the widget cdn as follows `"jupyter.widgetScriptSources": []`
* Now reload VS Code (basically we've turned off CDN)
* Connect to a remote server
* Run a cell that renders a Matplotlib plot
* Nothing happens for 1 full minute

Basically we're at a dead lock,
* The kernel messages are being synced and any subsequent kernel message cannot be procesed as we're still waiting for the UI kernel message to compete,
* However the UI side is waiting for widgets to load
* On the extension host side, the code is waiting for some new kernel request to complete (fetch list of widgets via an exec request)

Solutions
* Ensure we load list of widgets on remote early
* Ensure we don't wait for previous kernel messages to complete when sending internal requests (requests generated by the extension)

DonJayamanne:
Fixes #10510

### DIFF
--- a/src/kernels/ipywidgets/ipyWidgetScriptSource.ts
+++ b/src/kernels/ipywidgets/ipyWidgetScriptSource.ts
@@ -86,17 +86,7 @@ export class IPyWidgetScriptSource {
                 this.uriConverter.resolveUri(response.request, response.response);
             }
         } else if (message === IPyWidgetMessages.IPyWidgets_Ready) {
-            // Send to UI (even if there's an error) instead of hanging while waiting for a response.
-            (this.scriptProvider ? this.scriptProvider?.getBaseUrl() : Promise.resolve())
-                .then((baseUrl) => {
-                    if (baseUrl) {
-                        this.postEmitter.fire({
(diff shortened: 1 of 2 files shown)


src/kernels/ipywidgets/ipyWidgetScriptSource.ts
```diff
@@ -86,17 +86,7 @@ export class IPyWidgetScriptSource {
                 this.uriConverter.resolveUri(response.request, response.response);
             }
         } else if (message === IPyWidgetMessages.IPyWidgets_Ready) {
-            // Send to UI (even if there's an error) instead of hanging while waiting for a response.
-            (this.scriptProvider ? this.scriptProvider?.getBaseUrl() : Promise.resolve())
```
DonJayamanne, line 90: Nullables...
When the `scriptProvider` was null, we woudln't send this.
Hence fixed to send this baseUrl after the scriptProvider has been initialized as well.

rchiodo (approved):

PR #10673, opened 2022-07-01
Change to using msrCrypto to compute hash for serverId

rchiodo:
Fixes https://github.com/microsoft/vscode-jupyter-internal/issues/306

This algorithm is identical to hashjs so all existing cached values should compare the same after this change.


src/kernels/jupyter/jupyterUtils.ts
```diff
@@ -133,8 +138,18 @@ export function createRemoteConnectionInfo(
     };
 }
 
-export function computeServerId(uri: string) {
-    return hashjs.sha256().update(uri).digest('hex');
+export async function computeServerId(uri: string) {
+    try {
+        const inputBuffer = new TextEncoder().encode(uri);
+        const hashBuffer = await crypto.subtle.digest({ name: 'SHA-256' }, inputBuffer);
```
rchiodo, line 144: The difference is the digest function is a promise. That rippled all over the place.

src/kernels/jupyter/launcher/serverUriStorage.ts
```diff
@@ -122,7 +130,9 @@ export class JupyterServerUriStorage implements IJupyterServerUriStorage, IServe
             blob
         );
     }
-    public async getSavedUriList(): Promise<{ uri: string; time: number; displayName?: string | undefined }[]> {
+    public async getSavedUriList(): Promise<
+        { uri: string; serverId: string; time: number; displayName?: string | undefined }[]
```
rchiodo, line 134: Instead of having to get the server id for the saved uri list all over the place, I just changed this to return it too.

src/notebooks/controllers/controllerLoader.ts
```diff
@@ -75,26 +74,6 @@ export class ControllerLoader implements IControllerLoader, IExtensionSyncActiva
         // Make sure to reload whenever we do something that changes state
         const forceLoad = () => this.loadControllers(true);
         this.serverUriStorage.onDidChangeUri(forceLoad, this, this.disposables);
-        this.serverUriStorage.onDidRemoveUris(
```
rchiodo, line 78: Moved the logic for this filtering into the controllerRegistration. It has all the other logic for filtering, so figured this would make sense being there.

src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -213,6 +215,22 @@ export class ControllerRegistration implements IControllerRegistration {
         this.onDidChangeFilter();
     }
 
+    private async onDidRemoveUris(uris: string[]) {
```
rchiodo, line 218: This is the logic for filtering that moved from the controllerloader.

src/platform/msrCrypto/README.md
```diff
@@ -0,0 +1,7 @@
+This source comes from here:
+https://github.com/microsoft/MSR-JavaScript-Crypto
```
rchiodo, line 2: For some reason msrCrypto is not an npm package. So we have to copy it. And it's not a module, so it has to be included in the source.

DonJayamanne (approved):

PR #10679, opened 2022-07-03
Slimmer FileSystemNode class

DonJayamanne:
TLDR - the code is more portable to web when using a simpler (common IFileSystem), as the api is isomorphic.

Remove some of the unnecessary methods in IFileSystemNode and optimized the code.
* When using local files/folders, then use the node modules such as `fs-extra` instead of using VS Code API (as that's slower and unnecessary)
* Use of `Uris` instead of strings when dealing with files (consistency with how we deal with FS)


src/kernels/jupyter/launcher/notebookStarter.node.ts
```diff
@@ -291,7 +291,7 @@ export class NotebookStarter implements INotebookStarter {
         const args: string[] = [];
         // Check for a docker situation.
         try {
-            const cgroup = await this.fs.readLocalFile('/proc/self/cgroup').catch(() => '');
+            const cgroup = await this.fs.readFile(Uri.file('/proc/self/cgroup')).catch(() => '');
```
DonJayamanne, line 294: Whether it local or not is not a concern of the API, but its a concern of the implementation.
I.e. how we copy & read files should not be determined by the API, OOP can deal with the complications by having a custom implemementation of the IFileSystem interface.

Thus, in most places we should now be able to just use `IFileSystem` instead of `IFileSystemNode`

E.g. this class now just uses `IFileSystem` and not a custom interface.

src/platform/common/platform/fileSystem.ts
```diff
@@ -40,8 +40,8 @@ export class FileSystem implements IFileSystem {
 
     // URI-based filesystem functions
 
-    async copy(source: vscode.Uri, destination: vscode.Uri): Promise<void> {
-        await this.vscfs.copy(source, destination);
+    async copy(source: vscode.Uri, destination: vscode.Uri, options?: { overwrite: boolean }): Promise<void> {
+        await this.vscfs.copy(source, destination, options);
```
DonJayamanne, line 44: Both fs and VS Code support `options` and we were't passing them, but was hardcoded in the node implementation

src/platform/common/platform/types.node.ts
```diff
@@ -12,24 +12,8 @@ export type WriteStream = fs.WriteStream;
...
 export const IFileSystemNode = Symbol('IFileSystemNode');
 export interface IFileSystemNode extends IFileSystem {
-    areLocalPathsSame(path1: string, path2: string): boolean;
-    createLocalDirectory(path: string): Promise<void>;
-    /**
-     * Copies a local file/directory.
-     * @param {{ overwrite: boolean }} [options] Defaults to overwriting existing files/directories.
-     */
-    copyLocal(source: string, destination: string, options?: { overwrite: boolean }): Promise<void>;
-    deleteLocalFile(path: string): Promise<void>;
```
DonJayamanne, line 22: Thats a lot of unnecessary API that callers don't need to worry about.
In the custom implemetnation we check the scheme of the Uri and use the appropriate methods (i.e. when dealing with local files we use fs module else revert to using VS Code API)

Hence a slimmer interface and simpler to use.
Also optimized the code to always use fs modules instead of VS Code when dealing with local fs resources

src/platform/common/platform/fileSystem.node.ts
```diff
@@ -88,34 +66,70 @@ export class FileSystem extends FileSystemBase implements IFileSystemNode {
...
-        return arePathsSame(path1, path2);
+    async writeLocalFile(filename: string, text: string | Buffer): Promise<void> {
+        await fs.ensureDir(path.dirname(filename));
+        return fs.writeFile(filename, text);
     }
 
-    public async createLocalDirectory(path: string): Promise<void> {
-        await fs.ensureDir(path);
+    override async readFile(uri: Uri): Promise<string> {
+        if (isLocalFile(uri)) {
```
DonJayamanne, line 75: Optimized the code to always use fs modules instead of VS Code API when dealing with local fs resources

src/platform/common/application/extensions.node.ts
```diff
@@ -57,9 +57,9 @@ export class Extensions implements IExtensions {
                 let dirName = path.dirname(frame);
                 let last = frame;
                 while (dirName && dirName.length < last.length) {
-                    const possiblePackageJson = path.join(dirName, 'package.json');
-                    if (await this.fs.localFileExists(possiblePackageJson)) {
-                        const text = await this.fs.readLocalFile(possiblePackageJson);
+                    const possiblePackageJson = Uri.file(path.join(dirName, 'package.json'));
+                    if (await this.fs.exists(possiblePackageJson)) {
+                        const text = await this.fs.readFile(possiblePackageJson);
```
sadasant, line 62: Much nicer API!

sadasant (approved):
Excellent! I love what you're doing in this PR 👏

Issue #10080, opened 2022-05-20
Log parsing extension


rchiodo:
Logs in CI are rather large and difficult to find things.

One idea to alleviate this would be to provide an extension that would:

- Create an outline provider which provides an outline based on all the tests in the log
- Not sure if that automatically creates folding regions or not, but if not the extension would automatically create folding regions in the log
- Extension would detect .log file extension and then parse to see if it's a CI log

rebornix:
Since we are opening the logs in the editor, we can leverage the editor apis:

* Highlight the test result summary (how many fail and who fail), give them a red background, through decoration api
* Links to the test itself (which has call stacks) through link provider api

PR #10680, opened 2022-07-04
Improved test summary logging on CI

DonJayamanne:
Clean up of logging of test results

1. Must faster to view the test failures now
     - Today in web & desktop test failures we have to wait for github actions to load the logs & expand the output, this is basically instantaneous, as opposed to waiting for 15-20 seconds (**& even after that the logs displayed are useless**)
2. Colors in the output
3. Consistent outputs whether its web or desktop (the failures are displayed in a single job)
4. Generate an ipynb file with a list of failed tests, including the output & screenshots

# Old

# New (no you can see the failures & the details) in <1s as opposed to 15s

# Old
When you look at the logs,  you have no idea what output pertains to what test execution.
Hence when looking at a test failure, you have to:
* Scan through the logs & look for the start & end of the test & extract the text file contents
* You need to download all screenshots & look for the file

# New
Note: I find this very useful, as I always scrape the output by looking for the start and end of the tests (this is why we have `traceInfo(`Start Test ${this.currentTest?.title}`);` in the test files, so we can capture the logs that are specific to a test run. This notebook alone will save me a lot of time when looking at the test failures.

These are attached to the job as artificats.

* All failures in one ipynb
* Each failure is a separate cell with its own output & screenshots

## We have colorized outputs, which makes it much easier to find the errors (messages we're looking for in the logs)

I believe this PR also fixes https://github.com/microsoft/vscode-jupyter/issues/10080
If not, we can easily create a seprate ipynb file that contains the output from all tests (today the ipynb created only contains the failed test & their corresponding output & screenshots)


.github/workflows/build-test.yml
```diff
@@ -683,10 +683,6 @@ jobs:
         id: test_notebookWithoutPythonExt_vscode
         if: matrix.python == 'noPython' && matrix.os != 'windows-latest'
 
-      - name: Check test run
```
DonJayamanne, line 686: Not required anymore

build/webTestReporter.js
```diff
@@ -6,11 +6,17 @@ const path = require('path');
...
 const { EventEmitter } = require('events');
+const colors = require('colors');
+const core = require('@actions/core');
+const hashjs = require('hash.js');
+const glob = require('glob');
+const { ExtensionRootDir } = require('./constants');
 
 const settingsFile = path.join(__dirname, '..', 'src', 'test', 'datascience', '.vscode', 'settings.json');
-const webTestSummaryJsonFile = path.join(__dirname, '..', 'webtest.json');
-const webTestSummaryFile = path.join(__dirname, '..', 'webtest.txt');
```
DonJayamanne, line 13: Not required

build/webTestReporter.js
```diff
@@ -73,24 +78,131 @@ exports.startReportServer = async function () {
...
+                reportWriter.stats = { ...output.stats };
+                Object.assign(runner.stats, output.stats);
             }
             if (output.event === 'fail') {
                 reportWriter.failures.push(output);
             }
-            eventEmitter.emit(output.event, Object.assign({}, output));
+            runner.emit(output.event, output, output.err);
+
+            switch (output.event) {
```
DonJayamanne, line 112: Generate an ipynb file with:
* failed tests
* logs for each of the fail tests
* screenshots of the failed tests

build/webpack/webpack.extension.web.config.js
```diff
@@ -124,7 +124,8 @@ const config = {
             // for the list of Node.js core module polyfills.
             assert: require.resolve('assert'),
             buffer: require.resolve('buffer'),
-            stream: require.resolve('stream-browserify')
+            stream: require.resolve('stream-browserify'),
+            os: require.resolve('os-browserify')
```
DonJayamanne, line 128: required for mocha test reporter

src/test/client/api.vscode.test.ts
```diff
@@ -52,14 +52,14 @@ suite('3rd Party Kernel Service API', function () {
             await startJupyterServer();
             traceInfo(`Start Test (completed) ${this.currentTest?.title}`);
         } catch (e) {
-            await captureScreenShot(this.currentTest?.title || 'unknown');
+            await captureScreenShot(this);
```
DonJayamanne, line 55: This was repeated everywhere, now its simpler we just call `captureScreenShot`, 
and has the added benefit of being able to predict the file names for test failures, so we can attach them to the ipynb file.

build/webTestReporter.js
```diff
@@ -6,11 +6,17 @@ const path = require('path');
...
+const colors = require('colors');
+const core = require('@actions/core');
+const hashjs = require('hash.js');
+const glob = require('glob');
+const { ExtensionRootDir } = require('./constants');
 
 const settingsFile = path.join(__dirname, '..', 'src', 'test', 'datascience', '.vscode', 'settings.json');
-const webTestSummaryJsonFile = path.join(__dirname, '..', 'webtest.json');
-const webTestSummaryFile = path.join(__dirname, '..', 'webtest.txt');
+const webTestSummaryJsonFile = path.join(__dirname, '..', 'testresults.json');
```
DonJayamanne, line 18: new name, all tests will now generate this file (not just web tests).
I.e. even regular (desktop) tests will generate this file, so we can dump the test results in a separate job as we did in web test, thus producing a cleaner, simpler  & consistent output for all tests.

src/test/web/customReporter.ts
```diff
@@ -147,32 +244,61 @@ function CustomReporter(this: any, runner: mochaTypes.Runner, options: mochaType
             });
         })
         .on(constants.EVENT_TEST_FAIL, (test: mochaTypes.Test, err: any) => {
+            const consoleOutput = consoleHijacker.release();
             writeReportProgress({
                 event: constants.EVENT_TEST_FAIL,
                 title: test.title,
                 err: formatException(err),
+                titlePath: test.titlePath(),
```
DonJayamanne, line 252: Add some more data required by the test writers (e.g. when using xunit, json report writer or the like, these are required)

build/webTestReporter.js
```diff
@@ -73,24 +78,131 @@ exports.startReportServer = async function () {
...
+                        metadata: {
+                            collapsed: true
+                        },
+                        source: `#${output.title}`,
+                        execution_count: ++executionCount,
+                        outputs: [assertionError, consoleOutput, ...screenshots]
+                    });
+                    break;
+                }
+            }
```
sadasant, line 196: Could it be possible to have a case that doesn't fit any of the other cases? I'm thinking we could add a simple "default" case to throw (in case of an unexpected behavior or condition).

sadasant (approved):
This is a beautiful PR for me! The improvements are excellent. Thank you!

build/webTestReporter.js
DonJayamanne: No, for other cases we don't need to do anything, hence throwing an exception is wrong
basically this switch is not handling all possible cases, but only handling what we need.

sadasant: oooh perfect! thank you 👍

PR #10686, opened 2022-07-04
Update create new nb instruction, remove old ipynb

DonJayamanne:
The old file `Notebooks intro.ipynb` is not used anywhere & has old UI


sadasant (approved):
Love the deletions!

PR #10689, opened 2022-07-05
Read file contents instead of relying on webpack to inject contents into bundle

### LINKED ISSUE #10094: Using `require` to inject file contents into JS files causes issues with unit tests
We cannot write unit tests for some files if using code as follows:
```typescript
const addRunCellHook = require('../../pythonFiles/vscode_datascience_helpers/kernel/addRunCellHook.py');
```

* We might want to consider removing such webpack specific code from our code and use some other way to inject the file contents. 
* Or we need way another way to write unit tests for such files.
* Or we could have a simple provider that provides the contents of these files.
* We end up duplicating the files (its in the source code as well as vsix)

DonJayamanne:
Fixes #10094


sadasant (approved):
Much clearer to think through!

PR #10699, opened 2022-07-05
Include jquery as npm package

DonJayamanne:
Gets rid of the compiler warning we get about jquery not being in the src directory


build/webpack/webpack.extension.node.config.js
```diff
@@ -120,6 +120,14 @@ const config = {
                 }
             ]
         }),
+        new copyWebpackPlugin({
+            patterns: [
+                {
+                    from: './node_modules/jquery/dist/jquery.min.js',
+                    to: './node_modules/jquery/dist/jquery.min.js'
```
amunger, line 127: source and destination are the same?

DonJayamanne: Destination is rooted at `out`, 
if you look 5 lines above, we have a similar entry for `pdfkit`

amunger (approved):

PR #10713, opened 2022-07-06
re-enable export test

amunger:
Fixes https://github.com/microsoft/vscode-jupyter/issues/10649


amunger:
seems to be passing now. From what I could tell, it was failing every run previously

sadasant (approved):
Not a lot to review! What is here looks good to me. I'd let CI be the judge 😁✅

PR #10715, opened 2022-07-06
Better CI Job labels

DonJayamanne:
This basically makes it easy to read the Job labels & figure out which one is web and which runs against pre-release bits

# Before
Screen Shot 2022-07-07 at 09 30 36

# Now (with changes)
Screen Shot 2022-07-07 at 09 33 28


sadasant (approved):
I feel I'm not the best judge here, but this causes no harm 👍✅

PR #10724, opened 2022-07-07
Using monospace in the values of the variable viewer

### LINKED ISSUE #10223: Consider using a monospace font in the variable viewer
Testing #10170

We typically use monospace fonts for things that represent code or their values

Screen Shot 2022-06-01 at 7 09 10 am

See the debug viewlet for example:

Screen Shot 2022-06-01 at 7 11 12 am

Screen Shot 2022-06-01 at 7 11 43 am

sadasant:
This PR changes the font family of the cells in the table of the variable viewer to use monospace. Looks as follows:

Screen Shot 2022-07-07 at 5 00 14 PM

You can compare that screenshot with the screenshot available in the related issue.
Fixes #10223

### DIFF
--- a/src/webviews/webview-side/interactive-common/variableExplorerGrid.less
+++ b/src/webviews/webview-side/interactive-common/variableExplorerGrid.less
@@ -70,6 +70,7 @@
     color: var(--vscode-editor-foreground);
     border-style: none;
     padding: 0 2px;
+    font-family: monospace;
 }
 
 #variable-explorer-data-grid .react-grid-Cell:hover {


src/webviews/webview-side/interactive-common/variableExplorerGrid.less
```diff
@@ -70,6 +70,7 @@
     color: var(--vscode-editor-foreground);
     border-style: none;
     padding: 0 2px;
+    font-family: monospace;
```
DonJayamanne, line 73: Perhaps best disussed at standup or with UX, I'm not sure we want this or whether this is the right thing, 
this removes ability for users to costomize the fonts, however this is also required to fix the issue. Hence understand the need for something.

I.e. in effect this removes ability for users to use their own themes (fonts).

DonJayamanne (commented):
Perhaps we can discuss this in standup, 
@IanMatthewHuff @rebornix @roblourens  thoughts

src/webviews/webview-side/interactive-common/variableExplorerGrid.less
sadasant: (Replied here by mistake. I deleted my reply)

IanMatthewHuff (changes requested):
I believe that we want `debug.console.fontFamily` per my other comment.

sadasant:
`--debug-console-font-family` makes it look a bit too times-new-romany. Not sure what to do next.
Screen Shot 2022-07-08 at 4 18 59 PM

IanMatthewHuff:
So we only have access to some variables in a webview I don't believe the --debug-console-font-family is one of them. I was just referring to the VS Code setting, but I'm not 100% sure that's the correct setting. Looking at VS Code it's using 'SF Mono' via `--monaco-monospace-font`.

IanMatthewHuff:
> So we only have access to some variables in a webview I don't believe the --debug-console-font-family is one of them. I was just referring to the VS Code setting, but I'm not 100% sure that's the correct setting. Looking at VS Code it's using 'SF Mono' via `--monaco-monospace-font`.

Bleh, yeah, looking at the setting for debug.console.fontFamily it just pulls `default` so we can't load that setting from the extension side and send it to the webview.

sadasant:
@IanMatthewHuff thank you, Ian!

DonJayamanne:
> that we want debug.console.fontFamily per my other comment

id suggest filling a feature request and or subjugating a pr upstream.
i know @rebornix exposed some css variables for notebook output renderers too work as well. we could do the same here. else using settings feels more work

sadasant:
@DonJayamanne , @IanMatthewHuff I'll close this PR and add a comment on the issue to add a feature request @ upstream.

PR #10726, opened 2022-07-07
Load widget scripts from all known Jupyter paths

### LINKED ISSUE #10722: Support all Jupyter Paths when loading static resources
We recently updated IPyWidgets to load static resources from a couple of different places,
however Jupyter has a pre-defined & documented list

Today we load static resources from just 2 places, however jupyter has more than that.
E.g. if you look at the isntallation of pythreejs, if user installs widgets into `--system`, then they will not work in VS Code.
We don't load widgets from the `--system` location.

**Todo:**
* Load static resources from all directories that Jupyter loads from
* Use the same priority that Jupyter uses

**Ref**
* Documentation of the paths can be found here https://docs.jupyter.org/en/latest/use/jupyter-directories.html#data-files
* Source for priority & paths can be found in `jupyter_path` function in `site-packages/jupyter_core/paths.py`

DonJayamanne:
Fixes #10722

### DIFF
--- a/src/kernels/helpers.ts
+++ b/src/kernels/helpers.ts
@@ -1079,9 +1079,7 @@ export function getKernelId(spec: IJupyterKernelSpec, interpreter?: PythonEnviro
     }
     const prefixForRemoteKernels = serverId ? `${serverId}.` : '';
     const specPath = getFilePath(
-        getNormalizedInterpreterPath(
-            spec.interpreterPath ? Uri.file(spec.interpreterPath) : Uri.file(spec.executable)
-        )
+        getNormalizedInterpreterPath(spec.interpreterPath ? Uri.file(spec.interpreterPath) : Uri.file(spec.executable))
     );
     const interpreterPath = getFilePath(getNormalizedInterpreterPath(interpreter?.uri)) || '';
     return `${prefixForRemoteKernels}${


src/kernels/raw/finder/jupyterPaths.node.ts
```diff
@@ -174,64 +300,73 @@ export class JupyterPaths {
...
-                    return [];
-                }
-                const jupyterPathVars = vars.JUPYTER_PATH
-                    ? vars.JUPYTER_PATH.split(path.delimiter).map((jupyterPath) => {
-                          return path.join(jupyterPath, 'kernels');
-                      })
-                    : [];
-
-                if (jupyterPathVars.length > 0) {
-                    jupyterPathVars.forEach(async (jupyterPath) => {
```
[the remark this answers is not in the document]
DonJayamanne: Added a test for this as well

src/platform/common/platform/platformService.node.ts
```diff
@@ -9,13 +9,14 @@ import { Uri } from 'vscode';
...
 import { getOSType, OSType } from '../utils/platform';
+import { getUserHomeDir } from '../utils/platform.node';
 import { parseVersion } from '../utils/version.node';
 import { NON_WINDOWS_PATH_VARIABLE_NAME, WINDOWS_PATH_VARIABLE_NAME } from './constants.node';
 import { IPlatformService } from './types';
 
 @injectable()
 export class PlatformService implements IPlatformService {
-    private readonly _homeDir = Uri.file(os.homedir());
+    private readonly _homeDir = getUserHomeDir() || Uri.file(os.homedir());
```
DonJayamanne, line 19: Modified to ensure we use the same function everywhere.

src/kernels/ipywidgets/remoteIPyWidgetScriptManager.ts
```diff
@@ -49,36 +51,22 @@ export class RemoteIPyWidgetScriptManager extends BaseIPyWidgetScriptManager imp
         }
         return this.widgetEntryPointsPromise;
     }
+    private getCodeToExecute() {
+        if (!this.code) {
+            this.code = this.fs.readFile(
+                Uri.joinPath(this.context.extensionUri, 'pythonFiles', 'printJupyWidgetEntryPoints.py')
```
DonJayamanne, line 57: Moved code to a file, easier to manage.

src/platform/vscode-path/utils.ts
```diff
@@ -1,11 +1,5 @@
 // Miscellaneous functions from other spots in VS code
 
-import { Uri } from 'vscode';
-
-export function fsPathToUri(path: string | undefined) {
-    return path ? Uri.file(path) : undefined;
```
DonJayamanne, line 6: Removed this as this just compilcates the code unencessarily by adding nullables to the return values, even when the path is not undefined.

I.e. if we call `fsPathToUri(<some string value>)` the return value is still nullable, when this is impossible.
This unnecessary nullable complicates the code, hence removing it.

PR #10728, opened 2022-07-07
Simply script uri converter

DonJayamanne:


src/kernels/ipywidgets/scriptUriConverter.ts
```diff
@@ -1,122 +1,22 @@
...
      */
-    public async asWebviewUri(localResource: Uri): Promise<Uri> {
-        if (this.isWeb) {
-            // We cannot create folders on the web, return this as is.
-            return localResource;
-        }
-        // Make a copy of the local file if not already in the correct location
-        if (!this.isInScriptPath(localResource)) {
-            const key = getComparisonKey(localResource);
-            if (!this.resourcesMappedToExtensionFolder.has(key)) {
```
DonJayamanne, line 47: This is no longer required. We copy the entire nbextensions folders now, instead of copying a single file

src/kernels/ipywidgets/types.ts
```diff
@@ -80,16 +80,10 @@ export interface IWidgetScriptSourceProviderFactory {
     ): IWidgetScriptSourceProvider[];
 }
 
-export const ILocalResourceUriConverter = Symbol('ILocalResourceUriConverter');
-
 /**
  * Given a local resource this will convert the Uri into a form such that it can be used in a WebView.
  */
 export interface ILocalResourceUriConverter {
```
sadasant, line 86: Would this still be needed?

DonJayamanne: Yes, we don't want to hardcode the class everywhere, instead just use the interafce.

sadasant (approved):

src/kernels/ipywidgets/types.ts
```diff
@@ -102,15 +96,6 @@ export interface ILocalResourceUriConverter {
      * ```
      */
     asWebviewUri(localResource: Uri): Promise<Uri>;
-    /**
-     * The converter will post an event when it needs to convert the webview URI
-     */
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    requestUri: Event<Uri>;
```
DonJayamanne, line 109: Removed this event.
I personally think its simpler to read the code by firing a method to get information as opposed to firing an event to ask for information.
Events - fire and forget, and method (callbacks) mean we need a result.

IanMatthewHuff (approved):

PR #10731, opened 2022-07-07
Remove releasing extension from PR yml

DonJayamanne:


.github/actions/build-vsix/action.yml
```diff
@@ -1,11 +1,6 @@
 name: 'Build VSIX'
 description: "Build the extension's VSIX"
 
-inputs:
-  package_command:
-     description: 'Package.json package script'
-     required: false
-     default: 'package'
```
DonJayamanne, line 8: We don't need any of these anymore as we now use the azure pipelines for releases.

.github/workflows/build-test.yml
```diff
@@ -69,68 +59,10 @@ jobs:
...
-        run: |
-          echo "package_command=package" >> $GITHUB_ENV
-
-      - name: pr
-        if: github.event_name == 'pull_request'
-        run: |
-          echo "vsix_name=ms-toolsai-jupyter-insiders.vsix" >> $GITHUB_ENV
-          echo "test_matrix_os=[\"ubuntu-latest\", \"windows-latest\"]" >> $GITHUB_ENV
-
-      - name: pr release push
```
DonJayamanne, line 95: We don't need any of these anymore as we now use the azure pipelines for releases.

src/test/standardTest.node.ts
```diff
@@ -58,7 +58,7 @@ async function installPythonExtension(vscodeExecutablePath: string) {
     // Pick python extension to use based on environment variable. Insiders can be flakey so
     // have the capability to turn it off/on.
     const pythonVSIX =
-        process.env.VSC_JUPTYER_PYTHON_EXTENSION_VERSION === 'insiders'
+        process.env.VSC_JUPYTER_PYTHON_EXTENSION_VERSION === 'insiders'
```
sadasant, line 61: I'm not sure why this is labeled as a change 😅 I can't see the difference 😵

sadasant (approved):

src/test/standardTest.node.ts
DonJayamanne: JUPYTER vs JUPTYER , the `TY` are swapped.
I'm 200% certain I introduced this typo

PR #10732, opened 2022-07-07
Ensures string variables appear within single quotes in the variable viewer

### LINKED ISSUE #10225: Variables viewer: Add quotes around string values
Testing #10170

Without quotes the values look like variables, not strings

Screen Shot 2022-06-01 at 7 14 55 am

sadasant:
This PR changes the way strings are rendered in the variable viewer to wrap them in single quotes. Looks as follows:

Screen Shot 2022-07-07 at 7 51 21 PM

I'm achieving this by detecting the string type on the component that renders the string, and wrapping the value in quotes.

**IMPORTANT:** For some reason, all of the variable values are received with an empty space up front.

This PR aims to:
Fix #10225

Feedback appreciated!

### DIFF
--- a/src/kernels/variables/kernelVariables.ts
+++ b/src/kernels/variables/kernelVariables.ts
@@ -24,7 +24,7 @@ import {
 // kernels will add the ansi encoding.
 const TypeRegex = /.*?\[.*?;31mType:.*?\[0m\s+(\w+)/;
 const ValueRegex = /.*?\[.*?;31mValue:.*?\[0m\s+(.*)/;
-const StringFormRegex = /.*?\[.*?;31mString form:.*?\[0m\s*?([\s\S]+?)\n(.*\[.*;31m?)/;
+const StringFormRegex = /.*?\[.*?;31mString form:.*?\[0m\s+?([\s\S]+?)\n(.*\[.*;31m?)/;
 const DocStringRegex = /.*?\[.*?;31mDocstring:.*?\[0m\s+(.*)/;
 const CountRegex = /.*?\[.*?;31mLength:.*?\[0m\s+(.*)/;
 const ShapeRegex = /^\s+\[(\d+) rows x (\d+) columns\]/m;
--- a/src/webviews/webview-side/interactive-common/variableExplorer.tsx
+++ b/src/webviews/webview-side/interactive-common/variableExplorer.tsx
@@ -348,6 +348,14 @@ export class VariableExplorer extends React.Component<IVariableExplorerProps, IV
                 } else if (variable.count) {
                     newSize = variable.count.toString();
                 }
+                let value = variable.value;
+                if (variable.type === 'str' && variable.value) {
+                    if (variable.value[0] === ' ') {
+                        value = `'${variable.value.slice(1)}'`;
+                    } else {
+                        value = `'${variable.value}'`;
+                    }
+                }
                 return {
                     buttons: {
                         name: variable.name,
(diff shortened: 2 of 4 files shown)


src/webviews/webview-side/interactive-common/variableExplorer.tsx
```diff
@@ -348,6 +348,14 @@ export class VariableExplorer extends React.Component<IVariableExplorerProps, IV
...
                     newSize = variable.count.toString();
                 }
+                let value = variable.value;
+                if (variable.type === 'str' && variable.value) {
+                    if (variable.value[0] === ' ') {
+                        value = `'${variable.value.slice(1)}'`;
+                    } else {
+                        value = `'${variable.value}'`;
+                    }
+                }
```
sadasant, line 358: What part of the code would add an empty space at the beginning of all of the received values?

src/webviews/webview-side/interactive-common/variableExplorer.tsx
```diff
@@ -348,6 +348,14 @@ export class VariableExplorer extends React.Component<IVariableExplorerProps, IV
                 } else if (variable.count) {
                     newSize = variable.count.toString();
                 }
+                let value = variable.value;
+                if (variable.type === 'str' && variable.value) {
+                    if (variable.value[0] === ' ') {
```
DonJayamanne, line 353: I would rather we found where this space was added and removed that there instead of patching this here.

Else tomorrow if we remove this leading space from the other end (which evere code/layer adds that), then we'd end up stripping the first empty character (here) from strings which would be wrong.
Unless someone remembers to remove this patchy code, which I think is less likely to happen.

DonJayamanne (changes requested):
Please remove the leading space from the code that adds it instead of removing it at the UI layer

src/webviews/webview-side/interactive-common/variableExplorer.tsx
sadasant: Done here: https://github.com/microsoft/vscode-jupyter/pull/10732/commits/767759d3f2ad6c05d8bd75264315c4e1ba8083b5

src/kernels/ipywidgets/localIPyWidgetScriptManager.node.ts
```diff
@@ -48,7 +49,17 @@ export class LocalIPyWidgetScriptManager extends BaseIPyWidgetScriptManager impl
...
+        this.nbExtensionsParentPath = undefined;
+        super.onKernelRestarted();
+    }
     protected async getNbExtensionsParentPath(): Promise<Uri | undefined> {
+        if (!this.nbExtensionsParentPath) {
+            this.nbExtensionsParentPath = this.getNbExtensionsParentPathImpl();
+        }
+        return this.nbExtensionsParentPath;
+    }
+    private async getNbExtensionsParentPathImpl(): Promise<Uri | undefined> {
```
sadasant, line 62: Not sure how this got here. Cleaning this PR

src/kernels/variables/kernelVariables.ts
```diff
@@ -24,7 +24,7 @@ import {
 // kernels will add the ansi encoding.
 const TypeRegex = /.*?\[.*?;31mType:.*?\[0m\s+(\w+)/;
 const ValueRegex = /.*?\[.*?;31mValue:.*?\[0m\s+(.*)/;
-const StringFormRegex = /.*?\[.*?;31mString form:.*?\[0m\s*?([\s\S]+?)\n(.*\[.*;31m?)/;
+const StringFormRegex = /.*?\[.*?;31mString form:.*?\[0m\s+?([\s\S]+?)\n(.*\[.*;31m?)/;
```
sadasant, line 27: This solves the unnecessary space problem.

IanMatthewHuff (approved):

DonJayamanne (approved):

PR #10744, opened 2022-07-08
Variable viewer high contrast hover color fix

sadasant:
The rows in the table of the variable viewer have a small bug in high contrast mode in which the values disappear completely when the row is hovered. This PR aims to fix that by:

- Removing some lines that seemed unnecessary to me. **Please let me know if I'm missing something.**
- Used `--vscode-editor-foreground` (which is the inverse of the high contrast background color) instead of `--vscode-editor-selectionForeground` (which is the same as the high contrast background color).
- Used opacity to make a color that could be distinct from the default foreground color, but that could remain coherent independently of the background color.

Here's a short video demonstrating the problem:

https://user-images.githubusercontent.com/417016/178050421-b199b7ee-84bb-437a-ae11-1e2c38d14eba.mov

Here's a short video demonstrating how the solution looks:

https://user-images.githubusercontent.com/417016/178050020-0c4735d2-ab9c-422d-9410-d2d0b1bef46d.mov

Feedback appreciated 👍

### DIFF
--- a/src/webviews/webview-side/interactive-common/variableExplorerCellFormatter.css
+++ b/src/webviews/webview-side/interactive-common/variableExplorerCellFormatter.css
@@ -14,19 +14,11 @@
     color: var(--vscode-editor-foreground);
 }
 
+body.vscode-high-contrast .react-grid-Row:hover {
+    filter: opacity(70%);
+}
+
 /* High contrast uses an inverse for selection highlights, so just use foreground when hovering */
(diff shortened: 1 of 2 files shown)


rchiodo:
What does it look like without high contrast (before and after)?

sadasant:
@rchiodo it looks the same before and after (since this code did not change any of the non-high-contrast-styles). It looks good (which is why it has never been a problem).

I wanted to record a video, but I'm upgrading `brew` and that is making my computer unusable.

IanMatthewHuff (approved):

PR #10747, opened 2022-07-08
Adding more padding to each cell in the variable viewer

### LINKED ISSUE #10206: hovering over sort indicator shows column resize cursor
Testing #10170

Hovering over the right half of the sort indicator ("triangle") shows the column resize cursor and when clicking there the column is actually resized.

CleanShot 2022-06-01 at 09 54 55

sadasant:
As the issue #10206 says:

> Hovering over the right half of the sort indicator ("triangle") shows the column resize cursor and when clicking there the column is actually resized.

This PR solves that by increasing the padding on the cells. Rather than fixing this only in the headers, I decided to increase the horizontal padding on all the variable viewer cells in order to keep them looking more harmonious together.

Here's how it looks:

https://user-images.githubusercontent.com/417016/178078880-2de262fd-4133-4406-be73-535f90a34f60.mov

Let me know if you know of a better approach!

Fixes #10206

### DIFF
--- a/src/webviews/webview-side/interactive-common/variableExplorerGrid.less
+++ b/src/webviews/webview-side/interactive-common/variableExplorerGrid.less
@@ -40,7 +40,7 @@
     padding: 2px;
     font-family: var(--vscode-font-family);
     border-right: 1px solid var(--vscode-menu-border);
-    padding-top: 4px;
+    padding: 4px 8px;
 }
 
 .react-grid-Header {
@@ -69,7 +69,7 @@
     background-color: transparent;
     color: var(--vscode-editor-foreground);
     border-style: none;
-    padding: 0 2px;
+    padding: 0 8px;
 }
 
 #variable-explorer-data-grid .react-grid-Cell:hover {


src/webviews/webview-side/interactive-common/variableExplorerGrid.less
```diff
@@ -69,7 +69,7 @@
     background-color: transparent;
     color: var(--vscode-editor-foreground);
     border-style: none;
-    padding: 0 2px;
+    padding: 0 8px;
```
DonJayamanne, line 72: i think it's better not to hardcode 8px, instead hardcode top and bottom padding.
this way, if the 8px changes in the library or else where we don't need you worry about that... right now 8px is a magic number and it's not known why we use 8px.

sadasant: I think I like what I'm reading, but I don't fully understand!

> instead hardcode top and bottom padding

Top and bottom? Not sure how that relates to the horizontal padding.

DonJayamanne: what i meant is, change the code to 
```css
padding-top: 0
padding-bottom: 0
```

this way the original padding of 8px is not changed.
from what i know 8px is setup by the original library, not our code, hence if the library charges tomorrow to 4px this code will need to change yet again.
ie im merely suggesting a way to future proof this and inherit the css values instead of hardcoding it.
again totally optional, which is why i didn't block this pr

sadasant: I like what you're thinking! I just didn't notice the 8px were already used by the component of my changes 👀 trying your suggestion now.

sadasant: Looks good to me! https://github.com/microsoft/vscode-jupyter/pull/10747/commits/5d5d5727a405cf84adf0f58a3952ebbb081f0d07

https://user-images.githubusercontent.com/417016/178082998-2abd5ed4-4284-4514-972d-6af4bfb6f2ed.mov

src/webviews/webview-side/interactive-common/variableExplorerGrid.less
```diff
@@ -40,7 +40,7 @@
     padding: 2px;
     font-family: var(--vscode-font-family);
     border-right: 1px solid var(--vscode-menu-border);
-    padding-top: 4px;
+    padding: 4px 8px;
```
DonJayamanne, line 43: Can we do the same here, just add padding-top and padding-bottom

sadasant: Oh, my bad! I had forgotten I changed it in two places. On my way!

sadasant: Sorry, here it doesn't work!

sadasant: Oh I see why. Disregard!

sadasant: Fixed here: https://github.com/microsoft/vscode-jupyter/pull/10747/commits/0fb043be585c47f8aa30fc2bb7b69eb968b3c9ce

DonJayamanne (approved):

IanMatthewHuff (approved):

PR #10756, opened 2022-07-11
Validate package.nls files against a schema

DonJayamanne:
Part of https://github.com/microsoft/vscode-jupyter/issues/10729

The drawback with having the Azure pipeline fail is the fact that its not monitored and its too late in the process for something to fail (PR verifications should not allow any broken bits to get into the repo)

**This this ensures we don't break the release builds again via a PR.**
E.g. if someone makes a typo again in the package.nls file this PR will ensure its blocked.

We already have some validation for the package.nls files as we've broken builds due to faulty JSON files, looks like we need another layer of validation, hence added that.


sadasant (approved):
Nice!

Issue #10752, opened 2022-07-10
Restart & Interrupt kernel buttons are missing when using the Chinese Simplified Language Pack 中文没有重启内核


morning-start:
Issue Type: <b>Bug</b>

使用Chinese简体插件之后，notebook的restart就没了

Extension version: 2022.6.1001902341
VS Code version: Code 1.69.0 (92d25e35d9bf1a6b16f7d0758f25d48ace11e5b9, 2022-07-07T05:28:36.503Z)
OS version: Windows_NT x64 10.0.19043
Restricted Mode: No

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|AMD Ryzen 5 3500U with Radeon Vega Mobile Gfx   (8 x 2096)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: disabled_off<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_renderer: enabled_on<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|9.94GB (1.01GB free)|
|Process Argv|--crash-reporter-id cf0d668d-4329-46d7-829b-6dc58f8a1ba2 --crash-reporter-id cf0d668d-4329-46d7-829b-6dc58f8a1ba2|
|Screen Reader|no|
|VM|50%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368cf:30146710
vsreu685:30147344
python383:30185418
vspor879:30202332
vspor708:30202333
vspor363:30204092
vswsl492:30256859
vstes627:30244334
vslsvsres303:30308271
pythonvspyl392:30443607
vserr242:30382549
pythontb:30283811
vsjup518:30340749
pythonvspyt551cf:30345471
pythonptprofiler:30281270
vsdfh931cf:30280410
vshan820:30294714
vstes263:30335439
vscorecescf:30445987
pythondataviewer:30285071
vscod805:30301674
binariesv615:30325510
bridge0708:30335490
bridge0723:30353136
vsaa593cf:30376535
vsc1dst:30438360
pythonvs932:30410667
wslgetstarted:30449410
vscscmwlcmt:30465135
[...]

PR #10763, opened 2022-07-11
Undoing command localizations

sadasant:
At the time I did this PR I assumed that, given that the other aspects of localization were a success, my changes would not have any drawback. I was wrong! This PR #10712 introduces a serious issue: the lack of translations seems to be preventing the extension to load properly, causing the restart and interrupt buttons on the global toolbar to disappear (probably among other things).

This issue is related: #10752

How does this look? What else should we do?
Feedback appreciated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -272,19 +272,19 @@
         "commands": [
             {
                 "command": "dataScience.ClearCache",
-                "title": "%jupyter.command.dataScience.clearCache.title%",
+                "title": "Clear Cache",
                 "category": "Jupyter (Dev)",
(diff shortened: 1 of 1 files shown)


sadasant:
Some things we know:

- Switching to version 5.100 does fix the problem.
- The localization changes that I did on PR #10712 are not part of the released 6.100 version of our extension.
- The package.nls.es.json in 6.100 has nothing I can identify as wrong or broken.

That means, at the very least, that reverting my PR won’t do anything.

PR #10783, opened 2022-07-12
Fix problem when restarting right after debugging

### LINKED ISSUE #10741: No debugger available after restarting kernel
### Applies To

- [X] Notebooks (.ipynb files)

### What happened?

-> Debug a cell
-> Stop debugger with disconnect (didn't find any other way to stop debugger)

2022-07-08 10_59_48-Window

The cell dropdown menu with debug option disappears. Need to run the cell to get que dropdown menu back.

-> If restar kernel

2022-07-08 10_58_13-Window

These toasts appears. Sometimes it does not affect anything, but sometimes you need to close and reopen VSCode to be able to run debugger again.

### VS Code Version

1.69

### Jupyter Extension Version

v2022.7.1001902022

### Jupyter logs

_No response_

Python 3.8.6

_No response_

_No response_

Local

rchiodo:
Fixes #10741 

I don't think this was an actual regression, but rather a papercut. Super easy fix though.

### DIFF
--- a/src/kernels/debugger/kernelDebugAdapterBase.ts
+++ b/src/kernels/debugger/kernelDebugAdapterBase.ts
@@ -201,11 +201,11 @@ export abstract class KernelDebugAdapterBase implements DebugAdapter, IKernelDeb
 
     public async disconnect() {
         if (!this.disconnected) {
+            this.disconnected = true;
             if (this.debugService.activeDebugSession === this.session) {
                 await this.session.customRequest('disconnect', { restart: false });
             }
             this.endSession.fire(this.session);
-            this.disconnected = true;
             this.kernel?.removeEventHook(this.kernelEventHook);
         }
     }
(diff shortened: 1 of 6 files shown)


src/kernels/debugger/kernelDebugAdapterBase.ts
```diff
@@ -189,10 +200,14 @@ export abstract class KernelDebugAdapterBase implements DebugAdapter, IKernelDeb
...
-        await this.session.customRequest('disconnect', { restart: false });
-        this.endSession.fire(this.session);
-        this.disconnected = true;
-        this.kernel?.removeEventHook(this.kernelEventHook);
+        if (!this.disconnected) {
+            if (this.debugService.activeDebugSession === this.session) {
+                await this.session.customRequest('disconnect', { restart: false });
+            }
+            this.endSession.fire(this.session);
+            this.disconnected = true;
```
DonJayamanne, line 208: I think we should move this property initialization to the very top (just after checking it), else this method `disconnect` can be called multiple times and we could end up sending multiple `disconnect` requests

DonJayamanne (approved):
One suggested change to prevent duplicate `disconnect` requests being sent out

src/kernels/debugger/kernelDebugAdapterBase.ts
rchiodo: Ah good point.

PR #10784, opened 2022-07-12
Fix screen capture to work on windows

rchiodo:
Screen capture file names are too long. I can't open them on windows. 

This should achieve the same result (unique file names)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -223,7 +223,7 @@
             ],
             "env": {
                 "VSC_JUPYTER_FORCE_LOGGING": "1",
-                "VSC_JUPYTER_CI_TEST_GREP": "", // Leave as `VSCode Notebook` to run only Notebook tests.
+                "VSC_JUPYTER_CI_TEST_GREP": "Ensure ipykernel install prompt is displayed every time you try to run a cell in an Interactive Window and you can switch kernels", // Leave as `VSCode Notebook` to run only Notebook tests.
                 "VSC_JUPYTER_CI_TEST_INVERT_GREP": "", // Initialize this to invert the grep (exclude tests with value defined in grep).
                 "CI_PYTHON_PATH": "", // Update with path to real python interpereter used for testing.
                 "VSC_JUPYTER_CI_RUN_NON_PYTHON_NB_TEST": "", // Initialize this to run tests again Julia & other kernels.
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -223,7 +223,7 @@
             ],
             "env": {
                 "VSC_JUPYTER_FORCE_LOGGING": "1",
-                "VSC_JUPYTER_CI_TEST_GREP": "Ensure ipykernel install prompt is displayed every time you try to run a cell in an Interactive Window and you can switch kernels", // Leave as `VSCode Notebook` to run only Notebook tests.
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -223,7 +223,7 @@
             ],
             "env": {
                 "VSC_JUPYTER_FORCE_LOGGING": "1",
-                "VSC_JUPYTER_CI_TEST_GREP": "", // Leave as `VSCode Notebook` to run only Notebook tests.
+                "VSC_JUPYTER_CI_TEST_GREP": "Ensure ipykernel install prompt is displayed every time you try to run a cell in an Interactive Window and you can switch kernels", // Leave as `VSCode Notebook` to run only Notebook tests.
```
rchiodo, line 226: Whoops

src/test/common.node.ts
```diff
@@ -322,21 +323,15 @@ export async function captureScreenShot(contextOrFileName: string | Mocha.Contex
...
-            : hashjs
-                  .sha256()
-                  .update(contextOrFileName.currentTest?.fullTitle() || '')
-                  .digest('hex');
     const testTitle = typeof contextOrFileName === 'string' ? '' : contextOrFileName.currentTest?.title || '';
     const fileNamePrefix =
-        typeof contextOrFileName === 'string' ? contextOrFileName : `${testTitle}_${fullTestNameHash}`;
-    const name = `${fileNamePrefix}_${uuid()}`.replace(/[\W]+/g, '_');
+        typeof contextOrFileName === 'string' ? contextOrFileName : `${testTitle}_${screenShotIndex}`;
+    const name = fileNamePrefix.replace(/[\W]+/g, '_');
```
DonJayamanne, line 329: Can we keep the hash and the screenshot index and just remove the `uuid`.
I think that will reduce the length.
Optionally we could also reduce the length of the test title by taking just the first 100 characters `testTitle.substring(0, 100)`

The hash is used in another place, so we can determine the screenshots that belong to a particular test.
Basiscally we now generate ipynb files with the test failures and screenshots of the errors, i demnoe'd this last week, probably should share this again.

If  you go into the CI job and look for artifacts named 'TestResultNotebook...ipynb` that will contain the failed tests, with test logs for that test and the corresponding screenshots as well.
I find that very useful, else I have to parse the entire log file manually & then look for screenshots manually, this way the ipynb has everything in one place.

DonJayamanne (changes requested):
Please leave the hash, and you could just remove the uuid and keep the screenshot index.

src/test/common.node.ts
rchiodo: I wonder if that notebook works on windows at the moment. I should try that and see.

rchiodo: Hmm, I don't see a failure with screenshots.

rchiodo: Did you remove the separate screenshots now that we have a notebook? 

On this run, I would have expected there to be screenshots:
https://github.com/microsoft/vscode-jupyter/actions/runs/2659693681

rchiodo:
Looks like screenshots aren't working at all before this, so closing for now. I'll retry opening screenshots after we get screenshots working again.

DonJayamanne:
I don't think this is required anymore.

PR #10785, opened 2022-07-12
Fix #10733. Support interactive window restore on window reload

### LINKED ISSUE #10733: Interactive Window Hot Exit
VS Code is adding support for restoring Interactive Window editors on reload (hot exit), we would need to re-link the python file and the restored Interactive Window editor after the workspace is reloaded.

rebornix:
Fixes #10733

With changes in this PR and https://github.com/microsoft/vscode/pull/154974, users can now choose if they want to keep the interactive window on window reloads. Jupyter will now cache the interactive windows and validate them with open tabs (through the new Tab api).

This PR works when if https://github.com/microsoft/vscode/pull/154974 is merged late. If VS Code part is released first, since it's turned off by default, existing users won't see any difference. We could turn the hot exit setting default value to `true` once we release changes in both products and feel comfortable of doing that.

### DIFF
--- a/src/interactive-window/interactiveWindow.ts
+++ b/src/interactive-window/interactiveWindow.ts
@@ -166,26 +166,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         }
     }
 
-    public async start(preferredController: IVSCodeNotebookController | undefined) {
-        if (preferredController) {
-            this.preferredController = preferredController;
-        }
-        if (!this.notebookEditor) {
-            if (isInteractiveInputTab(this.notebookEditorOrTab)) {
-                const document = await workspace.openNotebookDocument(this.notebookEditorOrTab.input.uri);
-                const editor = await window.showNotebookDocument(document, {
-                    viewColumn: this.notebookEditorOrTab.group.viewColumn
-                });
-                this._notebookEditor = editor;
-                await this.startKernel(
-                    this.preferredController?.controller,
-                    this.preferredController?.connection
-                ).ignoreErrors();
-            } else {
-                this._notebookEditor = this.notebookEditorOrTab;
-            }
-        }
-
+    public start() {
         window.onDidChangeActiveNotebookEditor((e) => {
             if (e === this.notebookEditor) {
                 this._onDidChangeViewState.fire();
@@ -209,6 +190,25 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         } else if (this.isWebExtension) {
             this.insertInfoMessage(DataScience.noKernelsSpecifyRemote()).ignoreErrors();
         }
+    }
+
+    public async restore(preferredController: IVSCodeNotebookController | undefined) {
+        if (preferredController) {
+            this.preferredController = preferredController;
+        }
+        if (!this.notebookEditor) {
+            if (isInteractiveInputTab(this.notebookEditorOrTab)) {
+                const document = await workspace.openNotebookDocument(this.notebookEditorOrTab.input.uri);
+                const editor = await window.showNotebookDocument(document, {
+                    viewColumn: this.notebookEditorOrTab.group.viewColumn
+                });
+                this._notebookEditor = editor;
+            } else {
+                this._notebookEditor = this.notebookEditorOrTab;
+            }
+        }
+
+        this.start();
     }
 
     private async startKernel(
--- a/src/interactive-window/interactiveWindow.ts
+++ b/src/interactive-window/interactiveWindow.ts
@@ -159,6 +159,9 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         this._notebookUri = isInteractiveInputTab(notebookEditorOrTab)
             ? notebookEditorOrTab.input.uri
             : notebookEditorOrTab.notebook.uri;
+        if (!isInteractiveInputTab(notebookEditorOrTab)) {
+            this._notebookEditor = notebookEditorOrTab;
+        }
 
         // Set our owner and first submitter
         if (this._owner) {
--- a/src/interactive-window/interactiveWindow.ts
+++ b/src/interactive-window/interactiveWindow.ts
@@ -21,19 +21,14 @@ import {
     NotebookController,
     NotebookEdit
 } from 'vscode';
-import {
-    IApplicationShell,
-    ICommandManager,
-    IDocumentManager,
-    IWorkspaceService
-} from '../platform/common/application/types';
+import { ICommandManager, IDocumentManager, IWorkspaceService } from '../platform/common/application/types';
 import { Commands, defaultNotebookFormat, MARKDOWN_LANGUAGE, PYTHON_LANGUAGE } from '../platform/common/constants';
 import '../platform/common/extensions';
 import { traceError, traceInfoIfCI } from '../platform/logging';
 import { IFileSystem } from '../platform/common/platform/types';
 import * as uuid from 'uuid/v4';
 
-import { IConfigurationService, InteractiveWindowMode, Resource } from '../platform/common/types';
+import { IConfigurationService, InteractiveWindowMode, IsWebExtension, Resource } from '../platform/common/types';
 import { noop } from '../platform/common/utils/misc';
 import {
     IKernel,
@@ -53,8 +48,13 @@ import { INotebookExporter } from '../kernels/jupyter/types';
 import { IExportDialog, ExportFormat } from '../notebooks/export/types';
 import { generateCellsFromNotebookDocument } from './editor-integration/cellFactory';
 import { CellMatcher } from './editor-integration/cellMatcher';
-import { IInteractiveWindowLoadable, IInteractiveWindowDebugger, IInteractiveWindowDebuggingManager } from './types';
-import { generateInteractiveCode } from './helpers';
+import {
+    IInteractiveWindowLoadable,
+    IInteractiveWindowDebugger,
+    IInteractiveWindowDebuggingManager,
+    InteractiveTab
+} from './types';
+import { generateInteractiveCode, isInteractiveInputTab } from './helpers';
 import { IControllerSelection, IVSCodeNotebookController } from '../notebooks/controllers/types';
 import { DisplayOptions } from '../kernels/displayOptions';
 import { getInteractiveCellMetadata } from './helpers';
@@ -85,9 +85,6 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
     public get submitters(): Uri[] {
         return this._submitters;
     }
-    public get notebookUri(): Uri {
-        return this.notebookEditor.notebook.uri;
-    }
     public get notebookDocument(): NotebookDocument {
         return this.notebookEditor.notebook;
     }
@@ -98,7 +95,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
     private closedEvent = new EventEmitter<void>();
     private _submitters: Uri[] = [];
     private fileInKernel: Uri | undefined;
-    private cellMatcher;
+    private cellMatcher: CellMatcher;
 
     private internalDisposables: Disposable[] = [];
     private kernelDisposables: Disposable[] = [];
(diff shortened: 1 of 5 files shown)


src/interactive-window/interactiveWindow.ts
```diff
@@ -110,33 +107,85 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
-        private readonly storageFactory: IGeneratedCodeStorageFactory,
-        private readonly debuggingManager: IInteractiveWindowDebuggingManager,
-        private readonly isWebExtension: boolean
+        private _owner: Resource,
+        public mode: InteractiveWindowMode,
+        private preferredController: IVSCodeNotebookController | undefined,
+        private readonly notebookEditorOrTab: NotebookEditor | InteractiveTab,
+        public readonly inputUri: Uri
     ) {
+        this.documentManager = this.serviceContainer.get<IDocumentManager>(IDocumentManager);
```
rchiodo, line 142: Why did you change all of these to come from the service container? 

This has caused problems in the past when we ship because sometimes the items aren't available anymore in the service container (somebody refactors something somewhere else). With an explicit dependency this is impossible.

rchiodo: Sorry I take that back. This was already doing that. You basically moved this from the provider to here.

src/interactive-window/interactiveWindowProvider.ts
```diff
@@ -132,6 +156,15 @@ export class InteractiveWindowProvider
         if (!result) {
             // No match. Create a new item.
             result = await this.create(resource, mode, connection);
+            // We already check preferred kernel before creating the interactive window
+            // no need to update it anymore.
+            await result.start(undefined);
+        } else {
+            const preferredController = connection
```
DonJayamanne, line 163: I think we should cache the controller id as well.
Assume user starts IW, and then changes the kernel, now they reload VS Code, the IW will point to the wrong kernel.

rebornix: Good catch

> changes the kernel, now they reload VS Code

To double confirm, there is no "cell execution" between these two steps, right?

src/interactive-window/interactiveWindowProvider.ts
```diff
@@ -273,6 +291,18 @@ export class InteractiveWindowProvider
...
+        const windowCache = this._windows.map(
+            (iw) =>
+                ({
+                    owner: iw.owner,
+                    mode: iw.mode,
+                    uri: iw.notebookUri,
+                    inputBoxUri: iw.inputUri
+                } as IInteractiveWindowCache)
+        );
+        this.workspaceMemento.update(InteractiveWindowCacheKey, windowCache).then(noop, noop);
```
DonJayamanne, line 304: I do'nt tihnk we can store Uris in this manner in the memento.
If you try to get the value out it will not be de-serialized into Uri class objects, but will be some funky JSON object.
I think its safer to store uri.toString() or the like, and then when reading this  value from cache we should probably use Uri.parse.

This way there's no assumption that things are serialized or anyone trying to use this in the future knows that its not a Uri object, but a plain string that needs to be parsed.

Also, i think we should store the controllerId (or kernelConnection.id = both are the same) at this point.

src/interactive-window/types.ts
```diff
@@ -76,6 +78,22 @@ export interface IInteractiveWindow extends IInteractiveBase {
     export(cells?: ICell[]): void;
 }
 
+export interface IInteractiveWindowCache {
+    owner: Resource;
+    mode: InteractiveWindowMode;
+    uri: Uri;
+    inputBoxUri: Uri;
```
DonJayamanne, line 85: I don't think we should be storing Uris into memento, they don't get serialized and deserialized properly, its stored as plain JSON and this could lead to problems

rebornix: Valid point. Another approach is introducing `UriComponents` like in the core, where `UriComponents` is an interface. But function-wise it works the same as `string`.

src/interactive-window/interactiveWindowProvider.ts
```diff
@@ -132,6 +156,15 @@ export class InteractiveWindowProvider
...
             result = await this.create(resource, mode, connection);
+            // We already check preferred kernel before creating the interactive window
+            // no need to update it anymore.
+            await result.start(undefined);
+        } else {
+            const preferredController = connection
+                ? this.controllerRegistration.get(connection, InteractiveWindowView)
+                : await this.controllerDefaultService.computeDefaultController(resource, InteractiveWindowView);
+
+            await result.start(preferredController);
```
DonJayamanne, line 167: I think we should let the `create` method figure out the preferred controller and pass that into the contructor.
that way we don't have two ways of starting the IW with & without arguments for the notebook controller.
Also cleaner, else the controller can be changed from outside, when i think it should be a value passed into the ctor at the point of creating the IW.

Basically i think we can move this `preferedController = connection...` code into the `create` method & then remove the public method `start`

src/interactive-window/interactiveWindowProvider.ts
rebornix: > This way there's no assumption that things are serialized or anyone trying to use this in the future knows that its not a Uri object

Agree this is probably better to check if they need to perform `Uri.from`.

DonJayamanne (changes requested):
I'd consider moving the code to figure out the controller into the create method as that's where we already do the dterminination, this also removes the need for the new start method (which feels weird, as it can be called with and without an arument)

src/interactive-window/interactiveWindowProvider.ts
```diff
@@ -109,6 +102,37 @@ export class InteractiveWindowProvider
...
+        this.restoreWindows();
+    }
+
+    private restoreWindows() {
+        // VS Code controls if interactive windows are restored.
+        const interactiveWindowMapping = new Map<string, InteractiveTab>();
+        window.tabGroups.all.forEach((group) => {
+            group.tabs.forEach((tab) => {
+                if (isInteractiveInputTab(tab) && tab.input.uri) {
+                    interactiveWindowMapping.set(tab.input.uri.path, tab);
```
rchiodo, line 114: I think we have something like getkey for a uri? Would that be better than path? (Path wouldn't match say if the cases were different on windows)

src/interactive-window/interactiveWindowProvider.ts
rchiodo: Cell execution happens after the window is created, so yeah there shouldn't be any cell execution between these two steps.

src/interactive-window/interactiveWindowProvider.ts
rchiodo: +1, agree with Don. Controller in the constructor makes the logic simpler in the class itself.

rebornix: > I think we should let the create method figure out the preferred controller and pass that into the contructor.
that way we don't have two ways of starting the IW with & without arguments for the notebook controller

This is what `InteractiveWindowProvider#create` is already doing now, it creates controller based on connection first and then create `InteractiveWindow` objects. While on window reload, we create `InteractiveWindow` from cache, at which point we are not in this code path of `InteractiveWindowProvider#create` but in a `get` path.

src/interactive-window/interactiveWindowProvider.ts
rebornix: @rchiodo I pushed a fix to use `uri.toString()`.

rchiodo: Is that the same as getComparisonKey? https://github.com/microsoft/vscode-jupyter/blob/acb70fa84930ad77ccd08180b5351902664e04d4/src/platform/vscode-path/resources.ts#L57

src/interactive-window/interactiveWindow.ts
```diff
@@ -110,33 +107,85 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
-        private readonly fs: IFileSystem,
-        private readonly configuration: IConfigurationService,
-        private readonly commandManager: ICommandManager,
-        private readonly jupyterExporter: INotebookExporter,
-        private readonly workspaceService: IWorkspaceService,
-        private _owner: Resource,
-        private mode: InteractiveWindowMode,
-        private readonly exportDialog: IExportDialog,
-        private readonly notebookControllerSelection: IControllerSelection,
         private readonly serviceContainer: IServiceContainer,
```
amunger, line 135: I'm not a fan of passing the serviceContainer to constructors since it makes the class more coupled with the specific IoC container, but I suppose it's ok since it was already being passed in anyway

src/interactive-window/interactiveWindowProvider.ts
rebornix: If we need to compare `Uri`s we should use `getComparisonKey` which can normalize paths, here we are using it as an identifier so a `toString()` is sufficient.

src/interactive-window/interactiveWindow.ts
rebornix: Agree, I'm not a fan either. My motivation is avoiding doing `serviceContainer.get<>` in multiple places where we call `InteractiveWindow#ctor`. 

@DonJayamanne pointed out that it's explicit to us how many services external `InteractiveWindow` import but moving `serviceContainer.get<>` into `ctor` kinds of hiding it. We might want to revisit this along with other challenges I ran into (will leave a comment about this soon)

rebornix:
Synced with @DonJayamanne offline to touch base on https://github.com/microsoft/vscode-jupyter/pull/10785#discussion_r919500505, there are few things mixed together making it a bit complex to understand:

* Firstly we want the `InteractiveWindowProvider#create` to be synchronous (as requested in the comment https://github.com/microsoft/vscode-jupyter/blob/b8f5974a47b8e43e972aa876bacf5e302f17eedd/src/interactive-window/interactiveWindowProvider.ts#L185) when creating `InteractiveWindow` and inserting it to `InteractiveWindowProvider#_windows` to ensure that when the IW creation is called fast, we don't run into race conditions
  * 🐛  Though currently the race condition can happen as it's asynchronous to fetch default preferred controller https://github.com/microsoft/vscode-jupyter/blob/b8f5974a47b8e43e972aa876bacf5e302f17eedd/src/interactive-window/interactiveWindowProvider.ts#L201-L202 and create IW editor https://github.com/microsoft/vscode-jupyter/blob/b8f5974a47b8e43e972aa876bacf5e302f17eedd/src/interactive-window/interactiveWindowProvider.ts#L204
* Thus currently the `InteractiveWindow` creation is async when being newly created (we pass controller, notebook editor, and input box uri to the constructor), however on window reload, we are restoring the windows synchronously. 
  * When we restore `InteractiveWindowProvider#_windows` from cache, the controllers might not be all registered yet (@DonJayamanne pointed out that we are not caching controllers now), so caching controller ID does not always work as we might not have the registered controller yet. `preferredController` needs to be found later when users click on **Run Cell**
 codelens, that's the reason why we need to restore https://github.com/microsoft/vscode-jupyter/pull/10785/files#diff-ea1d2b78c1f2061826843821490a1383de7feb165ed7cc45a59efa17809852d9R166
  * On Window reload, `NotebookEditor` might not be available either. In IW creation code path, we always create the IW first (and `await` for it) so we have the `NotebookEditor` object before ctor. However, on window reload, an interactive window tab might be hidden (behind other tabs). `NotebookEditor` is only resolved when we bring it to the front of the workspace, we can't do that right after the window reload since it can break user's current workspace state.
    * That's why we postpone it also to when users click the code lens


It would involve quite some code changes to get this right so I'd love to have them resolved in follow up PRs. One solution we discussed is

* Create `InteractiveWindow` immediately with `owner` and `mode`, push it to `_windows` in both cases (create or restore)
* Have one explicit `start` method to get the `InteractiveWindow` prepared for cell execution, which includes
  * If it's new IW
    * Get `preferredController`
    * Call `interactive.open` to create the IW editor in VS Code
    * Start kernel
  * If it's a restored IW, we then have the `InteractiveTab`
    * Reveal the tab into view (currently we use `openNotebookDocument` API, in the future we would have Tab API to archive this)
    * Get `preferredController` and set it as the active controller (maybe we don't need this as the notebook controller provider already tells VS Code which controller to use)
    * Start kernel


@rchiodo @IanMatthewHuff @amunger would love to have your insights into this.

rchiodo:
I'm not sure why create needs to be synchronous? Because the user might try to create two interactive windows at the same time? It seems the comment there has already been ignored. 

The problem I believe Don was trying to workaround was inside the IW class itself. If the object isn't created with a preferred controller, it then needs to do a lot of stuff async. It felt to me like async creation was better than making parts of IW async.

DonJayamanne (approved):

src/interactive-window/interactiveWindowProvider.ts
```diff
@@ -132,6 +156,14 @@ export class InteractiveWindowProvider
         if (!result) {
             // No match. Create a new item.
             result = await this.create(resource, mode, connection);
+            // start the kernel
+            result.start();
```
rchiodo, line 160: Would it make sense to just put this into the create?

rchiodo (approved):

amunger (approved):

IanMatthewHuff (approved):

PR #10787, opened 2022-07-13
Warn when files could override Python modules

### LINKED ISSUE #7538: Warn users when files override built in packages that could have caused the kernel to fail
We still have a large number of users running into issues resulting from creating files such as `string.py`, `email.py`, `os.py`, `sys.py` etc
I think in the past 3 weeks we had 3 users running into this issue.

This causes a lot of issues, kernel don't start or when packages get imported dynamically kernels behave weirdly.

Solution:
* Keep a hardcoded list of reserved file names (use `help('modules')` to get a list of all built in modules)
This should also help Python extension, but not sure it matters as much for Python extension.

DonJayamanne:
Fixes #7538


src/extension.node.ts
```diff
@@ -169,7 +170,9 @@ async function activateUnsafe(
 
         const [serviceManager, serviceContainer] = initializeGlobals(context);
         activatedServiceContainer = serviceContainer;
-        initializeTelemetryGlobals(serviceContainer);
+        initializeTelemetryGlobals((interpreter) =>
```
DonJayamanne, line 173: Just pass a callback instead of passing the entire interface, as we dont want the telemetry layer to have any dependencies on interpreter code.

src/platform/telemetry/telemetry.ts
```diff
@@ -5,23 +5,23 @@
...
 
 type Context = {
     previouslySelectedKernelConnectionId: string;
 };
 export const trackedInfo = new Map<string, [ResourceSpecificTelemetryProperties, Context]>();
 export const pythonEnvironmentsByHash = new Map<string, PythonEnvironment>();
-let globalContainer: IServiceContainer | undefined = undefined;
-
-export function initializeGlobals(serviceContainer: IServiceContainer) {
-    globalContainer = serviceContainer;
```
DonJayamanne, line 24: We no longer need to inject the entire service container, just a callback is sufficient.

src/extension.node.ts
```diff
@@ -169,7 +170,9 @@ async function activateUnsafe(
...
+            serviceContainer.get<IInterpreterPackages>(IInterpreterPackages).getPackageVersions(interpreter)
```
DonJayamanne, line 174: Moved IInterpreterPackages from telemetry folder into Interprter folder, and passed just a callback into interpreters layer, we don't need everything, a callback is sufficient.

src/kernels/installer/productInstaller.node.ts
```diff
@@ -29,13 +29,13 @@ import {
 import { noop } from '../../platform/common/utils/misc';
 import { IServiceContainer } from '../../platform/ioc/types';
 import { sendTelemetryEvent, Telemetry } from '../../telemetry';
-import { InterpreterPackages } from '../../platform/telemetry/interpreterPackages.node';
+import { InterpreterPackages } from '../../platform/interpreter/interpreterPackages.node';
```
DonJayamanne, line 32: Fixed dependency issues, installation code does not need to depend on telemetry for things like getting packages

src/platform/interpreter/interpreterPackages.web.ts
```diff
@@ -2,15 +2,18 @@
 // Licensed under the MIT License.
 
 import { injectable } from 'inversify';
-import { InterpreterUri } from '../common/types';
+import { InterpreterUri, Resource } from '../common/types';
 import { PythonEnvironment } from '../pythonEnvironments/info';
-import { IInterpreterPackages } from '../../telemetry';
+import { IInterpreterPackages } from './types';
```
DonJayamanne, line 7: Moved out of telemetry into interpreter.

src/platform/interpreter/types.ts
```diff
@@ -0,0 +1,24 @@
...
+    getUriOverridingReservedPythonNames(cwd: Uri): Promise<{ uri: Uri; type: 'file' | '__init__' }[]>;
+    isReserved(uri: Uri): Promise<boolean>;
+    /**
+     * Keeps track of a Uri as a file that should be ignored from all warnings related to reserved names.
+     */
+    addToIgnoreList(uri: Uri): Promise<void>;
+}
+
+export const IInterpreterPackages = Symbol('IInterpreterPackages');
+export interface IInterpreterPackages {
```
DonJayamanne, line 19: Moved from telemetry folder

DonJayamanne: I need to reference this from the new code, hence moved this from telemtry into Interrpeter & ensured telemetry doesnt depend on this.

src/platform/serviceRegistry.web.ts
```diff
@@ -50,6 +54,15 @@ export function registerTypes(serviceManager: IServiceManager) {
         KernelProgressReporter
     );
 
+    serviceManager.addSingleton<IInterpreterPackages>(IInterpreterPackages, InterpreterPackages);
+    serviceManager.addSingleton<IExtensionSyncActivationService>(
```
DonJayamanne, line 58: These things that have to do with interpeters is now in interpeters folder and no longer in telemetry folder.

src/platform/telemetry/serviceRegistry.node.ts
```diff
@@ -1,23 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
DonJayamanne, line 1: Yay (i think), service registrations in telemetry folder anymore

src/telemetry.ts
```diff
@@ -106,13 +105,6 @@ export type ResourceSpecificTelemetryProperties = Partial<{
     kernelLiveCount: number;
 }>;
 
-export const IInterpreterPackages = Symbol('IInterpreterPackages');
-export interface IInterpreterPackages {
```
DonJayamanne, line 110: Moved out of here

src/platform/errors/errorUtils.ts
```diff
@@ -423,29 +435,73 @@ export function analyzeKernelErrors(
...
+        let fileName = '';
+        let folderName = '';
+        if (moduleName.split("'").length > 2) {
+            fileName = moduleName.split("'").slice(-2)[0] || '';
+            fileName = fileName ? `${fileName}.py` : '';
+            folderName = fileName ? fileName : '';
+            moduleName = moduleName.split("'")[1] || moduleName;
+        }
+        const modulesInCwd = filesInCwd
+            .filter((file) => file.toString().toLowerCase().endsWith('__init__.py'))
```
IanMatthewHuff, line 474: Kinda pedantic of me, but seems like this should be an exact match and not endsWith. I don't really see someone naming a file HELLO__init__.py but you never know :D

src/platform/interpreter/reservedNamedProvider.node.ts
```diff
@@ -0,0 +1,146 @@
...
+
+        // Use the name of the file as the module name.
+        const baseName = path.basename(uri, path.extname(uri)).toLowerCase();
+        // If its a __init__.py, get name of parent folder (as its a module).
+        const possibleModule = baseName === '__init__' ? path.basename(path.dirname(uri)).toLowerCase() : baseName;
+        if (this.cachedModules.has(possibleModule)) {
+            return true;
+        }
+
+        if (this.extensionChecker.isPythonExtensionInstalled) {
```
IanMatthewHuff, line 110: Would this check make sense inside of package.listPackages? Thinking about the case that I was looking at (which you've resolved in this PR I believe) where this was getting pulled in with telemetry. Someone consuming listPackage might not be aware that they are creating an install prompt for the PythonExtension when using it.

src/standalone/diagnostics/reservedFileNameDiagnostics.node.ts
```diff
@@ -0,0 +1,208 @@
...
+import { disposeAllDisposables } from '../../platform/common/helpers';
+import { Common, DataScience } from '../../platform/common/utils/localize';
+import * as path from '../../platform/vscode-path/path';
+import { IFileSystem } from '../../platform/common/platform/types';
+import { IWorkspaceService } from '../../platform/common/application/types';
+import { IReservedPythonNamedProvider } from '../../platform/interpreter/types';
+import { JupyterKernelStartFailureOverrideReservedName } from '../../platform/interpreter/constants';
+import { swallowExceptions } from '../../platform/common/utils/decorators';
+
+export const enabledSettingName = 'diagnostics.reservedPythonNames.enabled';
```
IanMatthewHuff, line 40: This is a bit funny to be just floating loose as an export. For testing?

src/standalone/diagnostics/reservedFileNameDiagnostics.node.ts
```diff
@@ -0,0 +1,208 @@
...
+
+        window.onDidChangeActiveTextEditor(this.provideDiagnosticsForEditor, this, this.disposables);
+        workspace.onDidCloseTextDocument((e) => this.diagnosticCollection.delete(e.uri), this, this.disposables);
+        window.visibleTextEditors.forEach((editor) => this.provideDiagnosticsForEditor(editor));
+        this.disposables.push(
+            commands.registerCommand(
+                'jupyter.ignoreReservedFileNamesDiagnostic',
+                async (uri: Uri) => {
+                    await this.reservedNameProvider.addToIgnoreList(uri);
+                    this._onDidChangeFileDecorations.fire(uri);
```
IanMatthewHuff, line 93: Ordering might be off here. I would think you'd fire the deco change after deleting.

IanMatthewHuff (approved):
Cool this will be great for customers (still want Python to own this, but that can always come later). One minor question was on the listPackages possibly guarding against python extension install internally.

src/platform/errors/errorUtils.ts
DonJayamanne: Fixed

src/platform/interpreter/reservedNamedProvider.node.ts
DonJayamanne: Good point

src/standalone/diagnostics/reservedFileNameDiagnostics.node.ts
DonJayamanne: Yes thats correct, do you think we should not export these and just hard code this value in the tests as well?
I used to not export and just hard code in the tests, this way its not exported unnecessarily and the tests also verify whether we're using the right values, however over time i've changed to just export it.

Thoughts?

src/standalone/diagnostics/reservedFileNameDiagnostics.node.ts
DonJayamanne: No its correct, because we basically tell VS Code the decorations have changed for the file and when vscode calls back into our code to get the decorations we should have ignored the file (meaning that should have happened first)

src/kernels/errors/kernelErrorHandler.web.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Resource } from '../../platform/common/types';
+import { DataScienceErrorHandler } from './kernelErrorHandler';
+
+@injectable()
+export class DataScienceErrorHandlerWeb extends DataScienceErrorHandler {
+    protected override async addErrorMessageIfPythonArePossiblyOverridingPythonModules(
+        _messages: string[],
+        _resource: Resource
+    ) {
+        //
```
rchiodo, line 13: I wonder if we need a remote check in the node version? Theoretically there could be a remote version of this too where we use the cwd of the kernel process to check for files (instead of reading from disk)

DonJayamanne: > we need a remote check in the node version? Theoretically there could be a remote version o

**Agreed, however** 
* I'm assuming things are already working on the remote because they already have jupyter server running
* reading the files via a kernel would still require the kernel to start (but this code is used to detect what could cause the kernel to fail)


I think this will be necessary when users have access to some kind of a file system where they can create files, and at that point I think we can try to look into this. Else I feel we can leave this for now (at least for remote).

src/platform/interpreter/reservedNamedProvider.node.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    public dispose() {
+        disposeAllDisposables(this.disposables);
+    }
+
+    public async getUriOverridingReservedPythonNames(cwd: Uri): Promise<{ uri: Uri; type: 'file' | '__init__' }[]> {
+        const [files, initFile] = await Promise.all([
+            this.fs.searchLocal('*.py', cwd.fsPath, true),
+            // Look for packages in the current directory (only interested in top level).
```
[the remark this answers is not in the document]
rchiodo: Oh the remote check is in the error handler. Maybe a comment here indicating this only works for local?

src/kernels/errors/kernelErrorHandler.web.ts
rchiodo: Okay maybe not the cwd of the kernel process, but rather the file system api from the Jupyter server.

rchiodo (approved):

src/standalone/diagnostics/reservedFileNameDiagnostics.node.ts
IanMatthewHuff: Not sure that I see it as a big deal. Since the tests have the class I guess having it as a public static could be a possibility.

ReservedFileNamesDiagnosticProvider.enabledSettingName

Constraints it a bit to the class in question. But it's already scoped via the file it's in, so again I'm fine with it as is.

DonJayamanne: Agreed, but not a fan of statics as well, as that's not really a const, even though we could create a `static readonly`
leaving this as is.

PR #10796, opened 2022-07-14
Stop waiting for idle kernel when swithing kernels

### LINKED ISSUE #10795: Couldn't switch kernels and run when there is a pending execution/connection
I have a broken/misbehaving kernel locally which can't be started. It's picked as the preferred kernel and when I run cells, we would attempt to start the kernels and re-try 4 more times if connection fails. While it's re-trying

* We can't stop the execution
* Or kill the kernel (connection)
* Switching to another kernel doesn't clear or stop the execution, so I couldn't start with another kernel. Have to wait until the 5 times of kernel connections fail

DonJayamanne:
Fixes #10795

### DIFF
--- a/news/2 Fixes/10795.md
+++ b/news/2 Fixes/10795.md
@@ -1,0 +1,1 @@
+Stop waiting for kernel to start (or be idle) when switching kernels.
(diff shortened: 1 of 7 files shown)


rebornix (approved):
👏

PR #10807, opened 2022-07-14
Enable an experimental setting for the new kernel picker UI

### LINKED ISSUE #10782: Experiment setting to control new kernel UI
In order to determine if the new UI is actually worthwhile or not, we should add an experiment setting that turns it on.

We'd check if remote usage changes or not with the new UI.

rchiodo:
Fixes #10782 

For the most part I left a copy of the code for the UI parts. This was simpler than trying to make the new code handle the old UI and the new.


package.json
```diff
@@ -2033,6 +2033,15 @@
...
                     "markdownDescription": "%jupyter.configuration.jupyter.enableExtendedKernelCompletions.markdownDescription%",
                     "scope": "machine"
+                },
+                "jupyter.showOnlyOneTypeOfKernel": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "%jupyter.configuration.jupyter.showOnlyOneTypeOfKernel.markdownDescription%",
+                    "scope": "machine",
+                    "tags": [
+                        "experimental"
```
rchiodo, line 2043: This tag enables the changing of this setting through an experiment service.

src/notebooks/controllers/commands/serverConnectionControllerCommands.ts
```diff
@@ -86,17 +88,32 @@ export class ServerConnectionControllerCommands implements IExtensionSingleActiv
...
-            .ignoreErrors();
-        this.showingRemoteNotWebContext.set(!isLocal && !this.isWeb).ignoreErrors();
-        this.showingRemoteContext.set(!isLocal && this.controllerRegistration.values.length > 0).ignoreErrors();
+            this.showingLocalOrWebEmptyContext
+                .set(isLocal || (this.isWeb && this.controllerRegistration.values.length === 0))
+                .ignoreErrors();
+            this.showingRemoteNotWebContext.set(!isLocal && !this.isWeb).ignoreErrors();
+            this.showingRemoteContext.set(!isLocal && this.controllerRegistration.values.length > 0).ignoreErrors();
+        } else {
+            this.showingLocalOrWebEmptyContext.set(false).ignoreErrors();
```
rchiodo, line 105: Commands are always hidden if the setting is turned off

src/notebooks/controllers/kernelFilter/kernelFilterService.ts
```diff
@@ -139,3 +144,10 @@ type InterpreterFiter = {
      */
     path: string;
 };
+type RemoteSpecFilter = {
```
rchiodo, line 147: Added support for remote filtering. Made testing this easier.

src/test/datascience/jupyter/serverSelector.unit.test.ts
```diff
@@ -88,206 +107,453 @@ suite('DataScience - Jupyter Server URI Selector', () => {
         disposeAllDisposables(disposables);
     });
 
-    test('Local pick server uri', async () => {
-        const { selector, storage } = createDataScienceObject('', true);
-        await selector.selectJupyterURI();
-        let value = await storage.getUri();
-        assert.equal(value, Settings.JupyterServerLocalLaunch, 'Default should pick local launch');
+    suite('Experimental', () => {
```
rchiodo, line 110: Two copies of the tests. 

One for new way the server selector works
One for the old way.

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -183,6 +183,13 @@ export class VSCodeNotebookController implements Disposable, IVSCodeNotebookCont
         this.controller.supportedLanguages = this.languageService.getSupportedLanguages(kernelConnection);
         // Hook up to see when this NotebookController is selected by the UI
         this.controller.onDidChangeSelectedNotebooks(this.onDidChangeSelectedNotebooks, this, this.disposables);
+        this.notebookApi.onDidCloseNotebookDocument(
```
rchiodo, line 186: Associated documents were surviving tests closing notebooks.

rchiodo: Which could also potentially happen in the real product.

User opens notebook
Connects to remote
User closes notebook
User opens another notebook
Connects to local
Kernel list still includes remote connection from notebook that isn't open anymore

rchiodo: New test I added is verifying this isn't the case

src/platform/common/application/notebook.ts
```diff
@@ -112,10 +113,11 @@ export class VSCodeNotebook implements IVSCodeNotebook {
     public async closeActiveNotebooks() {
         // We could have untitled notebooks, close them by reverting changes.
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        const documents = new Set<NotebookDocument>();
+        const documents = new Set<NotebookDocument>(this.notebookDocuments);
```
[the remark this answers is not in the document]
DonJayamanne: Never mind, i can see that we're relying on the events.. bummer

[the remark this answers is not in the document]
rchiodo: Yes it is a hack. The 'revertAndClose' command does not actually send a close event. It can't be in the test code because it can't fire the onDidCloseNotebook event. Only this class can fire that event.

src/kernels/jupyter/serverSelector.ts
```diff
@@ -284,3 +376,274 @@ export class JupyterServerSelector {
...
+                                  tooltip: DataScience.removeRemoteJupyterServerEntryInQuickPick()
+                              }
+                          ]
+                });
+            }
+        });
+
+        return items;
+    }
+}
```
sadasant, line 649: Cleanly done! I like it

src/notebooks/controllers/kernelFilter/kernelFilterUI.ts
```diff
@@ -65,7 +71,7 @@ export class KernelFilterUI implements IExtensionSyncActivationService, IDisposa
                     })
                     .map((item) => {
                         return <QuickPickType>{
-                            label: getDisplayNameOrNameOfKernelConnection(item),
+                            label: getKernelLabel(item),
                             picked: !this.kernelFilter.isKernelHidden(item),
                             description: getKernelConnectionPath(item, this.workspace),
                             detail:
```
sadasant, line 77: Nice!

src/notebooks/controllers/types.ts
```diff
@@ -34,9 +34,13 @@ export const IControllerRegistration = Symbol('IControllerRegistration');
...
 export interface IControllerRegistration {
     /**
-     * Gets the list of all of the controllers
+     * Gets the filtered list of all of the controllers
      */
     values: IVSCodeNotebookController[];
+    /**
+     * Gets every registered connection metadata
+     */
+    all: KernelConnectionMetadata[];
```
DonJayamanne, line 43: Can't we already get this same information from outside using the existing public property `values` as follows
`values.map(item => item.connection`

Adding another property seems redundant and complicates it as we have more than one way to get the same information. I.e. i though this would be different from `values`, however looking at the code this is identical.

I think we should remove the propert `registeredMetadatas` and make it just return as follows:
```typescript
private get metadatas() {
    return this.values.map(item => item.connection);
}
```

& then delete `this.regsiteredMetadatas` property.
I think i see a bug, instead of removing the controller when Uris change, we're only removing the metadata. Removing this duplicate property would resolve that bug

src/notebooks/controllers/vscodeNotebookController.ts
sadasant: This might improve performance and potentially fix other unnoticed bugs! I think? This looks really promising to me

src/notebooks/controllers/types.ts
rchiodo: values is not the same as 'all'. Values are the items active in the kernel picker. All is every kernel that's been registered.

When the URI changes we remove the metadatas and then we reapply the filter. This disposes all controllers that aren't allowed in the filter as shown here:
https://github.com/microsoft/vscode-jupyter/blob/4d1a5c7f58700fa93233226afddc061bd4d5878a/src/notebooks/controllers/controllerRegistration.ts#L245

DonJayamanne (changes requested):
looks ok, except for a problem with the `all` & `registeredMetadatas` properties

src/notebooks/controllers/types.ts
DonJayamanne: Oh ok, got it.

DonJayamanne (approved):

sadasant (approved):
LGTM

rchiodo:
Please see my reply to the 'values' and 'all' conversation. Maybe I should rename 'values' to 'filtered'. It is the filtered list, which when you're local it only has local values, but not remote and vice versa. The 'all' property is every kernel connection every registered.

DonJayamanne:
> he 'values' and 'all' conversation. Maybe I should rename 'values' to 'filtered'.

Or just `reigsteredControllers`, I'd avoid using the word `filter` if possible, as thats just a way of deciding what's registered and what's not. At the end of the day, we're only interested in whats been registered with vs code.
Just a thought.

PR #10811, opened 2022-07-14
HotFix: Port fix to "Show In Data Viewer" to release branch and update for hotfix

IanMatthewHuff:


package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "jupyter",
     "displayName": "Jupyter",
-    "version": "2022.6.110",
+    "version": "2022.6.120",
```
sadasant, line 4: I believe we need the same kind of change in the package-lock.json. There should be two instances there. Please, let the know if I'm wrong 🙇‍♂️

DonJayamanne (approved):
Please ensure this is also merged into the main branch.

sadasant (approved):

package.json
IanMatthewHuff: I did a regen at first, but it tweaked a bit more of the package lock so I just went with manual...

rchiodo (approved):

PR #10773, opened 2022-07-12
Hotfix: Stop using vscode-nls for some key localizations

sadasant:
This PR undoes localization on some localization keys that are breaking the related features of the `vscode-jupyter` extension.

Last week we released version 6.100 of the `vscode-jupyter`. Unbeknown to us, the extension shipped with some of its features broken. Customers began reporting that some of the commands, like the one for restarting kernels and the one for interrupting kernels, were missing from the UI.

In the developer tools of VSCode (version 1.69.0 and above), when 6.100 is loaded, we see a large influx of error messages:

Screen Shot 2022-07-11 at 8 17 20 PM

While debugging, we realized that:

- Version 5.100 does not have the issue, even though it shipped with the localization code (though it didn't ship with the new localizations themselves, only with the code that uses them).
- Version 6.100 does not include localization changes to the related localization IDs.
- The problem doesn't reproduce building a .VSIX of the extension locally.
- Changing the problematic entries in the `package.json` to use existing keys pointing to other localized phrases bypasses the problem.
- Changing the problematic entries in the `package.json` to use plain strings also bypasses the problem.

This PR aims to replace all of the problematic localizations for plain text English strings in order to deliver a hotfix for 6.100.

The changes included in this PR do make all of the errors shown in the screenshot go away.

(Once released)
Fixes #10752

PR #10814, opened 2022-07-14
Revert "Hotfix: Stop using vscode-nls for some key localizations"

### LINKED ISSUE #10774: [Localization] Re-introduce localization to the reverted localization keys
In this PR #10773 we reverted some of the localizations because adding them would break the .VSIX generated through our pipelines. We should re-incorporate these localizations.

sadasant:
Reverts microsoft/vscode-jupyter#10773

Fixes #10774

### DIFF
--- a/package.json
+++ b/package.json
@@ -593,7 +593,7 @@
             },
             {
                 "command": "jupyter.importnotebook",
-                "title": "Import Jupyter Notebook",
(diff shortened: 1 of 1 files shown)


rchiodo:
Does it require it to be submitted to verify the fix? Or can you build the translated version locally?

DonJayamanne (approved):

PR #10817, opened 2022-07-14
Strict IKernel with notebook document linked in core.

rebornix:
This PR tries to separate jupyter kernels created for notebook documents and non notebook documents. 

The motivation behind this is having a clear separation between the use of kernels created for notebooks or non-notebooks. Currently kernels created for non-notebooks are only created by 3rd party extensions and breaking the 1 to 1 mapping between `IKernel` to `NotebookDocument`. That means the core system (notebook and interactive window) needs to handle the `undefined` case when running `getAssociatedNotebookDocument`, even if that should never happen.

As @DonJayamanne and I discussed offline, with changes in this PR, internal components only need to care about Kernels that are attached to a `NotebookDocument`. The `src/kernels` component will offer
* `IKernelProvider
* `IKernel` which always ties to `NotebookDocument`

Kernels created by third party only matters to `src/kernels` component (e.g., execution related code) and API (this is why we have such kernels).


src/interactive-window/editor-integration/codeLensFactory.ts
```diff
@@ -62,12 +61,9 @@ export class CodeLensFactory implements ICodeLensFactory {
...
         this.configService.getSettings(undefined).onDidChange(this.onChangedSettings, this, disposables);
         notebook.onDidChangeNotebookCellExecutionState(this.onDidChangeNotebookCellExecutionState, this, disposables);
-        kernelProvider.onDidDisposeKernel(
+        kernelProvider.onDidDisposeNotebookKernel(
             (kernel) => {
-                const notebook = getAssociatedNotebookDocument(kernel);
-                if (notebook) {
-                    this.notebookData.delete(notebook.uri.toString());
-                }
+                this.notebookData.delete((kernel as INotebookKernel).notebook.uri.toString());
```
DonJayamanne, line 66: We don't need to cast this. 
If we haev a special event just for INotebookKernel, then the argument kernel should be of type INoteobookKernel

src/interactive-window/generatedCodeStoreManager.ts
```diff
@@ -28,15 +27,15 @@ export class GeneratedCodeStorageManager implements IExtensionSyncActivationServ
...
     }
     private onNotebookControllerSelected({ notebook }: { notebook: NotebookDocument }) {
         this.storageFactory.get({ notebook })?.clear();
         this.codeGeneratorFactory.get(notebook)?.reset();
     }
-    private onDidCreateKernel(kernel: IKernel) {
-        const notebook = getAssociatedNotebookDocument(kernel);
+    private onDidCreateKernel(kernel: INotebookKernel) {
+        const notebook = kernel.notebook;
         if (kernel.creator !== 'jupyterExtension' || notebook?.notebookType !== InteractiveWindowView) {
```
DonJayamanne, line 39: notebook won't be undefined.
```suggestion
        if (kernel.creator !== 'jupyterExtension' || notebook.notebookType !== InteractiveWindowView) {
```

src/kernels/kernelProvider.base.ts
```diff
@@ -8,38 +8,58 @@ import { IVSCodeNotebook } from '../platform/common/application/types';
...
      * Where as if a kernel is tied to a notebook, then the kernel dies along with notebooks.
      */
     private readonly kernelsByUri = new Map<string, { options: KernelOptions; kernel: IKernel }>();
     private readonly pendingDisposables = new Set<IAsyncDisposable>();
     protected readonly _onDidRestartKernel = new EventEmitter<IKernel>();
+    protected readonly _onDidRestartNotebookKernel = new EventEmitter<INotebookKernel>();
     protected readonly _onDidStartKernel = new EventEmitter<IKernel>();
+    protected readonly _onDidStartNotebookKernel = new EventEmitter<INotebookKernel>();
     protected readonly _onDidCreateKernel = new EventEmitter<IKernel>();
+    protected readonly _onDidCreateNotebookKernel = new EventEmitter<INotebookKernel>();
```
DonJayamanne, line 34: Do we really need these duplicate events
Can we not have the same event. & when using this we can just check if `kernel.notebook` exists, and if it does then typescript will automatically treat the type as INotebookKernel (after all INotebookKernel extends IKernel).

Else feels like a lot of duplication.

rebornix: This is currently intentional as we are falling into the same problem as before that the core system knows the existence of `IKernel` created by Data Wrangler. Ideally the core system only needs to care about events for `INotebookKernel` and in very few cases that we need to handle both `INotebookKernel` and `IKernel` created by 3rd party.

src/interactive-window/generatedCodeStoreManager.ts
```diff
@@ -28,16 +27,16 @@ export class GeneratedCodeStorageManager implements IExtensionSyncActivationServ
...
     }
     private onNotebookControllerSelected({ notebook }: { notebook: NotebookDocument }) {
         this.storageFactory.get({ notebook })?.clear();
         this.codeGeneratorFactory.get(notebook)?.reset();
     }
-    private onDidCreateKernel(kernel: IKernel) {
-        const notebook = getAssociatedNotebookDocument(kernel);
-        if (kernel.creator !== 'jupyterExtension' || notebook?.notebookType !== InteractiveWindowView) {
+    private onDidCreateKernel(kernel: INotebookKernel) {
+        const notebook = kernel.notebook;
```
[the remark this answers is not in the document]
rebornix: > & we can change the signature of onDidCreateKernel to be Event<IKernel | INotebookKernel>, this way we can use the same event

This way, in every place we listen to the event, we would need to check `kernel.notebook` again and would need to decide how to handle non-notebook case, right?

sadasant (approved):
Pretty neat and complete! From my limited perspective, this looks good!

src/interactive-window/editor-integration/codeLensFactory.ts
```diff
@@ -64,10 +63,7 @@ export class CodeLensFactory implements ICodeLensFactory {
         notebook.onDidChangeNotebookCellExecutionState(this.onDidChangeNotebookCellExecutionState, this, disposables);
         kernelProvider.onDidDisposeKernel(
             (kernel) => {
-                const notebook = getAssociatedNotebookDocument(kernel);
-                if (notebook) {
-                    this.notebookData.delete(notebook.uri.toString());
-                }
+                this.notebookData.delete(kernel.notebook.uri.toString());
```
IanMatthewHuff, line 66: Maybe out of scope for this PR, but I thought we removed almost all cases of using uri.toString(). Seems like this should maybe WeakMap the NotebookDocument instead like we do in other places.

IanMatthewHuff (approved):
I like it, nice to just have a .notebook instead of the helper function all over the place.

rchiodo (changes requested):
I think this is changing the 3rd party API. 

Before this change, 3rd parties could get kernels from live notebooks. Now they can only find kernels that they've created.

I think the 3rd party kernel provider has to search in the original kernel provider too.

rchiodo:
Sorry I didn't read all the way down to the API section. It's using both. My initial feeling was that the 3rd party provider should provide both but thinking about it more, it makes more sense for the API to do it in the API layer.

rchiodo (approved):

rchiodo (changes requested):
I wonder if there still might be problems though with internal code not seeing 3rd party kernels at all? If a 3rd party creates a kernel for a notebook and then the user runs a cell, it won't use it. So from the 3rd party's point of view they're using the wrong kernel.

rchiodo (approved):
No it looks like you've handled that case in the KernelConnector code.

DonJayamanne (approved):

PR #10824, opened 2022-07-15
Try number 2 for the Ensure install prompt test failure

### LINKED ISSUE #10734: Test failing consistently (prompt to install ipykernel)
Looking at the CI logs, this was fine this morning, even the last commit was green.
However most (if not all) PRs seems to be failing now for the same test.

Screen Shot 2022-07-08 at 10 26 45

rchiodo:
Fixes #10734 

On my local machine at least, the 'activeNotebookEditor' was consistently not being set when opening different windows. 

Specifying it explicitly fixes it for me locally plus makes the tests run faster. Not sure we need to validate active notebook editor is correct (as VS code may decide to change focus behavior), so this seems good to me in general.

### DIFF
--- a/src/test/datascience/notebook/remoteNotebookEditor.vscode.test.ts
+++ b/src/test/datascience/notebook/remoteNotebookEditor.vscode.test.ts
@@ -15,7 +15,7 @@ import { traceInfoIfCI, traceInfo } from '../../../platform/logging';
 import { captureScreenShot, IExtensionTestApi, initialize, waitForCondition } from '../../common';
 import { openNotebook } from '../helpers';
 import { JupyterServer } from '../jupyterServer.node';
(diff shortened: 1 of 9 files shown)


src/test/datascience/notebook/remoteNotebookEditor.vscode.test.ts
```diff
@@ -158,6 +159,7 @@ suite('DataScience - VSCode Notebook - (Remote Execution)', function () {
     });
 
     test('Local and Remote kernels are not listed', async function () {
+        await changeShowOnlyOneTypeOfKernel(true);
```
rchiodo, line 162: Oh and I noticed I hadn't changed this test. It was failing from my experiment submission.

rchiodo:
Crap. It's failing the same way as it did before:

image

It's not selecting the active interpreter. But I can't repro anymore on my own machine.

Issue #10638, opened 2022-06-30
Universal dependency installer (and universal debugger code evaluator)


sadasant:
This issue tracks the progress towards a more universal installer. To close this issue, we should end satisfying the following considerations:

- Have the Data Frame working on the web (therefore having #9665 fixed).
- To have a new service, `DebuggerEvaluator`, with an `evaluate` function that would allow the execution of code on the debugger.
  - This debugger evaluator should replace all (or at least most) of the current usages of the `DebugLocationTrackerFactory`.
- To have an `IInstaller` that supports installing dependencies given an `IKernel` and a `DebuggerSession`.
    - The installer should show the appropriate message to the user regarding whether the kernel or the debugger may need to be restarted.

To achieve the above, I will focus on making the following PRs in order.

- [x] (Urgent) Enable the Data Frame on the web (Fixing #9665)
  - By changing the `IDataViewerDependencyService` to receive an `IKernel` besides the `PythonEnvironment` that it receives today.
  - By adding a universal version of the `dataViewerDependencyService` that would only work if `IKernel` is provided and throw otherwise an error saying `Installation of "pandas" not supported while debugging. Please ensure you have "pandas" installed.`
  - By keeping the current node-specific version of the `dataViewerDependencyService` that would extract the interpreter from the `IKernel` if the `IKernel` is provided.
  - Besides ensuring the `dataViewerDependencyService` is registered on the web service registry, and removing the conditional that disables the Data Frame on the web.
- [x] #10639
  - This service would allow any other piece of code to run Python on the debugger.
  - This service will copy how the current `DebugLocationTrackerFactory` retrieves the frame ID.
  - This service will replace all the current usages of the `DebugLocationTrackerFactory` in the same PR.
- [x] #10640
  - Replace the usage of the `PythonEnvironment` (the interpreter) from the parameters of the `dataViewerDependencyService` methods and use the `DebugEvaluator` instead.
  - Evaluate code to get the interpreter path then use the old interpreter code.
- [x] #10641 
  - Enable installing `pandas` on the Debugger version of the `dataViewerDependencyService` by using the code underneath `%pip` and `%conda` in iPython.
  - Completely remove the usage of the interpreter from the `dataViewerDependencyService`.
  - If `pandas` needs to be installed (either because it's not yet installed or because the version is inadequate), show a prompt message saying that `You may need to restart the debugger` after the installation is complete.
- [x] #10642
[...]

PR #10825, opened 2022-07-15
Web version of the DataFrame viewer

### LINKED ISSUE #9665: DataFrame working in web extension

sadasant:
(The prior take on solving #9665 was PR: #10604)

Last month we tried to get the DataFrame viewer working on the web. In summary, we arrived at a set of bigger goals that are available here: #10638 . The first step actually does solve #9665 , but with an imperfect approach that is nonetheless good enough to get this feature out and get us started towards our longer term goals.

The first step, solved in this PR, goes as follows:

> Enable the Data Frame on the web (Fixing #9665)
>   - By changing the `IDataViewerDependencyService` to receive an `IKernel` besides the `PythonEnvironment` that it receives today.
>   - By adding a universal version of the `dataViewerDependencyService` that would only work if `IKernel` is provided and throw otherwise an error saying `Installation of "pandas" not supported while debugging. Please ensure you have "pandas" installed.`
>   - By keeping the current node-specific version of the `dataViewerDependencyService` that would extract the interpreter from the `IKernel` if the `IKernel` is provided.
>   - Besides ensuring the `dataViewerDependencyService` is registered on the web service registry, and removing the conditional that disables the Data Frame on the web.

This PR contains those changes, and includes unit tests.

Fixes #9665 

Feedback always appreciated!


src/webviews/extension-side/dataviewer/dataVIewerDependencyService.ts
```diff
@@ -0,0 +1,162 @@
...
+        sendTelemetryEvent(Telemetry.PythonModuleInstall, undefined, {
+            action: 'displayed',
+            moduleName: ProductNames.get(Product.pandas)!
+        });
+
+        const kernel = options.kernel;
+
+        let selection: string | undefined;
+        if (kernel) {
+            selection = this.isCodeSpace
```
sadasant, line 107: Just to understand, is it always the case that the browser version of the editor will be in codespaces?

src/webviews/extension-side/dataviewer/dataVIewerDependencyService.ts
```diff
@@ -0,0 +1,162 @@
...
+                ? Common.install()
+                : await this.applicationShell.showErrorMessage(
+                      DataScience.pandasRequiredForViewing(),
+                      { modal: true },
+                      Common.install()
```
sadasant, line 112: If I remove the check for `this.isCodeSpace` and run `showErrorMessage`, it does work in the web, but only in this scenario, after installing pandas, I need to restart the kernel to make it work. If I keep the `this.isCodeSpace` check, I do not need to restart the kernel after `pandas` gets installed. Does this make sense? It is weird to me. I wonder if I'm doing something wrong.

src/webviews/extension-side/dataviewer/dataVIewerDependencyService.ts
```diff
@@ -0,0 +1,162 @@
...
+ */
+@injectable()
+export class DataViewerDependencyService implements IDataViewerDependencyService {
+    constructor(
+        @inject(IApplicationShell) private readonly applicationShell: IApplicationShell,
+        @inject(IsCodeSpace) private isCodeSpace: boolean
+    ) {}
+
+    private kernelPackaging(kernel: IKernel): '%conda' | '%pip' {
+        const envType = kernel.kernelConnectionMetadata.interpreter?.envType;
```
sadasant, line 40: Although I have Conda configured, I'm getting `undefined` on `kernel.kernelConnectionMetadata.interpreter?.envType`. How do I ensure that Jupyter starts connected to the Conda environment? I'm starting Jupyter with:

```
conda activate base
jupyter notebook --no-browser --NotebookApp.allow_origin=* --NotebookApp.disable_check_xsrf=true
```

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    public async checkAndInstallMissingDependencies(options: IDataViewerDependencyServiceOptions): Promise<void> {
+        const kernel = options.kernel;
+        const interpreter = options.interpreter;
+
+        // Providing feedback as soon as possible.
+        if (!kernel && interpreter) {
+            sendTelemetryEvent(Telemetry.InterpreterNotSupportedOnTheWeb);
+            throw new Error(DataScience.interpreterNotSupportedOnTheWeb());
+        }
```
sadasant, line 55: Is it ok to throw if a developer provides an interpreter on the web and not a kernel? at least for now, while we work towards the debugger support.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,160 @@
...
+        sendTelemetryEvent(Telemetry.PythonModuleInstall, undefined, {
+            action: 'displayed',
+            moduleName: ProductNames.get(Product.pandas)!
+        });
+
+        const kernel = options.kernel;
+
+        let selection: string | undefined;
+        if (kernel) {
+            selection = this.isCodeSpace
```
sadasant, line 108: Just to understand, is it always the case that the browser version of the editor will be in codespaces? `this.isCodeSpace` is true when I run `openInBrowser`.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,160 @@
...
+ */
+@injectable()
+export class DataViewerDependencyService implements IDataViewerDependencyService {
+    constructor(
+        @inject(IApplicationShell) private readonly applicationShell: IApplicationShell,
+        @inject(IsCodeSpace) private isCodeSpace: boolean
+    ) {}
+
+    private kernelPackaging(kernel: IKernel): '%conda' | '%pip' {
+        const envType = kernel.kernelConnectionMetadata.interpreter?.envType;
```
sadasant, line 40: Although I have Conda configured, I'm getting `undefined` on `kernel.kernelConnectionMetadata.interpreter?.envType`. How do I ensure that Jupyter starts connected to the Conda environment? I'm starting Jupyter with:

```
conda activate base
jupyter notebook --no-browser --NotebookApp.allow_origin=* --NotebookApp.disable_check_xsrf=true
```

src/webviews/extension-side/dataviewer/dataViewerDependencyService.node.ts
```diff
@@ -39,7 +39,18 @@ export class DataViewerDependencyService implements IDataViewerDependencyService
...
+    public async checkAndInstallMissingDependencies(options: IDataViewerDependencyServiceOptions): Promise<void> {
+        const kernel = options.kernel;
+        let interpreter = options.interpreter;
+
+        if (!interpreter && kernel?.kernelConnectionMetadata.interpreter) {
+            interpreter = kernel.kernelConnectionMetadata.interpreter;
+        }
+
+        if (!interpreter) {
+            throw new Error('Kernel dependency installer is not available in Node.js. Use an interpreter.');
```
rchiodo, line 51: Seems like this should be localized? Does the user see this message?

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
rchiodo: If you're using remote then it's iffy whether or not we can find the interpreter data. Is there a reason you're using remote here?

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
rchiodo: Yeah seems fine to me but this version is not web specific. It's the generic one that works for both. You'd throw this in a web specific one if you made one.

If your intent is to have one of these functions for node and one for web, I think you should create this method as abstract here and then override it in a 'web' class.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
rchiodo: isWebExtension is more accurate to detect if in the web version.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,160 @@
...
+        const kernel = options.kernel;
+
+        let selection: string | undefined;
+        if (kernel) {
+            selection = this.isCodeSpace
+                ? Common.install()
+                : await this.applicationShell.showErrorMessage(
+                      DataScience.pandasRequiredForViewing(),
+                      { modal: true },
+                      Common.install()
```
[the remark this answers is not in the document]
rchiodo: Yeah that doesn't make any sense. The error message should have no effect on the state of the kernel.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
rchiodo: It looks to me like you're treating this whole class as the web version. It should be named that way then to force it to be web specific.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
sadasant: Well, I'm using remote because I'm on the web! Right?

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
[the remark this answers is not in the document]
sadasant: I have not been able to reproduce this again! I'll keep an eye on it.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
```diff
@@ -0,0 +1,121 @@
...
+                throw new Error(DataScience.failedToInstallPandas());
+            }
+        } else {
+            sendTelemetryEvent(Telemetry.UserDidNotInstallPandas);
+            throw new Error(DataScience.pandasRequiredForViewing());
+        }
+    }
+
+    async checkAndInstallMissingDependencies(kernel: IKernel): Promise<void> {
+        if (kernel && !kernel.session) {
```
DonJayamanne, line 100: checking if `kernel` is not undefined is unnecessary

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
```diff
@@ -0,0 +1,121 @@
...
+        if (kernel) {
+            selection = this.isCodeSpace
+                ? Common.install()
+                : await this.applicationShell.showErrorMessage(
+                      DataScience.pandasRequiredForViewing(),
+                      { modal: true },
+                      Common.install()
+                  );
+        } else {
+            await this.applicationShell.showErrorMessage(DataScience.pandasRequiredForViewing());
```
DonJayamanne, line 76: this is an impossible code path, kernel can never be undefined

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        if (!kernel) {
+            return;
```
DonJayamanne, line 80: not necessary, don't see how this is possible

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        const command = `${kernelPackaging(kernel)} install pandas`;
+
+        if (selection === Common.install()) {
+            try {
+                await this.execute(command, kernel);
+                sendTelemetryEvent(Telemetry.UserInstalledPandas);
+            } catch (e) {
+                sendTelemetryEvent(Telemetry.FailedToInstallPandas);
```
DonJayamanne, line 90: the, error is swallowed here.
we should log the error and also send it into the telemetry function.

DonJayamanne: if we don't log the original error, we'll never knows what went wrong

sadasant: Thank you, Don! This is great feedback! I will review my PR again using your perspective in these comments

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
sadasant: This is what happens when I take a PR out of draft a Friday night 😅 thank you! I'm reviewing my PR. Thank you for your patience 🙏

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
sadasant: Changed the throwing of the error to include the message: `throw new Error(DataScience.failedToInstallPandas().format(e.message));`. How does that sound?

DonJayamanne: Sorry should have been more specified,
this is what I'm suggesting `sendTelemetryEvent(Telemetry.UserInstalledPandas, undefined, undefined, e);`
This way we can use the same telemetry event (with a flag indicating it failed) and we have the error information as well.
Using error.message is risky as it could contain PII and the `sendTelemetryEvent` will only take the bits that won't contain PII

here's an example `sendTelemetryEvent(EventName.EXTENSION_LOAD, durations, props, ex);`

src/platform/common/utils/localize.ts
```diff
@@ -808,7 +808,7 @@ export namespace DataScience {
                 key: 'DataScience.pandasRequiredForViewing',
                 comment: ["{Locked='pandas'", 'This is the name of the pandas package']
             },
-            "Python package 'pandas' is required for viewing data."
+            "Python package 'pandas' is required for viewing data. Please ensure you have installed 'pandas' version {0} or above."
```
DonJayamanne, line 811: You need to update the package.nls.json file as well, coudlnt see that change in this PR

src/platform/common/utils/localize.ts
```diff
@@ -1396,6 +1396,18 @@ export namespace DataScience {
         );
     export const listOfFilesWithLinksThatMightNeedToBeRenamed = () =>
         localize('DataScience.listOfFilesWithLinksThatMightNeedToBeRenamed', 'File(s): {0} might need to be renamed.');
+    export const noActiveKernelSession = () =>
+        localize('DataScience.noActiveKernelSession', 'No no active kernel session.');
+    export const failedToGetVersionOfPandas = () =>
+        localize(
+            { key: 'DataScience.failedToGetVersionOfPandas', comment: ['{Locked="Pandas"}'] },
+            'Failed to get version of Pandas to use the Data Viewer'
```
DonJayamanne, line 1404: I think this message needs to end with a period.

DonJayamanne: Coudln't find changes to package.nls. in this PR for any of these. or is it in here and I'm missing it.

src/webviews/extension-side/dataviewer/constants.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+'use strict';
+
+export const pandasMinimumVersionSupported = '0.20.0';
```
DonJayamanne, line 6: I'd change this to be more descriptive or add comments.
```suggestion
export const pandasMinimumVersionSupportedByVariableViewer = '0.20.0';
```

src/webviews/extension-side/dataviewer/dataViewerDependencyService.node.ts
```diff
@@ -4,132 +4,51 @@
...
-            const response = await Promise.race([
-                this.installer.install(Product.pandas, interpreterToInstallDependenciesInto, tokenSource),
-                cancellationPromise
-            ]);
-            if (response === InstallerResponse.Installed) {
-                sendTelemetryEvent(Telemetry.UserInstalledPandas);
-            }
+    async checkAndInstallMissingDependencies(executionEnvironment: IKernel | PythonEnvironment): Promise<void> {
+        // IKernel and PythonEnvironment are only types, so I can't compare prototypes or instances of.
+        if (typeof (executionEnvironment as PythonEnvironment).sysPrefix === 'string') {
```
DonJayamanne, line 46: I thin a better way to test this is `if ('sysPrefix' in executionEnvironment && executionEnvironment.sysPrefix) {`
This way if we change the arguments tomorrow this line of code will automatically fail to compile (i.e. compilre will catch it) else something like an explicit cast will not be caught and could lead to bugs

src/webviews/extension-side/dataviewer/dataViewerDependencyService.node.ts
```diff
@@ -4,132 +4,51 @@
...
+            sendTelemetryEvent(Telemetry.DataViewerUsingInterpreter);
```
DonJayamanne, line 47: I think this telemetry should moved into the method `withInterpreter.checkAndInstallMissingDependnecies` , else if this method is called in 3 places tomorrow, we'll need to write the same code in 3 places to send the telemtry.

src/webviews/extension-side/dataviewer/dataViewerDependencyService.node.ts
```diff
@@ -4,132 +4,51 @@
...
-            const result = await launcher.exec(['-c', 'import pandas;print(pandas.__version__)'], {
-                throwOnStdErr: true,
-                token
-            });
-
-            return parseSemVer(result.stdout);
-        } catch (ex) {
-            traceWarning('Failed to get version of Pandas to use Data Viewer', ex);
-            return;
+            sendTelemetryEvent(Telemetry.DataViewerUsingKernel);
```
DonJayamanne, line 50: Same as previous comment (move sending telemetry into the method)

src/webviews/extension-side/dataviewer/dataViewerDependencyService.ts
```diff
@@ -0,0 +1,31 @@
...
+    constructor(
+        @inject(IApplicationShell) applicationShell: IApplicationShell,
+        @inject(IsCodeSpace) isCodeSpace: boolean
+    ) {
+        this.withKernel = new KernelDataViewerDependencyImplementation(applicationShell, isCodeSpace);
+    }
+
+    async checkAndInstallMissingDependencies(kernel: IKernel): Promise<void> {
+        sendTelemetryEvent(Telemetry.DataViewerUsingKernel);
+        return this.withKernel.checkAndInstallMissingDependencies(kernel);
```
DonJayamanne, line 29: This is what I would like to avoid, we shodulh't have to copy this code to send telemetry where ever this method is called, it should be done in the method itself.

src/webviews/extension-side/dataviewer/interpreterDataViewerDependencyImplementation.node.ts
```diff
@@ -0,0 +1,124 @@
...
+                return;
+            }
+
+            if (pandasVersion) {
+                if (pandasVersion.compare(pandasMinimumVersionSupported) > 0) {
+                    return;
+                }
+                sendTelemetryEvent(Telemetry.PandasTooOld);
+                // Warn user that we cannot start because pandas is too old.
+                const versionStr = `${pandasVersion.major}.${pandasVersion.minor}.${pandasVersion.build}`;
```
DonJayamanne, line 49: Why not just use `pandasVersions.raw` or the like, instead of formatting this manually.

src/webviews/extension-side/dataviewer/interpreterDataViewerDependencyImplementation.node.ts
```diff
@@ -0,0 +1,124 @@
...
+    private async installMissingDependencies(
+        interpreter: PythonEnvironment,
+        tokenSource: CancellationTokenSource
+    ): Promise<void> {
+        sendTelemetryEvent(Telemetry.PythonModuleInstall, undefined, {
+            action: 'displayed',
+            moduleName: ProductNames.get(Product.pandas)!,
+            pythonEnvType: interpreter?.envType
+        });
+        const selection = this.isCodeSpace
```
DonJayamanne, line 69: Woudln't it be better to have the code to compare the version and display messages and the like to be in a base class (or seprate class), and have separate implementations for getting version and installing the pacakges.
Else we have to copy this code for comparing versions and displaying messages into two or three places (one for kernel, one for interpreter, one for debugger).

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
```diff
@@ -0,0 +1,107 @@
...
+    private async installMissingDependencies(kernel: IKernel): Promise<void> {
+        sendTelemetryEvent(Telemetry.PythonModuleInstall, undefined, {
+            action: 'displayed',
+            moduleName: ProductNames.get(Product.pandas)!
+        });
+
+        let selection = this.isCodeSpace
+            ? Common.install()
+            : await this.applicationShell.showErrorMessage(
+                  DataScience.pandasRequiredForViewing().format(pandasMinimumVersionSupported),
```
DonJayamanne, line 64: This is the code I'd like to avoid duplicating, comparing versions, displaying messages and the like.
Basically have a single class to handle dependency management and different implementations for how we get versions and install the missing packages.

DonJayamanne: Note, tomorrow we could have another similar implementation for the debugger as well.

DonJayamanne (changes requested):
Some minor tweeks requested

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
sadasant: > Note, tomorrow we could have another similar implementation for the debugger as well.

Yup, that's the idea!

src/webviews/extension-side/dataviewer/interpreterDataViewerDependencyImplementation.node.ts
IanMatthewHuff: Debugger dependency checker (since it checks for an ipykernel version) will be super similar to this so that could be used as a guide to see how to abstract out the commonality.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/**
+ * Uses the Kernel to manage the dependencies of a Data Viewer.
+ */
+export class KernelDataViewerDependencyImplementation implements IDataViewerDependencyService {
+    constructor(private readonly applicationShell: IApplicationShell, private isCodeSpace: boolean) {}
+
+    protected async execute(command: string, kernel: IKernel): Promise<(string | undefined)[]> {
+        const outputs = await executeSilently(kernel.session!, command);
```
IanMatthewHuff, line 37: The assumption that the kernel has a session (with the !) seems like it might be sneaking around the type system a bit. I don't see, at least at this point, an explicit reason the session would be undefined. Per the usage of this function it look like it should maybe just take a Session as a non optional parameter? I don't believe anything else on the Kernel is used, and the caller should guarantee that there is a session.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
```diff
@@ -0,0 +1,107 @@
...
+    async checkAndInstallMissingDependencies(kernel: IKernel): Promise<void> {
+        if (!kernel.session) {
+            sendTelemetryEvent(Telemetry.NoActiveKernelSession);
+            throw new Error(DataScience.noActiveKernelSession());
+        }
+
+        const pandasVersion = await this.getVersion(kernel);
+
+        if (pandasVersion) {
+            if (pandasVersion.compare(pandasMinimumVersionSupported) > 0) {
```
IanMatthewHuff, line 94: Might be nice to have another telemetry event here? It's the type of thing that we can probably calculate using other events, but an explict Pandas === Ok event might help calculating how many Pandas failures we see.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
IanMatthewHuff: Yeah, I see the check at the start of the dependency check later in the code. Seems like after that you could just use the session.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
IanMatthewHuff: @sadasant just to check, your argument here is that we can get in one solid implementation here first, then take a second pass (once we add the second usage of it) to create a more general IInstaller? I don't think that I'm adverse to that, though it might mean that some of the choices we make now might be harder to generalize later.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
sadasant: This function is called by a function that checks for the session, but I like your idea! Passing the session directly, rather than the kernel. Thank you 🙏

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
```diff
@@ -0,0 +1,107 @@
...
+import { executeSilently } from '../../../kernels/helpers';
+import { IKernel } from '../../../kernels/types';
+import { parseSemVer } from '../../../platform/common/utils';
+import { IDataViewerDependencyService } from './types';
+import { pandasMinimumVersionSupported } from './constants';
+
+export const kernelGetPandasVersion =
+    'import pandas as _VSCODE_pandas;print(_VSCODE_pandas.__version__);del _VSCODE_pandas';
+
+function kernelPackaging(kernel: IKernel): '%conda' | '%pip' {
```
IanMatthewHuff, line 23: @DonJayamanne might know best here. But this will use %pip for envs like poetry. In those cases should we just not do the install and ask them to install manually?

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
sadasant: @IanMatthewHuff More or less. I'll say two arguments:

1. The debugger implementation might shake up our assumptions of what can be generalized and what can't.
2. The longer-term goal is not to have the installer logic in the dataViewerDependencyService, but in the `Installer`. Getting there means eventually merging the `IInstaller` with these `*DataViewerDependencyInstallers`.

Put in another way: The union between kernel+interpreter is likely going to be different than the union between kernel+interpreter+debugger, which is likely going to be different than the union between kernel+interpreter+debugger within `IInstaller`. Therefore, why try to formalize a more complex common API this early?

sadasant: (by "longer-term" I mean that my current focus is to get the DataFrame and the PlotViewer working, and then I'll resume with https://github.com/microsoft/vscode-jupyter/issues/10638 )

src/webviews/extension-side/dataviewer/dataViewerDependencyService.node.ts
sadasant: The closest I can do is `'sysPrefix' in executionEnvironment && typeof executionEnvironment.sysPrefix === 'string'`  (in its own function), to avoid having to set it to something every time we instantiate it in tests.

src/webviews/extension-side/dataviewer/interpreterDataViewerDependencyImplementation.node.ts
sadasant: The reason why this isn't done in the interpreter version is because `pandasVersions.raw` results in `0.20..` 🤔

src/webviews/extension-side/dataviewer/interpreterDataViewerDependencyImplementation.node.ts
sadasant: How does this look? https://github.com/microsoft/vscode-jupyter/pull/10825/commits/60dcbf9c6785a8e5841f58bc4b2afb5581491b3d as a first step. Can grow over time 🙂

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
DonJayamanne: > adverse to that, though it might mean that some of the choices we make now might be harder to generalize later.

I think this can be done now., 

> kernel+interpreter+debugger, which is likely going to be different than the union between

Not sure i agree about this, but anyways, for instance I can already see a lot of code duplication in this PR, but if this will be fixed in the next PR, then i'm fine with that.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
DonJayamanne: Yes agreed,

src/platform/common/utils/localize.ts
```diff
@@ -1396,6 +1396,18 @@ export namespace DataScience {
         );
     export const listOfFilesWithLinksThatMightNeedToBeRenamed = () =>
         localize('DataScience.listOfFilesWithLinksThatMightNeedToBeRenamed', 'File(s): {0} might need to be renamed.');
+    export const noActiveKernelSession = () =>
+        localize('DataScience.noActiveKernelSession', 'No no active kernel session.');
```
DonJayamanne, line 1400: I don't see this in package.nls.json?
Also not sure this needs to be localized, but not going to ask this to be removed, as this is an error case that should not occur.

DonJayamanne: Because when users recieve this error message, i'm 99% certain that they will not know what to do at that point in time, as the error message doesn't tell them anything apart from the fact that sometjhing is missing.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
sadasant: How about this? https://github.com/microsoft/vscode-jupyter/pull/10825/commits/c3341d788e63dff03d4f124a89bf57dfc72c592a

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
sadasant: what does that mean? how do I catch environments like poetry?

src/webviews/extension-side/dataviewer/interpreterDataViewerDependencyImplementation.node.ts
```diff
@@ -0,0 +1,125 @@
...
+            }
+
+            if (pandasVersion) {
+                if (pandasVersion.compare(pandasMinimumVersionSupportedByVariableViewer) > 0) {
+                    return;
+                }
+                sendTelemetryEvent(Telemetry.PandasTooOld);
+                // Warn user that we cannot start because pandas is too old.
+                const versionStr = `${pandasVersion.major}.${pandasVersion.minor}.${pandasVersion.build}`;
+                throw new Error(DataScience.pandasTooOldForViewingFormat().format(versionStr));
```
DonJayamanne, line 54: Shouldn't the prompt be displayed at this point in time?
I.e. if the version is too old, shouldn't we display the prmonpt and ask to install.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
```diff
@@ -0,0 +1,107 @@
...
+        if (await this.promptInstall()) {
+            try {
+                await this.execute(command, kernel);
+                sendTelemetryEvent(Telemetry.UserInstalledPandas);
+            } catch (e) {
+                sendTelemetryEvent(Telemetry.UserInstalledPandas, undefined, undefined, e);
+                throw new Error(DataScience.failedToInstallPandas());
+            }
+        } else {
+            sendTelemetryEvent(Telemetry.UserDidNotInstallPandas);
```
DonJayamanne, line 76: Sending the telemetry after installing and when failing, or when not required to install is still duplicated and feel this can be moved into the base class.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
```diff
@@ -0,0 +1,107 @@
...
+
+        const pandasVersion = await this.getVersion(kernel);
+
+        if (pandasVersion) {
+            if (pandasVersion.compare(pandasMinimumVersionSupportedByVariableViewer) > 0) {
+                return;
+            }
+            sendTelemetryEvent(Telemetry.PandasTooOld);
+            // Warn user that we cannot start because pandas is too old.
+            const versionStr = `${pandasVersion.major}.${pandasVersion.minor}.${pandasVersion.build}`;
```
DonJayamanne, line 99: This code can also be moved into the base class, all the implementations need to do is get the version and return it, and the base class can check if a version was available or not.

DonJayamanne (commented):
I still think a lof of the code can be moved into a base class, even if debugger implementation comes in later.

src/platform/common/utils/localize.ts
sadasant: I'll de-localize it!

src/webviews/extension-side/dataviewer/interpreterDataViewerDependencyImplementation.node.ts
```diff
@@ -0,0 +1,125 @@
...
+            // Always pass a cancellation token to `install`, to ensure it waits until the module is installed.
+            const response = await Promise.race([
+                this.installer.install(Product.pandas, interpreterToInstallDependenciesInto, tokenSource),
+                cancellationPromise
+            ]);
+            if (response === InstallerResponse.Installed) {
+                sendTelemetryEvent(Telemetry.UserInstalledPandas);
+            }
+        } else {
+            sendTelemetryEvent(Telemetry.UserDidNotInstallPandas);
```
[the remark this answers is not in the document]
rchiodo: I believe Don is just suggesting that shared code should be in a base class. Simply to eliminate duplicate code. The design should be relatively close to the same I'd think.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
sadasant: Once I have the picture clear with the debugger, the plan is to move all of the install code into the `IInstaller`. That will have to be fully generic for the three contexts: interpreter, kernel and debugger.

src/webviews/extension-side/dataviewer/kernelDataViewerDependencyImplementation.ts
rchiodo: I think it's unlikely this will ever pick 'conda' either. When we have an IKernel, we're in a remote situation. Otherwise we'd be using the interpreter.

I think we can leave this as is for now, and then figure out some other way to detect the presence of the package manager later.

src/webviews/extension-side/dataviewer/interpreterDataViewerDependencyImplementation.node.ts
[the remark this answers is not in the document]
sadasant: So, what do you think of this?

```ts
export abstract class BaseDataViewerDependencyImplementation<TExecuter> implements IDataViewerDependencyService {
    protected abstract _getVersion(executer: TExecuter): Promise<string | undefined>;
    protected async getVersion(executer: TExecuter): Promise<SemVer | undefined> {
        try {
            const version = await this._getVersion(executer);
            return typeof version === 'string' ? parseSemVer(version) : version;
        } catch (e) {
            traceWarning(DataScience.failedToGetVersionOfPandas(), e.message);
            return;
        }
    }
```

Then

```ts
export class KernelDataViewerDependencyImplementation extends BaseDataViewerDependencyImplementation<IKernel> {
    protected async _getVersion(kernel: IKernelWithSession): Promise<string | undefined> {
        const outputs = await this.execute(kernelGetPandasVersion, kernel);
        return outputs.map((text) => (text ? text.toString() : undefined)).find((item) => item);
    }
```

I'm thinking I'd rather focus on telemetry wrappers than on splitting the steps into common functions.

Feedback appreciated!

DonJayamanne (approved):

src/webviews/extension-side/dataviewer/interpreterDataViewerDependencyImplementation.node.ts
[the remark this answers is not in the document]
sadasant: Perhaps more like this?

```ts
    protected abstract _getVersion(executer: TExecuter): Promise<string | undefined>;

    @traceDecoratorWarn(DataScience.failedToGetVersionOfPandas())
    protected async getVersion(executer: TExecuter): Promise<SemVer | undefined> {
        const version = await this._getVersion(executer);
        return typeof version === 'string' ? parseSemVer(version) : version;
    }
```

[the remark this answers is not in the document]
sadasant: How does this look? https://github.com/sadasant/vscode-jupyter/pull/1

cc: @DonJayamanne , @IanMatthewHuff , @rchiodo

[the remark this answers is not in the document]
sadasant: I merged that PR!

rchiodo (approved):

sadasant:
Now I'm getting two different failures, one regarding DataScience Exports, and another one where the `Log test results` job `Failed to print test summary`. I'm assuming this PR is good to go. I'll merge it.

PR #10833, opened 2022-07-15
Temporary execution is created unnecessarily

### LINKED ISSUE #10831: Temporary execution is created unnecessarily
For https://github.com/microsoft/vscode/issues/150924, the extension is creating a temporary execution in case it needs to update outputs, but it turns out to be a case where the execution object isn't needed. Doing this creates unnecessary work on vscode's side

roblourens:
Fixes #10831


amunger (approved):

roblourens:
Note that I moved lazy.ts out of vscode-path to somewhere that looks like the right spot for a general util, let me know if that's not right.

rchiodo (approved):

IanMatthewHuff (approved):

rchiodo:
> Note that I moved lazy.ts out of vscode-path to somewhere that looks like the right spot for a general util, let me know if that's not right.

Actually I had put it there so we could try and keep it in sync with the VS code version. But maybe we don't need to.

roblourens:
Oh, I didn't realize that was just copied out of vscode. I can put it back, whatever you guys think.

rchiodo:
I'm okay with you moving it. I doubt we'll ever update it.

Issue #10399, opened 2022-06-09
Improve CI time


rebornix:
This issue captured ideas brought up in offline discussions with @rchiodo for improving the testing experience in the repo, including faster CI time and tighter feedback loop for changes (locally)

* Shared document/kernel for test suite when appropriate. We currently open and close document for every individual test, which will launch kernel for every single. The kernel startup contributes to main portion of our CI time. When document open/close and kernels state don't matter in a test suite, we could see how we can reuse the same document and kernel for faster testing.
* A `launch.json` option (with properly tuned envs) to test non-conda tests locally. Currently the tests on a reasonably powerful machine takes too long and some of us rely on the CI for the test results.

rchiodo:
We could also work on improving
- Compile time
- Conda install time
- Not install .NET/Julia for python tests

DonJayamanne:
Partial improvements made here https://github.com/microsoft/vscode-jupyter/pull/11867

DonJayamanne:
Closing this as majority of the work has been done (faster compilation, faser tests, etc)
If we can improve more, I think it can be done as a regular debt task.
This used to take 45 minutes or longer back a year ago, now its around 15 minutes

PR #10847, opened 2022-07-18
Skip unnecessary controller refreshes

rchiodo:
For #10399 

Was running some poor mans profiling (just trace statements in the debugger) and noticed we spend a lot of time continually refreshing the kernel list. 

This PR attempts to minimize some of the refreshes that occur.

### DIFF
--- a/src/notebooks/controllers/controllerLoader.ts
+++ b/src/notebooks/controllers/controllerLoader.ts
@@ -62,7 +62,11 @@ export class ControllerLoader implements IControllerLoader, IExtensionSyncActiva
         );
 
         // Make sure to reload whenever we do something that changes state
-        const forceLoad = () => this.loadControllers(true);
+        const forceLoad = () => {
+            this.loadControllers(true)
+                .then(noop)
+                .catch((ex) => traceError('Error reloading notebook controllers', ex));
+        };
         this.serverUriStorage.onDidChangeUri(forceLoad, this, this.disposables);
 
         // If we create a new kernel, we need to refresh if the kernel is remote (because
@@ -71,7 +75,7 @@ export class ControllerLoader implements IControllerLoader, IExtensionSyncActiva
         // to check for remote if the future when we support live sessions on local
         this.kernelProvider.onDidStartKernel((k) => {
             if (isRemoteConnection(k.kernelConnectionMetadata)) {
-                forceLoad().ignoreErrors();
+                forceLoad();
             }
         });
 
--- a/src/notebooks/controllers/controllerLoader.ts
+++ b/src/notebooks/controllers/controllerLoader.ts
@@ -50,22 +50,12 @@ export class ControllerLoader implements IControllerLoader, IExtensionSyncActiva
     }
 
     public activate(): void {
-        let timer: NodeJS.Timeout | number | undefined;
         this.interpreters.onDidChangeInterpreters(
-            () => {
-                if (timer) {
-                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                    clearTimeout(timer as any);
+            async () => {
+                // Don't do anything if the interpreter list is still being refreshed
+                if (!this.interpreters.refreshing) {
+                    await this.loadControllers(true);
                 }
-                timer = setTimeout(
-                    () =>
-                        this.loadControllers(true).catch((ex) =>
-                            traceError('Failed to re-query python kernels after changes to list of interpreters', ex)
-                        ),
-                    // This hacky solution should be removed in favor of https://github.com/microsoft/vscode-jupyter/issues/7583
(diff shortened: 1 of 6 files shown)


src/notebooks/controllers/controllerLoader.ts
```diff
@@ -74,12 +64,25 @@ export class ControllerLoader implements IControllerLoader, IExtensionSyncActiva
...
         this.serverUriStorage.onDidChangeUri(forceLoad, this, this.disposables);
-        this.kernelProvider.onDidStartKernel(forceLoad, this, this.disposables);
+
+        // If we create a new kernel, we need to refresh if the kernel is remote (because
+        // we have live sessions possible)
+        // Note, this is a perf optimization for right now. We should not need
+        // to check for remote if the future when we support live sessions on local
+        this.kernelProvider.onDidStartKernel((k) => {
+            if (isRemoteConnection(k.kernelConnectionMetadata)) {
+                forceLoad().ignoreErrors();
```
IanMatthewHuff, line 74: the ignoreErrors here looks like it might be a handling change from the previous code that might not be intended. From the comment on line 112 might be errors getting thrown by loadControllers that now won't bubble up.

src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -121,20 +121,17 @@ export class ControllerPreferredService implements IControllerPreferredService,
...
-                    preferredSearchToken.token,
-                    'ignoreCache',
-                    preferredInterpreter,
-                    serverId
-                );
-
                 // If we didn't find an exact match in the cache, try awaiting for the non-cache version
                 if (!preferredConnection) {
-                    ({ preferredConnection } = await ignoreCacheFindPreferredPromise);
+                    // Don't start this ahead of time to save some CPU cycles
```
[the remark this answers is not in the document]
rchiodo: No that's not true. The cache is updated elsewhere. This code is only for searching for a preferred connection. The kernel 'cache' is updated by the controller loader every time it refreshes.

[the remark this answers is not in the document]
rchiodo: I'd argue you don't need to update the cache if there's an exact match.

src/notebooks/controllers/controllerLoader.ts
rchiodo: Although all of the handlers are that way. Maybe it would be better to make forceLoad at least log the error and then have it not return a promise.

src/platform/api/pythonApi.ts
```diff
@@ -468,4 +485,32 @@ export class InterpreterService implements IInterpreterService {
             })
             .catch(noop);
     }
+    private getRefreshPromise(): Promise<void> | undefined {
+        if (this.canFindRefreshPromise === undefined) {
+            this.canFindRefreshPromise = false;
+            const api = this.tryGetApi();
+            if (!api) {
+                this.getApi()
```
amunger, line 493: `tryGetApi` will already have called `getApi`, so this code block doesn't feel necessary

rchiodo: The reason it needs to be called is to force the update of the 'canFindRefreshPromise' which getApi doesn't do.

rchiodo: The then clause.

src/notebooks/controllers/controllerPreferredService.ts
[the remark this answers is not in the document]
IanMatthewHuff: > I'd argue you don't need to update the cache if there's an exact match.

Well we do call the active interpreter an exact match. But due to what you said above about the cache being updated else where I don't think it matters, so good as is.

src/platform/api/pythonApi.ts
rchiodo: getRefreshPromise technically has to be synchronous. So it can't await because then the function itself would be part of the refresh promise.

This makes the order like so:

1. Somebody calls 'refreshing' getter
2. 'Refreshing' getter calls this function to get a refresh promise or not
3. canFindRefreshPromise is undefined
4. canFindRefreshPromise is set to false
5. API is not initialized, so trygetApi returns undefined
6. canFindRefreshPromise is still false, so undefined is returned from the function
7. A promise is scheduled that updates canFindRefreshPromise
8. Next call to 'refreshing' uses the new value.

The root of all this is the weird 'getRefreshPromise() : Promise<void> | undefined that the python extension has. If they'd just always returned a promise, and made the promise resolve when doing refreshing, this could have been a lot simpler.

amunger: I see where I'm confused - in step 5, tryGetApi sees that the api is undefined so it calls getApi, but **doesn't wait for it**. 
That will cause 2 calls to `this.apiProvider.getApi()` per call to this function though, right? Is it fine to stack up those calls?

Issue #10359, opened 2022-06-08
Refactor how env variables are loaded and passed into kernels and processes spawned


DonJayamanne:
We have a lot of places & different ways we load env variables from .env files & merges them with the activated Python envs & then we also get these env variables from the Python extension.

We also have a few issues where these env varaibles are getting cached when they shouldn't or we're not loading the updated .env files.
To get OCAML & a couple of other kernels working we need to load env variables form the shell and merge them again with the .env files.

Before we do that, I think we need to first fix up how env variables are parsed, loaded, merged, cached.

Required for :
* vscode not picking up environment variables (when opened from terminal)
* https://github.com/microsoft/vscode-jupyter/issues/9682
* https://github.com/microsoft/vscode-jupyter/issues/9682
* https://github.com/microsoft/vscode-jupyter/issues/9691 (load env variables from a specific file when dealing with Python kernels)
* https://github.com/microsoft/vscode-jupyter/issues/9774
* https://github.com/microsoft/vscode-jupyter/issues/10392
* https://github.com/microsoft/vscode-jupyter/issues/10755

Other possible issues:
* https://github.com/microsoft/vscode-jupyter/issues/9917
* https://github.com/microsoft/vscode-jupyter/issues/10392
* https://github.com/microsoft/vscode-jupyter/issues/1722
* #10392

DonJayamanne:
Personal notes:

- KernelEnvironmentVariablesService (one method to get kernel env vars)
- IEnvironmentActivationService (one method to get activated Python env vars)
- IEnvironmentVariablesService (helper, parsing .env file and merging vars)
- IEnvironmentVariablesProvider (helper, event when env vars change and ability to get env vars and custom env vars)

## PythonApi.getActivatedEnvironmentVariables

- Does this take .env variables into account
- Does this merge process.env into the return value
- Is this cached?
- If user updates .env file, will this be updated?
- Is user updates shell env variables, will this be updated?

## KernelEnvironmentVariablesService

- `getEnvironmentVariables <Top level>`
  - Used in one place to get the environment variables of the kernel when starting raw kernel process.
  - This calls `KernelEnvironmentVariablesService.getCustomEnvironmentVariables` and `IEnvironmentActivationService.getActivatedEnvironmentVariables` and merges the two results, also with the process.env.
    TODO: Won't both already have been merged with `process.env`.
  - Special code for handling `PYTHONPATH` and `PATH|Path` variables.
[...]

PR #10852, opened 2022-07-18
Cache kernels for a specific version of extension

DonJayamanne:
Related to #10359

Found a problem in our code:
* The cached list of raw kernels is pointing to kernelspec.json files in the extensions directory.
* So assume you have version 1 of extension installed
* Now you update to version 2, at this point the cache still points to version 1 and the kernelspec.json files are in the directory version 1.
* Those files in directory for version 1 could get deleted by VS Code at any point in time, as thats an old version of the extension and user has now installed version 2.

This PR ensures we don't use the cache when users updated their exensions from version 1 to version 2, else they end up using kernelspecs.json files that could get deleted at some point in time.
At worst this adds a slight delay in loading the kernels, however we have a lot of other cache points for interpeters and the liek to ensure this delay in loading kernels isn't too slow.

### DIFF
--- a/src/kernels/kernelFinder.base.ts
+++ b/src/kernels/kernelFinder.base.ts
@@ -282,6 +282,15 @@ export abstract class BaseKernelFinder implements IKernelFinder {
                 key,
                 { kernels: [], extensionVersion: '' }
             );
+
+            /**
+             * The cached list of raw kernels is pointing to kernelSpec.json files in the extensions directory.
+             * Assume you have version 1 of extension installed.
+             * Now you update to version 2, at this point the cache still points to version 1 and the kernelSpec.json files are in the directory version 1.
+             * Those files in directory for version 1 could get deleted by VS Code at any point in time, as thats an old version of the extension and user has now installed version 2.
+             * Hence its wrong and buggy to use those files.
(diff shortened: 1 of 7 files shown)


src/kernels/kernelFinder.base.ts
```diff
@@ -304,9 +309,36 @@ export abstract class BaseKernelFinder implements IKernelFinder {
...
+            this.removeOldCachedItems(),
+            ,
+            this.globalState.update(key, { kernels: serialized, extensionVersion: this.env.extensionVersion })
+        ]);
+    }
+    /**
+     * The old cached items can be quite large and we should clear them if we no longer need them.
+     */
+    private async removeOldCachedItems(): Promise<void> {
+        await Promise.all(
```
rchiodo, line 325: Maybe a comment as to why the cache can't survive version update?

IanMatthewHuff (approved):

amunger:
What is the result of the kernelspec being deleted when it was cached?

DonJayamanne:
> What is the result of the kernelspec being deleted when it was cached?

When running kernels things could go wrong, don't know exactly what can go wrong, but this isn't how the code is structured, we don't expect users to delete files in extension folder, same applies for kernelspecs which are in the extenssions folder.

PR #10853, opened 2022-07-18
Unify node/web kernel and remove iw reference in kernel component

rebornix:
Attempt to unify `kernel.web/node.base`. While trying to merge them, I found that the major difference between them is how we should generate start up code for interactive window debugging. It feels wrong that kernel needs to be aware of the logic of interactive window debugging so this PR tries to switch this to provider model.

@DonJayamanne let's meet offline and discuss about this PR.


src/kernels/kernelStartupCodeProvider.node.ts
```diff
@@ -0,0 +1,81 @@
...
+                !this.configService.getSettings(undefined).forceIPyKernelDebugger
+            )
+        ) {
+            return [];
+        }
+
+        if (
+            isLocalConnection(kernel.kernelConnectionMetadata) ||
+            isLocalHostConnection(kernel.kernelConnectionMetadata)
+            // && kernel.kernelConnectionMetadata.kind !== 'connectToLiveRemoteKernel' // Skip for live kernel. Don't change current directory on a kernel that's already running
```
rebornix, line 40: After moving the multiple `if` expressions to the same place, TS is now telling me this check is redundant.

src/interactive-window/debugger/startupCodeProvider.web.ts
```diff
@@ -0,0 +1,43 @@
...
+import { inject, injectable } from 'inversify';
+import { Uri } from 'vscode';
+import { getAssociatedNotebookDocument } from '../../kernels/helpers';
+import { IKernel, IStartupCodeProvider, StartupCodePriority } from '../../kernels/types';
+import { InteractiveWindowView } from '../../platform/common/constants';
+import { IFileSystem } from '../../platform/common/platform/types';
+import { IExtensionContext } from '../../platform/common/types';
+
+@injectable()
+export class InteractiveWindowDebuggingStartupCodeProvider implements IStartupCodeProvider {
```
rchiodo, line 13: Seems like these could be the same. Just use the fs.readFile on the extension URI for both.

DonJayamanne: From what I remember, this same code applies to desktop extension as well, when debugging using remote jupyter kernels. Hence we need this same code for non-web extension as well

src/kernels/kernel.ts
```diff
@@ -146,12 +147,14 @@ export abstract class BaseKernel implements IKernel {
...
         outputTracker: CellOutputDisplayIdTracker,
+        protected readonly workspaceService: IWorkspaceService,
         private readonly statusProvider: IStatusProvider,
         public readonly creator: KernelActionSource,
-        protected readonly context: IExtensionContext,
-        formatters: ITracebackFormatter[]
+        context: IExtensionContext,
+        formatters: ITracebackFormatter[],
+        private readonly startupCodeProviders: IStartupCodeProvider[],
+        private readonly sendTelemetryForPythonKernelExecutable: () => Promise<void>
```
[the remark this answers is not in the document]
DonJayamanne: here's the issue https://github.com/microsoft/vscode-jupyter/issues/10863

src/interactive-window/debugger/startupCodeProvider.web.ts
DonJayamanne: aaah, I can see that this is the same code in `InteractiveWindowDebuggingStartupCodeProvider.node & .web`
Yeah, would prefer if we could have the same class to do both, rather than having two seprate classes with the same logic, then the logic for debugging remote jupyter kernels is in one place, instead of being split across two files.

rebornix: yep, agree.

src/interactive-window/debugger/startupCodeProvider.ts
```diff
@@ -0,0 +1,55 @@
...
+    constructor(
+        @inject(IConfigurationService) private readonly configService: IConfigurationService,
+        @inject(IFileSystem) private readonly fs: IFileSystem,
+        @inject(IExtensionContext) private readonly context: IExtensionContext,
+        @inject(IsWebExtension) private readonly isWebExtension: boolean
+    ) {}
+
+    async getCode(kernel: IKernel): Promise<string[]> {
+        if (!this.isWebExtension) {
+            const useNewDebugger = this.configService.getSettings(undefined).forceIPyKernelDebugger === true;
```
rebornix, line 26: Now I merged the code for both Web and Node. I'm not sure though if the setting check on Node is still necessary but I kept it @rchiodo

src/kernels/kernel.ts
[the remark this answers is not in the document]
rebornix: @DonJayamanne proposal in #10863 is better 👍

src/kernels/kernelProvider.node.ts
```diff
@@ -51,24 +51,32 @@ export class KernelProvider extends BaseKernelProvider {
         const resourceUri = notebook?.notebookType === InteractiveWindowView ? options.resourceUri : uri;
         const waitForIdleTimeout = this.configService.getSettings(resourceUri).jupyterLaunchTimeout;
         const interruptTimeout = this.configService.getSettings(resourceUri).jupyterInterruptTimeout;
-        const kernel = new Kernel(
+        let kernel: Kernel;
+        kernel = new Kernel(
```
sadasant, line 55: why separate these two lines? 🤔

src/interactive-window/debugger/startupCodeProvider.ts
rchiodo: Yeah this is really confusing. Maybe a comment? The addRunCellHook is necessary when using ipykernel 6 with the old interactive window debugger.

The old interactive window debugger only supported local though.

rchiodo (approved):

src/kernels/kernelStartupCodeProvider.node.ts
```diff
@@ -0,0 +1,80 @@
...
+        @inject(IConfigurationService) private readonly configService: IConfigurationService,
+        @inject(IFileSystem) private readonly fs: IFileSystem,
+        @inject(IWorkspaceService) private readonly workspaceService: IWorkspaceService
+    ) {}
+
+    async getCode(kernel: IKernel): Promise<string[]> {
+        if (
+            !(
+                isLocalConnection(kernel.kernelConnectionMetadata) &&
+                !this.configService.getSettings(undefined).forceIPyKernelDebugger
```
sadasant, line 31: is it necessary to pass `undefined` as a parameter?

src/kernels/kernelStartupCodeProvider.node.ts
```diff
@@ -0,0 +1,80 @@
...
+                        kernel.resourceUri,
+                        this.workspaceService,
+                        this.configService.getSettings(kernel.resourceUri)
+                    )
+                );
+                if (suggestedDir && (await this.fs.exists(suggestedDir))) {
+                    traceInfo('UpdateWorkingDirectoryAndPath in Kernel');
+                    return this.getChangeDirectoryCode(kernel, suggestedDir);
+                }
+            }
```
sadasant, line 65: I'm not suer how much it would make sense to merge the tow ways to resolve the suggested dir, but I feel that with a new local method the code can be much simpler to parse. Something like:

```ts
private async resolveDirectory(kernel: IKernel, path?: Uri): Promise<Uri | undefined> {
    let dir: Uri | undefined;
    if (!path) {
        dir = await calculateWorkingDirectory(
            this.configService,
            this.workspaceService,
            this.fs,
            kernel.resourceUri
        )
    } else {
        dir = Uri.file(
            expandWorkingDir(
                getFilePath(path),
                kernel.resourceUri,
                this.workspaceService,
                this.configService.getSettings(kernel.resourceUri)
            )
        );
    }
    if (dir && (await this.fs.exists(dir))) {
        return dir;
    }
    return;
}
```

Would allow us to do something like:

```ts
        if (
             isLocalConnection(kernel.kernelConnectionMetadata) ||
             isLocalHostConnection(kernel.kernelConnectionMetadata)
         ) {
             let suggestedDir = await this.resolveDirectory(kernel);
             if (kernel.resourceUri && (await this.fs.exists(kernel.resourceUri))) {
                 suggestedDir = await this.resolveDirectory(kernel, suggestedDir);
             }
             if (suggestedDir) {
                 traceInfo('UpdateWorkingDirectoryAndPath in Kernel');
                 return this.getChangeDirectoryCode(kernel, suggestedDir);
             }
         }
```

sadasant: if this doesn't look valuable, feel free to resolve it!

DonJayamanne (approved):

src/kernels/kernelStartupCodeProvider.node.ts
rebornix: Thanks, I'll consider having a new PR taking this suggestion!

PR #10854, opened 2022-07-18
Add descriptions to classes in IW folder

rchiodo:

Also consolidated the interactiveWindowCommandListener class into the commands/commandRegistry. They were doing the same thing so put them into one class.


src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -15,7 +15,11 @@ import { getInteractiveCellMetadata } from '../../../interactive-window/helpers'
 import { KernelDebugAdapterBase } from '../../../kernels/debugger/kernelDebugAdapterBase';
 import { InteractiveCellMetadata } from '../../editor-integration/types';
 import { IDebugService } from '../../../platform/common/application/types';
-
+/**
+ * KernelDebugAdapter listens to debug messages in order to translate file requests into real files
+ * (Interactive Window generally executes against a real file)
+ * Files from the kernel are pointing to cells, whereas the user is looking at a real file.
```
rebornix, line 21: Does it mean that the difference between notebook debugging and interactive window debugging is IW has one more step: mapping DAP file requests from cell uris to file uri (python file)?

src/interactive-window/editor-integration/codewatcher.ts
```diff
@@ -53,6 +53,10 @@ function getIndex(index: number, length: number): number {
     }
 }
 
+/**
+ * CodeWatchers watch source files, provide code lenses in them, and handle the execution of
+ * the code lenses.
+ */
```
rebornix, line 59: The name for this module is `CodeWatcher`, does it listen to `TextDocument` content change and parse the document incrementally, or it scans the whole document each time when the content changes?

src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
rchiodo: Pretty much.

src/interactive-window/editor-integration/codewatcher.ts
rchiodo: It's actually more complicated than that. There's an update required event for the code lens factory:
https://github.com/microsoft/vscode-jupyter/blob/6f9fd031aeab67741fc36c1ee2d91a002a80e500/src/interactive-window/editor-integration/codewatcher.ts#L91

That fires when the execution count on the interactive window changes (well because the goto code lens needs to update)

And whenever VS code asks for code lenses.

rchiodo: It doesn't actually generate the code lenses itself. That's done by the codelensfactory

rchiodo: That's here:
https://github.com/microsoft/vscode-jupyter/blob/cf7da9f9c9d43f621d65c91e46c4a0847ff11cef/src/interactive-window/editor-integration/codeLensFactory.ts#L90

src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
rchiodo: At least once we switch to using the IPykernel6 debugger. The old debugger is slightly different as it does attach and detach at key spots and loads debugpy itself.

rchiodo:
@microsoft/datascience-vsc-team-reviewers does anybody want any changes?

sadasant (approved):

PR #10855, opened 2022-07-18
Documentation on all missing classes in kernels folder

rchiodo:

Added a blurb to every kernel class that was missing one


sadasant (approved):
I tried to find reasons to give feedback but this really looks good to me!

IanMatthewHuff (approved):

PR #10876, opened 2022-07-20
Use python.envFile to get env file name when running Python code

### LINKED ISSUE #9691: python.envFile is not respected by Python code executed in Notebook and IW
### Applies To

- [X] Notebooks (.ipynb files)

### What happened?

Environment variables defined in an `.env` file in the workspace root are available in notebooks, but it is surprising that only root `.env` files are supported, rather than using the [`python.envFile` setting](https://code.visualstudio.com/docs/python/environments#_environment-variable-definitions-file) like other VSCode Python functionality.

The `.env` file path appears to be hardcoded [here](https://github.com/microsoft/vscode-jupyter/blob/5a91b9cdd946903a92610cb630221e78d3744038/src/platform/common/variables/environmentVariablesProvider.node.ts#L70).

### VS Code Version

1.66.2

### Jupyter Extension Version

v2022.3.1000901801

### Jupyter logs

_No response_

_No response_

_No response_

_No response_

_No response_

DonJayamanne:
Fixes #9691

### DIFF
--- a/src/platform/common/variables/customEnvironmentVariablesProvider.node.ts
+++ b/src/platform/common/variables/customEnvironmentVariablesProvider.node.ts
@@ -13,6 +13,7 @@ import { traceDecoratorVerbose, traceError, traceInfoIfCI, traceVerbose } from '
 import { SystemVariables } from './systemVariables.node';
 import { disposeAllDisposables } from '../helpers';
 import { IPythonExtensionChecker } from '../../api/types';
+import { getDisplayPath } from '../platform/fs-paths';
 
 const CACHE_DURATION = 60 * 60 * 1000;
 @injectable()
@@ -70,6 +71,7 @@ export class CustomEnvironmentVariablesProvider implements ICustomEnvironmentVar
         this.trackedWorkspaceFolders.add(workspaceFolderUri ? workspaceFolderUri.fsPath : '');
 
         const envFile = this.getEnvFile(workspaceFolderUri, purpose);
+        traceInfoIfCI(`Env File => ${getDisplayPath(Uri.file(envFile))}`);
         this.createFileWatcher(envFile, workspaceFolderUri);
         return this.envVarsService.parseFile(envFile, process.env);
     }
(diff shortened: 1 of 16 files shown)


src/platform/common/variables/customEnvironmentVariablesProvider.node.ts
```diff
@@ -80,19 +91,37 @@ export class CustomEnvironmentVariablesProvider implements ICustomEnvironmentVar
...
-            this.disposables.push(envFileWatcher.onDidCreate(() => this.onEnvironmentFileCreated(workspaceFolderUri)));
-            this.disposables.push(envFileWatcher.onDidDelete(() => this.onEnvironmentFileChanged(workspaceFolderUri)));
+            envFileWatcher.onDidChange(() => this.onEnvironmentFileChanged(workspaceFolderUri), this, this.disposables);
+            envFileWatcher.onDidCreate(() => this.onEnvironmentFileCreated(workspaceFolderUri), this, this.disposables);
+            envFileWatcher.onDidDelete(() => this.onEnvironmentFileChanged(workspaceFolderUri), this, this.disposables);
         }
     }
-    private async _getEnvironmentVariables(resource?: Uri): Promise<EnvironmentVariables> {
-        let customEnvVars = await this.getCustomEnvironmentVariables(resource);
+    private getEnvFile(workspaceFolderUri: Uri, purpose?: 'RunPythonCode') {
```
rchiodo, line 107: When do we not want to 'runpythoncode'? I feel like if the opposite condition was defined as another string, it would make the use of this more obvious.

rchiodo (approved):

src/test/platform/interpreter/common/variables/customEnvironmentVariablesProvider.node.vscode.test.ts
```diff
@@ -0,0 +1,231 @@
...
+import { IEnvironmentVariablesService } from '../../../../../platform/common/variables/types';
+import { IS_REMOTE_NATIVE_TEST } from '../../../../constants';
+import { initialize } from '../../../../initialize';
+import * as fs from 'fs-extra';
+import { EXTENSION_ROOT_DIR_FOR_TESTS } from '../../../../constants.node';
+import * as dedent from 'dedent';
+import { IPythonExtensionChecker } from '../../../../../platform/api/types';
+import { createEventHandler } from '../../../../common';
+
+suite('Custom Environment Variables Provider', () => {
```
amunger, line 21: could these be unit tests? Doesn't seem to requre vscode to be running

amunger (approved):

src/test/platform/interpreter/common/variables/customEnvironmentVariablesProvider.node.vscode.test.ts
DonJayamanne: I guess this all depends on the definition of a unit test, should evertying be mocked? If yes, then thats not a good test.

I could mock a lot of things in the test, like reading and parsing the .env file and creating and editing files and then emitting file change evengts, etc... thats a heck of a lot to mock, and the effort to write the test using the real bits is significantly simpler and gets the job done.

basically I want to the the outputs for given inputs & that's exactly what this test it doing
& I rely on vscode here as I'm using the VS Code API to create the file watcher and i don't want to mock that, too much work and not worth it at all.

Here we have 6 test and everything runs in <1s.

DonJayamanne: @amunger looks like i'll have to create unit tests and mock the vscode classes, as the file change detection doesn't work on CI, possibly due to a large number of files or other.

PR #10878, opened 2022-07-20
Documentation for notebook classes

rchiodo:

### DIFF
--- a/src/notebooks/controllers/controllerDefaultService.ts
+++ b/src/notebooks/controllers/controllerDefaultService.ts
@@ -18,7 +18,7 @@ import {
 } from './types';
 
 /**
- * Determines the 'default' kernel for a notebook
+ * Determines the 'default' kernel for a notebook. Default is what kernel should be used if there's no metadata in a notebook.
  */
 @injectable()
 export class ControllerDefaultService implements IControllerDefaultService {
--- a/src/notebooks/controllers/controllerDefaultService.ts
+++ b/src/notebooks/controllers/controllerDefaultService.ts
@@ -17,6 +17,9 @@ import {
     IVSCodeNotebookController
 } from './types';
 
+/**
+ * Determines the 'default' kernel for a notebook
+ */
 @injectable()
 export class ControllerDefaultService implements IControllerDefaultService {
     private get isLocalLaunch(): boolean {
--- a/src/notebooks/export/exportUtil.ts
+++ b/src/notebooks/export/exportUtil.ts
@@ -3,6 +3,9 @@ import { NotebookCellData, NotebookData, NotebookDocument, Uri } from 'vscode';
 import { IExtensions } from '../../platform/common/types';
 import { ExportFormat, IExportDialog } from './types';
 
+/**
+ * Export utilities that are common to node/web
+ */
 @injectable()
 export class ExportUtilBase {
     constructor(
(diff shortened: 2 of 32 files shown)


src/notebooks/controllers/controllerDefaultService.ts
```diff
@@ -17,6 +17,9 @@ import {
     IVSCodeNotebookController
 } from './types';
 
+/**
```
rebornix, line 20: The `ControllerPreferredService` "Computes and tracks the preferred kernel for a notebook", what's the major difference between "default" and "preferred" here?

rchiodo: Preferred reads the data in the notebook. Default doesn't.

src/notebooks/export/fileConverter.ts
```diff
@@ -18,6 +18,9 @@ import { ExportFileOpener } from './exportFileOpener';
 import { ExportUtilBase } from './exportUtil';
 import { ExportFormat, IExport, IExportDialog, IFileConverter, INbConvertExport } from './types';
 
+/**
+ * Converts different file formats to others. Used in export.
```
DonJayamanne, line 22: Aren't some these documentations poinless?

rchiodo: I believe that would be subjective.

rchiodo: More constructive feedback here would be what would you rather have this say.

src/notebooks/export/exportUtil.ts
```diff
@@ -3,6 +3,9 @@ import { NotebookCellData, NotebookData, NotebookDocument, Uri } from 'vscode';
 import { IExtensions } from '../../platform/common/types';
 import { ExportFormat, IExportDialog } from './types';
 
+/**
+ * Export utilities that are common to node/web
```
DonJayamanne, line 7: E.g. this comment feels pointless to me,
no need to remove, but i don't think this would be useful
Having meaningful name and readable code makes this unnecessary (and the fact that the convention of having non .web and .node extension makes it obvious this is common for node & web)

DonJayamanne (approved):

src/notebooks/export/exportUtil.ts
rchiodo: These comments also show up when hovering over usages of a class, so the contents and the file name may not be apparent when hovering over a use.

rchiodo: Meaning they have more use than just opening this file and looking at the comment.

DonJayamanne: Here's what I'm getting at
`ExportUtilsBase` is the name of the class.

* `Export` tells us its related to export
* `Utils` tells us that its a utility class
* The fact that it doesn't have a .web.ts nor a .node.ts extension means (by convention in this repo) its common to both web and node.

Hence the argument, this comment is not necessary nor does it add any new information that can be determined from the name of the class & file (which is already meaningfully named).

Again, no need to remove & possible others find this useful.

src/notebooks/serverSelector.ts
```diff
@@ -13,6 +13,9 @@ import { JupyterServerSelector, SelectJupyterUriCommandSource } from '../kernels
 import { IJupyterServerUriStorage } from '../kernels/jupyter/types';
 import { IExtensionSyncActivationService } from '../platform/activation/types';
 
+/**
+ * Registers commands to allow the user to set the remote server URI.
+ */
```
rebornix, line 18: I guess we could have it merged into `notebookCommandListener`

src/notebooks/export/exportUtil.ts
rchiodo: Are you arguing we should skip documenting each class? Or maybe only document them if the description can offer something non obvious?

I find that the documentation comes up in hover useful (for this case it would be because the file name isn't obvious when just hovering over something)

rebornix:
:clap:

Issue #10832, opened 2022-07-15
Kernel architecture review and polish


rebornix:
This issue covers the architecture review and polish @DonJayamanne and I are working on. The main goals of this task are

* Finalize how the APIs from `src/kernels` are being used by the core system (`notebooks`, `interactive-window` etc) and 3rd party extensions. There will be restriction and clear separation between internal and external use of the kernel interfaces.
* Internal architecture review and polish of `src/kernels` component. The kernel component will stay minimal and limited responsibilities without worrying about business logics of how kernels are being used (i.e., how to read variables, how to install dependencies into the kernels, etc)

Here is a list of tasks we will tackle:

- [x] Strict IKernel with notebook document linked in core
  - https://github.com/microsoft/vscode-jupyter/pull/10817
  - Internal system always create `IKernel` with `NotebookDocument`, it's 1-to-1 mapping.
  - 3rd party can create kernels from `Uri` instead of `NotebookDocument`
- [x] Conditional notebook controller for cell execution https://github.com/microsoft/vscode-jupyter/pull/10884 
  - 3rd party can create kernels, which don't require a VS Code `NotebookController` to be created 
- [x] Remove explicit knowledge of [interactive window](https://github.com/microsoft/vscode-jupyter/blob/main/src/kernels/kernel.base.ts#L608) from `IKernel`
  - https://github.com/microsoft/vscode-jupyter/pull/10853
  - Both Interactive Window and Debugging feature require special initialization after jupyter session is started. `StartupCode` can be registered from IW and Debugging other than embedding the logic into `IKernel`
  - The fix would be similar to how we add `@multiInject(ITracebackFormatter)`
- [x] Debugging and Debug Variable can probably be moved out https://github.com/microsoft/vscode-jupyter/pull/10870
  - Debugging and debug variable provider are used by notebook and interactive window so they probably don't need to live in `kernels` at all.
- [x] 🏃 `KernelFinder` and `KernelProvider` https://github.com/microsoft/vscode-jupyter/issues/11153
- [x] Review misplaced internal components 
  - [x] ipywidgets https://github.com/microsoft/vscode-jupyter/pull/10903
  - [x] port
```[tasklist]

rebornix:
A gist of our discussion of how the kernel detection and launching works currently

<img width="1604" alt="image" src="https://user-images.githubusercontent.com/876920/179894686-5f61b1a7-bb1d-42ba-bebe-f4fa874e3c34.png">

[...]

PR #10884, opened 2022-07-21
Remove controllers from ThirdParty Kernels

DonJayamanne:
Part of https://github.com/microsoft/vscode-jupyter/issues/10832

* Split IKernel into 3rd party and regular kernels (regular = kernels used by us)
* Split IKernelProvider into 3rd party and regular provider

### DIFF
--- a/src/kernels/execution/kernelExecution.ts
+++ b/src/kernels/execution/kernelExecution.ts
@@ -52,13 +52,12 @@ export class BaseKernelExecution<TKernel extends IBaseKernel = IBaseKernel> impl
      * If we don't have a kernel (Jupyter Session) available, then just abort all of the cell executions.
      */
     public async interrupt(sessionPromise?: Promise<IKernelConnectionSession>): Promise<InterruptResult> {
-        // Possible we don't have a notebook.
         const session = sessionPromise ? await sessionPromise.catch(() => undefined) : undefined;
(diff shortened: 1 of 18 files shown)


src/kernels/execution/kernelExecution.ts
```diff
@@ -32,134 +34,58 @@ import { noop } from '../../platform/common/utils/misc';
  * Separate class that deals just with kernel execution.
  * Else the `Kernel` class gets very big.
  */
-export class KernelExecution implements IDisposable {
-    private readonly documentExecutions = new WeakMap<NotebookDocument, CellExecutionQueue>();
-    private readonly executionFactory: CellExecutionFactory;
-    private readonly disposables: IDisposable[] = [];
+export class BaseKernelExecution<TKernel extends IBaseKernel = IBaseKernel> implements IDisposable {
```
sadasant, line 37: Could the base class live in a different file? Not sure if that's preferable or not

DonJayamanne: Can, sometimes i find it useful to be in one place, else theres a lot of going back and forth..
However Will change

rchiodo (approved):

src/kernels/execution/kernelExecution.ts
DonJayamanne: Damn it, forgot this, will do in debt week

PR #10898, opened 2022-07-21
Fix conda activation failure test

### LINKED ISSUE #10896: test failing consistently in CI `Activate conda environment using conda run and activation commands`
e.g. https://github.com/microsoft/vscode-jupyter/actions/runs/2708076560

rchiodo:
Fixes #10896 

Root cause was the interpreter service is actually required now for this test to pass (even though it was effectively disabling it) but the python API wasn't caching the API promise correctly so sometimes it had the API and sometimes it didn't.

Change the test to better act like python isn't installed and fixed the caching of the api.


src/platform/api/pythonApi.ts
```diff
@@ -411,14 +410,14 @@ export class InterpreterService implements IInterpreterService {
         if (!this.extensionChecker.isPythonExtensionInstalled) {
             return;
         }
-        if (!this.api) {
```
rchiodo, line 414: It's only caching the result here. Not the actual promise. So this could be called over and over and all of the calls would update this.api.  In the test one of them would be disabled.

src/test/datascience/interpreters/environmentActivationService.vscode.test.ts
```diff
@@ -52,13 +55,23 @@ suite('DataScience - VSCode Notebook - (Conda Execution) (slow)', function () {
             api = await initialize();
             sinon.restore();
             pythonApiProvider = api.serviceContainer.get<IPythonApiProvider>(IPythonApiProvider);
-            const interpreter = await api.serviceContainer
-                .get<IInterpreterService>(IInterpreterService)
-                .getActiveInterpreter();
-            if (!interpreter) {
+            const interpreterService = api.serviceContainer.get<IInterpreterService>(IInterpreterService);
```
rchiodo, line 58: Improved the test to work locally more often. The active interpreter isn't always a conda one locally. This should ensure it's at least picking a conda one.

src/test/interpreters/interpreterService.node.ts
```diff
@@ -1,86 +0,0 @@
...
-import { getInterpreterInfo } from './index.node';
-import { IPythonExtensionChecker } from '../../platform/api/types';
-import { Resource } from '../../platform/common/types';
-import { IInterpreterService } from '../../platform/interpreter/contracts';
-import { PythonEnvironment } from '../../platform/pythonEnvironments/info';
-import { ResourceSet } from '../../platform/vscode-path/map';
-
-let interpretersCache: Promise<PythonEnvironment[]> | undefined;
-@injectable()
-export class InterpreterService implements IInterpreterService {
```
rchiodo, line 16: We never actually used this. At first I started making changes here but realized it wasn't even being used.

DonJayamanne (approved):

PR #10899, opened 2022-07-21
fix flaky debugging tests

amunger:
More diagnostic logging for CI test runs and clean up the debug session before the IW debugging test suite - a previous suite was leaving the debugging service in a weird state that still had a debugging context with no debugger running.

skipping java kernel test - https://github.com/microsoft/vscode-jupyter/issues/10900

### DIFF
--- a/src/interactive-window/editor-integration/codeLensFactory.ts
+++ b/src/interactive-window/editor-integration/codeLensFactory.ts
@@ -109,6 +109,7 @@ export class CodeLensFactory implements ICodeLensFactory {
             // Because we have all new ranges, we need to recompute ALL of our code lenses.
             cache.documentLenses = [];
             cache.gotoCellLens = [];
+            cache.cachedDocumentVersion = document.version;
             needUpdate = true;
         }
 
@@ -142,7 +143,6 @@ export class CodeLensFactory implements ICodeLensFactory {
                 });
                 firstCell = false;
             });
-            traceInfoIfCI(`Cached ${cache.documentLenses.length} code lenses`);
         } else {
             traceInfoIfCI(
                 `NOT Generating new code lenses for version ${document.version} of document ${document.uri} - needUpdate: ${needUpdate}, cellRanges.length: ${cache.cellRanges.length}`
@@ -170,7 +170,6 @@ export class CodeLensFactory implements ICodeLensFactory {
             }
         }
 
-        cache.cachedDocumentVersion = document.version;
         return cache;
     }
     private getDocumentExecutionCounts(key: string): number[] {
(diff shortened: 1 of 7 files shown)


amunger:
from the new logs, we're not getting any cell ranges back when calculating lenses - it's not finding any cell markers in the text. Perhaps it's looking at the wrong document, or it has an old version of the text?

src/interactive-window/editor-integration/codeLensFactory.ts
```diff
@@ -138,13 +138,15 @@ export class CodeLensFactory implements ICodeLensFactory {
                 commands.forEach((c) => {
                     const codeLens = this.createCodeLens(document, r, c, firstCell);
                     if (codeLens) {
-                        cache?.documentLenses.push(codeLens); // NOSONAR
+                        cache!.documentLenses.push(codeLens); // NOSONAR
```
IanMatthewHuff, line 141: This seems like such an odd typescript limitation. I poked around a bit, but didn't turn up a good quick explanation. Anyone happen to know the quick answer for needing the "!" here?

IanMatthewHuff: Code is fine BTW, just wondering about the language.

src/interactive-window/editor-integration/codelensprovider.ts
```diff
@@ -164,6 +164,13 @@ export class DataScienceCodeLensProvider implements IDataScienceCodeLensProvider
                     }
                     return false;
                 });
+            } else {
+                traceInfoIfCI(
+                    `Detected debugging context because activeDebugSession is name:"${this.debugService.activeDebugSession.name}", type: "${this.debugService.activeDebugSession.type}", ` +
+                        `but fell through with debugLocation: ${JSON.stringify(
+                            debugLocation
+                        )}, and doument.uri: ${document.uri.toString()}`
```
IanMatthewHuff, line 172: Just a logging message, but a misspell here on document

src/test/datascience/interactiveDebugging.vscode.common.ts
```diff
@@ -58,6 +58,17 @@ export function sharedIWDebuggerTests(
                 if (options?.suiteSetup) {
                     await options?.suiteSetup(debuggerType);
                 }
+
```
IanMatthewHuff, line 61: I think Don mentioned that this could be a general function. But ok to put it just here to unblock stuff.

IanMatthewHuff (approved):

src/interactive-window/editor-integration/codeLensFactory.ts
amunger: I'm guessing the forEach's scope disrupts the type narrowing done earlier. I can't see a reason why it would, so it's just probably hard to implement

Issue #10900, opened 2022-07-21
failing test: Automatically pick java kernel when opening a Java Notebook


amunger:
Seems like it's failing pretty often (if not always) in `raw, conda, 3.9, ubuntu-latest` setup

amunger:
screenshot confirms wrong kernel was selected
![image](https://user-images.githubusercontent.com/2019016/180479411-1a4a7c0e-3032-4ef6-80c2-02951a2d54a5.png)

amunger:
Still failing after the regression fix, I'm pretty sure CI machines do not have the kernel installed required for this test

DonJayamanne:
This isn't a flaky tet issue, the code is broken, we're not listing all of the kernels correctly.

DonJayamanne:
@IanMatthewHuff I tried to enable this test and found that the code was actually buggy, attempted fix is here https://github.com/microsoft/vscode-jupyter/pull/11053

Basically today we can run into situations where the kernels on disc are not shown.
The problem is with the new code (changes as a result of) written to display the new kernel picker ux.
Basically we have code that will fetch the controllers, and then subsequently we attempt to run the same method again and this sime the previous request is cancelled and that causes issues.

On a side note, the PR ensures that none of the `cancellation` tokens are nullable, if you follow the code path its impossible for them to be null, i.e. they are always valid.

DonJayamanne:
Probably blocked on https://github.com/microsoft/vscode-jupyter/issues/11153

DonJayamanne:
Adding my self, as I believe the problem is in the kernel finder (my area) and not necessarily in the code that matches kernels.

DonJayamanne:
# Note: This was an actual bug that users were running into and cause users not to see their kernels (including virtual envioronments)

PR #10916, opened 2022-07-25
Select preferred non-python kernel

DonJayamanne:
We weren't selecting the right kernel for non-python kernels


rchiodo (approved):

amunger:
linking related issue https://github.com/microsoft/vscode-jupyter/issues/10900, though the test is still failing

PR #10966, opened 2022-07-28
Handle restart failures with reserved file names

### LINKED ISSUE #10937: Different error handling of kernel startup and restart
Testing #10923

1. Create a new folder and a `ipynb` file with content `print(123)` in the first cell
2. Run the cell against a valid kernel (Python 3.6+)
3. Add an `email.py` file in the same folder
4. Restart the kernel

Now I got error message

```
Failed to start the Kernel. 
Cannot read properties of null (reading 'statusChanged'). 
View Jupyter [log](command:jupyter.viewOutput) for further details.
```

And in the Jupyter log it said

```
[E 10:34:35.419 NotebookApp] Error opening stream: HTTP 404: Not Found (Kernel does not exist: 4f479182-870f-4f57-a0b1-009e5853555e)
```

If we flip step 2 and 3, then we would get proper error message

```
Some of the following files found in the working directory may have prevented the Kernel from starting. Consider renaming them.
File(s): [~\email.py](file:///c%3A/Users/rebor/code/personal/notebooks/bugs/email.py?line=1) might need to be renamed.
```

DonJayamanne:
Fixes #10937

### DIFF
--- a/src/kernels/common/baseJupyterSession.ts
+++ b/src/kernels/common/baseJupyterSession.ts
@@ -6,14 +6,22 @@ import type { JSONObject } from '@lumino/coreutils';
 import type { Slot } from '@lumino/signaling';
 import { Observable } from 'rxjs/Observable';
 import { ReplaySubject } from 'rxjs/ReplaySubject';
-import { CancellationError, CancellationToken, CancellationTokenSource, Event, EventEmitter, Uri } from 'vscode';
+import {
+    CancellationError,
+    CancellationToken,
+    CancellationTokenSource,
+    Disposable,
+    Event,
+    EventEmitter,
+    Uri
+} from 'vscode';
 import { WrappedError } from '../../platform/errors/types';
 import { disposeAllDisposables } from '../../platform/common/helpers';
 import { traceInfo, traceVerbose, traceError, traceWarning, traceInfoIfCI } from '../../platform/logging';
 import { IDisposable, Resource } from '../../platform/common/types';
 import { createDeferred, sleep, waitForPromise } from '../../platform/common/utils/async';
 import * as localize from '../../platform/common/utils/localize';
-import { noop } from '../../platform/common/utils/misc';
+import { noop, swallowExceptions } from '../../platform/common/utils/misc';
 import { sendTelemetryEvent, Telemetry } from '../../telemetry';
 import { JupyterInvalidKernelError } from '../errors/jupyterInvalidKernelError';
 import { JupyterWaitForIdleError } from '../errors/jupyterWaitForIdleError';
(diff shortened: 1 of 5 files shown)


src/kernels/common/baseJupyterSession.ts
```diff
@@ -335,14 +343,33 @@ export abstract class BaseJupyterSession implements IBaseKernelConnectionSession
...
+                    new Disposable(() =>
+                        swallowExceptions(() => session.disposed.disconnect(sessionDisposed.resolve, sessionDisposed))
+                    )
+                );
+                const sleepPromise = sleep(timeout);
+                sessionDisposed.promise.ignoreErrors();
+                sleepPromise.ignoreErrors();
+                deferred.promise.ignoreErrors();
+                const result = await Promise.race([deferred.promise, sleepPromise, sessionDisposed.promise]);
+                if (session.isDisposed) {
```
DonJayamanne, line 365: This is what was missing, at this point its possible the session has been disposed.
We were getting errors because kernel was disposed and `session.kernel` was undefined.

We would be faster in handling cases where restarts fail before the timeout. 
Before this change we'd wait for 60s, now if the session dies before that we'd exit this block much sooner than 60s.

sadasant (approved):

rchiodo (approved):

src/kernels/common/baseJupyterSession.ts
```diff
@@ -335,14 +343,33 @@ export abstract class BaseJupyterSession implements IBaseKernelConnectionSession
...
+                session.disposed.connect(sessionDisposed.resolve, sessionDisposed);
+                disposables.push(
+                    new Disposable(() =>
+                        swallowExceptions(() => session.disposed.disconnect(sessionDisposed.resolve, sessionDisposed))
+                    )
+                );
+                const sleepPromise = sleep(timeout);
+                sessionDisposed.promise.ignoreErrors();
+                sleepPromise.ignoreErrors();
+                deferred.promise.ignoreErrors();
```
amunger, line 363: this seems a bit overly cautious - can these actually throw?

src/kernels/common/baseJupyterSession.ts
```diff
@@ -335,14 +343,33 @@ export abstract class BaseJupyterSession implements IBaseKernelConnectionSession
...
+                const result = await Promise.race([deferred.promise, sleepPromise, sessionDisposed.promise]);
```
amunger, line 364: `deferred` could be named better - `kernelConnected`?

amunger (approved):

src/kernels/common/baseJupyterSession.ts
```diff
@@ -335,14 +343,33 @@ export abstract class BaseJupyterSession implements IBaseKernelConnectionSession
...
+                if (session.isDisposed) {
+                    traceError('Session disposed while waiting for session to be idel.');
```
amunger, line 366: typo: idle

src/kernels/common/baseJupyterSession.ts
DonJayamanne: Changed to `kernelStatus` as its a promise with the value of status

PR #10967, opened 2022-07-28
Show variable snapshot in data viewer

### LINKED ISSUE #10942: Data Viewer: open the same variable from variable doesn't reuse Data Viewer
Testing #10915

* Click on the action "Show Variable in Data Viewer" from the variable view
* Data Viewer opened
* Click again
* It opens a new data viewer, instead of focusing on the Data Viewer again

sadasant:
How does this look?

Fixes #10942

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -660,7 +660,7 @@
     "DataScience.variableExplorerTypeColumn": "Type",
     "DataScience.variableExplorerCountColumn": "Size",
     "DataScience.variableExplorerValueColumn": "Value",
-    "DataScience.showDataExplorerTooltip": "Show variable in data viewer",
+    "DataScience.showDataExplorerTooltip": "Show variable current value in data viewer",
     "DataScience.dataExplorerInvalidVariableFormat": "'{0}' is not an active variable.",
     "DataScience.jupyterGetVariablesExecutionError": "Failure during variable extraction:\r\n{0}",
     "DataScience.loadingMessage": "loading ...",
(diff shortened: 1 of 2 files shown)


rchiodo (approved):

rchiodo (changes requested):
Doesn't this need changes to the localize.ts file too? For the translated version?

sadasant:
@rchiodo I don't think so. Thanks to your comment I realized that this other file needed to change: https://github.com/microsoft/vscode-jupyter/pull/10967/commits/1bafecb0686fcb0f8d916dbb01f38b5b3ab0ed34

That file isn't using our `localize` function, it's using `getLocString` from `../react-common/locReactSide`, which reads from the `package.nls.json` when the localization is initialized.

rchiodo (approved):

amunger:
I don't think this clarifies the action - it's not like we're going to show an old value of the variable when you click that button.
If the issue is that the variable view is static and won't keep in sync with the variable in the kernel, then I don't think this message really conveys that.
Maybe calling it a snapshot, or static view would help, but "current" doesn't imply to me that it will become stale.

sadasant:
@amunger I like how you think! How would you phrase it? `Show variable snapshot in data viewer?`

amunger:
> How would you phrase it? Show variable snapshot in data viewer?

yeah I would like that more

sadasant:
@amunger pushed!

PR #10970, opened 2022-07-28
First try at documenting an error

rchiodo:
@amunger thoughts?

### DIFF
--- a/src/platform/errors/interactiveCellResultError.ts
+++ b/src/platform/errors/interactiveCellResultError.ts
@@ -3,7 +3,16 @@
 
 import { NotebookCell } from 'vscode';
 
-// Small helper error to use in our class
+/**
+ * Error thrown when running cells in the interactive window.
+ *
+ * Cause:
+ * Kernel.executeCell() returns an error when running a cell in the interactive window.
+ *
+ * Handled by:
+ * Running cell in IW shows the error. If no running cell, then a notification is shown to the user.
+ *
+ */
 export class InteractiveCellResultError extends Error {
     constructor(public readonly cell: NotebookCell) {
         super('Cell failed to execute');


src/platform/errors/interactiveCellResultError.ts
```diff
@@ -3,7 +3,16 @@
 
 import { NotebookCell } from 'vscode';
 
-// Small helper error to use in our class
+/**
+ * Error thrown when running cells in the interactive window.
+ *
+ * Cause:
+ * Kernel.executeCell() returns an error when running a cell in the interactive window.
```
rchiodo, line 10: I could go into more detail here.

src/platform/errors/interactiveCellResultError.ts
```diff
@@ -3,7 +3,16 @@
...
+ *
+ * Handled by:
+ * Running cell in IW shows the error. If no running cell, then a notification is shown to the user.
+ *
+ */
```
rchiodo, line 15: Was thinking if we wanted a handling category, I could refactor all of the errors to implement some interface with an enum that indicates the handling category (similar to the category already on the BaseError)

rchiodo: It would be like:
```
enum HandlingType {
   'CellOutput'
   'Dialog'
   'Notification'
   'Swallowed'
}
```

amunger: I wouldn't want to constrict an error type to be handled in a specific way - perhaps they all have a single way of being handled now, but that might not hold true if a new context is introduced. 
I think the handled by clause here is fine, if more nuance is introduced, the comment could simply be updated

src/platform/errors/interactiveCellResultError.ts
amunger: An example would probably be the most useful. The details can change, so just searching for usages is probably the best method for digging in more

rchiodo:
@amunger so you're good with this? I wanted to submit this first and then start following this pattern everywhere else.

amunger (approved):

PR #10986, opened 2022-07-31
Stop creating hidden terms to look for Jupyter app

### LINKED ISSUE #10934: Respect VSCode's telemetry setting
## Environment data

-   VS Code version: 1.69.2
-   Jupyter Extension version (available under the Extensions sidebar): v2022.6.1201981810
-   OS (Windows | Mac | Linux distro) and version: Ubuntu 20.04 LTS

## Expected behaviour

When VSCode's telemetry preference is set to a restricted value (e.g: `off`, `crash`) the extension should not query for or send telemetry.

## Actual behaviour

I observed lines such as `printenv > /tmp/tmp-45935fdNhOTkd5hRYtxt` in my shell history and via https://twitter.com/ferrouswheel/status/1551353490949017600 traced this to having (hopefully!) come from this extension https://github.com/microsoft/vscode-jupyter/blob/545c0a7b00b073da4d73ed58bf6c49e0d31b7ee6/src/platform/terminals/helper.node.ts#L81-L93

I have `telemetry.telemetryLevel` set to `off` on the machine where this was observed.

Ideally the extension should not gather telemetry data outside the enabled telemetry remit.

DonJayamanne:
Fixes #10934


rchiodo (approved):

src/kernels/jupyter/jupyterDetectionTelemetry.node.ts
```diff
@@ -40,33 +38,6 @@ export class JupyterDetectionTelemetry implements IExtensionSyncActivationServic
         this.globalMemento.update(JupyterDetectionTelemetrySentMementoKey, true).then(noop, noop);
         this.detectJupyter('notebook', process.env).catch(noop);
         this.detectJupyter('lab', process.env).catch(noop);
```
amunger, line 40: aren't we still attempting to send telemetry here? I'm fine with the change, but it doesn't seem like it will resolve the concern raised in the attached issue.

amunger: oh, the attached issue is closed anyway. So doesn't really matter

amunger (approved):

src/kernels/jupyter/jupyterDetectionTelemetry.node.ts
DonJayamanne: The original issue is the fact that we're polluting terminal history.
this telemetry sent here is like any other telemetry.

Issue #11013, opened 2022-08-02
Problems with npm@8.3.0 and overrides


DonJayamanne:
**Problem: Currently we're unable to pin certain packages to specific versions due to a bug in the version of npm used**

We rely on `overrides` section in `package.json` to pin specific packages to particular versions

* Known issue with some versions of npm https://github.com/npm/cli/issues/4232
* Seems to be working in `8.15.0 (latest)` for me, even though others complain about `8.14.0`

**Solutions**
* Enforce node and npm/yarn versions  in package.json
    * i think it's over kill, but this prevents generation of lock files that are not reproducible (we have such lock files today)
* Option 1: Use `yarn`
* Option 2: Update `npm` to the latest version
    * Clearly document the version of node and npm required
    If a package is manually and entry added into `package.json`, then we should document this in package.json
   E.g. add a section, this won't be necessary if we use yarn or if we use the latest version of npm
    ```jsonc
    "depedencies-manually-added": {
        "added-to-fix-dependabot-issues": ["<package>:<version>"]
    }
    ```

PR #11025, opened 2022-08-02
Update packages and npm version

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/11013

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@
 ### Prerequisites
 
 1. [Node.js](https://nodejs.org/) 16.14.2
-2. [npm](https://www.npmjs.com/) 8.3.0
+2. [npm](https://www.npmjs.com/) 8.15.1
 3. [Python](https://www.python.org/) 3.6 or later
 4. Windows, macOS, or Linux
 5. [Visual Studio Code](https://code.visualstudio.com/)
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -2493,8 +2493,15 @@
         "fsevents": "^2.3.2"
     },
     "overrides": {
-        "simple-get": "3.1.1",
-        "fs-path": "0.0.25",
+        "simple-get@<3.1.1": "3.1.1",
+        "validator@<13.7.0": "13.7.0",
+        "jpeg-js@<0.4.4": "0.4.4",
```
DonJayamanne, line 2498: using `@<` so that any package that depends on a later version will still get the later version instead of what we specific here.
I.e. we need to ensure we pin only to a specific version unless absolutely necessary.

CONTRIBUTING.md
```diff
@@ -14,7 +14,7 @@
 ### Prerequisites
 
 1. [Node.js](https://nodejs.org/) 16.14.2
-2. [npm](https://www.npmjs.com/) 8.3.0
+2. [npm](https://www.npmjs.com/) 8.15.1
```
DonJayamanne, line 17: Older versions are buggy

rchiodo (approved):

amunger (approved):

package.json
```diff
@@ -2493,8 +2493,15 @@
         "fsevents": "^2.3.2"
     },
     "overrides": {
-        "simple-get": "3.1.1",
-        "fs-path": "0.0.25",
```
DonJayamanne, line 2497: `fs-path` is not used anywhere, hence this isn't required.

PR #11030, opened 2022-08-03
Verify package-lock.json file

DonJayamanne:


gulpfile.js
```diff
@@ -427,6 +427,16 @@ gulp.task('validateTelemetryMD', async () => {
     }
 });
 
+gulp.task('validatePackageLockJson', async () => {
+    const fileName = path.join(__dirname, 'package-lock.json');
+    const oldContents = fs.readFileSync(fileName).toString();
+    spawnSync('npm', ['install']);
+    const newContents = fs.readFileSync(fileName).toString();
+    if (oldContents.trim() !== newContents.trim()) {
+        throw new Error('package-lock.json has changed after running `npm install`');
```
amunger, line 436: could add suggested fix here: "make sure to run 'npm install' locally with the correct version of npm" - could we even get the version info from the environment?

amunger (approved):

gulpfile.js
rchiodo: Outputting the diff would be nice too? Otherwise how is the person going to know what to fix

DonJayamanne: i think it's obvious that something is setting with package-lock. json.
the suggested fix isn't just run npm install. it cities need, install right version of npm, install right version if node or others.

PR #11042, opened 2022-08-03
test logging

amunger:
debug tests are failing in 4 seconds, whereas they should be waiting 60. I can't figure out why from looking at the code, so I want to confirm that the wait period is too short


amunger:
maybe wrong, this example waited 60 seconds

2022-08-03T01:47:54.9076371Z info 1:47:54.883:  Start Test (completed) Stops at end of cell
2022-08-03T01:47:58.7124645Z info 1:47:58.696:  willSendRequest: disconnect
2022-08-03T01:48:58.4372515Z Test failing --- DebugSession should end

rchiodo (approved):

rchiodo (approved):

PR #11058, opened 2022-08-04
Initialize env variable to resolve a number of debugger issues

### LINKED ISSUE #8146: Jupyter cell debugging does not support "step into" the  third party library code with "justmycode:false"
Jupyter cell debugging does not support "step into" the  third party library code with "justmycode:false" in launch.json.

Please fix it !

### LINKED ISSUE #10106: Problems with Notebook debugger (using Jupyter Protocol)
**Note: These issues have been fixed for IW debugger, we need to fix them for Notebook**

* Problem 1:
	* Cell 1
	  ```python
	  < empty line>
	  < empty line>	  
	  < empty line>
	  < empty line>
	  print(1)
	  print(2)
	  ```
	When debugging the above code, the debugger will stop at line `print(1)`, however when you hit `step over` in RBL or `continue` in the cell debugger it will break there once again.

* Problem 2:
	* Cell 1
	  ```python
	  < empty line>
	  < empty line>	  
	  < empty line>
	  def foo():
	      print('inside foo')
	  ```
	  * Cell 2
	  ```python
	  
	  
	  foo()
	  ```
	  When debugging the above code, the debugger will stop at line `foo()`, however when you step into the code, it will stop way above the `def foo`, in a white space (basically on line 2 of the file, in the white space).
	  

* Problem 3:
	The above issues were uncovered while running some of the IW debugger tests.
	I think we need a few more tests for the notebook debugger to cover the same scenarios covered in IW debugger.

These issues have been fixed in the new IW when using debugger protocol.
Basically injecting `breakpiont()` along with a few other changes.

### LINKED ISSUE #10600: Debugger stops at the breakpoint twice
```python
print(1)
print(2)
print(3)
```

* Add a breakpoint to the second line
* Start debugging in IW or notebook
* Debugger will stop at the desired breakpoint
* Hit continue (F5), and debugger will stop at the same place (same breakpoint) again.

### LINKED ISSUE #11033: Set `"PYDEVD_IPYTHON_COMPATIBLE_DEBUGGING" = "1"` in the environment prior to importing `debugpy`
There was a recent fix in `debugpy` to work properly with IPython and the fix requires that `"PYDEVD_IPYTHON_COMPATIBLE_DEBUGGING" = "1"` is set in the environment prior to importing `debugpy`.

See: https://github.com/microsoft/debugpy/issues/869#issuecomment-1132141003 for details on why a custom support was needed.

DonJayamanne:
Fixes #11033
Fixes #10600
Fixes #10106
Fixes #8146

This also fixes the debugger test failures that we're running into, see here https://github.com/microsoft/debugpy/issues/1000.

Thought this could be merged, however we need to wait for a new version of debugpy to be published.
Until then this PR will be blocked.
However submitting this PR as i've already done the work (found while testing that theres no way to test as the changes haven't been released yet)


rchiodo (approved):

src/kernels/kernel.ts
```diff
@@ -536,16 +534,18 @@ abstract class BaseKernel<TKernelExecution extends BaseKernelExecution> implemen
...
-                result.push(...dir);
-            }
+        const startupCode = await Promise.all(
+            this.startupCodeProviders.sort((a, b) => b.priority - a.priority).map((provider) => provider.getCode(this))
+        );
+        for (let code of startupCode) {
+            result.push(...code);
+        }
 
+        // If this is a live kernel, we shouldn't be changing anything by running startup code.
```
amunger, line 544: we've already gathered startup code from the injected providers, so this comment isn't accurate.

src/kernels/kernelStartupCodeProvider.node.ts
```diff
@@ -25,6 +25,13 @@ export class KernelStartupCodeProvider implements IStartupCodeProvider {
     ) {}
 
     async getCode(kernel: IKernel): Promise<string[]> {
+        // If this is a live kernel, we shouldn't be changing anything by running startup code.
+        if (
+            !isPythonKernelConnection(kernel.kernelConnectionMetadata) &&
```
amunger, line 30: Might be easy to forget to add this check when creating new providers. We could alternatively change the interface to `getCode` and `getCodeForLiveKernel` to be more explicit

DonJayamanne: I don't think that's a good idea, the implementation should worry about the type of kernel we provide code for, after all the kernel is passed as an argument. 
Basically the implemetnation shouold worry about it.

i.e. if we have seprate interrfaces (seprate method) for each different implemtation then we end up with a single interface for every single enumeration which then makes it unnecessary to have an interface or even a base class.

src/kernels/kernel.ts
DonJayamanne: Nope it is still correct, if you check the code in the if condition, we're actually injecting some code, and that should be done only for non-live kernels. ie. the comment is required.

PR #11093, opened 2022-08-10
Add eslint rule for license and normalize headers

roblourens:
This adds the header eslint plugin, adds an eslint rule for a license header, and runs `npm lint -- --fix` on all files in the repo. 

Then, I did a search/replace to normalize the number of newlines after the license header. This caused the formatter to be run as well, so there are some minor formatting changes, which seem fine. If you don't want those changes in this PR, I can revert that second commit.


package-lock.json
```diff
@@ -34035,7 +34052,7 @@
             "requires": {
                 "extend": "^3.0.0",
                 "glob": "^7.1.1",
-                "glob-parent": "5.1.2",
+                "glob-parent": "^3.1.0",
```
DonJayamanne, line 34055: Not sure this is a valid change.
Please could you double check your version or node and npm
This looks like a downgrade

DonJayamanne: You need
1. [Node.js](https://nodejs.org/) 16.14.2
2. [npm](https://www.npmjs.com/) 8.15.1

IanMatthewHuff:
Looks like it triggered a ton of files to be pulled in with the prettier requirement (old files were grandfathered in for this until they were touched). You can see all the failure in the lint step. I think that I'd be ok with an admin push to get around the lint check for the prettier changes?

package-lock.json
roblourens: Seems like these are out of sync, I have that node version and npm 8.5.0 which is what it appears to have been published with, https://nodejs.org/en/download/releases/ is that right?

DonJayamanne: you need the latest version of npm, as npm 8.5.0 has a bug when using the overrides feature.

DonJayamanne: & yes, they are out of sync, either we go with thie route or use yarn, i mentioend this in standup & there was no preference to use yarn, hence i didn't want to make a bigger change by changing the lock file from npm to yarn

This was to fix the CG issues.

roblourens: Oh, I understand, ok thanks

PR #11094, opened 2022-08-10
Add eslint rule for license and normalize headers

roblourens:
This adds the header eslint plugin, adds an eslint rule for a license header, and runs `npm lint -- --fix` on all files in the repo. 

Then, I did a search/replace to normalize the number of newlines after the license header. This caused the formatter to be run as well, so there are some minor formatting changes, which seem fine. If you don't want those changes in this PR, I can revert that. I also ran prettier on all files, since the search/replace ran prettier on a bunch of them.


IanMatthewHuff:
Looks like Don's two new files snuck in with a bad header on the last merge with main. But aside from that looks good.

amunger (approved):

DonJayamanne (approved):

PR #11096, opened 2022-08-10
Use a single process to interrupt Python Kernels

### LINKED ISSUE #9909: Do not start multiple kernels when running a single cell in a notebook
Curious why the descrepancy.
It would be a lot cleaner and simpler to just have jupyter to restart local (non-raw) kernels as well instead of us shutting it down and restarting a kernel.

Behind the scenes jupyter does the same thing, they kill the kernel process and start a whole new process.

I feel this will greatly simplify our code, by removing a lot of the code related to restarting

### User bug
See issue #9700
Note: Marked as bug as we end up starting multiple R kernels that chew up resources on end user machines, i.e. starting multiple processes is not cheap.

DonJayamanne:
Fixes #9909

### DIFF
--- a/src/kernels/raw/finder/pythonKernelInterruptDaemon.node.ts
+++ b/src/kernels/raw/finder/pythonKernelInterruptDaemon.node.ts
@@ -3,51 +3,163 @@
 
 'use strict';
 
-import { ChildProcess } from 'child_process';
-import { MessageConnection, RequestType0 } from 'vscode-jsonrpc';
-import { traceInfo } from '../../../platform/logging';
-import { IPlatformService } from '../../../platform/common/platform/types';
-import { BasePythonDaemon } from '../../../platform/common/process/baseDaemon.node';
-import { IPythonExecutionService } from '../../../platform/common/process/types.node';
-import { PythonEnvironment } from '../../../platform/pythonEnvironments/info';
-import { IPythonKernelDaemon } from '../types';
+import { traceError, traceVerbose, traceWarning } from '../../../platform/logging';
+import { IPythonExecutionFactory, ObservableExecutionResult } from '../../../platform/common/process/types.node';
+import { EnvironmentType, PythonEnvironment } from '../../../platform/pythonEnvironments/info';
+import { inject, injectable } from 'inversify';
+import { IInterpreterService } from '../../../platform/interpreter/contracts';
+import { IAsyncDisposable, IDisposableRegistry, Resource } from '../../../platform/common/types';
+import { createDeferred, Deferred } from '../../../platform/common/utils/async';
+import { Disposable } from 'vscode';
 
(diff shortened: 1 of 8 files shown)


src/kernels/raw/finder/pythonKernelInterruptDaemon.node.ts
```diff
@@ -3,51 +3,171 @@
...
+        return true;
+    }
+    return false;
+}
+function isSupportedPythonVersion(interpreter: PythonEnvironment) {
+    if (
+        (interpreter?.version?.major ?? 3) >= 3 &&
+        // Even thought 3.6 is no longer supported, we know this works well enough for what we want.
+        // This way we don't need to update this every time the supported version changes.
+        (interpreter?.version?.minor ?? 6) >= 3
```
IanMatthewHuff, line 32: This supports 3.3, 3.4, 3.5? Copy paste issue?

src/kernels/raw/finder/pythonKernelInterruptDaemon.node.ts
```diff
@@ -3,51 +3,171 @@
...
+import { inject, injectable } from 'inversify';
+import { IInterpreterService } from '../../../platform/interpreter/contracts';
+import { IAsyncDisposable, IDisposableRegistry, IExtensionContext, Resource } from '../../../platform/common/types';
+import { createDeferred, Deferred } from '../../../platform/common/utils/async';
+import { Disposable, Uri } from 'vscode';
+import { EOL } from 'os';
+function isBestPythonInterpreterForAnInterruptDaemon(interpreter: PythonEnvironment) {
+    if (
+        isSupportedPythonVersion(interpreter) &&
+        (interpreter?.envType === EnvironmentType.Global ||
```
IanMatthewHuff, line 18: Maybe a comment? I'm not sure why the envType is restricted.

IanMatthewHuff (approved):
Seem legit, though I'm probably not the best reviewer for the Python code. It does look like there is a possible copy paste error in the version check for python interpreter.

src/kernels/raw/finder/pythonKernelInterruptDaemon.node.ts
DonJayamanne: yes thats correct

src/kernels/raw/finder/pythonKernelInterruptDaemon.node.ts
DonJayamanne: Done

src/kernels/raw/finder/pythonKernelInterruptDaemon.node.ts
DonJayamanne: Fixed

PR #11098, opened 2022-08-10
Put formatter setting in language override, otherwise it can be overridden by a language-scoped user setting.

roblourens:
Also add git.branchProtection


roblourens:
Also removed the `pwa-` prefixes in launch.json to fix warnings, there will be no behavior change

rebornix (approved):

PR #11104, opened 2022-08-11
Point release 20227

DonJayamanne:


package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "jupyter",
     "displayName": "Jupyter",
-    "version": "2022.7.100",
+    "version": "2022.7.120",
```
IanMatthewHuff, line 4: Not sure that it matters (so no need to change if it doesn't) but I would have thought that 7.110 would have been next, not 7.120.

DonJayamanne: oh shoot, yes you are right, no idea why i went with 120

IanMatthewHuff (approved):
Approved with a possible change of the version number if needed.

amunger (approved):

PR #11109, opened 2022-08-12
Prompt to select env when there is no Jupyter env

### LINKED ISSUE #1110: Jupyter interpreter can be invalid and break things
See the discussion here:
https://github.com/microsoft/vscode-jupyter/issues/513

Sometimes the jupyter interpreter can go out of date. We should verify it's still valid somehow before using it.

DonJayamanne:
Fixes #1110

### DIFF
--- a/src/kernels/jupyter/interpreter/jupyterInterpreterService.node.ts
+++ b/src/kernels/jupyter/interpreter/jupyterInterpreterService.node.ts
@@ -18,6 +18,8 @@ import { JupyterInterpreterOldCacheStateStore } from './jupyterInterpreterOldCac
 import { JupyterInterpreterSelector } from './jupyterInterpreterSelector.node';
 import { JupyterInterpreterStateStore } from './jupyterInterpreterStateStore.node';
 import { JupyterInterpreterDependencyResponse } from '../types';
+import { IApplicationShell } from '../../../platform/common/application/types';
+import { DataScience } from '../../../platform/common/utils/localize';
 
 /**
  * Manages picking an interpreter that can run jupyter.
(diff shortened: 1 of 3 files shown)


src/kernels/jupyter/interpreter/jupyterInterpreterService.node.ts
```diff
@@ -112,6 +115,17 @@ export class JupyterInterpreterService {
             // Use current interpreter.
             interpreter = await this.interpreterService.getActiveInterpreter(undefined);
             if (!interpreter) {
+                if (err) {
```
IanMatthewHuff, line 118: I read this once or twice and honestly it still looks a bit off so I'll just ask. I'm not sure that I understand the flow here. Per my reading of the bug this issue is for when there is a selected Jupyter interpreter, but for when something happens to that Jupyter interpreter which makes it fail to start. So in that case getSelectedInterpreter would still have the selected jupyter interpreter and it wouldn't go down this path at all.

IanMatthewHuff: This seem restricted to a very odd / unlikely case where we have an error, but we also don't have a selected Jupyter extension or an active python interpreter.

DonJayamanne: here's the work flow,
* user selects an en
* deletes the env as per issue
* now they get a quick pick without any error message


this charge merely displays the error message

DonJayamanne: ie workflow is better now, rather than displaying a quick pick for no reason, which would be very weird for the user. now the quick pick is after a user action (button click) from the prompt

IanMatthewHuff: I was under the impression that the getActiveInterpreter right above this almost always returns with something unless there is no python detected at all on the system. Is that not the case?

DonJayamanne: nope, because the active interpreter happened to be the same interpreter that was used for jupyter

IanMatthewHuff (approved):

PR #11124, opened 2022-08-15
Kernel Finder contribution

rebornix:
This is the first part of our Kernel Finder/Launcher refactoring. In the long run we want to move towards a contribution/push model instead of current pull model. 


src/kernels/jupyter/types.ts
```diff
@@ -333,3 +336,12 @@ export interface IServerConnectionType {
     isLocalLaunch: boolean;
     onDidChange: Event<void>;
 }
+
+export const IRemoteKernelFinder = Symbol('IRemoteKernelFinder');
+export interface IRemoteKernelFinder extends IContributedKernelFinder {
```
DonJayamanne, line 341: I would suggest creating two seprate interfaces, one to list kernel specs and one to list the live kernel connectiosn (active sessions).
The kernel sessions is what can change over time, we don't need to re-query the kernel specs everytime we need new kernel sessions.

src/kernels/internalTypes.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+'use strict';
+
+import { CancellationToken } from 'vscode';
+import { Resource } from '../platform/common/types';
+import { KernelConnectionMetadata } from './types';
+
+export interface IContributedKernelFinder {
```
DonJayamanne, line 10: Can we call these `IKernelFinder` or the like, feels weierd to call them `Contributed` after all everything will be contributed in the extension, as we're creating implementations of interafces, also everything is in this extension.
If its something that can be contributed by another extension, then i think thats fine

DonJayamanne: Not, necessary to change, but nit pick

src/kernels/kernelFinder.ts
```diff
@@ -0,0 +1,126 @@
...
+        resource: Resource,
+        cancelToken: CancellationToken | undefined,
+        useCache: 'ignoreCache' | 'useCache' = 'ignoreCache'
+    ): Promise<KernelConnectionMetadata[]> {
+        this.startTimeForFetching = this.startTimeForFetching ?? new StopWatch();
+
+        const kernels: KernelConnectionMetadata[] = [];
+
+        for (const finder of this._finders) {
+            const findKernels = await finder.listContributedKernels(resource, cancelToken, useCache).catch((ex) => {
```
DonJayamanne, line 83: You can parallelize these, instead of a serial operation, ie. instead of awaiting on each item one by one.
E.g. `Promise.all(this._finders.map(finder => ...))`  or similar

src/kernels/internalTypes.ts
rebornix: `IKernelFinder` is public to external components while `ContributedKernelFinder` are the raw/remote kernel finder impls and they are internal details (not exposed to external modules). Called it `contributed*` as they are contributions/plugins to the main `IKernelFinder`. I'm imagining it can be contributed by other extensions (say we want to have the Python envs from Python extension), with that said, I'm fine with calling it something like `IKernelFinderImpl`.

src/kernels/jupyter/types.ts
rebornix: @DonJayamanne yes we can do that. This PR is the first part of moving towards contribution model but not necessarily introducing new interfaces, that can come in follow up PRs.

src/kernels/kernelFinder.ts
rebornix: Good catch.

src/kernels/jupyter/types.ts
rebornix: Let's book sometime to discuss the separation though, it would be more productive (for me ;) ) if I have you look over my shoulder.

src/kernels/internalTypes.ts
DonJayamanne: no no, lets leave this.
Its just that when I think of `contribution`, to me thats basically `external` extensibility, or `external` contribution.
But i'm fine with this.

src/kernels/common/commonFinder.ts
```diff
@@ -0,0 +1,29 @@
...
+            'JUPYTER_LOCAL_KERNELSPECS',
+            'JUPYTER_LOCAL_KERNELSPECS_V1',
+            'JUPYTER_LOCAL_KERNELSPECS_V2',
+            'JUPYTER_LOCAL_KERNELSPECS_V3',
+            'JUPYTER_REMOTE_KERNELSPECS',
+            'JUPYTER_REMOTE_KERNELSPECS_V1',
+            'JUPYTER_REMOTE_KERNELSPECS_V2',
+            'JUPYTER_REMOTE_KERNELSPECS_V3'
+        ]
+            .filter((key) => LocalKernelSpecsCacheKey !== key && RemoteKernelSpecsCacheKey !== key) // Exclude latest cache key
```
IanMatthewHuff, line 25: Do you need this? The most recent V4 version strings are not in the list currently, so I don't think this filter is doing anything unless it's for future safety.

src/kernels/kernelFinder.ts
```diff
@@ -0,0 +1,136 @@
...
+
+        const kernels: KernelConnectionMetadata[] = [];
+
+        const allKernels = await Promise.all(
+            this._finders.map((finder) => {
+                return finder
+                    .listContributedKernels(resource, cancelToken, useCache)
+                    .catch((ex) => {
+                        // Sometimes we can get errors from the socket level or jupyter, with the message 'Canceled', lets ignore those
+                        if (!isCancellationError(ex, true)) {
```
IanMatthewHuff, line 88: This same catch chunk is also in the catch for listContributedKernels in both finders. Seems like it should only be here or in the finders, not both locations.

src/kernels/raw/finder/localKernelFinder.node.ts
```diff
@@ -3,26 +3,182 @@
...
+        const kernelsWithoutCachePromise: Promise<LocalKernelConnectionMetadata[]> = this.listKernels(
+            resource,
+            cancelToken
+        );
+        let kernels: LocalKernelConnectionMetadata[] = [];
+        if (useCache === 'ignoreCache') {
+            try {
+                kernels = await kernelsWithoutCachePromise;
+            } catch (ex) {
+                traceWarning(`Could not fetch kernels from the ${this.kind} server, falling back to cache: ${ex}`);
```
IanMatthewHuff, line 93: Warning message seems off here. This is the ignore cache case, so it's not falling back on the cache and it's not a server in the local case.

src/kernels/raw/finder/localKernelFinder.node.ts
```diff
@@ -3,26 +3,182 @@
...
+                updateCache = false;
+            }
+        } else {
+            let kernelsFromCache: LocalKernelConnectionMetadata[] | undefined;
+            kernelsFromCachePromise
+                .then((items) => {
+                    kernelsFromCache = items;
+                    updateCache = false;
+                })
+                .catch(noop);
```
IanMatthewHuff, line 104: Looks like this would eat errors from getFromCache, cache would be less important, but I'd like to see something in the logging at least.

src/kernels/raw/finder/localKernelFinder.node.ts
```diff
@@ -54,4 +210,38 @@ export class LocalKernelFinder implements ILocalKernelFinder {
             return true;
         });
     }
+
+    private async writeToCache(values: LocalKernelConnectionMetadata[]) {
+        this.cache = values;
+        const serialized = values.map(serializeKernelConnection);
+        await Promise.all([
+            removeOldCachedItems(this.globalState),
+            ,
```
IanMatthewHuff, line 219: Some sort of typo here it looks like.

src/kernels/common/commonFinder.ts
rebornix: We used to have this so I kept it still, not sure if we can remove it now.

IanMatthewHuff:
Change seems pretty legit. It did look to me like in the two finders the following functions were very similar to each other:
writeToCache
getFromCache
(kinda) listKernelsImpl
listContributedKernels

While the main differences were in:
listKernels vs listKernelsFromConnection
isValidCachedKernel

Would be nice with a refactor (especially how twisty the caching stuff is) if more of the common things could be in the base. E

src/kernels/common/commonFinder.ts
IanMatthewHuff: Yeah, that's the tricky bit with refactors, hard to see in the PR what is new versus a move. I doesn't seem harmful so I guess remove / keep at your discretion.

src/kernels/raw/finder/localKernelFinder.node.ts
rebornix: 👍 this is only valid in the remote case, thus can be removed.

src/kernels/raw/finder/localKernelFinder.node.ts
rebornix: This is originally from https://github.com/microsoft/vscode-jupyter/blob/main/src/kernels/kernelFinder.base.ts#L213

rebornix:
@IanMatthewHuff thanks for the thorough review, the comments are great!

> It did look to me like in the two finders the following functions were very similar to each other
> writeToCache
> getFromCache

I started by moving the logics in `kernelFinder.base` into raw and jupyter finders and then did some tree shaking as we are doing redundant checks for results from both finders. Will visit this again in follow up PRs when Don and I looked into reducing the multiple caches we have.

> (kinda) listKernelsImpl
> listContributedKernels

I'm not very happy with the naming as ideally they should not be public, but currently we have tests that are calling them directly. I'll see how I can do better naming to make them more intuitive.

IanMatthewHuff (approved):

IanMatthewHuff:
> I started by moving the logics in `kernelFinder.base` into raw and jupyter finders and then did some tree shaking as we are doing redundant checks for results from both finders. Will visit this again in follow up PRs when Don and I looked into reducing the multiple caches we have.

Sounds great. Like I was saying, I like the change so approving. My one thought looking at this PR was that I've always found the caching logic pretty twisted, and I didn't want to have to touch it in two different locations, but sounds like that's already on the radar for follow ups.

rebornix:
@DonJayamanne I have resolved your comments and as we discussed offline yesterday, we will move the kernel ranking/matching code out of `IKernelFinder` and then we consolidate how we do caching. I'll merge now to unblock my follow up PRs but comments/reviews are still welcome and I'll send updates once you have more.

PR #11128, opened 2022-08-15
Fix password validations with jupyter server selection

### LINKED ISSUE #10300: Jupyter server with incorrect password validates too late
1. start an external juptyer server with a password
2. Specify server for connections in VS Code and use that server with the wrong password

Expected:
Error message that connection did not succeed

Actual:
The prompt goes away without any notification, but no connection was made.
Next time you try to connect to the server, no password prompt appears, but the connection fails because the (saved) password is wrong.

IanMatthewHuff:
Fixes #10300 

### DIFF
--- a/src/kernels/jupyter/launcher/jupyterPasswordConnect.ts
+++ b/src/kernels/jupyter/launcher/jupyterPasswordConnect.ts
@@ -8,6 +8,7 @@ import { IApplicationShell } from '../../../platform/common/application/types';
 import { IAsyncDisposableRegistry, IConfigurationService, IDisposableRegistry } from '../../../platform/common/types';
 import { DataScience } from '../../../platform/common/utils/localize';
 import { IMultiStepInputFactory, IMultiStepInput } from '../../../platform/common/utils/multiStepInput';
+import { traceInfo } from '../../../platform/logging';
 import { captureTelemetry, sendTelemetryEvent, Telemetry } from '../../../telemetry';
 import {
     IJupyterPasswordConnect,
@@ -51,7 +52,15 @@ export class JupyterPasswordConnect implements IJupyterPasswordConnect {
         // See if we already have this data. Don't need to ask for a password more than once. (This can happen in remote when listing kernels)
         let result = this.savedConnectInfo.get(newUrl);
         if (!result) {
-            result = this.getNonCachedPasswordConnectionInfo(newUrl);
+            result = this.getNonCachedPasswordConnectionInfo(newUrl).then((value) => {
+                // If we fail to get a valid password connect info, don't save the value
+                traceInfo(`Password for ${newUrl} was invalid.`);
+                if (!value) {
+                    this.savedConnectInfo.delete(newUrl);
+                }
+
+                return value;
+            });
             this.savedConnectInfo.set(newUrl, result);
         }
 
--- a/src/platform/common/utils/multiStepInput.ts
+++ b/src/platform/common/utils/multiStepInput.ts
@@ -266,6 +266,9 @@ export class MultiStepInput<S> implements IMultiStepInput<S> {
                             resolve(inputValue);
                         } else {
                             input.validationMessage = validationMessage;
+
+                            // On validation error make sure we are showing our input
+                            input.show();
                         }
                         input.enabled = true;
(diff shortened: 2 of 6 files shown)


src/kernels/jupyter/launcher/jupyterPasswordConnect.ts
```diff
@@ -51,7 +52,15 @@ export class JupyterPasswordConnect implements IJupyterPasswordConnect {
         // See if we already have this data. Don't need to ask for a password more than once. (This can happen in remote when listing kernels)
         let result = this.savedConnectInfo.get(newUrl);
         if (!result) {
-            result = this.getNonCachedPasswordConnectionInfo(newUrl);
+            result = this.getNonCachedPasswordConnectionInfo(newUrl).then((value) => {
+                // If we fail to get a valid password connect info, don't save the value
+                traceInfo(`Password for ${newUrl} was invalid.`);
+                if (!value) {
+                    this.savedConnectInfo.delete(newUrl);
```
IanMatthewHuff, line 59: This fixes the main issue. The connection info was getting saved regardless of if we validated it or now. So if you entered an incorrect password it will continue to use that for the same VS Code session without a chance to fix it.

src/platform/common/utils/multiStepInput.ts
```diff
@@ -266,6 +266,9 @@ export class MultiStepInput<S> implements IMultiStepInput<S> {
                             resolve(inputValue);
                         } else {
                             input.validationMessage = validationMessage;
+
+                            // On validation error make sure we are showing our input
+                            input.show();
```
IanMatthewHuff, line 271: Smaller fix here, with the new validation flow the main validation all happens with the server selection multistep input. But when coming back from the failed password validation the server selection input was not actually getting shown again (this was exposed by fixing the bug again). Make sure we show our input again so the customer can see the password error.

IanMatthewHuff:
Note, Rich's experimental new UI has changed the flow here some. In the old UI you'll see the validation message and get another chance to hit enter on the URL entry dialog. In the new experimental UI flow you'll get a modal warning dialog about the password failure, but in both cases you'll get a chance to re-input a failed password.

src/test/datascience/mockInputBox.ts
```diff
@@ -38,27 +41,31 @@ export class MockInputBox implements InputBox {
         return this.didHideEmitter.event;
     }
     public show(): void {
-        // After 10 ms set the value, then accept it
-        setTimeout(() => {
-            this.value = this._value;
-            this.didChangeValueEmitter.fire(this._value);
+        // Unless we have specified more, we only want to fire the input one time (in case show is called more)
```
IanMatthewHuff, line 44: Tweak to get unit tests working. Due to the validation change above show can be called more often. Show has been rigged in the mock to input and fire the value as if a user entered it. Instead, just do it once (with the option for more for one test that expected to call it multiple times with user input).

src/test/datascience/jupyterPasswordConnect.unit.test.ts
```diff
@@ -227,6 +233,101 @@ suite('JupyterPasswordConnect', () => {
         fetchMock.verifyAll();
     });
 
+    test('getPasswordConnectionInfo bad password followed by good password.', async () => {
```
IanMatthewHuff, line 236: New test to make sure that we don't cache a failed password.

src/test/datascience/jupyterPasswordConnect.unit.test.ts
```diff
@@ -117,6 +117,7 @@ suite('JupyterPasswordConnect', () => {
     }
 
     test('getPasswordConnectionInfo', async () => {
+        when(appShell.showInputBox(anything())).thenReturn(Promise.resolve('Python'));
```
IanMatthewHuff, line 120: Updated previous good password check to also actually check the password in the fetch.

IanMatthewHuff:
There are a couple of test failures, the run by line debugging ones look unrelated, but I'll check it out before pushing.

DonJayamanne (approved):

IanMatthewHuff:
Test failures look to just be #11137 so pushing.

PR #11129, opened 2022-08-15
Interactive window execution ordering

### LINKED ISSUE #10671: Cells executed from the input box can be executed out of order
1. open a python file without having an active IW open
2. execute the cell from the file

```
# %%
print(n)
```

3. After the IW is created and while the kernel is connecting, run `n=1` from the input box

Result: the second cell added (from the input box) is executed first
image

The execution count in the `CodeGenerator` is also out of sync now, so debugging and error stack traces do not behave well

amunger:
Fixes #10671

-   [ ] ~Has telemetry for feature-requests.~
-   [ ] ~[Test plan](https://github.com/Microsoft/vscode-jupyter/blob/main/.github/test_plan.md) is updated as appropriate.~
-   [ ] ~[`package-lock.json`](https://github.com/Microsoft/vscode-jupyter/blob/main/package-lock.json) has been regenerated by running `npm install` (if dependencies have changed).~

### DIFF
--- a/src/interactive-window/interactiveWindow.ts
+++ b/src/interactive-window/interactiveWindow.ts
@@ -237,7 +237,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         if (this.kernelConnectionMetadata) {
             this.pendingCellAdd = cellAddedPromise;
             const controller = this.controllerRegistration.get(this.kernelConnectionMetadata, 'interactive');
-            controller?.setPendingCellAddition(cellAddedPromise);
+            controller?.setPendingCellAddition(this.notebookDocument, cellAddedPromise);
         }
     }
 
--- a/src/interactive-window/interactiveWindow.ts
+++ b/src/interactive-window/interactiveWindow.ts
@@ -105,6 +105,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
     private _submitters: Uri[] = [];
     private fileInKernel: Uri | undefined;
     private cellMatcher: CellMatcher;
+    private pendingCellAdd: Promise<void> | undefined;
 
     private internalDisposables: Disposable[] = [];
     private kernelDisposables: Disposable[] = [];
@@ -228,12 +229,15 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         this.start();
     }
 
-    private setPendingCellAdd() {
+    /**
+     * Inform the controller that a cell is being added and it should wait before adding any others to the execution queue.
+     * @param cellAddedPromise - Promise that resolves when the cell execution has been queued
+     */
+    private setPendingCellAdd(cellAddedPromise: Promise<void>) {
         if (this.kernelConnectionMetadata) {
+            this.pendingCellAdd = cellAddedPromise;
             const controller = this.controllerRegistration.get(this.kernelConnectionMetadata, 'interactive');
-            const deferred = createDeferred<void>();
-            controller?.setPendingCellAddition(deferred.promise);
-            return deferred;
+            controller?.setPendingCellAddition(cellAddedPromise);
         }
     }
 
@@ -264,6 +268,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
                 // Id may be different if the user switched controllers
                 this.currentKernelInfo.controller = k.controller;
                 this.currentKernelInfo.metadata = k.kernelConnectionMetadata;
+                !!this.pendingCellAdd && this.setPendingCellAdd(this.pendingCellAdd);
                 this.updateSysInfoMessage(
                     this.getSysInfoMessage(k.kernelConnectionMetadata, SysInfoReason.Start),
                     false,
@@ -573,7 +578,8 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
 
         // Multiple cells that have split our code.
         const promises = cells.map((c) => {
-            const deferred = this.setPendingCellAdd();
+            const deferred = createDeferred<void>();
+            this.setPendingCellAdd(deferred.promise);
             // Add the cell first. We don't need to wait for this part as we want to add them
             // as quickly as possible
             const notebookCellPromise = this.addNotebookCell(c, fileUri, line);
--- a/src/interactive-window/interactiveWindow.ts
+++ b/src/interactive-window/interactiveWindow.ts
@@ -56,7 +56,11 @@ import {
     InteractiveTab
 } from './types';
 import { generateInteractiveCode, isInteractiveInputTab } from './helpers';
-import { IControllerSelection, IVSCodeNotebookController } from '../notebooks/controllers/types';
+import {
+    IControllerRegistration,
+    IControllerSelection,
+    IVSCodeNotebookController
+} from '../notebooks/controllers/types';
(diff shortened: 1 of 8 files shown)


src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -211,6 +212,16 @@ export class VSCodeNotebookController implements Disposable, IVSCodeNotebookCont
         traceInfoIfCI(`${ConsoleForegroundColors.Green}Posting message to Notebook UI ${messageType}`);
         return this.controller.postMessage(message, editor);
     }
+    /**
+     * A cell has been added to the notebook, so wait for the execution to be queued before handling any more execution requests.
+     * This only applies to the Interactive Window since cells are added from both the extension and core.
+     * @param promise A promise that resolves when the notebook is ready to handle more executions.
+     */
+    public setPendingCellAddition(promise: Promise<void>): void {
```
IanMatthewHuff, line 220: Feels a bit funny to have this just hanging off the main controller when it only applies to the IW. No reason other users couldn't just call it and it does nothing without a warning or error. 

Could we consider something like extending IVsCodeNotebookController to IVsCodeInteractiveNotebookController with viewType locked to InteractiveWindowView and this function available? Or maybe an error or warning if you call it from a notebook?

src/interactive-window/interactiveWindow.ts
```diff
@@ -558,33 +578,39 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
 
         // Multiple cells that have split our code.
         const promises = cells.map((c) => {
+            const deferred = createDeferred<void>();
+            this.setPendingCellAdd(deferred.promise);
```
DonJayamanne, line 582: I think what you can do is
remove these two lines 

```suggestion
```

src/interactive-window/interactiveWindow.ts
```diff
@@ -558,33 +578,39 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
+                                    .then((message) => this.addErrorMessage(message, cell))
+                            )
+                            .catch(noop);
+                    }
+                })
+                .finally(() => {
+                    deferred?.resolve();
+                });
             return promise;
         });
```
DonJayamanne, line 616: I think what you can do is

```suggestion
const deferred = createDeferred<void>();
this.setPendingCellAdd(deferred.promise);
promises[promises.length - 1].finally(() => deferred.resolve()).ignoreErrors();
```

src/notebooks/controllers/vscodeNotebookController.ts
DonJayamanne: If we were to change this, I'd suggest throwing an error instead of a warning, as it should never be called for non-IW

src/interactive-window/interactiveWindow.ts
```diff
@@ -222,6 +229,18 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         this.start();
     }
 
+    /**
+     * Inform the controller that a cell is being added and it should wait before adding any others to the execution queue.
+     * @param cellAddedPromise - Promise that resolves when the cell execution has been queued
+     */
+    private setPendingCellAdd(cellAddedPromise: Promise<void>) {
```
DonJayamanne, line 236: I think we should be passing in the Notebook Document as an argument, so that the promise is scoped to a particular notebook, not all IW that use this controller.

DonJayamanne (changes requested):
We need to create a map to ensure we don't block all IW

src/interactive-window/interactiveWindow.ts
```diff
@@ -558,33 +578,39 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
 
         // Multiple cells that have split our code.
         const promises = cells.map((c) => {
+            const deferred = createDeferred<void>();
```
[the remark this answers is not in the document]
amunger: I disagree - If the input box ends up adding a cell between the cells from a file, then it seems like they should just be executed in the order that they appear in the notebook document.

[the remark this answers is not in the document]
DonJayamanne: This is my point, if I hit run all for 5 cells, then immediately go into the input box and run a cell there,
then its possible i have  cell1, cell2, cell3, cell4, Inputbox, cell5
For me as a user, i'd expect it to be cell1, cell2, cell3, cell4, cell5, and then inputbox as thats the order in which i ran stuff.

However if you're fine with this, then i'm also ok. as this is what we discussed yesterday.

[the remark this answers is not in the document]
DonJayamanne: > just be executed in the order that they appear in the notebook document.

I would have thought it should be executed in the order the user ran the cells, not the order in which we add the cells into the UI. again, i'm not fussed about this.

src/interactive-window/interactiveWindow.ts
```diff
@@ -101,6 +105,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
     private _submitters: Uri[] = [];
     private fileInKernel: Uri | undefined;
     private cellMatcher: CellMatcher;
+    private pendingCellAdd: Promise<void> | undefined;
```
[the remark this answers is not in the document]
amunger: an InteractiveWindow only has one NotebookDocument

src/interactive-window/interactiveWindow.ts
amunger: added the parameter to the controller's method

src/interactive-window/interactiveWindow.ts
amunger: I'd rather have a promise per cell add

src/interactive-window/interactiveWindow.ts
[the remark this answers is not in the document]
amunger: the controller now has a map though to achieve this

src/notebooks/controllers/vscodeNotebookController.ts
amunger: an IW override does sound like a clearer solution, but I'd probably still just throw for the notebook implementation to avoid a cast. I'll stick with a type check for now to avoid having to deal with the controller registration changes that would be necessary for the polymorphism solution.

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -97,7 +97,7 @@ export class VSCodeNotebookController implements Disposable, IVSCodeNotebookCont
         selected: boolean;
         notebook: NotebookDocument;
     }>();
-    private pendingCellAddition: Promise<void> | undefined;
+    private pendingCellAdditions = new Map<NotebookDocument, Promise<void>>();
```
DonJayamanne, line 100: Might want to change this to a weakmap
```suggestion
    private pendingCellAdditions = new WeakMap<NotebookDocument, Promise<void>>();
```

src/interactive-window/interactiveWindow.ts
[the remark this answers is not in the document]
IanMatthewHuff: FWIW I'm not sure about a right answer, but I would expect what Don was expecting. My queued up cells, then the input box cell.

DonJayamanne (approved):

PR #11132, opened 2022-08-16
Enable the flag esModuleInterop

DonJayamanne:
Required for PR https://github.com/microsoft/vscode-jupyter/pull/11133
Where I'm using an npm package that requires this change, also this is a good change.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -24,6 +24,7 @@
         "removeComments": true,
         "useUnknownInCatchVariables": false,
         "strictPropertyInitialization": false,
+        "esModuleInterop": true,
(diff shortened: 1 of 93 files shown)


tsconfig.json
```diff
@@ -24,6 +24,7 @@
         "removeComments": true,
         "useUnknownInCatchVariables": false,
         "strictPropertyInitialization": false,
+        "esModuleInterop": true,
```
DonJayamanne, line 27: This is the change in the tsconfig.json file

IanMatthewHuff (approved):

amunger (approved):

PR #11139, opened 2022-08-16
Add comments to telemetry events

DonJayamanne:


src/kernels/execution/cellExecution.ts
```diff
@@ -297,7 +298,7 @@ export class CellExecution implements IDisposable {
     }
 
     private sendPerceivedCellExecute() {
-        const props = { notebook: true };
+        const props = { notebook: this.controller.notebookType === JupyterNotebookView };
```
IanMatthewHuff, line 301: I was about to suggest just a string here for view type, but I guess this matches up with the old existing boolean type.

IanMatthewHuff (approved):

src/kernels/execution/cellExecution.ts
DonJayamanne: My plan is to add the string from viewtype everywhere, in debt, so that we have better data.
Today we have `resourceType` in the events, plan is to add that.

PR #11147, opened 2022-08-17
Log python environments discovered

DonJayamanne:
Required to diagnose user issues

### DIFF
--- a/src/platform/api/pythonApi.ts
+++ b/src/platform/api/pythonApi.ts
@@ -452,7 +452,9 @@ export class InterpreterService implements IInterpreterService {
         traceVerbose(
             `Full interpreter list for ${getDisplayPath(resource)} is length: ${
                 allInterpreters.length
-            }, ${allInterpreters.map((item) => getDisplayPath(item.uri))}`
+            }, ${allInterpreters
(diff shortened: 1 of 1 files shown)


src/platform/api/pythonApi.ts
```diff
@@ -449,7 +449,13 @@ export class InterpreterService implements IInterpreterService {
         if (activeInterpreter) {
             allInterpreters.push(activeInterpreter);
         }
-        traceInfoIfCI(`Full interpreter list length: ${allInterpreters.length}`);
+        traceVerbose(
+            `Full interpreter list for ${getDisplayPath(resource)} is length: ${
+                allInterpreters.length
+            }, ${allInterpreters
+                .map((item) => `${item.displayName}:${item.envType}:${getDisplayPath(item.uri)}`)
```
amunger, line 456: There has been concern in the past about logging PII (e.g. username in the path) since we ask them to share logs in issues. Can we avoid that here?

IanMatthewHuff (approved):

src/platform/api/pythonApi.ts
DonJayamanne: If the user doesn't want to share it, then its upto them.
We logs paths in a number of places, errors contain stack traces, hence this is not a new issue.

PR #11156, opened 2022-08-18
Improve kernel finder caching

rebornix:
Re https://github.com/microsoft/vscode-jupyter/issues/10832. This is still on-going and require offline discussion with Don.


rebornix:
Updated the branch to reflect the discussions @DonJayamanne and I had offline about moving the kernel detection/finder to push model. The new workflow of kernel detection would be

* On extension activation, we initialize `KernelFinder`
  * `RemoteKernelFinder` start loading cache from global memento.
    * Once cache loaded, store in in memory
    * Kick off a kernelspec search without cache
  * `LocalKernelFinder` start loading cache from global memento.
    * Once cache loaded, store in in memory
    * Kick off a kernelspec search without cache
* Controller Loader activation
  * Request kernels from `KernelFinder`
  * `KernelFinder` waits for all contributed kernel finder to finish initialization (loading cache)
  * `KernelFinder` asks all contributed kernel finder for kernels, this step is synchronous.
* When interpreters, python extension, conda environement changes
  * Contributed kernel finders perform another round of kernel detection
  * Once get the latest kernels, update cache
  * Broadcast kernels change event
* Controller Loader receives the event
  * Request kernels from `KernelFinder`
  * `KernelFinder` asks all contributed kernel finder for kernels, this step is synchronous.


With this new workflow, the only asynchronous part is the initial cache loading. Once we have the cache in memory, it's each contributed kernel finder's responsibility to update the cache per environment change. The main `KernelFinder` and `ControllerLoader` are just reading their cache synchronously.

rebornix:
@DonJayamanne with the changes in this PR, we now have push model / event based `IContributedKernelFinders` which can

* Load cache on activation
* Respond to relevant events and update cache
* On cache update, broadcast cache update event

`KernelFinder` still behave the same as before, whose `listKernels` method is still asynchronous. Whether we want to change the semantics of this method or what's the best shape of it is worth of a discussion with @IanMatthewHuff (i.e., whether we want to differeciate Remote/Local, KernelSpec/Session).

This PR reduces redundant kernel research for irrelevant events, but it didn't try to make the cache update in each individual contributed kernel finder search operations minimal. For example, the Local Kernel Finder will trigger research when conda/python interpreter changes, but they might not be necessary for local kernel specs. Besides, we didn't do a diff between newly found kernels and old ones to reduce the events we broadcast. These can be improved in the coming debt week or along with the adoption of the new Python API.

I'll leave the the decision to you @DonJayamanne  when we should merge this PR.

src/kernels/jupyter/finder/remoteKernelFinder.ts
```diff
@@ -67,102 +91,128 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingle
...
         }
 
-        const kernelsFromCachePromise =
-            useCache === 'ignoreCache' ? Promise.resolve([]) : this.getFromCache(cancelToken);
-        let updateCache = true;
-        const kernelsWithoutCachePromise = (async () => {
-            const connInfo = await this.getRemoteConnectionInfo(cancelToken);
-            return connInfo ? this.listKernelsFromConnection(resource, connInfo) : Promise.resolve([]);
-        })();
+        const loadCancellationToken = new CancellationTokenSource();
```
IanMatthewHuff, line 164: I'll confess that I might just be ignorant on the cancellation tokens (I know I've gotten mixed on them before). But don't they have to be passed down or saved and hooked up to something to actually perform the cancellation? From my quick reading this didn't look to me like it was hooked up to anything that could actually cancel the load operation.

rebornix: This should be cleaned up. For single operation, we want to drop the cancellation token and for operations can be performed multiple times, we want to reuse the cancellation token to dispose previous requests. Will update soon.

src/kernels/jupyter/finder/remoteKernelFinder.ts
```diff
@@ -55,6 +67,18 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingle
...
+    onDidChangeKernels: Event<void> = this._onDidChangeKernels.event;
+
+    private _initializeResolve: () => void;
+    private _initializedPromise: Promise<void>;
+
+    get initialized(): Promise<void> {
+        return this._initializedPromise;
+    }
+
+    private wasPythonInstalledWhenFetchingControllers = false;
```
IanMatthewHuff, line 80: Minor, but might be nice to have a name change on this (it's not fetching controllers).

src/kernels/kernelFinder.ts
```diff
@@ -30,20 +36,20 @@ export class KernelFinder implements IKernelFinder {
...
+            return [];
+        }
 
-        allKernels.forEach((kernelList) => {
-            kernels.push(...kernelList);
-        });
+        const kernels: KernelConnectionMetadata[] = [];
+
+        for (const finder of this._finders) {
+            const contributedKernels = finder.listContributedKernels(resource);
```
IanMatthewHuff, line 49: I know part of this is bolting the new push model onto the old method of working. But seems like now might be the PR to remove the useCache property from the kernelFinder api that the ControllerLoader is using. I think it's only use for telemetry at this point, and having the Loader call this function twice with a parameter that essentially does nothing sounds nice to get rid of.

rebornix: 👍 with the new model, the telemetry might be no longer valid. Probably we can remove it and then add proper telemetry later.

IanMatthewHuff: Yeah, this was the bit that I was wondering if we could get rid of at this point in the refactor process:
https://github.com/microsoft/vscode-jupyter/blob/f7428fb799dc912aaeaee66463f6f3b000484850/src/notebooks/controllers/controllerLoader.ts#L169
With your change hitting here I don't think this makes sense at all anymore.

rebornix: I pushed https://github.com/microsoft/vscode-jupyter/pull/11156/commits/3b205e9c5d42f9bbc88a14588117acaec455fcad to remove the cache but it might require more cleaning.

src/test/client/api.vscode.test.ts
```diff
@@ -23,7 +23,8 @@ import { workspace } from 'vscode';
 import { executeSilently } from '../../kernels/helpers';
 import { getPlainTextOrStreamOutput } from '../../kernels/kernel';
 
-suite('3rd Party Kernel Service API', function () {
+// eslint-disable-next-line no-only-tests/no-only-tests
+suite.only('3rd Party Kernel Service API', function () {
```
DonJayamanne, line 27: oops

rebornix: this is for troubleshooting the test failures.

src/kernels/jupyter/finder/remoteKernelFinder.ts
```diff
@@ -67,102 +92,137 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingle
...
+                this.updateCache().then(noop, noop);
             }
+        });
+
+        // For kernel dispose we need to wait a bit, otherwise the list comes back the
+        // same
+        this.kernelProvider.onDidDisposeKernel(
+            (k) => {
+                if (k && isRemoteConnection(k.kernelConnectionMetadata)) {
+                    setTimeout(() => {
```
DonJayamanne, line 133: Can we capture the return value so that the `timer` can be disposed in the `dipose` method.
This ensures we don't have any timers hanging around that could cause the node process from hanging instead of shutting down gracefully.
E.g. if i were to create this class and dispoose this immediately and end the tests, it would actually hang due to pending timers.

We've had a lot of issues in the past due to such dangling timers.

src/kernels/jupyter/finder/remoteKernelFinder.ts
```diff
@@ -67,102 +92,137 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingle
...
+            kernels = await kernelsWithoutCachePromise;
+        }
+
+        await this.writeToCache(kernels);
+        this._initializeResolve();
+    }
+
+    private async updateCache() {
+        let kernels: RemoteKernelConnectionMetadata[] = [];
+        this._cacheUpdateCancelTokenSource?.dispose();
```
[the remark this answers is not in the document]
DonJayamanne: Never mind, that will not be possible.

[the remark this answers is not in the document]
DonJayamanne: I think what you'll need to do is, check if the cancellation token is cancelled or not before writing to the cache and triggering the change event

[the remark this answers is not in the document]
DonJayamanne: Damn it, its ok, thats what we have today,

src/kernels/jupyter/finder/remoteKernelFinder.ts
```diff
@@ -67,102 +92,137 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingle
...
+
+        await this.writeToCache(kernels);
+
+        this._onDidChangeKernels.fire();
+    }
+
+    /**
+     * Used in test only
+     */
+    async _updateCacheForTest() {
```
DonJayamanne, line 217: Can you not mark `loadCache` as public that way you dont need a method specifically for tests.

DonJayamanne: No need to change, just a thought

src/kernels/raw/finder/localKernelFinder.node.ts
```diff
@@ -4,134 +4,226 @@
...
-                    updateCache = true;
-                }
+                kernels = await this.listKernels(resource, updateCacheCancellationToken.token);
             } catch (ex) {
                 traceError(`Exception loading kernels: ${ex}`);
+                return;
             }
+
+            // update resource cache
+            const resourceCacheKey = this.getResourceCacheKey(resource);
```
DonJayamanne, line 140: You'd need to return from here if the cancellation token has been cancelled.
Else we end up updating emptu items into the list and triggering the event.

src/kernels/raw/finder/localKernelFinder.node.ts
```diff
@@ -4,134 +4,226 @@
...
+            this.resourceCache.set(resourceCacheKey, kernels);
+
+            this._onDidChangeKernels.fire();
+        } else if (workspace.workspaceFolders) {
+            await Promise.all(
+                workspace.workspaceFolders.map((folder) => () => {
+                    return new Promise<void>(async (resolve) => {
+                        try {
+                            let kernels = await this.listKernels(folder.uri, updateCacheCancellationToken.token);
+                            await this.writeToCache(kernels);
```
DonJayamanne, line 150: Same as earlier, check if the cancel token has been cancelled and then return

src/kernels/raw/finder/localKernelFinder.node.ts
```diff
@@ -4,134 +4,226 @@
...
+                        } catch (ex) {
+                            traceError(`Exception loading kernels: ${ex}`);
+                        }
+                        resolve();
+                    });
+                })
+            );
+        }
+
+        this._onDidChangeKernels.fire();
```
DonJayamanne, line 162: Same as earlier, check if the cancel token has been cancelled and then return

DonJayamanne: Basically we shouldn't be updating the list with an empty list and triggering the event, else the callers could end up getting an empty list
all because we have this method getting executed twice.

DonJayamanne (approved):
Looks good, just got on issue with a possible race condition causing the kernel list to be empty

PR #11163, opened 2022-08-20
Fix/Reenable some RBL tests

### LINKED ISSUE #11137: Run by Line stops at the wrong line for a function call
test `Stops in same-cell function called from last line` is failing

Manual repro show there is a bug - the jump from the function call should go to line 2:
RBL

roblourens:
Fixes #11137

The same-cell test just needs to be adjusted for new behavior in debugpy 1.6.3, which is actually better.

The second test seems fine and I'll investigate if it's flaky

The third wasn't checking outputs correctly, and needs a slower timeout when stepping, that is fixed

And i also launch tests with a fixed extensions-dir, which is helpful when debugging, otherwise it uses my real code-insiders extensions dir

### DIFF
--- a/src/test/standardTest.node.ts
+++ b/src/test/standardTest.node.ts
@@ -56,7 +56,7 @@ async function createTempDir() {
 /**
  * Smoke tests & tests running in VSCode require Python extension to be installed.
  */
-async function installPythonExtension(vscodeExecutablePath: string) {
+async function installPythonExtension(vscodeExecutablePath: string, extensionsDir: string) {
     // Pick python extension to use based on environment variable. Insiders can be flakey so
     // have the capability to turn it off/on.
     const pythonVSIX =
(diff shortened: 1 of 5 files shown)


lramos15 (approved):

src/test/standardTest.node.ts
```diff
@@ -67,16 +67,16 @@ async function installPythonExtension(vscodeExecutablePath: string) {
         console.info('Python Extension not required');
         return;
     }
-    console.info(`Installing Python Extension ${pythonVSIX}`);
+    console.info(`Installing Python Extension ${pythonVSIX} to ${extensionsDir}`);
     const cliPath = resolveCliPathFromVSCodeExecutablePath(vscodeExecutablePath);
-    spawnSync(cliPath, ['--install-extension', pythonVSIX], {
+    spawnSync(cliPath, ['--install-extension', pythonVSIX, '--extensions-dir', extensionsDir], {
```
IanMatthewHuff, line 72: Nice, seems like a good change for test independence.

IanMatthewHuff (approved):

PR #11195, opened 2022-08-25
Add support for Kerberized Jupyter servers

pangav2001:
Made necessary changes to add support for authenticating with a Kerberized Jupyter Server. 

In the sake of fairness, I'd also like to provide a [gist](https://gist.github.com/mlucool/8c9dfb5767269c1ab2603d173576dea1) by [@mlucool](https://github.com/mlucool) showing how to make the necessary changes to the server-side code to support Kerberos Authentication.

After considering a lot of alternatives for making VSCode work with a Kerberized Jupyter Server, augmenting the extension to support this directly seemed to be the simplest, cleanest way of doing it without breaking any of the existing features. 

Having said that, I realize that adding a native dependency ([kerberos](https://www.npmjs.com/package/kerberos)) to the project might not be ideal / desirable. I'm more than happy to take suggestions for what would be an acceptable way of doing things from the extension. 
A couple of ideas are:
- Finding a way to use VSCode's ability to [negotiate](https://code.visualstudio.com/docs/setup/network#_authenticated-proxies) using Kerberos from our extension. However, there seems to be an [open issue](https://github.com/microsoft/vscode/issues/12588) since 2016 asking to expose this functionality to extensions via an API.
- I believe it would be a fair assumption to say that if someone wants to use Kerberos, then they would already have the binaries for interacting with GSSAPI or SSPI on their machines, so maybe it would be possible to add an extra .js file that would detect the OS used, locate those binaries and call them. Since we only need to (1) get the credentials, (2) initialize the security context and step once, and (3) delete the context, this wouldn't result in a lot of added code while at the same time enabling us to get rid of the kerberos dependency that can do a lot more things that we don't need.

List of commits:
- [Added Kerberos authentication support](https://github.com/microsoft/vscode-jupyter/commit/5bfb1cb755ace949cc272d2b5a75d5dadeba26c5)
- [Changed telemetry constants for Kerb and Pass auth](https://github.com/microsoft/vscode-jupyter/commit/c67ccffe952b5f63df76575bf6a5cb313ea71d88)
- [Added kerberos+types dependencies and reinstalled](https://github.com/microsoft/vscode-jupyter/commit/33c7c238f85a9fa35db4a7a73b5f5db627090157)
- [Added news entry](https://github.com/microsoft/vscode-jupyter/commit/f023a658e2669ae8c20c78bea6d82d7d512d237d)

Fixes #


pangav2001:
@DonJayamanne any thoughts / next steps on this?

DonJayamanne:
Thanks for the contribution. Apologies for not getting back, was unable to check this out during the weekend.
I'll have a look at this during the course of the week, I believe we can get this done via a contribution from a separate extension.

Thanks again.

pangav2001:
No need to apologize! I'll be eagerly waiting for any comments; appreciate you taking the time to review this. By the way, could you point me to the 'contribution from a separate extension' so I could take a look?

On an unrelated note, I couldn't find any mention of the [CLA](https://cla.opensource.microsoft.com/microsoft/vscode-jupyter?pullRequest=11195) before putting up the PR. I might have missed it, but I think it'd be good to add this in CONTRIBUTING.md; let me know if you'd like me to open an issue for this.

DonJayamanne:
> ouldn't find any mention of the [CLA](https://cla.opensource.microsoft.com/microsoft/vscode-jupyter?pullRequest=11195) before putting up the PR. I might

Thats an automated check, added for all Msft repos. Not sure adding the guidelines would help, as the links are generated based on a PR

pangav2001:
> the links are generated based on a PR

Seems like https://cla.opensource.microsoft.com/ is a static link. I believe it would be useful for some contributors to know about this beforehand; a CLA might be something that's subject to contract review, and learning about it only after contributing could be inconvenient.

pangav2001:
> I'll have a look at this during the course of the week, I believe we can get this done via a contribution from a separate extension.

Hey @DonJayamanne, did you get a chance to look at this?

pangav2001:
Hi @DonJayamanne , the PR itself is a bit outdated now. However, could you please add your thoughts on integrating Kerberos authentication in the extension?

DonJayamanne:
@pangav2001 
I've looked at this PR and have a few suggestions:
* Please could you create an issue to track this work (as you can see there's a new entry in this PR but no corresponding issue that it fixes)
* > Finding a way to use VSCode's ability to [negotiate](https://code.visualstudio.com/docs/setup/network#_authenticated-proxies) using Kerberos from our extension.
What's the extension you are referring to as `our` here. Is it Jupyter extension or another
* adding a native dependency ([kerberos](https://www.npmjs.com/package/kerberos)) to the project might not be ideal / desirable
I agree, we only have one native dependency in this extension for ZMQ and have enough issues with that as it is. I'd prefer to limit this to just zmq.
* s to add support for authenticating with a Kerberized Jupyter Server.
Can you provide more information about this. Is this just for your self or is this for some other product or server. 
Basically trying to understand how many users would end up using this feature.
Generally we have issues in the repo that track such work, and from the engagement (upvotes/comments) we can determine the interest/number of users who'd end up using this and evaluate the impact.
However without that I can only assume that you are the only one who'd end up using this feature.

**Alternative solution**
* Create your own extension that contributes a Jupyter server to the extension via the API defined here [registerRemoteServerProvider](https://github.com/microsoft/vscode-jupyter/blob/main/src/standalone/api/api.ts#L51)
Here's an excerpt of the API
* Basically you can provide the Uri of the Jupyter server to the extension
* & if and when the user chooses to connect to that we'll ask for the server info via `IJupyterServerUri` which would then contain all of the auth information such as the `authorizationHeader` and the like.
* & if that's still not sufficient, we can make the necessary changes to this to ensure it meets your needs.

I think this would be a better approach as:
* You could create an extension that says `Jupyter Server with Kerberos Auth` or the like in the marketplace place
* Its a completely different and stand alone extension with its own dependencies, what ever they may be.

```typescript
export interface IJupyterServerUri {
    baseUrl: string;
    token: string;
    authorizationHeader: any; // JSON object for authorization header.
    expiration?: Date; // Date/time when header expires and should be refreshed.
    displayName: string;
}

export type JupyterServerUriHandle = string;

export interface IJupyterUriProvider {
    /**
     * Should be a unique string (like a guid)
     */
    readonly id: string;
    onDidChangeHandles?: Event<void>;
    getQuickPickEntryItems?(): QuickPickItem[];
    handleQuickPick?(item: QuickPickItem, backEnabled: boolean): Promise<JupyterServerUriHandle | 'back' | undefined>;
    /**
     * Given the handle, returns the Jupyter Server information.
     */
    getServerUri(handle: JupyterServerUriHandle): Promise<IJupyterServerUri>;
    /**
     * Gets a list of all valid Jupyter Server handles that can be passed into the `getServerUri` method.
     */
    getHandles?(): Promise<JupyterServerUriHandle[]>;
}
```

pangav2001:
Hi, thanks for providing your thoughts! 

> * Please could you create an issue to track this work (as you can see there's a new entry in this PR but no corresponding issue that it fixes)
I could only create a bug report issue, so instead, I've opted for starting a new [discussion](https://github.com/microsoft/vscode-jupyter/discussions/11512). 
> * > Finding a way to use VSCode's ability to [negotiate](https://code.visualstudio.com/docs/setup/network#_authenticated-proxies) using Kerberos from our extension.
>   > What's the extension you are referring to as `our` here. Is it Jupyter extension or another
`our` was a mistype, I meant the vscode-jupyter extension.
> * adding a native dependency ([kerberos](https://www.npmjs.com/package/kerberos)) to the project might not be ideal / desirable
Agreed
>   I agree, we only have one native dependency in this extension for ZMQ and have enough issues with that as it is. I'd prefer to limit this to just zmq.
> * s to add support for authenticating with a Kerberized Jupyter Server.
>   Can you provide more information about this. Is this just for your self or is this for some other product or server.
>   Basically trying to understand how many users would end up using this feature.
>   Generally we have issues in the repo that track such work, and from the engagement (upvotes/comments) we can determine the interest/number of users who'd end up using this and evaluate the impact.
>   However without that I can only assume that you are the only one who'd end up using this feature.
I am explaining the motivation behind this in the discussion.
> 
> **Alternative solution**
> 
> * Create your own extension that contributes a Jupyter server to the extension via the API defined here [registerRemoteServerProvider](https://github.com/microsoft/vscode-jupyter/blob/main/src/standalone/api/api.ts#L51)
>   Here's an excerpt of the API
> * Basically you can provide the Uri of the Jupyter server to the extension
> * & if and when the user chooses to connect to that we'll ask for the server info via `IJupyterServerUri` which would then contain all of the auth information such as the `authorizationHeader` and the like.
> * & if that's still not sufficient, we can make the necessary changes to this to ensure it meets your needs.
> 
> I think this would be a better approach as:
> 
> * You could create an extension that says `Jupyter Server with Kerberos Auth` or the like in the marketplace place
> * Its a completely different and stand alone extension with its own dependencies, what ever they may be.
> 
> ```ts
> export interface IJupyterServerUri {
>     baseUrl: string;
>     token: string;
>     // eslint-disable-next-line @typescript-eslint/no-explicit-any
>     authorizationHeader: any; // JSON object for authorization header.
>     expiration?: Date; // Date/time when header expires and should be refreshed.
>     displayName: string;
> }
> 
> export type JupyterServerUriHandle = string;
> 
> export interface IJupyterUriProvider {
>     /**
>      * Should be a unique string (like a guid)
>      */
>     readonly id: string;
>     onDidChangeHandles?: Event<void>;
>     getQuickPickEntryItems?(): QuickPickItem[];
>     handleQuickPick?(item: QuickPickItem, backEnabled: boolean): Promise<JupyterServerUriHandle | 'back' | undefined>;
>     /**
>      * Given the handle, returns the Jupyter Server information.
>      */
>     getServerUri(handle: JupyterServerUriHandle): Promise<IJupyterServerUri>;
>     /**
>      * Gets a list of all valid Jupyter Server handles that can be passed into the `getServerUri` method.
>      */
>     getHandles?(): Promise<JupyterServerUriHandle[]>;
> }
> ```
> 
Thanks a lot for providing this alternative approach! I am not quite sure if this would fit the purpose, I'll need to look more into it and get back to you.

DonJayamanne:
@pangav2001 I've created an issue for you here https://github.com/microsoft/vscode-jupyter/issues/11518

PR #11229, opened 2022-08-30
Pass one telemetry ownership update.

IanMatthewHuff:


src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -580,26 +579,8 @@ export class VSCodeNotebookController implements Disposable, IVSCodeNotebookCont
             traceInfo('Switch kernel did not change kernel.');
             return;
         }
-        switch (this.connection.kind) {
```
IanMatthewHuff, line 583: Old code here was to distinguish between existing kernels and "interpreters as kernels" the other code path doesn't exist anymore, and the language info here it capture in the Telemetry.SwitchKernel event below.

src/kernels/jupyter/launcher/commandLineSelector.ts
```diff
@@ -36,7 +36,6 @@ export class JupyterCommandLineSelector {
         workspaceService.onDidChangeConfiguration(this.onDidChangeConfiguration.bind(this));
     }
 
-    @capturePerfTelemetry(Telemetry.SelectJupyterURI)
```
IanMatthewHuff, line 39: This looks like it's been wrong forever. Glad we checked. Would have had too many entries for this event.

DonJayamanne (approved):

src/telemetry.ts
```diff
@@ -1626,23 +1823,119 @@ export class IEventNamePropertyMapping {
...
+         * Count of columns in the target data frame.
+         */
+        columns: number | undefined;
+    }> = {
+        owner: 'IanMatthewHuff',
+        source: 'N/A',
+        feature: ['DataFrameViewer'],
+        properties: {
+            rows: {
+                classification: 'PublicNonPersonalData',
```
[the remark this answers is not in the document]
IanMatthewHuff: @DonJayamanne I went back and forth on it a bit. From my reading I thought that it might be PublicNonPersonal as it's not under our control, it's up to the customer who had made the size of the data frame. But it's non personal.

Reading the categories more, you might be right? Feel it's a bit hard to tell.

PR #11231, opened 2022-08-30
Fixes to ensure we are forward compatible with TS 4.8

DonJayamanne:


src/telemetryGenerator.node.ts
```diff
@@ -823,10 +830,7 @@ export default async function generateTelemetryOutput() {
         });
     });
     // Print out the source tree
-    generateDocumentation(files, {
-        target: ts.ScriptTarget.ES5,
-        module: ts.ModuleKind.CommonJS
```
DonJayamanne, line 828: This is hacky, we weren't using the tsconfig at all. The new change ensures we use our tsconfig when attempting to parse the ts files.

amunger (approved):

PR #11251, opened 2022-08-31
Ensure we have one way to generate hashes

DonJayamanne:
This is just step 1 of using a different hashing package.


src/platform/common/crypto.ts
```diff
@@ -6,33 +6,23 @@
...
-            return result as any;
-        }
-        return hash as any;
+    public createHash(data: string, algorithm: 'SHA512' | 'SHA256' = 'SHA256'): string {
+        return computeHash(data, algorithm);
+    }
+}
+
+export function computeHash(data: string, algorithm: 'SHA512' | 'SHA256') {
+    if (algorithm === 'SHA256') {
```
DonJayamanne, line 23: This is one of the only two places where we generate hashes,
plan is to merge the others with this and have this code to return the async version of hashing, which would be a bigger change (hence a separate PR).
Big chage because the new hashing code will be async.

src/telemetryGenerator.node.ts
```diff
@@ -1,6 +1,11 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+// This line should always be right on top.
```
IanMatthewHuff, line 4: Was this intended on being in?

DonJayamanne: Yes, this is requried, beacuse i'm changing some of the code, and now this code ends up pulling a few of our extension code which relines on this being set. Our code only has 1 entry point, extension.ts 
this file is called from node.js which ends up pulling in some of our extension code, i.e. we have another entry point which pulls in other stuff and things can fall over when running this. Hope that makes sense.

IanMatthewHuff (approved):

PR #11252, opened 2022-08-31
Turn ICryptoUtils.createHash into an async function

DonJayamanne:


src/platform/common/types.ts
```diff
@@ -288,7 +288,7 @@ export interface ICryptoUtils {
      * @param data The string to hash
      * @param [algorithm]
      */
-    createHash(data: string, algorithm?: 'SHA512' | 'SHA256'): string;
+    createHash(data: string, algorithm?: 'SHA512' | 'SHA256'): Promise<string>;
```
DonJayamanne, line 291: This is the change, from sync to async

src/platform/common/crypto.ts
```diff
@@ -14,7 +14,7 @@ import * as hashjs from 'hash.js';
  */
 @injectable()
 export class CryptoUtils implements ICryptoUtils {
-    public createHash(data: string, algorithm: 'SHA512' | 'SHA256' = 'SHA256'): string {
+    public async createHash(data: string, algorithm: 'SHA512' | 'SHA256' = 'SHA256'): Promise<string> {
```
amunger, line 17: We're just pretending it's async? Why would we want to do this?

DonJayamanne: next pr will add the real bits, wanted to break this work into smaller chunks. one pr already touched almost 70 files

DonJayamanne: i merged another pr into this, this pr now has 74 files changed.
hopefully it makes sense why im trophy to break the work into multiple prs.

again, another pr will add the others changes, adding that here will only make this much larger and more difficult to review

src/platform/common/crypto.ts
```diff
@@ -14,12 +14,12 @@ import * as hashjs from 'hash.js';
...
         return computeHash(data, algorithm);
     }
 }
 
-export function computeHash(data: string, algorithm: 'SHA512' | 'SHA256' | 'SHA1') {
+export async function computeHash(data: string, algorithm: 'SHA512' | 'SHA256' | 'SHA1') {
```
DonJayamanne, line 22: This is the other change, changed from sync to async.
This change was supposed to be in another PR, but was accidentally meged into this, now we have 74 file changes here.

IanMatthewHuff (approved):
It's kinda a pain that we have to do this (wish there was an approved sync option, seems like it should be a small operation). But code is fine.

DonJayamanne:
> wish there was an approved sync option, seems like it should be a small operation

agreed

PR #11254, opened 2022-09-01
Turn hashing function into an async function

DonJayamanne:


src/platform/common/crypto.ts
```diff
@@ -19,7 +19,7 @@ export class CryptoUtils implements ICryptoUtils {
     }
 }
 
-export function computeHash(data: string, algorithm: 'SHA512' | 'SHA256' | 'SHA1') {
+export async function computeHash(data: string, algorithm: 'SHA512' | 'SHA256' | 'SHA1') {
```
DonJayamanne, line 22: This single change had a cascading effect, 
I.e. all of the changes in this PR are a result of this one change, changes from sync to async.

amunger: why would this be async when there's no await in this function?

DonJayamanne: next PR will add that code

PR #11263, opened 2022-09-02
Use a new hashing lib

DonJayamanne:
Remove the usage of `hash.js` and use the `mscrypto.js` library from  here https://github.com/microsoft/MSR-JavaScript-Crypto


package.json
```diff
@@ -2192,15 +2192,13 @@
         "bootstrap": "^4.3.1",
         "bootstrap-less": "^3.3.8",
         "cross-fetch": "^3.1.5",
-        "crypto": "^1.0.1",
```
DonJayamanne, line 2195: No idea why this was added in the first place, its an empty package
Possible some hacky way to dismiss some warnings , no idea.
Removing this as this is not required and marked as deprecated

package.json
```diff
@@ -2192,15 +2192,13 @@
...
         "decompress-response": "^4.2.0",
         "encoding": "^0.1.13",
         "fast-deep-equal": "^2.0.1",
         "font-awesome": "^4.7.0",
         "format-util": "^1.0.5",
         "fs-extra": "^4.0.3",
         "glob": "^7.1.2",
-        "hash.js": "^1.1.7",
```
DonJayamanne, line 2203: This is what we want to do, get rid of hash.js

build/webTestReporter.js
```diff
@@ -10,9 +10,9 @@ const dedent = require('dedent');
 const { EventEmitter } = require('events');
 const colors = require('colors');
 const core = require('@actions/core');
-const hashjs = require('hash.js');
 const glob = require('glob');
 const { ExtensionRootDir } = require('./constants');
+const { computeHash } = require('../src/platform/msrCrypto/hash');
```
DonJayamanne, line 15: Re-use common approhach

build/webpack/webpack.datascience-ui.config.builder.js
```diff
@@ -198,6 +198,7 @@ function buildConfiguration(bundle) {
             fallback: {
                 fs: false,
                 path: require.resolve('path-browserify'),
+                crypto: require.resolve(path.join(constants.ExtensionRootDir, 'src/platform/msrCrypto/msrCrypto.js')),
```
DonJayamanne, line 201: jupyter packages needs this.

src/platform/msrCrypto/hash.js
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Use window crypto if it's available, otherwise use the msrCrypto module, else fall back to importing.
+const windowCrypto =
```
DonJayamanne, line 5: Created this so that we have a common library to generate hashes, this is used in a number of places even in JS code.

DonJayamanne: All of the hashing code eventually comes here, i.e. we now have a single way to generate hashes and this is the function that does that.

DonJayamanne: This is a simple wrapper around the mscrypto js file,

src/telemetryGenerator.node.ts
```diff
@@ -776,6 +776,11 @@ function generateDocumentation(fileNames: string[]): string | undefined {
                             const properties = comptePropertyGroupsForIntersectionTypes(type, typeChecker);
                             const comment = getCommentForUnions(type);
                             groups.push({ description: comment, properties });
+                        } else if (
+                            type.kind === ts.SyntaxKind.UndefinedKeyword ||
+                            type.kind === ts.SyntaxKind.NeverKeyword
```
DonJayamanne, line 781: Minor tweek to the telemetry generator, we were generating warnings about this in subsequent line when we didn't have to.

amunger (approved):

roblourens (approved):

PR #11269, opened 2022-09-02
enable settings telemetry

amunger:
https://github.com/microsoft/vscode-jupyter-internal/issues/340

re-enabled telemetry for user settings. For string-based settings, only `interactiveWindowMode` has been white-listed to send the actual value.

### DIFF
--- a/TELEMETRY.csv
+++ b/TELEMETRY.csv
@@ -2526,6 +2526,8 @@ selectAnotherInterpreter - Selected another interpreter.","","'notSelected'
 'selected' 
 'installationCancelled' 
 'selectAnotherInterpreter' ",true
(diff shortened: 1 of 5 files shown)


DonJayamanne:
there's address at least 2 settings that include paths.
jupyter.notebookFileRoot and debugpydistpath.
also the code in place doesn't protect us from creating another similar steering setting that could capture similar information.
ideally we should modify the code to exclude anything other than numbers or booleans, ie filter the json value do that's its future proof, as it has already failed in assuming that the current naive approach works.
once that's done we don't need to json stringify it, this can send the properties as a full list, but that's optional

DonJayamanne (changes requested):
oops, meant to request changes with comments, not close

amunger:
> there's are at least 2 settings that include paths.

All string settings are translated to 'default' or 'non-default', unless they are white listed (only interactiveWindowMode right now)

DonJayamanne (approved):

PR #11271, opened 2022-09-02
github action to collect and upload past test results

amunger:
By doing this in a Github action, we can use the automatic auth token (which hopefully has the right permissions) and get the heavy tasks of downloading/parsing json files done incrementally to speed up the process


amunger:
I probably need to install the `requests` package in the python environment for this to work

roblourens (approved):

PR #11277, opened 2022-09-02
split the python steps for the test aggregator job

amunger:
fixing:
Error: .github#L1
a step cannot have both the `uses` and `run` keys


DonJayamanne:
would be good to have a more meaningful name in the commit message

DonJayamanne (approved):
please change the commit message when committing

PR #11278, opened 2022-09-02
Avoid rewriting package.json, it's unnecessary. Set a user setting instead

roblourens:
It seems to work the same if I set this logging setting in the user settings.json. Is this ok @DonJayamanne or is there something I'm missing?


roblourens:
As soon as I sent this @DonJayamanne I realized that maybe this is needed for web tests? I don't know how to set user settings for web tests, seems like there is probably a way but I don't know what it is. So we can leave it alone for web tests but I'd like to do this for desktop tests if that's ok with you.

DonJayamanne (approved):
yes this should work given we create a user setting file now, i think that code u changed is before i created a user setting file

PR #11323, opened 2022-09-07
interpolate auth header string correctly

amunger:
I was not adding the auth header correctly and getting a 401 when retrieving the artifacts.


DonJayamanne (approved):

DonJayamanne (changes requested):
You might wnt to remove this from the logs
`print("Using authtoken with prefix: " + authtoken[:4])`

DonJayamanne (approved):

PR #11326, opened 2022-09-07
Ensure we properly identify an insider/pre-relase extension

DonJayamanne:


joyceerhl (approved):
Our guidance for extensions which publish pre-releases is that extensions should have an even minor version for stable releases and an odd minor version for pre-releases: https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions

I realize Jupyter has a YYYY.MM.DD___ extension versioning scheme, but if you adopted the pre-release numbering convention, you'd be able to tell if the extension is a pre-release by looking at the extension version from the package.json via `vscode.ExtensionContext`.

DonJayamanne:
>  extensions which publish pre-releases is that extensions should have an even minor version for stable releases and an odd minor version for 

Yes, but we decided not to change our versioning mechanism, as this helps us easily identify releases.

PR #11350, opened 2022-09-11
Optimize to handling of large outputs in notebooks

### LINKED ISSUE #11031: VS Code crashes when running a cell that produces a lot of output
I am trying to print model summary in tensorflow and I think the model is large and it's crashing my notebook. The model is ResNet101.

image

The whole computer comes to a halt, memory usage goes up to 99% and VS Code crashes. I have 16 GB of ram, so I didn't think printing something large would actually eat all my ram. Also, because the kernel crashes, all the variables are lost like history = model.fit() which I need to fine-tune the model afterwards. Moreover, I need to print base_model summary in order to choose from which layer to fine-tune from.

Is there a way to print the summary in another way and save the entire notebook with the variables, so I can continue working? I have checkpoints for model weights, but I need to keep track of past epochs through history to resume training afterwards.

I will try using`.py` files, but I want to know if there is a way to solve this problem for jupyter.

DonJayamanne:
Fixes #11031


roblourens (approved):

roblourens:
🎉🎉🎉

rebornix (approved):

roblourens (approved):

PR #11354, opened 2022-09-12
Fix incorrect event with incorrect properties

lramos15:
Not sure what this properties and measures key is, but this is incorrect syntax so I've removed it


joyceerhl (approved):

lramos15:
@DonJayamanne Need assistance again with the auto generation portion

DonJayamanne (approved):

PR #11355, opened 2022-09-12
Sort properties so its easier to find something

DonJayamanne:
While going through the GDPR properties (CSV genearted by Logan) i found that its easier to have the properties sorted

### DIFF
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -1043,11 +1043,8 @@ export class IEventNamePropertyMapping {
         feature: ['InteractiveWindow', 'Notebook'],
         tags: ['Cell Execution'],
         source: 'N/A',
-        properties: commonClassificationForResourceSpecificTelemetryProperties().properties,
-        measures: {
-            ...commonClassificationForDurationProperties(),
-            ...commonClassificationForResourceSpecificTelemetryProperties().measures
-        }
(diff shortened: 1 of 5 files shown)


src/telemetry.ts
```diff
@@ -1043,11 +1043,8 @@ export class IEventNamePropertyMapping {
         feature: ['InteractiveWindow', 'Notebook'],
         tags: ['Cell Execution'],
         source: 'N/A',
-        properties: commonClassificationForResourceSpecificTelemetryProperties().properties,
-        measures: {
-            ...commonClassificationForDurationProperties(),
-            ...commonClassificationForResourceSpecificTelemetryProperties().measures
-        }
```
DonJayamanne, line 1050: These telemetry events incorrectly defined, we didn't need these additional properties

src/telemetry.ts
```diff
@@ -1059,11 +1056,8 @@ export class IEventNamePropertyMapping {
         feature: ['InteractiveWindow', 'Notebook'],
         tags: ['Cell Execution'],
         source: 'N/A',
-        properties: commonClassificationForResourceSpecificTelemetryProperties().properties,
-        measures: {
```
DonJayamanne, line 1063: These telemetry events incorrectly defined, we didn't need these additional properties

PR #11378, opened 2022-09-15
no unclickable links in server connection dialogs in the original or experimental picker

### LINKED ISSUE #11285: Don't show unclickable links in dialogs
image

We usually show "Learn more..." buttons instead.

IanMatthewHuff:
Fixes #11285 


src/kernels/jupyter/serverSelector.ts
```diff
@@ -587,12 +588,16 @@ class JupyterServerSelector_Original implements IJupyterServerSelector {
                 if (!handled) {
                     return DataScience.jupyterSelfCertExpiredErrorMessageOnly();
                 }
-            } else if (!this.isWebExtension) {
-                return DataScience.remoteJupyterConnectionFailedWithoutServerWithError().format(
-                    err.message || err.toString()
-                );
             } else {
-                return DataScience.remoteJupyterConnectionFailedWithoutServerWithErrorWeb().format(
+                // Notify the user of web extension specific connection help
```
IanMatthewHuff, line 592: We have two different server selection flows currently. This is the original flow that uses an input box. For this scenario keeping the validation returned is nice. So I broke this up into a smaller specific error message for the web specific connection error info, then the normal connection error message that shows in the red validation box.

src/platform/common/utils/multiStepInput.ts
```diff
@@ -206,7 +206,7 @@ export class MultiStepInput<S> implements IMultiStepInput<S> {
                                     input.enabled = false;
                                     input.busy = true;
                                     // No validation allowed on a quick pick. Have to put up a dialog instead
-                                    await this.shell.showErrorMessage(validationMessage, { modal: true });
+                                    await this.shell.showErrorMessage(validationMessage);
```
IanMatthewHuff, line 209: This fixes the new experimental flow. It was using a quick pick instead of an input box. There error message that it was showing was being forced to modal. But at this point the flow just ends regardless of modal or not. So there is no need to make this a modal dialog that can't show links. Just use a normal error message. This is the only multistep quick pick flow that uses validation, so we know it won't change any other flows.

amunger (approved):

PR #11391, opened 2022-09-16
Fix the typo

JoonHong-Kim:
duplicated "the" in the error message


amunger (approved):

joyceerhl (approved):

joyceerhl (commented):
Does the message also need to be updated here? https://github.com/microsoft/vscode-jupyter/blob/e3eee370c97505ce390858a4645f73bc1b3912fc/src/platform/common/utils/localize.ts#L335

amunger (changes requested):
this just changes the expected value in the test, it should be updated in the prod localize file as well https://insiders.vscode.dev/github/microsoft/vscode-jupyter/blob/aamunger/IWDecorations/src/platform/common/utils/localize.ts#L335

PR #11397, opened 2022-09-16
Fix for issue #11396

### LINKED ISSUE #11396: Cell debug doesn't stop at breakpoints when VSCode is run on Windows but a remote kernel runs on Unix
## Environment data

-   VS Code version: 1.71.2
-   Jupyter Extension version (available under the Extensions sidebar): v2022.8.1002431955
-   Python Extension version (available under the Extensions sidebar): v2022.14.0
-   OS (Windows | Mac | Linux distro) and version: Windows 10 Enterprise 19044.1889
-   Python and/or Anaconda version: Python 3.9.12, Anaconda 4.14.0
-   Jupyter server running: Remote

## Expected behaviour

After setting a breakpoint in a Jupyter Notebook cell and clicking on Debug cell, the debugger should stop at the breakpoint and show runtime info.

## Actual behaviour

Instead of stopping on the breakpoint, VSCode simply executes the cell. The behavior is due to a bug in the kernelDebugAdapter, see later.

## Steps to reproduce:

Start VSCode on a Windows machine and connect to a kernel that runs for example on Ubuntu (or some other unix system), then try debugging a cell.

## Logs

The issue was evident from the debugpy logs on the server. In a nutshell: even though the kernel was running on unix, the SetBreakpoint requests included paths of the form "\tmp\kernel_id\dumped_cell_content_id.py", and the responses were File Not Found exceptions - as it could be expected.

## Solution

The problem stems from path normalization in the dumpCell function of the kernelDebugAdapter. It normalizes the path based on the OS where VSCode runs, which doesn't necessarily coincide with the OS where the kernel runs. The normalization is unnecessary, as on the VSCode side the path on the kernel is simply a key in a hashmap - it works just as fine with whatever path the kernel returned, but in return when VSCode sends the path back to the kernel this way it's guaranteed that the path exists.

I'll be sending a pull-request with the solution when the change passes the CI.

adam-zlatniczki:
Fixes #11396

### DIFF
--- a/news/2 Fixes/11396.md
+++ b/news/2 Fixes/11396.md
@@ -1,0 +1,2 @@
+Fixed an interworking issue when running VSCode on Windows and connecting to a remote kernel that runs on some Unix distribution. Due to file path naming convention collisions, cell debugging failed.
+(thanks [Adam Zlatniczki](https://github.com/adam-zlatniczki))
--- a/src/notebooks/debugger/kernelDebugAdapter.ts
+++ b/src/notebooks/debugger/kernelDebugAdapter.ts
@@ -3,7 +3,6 @@
 
 'use strict';
 
-import * as path from '../../platform/vscode-path/path';
 import { IDumpCellResponse } from './debuggingTypes';
 import { traceError } from '../../platform/logging';
 import { KernelDebugAdapterBase } from './kernelDebugAdapterBase';
@@ -22,9 +21,9 @@ export class KernelDebugAdapter extends KernelDebugAdapterBase {
             const response = await this.session.customRequest('dumpCell', {
                 code: cell.document.getText().replace(/\r\n/g, '\n')
             });
-            const norm = path.normalize((response as IDumpCellResponse).sourcePath);
-            this.fileToCell.set(norm, cell.document.uri);
-            this.cellToFile.set(cell.document.uri.toString(), norm);
+            const pathOnKernelSide = (response as IDumpCellResponse).sourcePath;
+            this.fileToCell.set(pathOnKernelSide, cell.document.uri);
+            this.cellToFile.set(cell.document.uri.toString(), pathOnKernelSide);
         } catch (err) {
             traceError(err);
         }


roblourens (commented):
Thanks for the investigation and fix! I see a bunch of places around the debugger that use the path module, and they are all possibly suspicious. You haven't seen any other issues while debugging cross-OS? Also, are you sure there is no scenario where it's needed? I can't think of anything, I found this commit from the PR that originally added it, https://github.com/microsoft/vscode-jupyter/commit/e9bfc0df02603261a18e8d50bd90215c075f8522, but no explanation unfortunately

adam-zlatniczki:
> Thanks for the investigation and fix! I see a bunch of places around the debugger that use the path module, and they are all possibly suspicious. You haven't seen any other issues while debugging cross-OS? Also, are you sure there is no scenario where it's needed? I can't think of anything, I found this commit from the PR that originally added it, [e9bfc0d](https://github.com/microsoft/vscode-jupyter/commit/e9bfc0df02603261a18e8d50bd90215c075f8522), but no explanation unfortunately

Regarding the kernelDebugAdapter.ts for **notebooks**, I don't really see any risks: if you search for fileToCell and cellToFile, the only results are in the kernelDebugAdapter and kernelDebugAdapterBase, and they are only used as lookups. Manual testing also didn't reveal any issues for me.

Now that you mentioned it I got curious, so I played around a bit with different setups and debug scenarios. It looks like there's a problem with debugging inside an interactive window kernel too. Judging by a quick glance at the error log, it seems to be something with the path convention too, but I need to dig deeper into this. I'll be getting back to you as soon as I discover something.

roblourens:
I figured out why we have that, it's because dumpCell used to return mixed slashes, \ and / on windows, and those need to be normalized before we use them to set breakpoints. I filed this issue before realizing it was fixed this year: https://github.com/ipython/ipykernel/issues/995 I think we need to keep the code because there will be plenty of users with environments that have the bug.

Instead of calling the generic normalize method, we can do a simple check for whether the path looks like a windows path with a drive letter (matches `[a-z]:\\`) and call the correct normalize method.

If you see something weird going on with the interactive window too, this call is suspicious

https://github.com/microsoft/vscode-jupyter/blob/947e18f8b0252877362ea2491019b9fb2872e755/src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts#L93

adam-zlatniczki:
Ooh, I see, that ipython issue is something that I wouldn't have figured
out in a thousand years. I guess this renders my pull-request obsolote. If
you agree, I'll remove the PR and move this conversation to the comment
section of issue #11396.

In the meantime I did some preliminary investigation for the interactive
window debugger problem. I also started exactly where you pointed to, but
it didn't solve the problem entirely. After normalization is done properly,
there are no failed requests on the kernel side, but the UI still ignores
the breakpoints. Unfortunately I wasn't able to figure out why, but I saw
something interesting: the notebook uri looks like file:///c:/... Notice
that there's an extra /, making the uri translate to a file located at
/c:/..., which is clearly an invalid windows path. Any ideas on this?

Rob Lourens ***@***.***> ezt írta (időpont: 2022. szept. 21.,
Sze 5:13):

roblourens:
Continuing in the issue

PR #11413, opened 2022-09-20
Add ability to customize a server display name in the picker list

IanMatthewHuff:


src/notebooks/serviceRegistry.node.ts
```diff
@@ -28,7 +28,7 @@ import { Identifiers } from '../platform/common/constants';
 import { JupyterDebugService } from './debugger/jupyterDebugService.node';
 import { NotebookIPyWidgetCoordinator } from './controllers/notebookIPyWidgetCoordinator';
 import { RemoteKernelConnectionHandler } from './controllers/remoteKernelConnectionHandler';
-import { JupyterServerSelectorCommand } from './serverSelector';
```
IanMatthewHuff, line 31: I renamed this as I didn't like how we had two files called 'serverSelector.ts' this file has the command, so name it with just the command.

src/kernels/jupyter/types.ts
```diff
@@ -234,13 +234,21 @@ export interface IJupyterUriProviderRegistration {
...
 export const IJupyterServerUriStorage = Symbol('IJupyterServerUriStorage');
 export interface IJupyterServerUriStorage {
     readonly currentServerId: string | undefined;
     readonly onDidChangeUri: Event<void>;
-    readonly onDidRemoveUris: Event<string[]>;
+    readonly onDidRemoveUris: Event<IJupyterServerUriEntry[]>;
+    readonly onDidAddUri: Event<IJupyterServerUriEntry>;
     addToUriList(uri: string, time: number, displayName: string): Promise<void>;
-    getSavedUriList(): Promise<{ uri: string; serverId: string; time: number; displayName?: string }[]>;
+    getSavedUriList(): Promise<IJupyterServerUriEntry[]>;
```
IanMatthewHuff, line 251: I changed the public interfaces on the URI list to return this whole type (now given a specific type name) from get, onDidRemote, onDidAdd. Felt like it was a good readability improvement and makes it so we don't need to recalculate serverID in a place or two.

src/kernels/jupyter/serverSelector.ts
```diff
@@ -663,3 +676,288 @@ class JupyterServerSelector_Original implements IJupyterServerSelector {
         return items;
     }
 }
+
+/**
+ * Inisders version of the JupyterServerSelector.
+ */
+class JupyterServerSelector_Insiders implements IJupyterServerSelector {
```
IanMatthewHuff, line 683: This is the new insiders version of the server selector, it's basically the same as the Original version (not 'Experimental'). Don't love the juggling here, but to prototype + work with stable and the previous experiment this seems the best way. I'll call out the different areas in this PR, but it's 98% the same.

src/kernels/jupyter/serverSelector.ts
```diff
@@ -663,3 +676,288 @@ class JupyterServerSelector_Original implements IJupyterServerSelector {
...
+        const uri = await input.showInputBox({
+            title: DataScience.jupyterSelectURIPrompt(),
+            value: initialValue || defaultUri,
+            validate: this.validateSelectJupyterURI,
+            prompt: ''
+        });
+
+        // Offer the user a change to pick a display name for the server
+        // Prepopulate the value with the uri as the default suggestiong
+        const newDisplayName = await this.applicationShell.showInputBox({
```
IanMatthewHuff, line 852: This is the only big change from the previous server picker, new input box to select a display name. Pre-populated with the URL, so you can just hit enter again to get the previous UI.

src/kernels/jupyter/types.ts
```diff
@@ -234,13 +234,21 @@ export interface IJupyterUriProviderRegistration {
     getJupyterServerUri(id: string, handle: JupyterServerUriHandle): Promise<IJupyterServerUri>;
 }
 
+export interface IJupyterServerUriEntry {
+    uri: string;
+    serverId: string;
+    time: number;
```
DonJayamanne, line 240: Some comment here would be good, is this time this was added or last accessed?

DonJayamanne (approved):

src/kernels/jupyter/types.ts
IanMatthewHuff: Good call. I'll add that in and update.

src/kernels/jupyter/types.ts
```diff
@@ -234,11 +234,12 @@ export interface IJupyterUriProviderRegistration {
...
 
+// An entry into our list of saved servers
 export interface IJupyterServerUriEntry {
-    uri: string;
-    serverId: string;
-    time: number;
-    displayName?: string;
+    uri: string; // URI of the server to connect to
+    serverId: string; // Unique ID using a hash of the full uri
+    time: number; // The most recent time that we connected to this server
```
DonJayamanne, line 241: Can we add jsDoc style comments
```suggestion
    // The most recent time that we connected to this server
    time: number; 
```

```suggestion
    /** 
     * The most recent time that we connected to this server
     */
    time: number; 
```

These comments show up better in typescript. Else one would have to use `go to definition` to see the above comments.

IanMatthewHuff: Bleh, I agree in this case for the definition look up, so I'll change 'em. But I never like how much space the jsdoc style takes up vertically.

IanMatthewHuff: Update incoming

PR #11414, opened 2022-09-20
Skip processing imports and mime types if telemetry is disabled

DonJayamanne:


IanMatthewHuff:
@DonJayamanne Sorry had a question on this. Shouldn't this be happening at a lower level? I would assume disabling telemetry would disable it all at a lower point where we send telemetry.

DonJayamanne:
>  Shouldn't this be happening at a lower level? I would assume disabling telemetry would disable it all at a lower point where we send telemetry.

Yes thats correct, however this whole class performa a lot of processing which is completely unnecessary.
This class is written with the intention of capturing telemetry, however if telemetry is disabled, then all of the processing is pointless & unnecessarily chews up resources (CPU/memory)

IanMatthewHuff (approved):

PR #11418, opened 2022-09-20
Add code folding regions for # %% cells within python files

### LINKED ISSUE #1527: Make cells (# %%) collapsible python files
## Environment data

- VS Code version: 1.33
- Extension version (available under the Extensions sidebar): Current
- OS and version: Win10 1703
- Python version (& distribution if applicable, e.g. Anaconda): Anaconda 3.7
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): Conda
- Relevant/affected Python packages and their versions: n/a
## Expected behaviour

In a document with multiple cells, collapsing some of them improves usability.

## Actual behaviour

It is not possible to collapse cells.

## Some Details

I think defining `#%%` as (collapsible region start) shouldn't be very hard (I can try to submit a PR in 1-2 weekends if you agree with the idea). 

Some implementation details should be evaluated though:

- Behavior of Shift+Enter in the cell above
- What to display if a region is collapsed (`#%% Comment ...` would be the default I think)
- Behavior of collapsed regions affected by `Run below|above`
- Visibility of `Run Cell | Run Above | Run Below`

amunger:
fixes #1527

re-use the ranges that the codelense provider is already calculating

### DIFF
--- a/src/interactive-window/serviceRegistry.node.ts
+++ b/src/interactive-window/serviceRegistry.node.ts
@@ -21,8 +21,7 @@ import {
     ICodeLensFactory,
     IDataScienceCodeLensProvider,
     IGeneratedCodeStorageFactory,
-    ICodeGeneratorFactory,
-    IPythonCellFoldingProvider
+    ICodeGeneratorFactory
 } from './editor-integration/types';
 import { GeneratedCodeStorageManager } from './generatedCodeStoreManager';
 import { InteractiveWindowTracebackFormatter } from './outputs/tracebackFormatter';
@@ -43,7 +42,10 @@ export function registerTypes(serviceManager: IServiceManager) {
         IDataScienceCodeLensProvider,
         DataScienceCodeLensProvider
     );
-    serviceManager.addSingleton<IPythonCellFoldingProvider>(IPythonCellFoldingProvider, PythonCellFoldingProvider);
+    serviceManager.addSingleton<IExtensionSyncActivationService>(
+        IExtensionSyncActivationService,
+        PythonCellFoldingProvider
+    );
     serviceManager.addSingleton<IExtensionSingleActivationService>(IExtensionSingleActivationService, Decorator);
     serviceManager.addSingleton<IExtensionSyncActivationService>(
         IExtensionSyncActivationService,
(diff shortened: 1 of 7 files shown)


mjbvz (approved):

src/interactive-window/serviceRegistry.node.ts
```diff
@@ -41,6 +43,7 @@ export function registerTypes(serviceManager: IServiceManager) {
         IDataScienceCodeLensProvider,
         DataScienceCodeLensProvider
     );
+    serviceManager.addSingleton<IPythonCellFoldingProvider>(IPythonCellFoldingProvider, PythonCellFoldingProvider);
```
DonJayamanne, line 46: Won't you need the same for web as well.

src/standalone/activation/globalActivation.ts
```diff
@@ -61,6 +67,11 @@ export class GlobalActivation implements IExtensionSingleActivationService {
                 vscode.languages.registerCodeLensProvider([PYTHON_FILE_ANY_SCHEME], this.dataScienceCodeLensProvider)
             );
         }
+        if (this.pythonCellFoldingProvider) {
+            this.extensionContext.subscriptions.push(
```
DonJayamanne, line 71: Instead of modiying globalActivation you can do this in the `activate` method of the folding provider.

DonJayamanne: YOu might as well register these providers in the activate method of the `FoldingProvider` class itself.

Have a look at the `NotebookCellBangInstallDiagnosticsProvider` & `PythonKernelCompletionProviderRegistration`
Thats what we do in other places as well, else we unnecessarily inject these dependencies into the `globalActivation.ts` file or other places.

DonJayamanne (changes requested):
I'd change this to register the providers like in `PythonKernelCompletionProviderRegistration`

src/standalone/activation/globalActivation.ts
amunger: thanks for pointing that out - I was following the codeLensProvider, so that should probably also be moved: https://github.com/microsoft/vscode-jupyter/issues/11424

DonJayamanne (approved):

rgbmrc:
Is it possible to disable this? I am afraid it messes with preexisting `# region`  blocks.

amunger:
@rgbmrc - the pre-release version has a `jupyter.pythonCellFolding` setting, and it should be available in release next week

PR #11420, opened 2022-09-20
Exclude proposed.d.ts and eslintrc.js from eslint

roblourens:


DonJayamanne (approved):

roblourens:
@DonJayamanne do you understand the purpose of this check that failed?

`Files are being ignored that should be linted: .eslintrc.js` 

https://github.com/microsoft/vscode-jupyter/blob/515ad59e73cb734ae2066c286ef964c6efd60b31/.github/actions/check-ignore-list/index.js#L41-L61

I don't get it... seems that it just complains if you change anything on the ignore list

Come to think of it, I don't understand why there are so many files on the ignore list, including a bunch of test files that don't exist anymore.

DonJayamanne:
> ou understand the purpose of this check that failed?

Yes, this is very similar to how vscode implemented support for `strict` in the codebase.
I believe we added `eslint` and didn't want to go through and fix all of the files.
Instead the solution was:
* Day 1 All files are ignored from eslint
* Any files added or updated will be checked against eslint
* Any files not motified will be ignored from eslint

This way over time files will be linted against eslint.

We did this exact same thing with implmenting `strict` checks for the codebase. However it looks like we forgot to go through this list clean these as part of debt. 

>  including a bunch of test files that don't exist anymore.

Probably needs to be cleaned up in debt. The file list is pretty small now.

.eslintrc.js
```diff
@@ -142,7 +142,9 @@ module.exports = {
         'src/test/workspaceSymbols/main.unit.test.ts',
         'src/test/workspaceSymbols/generator.unit.test.ts',
         'src/test/datascience/extensionapi/exampleextension/ms-ai-tools-test/src/serverPicker.ts',
-        'vscode.d.ts'
+        'vscode.d.ts',
+        'vscode.proposed.*.d.ts',
+        '.eslintrc.js'
```
DonJayamanne, line 147: Since `.eslintrc.js` is edited by us, i'd not exclude this file from eslint.
Else we could have the same content but  different styles by different people, which is what a linter resolves.

roblourens: Thanks for the explanation. Our lint npm script only looks at ts file in src, so files like .eslintrc.js and gulpfile.js are not linted. But the eslint extension doesn't know this and shows this error 
image
 because they are not included in a tsconfig file. If we want to lint these files it seems we need a tsconfig for each one, or a different eslintrc for them that isn't ts-based, I think. We could continue to ignore the errors for now and clean it up later

DonJayamanne: Got it, lets just ignore this for now.

Tyriar:
This was done in https://github.com/microsoft/vscode-jupyter/pull/12107

PR #11490, opened 2022-09-29
Support filing issue for flaky test.

rebornix:
Since we already have all the data, let's automate the last step.

### DIFF
--- a/.vscode/notebooks/CITestResults.ipynb
+++ b/.vscode/notebooks/CITestResults.ipynb
@@ -14,7 +14,8 @@
    "outputs": [],
    "source": [
     "%pip install requests\n",
-    "%pip install pandas"
+    "%pip install pandas\n",
+    "%pip install datetime"
    ]
   },
   {
@@ -184,7 +185,7 @@
    "outputs": [],
    "source": [
     "import pandas as pd\n",
-    "from datetime import timedelta\n",
+    "from datetime import timedelta, datetime\n",
     "\n",
     "testResults = []\n",
     "for file in os.listdir(resultsDownloadLocation):\n",
@@ -232,6 +233,13 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Failure of a specific test"
+   ]
+  },
+  {
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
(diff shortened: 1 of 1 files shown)


.vscode/notebooks/CITestResults.ipynb
```diff
@@ -295,12 +334,12 @@
...
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.13"
+   "version": "3.9.6"
   },
   "orig_nbformat": 4,
   "vscode": {
    "interpreter": {
-    "hash": "05cadb55cbb23d0d156eeac1bfe39405b391b4fa73e06b1a6140e7007eb56649"
+    "hash": "33193fed03cec9a59993558aba41e5916503fab5abae601cbbc0cdfec3d419dd"
```
rebornix, line 342: This probably should be transient metadata.

DonJayamanne: no it can't, we use this to reselect the right interpreter/kernel
once ian makes ALL necessary changes to kernel auto selection, we can review removing this completely.
@IanMatthewHuff /cc

i think it's a hack to store this in metadata, only works for a specific machine, hence not transferable to other machines.
besides we can track last selected kernel other ways without storing metadata in the nb.

DonJayamanne (approved):

.vscode/notebooks/CITestResults.ipynb
```diff
@@ -274,14 +282,45 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import io\n",
+    "from urllib import request\n",
+    "\n",
+    "# post to create new github issue\n",
```
amunger, line 300: The rest of this notebook could be run without any side effects, but this would create a new issue each time right?
Kai ran all cells when he tried this notebook out. We probably want some confirmation to actually create the issue

DonJayamanne: Yeah I agree

amunger (changes requested):
we shouldn't automatically create issues just when the notebook is run

.vscode/notebooks/CITestResults.ipynb
rebornix: agree that it should go away completely (`vscode` section) otherwise there is always dirty change when the notebook is being worked on collaboratively.

.vscode/notebooks/CITestResults.ipynb
rebornix: @amunger added a widget for confirmation before creating issue so the notebook can be executed without creating issues unexpectedly.

amunger (approved):

PR #11494, opened 2022-09-29
include all `build and test` runs that run against main

amunger:
also includes `scheduled` and manually `dispatched` runs

### DIFF
--- a/pythonFiles/aggregateTestResults.py
+++ b/pythonFiles/aggregateTestResults.py
@@ -11,8 +11,8 @@ print("Using authtoken with prefix: " + authtoken[:4])
 # %%
 def getRuns(createdDate):
     runsResponse = requests.get(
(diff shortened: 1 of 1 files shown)


pythonFiles/aggregateTestResults.py
```diff
@@ -11,8 +11,8 @@
 # %%
 def getRuns(createdDate):
     runsResponse = requests.get(
-        "https://api.github.com/repos/microsoft/vscode-jupyter/actions/runs",
-        params={"event": "push", "created": createdDate},
+        "https://api.github.com/repos/microsoft/vscode-jupyter/actions/workflows/build-test.yml/runs?per_page=50",
```
IanMatthewHuff, line 14: Github api looks like there is a query parameter for branch. Can we use that?

amunger: yes, I thought that would refer to the wrong branch attribute, but it seems to only select the ones we want

IanMatthewHuff (approved):

PR #11500, opened 2022-09-30
Set jupyter functions for pylance as well when it sets those functions for python core ext and use pylance completion api to do dedup

heejaechang:
-   [x] ~Pull request represents a single change (i.e. not fixing disparate/unrelated things in a single PR).~
-   [x] ~Title summarizes what is changing.~
-   [x] ~Appropriate comments and documentation strings in the code.~
-   [x] ~Has telemetry for feature-requests.~

### DIFF
--- a/src/standalone/intellisense/notebookPythonPathService.node.ts
+++ b/src/standalone/intellisense/notebookPythonPathService.node.ts
@@ -16,6 +16,13 @@ import * as semver from 'semver';
 import { traceInfo, traceVerbose } from '../../platform/logging';
 import { IControllerSelection } from '../../notebooks/controllers/types';
 
+interface PylanceApi {
+    notebook?: {
+        registerJupyterPythonPathFunction(func: (uri: Uri) => Promise<string | undefined>): void;
+        registerGetNotebookUriForTextDocumentUriFunction(func: (textDocumentUri: Uri) => Uri | undefined): void;
+    };
+}
+
 /**
  * Manages use of the Python extension's registerJupyterPythonPathFunction API which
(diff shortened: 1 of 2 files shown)


src/standalone/intellisense/notebookPythonPathService.node.ts
```diff
@@ -44,6 +51,14 @@ export class NotebookPythonPathService implements IExtensionSingleActivationServ
             return;
         }
 
+        const pylanceApi = await this._getPylanceApi();
```
IanMatthewHuff, line 54: This looks good and pretty straightforward to me. One question, is the python extension going to remove the registerJupyterPythonPathFunction and registerGetNotebookUriForTextDocumentUriFunction functions? I think that I mentioned this over Teams, but from the Jupyter perspective I don't see an issue with multiple APIs hooking up here. Say someone has a new Pylance, but an older Python, they might both register here. We could say that if we register pylance then don't register python. But from a Jupyter perspective that type of limitation doesn't seem needed. We just provide the functions for whoever wants them.

IanMatthewHuff (approved):

src/standalone/intellisense/notebookPythonPathService.node.ts
heejaechang: for now, both of them will register and only one will be used depending on an option set. once we think everything is stable, it will be removed from python core extension (and any client related code).

IanMatthewHuff (changes requested):
There has been some more conversation on these changes. I'll look to set up a meeting, but holding this for now.

DonJayamanne:
Closing based on feedback from @heejaechang

PR #11507, opened 2022-10-02
Refactor to change async to sync method and ensure we correctly support widgets in split views 

### LINKED ISSUE #11446: IPyWidgets no longer works with split editors
This used to work, looks list this is broken.
Based on the code we can have only one `CommonMessageCoordinator` per Notebook now, and this is tied to a single WebView.
However when we have multiple editors, we can have multiple webvierws, one webview for each editor (notebook UI).

DonJayamanne:
Found some debt while reviewing some of the Widget code for IPyWidgets 8
* Change async method to sync (it was unnecessarily async and complicated the flow)
* Ensure we support widgets with multiple editors for the same notebook , i.e split views (code is meant to do that but it doesn't support it correctly - i.e. code is wrong), hence fixes https://github.com/microsoft/vscode-jupyter/issues/11446

### DIFF
--- a/src/notebooks/controllers/ipywidgets/message/commonMessageCoordinator.ts
+++ b/src/notebooks/controllers/ipywidgets/message/commonMessageCoordinator.ts
@@ -34,7 +34,6 @@ import { IPyWidgetMessageDispatcherFactory } from './ipyWidgetMessageDispatcherF
 import { IPyWidgetScriptSource } from '../scriptSourceProvider/ipyWidgetScriptSource';
 import { IIPyWidgetMessageDispatcher, IWidgetScriptSourceProviderFactory } from '../types';
 import { ConsoleForegroundColors } from '../../../../platform/logging/types';
-import { createDeferred } from '../../../../platform/common/utils/async';
 import { IWebviewCommunication } from '../../../../platform/webviews/types';
 import { swallowExceptions } from '../../../../platform/common/utils/decorators';
 
@@ -63,7 +62,7 @@ export class CommonMessageCoordinator {
     private disposables: IDisposableRegistry;
     private jupyterOutput: IOutputChannel;
     private readonly configService: IConfigurationService;
-    private webview: IWebviewCommunication | undefined;
+    private readonly attachedWebviews = new WeakSet<IWebviewCommunication>();
     private modulesForWhichWeHaveDisplayedWidgetErrorMessage = new Set<string>();
 
     public constructor(
@@ -75,6 +74,7 @@ export class CommonMessageCoordinator {
         this.appShell = this.serviceContainer.get<IApplicationShell>(IApplicationShell, IApplicationShell);
         this.commandManager = this.serviceContainer.get<ICommandManager>(ICommandManager);
         this.configService = this.serviceContainer.get<IConfigurationService>(IConfigurationService);
+        this.initialize();
     }
 
     public dispose() {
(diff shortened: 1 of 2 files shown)


src/notebooks/controllers/ipywidgets/message/commonMessageCoordinator.ts
```diff
@@ -83,63 +83,49 @@ export class CommonMessageCoordinator {
         this.ipyWidgetScriptSource?.dispose(); // NOSONAR
     }
 
-    public async attach(webview: IWebviewCommunication) {
```
DonJayamanne, line 86: This was async, but if you check the call path to this method, we don't await at all, hence unnecessarily async.
Also, the webview is assumed to be always one-one for notebooks, but we can have multiple and that's why we have an attach method else it could have been passed via CTOR..

Looks like the code was broken after some refactoring.

src/notebooks/controllers/ipywidgets/message/commonMessageCoordinator.ts
```diff
@@ -83,63 +83,49 @@ export class CommonMessageCoordinator {
...
-            // In case the webview loaded earlier and it already sent the IPyWidgetMessages.IPyWidgets_Ready message
-            // This way we don't make assumptions, we just query widgets and ask its its ready (avoids timing issues etc).
-            webview
-                .postMessage({ type: IPyWidgetMessages.IPyWidgets_IsReadyRequest, payload: undefined })
-                .then(noop, noop);
-
-            // Wait for the widgets ready message
-            await promise.promise;
+    public attach(webview: IWebviewCommunication) {
+        if (this.attachedWebviews.has(webview)) {
```
DonJayamanne, line 87: Its now sync and we can attach to multiple webviews.

src/notebooks/controllers/notebookIPyWidgetCoordinator.ts
```diff
@@ -159,30 +149,19 @@ export class NotebookIPyWidgetCoordinator implements IExtensionSyncActivationSer
...
         }
-        traceVerbose(`Intiailize notebook communications for editor ${getDisplayPath(editor.notebook.uri)}`);
+        traceVerbose(`Initialize notebook communications for editor ${getDisplayPath(editor.notebook.uri)}`);
         const comms = new NotebookCommunication(editor, controller);
         this.addNotebookDisposables(notebook, [comms]);
         this.notebookCommunications.set(editor, comms);
-        const { token } = new CancellationTokenSource();
-        this.resolveKernel(notebook, comms, token).catch(noop);
-    }
-    private resolveKernel(
```
DonJayamanne, line 169: This was unnecessarily async, and an unnecessary method as well. Just made all of this inline to make it easier to follow the code.

src/notebooks/controllers/notebookIPyWidgetCoordinator.ts
```diff
@@ -202,8 +181,7 @@ export class NotebookIPyWidgetCoordinator implements IExtensionSyncActivationSer
         disposeAllDisposables(this.notebookDisposables.get(notebook) || []);
         editors.forEach((editor) => this.notebookCommunications.get(editor)?.dispose());
 
-        const coordinator = this.messageCoordinators.get(notebook);
-        coordinator?.then((c) => c.dispose()).catch(noop);
+        this.messageCoordinators.get(notebook)?.dispose();
```
DonJayamanne, line 184: No more promises, simpler

alexdima (approved):

PR #11525, opened 2022-10-04
Fix: vscode-jupyter fails to redirect output when multiple cells are run concurrently

ppeetteerrs:
Fixes #

The current behaviour of `VSCodeNotebookController.handleExecution` calls `executeCell` asynchronously, which leads to some synchronization issues when redirecting output between cells. The example notebook below should be able to demonstrate the bug.

image

Before this pull request, VSCode would be unable to output the two "HI"s when:
1. We restart the kernel and `Run All` (one `handleExecution` call with multiple cells)
2. We press `Shift+Enter` quickly to run all the cells consecutively (multiple `handleExecution` call with one cell each)

In order to produce the expected behaviour (displaying the desired outputs) under situations 1 and 2, a lock is added to ensure that multiple calls to `executeCell` are processed sequentially.

### DIFF
--- a/src/notebooks/controllers/vscodeNotebookController.ts
+++ b/src/notebooks/controllers/vscodeNotebookController.ts
@@ -89,6 +89,7 @@ import { initializeInteractiveOrNotebookTelemetryBasedOnUserAction } from '../..
 import { NotebookCellLanguageService } from '../languages/cellLanguageService';
 import { IDataScienceErrorHandler } from '../../kernels/errors/types';
 import { IJupyterServerUriStorage } from '../../kernels/jupyter/types';
+import AwaitLock from 'await-lock';
 
 /**
  * Our implementation of the VSCode Notebook Controller. Called by VS code to execute cells in a notebook. Also displayed
@@ -108,6 +109,7 @@ export class VSCodeNotebookController implements Disposable, IVSCodeNotebookCont
     private readonly disposables: IDisposable[] = [];
     private notebookKernels = new WeakMap<NotebookDocument, IKernel>();
     public readonly controller: NotebookController;
+    private lock = new AwaitLock();
     /**
      * Used purely for testing purposes.
      */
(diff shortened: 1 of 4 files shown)


ppeetteerrs:

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -295,7 +297,11 @@ export class VSCodeNotebookController implements Disposable, IVSCodeNotebookCont
         }
         await initializeInteractiveOrNotebookTelemetryBasedOnUserAction(notebook.uri, this.connection);
         // Notebook is trusted. Continue to execute cells
-        await Promise.all(cells.map((cell) => this.executeCell(notebook, cell)));
+        await Promise.all(cells.map(async (cell) => {
+            await this.lock.acquireAsync();
```
DonJayamanne, line 301: Thanks for the PR, I'll take a look at this.
Though this fixes the issue, we don't end up running `executeCell` for the other cells, when we should.
E.g. we need to wait for other cells to change state, I don't think this PR will ensure other cells go into a pending state.
Once again thanks & I'll take a look into this.
Will also file an issue so we can track the bug there.

ppeetteerrs: Yup, I agree with that. I just wanted to flag the bug and demo a fix. This can be done within executeCell as well, just that I am not familiar enough with the source code to decide which is best.

DonJayamanne: FYI - Filed an issue here https://github.com/microsoft/vscode-jupyter/issues/11526
I'll try to submit changes to this (your) PR, thanks, I can confirm this is a bug.

DonJayamanne:
@ppeetteerrs I coudn't push changes to your PR, hence ended up creating another PR https://github.com/microsoft/vscode-jupyter/pull/11542
Please feel free to chagne the changes from that PR and update yours and that should be good.
The fix was to ensure we didn't dispose a particular object.

Let me know once you've merged those changes and I'll approve this PR.

DonJayamanne:
@ppeetteerrs 
Thanks for your PR, as I haven't heard back I'm closing this and going with my PR.
Once again thanks for identifying this issue and submitting a PR to fix it

ppeetteerrs:
No problem!

PR #11542, opened 2022-10-05
Do not dispose kernel message handler for a cell when cell completes execution

### LINKED ISSUE #11526: Cell execution does not seem to work when using async Python code
Originally filed by [ppeetteerrs](https://github.com/ppeetteerrs) here https://github.com/microsoft/vscode-jupyter/pull/11525
Create a Python nb with the following cells
### Cell1
```python
from asyncio import sleep, create_task, gather
from typing import Awaitable, List
from IPython.display import display
from IPython import get_ipython

def get_msg_id() -> str:
	return get_ipython().kernel.get_parent()["header"]['msg_id']

async def say_hi_to_after(x:str, tasks: Awaitable[None]):
	current_cell_msg_id = get_msg_id()
	await tasks
	get_ipython().kernel.get_parent()["header"]["msg_id"] = x
	display("HI")
	get_ipython().kernel.get_parent()["header"]["msg_id"] = current_cell_msg_id
```
### Cell2
```python
x = get_msg_id()
```
### Cell3
```python
y = say_hi_to_after(x, sleep(2))
create_task(y);
```
### Cell4
```python
z = say_hi_to_after(x, sleep(3))
create_task(z);
```

Steps:
1. We restart the kernel and `Run All` (one `handleExecution` call with multiple cells)
2. We press `Shift+Enter` quickly to run all the cells consecutively (multiple `handleExecution` call with one cell each)

In both cases we should see `HI` displayed twice as the output for the second cell.

DonJayamanne:
Fixes #11526

### DIFF
--- a/src/kernels/execution/cellExecution.ts
+++ b/src/kernels/execution/cellExecution.ts
@@ -210,7 +210,6 @@ export class CellExecution implements IDisposable {
     public dispose() {
         traceCellMessage(this.cell, 'Execution disposed');
         disposeAllDisposables(this.disposables);
-        this.cellExecutionHandler?.dispose();
     }
     private completedWithErrors(error: Partial<Error>) {
         traceWarning(`Cell completed with errors`, error);
--- a/news/2 Fixes/11526.md
+++ b/news/2 Fixes/11526.md
@@ -1,0 +1,1 @@
+Ensure kernel messages are handled by a cell even after cell execution completes.
--- a/package-lock.json
+++ b/package-lock.json
@@ -172,6 +172,7 @@
                 "@vscode/test-electron": "^2.1.5",
                 "@vscode/test-web": "^0.0.29",
                 "acorn": "^6.4.1",
+                "await-lock": "^2.2.2",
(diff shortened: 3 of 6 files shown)


src/kernels/execution/cellExecution.ts
```diff
@@ -210,7 +210,6 @@ export class CellExecution implements IDisposable {
     public dispose() {
         traceCellMessage(this.cell, 'Execution disposed');
         disposeAllDisposables(this.disposables);
-        this.cellExecutionHandler?.dispose();
```
IanMatthewHuff, line 213: Does someone need to dispose this then? If it's disposable I'd assume that someone would need to call it? Or is someone else already cleaning it up?

DonJayamanne: It is tracked & gets disposed where this is created in `cellExecutionMessageHandlerService.ts`

DonJayamanne: I.e. we don't need to dispose this here.

IanMatthewHuff (approved):

PR #11544, opened 2022-10-05
clean up interactive window creation

### LINKED ISSUE #10808: Interactive Window creation and restore
Follow up item of https://github.com/microsoft/vscode-jupyter/pull/10785#issuecomment-1182646829

Currently the interactive window is created in an asynchronous fashion (due to the [async kernel detection](https://github.com/microsoft/vscode-jupyter/blob/b8f5974a47b8e43e972aa876bacf5e302f17eedd/src/interactive-window/interactiveWindowProvider.ts#L201-L202)), there is a chance of race condition (`InteractiveWindowProvider._windows` still empty when multiple clicks happen on **Run Cell** code lens). Supporting hot exit for Interactive Window also makes the workflow more complex as the hot exit / window restore happens synchronous at extension activation, before we have all notebook controllers ready.

A proposal coming from discussions in https://github.com/microsoft/vscode-jupyter/pull/10785#issuecomment-1182646829 is 

* Create `InteractiveWindow` immediately with `owner` and `mode`, push it to `_windows` in both cases (create or restore)
* Have one explicit `start` method to get the `InteractiveWindow` prepared for cell execution, which includes
  * If it's new IW
    * Get `preferredController`
    * Call `interactive.open` to create the IW editor in VS Code
    * Start kernel
  * If it's a restored IW, we then have the `InteractiveTab`
    * Reveal the tab into view (currently we use `openNotebookDocument` API, in the future we would have Tab API to archive this)
    * Get `preferredController` and set it as the active controller (maybe we don't need this as the notebook controller provider already tells VS Code which controller to use)
    * Start kernel

amunger:
Part of #10808

- Remove the excessive thread safety, the deferred creation promise should be enough to ensure duplicate IWs are not created
- Stop subscribing to events each time a cell is added by making the restore method more idempotent

### DIFF
--- a/src/interactive-window/interactiveWindowProvider.ts
+++ b/src/interactive-window/interactiveWindowProvider.ts
@@ -30,7 +30,6 @@ import {
     Resource,
     WORKSPACE_MEMENTO
 } from '../platform/common/types';
-import { chainable } from '../platform/common/utils/decorators';
 import * as localize from '../platform/common/utils/localize';
 import { noop } from '../platform/common/utils/misc';
 import { IServiceContainer } from '../platform/ioc/types';
@@ -139,7 +138,6 @@ export class InteractiveWindowProvider
         this._updateWindowCache();
     }
 
-    @chainable()
     public async getOrCreate(resource: Resource, connection?: KernelConnectionMetadata): Promise<IInteractiveWindow> {
         if (!this.workspaceService.isTrusted) {
             // This should not happen, but if it does, then just throw an error.
@@ -160,8 +158,8 @@ export class InteractiveWindowProvider
         if (!result) {
             // No match. Create a new item.
             result = await this.create(resource, mode, connection);
-            // start the kernel
-            result.start();
+            // ensure events are wired up and kernel is started.
+            result.initialize();
         } else {
             const preferredController = connection
                 ? this.controllerRegistration.get(connection, InteractiveWindowView)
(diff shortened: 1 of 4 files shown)


src/interactive-window/interactiveWindowProvider.ts
```diff
@@ -186,17 +184,11 @@ export class InteractiveWindowProvider
...
 
-    // Note to future devs: this function must be synchronous. Do not await on anything before calling
-    // the interactive window ctor and adding the interactive window to the provider's list of known windows.
-    // Otherwise we risk a race condition where e.g. multiple run cell requests come in quick and we end up
-    // instantiating multiples.
     private async create(resource: Resource, mode: InteractiveWindowMode, connection?: KernelConnectionMetadata) {
+        // track when a creation is pending, so consumers can wait before checking for an existing one.
         const creationInProgress = createDeferred<void>();
-        // Ensure we don't end up calling this method multiple times when creating an IW for the same resource.
         this.pendingCreations.push(creationInProgress.promise);
```
rebornix, line 190: Reading `pendingCreation`, it gave me a feeling that we can have multiple creation in parallel. However we are awaiting for all pending creations to be complete before creating new ones, does it mean that `pendingCreation.length <= 1` always? If so we can probably change it to `pendingCreations: Promise<void> | null`

rebornix (approved):
Looks good to me in overall. Only nick pick of how we manage `pendingCreation`

src/interactive-window/interactiveWindowProvider.ts
amunger: yeah I noticed that as well, I'll change it to better express that only one creation is happening at a time

Issue #11550, opened 2022-10-06
Custom startup directory not working for notebooks


DonJayamanne:
Found while looking into issue  https://github.com/microsoft/vscode-jupyter/issues/11534

PR #11551, opened 2022-10-06
Fix to support custom startup directory for kernel

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/11550

### DIFF
--- a/src/kernels/kernelStartupCodeProvider.node.ts
+++ b/src/kernels/kernelStartupCodeProvider.node.ts
@@ -32,12 +32,7 @@ export class KernelStartupCodeProvider implements IStartupCodeProvider {
         ) {
             return [];
         }
-        if (
-            !(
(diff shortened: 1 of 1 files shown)


src/kernels/kernelStartupCodeProvider.node.ts
```diff
@@ -32,12 +32,7 @@ export class KernelStartupCodeProvider implements IStartupCodeProvider {
         ) {
             return [];
         }
-        if (
-            !(
-                isLocalConnection(kernel.kernelConnectionMetadata) &&
-                !this.configService.getSettings(undefined).forceIPyKernelDebugger
```
DonJayamanne, line 38: Looks like a condition was accidentally added while refactoring the code.
This belongs to startup code of debuggers, not regular code.

joaomoreno (approved):

Issue #11552, opened 2022-10-06
Too many ways to calculate and setup CWD of kernels


DonJayamanne:
We have at least 3 ways to calcularte and setup CWD of the kernel
* One is `IWorkspaceService.computeWorkingDirectory`
    We have code in a generic location, however this is very specific to Notebook CWD, hence doesn't belong in `IWorkspaceService`
* & the other in `KernelStartupCodeProvider.getCode`
* & finally the third in `calculateWorkingDirectory`

We also setup the CWD in a few places
* When starting the kernel (via `cwd` option in procs and jupyter servers)
* After starting the kernel (via startup code)

DonJayamanne:
As a result of this
If we have startup folder as `WorksapceFolder` and we open an ipynb file in a nested folder, 
then:
* The path to `PYTHONPATH` is calculated relative to the ipynb file, when it should be relative to the workspace folder
* Hence things start to fall over.

DonJayamanne:
See here for original issue https://github.com/microsoft/vscode-jupyter/issues/11534

DonJayamanne:
Problems idenified:
* We don't set CWD for remote kernels
* However if we have a remote kernel thats local host then we set CWD, i.e. we have special case for local host
* When starting kernels we compute a working directory and use that (even in the case of raw kernels)
* However after we start the kernel (local kernels), we change the CWD by running code against the kernel

I.e. we have 2 ways to setup CWD for raw kernels:
1. When we spawn the process we setup CWD (one version)
2. After we spawn the process we run code against the kernel to setup CWD (this only works for Python)

This must be fixed.


TODO:
* We should avoid having special cases for local host (complicates the code). Not sure why we'd want to do this, remote should be treated as completely remote, instead of treating it as local but using users using  Jupyter to run locally (maintenance is difficult_
     * Add telemetry to see how many users use such features (as it is we know very few use remote, hence this subset would be much smaller)
     * Not supporting this might be worth the benefit of code maintenance
*

DonJayamanne:
PRs merged and fixed

PR #11560, opened 2022-10-06
Simplify code to calculate working dir of kernel & add tests 

DonJayamanne:
Part of #11552

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -266,7 +266,7 @@ module.exports = {
                         message: 'Importing test modules from ./src/test into extension code is not allowed.'
                     },
                     {
-                        target: './src/**[!test]**/**/*[!.node].ts',
+                        target: './src/**[!test]**/**/*[!.node|.unit].ts',
                         from: './src/**/*.node.ts',
                         message: 'Importing node modules into non node files is not allowed.'
                     },
--- a/src/kernels/kernelStartupCodeProvider.node.ts
+++ b/src/kernels/kernelStartupCodeProvider.node.ts
@@ -2,80 +2,34 @@
 // Licensed under the MIT License.
 
 import { inject, injectable } from 'inversify';
-import { Uri } from 'vscode';
-import { IWorkspaceService } from '../platform/common/application/types';
 import { CodeSnippets } from '../platform/common/constants';
 import { getFilePath } from '../platform/common/platform/fs-paths';
-import { IFileSystem } from '../platform/common/platform/types';
-import { IConfigurationService } from '../platform/common/types';
-import { calculateWorkingDirectory } from '../platform/common/utils.node';
 import { traceInfo } from '../platform/logging';
 import { isLocalHostConnection, isPythonKernelConnection } from './helpers';
-import { expandWorkingDir } from './jupyter/jupyterUtils';
+import { KernelWorkingFolder } from './kernelWorkingFolder.node';
 import { IKernel, isLocalConnection, IStartupCodeProvider, StartupCodePriority } from './types';
 
 @injectable()
 export class KernelStartupCodeProvider implements IStartupCodeProvider {
     public priority = StartupCodePriority.Base;
 
-    constructor(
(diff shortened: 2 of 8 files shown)


src/platform/common/application/workspace.node.ts
```diff
@@ -6,32 +6,27 @@ import { BaseWorkspaceService } from './workspace.base';
...
-    return workspace.getWorkspaceFolder(resource)?.uri.fsPath || workspace.rootFolder?.fsPath || process.cwd();
-}
 
 /**
  * Node implementation of the workspace service. Computing working directory is different for node.
  */
 @injectable()
 export class WorkspaceService extends BaseWorkspaceService {
-    public computeWorkingDirectory(resource: Resource): Promise<string> {
-        return computeWorkingDirectory(resource, this);
```
DonJayamanne, line 35: Not sure why we were doing this, passing `this` into a function.
Inlined the function here for improved readability

src/platform/common/utils.node.ts
```diff
@@ -29,45 +23,3 @@ export async function tryGetRealPath(expectedPath: Uri): Promise<Uri> {
         return expectedPath;
     }
 }
-
-export async function calculateWorkingDirectory(
-    configService: IConfigurationService,
```
DonJayamanne, line 34: This code doesn't belong here, this is very specific to kernels hence should not be under `platnform/common `

src/kernels/kernelStartupCodeProvider.node.ts
```diff
@@ -2,80 +2,43 @@
...
-            isLocalConnection(kernel.kernelConnectionMetadata) ||
-            isLocalHostConnection(kernel.kernelConnectionMetadata)
-        ) {
-            let suggestedDir = await calculateWorkingDirectory(
-                this.configService,
-                this.workspaceService,
-                this.fs,
-                kernel.resourceUri
-            );
-            if (suggestedDir && (await this.fs.exists(suggestedDir))) {
```
DonJayamanne, line 49: As you can see we have more complex logic here and its partially split into two files.
With this PR its all in 1 place with tests.

.eslintrc.js
```diff
@@ -266,7 +266,7 @@ module.exports = {
                         message: 'Importing test modules from ./src/test into extension code is not allowed.'
                     },
                     {
-                        target: './src/**[!test]**/**/*[!.node].ts',
+                        target: './src/**[!test]**/**/*[!.node|.unit].ts',
```
DonJayamanne, line 269: Unit tests can import files from `*.node.ts` files for testing.

src/kernels/kernelStartupCodeProvider.node.ts
```diff
@@ -2,80 +2,43 @@
...
-                        this.workspaceService,
-                        this.configService.getSettings(kernel.resourceUri)
-                    )
-                );
-                if (suggestedDir && (await this.fs.exists(suggestedDir))) {
-                    traceInfo('UpdateWorkingDirectoryAndPath in Kernel');
-                    return this.getChangeDirectoryCode(kernel, suggestedDir);
-                }
-            }
+        // If this is a live kernel, we shouldn't be changing anything by running startup code.
```
IanMatthewHuff, line 34: Comment might be in the wrong place now?

IanMatthewHuff: Small suggestion here, these checks are doubled up both here and in getWorkingDirectory. Maybe could just be in one place? If get working directory just returns undefined for remote or non-python then these checks here don't seem needed as a function prereq.

IanMatthewHuff (approved):
Looks good. Much cleaner to read now. Had one possible suggestion (looks like some error cases are checked twice).

src/kernels/kernelStartupCodeProvider.node.ts
DonJayamanne: Agreed, Fixed

PR #11572, opened 2022-10-07
Universal Remote Kernel finder under new kernel picker experiment

IanMatthewHuff:
Tagging @rebornix on the review since we already talked over much of this already.

Now when using the Insiders kernel picker experiment all servers currently in the URI list will be checked for kernels and listed in the kernel list, regardless of current server setting. May be some quirks here still, but enough for the next step in the picker experiments.

image

image


src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
```diff
@@ -0,0 +1,393 @@
...
+        private readonly env: IApplicationEnvironment,
+        private readonly cachedRemoteKernelValidator: IJupyterRemoteCachedKernelValidator,
+        kernelFinder: KernelFinder,
+        private readonly kernelProvider: IKernelProvider,
+        private readonly extensions: IExtensions,
+        private isWebExtension: boolean,
+        private readonly serverUri: IJupyterServerUriEntry
+    ) {
+        // Register with remote-serverId as our ID
+        this.id = `${this.kind}-${serverUri.serverId}`;
```
IanMatthewHuff, line 91: Note: class is mostly the same as remoteKernelFinder. Changes to all using serverId of the given serverUri as the key for cache + for registering with the main kernel finder and disposing if removed.

src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
```diff
@@ -0,0 +1,393 @@
...
+                return await this.interpreterService.getInterpreterDetails(Uri.file(spec.argv[0]));
+            } catch (ex) {
+                traceError(`Failure getting interpreter details for remote kernel: `, ex);
+            }
+        }
+    }
+
+    private getCacheKey() {
+        // For Universal finders key each one per serverId
+        return `${RemoteKernelSpecsCacheKey}-${this.serverUri.serverId}`;
```
IanMatthewHuff, line 371: Cache keyed on serverId

src/kernels/jupyter/finder/universalRemoteKernelFinderController.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        // Check for when more URIs are added
+        this.serverUriStorage.onDidAddUri(this.createRemoteKernelFinder, this, this.disposables);
+
+        // Also check for when a URI is removed
+        this.serverUriStorage.onDidRemoveUris(this.urisRemoved, this, this.disposables);
+    }
+
+    createRemoteKernelFinder(serverUri: IJupyterServerUriEntry) {
+        if (!this.serverFinderMapping.has(serverUri.serverId)) {
```
IanMatthewHuff, line 68: UniversalRemoteKernelFinders get created, added to the mapping and activated when URIs are added the serverURI storage.

src/kernels/jupyter/serviceRegistry.node.ts
```diff
@@ -155,4 +153,16 @@ export function registerTypes(serviceManager: IServiceManager, _isDevMode: boole
     );
     serviceManager.addSingleton<JupyterDetectionTelemetry>(IExtensionSyncActivationService, JupyterDetectionTelemetry);
     serviceManager.addSingleton<IDataScienceErrorHandler>(IDataScienceErrorHandler, DataScienceErrorHandlerNode);
+    const configuration = serviceManager.get<IConfigurationService>(IConfigurationService);
+    if (configuration.getSettings().kernelPickerType === 'Insiders') {
```
IanMatthewHuff, line 157: Need to be in insiders kernel picker to get the new UniversalRemoteKernel finder instead of the old one.

src/kernels/kernelFinder.ts
```diff
@@ -22,9 +22,27 @@ export class KernelFinder implements IKernelFinder {
 
     constructor(@inject(IDisposableRegistry) private readonly disposables: IDisposableRegistry) {}
 
-    public registerKernelFinder(finder: IContributedKernelFinder) {
+    public registerKernelFinder(finder: IContributedKernelFinder): IDisposable {
         this._finders.push(finder);
-        this.disposables.push(finder.onDidChangeKernels(() => this._onDidChangeKernels.fire()));
+        const onDidChangeDisposable = finder.onDidChangeKernels(() => this._onDidChangeKernels.fire());
```
IanMatthewHuff, line 27: Some small changes to the main KernelFinder. Notify that kernels changed if a new finder is registered, and return a disposable on registration so that Finders can clean themselves out of the list as they dispose.

src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -195,6 +195,9 @@ export class ControllerRegistration implements IControllerRegistration {
 
         if (this.inKernelExperiment) {
             return userFiltered || connectionTypeFiltered || urlFiltered;
+        } else if (this.configuration.getSettings().kernelPickerType === 'Insiders') {
```
IanMatthewHuff, line 198: Minor change to controllerRegistration. Without this urls that are not the current url get filtered out.

src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
```diff
@@ -0,0 +1,393 @@
...
+// Re-query after 2 seconds to ensure we don't get stale information.
+const REMOTE_KERNEL_REFRESH_INTERVAL = 2_000;
+
+// This class watches a single jupyter server URI and returns kernels from it
+export class UniversalRemoteKernelFinder implements IRemoteKernelFinder, IContributedKernelFinderInfo, IDisposable {
+    /**
+     * List of ids of kernels that should be hidden from the kernel picker.
+     */
+    private readonly kernelIdsToHide = new Set<string>();
+    kind: string = 'remote';
```
rebornix, line 56: We might want to change this to `enum`.

src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
```diff
@@ -0,0 +1,393 @@
...
+
+    async activate(): Promise<void> {
+        // warm up the cache
+        this.loadCache().then(noop, noop);
+
+        // If we create a new kernel, we need to refresh if the kernel is remote (because
+        // we have live sessions possible)
+        // Note, this is a perf optimization for right now. We should not need
+        // to check for remote if the future when we support live sessions on local
+        this.kernelProvider.onDidStartKernel((k) => {
```
rebornix, line 119: Add this and listeners below to `this.disposables`?

rebornix (approved):
Looks good to me in overall, only one nit pick: adding listeners to local disposable store.

src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
IanMatthewHuff: @rebornix Ahh, good catch. So the next two listeners below are added to disposables, they are just added via using the third function parameter. However this onDidStartKernel was missing that. It was copied over from remoteKernelFinder.ts so I also fixed it up there as well.

src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
```diff
@@ -0,0 +1,397 @@
...
+                    sessionManager.getRunningKernels(),
+                    sessionManager.getKernelSpecs(),
+                    sessionManager.getRunningSessions(),
+                    computeServerId(connInfo.url)
+                ]);
+
+                // Turn them both into a combined list
+                const mappedSpecs = await Promise.all(
+                    specs.map(async (s) => {
+                        await sendKernelSpecTelemetry(s, 'remote');
```
DonJayamanne, line 296: i think we don't need the await here, it does the whole process of fetching the kernels

src/kernels/jupyter/finder/universalRemoteKernelFinderController.ts
```diff
@@ -0,0 +1,99 @@
...
+        @inject(IKernelFinder) private readonly kernelFinder: KernelFinder,
+        @inject(IDisposableRegistry) private readonly disposables: IDisposableRegistry,
+        @inject(IKernelProvider) private readonly kernelProvider: IKernelProvider,
+        @inject(IExtensions) private readonly extensions: IExtensions,
+        @inject(IsWebExtension) private isWebExtension: boolean
+    ) {}
+
+    async activate(): Promise<void> {
+        // Add in the URIs that we already know about
+        const currentServers = await this.serverUriStorage.getSavedUriList();
```
DonJayamanne, line 57: this does the loading of the extension, can this be done in the background, 
all of the activate methods think when extension loads and we wait for all of them to complete.
i would suggest using the IExtensionSync.... interface

src/kernels/jupyter/finder/remoteKernelFinder.ts
```diff
@@ -50,21 +50,22 @@ import { IApplicationEnvironment } from '../../../platform/common/application/ty
...
 @injectable()
 export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingleActivationService {
     /**
      * List of ids of kernels that should be hidden from the kernel picker.
      */
     private readonly kernelIdsToHide = new Set<string>();
-    kind: string = 'remote';
+    kind = ContributedKernelFinderKind.Remote;
+    id: string = 'currentremote';
+    displayName: string = 'Current Remote';
```
DonJayamanne, line 68: in local kernel finder this is localized, but not here

src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
```diff
@@ -0,0 +1,397 @@
...
+import { RemoteKernelSpecsCacheKey, removeOldCachedItems } from '../../common/commonFinder';
+import { ContributedKernelFinderKind, IContributedKernelFinderInfo } from '../../internalTypes';
+import { disposeAllDisposables } from '../../../platform/common/helpers';
+
+// Even after shutting down a kernel, the server API still returns the old information.
+// Re-query after 2 seconds to ensure we don't get stale information.
+const REMOTE_KERNEL_REFRESH_INTERVAL = 2_000;
+
+// This class watches a single jupyter server URI and returns kernels from it
+export class UniversalRemoteKernelFinder implements IRemoteKernelFinder, IContributedKernelFinderInfo, IDisposable {
```
DonJayamanne, line 51: curious why the name universal?
if the old class didn't exist would we still call this universal or just remotekernel finder?

src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
```diff
@@ -0,0 +1,397 @@
...
+     * List of ids of kernels that should be hidden from the kernel picker.
+     */
+    private readonly kernelIdsToHide = new Set<string>();
+    kind = ContributedKernelFinderKind.Remote;
+    id: string;
+    displayName: string;
+    private _cacheUpdateCancelTokenSource: CancellationTokenSource | undefined;
+    private cache: RemoteKernelConnectionMetadata[] = [];
+
+    private _onDidChangeKernels = new EventEmitter<void>();
```
DonJayamanne, line 62: this needs to be added to list of disposals, couldn't find it

src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
```diff
@@ -0,0 +1,397 @@
...
+        } else {
+            await promise;
+        }
+        return validValues;
+    }
+
+    // Talk to the remote server to determine sessions
+    @capturePerfTelemetry(Telemetry.KernelListingPerf, { kind: 'remote' })
+    public async listKernelsFromConnection(
+        connInfo: INotebookProviderConnection
```
DonJayamanne, line 276: i think we can pass the cancellation token in here, there's a lot of async called in here

IanMatthewHuff:
@DonJayamanne The goal here was to have (behind an experiment) just enough work to load kernels from all sources without massive changes to current kernel finding and controller loading mechanisms. It's not intended that this is the final shape of remote kernel loading and the bulk of it should only be active when the kernel picker experiment is set to 'Insiders.' The next checkins should add on the kernel picker commands for selecting sources which will give us the ability to consider which direction to take the UI.

If any of this seems risky to current stable bits just tell me and I can look into changing it.

src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
```diff
@@ -0,0 +1,397 @@
...
+                }
+            },
+            this,
+            this.disposables
+        );
+        this.wasPythonInstalledWhenFetchingKernels = this.extensionChecker.isPythonExtensionInstalled;
+    }
+
+    public async loadCache() {
+        traceInfoIfCI(`Remote Kernel Finder load cache Server: ${this.id}`);
```
DonJayamanne, line 160: this is an internal  implementation, why do we need to make this public
couldn't find it being used

DonJayamanne (approved):

src/kernels/jupyter/finder/remoteKernelFinder.ts
IanMatthewHuff: Added the string, but forgot to link it in here!

src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
IanMatthewHuff: Yeah, it was just a way to distinguish between the existing remote kernel finder and the prototype finder. Universal being universally considering all servers.

src/kernels/jupyter/finder/universalRemoteKernelFinder.ts
IanMatthewHuff: Ahh, looks like just a copy from the existing one. That that was used in unit tests. Cleaned up.

PR #11575, opened 2022-10-07
fix interactive window restore

### LINKED ISSUE #11574: Interactive window creation fails after a restored one for the same document was closed
repro steps
1. set "interactiveWindow.restore": true
2. run a cell from a .py file to create an interactive window
3. reload VScode
4. close the restored IW
5. run a cell from that same file
:bug: creation fails sometimes:

image

amunger:
Fixes #11574
part of #10808

- Most of the restoration is done just by recovering the editor, the controller is automatically selected.
- We need to make sure the events are all hooked up correctly even if no cells are run in the IW. (closing the IW should remove it from the list of windows)
- Get the editor information from the `windows` API instead of calling an API that might end up creating a new one. 
    - If the editor is closed before the extension does it's restore logic, we don't want to restore that editor

### DIFF
--- a/src/interactive-window/interactiveWindow.ts
+++ b/src/interactive-window/interactiveWindow.ts
@@ -437,6 +437,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
                         controller: e.controller.controller,
                         metadata: e.controller.connection
                     };
+                    // don't start the kernel if the IW has only been restored from a previous session
                     if (this.initialized) {
                         this.startKernel().ignoreErrors();
                     }
--- a/src/interactive-window/interactiveWindow.ts
+++ b/src/interactive-window/interactiveWindow.ts
@@ -49,13 +49,8 @@ import { INotebookExporter } from '../kernels/jupyter/types';
 import { IExportDialog, ExportFormat } from '../notebooks/export/types';
 import { generateCellsFromNotebookDocument } from './editor-integration/cellFactory';
 import { CellMatcher } from './editor-integration/cellMatcher';
-import {
-    IInteractiveWindowLoadable,
-    IInteractiveWindowDebugger,
-    IInteractiveWindowDebuggingManager,
-    InteractiveTab
-} from './types';
-import { generateInteractiveCode, isInteractiveInputTab } from './helpers';
+import { IInteractiveWindowLoadable, IInteractiveWindowDebugger, IInteractiveWindowDebuggingManager } from './types';
+import { generateInteractiveCode } from './helpers';
 import {
     IControllerRegistration,
     IControllerSelection,
@@ -100,7 +95,6 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
     public get kernelConnectionMetadata(): KernelConnectionMetadata | undefined {
         return this.currentKernelInfo.metadata;
     }
-    private initialized = false;
     private _onDidChangeViewState = new EventEmitter<void>();
     private closedEvent = new EventEmitter<void>();
     private _submitters: Uri[] = [];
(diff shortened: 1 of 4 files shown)


src/interactive-window/interactiveWindow.ts
```diff
@@ -461,7 +432,14 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
                 if (e.controller.id !== this.currentKernelInfo.controller?.id) {
                     this.disconnectKernel();
-                    this.startKernel(e.controller.controller, e.controller.connection).ignoreErrors();
+                    this.currentKernelInfo = {
+                        controller: e.controller.controller,
+                        metadata: e.controller.connection
+                    };
+                    // don't start the kernel if the IW has only been restored from a previous session
+                    if (this.initialized) {
+                        this.startKernel().ignoreErrors();
```
amunger, line 441: this event is what sets the controller on restore. It is called even when the controller is initially set, and will also take care of the case that the user changes the controller before running any cells

src/interactive-window/interactiveWindowProvider.ts
```diff
@@ -124,15 +124,28 @@ export class InteractiveWindowProvider
                 return;
             }
 
+            const tab = interactiveWindowMapping.get(iw.uriString);
+            const notebookEditor = window.visibleNotebookEditors.find(
```
rebornix, line 128: `window.visibleNotebookEditors` is probably not safe when we have multiple IW windows.

rebornix (commented):
I'm afraid that checking `visibleNotebookEditors` might not work with hidden IW tabs.

src/interactive-window/interactiveWindow.ts
```diff
@@ -95,7 +95,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
         return this._submitters;
     }
     public get notebookDocument(): NotebookDocument {
-        return this.notebookEditor.notebook;
+        return this.notebookEditor?.notebook;
```
rebornix, line 98: Would this change the typing to `notebookDocument: NotebookDocument | undefined`?

src/interactive-window/interactiveWindow.ts
```diff
@@ -168,72 +164,78 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
...
-    public async restore(preferredController: IVSCodeNotebookController | undefined) {
-        if (preferredController) {
-            this.preferredController = preferredController;
-        }
-        if (!this.notebookEditor) {
-            if (isInteractiveInputTab(this.notebookEditorOrTab)) {
-                const document = await workspace.openNotebookDocument(this.notebookEditorOrTab.input.uri);
-                const editor = await window.showNotebookDocument(document, {
-                    viewColumn: this.notebookEditorOrTab.group.viewColumn
+    public async ensureInitialized() {
```
rebornix, line 210: We can probably use some typescript type checking trick here since `notebookDocument` can be `undefined`:

```ts
interface IResolvedInteractiveWindow {
    notebookDocument: NotebookDocument;
}

class InteractiveWindow {
    ensureInitialized(): this is IResolvedInteractiveWindow;
}
```

rebornix (approved):
It looks good to me. Only nit about typings check.

src/interactive-window/interactiveWindow.ts
amunger: It actually was already that, but I suppose I should go through with the rest of the corrections for that

amunger: I think I'd rather tackle this in debt week - there is a lot of code in `InteractiveWindow` that relies on notebookDocument not being null, which would add a lot of extra `undefined` checking. Now that I've poked around quite a bit this iteration, I have a better idea of what could be done, including not relying on the editor to be the same throughout the lifetime of the IW object (separate issue).

Issue #7583, opened 2021-09-17
Adopt new Python extensions non-blocking API for interpreter disocvery


karrtikr:
Three discovery APIs are being added for Jupyter which is meant to replace existing `getInterpreters()` API: https://github.com/microsoft/vscode-python/pull/17452

```typescript
type PythonApiForJupyterExtension = {
     ...
    /**
     * IInterpreterService
     */
    readonly refreshPromise: Promise<void> | undefined;
    /**
     * IInterpreterService
     */
    readonly onDidChangeInterpreters: Event<PythonEnvironmentsChangedEvent>;
    /**
     * Equivalent to getInterpreters() in IInterpreterService
     */
    getKnownInterpreters(resource?: Uri): PythonEnvironment[];
    /**
     * Equivalent to getAllInterpreters() in IInterpreterService
     */
    getInterpreters(resource?: Uri): Promise<PythonEnvironment[]>;
    ...
```

### Previous behavior:

Existing `getInterpreters()` is an asynchronus blocking API, so it waits until all of discovery is completed the first time it is called. Subsequent calls return instantaneously from cache if the same resource is queried again.

### New behavior:

- New `getKnownInterpreters()` simply returns whatever is currently in cache. It is useful when we do not need all of the interpreters at once, which is usually the case. In the Python extension we are able to replace all instances of `getInterpreters()` with `getKnownInterpreters()`. Do note that the kind info returned by this API may change as refresh progresses.
- `onDidChangeInterpreters` fires when an environment gets added/removed/updated in cache. Event properties tracks the old and new environments accordingly.
- Python extension triggers a refresh in background when the extension loads for the first time, which is when `refreshPromise` is set. It can be used to implement now deprecated `getInterpreters()` in the following way:
```typescript
    /** Gets all interpreters **/
    public async getInterpreters(resource?: Uri): Promise<PythonEnvironment[]> {
        await this.refreshPromise; // Wait for the refresh to finish so that the cache contains all environments.
        return getKnownInterpreters(resource); // Return from cache.
    }
```
- `getInterpreterDetails()` is still available as before, and can be used to get complete and reliable information on environments. As it is a blocking call, it's recommend to only use it if complete information is needed, which is generally only the case for selected interpreters.

You can the see the new API in action here in our dynamic quickpick https://github.com/microsoft/vscode-python/pull/17043#issue-975156935.

> Note: Any internal interpreter list caching that was recently added should go away once `onDidChangeInterpreters` events are available.

[...]

PR #11591, opened 2022-10-10
Add Id property to PythonEnvironment type

DonJayamanne:
Part of #7583
* Add the `id` property to PythonEnvironment

### DIFF
--- a/src/test/client/api.vscode.test.ts
+++ b/src/test/client/api.vscode.test.ts
@@ -30,7 +30,7 @@ suite('3rd Party Kernel Service API', function () {
     const disposables: IDisposable[] = [];
     this.timeout(120_000);
     suiteSetup(async function () {
-        traceInfo('Suite Setup VS Code Notebook - Execution');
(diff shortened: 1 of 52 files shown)


src/test/client/api.vscode.test.ts
```diff
@@ -24,13 +24,13 @@ import { executeSilently } from '../../kernels/helpers';
 import { getPlainTextOrStreamOutput } from '../../kernels/kernel';
 import { IInterpreterService } from '../../platform/interpreter/contracts';
 
-suite('3rd Party Kernel Service API', function () {
+suite.only('3rd Party Kernel Service API', function () {
```
IanMatthewHuff, line 27: Got a .only in here. The PR into non main makes it a bit hard to catch these since the linting step doesn't run.

IanMatthewHuff (changes requested):
Changes look fine, but have to mark as changes for now since I think there is a test.only lint failure in here. Don't want that to leak into main. Will approve on that change.

IanMatthewHuff (approved):

PR #11593, opened 2022-10-10
Remove more unused Python execution methods

### LINKED ISSUE #7583: Adopt new Python extensions non-blocking API for interpreter disocvery

DonJayamanne:
Fixes #7583
* Remove unused Python execution methods.

### DIFF
--- a/src/kernels/raw/finder/localKernelFinder.node.ts
+++ b/src/kernels/raw/finder/localKernelFinder.node.ts
@@ -79,10 +79,7 @@ export class LocalKernelFinder implements ILocalKernelFinder, IExtensionSingleAc
 
         this.interpreters.onDidChangeInterpreters(
             async () => {
-                // Don't do anything if the interpreter list is still being refreshed
-                if (!this.interpreters.refreshing) {
-                    this.updateCache(undefined).then(noop, noop);
-                }
+                this.updateCache(undefined).then(noop, noop);
             },
             this,
             this.disposables
--- a/src/platform/api/pythonApi.ts
+++ b/src/platform/api/pythonApi.ts
@@ -4,13 +4,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
 import { EventEmitter, Event, Uri, ExtensionMode } from 'vscode';
-import {
-    IPythonApiProvider,
-    IPythonExtensionChecker,
-    IPythonProposedApi,
-    PythonApi,
-    PythonEnvironment_PythonApi
-} from './types';
+import { IPythonApiProvider, IPythonExtensionChecker, PythonApi, PythonEnvironment_PythonApi } from './types';
 import * as localize from '../common/utils/localize';
 import { injectable, inject } from 'inversify';
 import { sendTelemetryEvent } from '../../telemetry';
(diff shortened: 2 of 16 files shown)


IanMatthewHuff:
@DonJayamanne Not entirely sure how the ordering is supposed to work on these PRs, but approving the removals on this one. Do they need to be looked at in order?

IanMatthewHuff (approved):

DonJayamanne:
> Not entirely sure how the ordering is supposed to work on these PRs, but approving the removals on this one. Do they need to be looked at in order?

No need to look at them in any order,

PR #11605, opened 2022-10-10
Remove Uri arg from python API

DonJayamanne:
Part of #7583


joyceerhl (approved):
There are a few references to the `Uri` arg in some code that's been commented out (including the notebook functional tests), maybe worth updating if we ever bring those tests back?

DonJayamanne:
> arg in some code that's been commented out (including the notebook functional tests), maybe worth updating if we ever bring those tests back?

Thanks, the compiler should pick that up.

PR #11625, opened 2022-10-12
Lazy validate remote servers

rebornix:
Validate remote servers when the providers are registered.

* If the server is `local` or user provided, then it's validated immediately (no real validation will happen)
* If the server is provided by a third party provider, it will be validated by the extension. If the provider is not registered yet, it stays `invalidated` until the provider is registered.


src/kernels/jupyter/types.ts
```diff
@@ -254,6 +254,10 @@ export interface IJupyterServerUriEntry {
      * An optional display name to show for this server as opposed to just the Uri
      */
     displayName?: string;
+    /**
+     * Whether the server is validated by its provider or not
+     */
+    isValidated?: boolean;
```
DonJayamanne, line 260: When would this not be validated? I don't quite understand the flow and the purpose of this.

DonJayamanne (approved):

src/kernels/jupyter/types.ts
rebornix: Discussed offline about why we want to do this.

Issue #11576, opened 2022-10-07
Ship prototype of per document kernel source selection under an experimental flag


IanMatthewHuff:
Per notebook document, provide kernel picker commands that allow you to select the source of kernels available for this document. Sources can be kernels on the local system or any of the jupyter servers currently in the list of previous / current jupyter server connections.

https://user-images.githubusercontent.com/812783/194664102-737d5fe9-dd48-45c8-8e0e-1ec1bf682cc0.mov

PR #11632, opened 2022-10-12
Select Kernel Source Per Document Experiment

IanMatthewHuff:
#11576 

This PR completes shipping the basic prototype for selecting a kernel source per document. There are quite a few known hitches here still so I'll list a few.

1. Right now it's a two step process, first you pick a kernel source with the command, then you open the picker again to pick it. My first follow up will be looking at adding a kernel selection step to the second step in picking a kernel source so that you only have to open the picker once.
2. Suggestions are currently not wired up, I don't consider this a huge loss as I'd like to handle them differently in general, but when you select a kernel source everything is now just in a flat list.
3. Kernel source selections are not persisted, when documents open they start in the fresh "no source selected" state
4. Don't add the url as the default value for the server name, makes it hard to pick the name out
5. Invalidated server entries in the list break the loading of the full set of kernel finders possibly resolved by: https://github.com/microsoft/vscode-jupyter/pull/11625/files
6. Current selected kernel is not deselected if the notebook controller is hidden for that document


src/kernels/kernelFinder.ts
```diff
@@ -61,7 +61,11 @@ export class KernelFinder implements IKernelFinder {
         const kernels: KernelConnectionMetadata[] = [];
 
         for (const finder of this._finders) {
-            const contributedKernels = finder.listContributedKernels(resource);
+            const contributedKernels = finder.listContributedKernels(resource).map((kernelConnection) => {
```
IanMatthewHuff, line 64: @DonJayamanne Most of this change is all experimental code at the UI level under controllers. However there were a couple small changes at the KernelFinder. I want to make sure that I have your ok with how I'm handling this. 

We can chat online if needed.

This spot here is where I tag the kernel connections with the kernelFinderInfo that contributed them. I added this kernelFinderInfo as an optional thing on the type.

If desired I could check the configuration here to see if the experiment is active to see if this should be done or not.

src/kernels/kernelFinder.ts
```diff
@@ -73,4 +77,9 @@ export class KernelFinder implements IKernelFinder {
 
         return kernels;
     }
+
+    // Give the info for what kernel finders are currently registered
```
IanMatthewHuff, line 81: Also added this so that the UI could see what kernel finders are registered. I didn't just go off the controllers that we see registered since if you add a kernel source with no controllers you would still want to see it as an option in the UI so you know that you connected to the server correctly.

src/notebooks/controllers/commands/pickDocumentKernelSourceCommand.ts
```diff
@@ -0,0 +1,59 @@
...
+    private async pickDocumentKernelSource(_notebook?: NotebookDocument) {
+        // We want to get the context here from the command, but that needs a fix on the core side: https://github.com/microsoft/vscode/issues/161445
+        if (window.activeNotebookEditor && this.configService.getSettings().kernelPickerType === 'Insiders') {
+            await this.kernelSourceSelector.selectKernelSource(window.activeNotebookEditor.notebook);
+        }
+    }
+
+    // Only show this command if we are in our Insiders picker type
+    private updateVisibility() {
+        if (this.configService.getSettings().kernelPickerType === 'Insiders') {
```
rebornix, line 53: It seems we are not picking up the context key change in the kernel picker so after changing the setting, a window reload is required for it to work. We might need a fix in the core.

rebornix:
The code looks good to me. Played around with it a bit and ran into this interesting scenario:

* Select kernel source local
* Pick a kernel
* Run cells
* Select another kernel source
* Now the kernel picker will not show the local kernels anymore, but the kernel status bar and the notebook editor is still using the previous used local kernel, thus it should be `hidden` now.

image

src/kernels/kernelFinder.ts
```diff
@@ -61,7 +61,11 @@ export class KernelFinder implements IKernelFinder {
         const kernels: KernelConnectionMetadata[] = [];
 
         for (const finder of this._finders) {
-            const contributedKernels = finder.listContributedKernels(resource);
+            const contributedKernels = finder.listContributedKernels(resource).map((kernelConnection) => {
+                // Tag our connections with the kernel finder info that located them
+                kernelConnection.kernelFinderInfo = finder;
```
DonJayamanne, line 66: I think the `finder` should update the `kernelFinderInfo` as thats the class that creates and returns the `KernelConnection`.
This way no matter who creats these we know it will always be setup.

src/kernels/types.ts
```diff
@@ -94,7 +95,12 @@ export type PythonKernelConnectionMetadata = Readonly<{
  * This ensure we don't update is somewhere unnecessarily (such updates would be unexpected).
  * Unexpected as connections are defined once & not changed, if we need to change then user needs to create a new connection.
  */
-export type KernelConnectionMetadata = RemoteKernelConnectionMetadata | LocalKernelConnectionMetadata;
+export type KernelConnectionMetadata = (RemoteKernelConnectionMetadata | LocalKernelConnectionMetadata) & {
+    /**
+     * Kernel Connections may optionally be tagged with the info for the kernel finder that contributed them
+     */
+    kernelFinderInfo?: IContributedKernelFinderInfo;
```
DonJayamanne, line 102: Why not make them mandatory? Based on the code this looks like they will always be present (execpt for the cases where we load these from cache, and we can still update them from within the finders)

src/kernels/types.ts
```diff
@@ -612,6 +618,10 @@ export const IKernelFinder = Symbol('IKernelFinder');
 export interface IKernelFinder {
     onDidChangeKernels: Event<void>;
     listKernels(resource: Resource, cancelToken?: CancellationToken): Promise<KernelConnectionMetadata[]>;
+    /*
+     * Return basic info on all currently registered kernel finders
+     */
+    getRegisteredKernelFinderInfos(): IContributedKernelFinderInfo[];
```
DonJayamanne, line 624: Can we make this a property?

DonJayamanne (approved):
LGTM, however please move the code to update `kernelFinderInfo` into the finder itself, as the finder owns creation of the `KernelConnectionMetadata`, this way we know that the `KernelConnectionMeatadata` is created and populated in one place as opposed to multiple different locations.

src/kernels/types.ts
IanMatthewHuff: I'll take a second look here. Initially the reason that I did this was because we manually register controllers from outside the main finder path at a few places where they don't come from the finders. I don't love that we do this as it complicates things, so maybe I can get rid of these cases or manually put in some type of fake KernelFinderInfo. If those paths seem like they would mess with the stable path functionality I'll leave this as optional.

src/kernels/kernelFinder.ts
IanMatthewHuff: Sounds good. I'll make the change.

src/kernels/types.ts
DonJayamanne: i still think it could cause issues
if you check, these types are all read-only.
we want to prevent updating properties during runtime.
optionally you could have another class/dictionary to map the info by the kernel connection id

ie i don't think so should add these write properties, we had a lot of issues in the party as we didn't know when it could change

DonJayamanne (changes requested):
sorry but the connections should be ready only

src/kernels/types.ts
DonJayamanne: If you are able to make this a readonly property can you add this to the individual types instead of adding this to the top level type.
I.e. add the property to `LocalKernelSpecConnectionMetadata` itself (add to the 4 different types).

After all, if you look at an individual connection that should contain the `kernelFinderInfo`, but based on the above types it will not, as the proeprty is added to the top level union.

DonJayamanne (approved):

PR #11635, opened 2022-10-13
Lazy load Controllers

DonJayamanne:
Part of #7583

### DIFF
--- a/src/kernels/internalTypes.ts
+++ b/src/kernels/internalTypes.ts
@@ -13,7 +13,6 @@ export enum ContributedKernelFinderKind {
 
 export interface IContributedKernelFinder<T extends KernelConnectionMetadata> extends IContributedKernelFinderInfo {
     kind: ContributedKernelFinderKind;
-    initialized: Promise<void>;
     onDidChangeKernels: Event<void>;
     kernels: T[];
 }
--- a/src/kernels/jupyter/finder/remoteKernelFinder.ts
+++ b/src/kernels/jupyter/finder/remoteKernelFinder.ts
@@ -71,13 +71,6 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingle
     private _onDidChangeKernels = new EventEmitter<void>();
     onDidChangeKernels: Event<void> = this._onDidChangeKernels.event;
 
-    private _initializeResolve: () => void;
-    private _initializedPromise: Promise<void>;
-
-    get initialized(): Promise<void> {
-        return this._initializedPromise;
-    }
-
     private wasPythonInstalledWhenFetchingKernels = false;
 
     constructor(
@@ -97,10 +90,6 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingle
         @inject(IExtensions) private readonly extensions: IExtensions,
         @inject(IsWebExtension) private isWebExtension: boolean
     ) {
-        this._initializedPromise = new Promise<void>((resolve) => {
-            this._initializeResolve = resolve;
-        });
-
         kernelFinder.registerKernelFinder(this);
     }
 
@@ -163,7 +152,6 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingle
 
         if (this.serverConnectionType.isLocalLaunch) {
             await this.writeToCache([]);
-            this._initializeResolve();
             return;
         }
 
@@ -190,7 +178,6 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingle
         await this.writeToCache(kernels);
 
         traceVerbose('RemoteKernelFinder: load cache finished');
-        this._initializeResolve();
     }
 
     private async updateCache() {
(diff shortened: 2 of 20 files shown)


src/kernels/internalTypes.ts
```diff
@@ -13,7 +13,6 @@ export enum ContributedKernelFinderKind {
 
 export interface IContributedKernelFinder<T extends KernelConnectionMetadata> extends IContributedKernelFinderInfo {
     kind: ContributedKernelFinderKind;
-    initialized: Promise<void>;
```
DonJayamanne, line 16: Not required

src/kernels/jupyter/finder/remoteKernelFinder.ts
```diff
@@ -71,13 +71,6 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingle
     private _onDidChangeKernels = new EventEmitter<void>();
     onDidChangeKernels: Event<void> = this._onDidChangeKernels.event;
 
-    private _initializeResolve: () => void;
-    private _initializedPromise: Promise<void>;
-
-    get initialized(): Promise<void> {
-        return this._initializedPromise;
```
DonJayamanne, line 78: Now that we're push based (using events) this isn't required.

src/kernels/jupyter/types.ts
```diff
@@ -360,8 +359,4 @@ export interface IServerConnectionType {
     onDidChange: Event<void>;
 }
 
-export interface IRemoteKernelFinder extends IContributedKernelFinder<RemoteKernelConnectionMetadata> {
-    listKernelsFromConnection(
```
DonJayamanne, line 364: This method was only used in tests, hence removed from the interface, not to be used in the extension code.
Was confusing why we had this when we're implemeneting an existing interface

src/kernels/types.ts
```diff
@@ -611,7 +611,7 @@ export const IKernelFinder = Symbol('IKernelFinder');
 
 export interface IKernelFinder {
     onDidChangeKernels: Event<void>;
-    listKernels(cancelToken?: CancellationToken): Promise<KernelConnectionMetadata[]>;
+    kernels: KernelConnectionMetadata[];
```
DonJayamanne, line 614: Push based API

src/notebooks/controllers/controllerLoader.ts
```diff
@@ -70,12 +69,7 @@ export class ControllerLoader implements IControllerLoader, IExtensionSyncActiva
     public loadControllers(refresh?: boolean | undefined): Promise<void> {
         if (!this.controllersPromise || refresh) {
             const stopWatch = new StopWatch();
-            // Cancel previous load
-            if (this.loadCancellationToken) {
-                this.loadCancellationToken.cancel();
-            }
-            this.loadCancellationToken = new vscode.CancellationTokenSource();
-            this.controllersPromise = this.loadControllersImpl(this.loadCancellationToken.token)
+            this.controllersPromise = this.loadControllersImpl()
```
DonJayamanne, line 72: Cancellation token is no longer useful as the API is not pull based anymore, hence nothing can be cancelled.
I.e. nothing is async

src/notebooks/controllers/controllerLoader.ts
```diff
@@ -125,16 +119,11 @@ export class ControllerLoader implements IControllerLoader, IExtensionSyncActiva
         }
     }
 
-    private async loadControllersImpl(cancelToken: vscode.CancellationToken) {
+    private async loadControllersImpl() {
         // First off set our context to false as we are about to refresh
```
DonJayamanne, line 123: To be discussed with @IanMatthewHuff 
THe only thing that makes this async is updateing a context variable, which doesn't seem to be valid anymore.
I think we can update the context variable without relying on controlers,
instead we can directly use Python API to check if tehre are any interpreters and update the context var accordingly, and that can be done in a seprate class, making this class cleaner

src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -249,12 +249,15 @@ export class ControllerPreferredService implements IControllerPreferredService,
         let preferredConnection: KernelConnectionMetadata | undefined;
         const rankedConnections = await this.kernelRankHelper.rankKernels(
             uri,
+            this.registration.all,
```
DonJayamanne, line 252: I don't think rankKernels should internall call the finder API, 
its not possible to rank kernels when we don't already have kernels, hence i don't think rank kernels should call any api to get the kernels, instead it should be given a list of kernsl to rank.
Else the method should be named `getRankedKernels`, and that complicates things.
I think this is simpler now (simpler to adopt the lazy loading of controllers)

src/notebooks/controllers/kernelRanking/kernelRankingHelper.ts
```diff
@@ -19,36 +19,38 @@ import { IKernelRankingHelper } from '../types';
 @injectable()
 export class KernelRankingHelper implements IKernelRankingHelper {
     constructor(
-        @inject(IKernelFinder) private readonly kernelFinder: IKernelFinder,
```
DonJayamanne, line 22: See comments earlier

IanMatthewHuff (approved):

PR #11651, opened 2022-10-14
Re compute preferred controllers when controllers are added

### LINKED ISSUE #11646: Re-compute preferred controller when controllers are added dynamically
As a result of the changes in #7583 we need to re-compute the preferred controller everytime controllers are added

DonJayamanne:
Fixes #11646

### DIFF
--- a/src/notebooks/controllers/controllerPreferredService.ts
+++ b/src/notebooks/controllers/controllerPreferredService.ts
@@ -87,12 +87,10 @@ export class ControllerPreferredService implements IControllerPreferredService, 
             }, this)
         );
         this.disposables.add(
-            this.registration.onCreated(() => {
-                if (this.notebook.activeNotebookEditor) {
-                    this.onDidOpenNotebookDocument(this.notebook.activeNotebookEditor.notebook);
-                }
-                this.notebook.notebookDocuments.map((nb) => this.onDidOpenNotebookDocument(nb));
-            }, this)
+            this.registration.onCreated(
+                () => this.notebook.notebookDocuments.map((nb) => this.onDidOpenNotebookDocument(nb)),
+                this
+            )
         );
     }
     public dispose() {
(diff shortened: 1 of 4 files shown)


src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -49,35 +50,51 @@ import {
...
+        @inject(IDisposableRegistry) disposables: IDisposableRegistry,
         @inject(IPythonExtensionChecker) private readonly extensionChecker: IPythonExtensionChecker,
         @inject(IServerConnectionType) private readonly serverConnectionType: IServerConnectionType,
         @inject(IKernelRankingHelper) private readonly kernelRankHelper: IKernelRankingHelper
-    ) {}
-    public async activate() {
+    ) {
+        disposables.push(this);
+    }
+    public activate() {
```
DonJayamanne, line 73: We don't need an `async` activate

src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -88,7 +105,10 @@ export class ControllerPreferredService implements IControllerPreferredService,
     }> {
         traceInfoIfCI(`Clear controller mapping for ${getDisplayPath(document.uri)}`);
         // Keep track of a token per document so that we can cancel the search if the doc is closed
+        this.preferredCancelTokens.get(document)?.cancel();
+        this.preferredCancelTokens.get(document)?.dispose();
```
DonJayamanne, line 109: Cancel previous

src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -88,7 +105,10 @@ export class ControllerPreferredService implements IControllerPreferredService,
...
         const preferredSearchToken = new CancellationTokenSource();
+        this.disposables.add(preferredSearchToken);
```
DonJayamanne, line 111: This could get called a lot for every document, over & over when new controllers are added
As a result we could end up with a lot of CancellationTokens, yes we can add them into the global Disposables list, however that means we end up slowing the main dispose method of VS Code just processing 100s if not 1000s of these when the extension is unloaded or the like.

Hence all we do here is, we keep a set of the disposables and clean up as and when they're done with, this way we end up with a significantly smaller list of disposables that need to be disposed when the extenion is disposed (in fact we could end up with just 0 being contributed by this code)

rebornix (approved):

Issue #7670, opened 2021-09-27
Can't "restart" notebook debug session


roblourens:
- Start debugging a cell
- Click "restart"
- Get an error about a missing debug adapter descriptor

Need to handle debugging being started this way too, not just through the command

DavidKutu:
![image](https://user-images.githubusercontent.com/20347396/134966895-ea65644b-2535-48b6-a447-3d6b256b2cb6.png)
Is this it? this is the only error I get

roblourens:
That's not it - it will fail to start because it calls our debug adapter descriptor factory without doing the setup that we do when the command is run.

roblourens:
Why wouldn't it be supported?

DonJayamanne:
Its not supported today in Interactive window, hence I'm assuming its the same reason its not supported in run by line as well.
Either we fix it (which we haven't for over 2 years the bug has been around) or we notify users via a message.

I'm up for fixing it, thats better than some hacky warning about not supporting something (when it works in traditional Python).

PR #11668, opened 2022-10-14
Implement restart 

roblourens:
Towards #7670

Also includes some refactoring so that this flow reads more nicely.
I had to
- Flip around the flow of starting a cell debug session- previously it did init work in the command, starting a debug session, instantiating the DA then passing the session object back to the command handler. Now the command only kicks off a debug session, and all the interesting work happens inside the debug adapter factory, which is more straightforward
- Modify the initialize request to add the capability flag that the DA will implement the "restart" request
- Intercept the restart request, and start a new debug session. The reason we have to implement the restart request is because the index of the notebook cell may have changed since the original request, and we will start a launch config using the new index.

Remaining: This is not hooked up for IW debugging yet


src/notebooks/debugger/commandRegistry.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    private async runByLine(cell: NotebookCell | undefined) {
+        sendTelemetryEvent(DebuggingTelemetry.clickedRunByLine);
+        cell ??= this.getCellFromActiveEditor();
+        if (!cell) {
+            return;
+        }
+
+        if (!cell) {
```
IanMatthewHuff, line 45: Looks like two copies of this code chunk?

src/notebooks/debugger/debuggingManagerBase.ts
```diff
@@ -167,7 +159,43 @@ export abstract class DebuggingManagerBase implements IDisposable {
         return kernel;
     }
 
-    protected async checkForIpykernel6(doc: NotebookDocument): Promise<IpykernelCheckResult> {
+    private findEditorForCell(cell: NotebookCell): NotebookEditor | undefined {
+        const notebookUri = cell.notebook.uri.toString();
```
IanMatthewHuff, line 163: Doesn't really matter, but looks like this is toStringed twice.

IanMatthewHuff (approved):
Awesome. Looks great.

connor4312 (approved):

sbatten (approved):

PR #11674, opened 2022-10-16
Lazy load list of all Python Environments

### LINKED ISSUE #7583: Adopt new Python extensions non-blocking API for interpreter disocvery

DonJayamanne:
Fixes #7583

### DIFF
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -1913,22 +1913,6 @@ Expand each section to see more information about that event.
         Common measurement used across a number of events.  
 
 
-* DS_INTERNAL.INTERPRETER_LISTING_PERF  (Telemetry.InterpreterListingPerf)  
-      Owner: [@donjayamanne](https://github.com/donjayamanne)  
-    ```
-    Time taken to list the Python interpreters.  
-    ```
-
-    - Properties:  
-        - `firstTime`?: `boolean`  
-        Whether this is the first time in the session.  
-        (fetching kernels first time in the session is slower, later its cached).  
-        This is a generic property supported for all telemetry (sent by decorators).  
-    - Measures:  
-        - `duration`: `number`  
-        Total time taken to list interpreters.  
-
-
 * DS_INTERNAL.IPYWIDGET_DISCOVER_WIDGETS_NB_EXTENSIONS  (Telemetry.DiscoverIPyWidgetNamesPerf)  
       Owner: [@donjayamanne](https://github.com/donjayamanne)  
     ```
--- a/src/kernels/raw/finder/localKernelFinder.node.ts
+++ b/src/kernels/raw/finder/localKernelFinder.node.ts
@@ -27,9 +27,9 @@ import * as localize from '../../../platform/common/utils/localize';
 import { debounceAsync } from '../../../platform/common/utils/decorators';
 import { IPythonExtensionChecker } from '../../../platform/api/types';
 import { IInterpreterService } from '../../../platform/interpreter/contracts';
-import { EnvironmentType } from '../../../platform/pythonEnvironments/info';
 import { ContributedKernelFinderKind } from '../../internalTypes';
 import { PYTHON_LANGUAGE } from '../../../platform/common/constants';
+import { KnownEnvironmentTypes } from '../../../platform/api/pythonApiTypes';
 
 // This class searches for local kernels.
 // First it searches on a global persistent state, then on the installed python interpreters,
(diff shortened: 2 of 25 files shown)


TELEMETRY.md
```diff
@@ -1913,22 +1913,6 @@ Expand each section to see more information about that event.
         Common measurement used across a number of events.  
 
 
-* DS_INTERNAL.INTERPRETER_LISTING_PERF  (Telemetry.InterpreterListingPerf)
```
DonJayamanne, line 1916: This telemetry is no longer valid, as we no longer have an API to fetch all python enviornments.

src/test/datascience/execution.unit.test.ts
```diff
@@ -1,1135 +0,0 @@
-// Copyright (c) Microsoft Corporation.
```
DonJayamanne, line 1: See here https://github.com/microsoft/vscode-jupyter/pull/11673
This PR depends on Removed in separate PR https://github.com/microsoft/vscode-jupyter/pull/11673

src/kernels/raw/finder/localKernelFinder.node.ts
```diff
@@ -90,38 +79,18 @@ export class LocalKernelFinder implements ILocalKernelFinder, IExtensionSyncActi
...
 
     private async loadInitialState() {
-        traceVerbose('LocalKernelFinder: load cache');
-        // loading cache, which is resource agnostic
-        const kernelsFromCache = await this.getFromCache();
-        let kernels: LocalKernelConnectionMetadata[] = [];
-
-        this.updateCache().catch(noop);
-
-        if (Array.isArray(kernelsFromCache) && kernelsFromCache.length > 0) {
```
DonJayamanne, line 117: We need to cache anything anymore, seems pointless and complicated.
Can add this later if we need this.

src/platform/api/pythonApi.ts
```diff
@@ -345,28 +355,23 @@ export class InterpreterService implements IInterpreterService {
...
+    public get environments(): readonly PythonEnvironmentV2[] {
+        this.getApi().catch(noop);
+        this.hookupOnDidChangeInterpreterEvent();
+        return this.api?.environments?.known || [];
+    }
 
-    @traceDecoratorVerbose('Get Interpreters', TraceOptions.Arguments | TraceOptions.BeforeCall)
-    public getInterpreters(): Promise<PythonEnvironment[]> {
-        const stopWatch = new StopWatch();
+    private getInterpreters(): Promise<PythonEnvironment[]> {
```
DonJayamanne, line 369: The method `getInterrpeters` is no longer used, we will not rely on a push model completely.

DonJayamanne: Fortunately this was used only in a couple of places in the extension code.

src/kernels/raw/finder/localKernelSpecFinderBase.node.ts
```diff
@@ -72,11 +71,6 @@ export abstract class LocalKernelSpecFinderBase<
         disposables.push(this);
     }
 
-    @testOnlyMethod()
-    public clearCache() {
```
DonJayamanne, line 76: Unused

alexr00 (approved):

PR #11676, opened 2022-10-17
Fix #11675. Show 3rd party connection display name when available.

### LINKED ISSUE #11675: Misleading connection error message from 3rd party jupyter server
When 3rd party jupyter server failed to connect, we often show **`https://remote` is not reachable**. The url in the error message is not useable as it's not telling users which server is unavailable.

image

rebornix:
Fixes #11675. When a server url is from 3rd party (baseUrl is `https://remote`), we should try to resolve its display name.

### DIFF
--- a/src/kernels/jupyter/jupyterConnection.ts
+++ b/src/kernels/jupyter/jupyterConnection.ts
@@ -129,7 +129,7 @@ export class JupyterConnection implements IExtensionSyncActivationService {
         }
     }
 
-    private getServerUri(uri: string): IJupyterServerUri | undefined {
+    public getServerUri(uri: string): IJupyterServerUri | undefined {
         const idAndHandle = extractJupyterServerHandleAndId(uri);
         if (idAndHandle) {
             return this.uriToJupyterServerUri.get(uri);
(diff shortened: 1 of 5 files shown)


src/kernels/jupyter/jupyterConnection.ts
```diff
@@ -129,7 +129,7 @@ export class JupyterConnection implements IExtensionSyncActivationService {
         }
     }
 
-    private getServerUri(uri: string): IJupyterServerUri | undefined {
+    public getServerUri(uri: string): IJupyterServerUri | undefined {
```
DonJayamanne, line 132: I think its best to use existing API.
You can use 
```
const idAndHandle = extractJupyterServerHandleAndId(uri);
if (idAndHandle) {
    const serverUri = IJupyterUriProviderRegistration.getJupyterServerUri(idAndHandle.id, idAndHandle.handle)
}
```

This way we don't have to make private stuff public.

DonJayamanne (changes requested):
I think its better to use existing API instead of making private stuff public

src/kernels/jupyter/jupyterConnection.ts
rebornix: @DonJayamanne thanks, I'm not aware of that API, `IJupyterUriProviderRegistration.getJupyterServerUri` is better!

rebornix:
@DonJayamanne updated the PR as suggested, thanks!

DonJayamanne (approved):

PR #11685, opened 2022-10-18
Fix #11681. Emit change event when multiple kernels are registered.

### LINKED ISSUE #11681: Jupyter doesn't finish kernel detection before resolving api method addRemoteJupyterServer
Neither `addRemoteJupyterServer` nor `getSuggestedController` will guarantee creating controllers for the kernels found in the kernel finders as now they are push model. 

This is breaking the workflow:

* Users click run cell
* Kernels not available, users pick a kernel source command from the kernel picker
* kernel source command calls the `addRemoteJupyterServer` api to register a remote server
* 🐛 Kernels are populated to VS Code and VS Code continues to run the cells

rebornix:
Fixes #11681.

This PR made two changes:

* Emit change event when multiple notebook controllers are created in a kernel finder update event cycle.
* Ensure that we have controllers registered from a remote server when the remote server is added

### DIFF
--- a/src/notebooks/controllers/controllerRegistration.ts
+++ b/src/notebooks/controllers/controllerRegistration.ts
@@ -29,7 +29,12 @@ import { traceError, traceVerbose } from '../../platform/logging';
 import { sendTelemetryEvent, Telemetry } from '../../telemetry';
 import { NotebookCellLanguageService } from '../languages/cellLanguageService';
 import { KernelFilterService } from './kernelFilter/kernelFilterService';
-import { IControllerRegistration, InteractiveControllerIdSuffix, IVSCodeNotebookController } from './types';
+import {
+    IControllerRegistration,
+    InteractiveControllerIdSuffix,
+    IVSCodeNotebookController,
+    IVSCodeNotebookControllerUpdateEvent
+} from './types';
 import { VSCodeNotebookController } from './vscodeNotebookController';
 
 /**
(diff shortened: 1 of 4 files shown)


src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -43,12 +48,16 @@ export class ControllerRegistration implements IControllerRegistration {
     }
     private registeredControllers = new Map<string, VSCodeNotebookController>();
     private creationEmitter = new EventEmitter<IVSCodeNotebookController>();
+    private changeEmitter = new EventEmitter<IVSCodeNotebookControllerUpdateEvent>();
     private registeredMetadatas = new Map<string, KernelConnectionMetadata>();
     private inKernelExperiment = false;
 
     public get onCreated(): Event<IVSCodeNotebookController> {
         return this.creationEmitter.event;
```
DonJayamanne, line 56: can you remove the onCreated event, so we'd have a consistent way to look for changes

DonJayamanne (changes requested):

src/notebooks/controllers/controllerRegistration.ts
rebornix: @DonJayamanne I like the idea of removing `onCreated` event and move towards `onChange`. Currently `onCreated` is being used in half a dozen places, removing it would require changes to them and I'm not inclined to replacing `onCreated(handler)` to `onChange(e => e.added.foreach(controller => handler(controller)` naively, but rather optimizing how we handle batch update events in each component. Does this make better sense? I can also remove them simply if that's more preferred.

DonJayamanne: Since I'm working in this space, I'll clean this up

DonJayamanne (approved):

src/notebooks/controllers/controllerRegistration.ts
rebornix: @DonJayamanne ❤️ thanks, I created https://github.com/microsoft/vscode-jupyter/issues/11694 to track this. I can help with the controllers part too.

PR #11687, opened 2022-10-18
Fix #11686. Trigger a cache update immediately after reading from cache.

### LINKED ISSUE #11686: Remote kernel list is partial on window reload
image

Sometimes after window reload, the remote kernel list is partial, e.g., showing only one kernel as above.

rebornix:
Fixes #11686.

### DIFF
--- a/src/kernels/jupyter/finder/remoteKernelFinder.ts
+++ b/src/kernels/jupyter/finder/remoteKernelFinder.ts
@@ -162,6 +162,8 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IExtensionSingle
         // If we finish the cache first, and we don't have any items, in the cache, then load without cache.
         if (Array.isArray(kernelsFromCache) && kernelsFromCache.length > 0) {
             kernels = kernelsFromCache;
+            // kick off a cache update request
+            this.updateCache().then(noop, noop);
         } else {
             try {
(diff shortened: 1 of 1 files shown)


aeschli (approved):

DonJayamanne (changes requested):
Please could you check the changes here https://github.com/microsoft/vscode-jupyter/pull/11692
I think that might be a better way, I didn't want to push changes to your PR
please use my PR as a suggestion.

I think its better to trigger the change event closer to where the values are updated, 
the problem today is the `getFromCache` updates the cached property when i think that's incorrect, as it should just return as the method implies, else the method should be named `initializeFromCache`

DonJayamanne (approved):
Appoving the PR as my nit with the current approach doesn't warrant a rejection,
it's merely a code smell

rebornix:
@DonJayamanne thanks for reviewing the changes. I guess the naming of the functions is a bit confusing but the reason we need one more `updateCache` is not that we rely on its side effect, it's ensuring that if we initially `loadCache` successfully, making sure that we trigger a real kernel find. It's possible that we found something in the cache, and there is no relevant emitted to trigger a research, we never do the real kernel find before users run a cell. Changes https://github.com/microsoft/vscode-jupyter/pull/11692 here are good but they don't fix the particular bug.

Issue #11694, opened 2022-10-18
Move off IControllerRegistration#onCreated


rebornix:
We now introduced `onChanged` to support sending multiple controllers creation in one event and also support delete event, it can replace the existing `onCreated` event. https://github.com/microsoft/vscode-jupyter/pull/11685

We want to go through `onCreated` usage in each component to see if we can optimize with the batch event, instead of a simple `onCreated(handler)` to `onChange(e => e.added.foreach(controller => handler(controller)` replacement.

DonJayamanne:
I've got a PR for this https://github.com/microsoft/vscode-jupyter/pull/11695

PR #11695, opened 2022-10-18
Remove redundant `onCreated` event

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/11694

### DIFF
--- a/src/notebooks/controllers/connectionTracker.ts
+++ b/src/notebooks/controllers/connectionTracker.ts
@@ -14,7 +14,7 @@ import {
     IKernelRankingHelper,
     IConnectionMru,
     IConnectionTracker,
-    IVSCodeNotebookController
+    IVSCodeNotebookControllerUpdateEvent
 } from './types';
 
 @injectable()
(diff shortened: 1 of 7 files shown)


alexdima (dismissed):

src/notebooks/controllers/commands/serverConnectionControllerCommands.ts
```diff
@@ -283,35 +283,37 @@ export class ServerConnectionControllerCommands implements IExtensionSingleActiv
...
+                            kind: QuickPickItemKind.Separator,
+                            label: kind
+                        });
+                        quickPickItems.push(pick);
+                        kindIndexes.set(kind, quickPickItems.length);
+                    } else {
+                        quickPickItems.splice(index, 0, pick);
+                        kindIndexes.set(kind, quickPickItems.length);
+                    }
+                    changeEmitter.fire(quickPickItems);
```
rebornix, line 314: Do we want to emit the changes once all `added` controllers are handled?

DonJayamanne: I think it should be ok, I don't see how it ocul dcause any issues (i.e. the behaviour is the same even in the past)

rebornix (approved):

src/notebooks/controllers/connectionTracker.ts
```diff
@@ -87,11 +87,15 @@ export class ConnectionTracker implements IExtensionSyncActivationService, IConn
...
-    private async onCreatedController(controller: IVSCodeNotebookController) {
+    private async onChangeController(e: IVSCodeNotebookControllerUpdateEvent) {
         await Promise.all(
-            workspace.notebookDocuments.map(async (notebook) => {
-                await this.updateAffinity(notebook, controller.connection);
-            })
+            e.added.map((controller) =>
+                Promise.all(
+                    workspace.notebookDocuments.map(async (notebook) => {
+                        await this.updateAffinity(notebook, controller.connection);
```
rebornix, line 95: Maybe we can have a batch affinity update impl.

rebornix (approved):

src/notebooks/controllers/connectionTracker.ts
DonJayamanne: Will create a debt item for that.

Issue #11433, opened 2022-09-22
implement incremental updates for .py code lenses to improve performance


amunger:
We should be able to avoid re-parsing the entire file on every edit when providing code lenses

amunger:
marking this as complete for now. We'll look at performance again when we have more data and decide where to go from there

PR #11696, opened 2022-10-18
improve code lens provider performance

amunger:
for #11433 

When looking at a large file with many code lenses, most of the code lens provider's is spent creating the code lens objects that are returned, and resulting gc calls. Even more is spent serializing those objects for the response, but that is outside the extension's scope.

This change focuses on speeding up the iteration through code lens range, reducing the `getDocument` and `localize` calls.

Parsing the file itself to divide it into cells isn't that large of an impact, so I didn't spend the time implementing a more complex incremental update.

### DIFF
--- a/src/interactive-window/editor-integration/codeLensFactory.ts
+++ b/src/interactive-window/editor-integration/codeLensFactory.ts
@@ -24,7 +24,8 @@ import { getInteractiveCellMetadata } from '../helpers';
 import { IKernelProvider } from '../../kernels/types';
 import { CodeLensCommands, Commands, InteractiveWindowView } from '../../platform/common/constants';
 import { generateCellRangesFromDocument } from './cellFactory';
-import { ICodeLensFactory, IGeneratedCode, IGeneratedCodeStorageFactory } from './types';
+import { CodeLensPerfMeasures, ICodeLensFactory, IGeneratedCode, IGeneratedCodeStorageFactory } from './types';
+import { StopWatch } from '../../platform/common/utils/stopWatch';
 
 type CodeLensCacheData = {
     cachedDocumentVersion: number | undefined;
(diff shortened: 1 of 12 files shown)


src/interactive-window/editor-integration/codeLensFactory.ts
```diff
@@ -125,21 +139,28 @@ export class CodeLensFactory implements ICodeLensFactory {
             needUpdate = true;
         }
 
+        // Do not generate interactive window codelenses for TextDocuments which are part of NotebookDocuments
+        if (workspace.notebookDocuments.find((notebook) => notebook.uri.toString() === document.uri.toString())) {
+            return cache;
```
rebornix, line 144: should it return `[]` then instead of cache?

rebornix:
It was probably not introduced in this PR but the check for text document uris seem wrong.

src/interactive-window/editor-integration/codeLensFactory.ts
```diff
@@ -125,21 +139,28 @@ export class CodeLensFactory implements ICodeLensFactory {
             needUpdate = true;
         }
 
+        // Do not generate interactive window codelenses for TextDocuments which are part of NotebookDocuments
+        if (workspace.notebookDocuments.find((notebook) => notebook.uri.toString() === document.uri.toString())) {
```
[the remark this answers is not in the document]
amunger: I can just remove this, the provider filters out by scheme immediately and we can filter even earlier to only python files when registering the providers in the first place

src/interactive-window/editor-integration/codelensProviderActivator.ts
```diff
@@ -17,7 +17,7 @@ export class CodeLensProviderActivator implements IExtensionSyncActivationServic
 
     public activate() {
         this.extensionContext.subscriptions.push(
-            languages.registerCodeLensProvider([PYTHON_FILE_ANY_SCHEME], this.dataScienceCodeLensProvider)
+            languages.registerCodeLensProvider([PYTHON_FILE, PYTHON_UNTITLED], this.dataScienceCodeLensProvider)
```
rebornix, line 20: `export const PYTHON_FILE = { scheme: 'file', language: PYTHON_LANGUAGE };` we won't have codelens on `.dev` if we enforce `file` scheme.

rebornix (changes requested):

src/interactive-window/editor-integration/codelensProviderActivator.ts
amunger: what scheme is used on .dev?

amunger: reverted

rebornix (approved):

PR #11707, opened 2022-10-19
Ensure we catch unhandled promise errors

DonJayamanne:
@roblourens Found that `handleMessage` is returning a promise and sometimes in our tests these methods can throw errors, however VS Code doesn't handle the promises returned, after all VS Code is expecting this method to be sync.

As a result of this, we now have unhandled promise rejections in the extesnion (which node.js doesn't like & I too hate such warnings)

This PR ensures we swallow the promise exceptions from the `handleMessage` metjod (which is ideally supposed to be a sync method).
I'll leave it upto you to either clean this up to ensure the super method is not async, but this should fix failreus due to `unhandled rejections`

Took me a few days to figure out this failure (when I've been running into this locally and sometimes on CI)

### DIFF
--- a/src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
+++ b/src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
@@ -15,6 +15,7 @@ import { getInteractiveCellMetadata } from '../../../interactive-window/helpers'
 import { KernelDebugAdapterBase } from '../../../notebooks/debugger/kernelDebugAdapterBase';
 import { InteractiveCellMetadata } from '../../editor-integration/types';
 import { IDebugService } from '../../../platform/common/application/types';
+import { noop } from '../../../platform/common/utils/misc';
 /**
  * KernelDebugAdapter listens to debug messages in order to translate file requests into real files
  * (Interactive Window generally executes against a real file)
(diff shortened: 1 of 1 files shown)


src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -67,15 +68,19 @@ export class KernelDebugAdapter extends KernelDebugAdapterBase {
...
             this.debugLocationTracker.onWillReceiveMessage(message);
         }
         if (message.type === 'response' && this.debugLocationTracker?.onDidSendMessage) {
             this.debugLocationTracker.onDidSendMessage(message);
         }
-        return super.handleMessage(message);
+        const promise = super.handleMessage(message);
+        // The VS Code debugger class does not support an async `handleMessage`, its supposed to be sync.
+        // As a result, any errors here will not be handled and they'll be treated as unhandled errors.
+        promise.catch(noop);
```
roblourens, line 82: Thanks, but it looks like there is already a try/catch around handleMessage in the base class, right? Even so, would you want to log the error?

src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
```diff
@@ -67,15 +68,19 @@ export class KernelDebugAdapter extends KernelDebugAdapterBase {
         }
     }
 
-    override handleMessage(message: DebugProtocol.ProtocolMessage): Promise<KernelMessage.IDebugReplyMsg | undefined> {
+    override handleMessage(message: DebugProtocol.ProtocolMessage) {
```
roblourens, line 71: Just curious why you don't like method return signatures, I always add them, just my personal preference

roblourens:
> however VS Code doesn't handle the promises returned,

I think the technically right thing for us to do would be to send an error response back (if we failed handling a request), I can look into that.

> Took me a few days to figure out this failure

Can you tell me what the actual error/failure is?

src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
DonJayamanne: i think we should change this, as is not supposed to be async.
ie i think we should change this to sync, probably should have left it there's as the code is still async.

src/interactive-window/debugger/jupyter/kernelDebugAdapter.ts
DonJayamanne: oh, didn't realise that, will close this then.
i guess i added it at some point then

PR #11720, opened 2022-10-19
Ensure CI pipelines run against vscode stable

DonJayamanne:


rzhao271 (approved):

src/test/datascience/notebook/helper.ts
```diff
@@ -656,18 +656,10 @@ export async function waitForKernelToGetAutoSelectedImpl(
         let preferred: IVSCodeNotebookController | undefined;
 
         // Wait for one of them to have affinity as the preferred (this may not happen)
-        try {
-            await waitForCondition(
-                async () => {
-                    preferred = controllerPreferred.getPreferred(notebookEditor!.notebook);
-                    return preferred != undefined;
-                },
-                30_000,
```
DonJayamanne, line 665: This condition was causing things to timeout, looks like even in our stable tests we wait for 30s unnecessarily.
With the recent changes we end up calling this code over & over, hence we have 30+30+30+30+.... resulting in a timeout of the tests, this should make the tests faster

& we need to fix this (test) code, its just a giant hack with waits and retries, all to ensure we either wait for the right controller to get selected or to select the right controller when its available.

During debt I plan on working on this area - simplifying and speeding up the tests

amunger (approved):

roblourens (approved):

PR #11763, opened 2022-10-24
update install python controller command to use api directly and remove controller loaded context

### LINKED ISSUE #11647: Change how context key `jupyter.controllersLoaded` is updated
@IanMatthewHuff 
As per discussion I think we can:
* Directly use the Python API to determine if python envs are available or not instead of checking the controllers.
* Remove the code to update `jupyter.controllersLoaded` from the `ControllerLoaded` class

IanMatthewHuff:
Fixes #11647 

### DIFF
--- a/src/notebooks/controllers/commands/installPythonControllerCommands.ts
+++ b/src/notebooks/controllers/commands/installPythonControllerCommands.ts
@@ -12,7 +12,6 @@ import {
     window
 } from 'vscode';
 import { IDataScienceErrorHandler } from '../../../kernels/errors/types';
-import { isPythonKernelConnection } from '../../../kernels/helpers';
 import { IExtensionSingleActivationService } from '../../../platform/activation/types';
 import { IPythonApiProvider, IPythonExtensionChecker } from '../../../platform/api/types';
 import { IApplicationShell, ICommandManager } from '../../../platform/common/application/types';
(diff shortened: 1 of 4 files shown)


src/notebooks/controllers/commands/installPythonControllerCommands.ts
```diff
@@ -109,7 +107,7 @@ export class InstallPythonControllerCommands implements IExtensionSingleActivati
                     return;
                 }
 
-                if (!this.foundPythonConnections) {
+                if (this.interpreterService.environments.length === 0) {
                     // Extension is installed, but we didn't find any python connections
                     // recommend installing python in this case
                     await this.showInstallPythonExtensionContext.set(false);
```
DonJayamanne, line 113: I don't get code.
Shouldn't this be `this.interpreterService.environments.length > 0`

DonJayamanne (changes requested):
Need clarification on the logic

src/notebooks/controllers/commands/installPythonControllerCommands.ts
IanMatthewHuff: @DonJayamanne the code in this block should be hit if:

1. Python Extension is installed
2. We don't have any valid python environments to use (length of envs === 0)

I would think your suggestion would be saying if we did find environments to use. The translation from previous code is from 'did not find python connections' to 'did not find python environments'. Or am I misunderstanding?

DonJayamanne: Oh ok, yes its correct, the next line is `await this.showInstallPythonContext.set(true);`
& Github conviniently failed to display that crucial line.
All good.

DonJayamanne (approved):

src/notebooks/controllers/commands/installPythonControllerCommands.ts
IanMatthewHuff: No problem! I just always worry that I flubbed the logic for things like this, so I was looking it over again and again 😆 .

PR #11770, opened 2022-10-25
Don't enable justMyCode for run-by-line

roblourens:
Part of #8146


karthiknadig (approved):

fabioz:
I'm not sure this is ideal... rather, the debugger should have a better API so that vscode-jupyter can ask to stop just in user code (see comment: https://github.com/microsoft/vscode-jupyter/issues/8146#issuecomment-1290324384).

roblourens:
This change is only about enabling justMyCode, not related to stopping in user code.

PR #11807, opened 2022-10-25
Exclude python envs from other workspace folders

### LINKED ISSUE #11779: Consider not showing venvs from other directories
Testing #11755

I think this work would actually fall on me on the Controller level (so I don't expect you to have to do this), but putting this item here for consideration since the new Python API did expose this. One thing that I always liked about using venvs is that they were restricted to just the folder path that I had open. Now I'm getting way too many non-relevant venvs from all over my system. 

image

I'm no python expert so I'm totally fine if folks disagree and we don't do this. It's just a downgrade for how I use Python + VS Code notebooks.

DonJayamanne:
Fixes #11779

### DIFF
--- a/src/platform/api/pythonApi.ts
+++ b/src/platform/api/pythonApi.ts
@@ -589,9 +589,12 @@ export class InterpreterService implements IInterpreterService {
         );
         return allInterpreters;
     }
+    /**
+     * Python extension API returns all envs from all known workspace folders, including those that do not
+     * belong to the currently opened workspace folders.
+     * This will is used to determine whether an env belongs to the currently opened workspace folders.
+     */
     private isValidWorkSpaceRelatedEnvironment(env: PythonEnvironmentV2 | ResolvedEnvironment) {
-        // If this environment belongs to a different workspace folder, then ignore this.
-        // Python API returns envs from all workspace folders, not just the active ones in VS Code.
         const envWorkspaceFolder = env?.environment?.workspaceFolder;
         if (envWorkspaceFolder) {
             if (!this.workspace.workspaceFolders) {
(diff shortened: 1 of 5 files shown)


src/platform/api/pythonApi.ts
```diff
@@ -548,19 +548,31 @@ export class InterpreterService implements IInterpreterService {
...
+                            try {
+                                const env = await api.environments.resolveEnvironment(item.id);
+                                const resolved = this.trackResolvedEnvironment(env);
+                                traceVerbose(
+                                    `Python environment ${env?.id} from Python Extension API is ${JSON.stringify(
+                                        env
+                                    )} and translated is ${JSON.stringify(resolved)}`
+                                );
+                                if (!this.isValidWorkSpaceRelatedEnvironment(item)) {
+                                    return;
```
DonJayamanne, line 563: Filter after we get the details, in case the previous env information was partial (empty)

src/platform/api/pythonApi.ts
```diff
@@ -548,19 +548,31 @@ export class InterpreterService implements IInterpreterService {
...
-                    api.environments.known.map(async (item) => {
-                        try {
-                            const env = await api.environments.resolveEnvironment(item.id);
-                            const resolved = this.trackResolvedEnvironment(env);
-                            if (resolved) {
-                                allInterpreters.push(resolved);
-                            } else {
-                                traceError(`Failed to get env details from Python API for ${item.id} without an error`);
+                    api.environments.known
+                        .filter((item) => this.isValidWorkSpaceRelatedEnvironment(item))
```
DonJayamanne, line 552: Filter with what information we have

DonJayamanne: This is a requirement of the new Python API, they return envs for all worksapce folders

roblourens (approved):

PR #11809, opened 2022-10-26
Fix display names of Python Envs in kernel Picker

### LINKED ISSUE #11783: Display names of Python environments are incorrect in the kernel picker
Testing #11755

VS Code stable

image

Insiders

image

Note that in Insiders we don't have the complete Python version info.

DonJayamanne:
Fixes #11783

### DIFF
--- a/src/platform/api/pythonApi.ts
+++ b/src/platform/api/pythonApi.ts
@@ -79,7 +79,7 @@ export function pythonEnvToJupyterEnv(env: ResolvedEnvironment): PythonEnvironme
     return {
         id: env.id,
         sysPrefix: env.executable.sysPrefix,
-        envPath: Uri.file(env.path),
+        envPath: env.environment?.folderUri,
         envName: env.environment?.name,
         uri: env.executable.uri!,
         displayName: env.environment?.name,
--- a/src/platform/api/pythonApi.ts
+++ b/src/platform/api/pythonApi.ts
@@ -613,6 +613,7 @@ export class InterpreterService implements IInterpreterService {
                     this.eventHandlerAdded = true;
                     api.environments.onDidChangeActiveEnvironmentPath(
                         () => {
+                            traceVerbose(`Detected change in Active Python environment via Python API`);
                             this.interpreterListCachePromise = undefined;
                             this.workspaceCachedActiveInterpreter.clear();
                             this.didChangeInterpreter.fire();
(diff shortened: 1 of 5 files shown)


src/platform/api/pythonApi.ts
```diff
@@ -79,7 +79,7 @@ export function pythonEnvToJupyterEnv(env: ResolvedEnvironment): PythonEnvironme
     return {
         id: env.id,
         sysPrefix: env.executable.sysPrefix,
-        envPath: Uri.file(env.path),
+        envPath: env.environment?.folderUri,
```
DonJayamanne, line 82: This mapping was incorrect with the new API

DonJayamanne: In the past `path` pointed to the env path, now `path` seems to point to the executable

karthiknadig: @karrtikr can you clarify this?

DonJayamanne: Yup, if you check the old API, `pythonVersion.envPath` pointed to the path of the enviorinment, not the executable.
See here https://github.com/DonJayamanne/pythonVSCode/blob/main/src/client/pythonEnvironments/info/index.ts#L84

```typescript
/**
 * Details about a Python environment.
 *
 * @prop companyDisplayName - the user-facing name of the distro publisher
 * @prop displayName - the user-facing name for the environment
 * @prop type - the kind of Python environment
 * @prop envName - the environment's name, if applicable (else `envPath` is set)
 * @prop envPath - the environment's root dir, if applicable (else `envName`)
 * @prop cachedEntry - whether or not the info came from a cache
 */
// Note that "cachedEntry" is specific to the caching machinery
// and doesn't really belong here.
export type PythonEnvironment = InterpreterInformation & {
    id?: string;
    companyDisplayName?: string;
    displayName?: string;
    detailedDisplayName?: string;
    envType: EnvironmentType;
    envName?: string;
    envPath?: string;
    cachedEntry?: boolean;
};
```

DonJayamanne: Note: its not a problem with the Python extensoin, this is basically an issue at our end, as menioend here `This mapping was incorrect with the new API`

roblourens (approved):

src/platform/api/pythonApi.ts
karrtikr: `env.path` is different from the former `envPath` property as Don just described, here's the doc for what it represents:

image

It's useful for the display path of the environment:

image

roblourens (approved):

PR #11819, opened 2022-10-26
Do not trigger unnecessary environment changes

DonJayamanne:
Fail safe for https://github.com/microsoft/vscode-python/issues/20070

### DIFF
--- a/src/platform/api/pythonApi.ts
+++ b/src/platform/api/pythonApi.ts
@@ -418,7 +418,7 @@ export class InterpreterService implements IInterpreterService {
             }
             const envPath = api.environments.getActiveEnvironmentPath(resource);
             const env = await api.environments.resolveEnvironment(envPath);
-            return this.trackResolvedEnvironment(env);
+            return this.trackResolvedEnvironment(env, false);
         });
 
         // If there was a problem in getting the details, remove the cached info.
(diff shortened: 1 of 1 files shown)


src/platform/api/pythonApi.ts
```diff
@@ -487,19 +487,29 @@ export class InterpreterService implements IInterpreterService {
...
             if (
                 !this._interpreters.get(env.id) ||
                 !areObjectsWithUrisTheSame(resolved, this._interpreters.get(env.id)?.resolved)
             ) {
-                changed = true;
                 this._interpreters.set(env.id, { resolved });
-            }
-            if (changed) {
-                this.didChangeInterpreters.fire();
+                if (triggerChangeEvent) {
```
amunger, line 510: should this still also check `changed`?

amunger: it looks like we're now just firing this event for every interpreter found

DonJayamanne: No need, if you check the code, previously `changed` was initialize din line 498, all i've done is moved this code into the same if block, which was effectively the same, & there by removing the unnecessary variable as well

DonJayamanne: I.e. it will not fire everytime, only when there are changes (where previously we'd set `changed = true` in the same if condition

amunger (approved):

PR #11827, opened 2022-10-27
Do not recommend controller if its the only one

### LINKED ISSUE #11826: Do not reccommend the only controller as the preferred controller even if it doesn't match
Details described here https://github.com/microsoft/vscode/issues/164827

* Jupyter extension loads and detects kernel specs from disc and creates a controller for that (which happens to be typescript_
* Note: In the back ground Juyter extnesion is still busy finding other kernelspecs and Pytohn environments
* Jupyter extension has only one controller and sets that controller as the preferred controller.
* Jupyter then sees other controllers later, but doesn't find any matching controllers.

I think there are a few problems here:
* We should un-set the preffered controller once we get other controllers loaded

DonJayamanne:
Fixes #11826

### DIFF
--- a/src/notebooks/controllers/controllerPreferredService.ts
+++ b/src/notebooks/controllers/controllerPreferredService.ts
@@ -305,6 +305,18 @@ export class ControllerPreferredService implements IControllerPreferredService, 
 
                 // Save in our map so we can find it in test code.
                 this.preferredControllers.set(document, targetController);
+            } else if (!preferredConnection && this.preferredControllers.get(document)) {
+                // Possible previously we had just 1 controller and that was setup as the preferred
+                // & now that we have more controllers, we know more about what needs to be matched
+                // & since we no longer have a preferred, we should probably unset the previous preferred
+                traceVerbose(
+                    `Resetting the previous preferred controller ${
+                        this.preferredControllers.get(document)?.id
+                    } to default affinity for document ${getDisplayPath(document.uri)}`
+                );
+                await this.preferredControllers
+                    .get(document)
+                    ?.controller.updateNotebookAffinity(document, NotebookControllerAffinity.Default);
             }
             traceInfoIfCI(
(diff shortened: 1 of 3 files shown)


src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -397,9 +409,18 @@ export class ControllerPreferredService implements IControllerPreferredService,
...
                 languageInNotebookMetadata !== PYTHON_LANGUAGE &&
                 getKernelConnectionLanguage(potentialMatch) === languageInNotebookMetadata;
+            const isPythonLanguageMatch =
+                languageInNotebookMetadata &&
+                languageInNotebookMetadata === PYTHON_LANGUAGE &&
+                getKernelConnectionLanguage(potentialMatch) === languageInNotebookMetadata;
 
             // Match on our possible reasons
-            if (onlyConnection || topMatchIsPreferredInterpreter || isExactMatch || isNonPythonLanguageMatch) {
+            if (
```
DonJayamanne, line 418: Previously if we had jsut 1 controller we recommend that as the preferred.
Even if user opens a blank notebook.

E.g. previous logic
* If you have only typescript kernel
* You open a Python notebook
* We will still recommend using typescript as the preferred
* & then when we get other controllers, we no longer update the previous controller as not preferred.

I believe that's whats going on
Difficult to repro this as its all about timing, but based on the control flow that is a definite possibility.

Desperately need unit tests for this change, will work on that

IanMatthewHuff: Can we just remove the onlyConnection check totally (versus adding the isPythonLanguageMatch)? Given how push model works it seems pretty dubious in general. Too many weird cases in our code where kernels are added in out of sequence that might trigger odd suggestions here.

IanMatthewHuff (approved):

Issue #11800, opened 2022-10-25
"Discovering Python Interpreters" is occasionally very slow after window reload. 


IanMatthewHuff:
Testing #11755

I'm having a bit of trouble pinning down exactly when this happens, but I had assume that with caching this wouldn't be the case as much any more. Sometimes when I reload the window (no changes to python interpreters) the "Discovering Python Interpreters" step takes 30s or so. The whole wait time there I don't have access to any of the python controllers. 

Here is a semi-bad recording that I captured. Just a reload (nothing new added). 

https://user-images.githubusercontent.com/812783/197867350-4ef07bf0-642b-4147-bfce-8f3a55810da8.mov

DonJayamanne:
@IanMatthewHuff please could you provide the debug logs

IanMatthewHuff:
@DonJayamanne Looked to me like the time might be on the Python side, so I captured Jupyter and Python logs for one of these slow instances. 
[JupyterLog.txt](https://github.com/microsoft/vscode-jupyter/files/9864415/JupyterLog.txt)
[PythonLog.txt](https://github.com/microsoft/vscode-jupyter/files/9864416/PythonLog.txt)

DonJayamanne:
@IanMatthewHuff you should see a progress indicator contributed by pytohn extension on the bottom left of the status bar (coudln't see it in the video).
Let me know if you see that is visible.

I'm adding some logging that would help us figure out where the delay is.

PR #11831, opened 2022-10-27
Cache Kernels for faster loading

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/11800

### DIFF
--- a/src/kernels/raw/finder/localKernelSpecFinderBase.node.ts
+++ b/src/kernels/raw/finder/localKernelSpecFinderBase.node.ts
@@ -133,7 +133,7 @@ export abstract class LocalKernelSpecFinderBase<
         // ! as the has and set above verify that we have a return here
         return this.kernelSpecCache.get(cacheKey)!.promise;
     }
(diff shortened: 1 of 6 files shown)


sbatten (approved):

src/kernels/raw/finder/localKernelSpecFinderBase.node.ts
```diff
@@ -125,7 +133,64 @@ export abstract class LocalKernelSpecFinderBase<
...
+                }
+            })
+        );
+        return validValues;
+    }
+
+    protected async writeToMementoCache(values: T[], cacheKey: string) {
+        const serialized = values.map(serializeKernelConnection);
+        await Promise.all([
+            removeOldCachedItems(this.globalState),
```
IanMatthewHuff, line 171: Just to note, previously the cache keys used to be defined in the commonFinder.ts file where removeOldCachedItems was defined. Might still want to define them there so that they can be removed when versions update? I would still take this PR and just file an issue for that as it shouldn't hurt anything now.

IanMatthewHuff (approved):

PR #11832, opened 2022-10-28
Disable lazy loading of controllers

DonJayamanne:
This branch is based on https://github.com/microsoft/vscode-jupyter/pull/11831
Review only the last comment


src/notebooks/controllers/controllerLoader.ts
```diff
@@ -50,7 +50,7 @@ export class ControllerLoader implements IControllerLoader, IExtensionSyncActiva
         this.loadControllers();
     }
     private loadControllers() {
-        this.loadControllersImpl();
+        this.loadControllersImpl().ignoreErrors();
```
IanMatthewHuff, line 53: Looks like the two following lines for telemetry should be sent .then from the load controllers impl.

src/platform/api/pythonApi.ts
```diff
@@ -379,7 +379,22 @@ export class InterpreterService implements IInterpreterService {
...
+                .get<boolean>('lazyLoadControllers', false);
+            if (lazyLoadControllers) {
+                return;
+            }
+            this._waitForAllInterpretersToLoad = (async () => {
+                await this.refreshInterpreters();
+                await this.getInterpreters();
+            })();
+        }
+        return this._waitForAllInterpretersToLoad;
```
IanMatthewHuff, line 396: I'm going to dig a bit more, but I'm not sure this promise would ensure that the finders have actually recalculated their kernels when this promise is done. The gap I'm trying to consider is if the controllerLoader is going to move forward while local kernel finder is still recalculating its cache.

rebornix:
@DonJayamanne @IanMatthewHuff I have just merged the branch with main and retargeted it against main branch. The changes now are very minimal and considering we have already did a test run for this, IMHO we should merge it and have it released to Insiders users.

rebornix (approved):

PR #11839, opened 2022-10-28
Updates for next release

roblourens:


package.json
```diff
@@ -29,7 +29,7 @@
         "theme": "light"
     },
     "engines": {
-        "vscode": "^1.73.0-insider"
+        "vscode": "^1.74.0-insider"
```
IanMatthewHuff, line 32: Is this bump needed? I would think that main / insiders could keep releasing for 1.73.

IanMatthewHuff: Heh, it varies between the last three times this merge has been done:
https://github.com/microsoft/vscode-jupyter/commit/6119d4e801a992012979ef8774427880ec95e627
https://github.com/microsoft/vscode-jupyter/commit/be3a5f5fafcfb61509ebabb23a3ef56e5971e460
https://github.com/microsoft/vscode-jupyter/commit/63fd00ffcff810dc1bf07ff4a3653a3065a1edc5

roblourens: The release plan is ambiguous, it says

>  Bump the vscode engine version in package.json

right away, but also

>  Bump the engines.vscode version on the main branch to point to the next insider version. For example, from 1.58.0-insider to 1.59.0-insider

when the release happens, which is probably what we should do

IanMatthewHuff: Had to stop and think for a second. I think the issue here might be due to this PR not being a merge of a previous commit in the release branch. In this step here:

> Merge the changes from release (Changelog, delete news, ThirdPartyNotices) into main branch

Steps should look like this:
Release PR: https://github.com/microsoft/vscode-jupyter/pull/11498
Update Main: https://github.com/microsoft/vscode-jupyter/pull/11499

The main update should automatically pull in the bump of the vscode engine version in package.json. So 1.73.0-insiders should be going to 1.73.0. Like in #11499. There isn't an explicit bump in the main update needed as insiders can continue to release targetting 1.73.0 with the updated version number.

roblourens: I'm actually not sure what the effect of the `-insider` is on the vscode engine...

package-lock.json
```diff
@@ -284,7 +284,7 @@
                 "yargs-parser": "^13.1.2"
             },
             "engines": {
-                "vscode": "^1.73.0"
+                "vscode": "^1.74.0-insider"
```
IanMatthewHuff, line 287: Looks like package-lock still picked up the previous change. Just needs a regen I assume.

package.json
IanMatthewHuff: @roblourens Honestly wasn't sure myself. Looking at the docs it's...unclear to me?

https://code.visualstudio.com/api/working-with-extensions/publishing-extension#visual-studio-code-compatibility

Honestly my read of that section is that -insider might not mean anything and only the version matters?

IanMatthewHuff (approved):

PR #11850, opened 2022-10-31
Fix the wait for when interpreters are fully loaded, and wait for this refresh only on local kernels refresh

IanMatthewHuff:


IanMatthewHuff:
Note, this is a PR to merge into Don's PR in progress

rebornix (approved):

PR #11857, opened 2022-11-01
Simplify Kernel dependencies

DonJayamanne:
Simplify by removing some functionality into another class
* E.g. displaying status of the kernel can be done outside the kernel class

This simple change reduces the coupling and the number of arguments passed all around

### DIFF
--- a/src/kernels/kernelStatusProvider.ts
+++ b/src/kernels/kernelStatusProvider.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { inject } from 'inversify';
+import { inject, injectable } from 'inversify';
 import { IExtensionSyncActivationService } from '../platform/activation/types';
 import { disposeAllDisposables } from '../platform/common/helpers';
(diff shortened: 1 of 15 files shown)


src/kernels/kernelStatusProvider.ts
```diff
@@ -0,0 +1,68 @@
...
+import { IExtensionSyncActivationService } from '../platform/activation/types';
+import { disposeAllDisposables } from '../platform/common/helpers';
+import { IDisposable, IDisposableRegistry } from '../platform/common/types';
+import { DataScience } from '../platform/common/utils/localize';
+import { KernelProgressReporter } from '../platform/progress/kernelProgressReporter';
+import { IStatusProvider } from '../platform/progress/types';
+import { getDisplayNameOrNameOfKernelConnection } from './helpers';
+import { IKernel, IKernelProvider } from './types';
+
+export class KernelStatusProvider implements IExtensionSyncActivationService {
```
DonJayamanne, line 14: This simple change reduces the coupling and the number of arguments passed all around

lramos15 (approved):

PR #11861, opened 2022-11-01
Seprate kernel execution and reduced coupling

DonJayamanne:
Further refactoring to reduces coupling in kernels

### DIFF
--- a/src/kernels/kernel.ts
+++ b/src/kernels/kernel.ts
@@ -48,7 +48,8 @@ import {
     KernelActionSource,
     KernelHooks,
     IKernelSettings,
-    IKernelController
+    IKernelController,
+    IThirdPartyKernel
 } from './types';
 import { Cancellation, isCancellationError } from '../platform/common/cancellation';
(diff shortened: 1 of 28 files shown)


DonJayamanne:
@rebornix /cc

src/kernels/kernel.ts
```diff
@@ -661,7 +664,7 @@ abstract class BaseKernel<TKernelExecution extends BaseKernelExecution> implemen
     }
 }
 
-export class ThirdPartyKernel extends BaseKernel<ThirdPartyKernelExecution> {
+export class ThirdPartyKernel extends BaseKernel<ThirdPartyKernelExecution> implements IThirdPartyKernel {
```
DonJayamanne, line 667: @rebornix both ThirdPartyKernel and internal Kernel are almost identical.
The internal kernel has 2 additoinal properties, that's the only difference now.

roblourens (approved):

rebornix (approved):

Issue #11403, opened 2022-09-19
Flaky test __file__ exists even after restarting a kernel:


DonJayamanne:
https://github.com/microsoft/vscode-jupyter/actions/runs/3080681677/jobs/4978318234

PR #11876, opened 2022-11-02
renew the cancel token right after cancelling for kernel restart

amunger:
Fixes https://github.com/microsoft/vscode-jupyter/issues/11403

for flaky test `__file__ exists even after restarting a kernel`

An already cancelled token was being passed into the `waitForIdle` function [here](https://github.com/microsoft/vscode-jupyter/blob/aamunger/cancelledKernelRestart/src/kernels/kernel.ts#L709), so if it wasn't already idle, the restart operation would throw because it was cancelled.


DonJayamanne (approved):
Looks like debt, need to review the `startCancellation`
Approving, but I'm not even sure about the usage of this cancellation token (nothing to do with changes from this PR, but the existing code seems incorrect).

DonJayamanne:
I'll work on this soon, https://github.com/microsoft/vscode-jupyter/issues/11877
But please feel free to merge this PR

DonJayamanne (changes requested):
This change will not work as expected, let me have a closer look

src/kernels/kernel.ts
```diff
@@ -191,7 +191,6 @@ abstract class BaseKernel implements IBaseKernel {
             Array.from(this.hooks.get('willInterrupt') || new Set<Hook>()).map((h) => h())
         );
         traceInfo(`Interrupt requested ${getDisplayPath(this.resourceUri || this.uri)}`);
-        this.startCancellation.cancel();
```
DonJayamanne, line 194: This alone will not work, as the previous code line 280 was used to cancel the pending start, hence cancelling the previous and immediately initializing it with a new item will not have the desired effect of the cancellation.

Personally not sure why we're cancellign either, but this is existing code and reluctant to change that at this stage, will look into that in next iteration debt.

DonJayamanne, line 194: We don't need to cancel when interrupting a kernel, its wrong.
If we do i'd like to see more information (comments/tests for that case).
I'm removing this here as its not requried.

src/kernels/kernel.ts
```diff
@@ -277,11 +276,20 @@ abstract class BaseKernel implements IBaseKernel {
                 Array.from(this.hooks.get('willRestart') || new Set<Hook>()).map((h) => h(this._jupyterSessionPromise))
             );
             traceInfo(`Restart requested ${this.uri}`);
-            this.startCancellation.cancel();
+            this.startCancellation.cancel(); // Cancel any pending starts.
```
DonJayamanne, line 279: This alone will not work, as the previous code line 280 was used to cancel the pending start, hence cancelling the previous and immediately initializing it with a new item will not have the desired effect of the cancellation.

Personally not sure why we're cancellign either, but this is existing code and reluctant to change that at this stage, will look into that in next iteration debt.

DonJayamanne (approved):

src/kernels/kernel.ts
```diff
@@ -277,11 +276,20 @@ abstract class BaseKernel implements IBaseKernel {
...
             const stopWatch = new StopWatch();
             try {
-                // If the session died, then start a new session.
-                await (this._jupyterSessionPromise ? this.restartImpl() : this.start(new DisplayOptions(false)));
+                // Try to restart the current session if possible.
+                const result = await this.restartCurrentSession();
+                if (result === 'currentSessionRestarted') {
+                    // This happens when we've been unable to stop the pending start.
+                    // Of the session has already been started.
```
amunger, line 286: should that be "Or"?

amunger:
more test failures now

src/kernels/kernel.ts
amunger: initializing a new `cancellationSource` won't affect the token that was previously passed to the tasks that would need to be cancelled.

DonJayamanne:
> more test failures now

Branch needs to be updated, nothing to do with the PR changes.

PR #11892, opened 2022-11-03
remove OnlyOneTypeOfKernel picker experiment

IanMatthewHuff:
Removing the "onlyOneTypeOfKernel" experiment since we won't be heading that direction. Helps make it easier to look at just the difference between stable and insiders experiment.


src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
```diff
@@ -165,3 +164,21 @@ export class NotebookKernelSourceSelector implements INotebookKernelSourceSelect
         return { kernelFinderInfo, label: kernelFinderInfo.displayName };
     }
 }
+
+function groupBy<T>(data: ReadonlyArray<T>, compare: (a: T, b: T) => number): T[][] {
```
IanMatthewHuff, line 168: Moved from the deleted file serverConnectionControllerCommands

amunger (approved):

rebornix (approved):

PR #11909, opened 2022-11-04
Faster builds

DonJayamanne:
These two packages were excluded from the main bundle, however they are very small and inlined into the web.
Hence inlinine int the node bundle doesn't add much (some of the larger packages have been removed in another PR)


build/webpack/webpack.extension.web.config.js
```diff
@@ -89,6 +90,7 @@ const config = {
     },
     externals: ['vscode', 'commonjs', 'electron'], // Don't bundle these
     plugins: [
+        ...common.getDefaultPlugins('web'),
```
DonJayamanne, line 93: So we can analyze the bundle using `bundle-analyzer`, this change doesn't impact the build only generates an html file during dev build based on an env variable.

karthiknadig (approved):

DonJayamanne:
@karthiknadig please re-approve, had a merge conflict which required manual updates

build/webpack/common.js
```diff
@@ -6,15 +6,7 @@ const glob = require('glob');
 const path = require('path');
 const webpack_bundle_analyzer = require('webpack-bundle-analyzer');
 const constants = require('../constants');
-exports.nodeModulesToExternalize = [
-    '@jupyterlab/services',
-    'node-stream-zip',
```
DonJayamanne, line 11: These are either inlined today, hence marking them as externals is a moot point (just adding more into a seprate bundle).

rebornix (approved):

Issue #11663, opened 2022-10-14
Don't store the `NotebookEditor` instance as a member of the `InteractiveWindow`


amunger:
We should avoid keeping around the same instance of the NotebookEditor since it could become invalid after it becomes hidden, or another notebook document is opened in the same editor.

From a quick look, we might be able to just hold onto the `NotebookDocument` instead.

PR #11947, opened 2022-11-08
Reduce usage of editor in InteractiveWindow

amunger:
the simple parts of https://github.com/microsoft/vscode-jupyter/issues/11663

Don't use the editor to access the notebook since we already hold the notebook, and compare the notebook URI instead of the editor instance since it's not guaranteed that the editor is the same.


mjbvz (approved):

src/interactive-window/interactiveWindow.ts
```diff
@@ -200,7 +200,7 @@ export class InteractiveWindow implements IInteractiveWindowLoadable {
             }
         }, this.internalDisposables);
 
-        if (window.activeNotebookEditor === this.notebookEditor) {
+        if (window.activeNotebookEditor?.notebook.uri.toString() === this.notebookUri.toString()) {
```
DonJayamanne, line 203: No need to change, however you could compare the instances of the Notebook objects as well, as that will be the same.

PR #11956, opened 2022-11-08
Ended up with a recursive loop, fixes that

DonJayamanne:


rebornix (approved):

src/notebooks/controllers/connectionDisplayData.ts
```diff
@@ -45,8 +45,7 @@ export class ConnectionDisplayDataProvider {
         const description = getKernelConnectionDisplayPath(connection, this.workspace, this.platform);
         const detail =
             connection.kind === 'connectToLiveRemoteKernel' ? getRemoteKernelSessionInformation(connection) : '';
-        const category =
-            connection.kind === 'connectToLiveRemoteKernel' ? getRemoteKernelSessionInformation(connection) : '';
+        const category = '';
```
roblourens, line 48: If this was a recursive loop, why wasn't the line above?

DonJayamanne: The category ends up with a different value everytime, 
If you check further below we update this later on via an async callback.
However when we access this yet again, then the value for category is overwritten with some other value, which then results in the ayns callback causing an update with a different value.

DonJayamanne: Only applies to `category` as thats the only one updated asynchronously

PR #11989, opened 2022-11-11
Reduce kernel status flickr, fewer affinity change

### LINKED ISSUE #11988: Refactor/Cleanup preferred controller and controller registration and related code
* Remove the following code in loadControllersImpl
```
// This is temporary, when we create an MRU list in VS Code or the like, this should go away.
```

DonJayamanne:
Fixes #11988

### DIFF
--- a/src/notebooks/controllers/controllerRegistration.ts
+++ b/src/notebooks/controllers/controllerRegistration.ts
@@ -82,26 +82,30 @@ export class ControllerRegistration implements IControllerRegistration {
         this.serverUriStorage.onDidRemoveUris(this.onDidRemoveUris, this, this.disposables);
     }
     batchAdd(metadatas: KernelConnectionMetadata[], types: ('jupyter-notebook' | 'interactive')[]) {
-        const added: IVSCodeNotebookController[] = [];
+        const addedList: IVSCodeNotebookController[] = [];
         metadatas.forEach((metadata) => {
-            added.push(...this.addImpl(metadata, types, false));
+            const { added } = this.addImpl(metadata, types, false);
+            addedList.push(...added);
         });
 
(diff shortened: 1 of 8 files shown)


src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -82,26 +82,30 @@ export class ControllerRegistration implements IControllerRegistration {
...
     }
     addImpl(
         metadata: KernelConnectionMetadata,
         types: ('jupyter-notebook' | 'interactive')[],
         triggerChangeEvent: boolean
-    ): IVSCodeNotebookController[] {
-        let results: IVSCodeNotebookController[] = [];
+    ): { added: IVSCodeNotebookController[]; existing: IVSCodeNotebookController[] } {
+        const added: IVSCodeNotebookController[] = [];
+        const existing: IVSCodeNotebookController[] = [];
```
DonJayamanne, line 108: We were triggering change events even when we're updating existing items
This had other side effects, the change event only supports `added, removed`, not `updated`

src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -82,26 +82,30 @@ export class ControllerRegistration implements IControllerRegistration {
...
         metadatas.forEach((metadata) => {
-            added.push(...this.addImpl(metadata, types, false));
+            const { added } = this.addImpl(metadata, types, false);
+            addedList.push(...added);
         });
 
-        this.changeEmitter.fire({ added, removed: [] });
-        return added;
+        if (addedList.length) {
+            this.changeEmitter.fire({ added: addedList, removed: [] });
```
DonJayamanne, line 92: No more unnecessary change events, which had side effects

mjbvz (approved):

PR #12004, opened 2022-11-14
Display preferred kernel in new kernel quickpick

### LINKED ISSUE #11992: Display recommended Python environment in the Local Python  quick pick (use the star just like Python ext)

### LINKED ISSUE #12003: Python Environment creation using Python Extension Api

DonJayamanne:
Fixes #11992 

Screenshot 2022-11-14 at 14 59 22
Screenshot 2022-11-14 at 15 00 15

### DIFF
--- a/src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
+++ b/src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
@@ -119,11 +119,34 @@ type MultiStepResult = {
     connection?: KernelConnectionMetadata;
     disposables: IDisposable[];
 };
-
-function updateQuickPickWithNewItems(
-    quickPick: QuickPick<KernelSourceQuickPickItem | QuickPickItem>,
-    items: (KernelSourceQuickPickItem | QuickPickItem)[],
-    activeItem?: KernelSourceQuickPickItem | QuickPickItem
+function isKernelPick(item: ConnectionQuickPickItem | QuickPickItem): item is ConnectionQuickPickItem {
+    return 'connection' in item;
+}
+function updateKernelSourceQuickPickWithNewItems<T extends ConnectionQuickPickItem | QuickPickItem>(
+    quickPick: QuickPick<T>,
+    items: T[],
+    activeItem?: T
+) {
+    const activeItems = quickPick.activeItems.length ? [quickPick.activeItems[0]] : activeItem ? [activeItem] : [];
+    if (activeItems.length && !items.includes(activeItems[0])) {
+        const oldActiveItem = activeItems[0];
+        const newActiveKernelQuickPickItem =
+            isKernelPick(oldActiveItem) &&
+            items.find((item) => isKernelPick(item) && item.connection.id === oldActiveItem.connection.id);
+        // Find this same quick pick item.
+        if (newActiveKernelQuickPickItem) {
+            activeItems[0] = newActiveKernelQuickPickItem;
+        } else {
+            activeItems.length = 0;
+        }
+    }
+    quickPick.items = items;
+    quickPick.activeItems = activeItems;
+}
+function updateKernelQuickPickWithNewItems<T extends ConnectionQuickPickItem | QuickPickItem>(
+    quickPick: QuickPick<T>,
+    items: T[],
+    activeItem?: T
 ) {
     const activeItems = quickPick.activeItems.length ? [quickPick.activeItems[0]] : activeItem ? [activeItem] : [];
     quickPick.items = items;
@@ -238,7 +261,7 @@ export class NotebookKernelSourceSelector implements INotebookKernelSourceSelect
                     provider: p
                 });
             });
-            updateQuickPickWithNewItems(quickPick, items);
+            updateKernelSourceQuickPickWithNewItems(quickPick, items);
             quickPick.busy = false;
         })().ignoreErrors();
 
@@ -251,7 +274,7 @@ export class NotebookKernelSourceSelector implements INotebookKernelSourceSelect
             switch (selectedSource.type) {
                 case KernelSourceQuickPickType.LocalKernelSpec:
                 case KernelSourceQuickPickType.LocalPythonEnv:
-                    return this.selectKernelFromKernelFinder(selectedSource.kernelFinderInfo, state, token);
+                    return this.selectKernelFromLocalKernelFinder(selectedSource.kernelFinderInfo, state, token);
                 case KernelSourceQuickPickType.ServerUriProvider:
                     return this.getRemoteServersFromProvider.bind(this, selectedSource.provider, token);
                 default:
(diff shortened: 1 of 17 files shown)


src/notebooks/controllers/preferredKernelConnectionService.ts
```diff
@@ -0,0 +1,329 @@
...
+    RemoteKernelSpecConnectionMetadata
+} from '../../kernels/types';
+import { disposeAllDisposables } from '../../platform/common/helpers';
+import { IDisposable } from '../../platform/common/types';
+import { getNotebookMetadata, translateKernelLanguageToMonaco } from '../../platform/common/utils';
+import { IInterpreterService } from '../../platform/interpreter/contracts';
+import { ServiceContainer } from '../../platform/ioc/container';
+import { getLanguageOfNotebookDocument } from '../languages/helpers';
+
+/**
```
rebornix, line 23: I wonder if we can separate this change from "Create Python Env" to give us smaller PRs to review.

DonJayamanne: Thats how it started, then thought this would be small enough
Sure thing will do.

src/notebooks/controllers/preferredKernelConnectionService.ts
```diff
@@ -0,0 +1,329 @@
...
+ * Attempt to clean up https://github.com/microsoft/vscode-jupyter/issues/11914
+ * Provides the ability to look for exact or preferred kernel connections.
+ * Note: This class has zero side effects unlike `ControllerPreferredService`, and is meant to be a replacement for it
+ */
+export class PreferredKernelConnectionService {
+    private readonly disposables: IDisposable[] = [];
+    public dispose() {
+        disposeAllDisposables(this.disposables);
+    }
+    public async findExactRemoteKernelConnection(
```
DonJayamanne, line 33: @rebornix @Tyriar 
this is a very simple approach to not injecting everything
i personally think making everything as injectable and injecting all classes instead if instantiating then when we need them is not absolutely necessary, but that depends on whether one believes everything must be injected or not. i don't think we'd need to

Tyriar: Do you mean fetching the `kernelFinder`? Removing injection would end up making code less readable wouldn't it, reducing code quality in the end?

DonJayamanne: > Removing injection would end up making code less readable wouldn't it, reducing code quality in the end?

Not sure about readability, but we're still using a DI container, instead of injection we're using a service container to pull in the necessary dependencies as and when we need them.
Happy to change this back to an injectable.

DonJayamanne: I.e. we still use the DI container, but don't inject, instead use the service container to create/get what we need.

rebornix: This can help us with the startup since inversify doesn't need to resolve the dependency graph for it. Though as @Tyriar mentioned it can potentially make code less readable. For `IKernelFinder` using the DI to resolve the service other than injectable seems identical, for others I think we still need to go through them and see if they should just be a local instance other than a global service.

DonJayamanne: I'm not following, are you suggesting I change this?
If so, what's the difference between this approach and the use of `IInstantiationService.createInstance` in VS Code?
Just trying to understand the use case for the latter in VS Code, so I can follow the same pattern in this repo (without having to worry about any issues with maintainability).

src/platform/ioc/container.ts
```diff
@@ -21,7 +21,13 @@ try {
  */
 @injectable()
 export class ServiceContainer implements IServiceContainer {
-    constructor(private container: Container) {}
+    public static get instance(): IServiceContainer {
```
rebornix, line 24: :thumbsup: I'd like to use the service container in some cases, instead of having all injected in the ctor.

src/notebooks/controllers/preferredKernelConnectionService.ts
rebornix: @DonJayamanne sorry for the confusion. There are two scenarios:

In this case, `KernelFinder` might be used in mulitple places, so it makes good sense to register it to `ServiceContainer` but not `injectable`. And when we want to use it, we do `ServiceContainer.get`.

The other one is: Service A depends on service B, service B is only used in Service A and it's not required in `ctor`. Service B has its own service dependencies. When we need to use an instance of service B, then we would use `IInstantiationService.createInstance` to create an instance of service B with the right services dependencies.

With this approach, Service B is never registered on ServiceContainer, it won't be available to anyone else.

src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
```diff
@@ -108,11 +114,21 @@ interface ConnectionQuickPickItem extends QuickPickItem {
...
     connection?: KernelConnectionMetadata;
     disposables: IDisposable[];
 };
 
+function updateQuickPickWithNewItems(
+    quickPick: QuickPick<KernelSourceQuickPickItem | QuickPickItem>,
+    items: (KernelSourceQuickPickItem | QuickPickItem)[],
+    activeItem?: KernelSourceQuickPickItem | QuickPickItem
+) {
+    const activeItems = quickPick.activeItems.length ? [quickPick.activeItems[0]] : activeItem ? [activeItem] : [];
```
rebornix, line 128: Is it possible that `quickPick.activeItems` no longer exists in `items`?

rebornix: I did this recalculation of active items in core kernel picker https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/notebook/browser/contrib/editorStatusBar/editorStatusBar.ts#L226-L240

src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
```diff
@@ -433,24 +473,64 @@ export class NotebookKernelSourceSelector implements INotebookKernelSourceSelect
...
+                return recommended;
+            },
             refresh: () => source.refresh()
         };
         const disposable = source.onDidChangeKernels(() => onDidChange.fire());
         state.disposables.push(disposable);
         state.disposables.push(onDidChange);
+        state.disposables.push(onDidChangeRecommended);
+
+        if (
```
rebornix, line 500: Maybe this is a good time to split `selectKernelFromKernelFinder` to `selectKernelFromRemoteKernelFinder` and `selectKernelFromLocalKernelFinder`. It seems the local one is getting more complicated.

src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
```diff
@@ -534,23 +618,61 @@ export class NotebookKernelSourceSelector implements INotebookKernelSourceSelect
                 }
             }
         });
-        if (provider.status === 'discovering') {
-            quickPick.busy = true;
-        }
+        // Assume we're always busy.
+        quickPick.busy = true;
+        let noLongerBusy = true;
+        // If we don't get any updates after 1s, then hide the busy indicator
```
rebornix, line 624: I guess you mean 2s ;)

rebornix (commented):
Looks good to me in overall, a couple of questions only.

src/platform/ioc/container.ts
DonJayamanne: Leaving this for now, until we replace the DI conainer, thanks

src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
DonJayamanne: Done.

src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
```diff
@@ -534,23 +648,61 @@ export class NotebookKernelSourceSelector implements INotebookKernelSourceSelect
...
+                });
+            }
+            const recommendedItem = connectionToQuickPick(provider.recommended);
+            recommendedItem.label = `$(star-full) ${recommendedItem.label}`;
+            if (recommendedItems.length === 2) {
+                recommendedItems[1] = recommendedItem;
+            } else {
+                recommendedItems.push(recommendedItem);
+            }
+            updateKernelQuickPickWithNewItems(quickPick, recommendedItems.concat(quickPickItems), recommendedItems[1]);
```
rebornix, line 702: Do we want to use `updateKernelSourceQuickPickWithNewItems` here too?

rebornix (approved):
:thumbsup:

src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
DonJayamanne: Thats was a wrong function, not re-usable.

PR #12007, opened 2022-11-14
Add ability to create Python env from new Kernel QuickPick

### LINKED ISSUE #12003: Python Environment creation using Python Extension Api

DonJayamanne:
Fixes #12003 

# NOTE
Should be reviewed after https://github.com/microsoft/vscode-jupyter/pull/12004
### OR
Please review commit `Add create Python env in kernel picker` (ignoring the first commit, as thats the chagnes from https://github.com/microsoft/vscode-jupyter/pull/12004)

Work flow:
* User hits run cell
* They are presented with quick pick to select a kenel
* User selects the option to create a new env
* A new env is created
* The necessary dependencies are installed
* The cell is executed

I.e. user doesn't need to run cell again, it is one single smooth flow (though creation of env and installing packages takes a while).

Screenshot 2022-11-14 at 15 19 42

### DIFF
--- a/src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
+++ b/src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
@@ -52,10 +52,17 @@ import {
     IQuickPickParameters
 } from '../../../platform/common/utils/multiStepInput';
 import { IInterpreterService } from '../../../platform/interpreter/contracts';
+import { ServiceContainer } from '../../../platform/ioc/container';
 import { traceError } from '../../../platform/logging';
 import { ConnectionDisplayDataProvider } from '../connectionDisplayData';
 import { PreferredKernelConnectionService } from '../preferredKernelConnectionService';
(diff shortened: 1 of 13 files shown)


src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
```diff
@@ -627,13 +627,21 @@ export class NotebookKernelSourceSelector implements INotebookKernelSourceSelect
...
         };
 
+        const createPythonItems: (ConnectionQuickPickItem | QuickPickItem)[] = [];
+        let createPythonQuickPickItem: QuickPickItem | undefined;
+        if (this.extensionChecker.isPythonExtensionInstalled) {
+            createPythonQuickPickItem = {
+                label: `$(add) ${DataScience.createPythonEnvironmentInQuickPick()}`
+            };
+            createPythonItems.push(createPythonQuickPickItem);
+        }
```
rebornix, line 637: Not sure if we want to add a separator here.

rebornix (approved):

PR #12009, opened 2022-11-14
Fix flaky widget and IW tests

DonJayamanne:


src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -246,7 +246,7 @@ export class VSCodeNotebookController implements Disposable, IVSCodeNotebookCont
         traceVerbose(
             `Disposing controller ${this.id} associated with connection ${this.connection.id} ${
                 nbDocumentUris.length ? 'and documents ' + nbDocumentUris.join(', ') : ''
-            }`
+            } called from ${new Error('').stack}`
```
DonJayamanne, line 249: Leaving this here so we can see this in the future.

src/test/datascience/ipywidgets/ipyWidgetScriptManager.vscode.common.test.ts
```diff
@@ -66,6 +70,23 @@ suite('IPyWidget Script Manager @widgets', function () {
             disposables
         );
         const { notebook, editor } = await openNotebook(testWidgetNb);
+        if (!isWeb()) {
+            // Create the controller and select it for the tests.
```
DonJayamanne, line 74: Now that we're lazy loading controllers, we should create the controller for the tests,
this is much faster and more reliable, will also be required when we adopt the new kernel picker (as user will always have to select a kernel)

DonJayamanne: This fixes a flaky test with Widgets

src/test/datascience/notebook/remoteNotebookEditor.vscode.test.ts
```diff
@@ -180,7 +180,9 @@ suite('Remote Kernel Execution', function () {
             },
             defaultNotebookTestTimeout,
             () =>
-                `Should not have any remote controllers, existing ${JSON.stringify(controllerRegistration.registered)}`
+                `Should not have any remote controllers, existing ${controllerRegistration.registered
```
DonJayamanne, line 183: We ended up with a circular reference in attempting to serialize a class.

src/notebooks/controllers/controllerLoader.ts
```diff
@@ -126,6 +133,11 @@ export class ControllerLoader implements IControllerLoader, IExtensionSyncActiva
                 if (connectionIsNoLongerValid && controller.connection.kind === 'connectToLiveRemoteKernel') {
                     return true;
                 }
+
+                // Don't dispose this controller if it's attached to a document.
+                if (this.notebook.notebookDocuments.some((notebook) => controller.isAssociatedWithDocument(notebook))) {
+                    return false;
```
DonJayamanne, line 139: This fixes a flaky IW test

alexr00 (approved):

PR #12026, opened 2022-11-15
Changes to labels and display a controller

DonJayamanne:


src/notebooks/controllers/controllerPreferredService.ts
```diff
@@ -422,25 +420,6 @@ export class ControllerPreferredService implements IControllerPreferredService,
...
-        if (KernelPickerType.useNewKernelPicker) {
-            // With the new kernel picker always find exact matches when updating the affinity.
-            // Attempt to clean up https://github.com/microsoft/vscode-jupyter/issues/11914
-            // Here we actually need an exact match instead of close matches,
-            // For the new kernel picker we will always try to find an exact match, if not found, then we
-            // do not show anything in the VS Codes kernel picker.
-            const preferredService = new PreferredKernelConnectionService();
-            try {
-                if (serverId) {
-                    return preferredService.findExactRemoteKernelConnection(notebook, cancelToken);
```
DonJayamanne, line 434: No need to match exactly, fuzzy is good enough (this way we still have a controller dipslayed in vs code kernel picker)
Removing code path for the new kernel picker to make it behave the same as old (stable) kernel picker

src/notebooks/controllers/kernelSource/notebookKernelSourceSelector.ts
```diff
@@ -236,6 +224,19 @@ export class NotebookKernelSourceSelector implements INotebookKernelSourceSelect
             });
         }
 
+        const localKernelFinder = allKernelFinders.find(
```
DonJayamanne, line 227: Display local python env before localkernel specs (moved a block of code around)

TylerLeonhardt (approved):

Issue #11996, opened 2022-11-11
Investigate reducing bundle size


Tyriar:

Tyriar:
A profile when Jupyter activates is interesting, this is on a particularly fast activation on my fast machine:

![image](https://user-images.githubusercontent.com/2193314/202017071-7f3f172c-c8ed-46c0-9cfd-3013621808d2.png)

- `_loadCommonJSModule` (load jupyter code) 987.33ms
- `_callActivate` 50.59ms
- `activateLegacy` 206.24ms - Not sure what the legacy is about here?
  - Inversify init 165.33ms https://github.com/microsoft/vscode-jupyter/issues/12008
  - `activateSync` 10.20ms
  - `activate` 30.70ms
- `_loadCommonJSModule` (load python code) 126.5ms
- `activate` (python) 13.26ms 👏 
- `refreshEnvironments` 879.74ms - What's the spawn about? Is that really locking up the exthost just to read some registry keys?

Tyriar:
I'm thinking this for splitting strategy:

- We could create a tsconfig with project references to enforce a strict API to sub-components that can be lazily loaded like interactive-window, this could be done in a more naive way via layer eslint rules 
- Make all access to interactive window from the main bundle async
- Only allow importing interfaces or the strictly defined API from the main bundle. I think these are the main usages we need to worry about wrt the API
  ![image](https://user-images.githubusercontent.com/2193314/202019481-9d04ef78-2fea-46ea-b24b-eeaaf7b1c61d.png)
- Have a web and a node impl of the object that accesses the interactive window code, the node one would be async as dynamic imports work

DonJayamanne:
> `refreshEnvironments ` 879.74ms - What's the spawn about? Is that really locking up the exthost just to read some registry keys?

This should be in the Python extension, they end up spawning Python processes (its part of discovering python environments on the user machine).
One possible speed bump I've thought of in the past is using a worker process for spawning processes , but not sure tha'ts a viable solution or not (ther'es an issue for that).
@karrtikr  /cc

& I believe they do end up reading registry keys (at least at one point they did),

Again this is a delay in the Python extension.

karrtikr:
Once Jupyter switches to the new kernel picker, I believe 'refreshEnvironments' need not to called at startup, so this shouldn't be an issue. And yes we do block it on fetching registry keys atm.

Tyriar:
[...]

PR #12034, opened 2022-11-15
Avoid require when initializing logging

Tyriar:
This should be equivalent.

Part of https://github.com/microsoft/vscode-jupyter/issues/11996


rebornix (approved):
👍 I see no difference in compiled code.

PR #12035, opened 2022-11-15
Remove named-js-group

Tyriar:
uglify-js doesn't appear to be in the project anymore and it seems that was the only reason it was included based on a comment? Regardless, support is good for named capture group in all browsers/node now.

Part of #11996

image

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#browser_compatibility


Tyriar:
@types/uglify-js is actually used via @types/html-minifier-terser

DonJayamanne (approved):

rebornix (approved):

Issue #11745, opened 2022-10-22
Variables viewer causes slowness and instability


ramajoballester:
I am creating this issue because #8320 and #7069 did not solve the problem.

## Environment data

-   VS Code: Version: 1.72.2
Commit: d045a5eda657f4d7b676dedbfa7aab8207f8a075
Date: 2022-10-12T22:16:26.920Z
Electron: 19.0.17
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Linux x64 5.15.0-52-generic
Sandboxed: No
-   Jupyter Extension version: v2022.9.1202862440
-   Python Extension version: v2022.16.1
-   OS: Ubuntu 20.04.5 LTS (remote and local)
-   Python version: 3.10.6
-   Anaconda version: 22.9.0
-   Jupyter server running: Remote

## Expected behaviour

Normal and fast execution time, given the simplicity of the cell commands. Running the same jupyter kernel (remotely via ssh) in the browser executes every single cell instantaneously.

## Actual behaviour

Some cells take a lot of time to execute. I have experienced this behaviour recently after loading a pickle object, but it has happened with other different commands. The time it takes to complete is highly variable and non-repeatable. Memory usage is low.

Waiting for some time before running the next cell avoids this problem (see last cell that executes ```1+5``` in the expected time). Same in VSCode Insiders.

<img src="https://user-images.githubusercontent.com/47662166/197356530-c1a3ac28-f8b5-4826-b4da-7f63b6e40d11.png" width="500">
<img src="https://user-images.githubusercontent.com/47662166/197356721-82e112db-8a44-47f9-9a09-84e14cbefc52.png" width="500">

## Logs

<details>

<summary>With jupyter logging level to verbose. The log corresponds to the screenshots posted:
</summary>

<p>

```
debug 19:45:55.442: VSCodeNotebookController::handleExecution, Class name = x (started execution)
debug 19:45:55.443: Execute Cell 0 /home/messi/alvaro/loci-sensus/notebooks/first_try.ipynb
debug 19:45:55.444: start the kernel, options.disableUI=false
debug 19:45:55.445: Start Jupyter Session in kernel.ts with disableUI = false
debug 19:45:56.312: Cell 0 executed with state Success
debug 19:45:56.314: VSCodeNotebookController::handleExecution, Class name = x, completed in 872ms, has a falsy return value
[...]

PR #12042, opened 2022-11-16
Variables perf improvements

roblourens:
For #11745


roblourens:
Hey Don, I think you have the most context on Variables, please take a look and let me know if any of this is a bad idea. The commits can be reviewed individually

DonJayamanne (approved):

PR #12051, opened 2022-11-16
Add Kernel refresh indicator

DonJayamanne:


Jimimaku:
> _ No description provided. _

src/kernels/kernelRefreshIndicator.node.unit.test.ts
```diff
@@ -0,0 +1,230 @@
...
+import { IDisposable } from '../platform/common/types';
+import { KernelRefreshIndicator } from './kernelRefreshIndicator.node';
+import { IKernelFinder } from './types';
+import { mockedVSCodeNamespaces } from '../test/vscode-mock';
+import { IPythonExtensionChecker } from '../platform/api/types';
+import { IInterpreterService } from '../platform/interpreter/contracts';
+import { createDeferred } from '../platform/common/utils/async';
+import { InteractiveWindowView, JupyterNotebookView } from '../platform/common/constants';
+
+suite('Kernel Refresh Indicator (node)', () => {
```
Tyriar, line 17: Should this live in test/kernels?

src/kernels/kernelRefreshIndicator.web.unit.test.ts
```diff
@@ -0,0 +1,94 @@
...
+import { instance, mock, reset, verify, when } from 'ts-mockito';
+import { EventEmitter, NotebookControllerDetectionTask } from 'vscode';
+import { disposeAllDisposables } from '../platform/common/helpers';
+import { IDisposable } from '../platform/common/types';
+import { KernelRefreshIndicator } from './kernelRefreshIndicator.web';
+import { IKernelFinder } from './types';
+import { mockedVSCodeNamespaces } from '../test/vscode-mock';
+import { InteractiveWindowView, JupyterNotebookView } from '../platform/common/constants';
+
+suite('Kernel Refresh Indicator (web)', () => {
```
Tyriar, line 13: Should this live in test/kernels?

src/kernels/kernelRefreshIndicator.node.ts
```diff
@@ -0,0 +1,120 @@
...
+export class KernelRefreshIndicator implements IExtensionSyncActivationService {
+    private refreshedOnceBefore?: boolean;
+    private readonly disposables: IDisposable[] = [];
+    constructor(
+        @inject(IDisposableRegistry) disposables: IDisposableRegistry,
+        @inject(IPythonExtensionChecker) private readonly extensionChecker: IPythonExtensionChecker,
+        @inject(IInterpreterService) private readonly interpreterService: IInterpreterService,
+        @inject(IKernelFinder) private readonly kernelFinder: IKernelFinder
+    ) {
+        disposables.push(this);
```
Tyriar, line 29: Is this a normal pattern to pass in a disposable registry? What's the purpose?

src/kernels/kernelRefreshIndicator.node.ts
```diff
@@ -0,0 +1,120 @@
...
+import { IDisposable, IDisposableRegistry } from '../platform/common/types';
+import { IInterpreterService } from '../platform/interpreter/contracts';
+import { IKernelFinder } from './types';
+
+/**
+ * Ensures we refresh the list of Python environments upon opening a Notebook.
+ */
+@injectable()
+export class KernelRefreshIndicator implements IExtensionSyncActivationService {
+    private refreshedOnceBefore?: boolean;
```
Tyriar, line 21: nit: Better to init to false so narrow the type, V8 maybe be able to optimize a `false | true` where it could not a `false | true | undefined`

Tyriar (changes requested):

src/kernels/kernelRefreshIndicator.node.unit.test.ts
DonJayamanne: no, withbnew code we try to keep unit tests closev to the corresponding source file.

src/kernels/kernelRefreshIndicator.node.ts
DonJayamanne: look, i thought it was the opposite, i thought allocating space for a default value of false unnecessarily was less optimal

Tyriar: I think you're right wrt allocating space, it will likely deopt something in the class as its properties change after the ctor has run. That's getting a bit too unnecessarily deep though, the main reason I commented was for maintainability as I would expect a `?: boolean` to have 3 states

src/kernels/kernelRefreshIndicator.node.ts
DonJayamanne: this class is instantiated when the extension starts and it registers itself in to the list of disposables.
thus when the extension gets deactiavated this will get dispose


these are singletons, hence caller cannot dispose them untill the extension gets deactivated

Tyriar: Ah ok, we could perhaps move this all into activationManager.ts by checking `'dispose' in item`?

Tyriar (dismissed):

src/kernels/kernelRefreshIndicator.node.ts
DonJayamanne: yes that's the suggestion @rebornix had

DonJayamanne:
@Tyriar sorry had to rebase to get some fixes from the main branch

Tyriar (approved):

PR #12066, opened 2022-11-17
re-enable tests

amunger:
re-enabling tests after fixing https://github.com/microsoft/vscode/issues/166254


amunger:
this will need the next insiders build, 11/18/22

PR #12082, opened 2022-11-18
Remove redundant KernelPicker type setting

DonJayamanne:


package.json
```diff
@@ -1947,30 +1947,6 @@
                     "default": false,
                     "markdownDescription": "%jupyter.configuration.jupyter.enableExtendedKernelCompletions.markdownDescription%",
                     "scope": "machine"
-                },
-                "jupyter.showOnlyOneTypeOfKernel": {
```
DonJayamanne, line 1951: Old setting was never used, hence no point leaving this around

package.json
```diff
@@ -1947,30 +1947,6 @@
...
-                    "default": false,
-                    "description": "%jupyter.configuration.jupyter.showOnlyOneTypeOfKernel.markdownDescription%",
-                    "deprecationMessage": "%jupyter.configuration.jupyter.showOnlyOneTypeOfKernel.deprecationMessage%",
-                    "scope": "application",
-                    "tags": [
-                        "experimental"
-                    ]
-                },
-                "jupyter.experimental.kernelPickerType": {
-                    "type": "string",
```
DonJayamanne, line 1962: Removing this

src/platform/common/featureManager.ts
```diff
@@ -84,10 +84,7 @@ export class FeatureManager implements IFeaturesManager {
 
     private _updateFeatures() {
         const kernelPickerType =
-            this.workspace.getConfiguration('notebook.kernelPicker').get('type') === 'mru' ||
-            this.configService.getSettings().kernelPickerType === 'Insiders'
```
DonJayamanne, line 88: @rebornix didn't you want to use a single setting for this?
I don't see why we need both VS Code and Jupyter setting, if we want to use just one setting

rebornix: @DonJayamanne yes, we should only use one from VS Code.

rebornix (approved):

PR #12104, opened 2022-11-18
Ignore eslint in types folder

Tyriar:
This folder is for external typings so it shouldn't be linted


mjbvz (approved):

Jimimaku:
> This folder is for external typings so it shouldn't be linted

PR #12148, opened 2022-11-23
Delete ctrl if associated kernel is deleted

DonJayamanne:


src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -302,7 +341,7 @@ export class ControllerRegistration implements IControllerRegistration {
             // TODO: Don't hide controllers that are already associated with a notebook.
             // If we have a notebook opened and its using a kernel.
             // Else we end up killing the execution as well.
-            if (this.isFiltered(item.connection) && !this.isControllerAttachedToADocument(item)) {
+            if (this.isFiltered(item.connection) && this.canControllerBeDisposed(item)) {
```
DonJayamanne, line 344: This was incorrect, wonder how many users use the old filter UI, cant wait for that to be removed
too much stuff to manage due to kernel filters

rebornix (approved):

PR #12154, opened 2022-11-24
Support hot swapping between kernel  finders

DonJayamanne:
Will submit a separate PR to delete unnecessary controllers created with old kernle picker and ensure we handle hot swapping in other places


rebornix (approved):
It looks great!

PR #12157, opened 2022-11-24
Move commands to install python into quick picks

DonJayamanne:


src/notebooks/controllers/commands/installPythonControllerCommands.ts
```diff
@@ -162,11 +162,6 @@ export class InstallPythonControllerCommands implements IExtensionSingleActivati
         if (!this.extensionChecker.isPythonExtensionInstalled) {
             sendTelemetryEvent(Telemetry.PythonExtensionNotInstalled, undefined, { action: 'displayed' });
 
-            if (!(await this.shouldInstallExtensionPrompt())) {
```
DonJayamanne, line 165: No point prompting users to install an extension, VS Code already does this.
If we leave this, then users will get two prompts to install an extension.

src/notebooks/controllers/kernelSource/kernelSelector.ts
```diff
@@ -77,13 +95,39 @@ export class KernelSelector implements IDisposable {
...
+        };
+        this.installPythonItem = {
+            label: DataScience.installPythonTitle(),
+            command: async () => {
+                // Timeout as we want the quick pick to close before we start this process.
+                setTimeout(() => commands.executeCommand(Commands.InstallPythonViaKernelPicker).then(noop, noop));
+                throw new SomeOtherActionError();
+            }
+        };
+        this.installPythonExtension = {
```
DonJayamanne, line 110: Support:
* Install Python extension
* Install Python runtime
In the new kernel picker

These are commands displayed in the old (currently stable) version of the kernel picker ui.
We need to bring them into the new kernel picker as well.
All we're doing here is adding the items into the quick pick

andreamah (approved):

Issue #11774, opened 2022-10-25
Stepping into N times on the same statement


jrieken:
Testing #11768

*  "jupyter.debugJustMyCode": true,
* use the snippet below
* press step into on line 5
* :bug: you need to press step into a couple of times before it resumes on the next line

```py
import ipywidgets as widgets

print(12)

a = widgets.IntSlider(description='a')
a

```

roblourens:
Filed upstream https://github.com/microsoft/debugpy/issues/1104, not sure whether debugpy can do better or not

roblourens:
Reopening since I'll do a workaround on our end.

roblourens:
Real fix from debugpy is en route. Added a workaround on our end.

PR #12181, opened 2022-11-27
Workaround for #11774

roblourens:
This issue causes step-in with justMyCode enabled to stop multiple times on the same line.


DonJayamanne:
> This issue causes step-in with justMyCode enabled to stop multiple times on the same line.

Does this mean the change in this PR fixes the above issue, or causes this issue

roblourens:
This PR fixes that issue

roblourens:
Specifically, when using RBL after our change to set the environment variable needed for justMyCode, you sometimes have to click multiple times to get past a certain statement. This works around it by auto-stepping if we are stuck on the same line.

DonJayamanne (approved):

Issue #12190, opened 2022-11-28
Broken interactive window test 'Raising an exception from within a function has a stack trace'


DonJayamanne:
https://github.com/microsoft/vscode-jupyter/actions/runs/3565779955/jobs/5991373158
https://github.com/microsoft/vscode-jupyter/actions/runs/3568555079

DonJayamanne:
@amunger looks like the tests are failing consistently, hence looks more like a break in the code., 
You might want to look into this sooner than later, hence assigned to nov and added the bug label
Please re-adjust as you see fit

amunger:
I'll take a look, but it actually doesn't seem to have failed on main for a few weeks
![image](https://user-images.githubusercontent.com/2019016/204388943-1df0e683-de94-4c1f-b08c-4ae71b668eab.png)

amunger:
never mind, looks like the latest ipykernel version breaks our regex again

amunger:
verification steps:
create and save a python file
create a python environment and install ipykernel  6.18.1 (`pip install ipykernel==6.18.1)
run the file in the interactive window with something that will throw e.g. `print(x)`
ensure the stacktrace has links back to the errored line in the file

Issue #12189, opened 2022-11-28
Broken Interactive window test ' Error stack traces have correct line hrefs with mix of cell sources'


DonJayamanne:
https://github.com/microsoft/vscode-jupyter/actions/runs/3565779955/jobs/5991373158
https://github.com/microsoft/vscode-jupyter/actions/runs/3568555079

DonJayamanne:
@amunger looks like the tests are failing consistently, hence looks more like a break in the code., 
You might want to look into this sooner than later, hence assigned to nov and added the bug label
Please re-adjust as you see fit

amunger:
same cause as https://github.com/microsoft/vscode-jupyter/issues/12190

PR #12197, opened 2022-11-28
update regex for new traceback format

amunger:
for #12189
for #12190

the new ipykernel version provides a slightly different stack trace format


DonJayamanne (commented):
Can we write some unit tests
These tests are great but will only be useful with the latest ipykernel
hence if we were to break the regex for a previous version we'd never know until a user reports this or we find out some other way its broken

amunger:
good call: https://github.com/microsoft/vscode-jupyter/issues/12199

lramos15 (approved):

PR #12256, opened 2022-12-01
Return empty item when Uri points to local

DonJayamanne:


src/kernels/jupyter/launcher/serverUriStorage.ts
```diff
@@ -269,7 +269,9 @@ export class JupyterServerUriStorage implements IJupyterServerUriStorage {
         try {
             const uri = await this.getUri();
             traceInfoIfCI(`getRemoteUri: ${uri?.uri}`);
-
+            if (uri?.uri === Settings.JupyterServerLocalLaunch) {
```
DonJayamanne, line 272: @rebornix looks like this was missed as part of some refactoring.

PR #12347, opened 2022-12-07
Bump engine version

roblourens:


daviddossett (approved):

package.json
```diff
@@ -29,7 +29,7 @@
         "theme": "light"
     },
     "engines": {
-        "vscode": "^1.74.0-insider"
+        "vscode": "^1.75.0"
```
DonJayamanne, line 32: Shouldn't this be 1.75.0-insider

DonJayamanne (changes requested):
I think the engine version is incorrect

package.json
roblourens: No, we decided in Slack that `-insider` doesn't do anything.

DonJayamanne: Oh it doesn't, ok, i ididn't know that.

PR #12372, opened 2022-12-09
Add "delete temp debugging files" to @mandatory

### LINKED ISSUE #10041: Fill test hole for checking that dumpCell debugging files are deleted + re-enable debugging tests
Doing the work for https://github.com/microsoft/vscode-jupyter/issues/9973

I believe that we have a test hole for not checking to see if we delete the .py files we create during debugging. Also it looks like a number of the tests in debugger.vscode.test.ts are disabled or skipped. I'd like to add a test for temp file deletion and also revisit the skipped test as part of the work for #9973.

roblourens:
to ensure that it runs on Windows and also so that we have a debugging test in @mandatory
Fix #10041

### DIFF
--- a/src/test/datascience/debugger.vscode.test.ts
+++ b/src/test/datascience/debugger.vscode.test.ts
@@ -98,7 +98,7 @@ suite('Run By Line @debugger', function () {
     // Cleanup after suite is finished
     suiteTeardown(() => closeNotebooksAndCleanUpAfterTests(disposables));
 
-    test('Delete temp debugging files', async function () {
+    test('Delete temp debugging files @mandatory', async function () {
         let tempWatcher;
         let folderName;
         try {


Tyriar (approved):

src/test/datascience/debugger.vscode.test.ts
```diff
@@ -98,7 +98,7 @@ suite('Run By Line @debugger', function () {
     // Cleanup after suite is finished
     suiteTeardown(() => closeNotebooksAndCleanUpAfterTests(disposables));
 
-    test('Delete temp debugging files', async function () {
+    test('Delete temp debugging files @mandatory', async function () {
```
Tyriar, line 101: If you're parsing this in CI or something, it might be nicer if you had a util to make it more strongly typed? `testMandatory(...)`

roblourens: cc @DonJayamanne

Tyriar: https://github.com/microsoft/vscode-jupyter/issues/12373

PR #12382, opened 2022-12-10
Strip ansi escape codes from jupyter 'inspect' result before parsing so regexes can be stricter and faster

### LINKED ISSUE #11799: VSCode Crashes when assigning list[many tuples] as a default function parameter (only when checking variables tab)
## Environment data

-   VS Code version: Version: 1.72.2
-   Jupyter Extension version (available under the Extensions sidebar): v2022.10.1103001107
-   Python Extension version (available under the Extensions sidebar): v2022.17.12981046
-   OS: Linux Mint 21 x64 5.15.0-52-generic (also happening on Windows)
-   Python and/or Anaconda version: Python 3.10.6
-   Type of virtual environment used: venv (also happening with miniconda)

## Steps to reproduce:
```python
# run this block first
list_of_tuples = [("1","2","3", "4", "5") for i in range(350)]

# then this random function 
def test(argument = list_of_tuples):
    return 0
```
Open variables menu, vscode freezes, vscode cpu process stays at 12% constantly, nothing can be executed anymore. I think this is a major bug.

## Logs

<details>

<p>

```
info 22:10:56.524: Loading webview. View is notset
info 22:10:56.525: Loading web view...
info 22:10:56.526: Webview panel created.
info 22:10:56.761: Web view react rendered
```

</p>
</details>

roblourens:
Fix #11799

### DIFF
--- a/src/kernels/variables/kernelVariables.ts
+++ b/src/kernels/variables/kernelVariables.ts
@@ -8,6 +8,7 @@ import { CancellationError, CancellationToken, Event, EventEmitter } from 'vscod
 import { Identifiers, PYTHON_LANGUAGE } from '../../platform/common/constants';
 import { IConfigurationService, IDisposableRegistry } from '../../platform/common/types';
 import { createDeferred } from '../../platform/common/utils/async';
+import { stripAnsi } from '../../platform/common/utils/regexp';
 import { getKernelConnectionLanguage, isPythonKernelConnection } from '../helpers';
 import { IKernel, IKernelConnectionSession, IKernelProvider } from '../types';
 import {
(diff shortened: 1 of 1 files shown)


src/kernels/variables/kernelVariables.ts
```diff
@@ -335,7 +336,7 @@ export class KernelVariables implements IJupyterVariables {
             // Should be a text/plain inside of it (at least IPython does this)
             if (output && output.hasOwnProperty('text/plain')) {
                 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                const text = (output as any)['text/plain'].toString() as string;
+                const text = stripAnsi((output as any)['text/plain'].toString() as string);
```
DonJayamanne, line 339: do we still need `as string` when we use `.toString()`

roblourens: No, we don't even need the `any`

joyceerhl (approved):

PR #12428, opened 2022-12-21
Determine whether users use create Env feature

DonJayamanne:
Added some telemetry

### DIFF
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -2918,6 +2918,44 @@ export class IEventNamePropertyMapping {
         }
     };
     /**
+     * Telemetry sent when user attempts to create a Python env via the kernel picker.
+     */
+    [Telemetry.CreatePythonEnvironment]: TelemetryEventInfo<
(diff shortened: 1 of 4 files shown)


src/telemetry.ts
```diff
@@ -2906,6 +2906,44 @@ export class IEventNamePropertyMapping {
...
+              reason: 'cancelled' | 'kernelConnectionNotCreated';
+          }
+        | /* Creation succeeded */ {
+              /**
+               * Whether the kernel dependencies were installed or not.
+               */
+              dependenciesInstalled: boolean;
+          }
+    > = {
+        owner: 'IanMatthewHuff',
```
Tyriar, line 2932: This should change

Tyriar (changes requested):

src/telemetry.ts
DonJayamanne: Good catch, updated to assign myself to other items as well.

Tyriar (approved):

PR #12446, opened 2022-12-27
Support remote jupyter server working directory mapping

rebornix:
Allow already synced file to be used instead of creating backup file.

### DIFF
--- a/src/kernels/jupyter/session/jupyterSession.ts
+++ b/src/kernels/jupyter/session/jupyterSession.ts
@@ -22,11 +22,12 @@ import {
     IJupyterConnection,
     ISessionWithSocket,
     KernelActionSource,
-    IJupyterKernelConnectionSession
+    IJupyterKernelConnectionSession,
+    isRemoteConnection
 } from '../../types';
 import { DisplayOptions } from '../../displayOptions';
 import { IBackupFile, IJupyterBackingFileCreator, IJupyterKernelService, IJupyterRequestCreator } from '../types';
-import { CancellationError, Uri } from 'vscode';
+import { CancellationError, Uri, workspace } from 'vscode';
 import { generateBackingIPyNbFileName } from './backingFileCreator.base';
 import { noop } from '../../../platform/common/utils/misc';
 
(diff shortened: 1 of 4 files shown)


src/kernels/jupyter/session/jupyterSession.ts
```diff
@@ -240,14 +241,28 @@ export class JupyterSession extends BaseJupyterSession implements IJupyterKernel
...
+            // Create our backing file for the notebook
+            backingFile = await this.backingFileCreator.createBackingFile(
+                this.resource,
+                this.workingDirectory,
+                this.kernelConnectionMetadata,
+                this.connInfo,
+                this.contentsManager
+            );
+            remoteFilePath = backingFile?.filePath;
+        }
```
DonJayamanne, line 265: I'd change the code to ensure we always have a value for `remoteFilepath` which matches closely to the previous code.

```suggestion
        if (isRemoteConnection(this.kernelConnectionMetadata) && this.connInfo.workingDirectory && this.resource) {
            const currentWorkingDirectory = workspace.getWorkspaceFolder(Uri.from(this.resource));
            if (currentWorkingDirectory) {
                remoteFilePath = this.resource.path.replace(
                    currentWorkingDirectory.uri.path,
                    this.connInfo.workingDirectory
                );
            }
        } 
        if (!remoteFilePath) {
            // Create our backing file for the notebook
            backingFile = await this.backingFileCreator.createBackingFile(
                this.resource,
                this.workingDirectory,
                this.kernelConnectionMetadata,
                this.connInfo,
                this.contentsManager
            );
            remoteFilePath = backingFile?.filePath;
        }
```

src/kernels/types.ts
```diff
@@ -511,6 +511,7 @@ export interface IJupyterConnection extends Disposable {
     readonly url: string;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     getAuthHeader?(): any; // Snould be a json object
+    readonly workingDirectory?: string;
```
DonJayamanne, line 514: Some doc string would be good to explain what this is,

DonJayamanne: @rebornix should this be a Uri? I think that would be better, its more explicit that we're dealing with paths.
Also instead of making this a readonly property woudln't this better be served as a function?

E.g. what if the local machine is Mac and the remote machine is Windows, at this point a simple string replace might not work well, and its best to get a path constructed from the extension author instead .

rebornix: @DonJayamanne it's not exactly a Uri, but more or less a folder/path prefix as we will create at the end is a relative path to the server root. If I understand correctly, when we are creating sessions with a path, we are not really pointing to an absolute path on the server (as the root/pwd in which the server runs is unknown to the client) but we are creating relative paths like `a.ipynb` or `folder/a.ipynb` and the jupyter server will resolve it to a real remote file path on disk.

If the server is started in the synced workspace folder, then `workingDirectory` is not even needed as all we need to do is `resource.fsPath.replace(${workspaceFolder}, '')` (the `workspaceFolder` is local path). However it's possible that a server is started in an ancestor folder of the synced workspace, thus the prefix is needed.

> what if the local machine is Mac and the remote machine is Windows

I'm not sure if the session path is always a normalized unix path (as it's a REST request), we can double check Jupyter's implementation.

DonJayamanne: > I'm not sure if the session path is always a normalized unix path (as it's a REST request), we can double check Jupyter's implementation.

Sorry my concern is the opposite, if the extension author works on windows and the remote ends up being a unix. then extension author mgith always send windows path like `xyz\abc`

I think we'll need to revise this API, feels very much like a kneejerk reaction, i understand you need it for your extension, and lets go with this, and start reviewing the API in a another sprint

DonJayamanne (approved):

PR #12451, opened 2022-12-28
Increase test timeout

DonJayamanne:


src/test/datascience/notebook/helper.ts
```diff
@@ -597,18 +598,21 @@ async function getActiveInterpreterKernelConnection() {
...
+    const stopWatch = new StopWatch();
     return waitForCondition(
         () =>
             kernelFinder.kernels.find(
                 (item) =>
                     item.kind === 'startUsingPythonInterpreter' &&
                     areInterpreterPathsSame(item.interpreter.uri, interpreter.uri)
             ) as PythonKernelConnectionMetadata,
-        defaultNotebookTestTimeout,
+        defaultNotebookTestTimeout * 2,
```
Tyriar, line 609: 2 minutes is an enormous timeout, this will increase CI time when failures occur?

Issue #11422, opened 2022-09-21
Check setting scopes


roblourens:
There are a bunch of settings that have `"scope": "machine"`, for example, the log level: https://github.com/microsoft/vscode-jupyter/blob/515ad59e73cb734ae2066c286ef964c6efd60b31/package.json#L1518-L1532

This means that if I configure these in my user settings, then open a Remote-SSH window, those settings won't apply. I don't think `"machine"` is right for that, I think that the log level is a user preference which should carry across between machines. `"window"` would be the right scope for that.

https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema:~:text=machine%20%2D%20Machine%20specific%20settings%20that%20can%20be%20set%20only%20in%20user%20settings%20or%20only%20in%20remote%20settings.%20For%20example%2C%20an%20installation%20path%20which%20shouldn%27t%20be%20shared%20across%20machines.

There are about 10 machine-scope settings, I'm not sure any of them should be machine-scoped.

DonJayamanne:
> configure these in my user settings, then open a Remote-SSH window, those settings won't apply. I don't think "machine" i

i would have thought users would only change this when they need to diagnose a problem, on a specific env. not something they want to carry with them wherever they go.

roblourens:
I guess I want to run with that always enabled, but I'm not typical. You're probably right I don't think that's something we need to decide for the user. And I don't think that's true for the other settings.

roblourens:
I'm a little suspicious of a lot of the setting scopes. 

- I see the `machine` scope on a lot of settings that seem like they are essentially user preferences, and should be able to be set in user settings
- The `application` scope is used on a lot of settings that seem like they would work as `window` scope or even `resource` scope
- `resource` scope is used on some settings, and I don't know what differentiates those from the others

We should use the most permissive scope reasonable for each setting, so let's check the scopes for all settings

Please check over the scopes for settings in your area

- [ ] "jupyter.experiments.enabled"
- [ ] "jupyter.showVariableViewWhenDebugging"
- [ ] "jupyter.experiments.optInto"
- [ ] "jupyter.experiments.optOutFrom"
- [ ] "jupyter.widgetScriptSources"
- [ ] "jupyter.askForLargeDataFrames"
- [ ] "jupyter.askForKernelRestart"
- [ ] "jupyter.exportWithOutputEnabled"
- [ ] "jupyter.jupyterLaunchTimeout"
[...]

PR #12457, opened 2022-12-29
Fix settings scope

DonJayamanne:
For https://github.com/microsoft/vscode-jupyter/issues/11422


package.json
```diff
@@ -1561,25 +1554,24 @@
                     "type": "number",
                     "default": 60000,
                     "description": "%jupyter.configuration.jupyter.jupyterLaunchTimeout.description%",
-                    "scope": "resource"
+                    "scope": "application"
```
DonJayamanne, line 1557: I don't see why this would be different for each workspace. hence changed to be per a user setting

package.json
```diff
@@ -1561,25 +1554,24 @@
...
                     "description": "%jupyter.configuration.jupyter.jupyterLaunchRetries.description%",
-                    "scope": "resource"
+                    "scope": "application"
                 },
                 "jupyter.jupyterCommandLineArguments": {
                     "type": "array",
                     "default": [],
                     "description": "%jupyter.configuration.jupyter.jupyterCommandLineArguments.description%",
-                    "scope": "resource"
+                    "scope": "machine"
```
DonJayamanne, line 1569: This should not have been added to being with, this only applies in the case of non-zmq, which means this works on some machines and not others.
Also tomorrow if we ship zmq for all machines, then this property automatically becomes redunant and some feature that users might expect to work will not work 

Will file an issue to deprecate this

package.json
```diff
@@ -1537,7 +1530,7 @@
                     },
                     "uniqueItems": true,
                     "markdownDescription": "%jupyter.configuration.jupyter.widgetScriptSources.markdownDescription%",
-                    "scope": "machine"
+                    "scope": "application"
```
roblourens, line 1533: Seems like it would make sense as a workspace setting, since you could have widgets from different sources?

package.json
```diff
@@ -1549,7 +1542,7 @@
                     "type": "boolean",
                     "default": true,
                     "description": "%jupyter.configuration.jupyter.askForKernelRestart.description%",
-                    "scope": "resource"
+                    "scope": "application"
```
roblourens, line 1545: Also seems like it might make sense as a workspace setting

roblourens (commented):
It seems like some of these "application" scope settings could work as workspace settings, and if we move them to that scope, we could break them for users who have already configured them that way in their workspace. Is that worth it?

package.json
DonJayamanne: i don't see why one would have widgets come from a cdn for one and not the other.

package.json
DonJayamanne: i don't see why one would want to get prompted to restart kernels in one workspace and not the other. 

agreed existing data should be moved to application level of we change this

DonJayamanne:
> we move them to that scope, we could break them for users who have already configured them that way in their

agreed, perhaps we should migrate them to user settings if and when we come across them.

package.json
roblourens: But someone might want to do that, and if it's working fine, why restrict it? Could be a team preference that they want to enforce, could be that someone has a different workflow in two projects and just finds that useful.

roblourens:
> agreed, perhaps we should migrate them to user settings if and when we come across them.

I think we should avoid writing new settings to the user's settings, and here it will change the behavior in other workspaces. Do you want to write a different value to the user's settings every time they open a new workspace where it's configured differently?

DonJayamanne:
@roblourens I've updated the PR to 
* change from machine to Application
* if we have `resource`, then remove it so that its setup as the default `window`

Will revisit the settings that have `window/resource`, personally I think most of them just need to be application/machine instead of window/resource. But we can revisit that later based on the usage telemetry & based on the results possibly migrate them to the user settings file.

package.json
DonJayamanne: Got it, thanks

roblourens (dismissed):

DonJayamanne:
@roblourens I had to resolve a merge conflict

roblourens (approved):

Issue #9464, opened 2022-03-24
Conda install on CI can take a really long time


rchiodo:
![image](https://user-images.githubusercontent.com/19672699/159819533-32296ff8-8ca3-4b4c-8995-4cc1fa84df96.png)

One possible solution is to:
- Install conda normally (so the product works)
- Install micromamba
- Run micromamba to create envs

rchiodo:
Seems micromamba envs are not visible by conda except by path. You can install stuff into them, but you can't reference them by name. Might have to do something hacky to force conda to recognize them.

PR #12460, opened 2022-12-29
Use micromamba instead of miniconda for conda env tests

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/9464
Previously the conda CI job would take 10 minutes to setup Conda, now that same step takes <2 minutes.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -349,7 +349,7 @@ jobs:
           python-version: ${{matrix.pythonVersion}}
 
       - name: Install Conda environment from environment.yml
-        uses: mamba-org/provision-with-micromamba@v14
+        uses: mamba-org/provision-with-micromamba@f347426e5745fe3dfc13ec5baf20496990d0281f
         if: matrix.python == 'conda'
         with:
           cache-downloads: true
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test.yml
```diff
@@ -348,23 +348,15 @@ jobs:
         with:
           python-version: ${{matrix.pythonVersion}}
 
-      - name: Cache conda on linux
-        uses: actions/cache@v2
+      - name: Install Conda environment from environment.yml
+        uses: mamba-org/provision-with-micromamba@main
```
Tyriar, line 352: We want CI to be repeatable so pointing at main isn't a good idea. Also branches aren't the best as they're mutable, it's best to point at commits for third party actions.

Tyriar (changes requested):

.github/workflows/build-test.yml
DonJayamanne: Done

.github/workflows/build-test.yml
```diff
@@ -348,23 +348,15 @@ jobs:
         with:
           python-version: ${{matrix.pythonVersion}}
 
-      - name: Cache conda on linux
-        uses: actions/cache@v2
+      - name: Install Conda environment from environment.yml
+        uses: mamba-org/provision-with-micromamba
```
Tyriar, line 352: I meant uses: `mamba-org/provision-with-micromamba#<commit>`, or whatever the syntax is

Tyriar (changes requested):

Tyriar (approved):

PR #12466, opened 2022-12-29
Deprecated setting jupyterCommandLineArguments

### LINKED ISSUE #12463: Deprecate setting `jupyterCommandLineArguments`
This setting assumes that we always use Jupyter to start kernels.
However this is not the case, only when zmq is not available do we use Jupyter.
Hence any user relying on this setting would see inconsistent behaviour across machines or when zmq is shipped for other machines as well.

Its surprising that users haven't complained about this, either only a handful (or none at all) of users use this or users know how to turn off zmq (& the latter isn't supported).

**Solution:**
* Display warning message if this setting is used (prompt?)
* Remove the setting but keep the code
* Check if we have users using this setting.

DonJayamanne:
For #12463


package.json
```diff
@@ -1569,12 +1557,6 @@
                     "description": "%jupyter.configuration.jupyter.jupyterLaunchRetries.description%",
                     "scope": "resource"
                 },
-                "jupyter.jupyterCommandLineArguments": {
```
DonJayamanne, line 1572: Based on telemetry this is not used, hence this is being deprecated, will remove the code that uses this setting in a later release.

aeschli (approved):

Issue #11648, opened 2022-10-13
Remove `loaded` property from IController


DonJayamanne:
Now that all controllers are lazily loaded, we should remove the `loaded` property from the IController.
Ideally just awaiting on an event for refreshed or awaiting on controllers to get regisgtered (checking the count of controllers in `IControllerRegistration` might be a simpler appropach).

DonJayamanne:
Done

DonJayamanne:
Blocked on https://github.com/microsoft/vscode-jupyter/issues/13414
However since this is only used in one place now, we can close this issue.

PR #12477, opened 2023-01-03
Merge loader and registration interfaces

DonJayamanne:
Part of #11648


src/notebooks/controllers/controllerRegistration.ts
```diff
@@ -82,22 +94,261 @@ export class ControllerRegistration implements IControllerRegistration {
...
+    private loadControllers() {
+        this.controllersPromise = this.loadControllersImpl();
+        sendKernelListTelemetry(this.registered.map((v) => v.connection));
+
+        traceInfoIfCI(`Providing notebook controllers with length ${this.registered.length}.`);
+    }
+    public get loaded() {
+        return this.controllersPromise;
+    }
+    private async onDidOpenNotebookDocument(document: NotebookDocument) {
```
DonJayamanne, line 176: Brought all of this from another class.
Most of the code is very closely related, hence merged into a single class for easier maintenance
Besides a lot of this code will get deleted once we ship the new kernel picker.

aeschli (approved):

PR #12517, opened 2023-01-09
Display kernelSpec src only if kernels are found

### LINKED ISSUE #12224: Local Kernel Spec is a dead end
Testing #11963

* selected "Select Another Kernel > Local Kernel Spec"
* I see an empty QP with nothing

I honest don't really know "Local Kernel Spec" is all about but I'd suggest to (1) don't give me that option to begin with or (2) show a message why this is empty and how I can populate it

Screenshot 2022-11-30 at 11 36 23

DonJayamanne:
Fixes #12224


src/notebooks/controllers/kernelSource/kernelSourceCommandHandler.ts
```diff
@@ -76,16 +74,47 @@ export class KernelSourceCommandHandler implements IExtensionSyncActivationServi
...
+                                        command: 'jupyter.kernel.selectLocalKernelSpec'
+                                    }
+                                ];
+                            }
+                        })
+                    );
+                }
+            };
+            registerKernelSpecsSource();
+            this.kernelFinder.onDidChangeKernels(() => registerKernelSpecsSource(), this, this.localDisposables);
```
Tyriar, line 117: Does this need to be added to `localDisposables`?

Tyriar (changes requested):

Tyriar (dismissed):

src/notebooks/controllers/kernelSource/kernelSourceCommandHandler.ts
DonJayamanne: Yes, we want to be able to dispose these later in line 42.

rebornix (approved):

PR #12559, opened 2023-01-17
Handle failure to fetch remote kernel specs

### LINKED ISSUE #12203: cell stuck in running for non-existent remote connection
Testing #11963

enable mru kernel selection
create a jupyter server outside of vscode
connect to a kernel on the remote server and run a cell
kill the server and reload vscode
run a cell, this will open the kernel selection dialog
select the existing connection to the remote server

:bug: cell is in running state forever

DonJayamanne:
Fixes #12203

### DIFF
--- a/src/kernels/internalTypes.ts
+++ b/src/kernels/internalTypes.ts
@@ -15,6 +15,11 @@ export enum ContributedKernelFinderKind {
 export interface IContributedKernelFinder<T extends KernelConnectionMetadata = KernelConnectionMetadata> {
     status: 'discovering' | 'idle';
     onDidChangeStatus: Event<void>;
+    /**
+     * Last error thrown when listing the kernels.
+     * Use this property to determine if there was an error fetching kernels when there are no kernels listed.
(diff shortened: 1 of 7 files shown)


src/kernels/internalTypes.ts
```diff
@@ -15,6 +15,11 @@ export enum ContributedKernelFinderKind {
 export interface IContributedKernelFinder<T extends KernelConnectionMetadata = KernelConnectionMetadata> {
     status: 'discovering' | 'idle';
     onDidChangeStatus: Event<void>;
+    /**
+     * Last error thrown when listing the kernels.
+     * Use this property to determine if there was an error fetching kernels when there are no kernels listed.
+     */
+    lastError?: Error;
```
roblourens, line 22: I don't know the full flow here, but this pattern seems a little suspicious or like it could lead to async issues- like maybe the error should be associated with a particular call to a method or an event.

roblourens (approved):

Issue #7683, opened 2021-09-27
Add dropdown contribution to file editor `Run` icon for `Run file in Interactive window`


DonJayamanne:
Python extension is now using a new run icon, see below

We should contribute to this (if possible) for the `Run current file in interactive window`
This will improve discoverability of `Interactive window`

![Screen Shot 2021-09-27 at 13 23 01](https://user-images.githubusercontent.com/1948812/134979859-200b5d5a-ee54-4ce1-9751-2ac481aefd8d.png)

PR #12617, opened 2023-01-23
title run menu contribution to run file in IW

amunger:
Fixes https://github.com/microsoft/vscode-jupyter/issues/7683

add a launch contribution to run a python file in the interactive window.

The title should be changed to match up with Python's wording better, but the title only seems to be pulled from where the command is contributed, not the specific menu item contribution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -527,6 +527,7 @@
             {
                 "command": "jupyter.runFileInteractive",
                 "title": "%jupyter.command.jupyter.runFileInteractive.title%",
+                "icon": "$(play)",
                 "category": "Jupyter",
                 "enablement": "isWorkspaceTrusted"
             },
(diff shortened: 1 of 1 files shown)


amunger:
image

package.json
```diff
@@ -1463,6 +1464,13 @@
                     "group": "1_view",
                     "when": "debugProtocolVariableMenuContext == 'viewableInDataViewer'"
                 }
+            ],
+            "editor/title/run": [
+                {
+                    "command": "jupyter.runFileInteractive",
+                    "group": "jupyter",
+                    "when": "resourceLangId == python && !isInDiffEditor && !notebookEditorFocused && isWorkspaceTrusted"
```
roblourens, line 1472: So this can work for any python file, even if it doesn't have `#%%`?

roblourens: And is there a debug option?

roblourens (approved):

package.json
amunger: correct, cells are not required.
No current debug option - debugFileInteractive has very low usage relative to runFile or debugCell, though perhaps there is an argument for discoverability to be made.

PR #12694, opened 2023-01-30
parse the URI to get a platform compatible string

amunger:
Tests were failing locally on Windows with invalid JSON for the notebook contents.
A path URI was being written directly to the file, but failed on windows with unescaped file separators.


DonJayamanne (approved):

DonJayamanne (changes requested):
Not sure why we need this way, sorry should not have approved

DonJayamanne (approved):
> A path URI was being written directly to the file, but failed on windows with unescaped file separators.

This is a tricky one, we cannot assume that the ID will always contain File Paths,
However give that this is just for tests it should be fine

PR #12716, opened 2023-02-01
Stop using the Python API to get custom env vars

### LINKED ISSUE #12714: Starting a Jupyter server kernel fails (when zmq does not work)
* Use stable 1.74
* Use stable jupyter and python extension
* Ensure zmq does not work
* Use a virtual env to start up jupyter
* Ensure jupyter is installed into the above virtual env
* Ensure Jupyter is uninstalled from the global python env (which ever is in path)
* Run a cell,
This works in stable, this does not work in the latest bits

This is because the Python extension API is used to resolve the custom env variables from the .env file and even if .env does not exist we get env variables and we always get values for PATH and other such env variables which override the users python env variables, & that is wrong.

I.e. PATH in virtual env will be `.venv/bin,$PATH` and the PATH variable returned by python extension will be non-empty and will be the default PATH value. We always override virtual env variables with the custom env variables, thus PATH from python extension trumps,
As a result of this Python runtime will have an invalid PATH value and will then not be able to load the packages.

DonJayamanne:
Fixes #12714

This PR basically reverts a PR that makes use of the Python API to get the env variables.

### DIFF
--- a/src/platform/common/variables/customEnvironmentVariablesProvider.node.ts
+++ b/src/platform/common/variables/customEnvironmentVariablesProvider.node.ts
@@ -3,7 +3,7 @@
 
 import { inject, injectable, optional } from 'inversify';
 import * as path from '../../vscode-path/path';
-import { Disposable, Event, EventEmitter, RelativePattern, Uri } from 'vscode';
+import { ConfigurationChangeEvent, Disposable, Event, EventEmitter, RelativePattern, Uri } from 'vscode';
 import { TraceOptions } from '../../logging/types';
 import { sendFileCreationTelemetry } from '../../telemetry/envFileTelemetry.node';
 import { IWorkspaceService } from '../application/types';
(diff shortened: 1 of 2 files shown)


src/platform/common/variables/customEnvironmentVariablesProvider.node.ts
```diff
@@ -81,30 +78,11 @@ export class CustomEnvironmentVariablesProvider implements ICustomEnvironmentVar
             : this.workspaceService.workspaceFolders?.length
             ? this.workspaceService.workspaceFolders[0].uri
             : undefined;
-        const workspaceFolderUri = this.getWorkspaceFolderUri(resource);
-        if (!workspaceFolderUri) {
-            traceInfoIfCI(`No workspace folder found for ${resource ? resource.fsPath : '<No Resource>'}`);
-            return;
-        }
 
-        if (purpose === 'RunPythonCode' && this.extensionChecker.isPythonExtensionInstalled) {
```
DonJayamanne, line 90: We no longer rely on Python API to get the env variables.

rebornix (approved):

PR #12720, opened 2023-02-02
Revert "Stop using the Python API to get custom env vars (#12716)"

### LINKED ISSUE #12714: Starting a Jupyter server kernel fails (when zmq does not work)

DonJayamanne:
This reverts commit 746bacbde9fcebb73e5e8629f1b7b196bbad85be.

Related to #12714
Partially addresses https://github.com/microsoft/vscode-jupyter/issues/12805

### DIFF
--- a/src/kernels/raw/launcher/kernelEnvVarsService.node.ts
+++ b/src/kernels/raw/launcher/kernelEnvVarsService.node.ts
@@ -98,7 +98,7 @@ export class KernelEnvironmentVariablesService {
 
         if (isPythonKernel || interpreter) {
             // Merge the env variables with that of the kernel env.
-            interpreterEnv = interpreterEnv || {};
+            interpreterEnv = interpreterEnv || customEnvVars;
 
             if (this.configService.getSettings(resource).useOldKernelResolve) {
                 this.envVarsService.mergeVariables(interpreterEnv, mergedVars); // interpreter vars win over proc.
@@ -140,6 +140,15 @@ export class KernelEnvironmentVariablesService {
                 }
             } else {
                 Object.assign(mergedVars, interpreterEnv, kernelEnv); // kernels vars win over interpreter.
+                if (interpreter && Object.keys(interpreterEnv).length === 0) {
+                    // Possible Python extension fails to get the activated Python env variables,
+                    // Regardless, we need to ensure the Python executable is always at the beginning of the path.
+
+                    // This way all executables from that env are used.
+                    // This way shell commands such as `!pip`, `!python` end up pointing to the right executables.
+                    // Also applies to `!java` where java could be an executable in the conda bin directory.
+                    this.envVarsService.prependPath(mergedVars, path.dirname(interpreter.uri.fsPath));
+                }
             }
 
             // If user asks us to, set PYTHONNOUSERSITE
--- a/src/kernels/raw/launcher/kernelEnvVarsService.node.ts
+++ b/src/kernels/raw/launcher/kernelEnvVarsService.node.ts
@@ -140,15 +140,6 @@ export class KernelEnvironmentVariablesService {
                 }
             } else {
                 Object.assign(mergedVars, interpreterEnv, kernelEnv); // kernels vars win over interpreter.
-                if (interpreter && Object.keys(interpreterEnv).length === 0) {
-                    // Possible Python extension fails to get the activated Python env variables,
-                    // Regardless, we need to ensure the Python executable is always at the beginning of the path.
-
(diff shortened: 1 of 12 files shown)


src/platform/common/variables/customEnvironmentVariablesProvider.node.ts
```diff
@@ -94,30 +116,17 @@ export class CustomEnvironmentVariablesProvider implements ICustomEnvironmentVar
...
 
-        const envFile = this.getEnvFile(workspaceFolderUri, purpose);
+        const envFile = this.getEnvFile(workspaceFolderUri);
         this.createFileWatcher(envFile, workspaceFolderUri);
 
         const promise = this.envVarsService.parseFile(envFile, process.env);
         promise.then((result) => (cacheStoreIndexedByWorkspaceFolder.data = result)).ignoreErrors();
         return promise;
     }
-    public configurationChanged(e: ConfigurationChangeEvent) {
```
DonJayamanne, line 111: Just revert a lot of the changes that were made in a previous PR (reverting a revert commit)

src/kernels/raw/launcher/kernelEnvVarsService.node.ts
```diff
@@ -96,52 +95,68 @@ export class KernelEnvironmentVariablesService {
...
-            traceInfo(`Adding env Variable PYTHONNOUSERSITE to ${getDisplayPath(interpreter?.uri)}`);
-            mergedVars.PYTHONNOUSERSITE = 'True';
-        }
-        if (isPythonKernel) {
-            mergedVars.PYDEVD_IPYTHON_COMPATIBLE_DEBUGGING = '1';
+        if (isPythonKernel || interpreter) {
+            // Merge the env variables with that of the kernel env.
+            interpreterEnv = interpreterEnv || {};
+
+            if (this.configService.getSettings(resource).useOldKernelResolve) {
```
DonJayamanne, line 103: Simple fallback in case users run into issues, will be removed in a later release (debt item)

src/kernels/raw/launcher/kernelEnvVarsService.node.ts
```diff
@@ -96,52 +95,68 @@ export class KernelEnvironmentVariablesService {
...
+                    this.envVarsService.appendPythonPath(mergedVars, kernelEnv.PYTHONPATH);
+                }
+                // Ensure the python env folder is always at the top of the PATH, this way all executables from that env are used.
+                // This way shell commands such as `!pip`, `!python` end up pointing to the right executables.
+                // Also applies to `!java` where java could be an executable in the conda bin directory.
+                if (interpreter) {
+                    this.envVarsService.prependPath(mergedVars, path.dirname(interpreter.uri.fsPath));
+                }
+            } else {
+                Object.assign(mergedVars, interpreterEnv, kernelEnv); // kernels vars win over interpreter.
```
DonJayamanne, line 142: With the new Python API we can just use their API as is to get the resolve variables and merge the kernel variables

src/platform/common/process/environmentActivationService.node.ts
```diff
@@ -110,33 +110,20 @@ export class EnvironmentActivationService implements IEnvironmentActivationServi
...
-                            `Failed to get activated env variables from Python Extension for ${getDisplayPath(
-                                interpreter.uri
-                            )}`,
-                            ex
-                        );
-                        reasonForFailure = 'failedToGetActivatedEnvVariablesFromPython';
-                        return undefined;
-                    })
-            ),
-            this.envVarsService.getCustomEnvironmentVariables(resource, 'RunPythonCode').catch((ex) => {
```
DonJayamanne, line 128: With the new Python API we do not need to merge any of the custom env variables.

rebornix (approved):

PR #12762, opened 2023-02-07
IW creation telemetry

amunger:
for https://github.com/microsoft/vscode-jupyter-internal/issues/378


lramos15 (changes requested):
Please fill out the comments in the GDPR annotations about what each property means, even if they're self explanatory they help users who just read the manifest :)

src/platform/common/constants.ts
```diff
@@ -363,10 +363,6 @@ export enum Telemetry {
     RunSelectionOrLine = 'DATASCIENCE.RUN_SELECTION_OR_LINE',
     RunToLine = 'DATASCIENCE.RUN_TO_LINE',
     RunFromLine = 'DATASCIENCE.RUN_FROM_LINE',
-    /**
-     * Whether auto save feature in VS Code is enabled or not.
-     */
-    CreateNewInteractive = 'DATASCIENCE.CREATE_NEW_INTERACTIVE',
```
amunger, line 369: this event was already redundant since we capture telemetry for every command run, but more specific than the one being added here (the create new `interactive window command` was run)

amunger:
@lramos15 - you were just requesting to fill out the comment field, right? Does that look good now?

lramos15 (approved):

PR #12788, opened 2023-02-08
Append server name to remote kernel description in MRU.

rebornix:
MRU will not show controller kind/group thus we show the server info in the description.

### DIFF
--- a/src/notebooks/controllers/connectionDisplayData.ts
+++ b/src/notebooks/controllers/connectionDisplayData.ts
@@ -13,7 +13,7 @@ import { IJupyterServerUriStorage } from '../../kernels/jupyter/types';
 import { KernelConnectionMetadata, RemoteKernelConnectionMetadata } from '../../kernels/types';
 import { IWorkspaceService } from '../../platform/common/application/types';
 import { IPlatformService } from '../../platform/common/platform/types';
-import { IDisposableRegistry, ReadWrite } from '../../platform/common/types';
+import { IDisposableRegistry, IFeaturesManager, ReadWrite } from '../../platform/common/types';
 import { DataScience } from '../../platform/common/utils/localize';
 import { noop } from '../../platform/common/utils/misc';
 import { EnvironmentType } from '../../platform/pythonEnvironments/info';
@@ -24,6 +24,7 @@ export type ConnectionDisplayData = {
     readonly description: string | undefined;
     readonly detail: string;
     readonly category: string;
+    readonly serverDisplayName?: string;
 };
 
 @injectable()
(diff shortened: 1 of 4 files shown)


src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -309,7 +309,12 @@ export class VSCodeNotebookController implements Disposable, IVSCodeNotebookCont
     private updateDisplayData() {
         const displayData = this.displayDataProvider.getDisplayData(this.connection);
         this.controller.label = displayData.label;
-        this.controller.description = displayData.description;
+        if (this.featureManager.features.kernelPickerType === 'Insiders' && displayData.serverDisplayName) {
+            // MRU kernel picker doesn't show controller kind/category, so add server name to description
+            this.controller.description = displayData.description
+                ? `${displayData.description} (${displayData.serverDisplayName})`
```
DonJayamanne, line 315: We can do this concatenation in the `ConnectionDisplayDataProvider` it self
The goal of the `ConnectionDisplayDataProvbider` is to have all of the logic for `deatils`, `description` and `lable` in one place
Else this same code could end up having to be copied to other places if we use quick pick or a controller.

src/notebooks/controllers/connectionDisplayData.ts
```diff
@@ -60,6 +62,22 @@ export class ConnectionDisplayDataProvider {
...
         this.details.set(connection.id, details);
 
+        if (
+            this.featuresManager.features.kernelPickerType === 'Insiders' &&
+            (connection.kind === 'connectToLiveRemoteKernel' || connection.kind === 'startUsingRemoteKernelSpec')
+        ) {
+            getRemoteServerDisplayName(connection, this.serverUriStorage)
+                .then((displayName) => {
+                    if (details.serverDisplayName !== displayName) {
+                        details.serverDisplayName = displayName;
```
DonJayamanne, line 72: Instead of this, we can just 
```typescript
details.description = 
details.includes(displayName) ? details:`${details.description} (${displayName})`
```

src/notebooks/controllers/vscodeNotebookController.ts
DonJayamanne: This code would not be necesary, i.e. no need to change any code in this file.

DonJayamanne (changes requested):
The goal of the ConnectionDisplayDataProvbider is to have all of the logic for details, description and lable in one place

rebornix:
The concern @DonJayamanne shared above is we don't want to end up with generating kernel display data all over the places. It's a challenge as now the controller list view and the kernel list view in "Kernel Source" are rendering differently. That's the main motivation why this PR only handles the `description` differently in the controller registration, instead of baking this into the `ConnectionDisplayDataProvbider`. We agreed to stay this with approach for now but note that this can potentially be debt in the near future.

PR #12790, opened 2023-02-09
fix: Fix the inaccurate judgment of ipykernel version in debug feature.

realDuang:
Fixes #https://github.com/microsoft/vscode-jupyter/issues/12801


src/notebooks/debugger/helper.ts
```diff
@@ -25,10 +25,18 @@ import ipykernel
...
-        const version = output[0].text.toString().split('.');
-        const majorVersion = Number(version[0]);
-        return majorVersion >= 6 ? IpykernelCheckResult.Ok : IpykernelCheckResult.Outdated;
+    const versionRegex = /^\d+\.\d+\.\d+$/;
+
+    // It is necessary to traverse all the output to determine the version of ipykernel, some jupyter servers may return extra status metadata
+    for (const line of output) {
+        const lineText = line.text?.toString();
+        if (lineText && versionRegex.test(lineText)) {
+            const [majorVersion] = lineText.split('.').map(Number);
```
roblourens, line 34: Why use a regex then ignore the result and do a split instead?

roblourens: You resolved this but it doesn't look like anything changes

realDuang: > You resolved this but it doesn't look like anything changed

Oh hey Rob, I have responsed in channel 3weeks ago and then click the resolve button last week just want to ping you kindly. but somehow my reply disappear. Sorry to confuse you. I will rewrite it.

realDuang: Have changed the regex way to pick up the major version.

roblourens:
I don't know why these tests are failing...

roblourens:
This test is actually failing consistently, maybe it's actually related to your changes, can you check that? `Delete temp debugging files`

realDuang:
> This test is actually failing consistently, maybe it's actually related to your changes, can you check that? `Delete temp debugging files`

That's weird. I just ran the unit tests locally and they all seem to pass. Is the test case in question included in script `test:unittests`?

17525129-eb49-4233-9284-f6439b3b7b69

roblourens:
When I run this, there is a newline at the end of the result, so your regex doesn't match. A `.trim()` should do it

realDuang:
> When I run this, there is a newline at the end of the result, so your regex doesn't match. A `.trim()` should do it

Yeash you're right. I reproduced this with a local notebook。 I also found that the output type only can be a stream.

roblourens (approved):
Thanks!

lramos15 (approved):

PR #12846, opened 2023-02-14
Update kernel selection logic in Interactive window

### LINKED ISSUE #11967: Review default kernel for Interactive Window
Today when run a cell in IW, we have the following logic:
* if we don't have remotes, then always use local active python env
* If we have a remote, use a kernel from remote (what' is used, not very clear)
* If we have more than one remote, what do we do

Should we instead force the user to pick a kernel?

amunger:
Fixes #11967

modified the controller selection logic for IW as seen in `InteractiveControllerHelper.ts`.
Also reduced some redundant code for kernel startup.

The behavior for starting an IW with a remote that no longer exists is not what I had planned - the kernel picker will pop open because of the way we handle kernel connection errors (shared with notebooks), but I think this is OK for now.

### DIFF
--- a/src/notebooks/controllers/helpers.ts
+++ b/src/notebooks/controllers/helpers.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 import { createInterpreterKernelSpec, getKernelId } from '../../kernels/helpers';
-import { PythonKernelConnectionMetadata } from '../../kernels/types';
+import { KernelConnectionMetadata, PythonKernelConnectionMetadata } from '../../kernels/types';
 import { JupyterNotebookView, InteractiveWindowView } from '../../platform/common/constants';
 import { getDisplayPath } from '../../platform/common/platform/fs-paths';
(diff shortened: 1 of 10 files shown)


src/notebooks/controllers/helpers.ts
```diff
@@ -39,3 +39,12 @@ export async function createActiveInterpreterController(
...
     }
 }
+
+export async function isActiveInterpreter(
+    metadata: KernelConnectionMetadata,
+    resource: Resource,
+    interpreters: IInterpreterService
+) {
+    const pythonInterpreter = await interpreters.getActiveInterpreter(resource);
+    return pythonInterpreter?.envPath?.toString() === metadata.interpreter?.envPath?.toString();
```
DonJayamanne, line 49: use `pythonInterpreter?.id === metadata.interpreter?.id`
That is the right way to compare interpreters (Id is provided by Python extension and is a unique identifier)

rebornix (approved):

PR #12896, opened 2023-02-21
Ensure local scripts can be loaded

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/12843

### DIFF
--- a/src/notebooks/controllers/ipywidgets/message/commonMessageCoordinator.ts
+++ b/src/notebooks/controllers/ipywidgets/message/commonMessageCoordinator.ts
@@ -119,70 +119,81 @@ export class CommonMessageCoordinator {
             this,
             this.disposables
         );
+        const deferred = createDeferred<7 | 8>();
+        const sendIPyWidgetsVersion = async () => {
+            const stopWatch = new StopWatch();
+            if (!deferred.completed) {
+                // Determine the version of ipywidgets and send the appropriate script url to the webview.
+                traceVerbose('Attempting to determine version of IPyWidgets');
+                const disposables: IDisposable[] = [];
+                const kernelProvider = this.serviceContainer.get<IKernelProvider>(IKernelProvider);
+                const kernelPromise = createDeferred<IKernel>();
+                if (kernelProvider.get(this.document)) {
+                    kernelPromise.resolve(kernelProvider.get(this.document));
+                } else {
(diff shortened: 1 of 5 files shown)


src/webviews/webview-side/ipywidgets/dummy.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/**
+ * We need to tell VS Code some way to register the Jupyter extension directory
+ * as a known root directory where from scripts can be loaded.
+ * If we do not do this and attempt to load scripts from the Jupyter extension directory,
+ * then VS Code will return a 401 error.
+ * One way to do this is to create a script file and define this as a pre-load script for the controller in the controller code.
+ * NOTE: Defining pre-load scripts in the package.json file does not work.
```
DonJayamanne, line 10: Will file an issue and get this fixed within VS Code itself, this is a simple quick and dirty fix

src/notebooks/controllers/ipywidgets/message/commonMessageCoordinator.ts
```diff
@@ -193,6 +204,9 @@ export class CommonMessageCoordinator {
         // In case the webview loaded earlier and it already sent the IPyWidgetMessages.IPyWidgets_Ready message
         // This way we don't make assumptions, we just query widgets and ask its its ready (avoids timing issues etc).
         webview.postMessage({ type: IPyWidgetMessages.IPyWidgets_IsReadyRequest, payload: undefined }).then(noop, noop);
+        // Send the IPyWidgets message immediately, sometimes
+        // the webview is ready before we get the request message.
+        sendIPyWidgetsVersion().catch(noop);
```
DonJayamanne, line 209: This change was to address a race condition

src/webviews/webview-side/ipywidgets/dummy.ts
DonJayamanne: @roblourens /cc Here's the temporary & easy fix.
Will try to look into the issue in core as well (but this is an easy fix for now)

amunger (approved):

DonJayamanne:
Damn it, looks like a bug in github actions, the smoke tests have run to completion, but the action has not detected it

Issue #12808, opened 2023-02-10
Home Brew or python.org Python does not work on Mac, apt-get python3 does not work on linux & windows store Python too does not work


DonJayamanne:
Tested the following scenarios:
* 1. HomeBrew Python on Mac
* 2. Python install from Python org on Mac
* 3. apt-get install python3 on Ubuntu
* 4. Windows Store Python on Windows

In all of these cases when we install packages a warning is displayed in the terminal indicating the fact that packages are being installed in a directory that is not on the PATH.
Upon further investigation it is found that this directory is a USER_SITE directory.

After all, when we install packages into the global envs we use the `--user` flag.
Which results in installing the packages in a user directory (hence USER_SITE).

The work around here is to ensure we add that path into the PATH
This service merely returns the path that needs to be added to the PATH.

# Problem
When zmq does not work,  then starting jupyter using a global python env fails.
As a result kernel startup fails.

# Why don't we see this today for most users
zmq makes it unnecessary to start Jupyter using global python envs
Jupyter executables are stored in the directory that is not in the current path (warning above) as a result, jupyter cannot be located by the packages.

# Solution
* Ensure path is added as per the warning message

Here is the installation output for `python3 -m jupyter notebook --user` when running from the terminal

Notice the message
```
  NOTE: The current PATH contains path(s) starting with `~`, which may not be expanded by all applications.
  WARNING: The script jupyter-events is installed in '/Users/donjayamanne/Library/Python/3.10/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
```

Based on the docs on Home Brew we need to add the path, and this is documented on their site too
https://docs.brew.sh/Homebrew-and-Python

```
site-packages and the PYTHONPATH
The site-packages is a directory that contains Python modules, including bindings installed by other formulae. Homebrew creates it here:

$(brew --prefix)/lib/pythonX.Y/site-packages
So, for Python 3.y.z, you’ll find it at /usr/local/lib/python3.y/site-packages.

Python 3.y also searches for modules in:

/Library/Python/3.y/site-packages
~/Library/Python/3.y/lib/python/site-packages
Homebrew’s site-packages directory is first created (1) once any Homebrew formulae with Python bindings are installed, or (2) upon brew install python.
```

```
ollecting arrow>=0.15.0
  Using cached arrow-1.2.3-py3-none-any.whl (66 kB)
[...]

PR #12958, opened 2023-02-28
Exclude string python from path (#12953)

DonJayamanne:
For #12808 and #12632


amunger:
this will go into a recovery build

amunger (approved):

PR #12969, opened 2023-03-01
Deprecate API, no longer valid with kernel picker

DonJayamanne:


src/gdpr.ts
```diff
@@ -381,7 +381,7 @@
 //Telemetry.JupyterApiUsage
 /* __GDPR__
    "DATASCIENCE.JUPYTER_API_USAGE" : {
-     "extensionId": {"classification":"PublicNonPersonalData","purpose":"FeatureInsight","comment":"Extension Id that's attempting to use the API.","owner":"donjayamanne"},
+     "clientExtId": {"classification":"PublicNonPersonalData","purpose":"FeatureInsight","comment":"Extension Id that's attempting to use the API.","owner":"donjayamanne"},
```
DonJayamanne, line 384: extensionId might be too generic as we're not getting any names, hence changing the property to somethign else.

mjbvz (approved):

PR #12970, opened 2023-03-01
Fix electron 22 crash: update zeromq.

rebornix:
Prepare for the upcoming 1.77 Insiders. 1.77 ships with Electron 22, which introduces v8 memory cage. Current zeromq will crash with Electron 22. The issue was already fixed in a newer version of zeromq https://github.com/zeromq/zeromq.js/issues/514 .

More details about the v8 memory cage https://github.com/electron/electron/pull/36624

### DIFF
--- a/.github/actions/build-vsix/action.yml
+++ b/.github/actions/build-vsix/action.yml
@@ -10,6 +10,8 @@ runs:
   using: 'composite'
   steps:
     - run: npm ci --prefer-offline
+      env:
+        npm_config_build_from_source: true
       shell: bash
 
     - run: npm run package
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -112,6 +112,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Install dependencies
+        run: |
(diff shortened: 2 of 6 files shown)


DonJayamanne (changes requested):
Why is this required, not sure why we are updating the package apart from the fact that its a little bit more recent that the previous.
Some information (probably an issue would be great)

This way when we go bakc and look at ths history we know why things were updated.

DonJayamanne (approved):

rebornix:
@DonJayamanne thanks for pointing that out. PR created in a hurry. I have updated the issue description to explain why this update is needed.

DonJayamanne:
@rebornix no worries, saw the messages, and realized what it was ,

DonJayamanne (commented):
please could you test this vsix on windows
I don’t think it will work, I haven’t tried it myself

Please have a look at my pr 
You might also need to copy the nodegyp files
Else the vsix will not work on windows or apple intel now will it work on Linux
It will still fallback to Jupyter and Zamboanga will not load

DonJayamanne:
I think we can close this now
@rebornix feel free to reopen if you think we still need this.

PR #12978, opened 2023-03-02
Remove unwanted packages

DonJayamanne:


src/test/common/asyncDump.ts
```diff
@@ -1,10 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-/* eslint-disable @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires */
-const log = require('why-is-node-running');
-
 // Call this function to debug async hangs. It should print out stack traces of still running promises.
 export function asyncDump() {
-    log();
+    require('why-is-node-running')();
```
DonJayamanne, line 6: This is almost never used (only used when trying to figure out why tests hang), 
hence package can be installed when we need it.

sandy081 (approved):

PR #12992, opened 2023-03-03
more separation of IW responsibilities

amunger:
Another step towards cleaning up the IW code. There is a lot more to do, but I feel like this is good stopping point unless you see something glaring and easy to tackle.


bpasero (dismissed):

amunger:
found a regression that I should fix before this goes in - kernel restarts and new connections won't add an info cell or update it correctly on connection

DonJayamanne (approved):

PR #13049, opened 2023-03-10
Update zeromq to beta 16

### LINKED ISSUE #5206: Zeromq for Mac M1  & other linux distros
Zeromq is used to communicate with kernels. If zeromq didn't work on an os, then we rely on users installing jupyter, which will install python and zeromq will get built on the os.
Mac m1 is an os for which we don't have a prebuilt binary.

* I can confirm building zeromq works on Mac M1, & this works in a Mac

**TODO:**
* Ship this as an experiment (for controlled roll out)
* Consider contributing upstream
* Sign & Build dependencies (build pipeline)
* Should we sign & build dependencies for all OS & linux distros (even though we already have pre-built packages via npm)
* How to detect what's a Mac M1
	* We could optionally load currrenly shipped zeromq and then if that fails, try our build.
	* Would require some changes to the zeromq package code or we'll need to update the files during runtime.

DonJayamanne:
Fixes #5206

### DIFF
--- a/build/webpack/webpack.extension.node.config.js
+++ b/build/webpack/webpack.extension.node.config.js
@@ -129,6 +129,7 @@ const config = {
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromq/**/*.node' }] }),
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromq/**/*.json' }] }),
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/node-gyp-build/**/*' }] }),
+        new copyWebpackPlugin({ patterns: [{ from: './node_modules/@aminya/node-gyp-build/**/*' }] }),
         new webpack.DefinePlugin({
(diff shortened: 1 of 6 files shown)


build/webpack/webpack.extension.node.config.js
```diff
@@ -129,6 +129,7 @@ const config = {
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromq/**/*.node' }] }),
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromq/**/*.json' }] }),
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/node-gyp-build/**/*' }] }),
+        new copyWebpackPlugin({ patterns: [{ from: './node_modules/@aminya/node-gyp-build/**/*' }] }),
```
rzhao271, line 132: Does this collide with the node-gyp line above?

DonJayamanne: Will remove that, thanks

rzhao271 (approved):

PR #13117, opened 2023-03-20
Fallback zeromq binaries

DonJayamanne:
Add support for falling back to older version of zmq if the new version fails to load the binaries built by us.

### DIFF
--- a/build/webpack/webpack.extension.node.config.js
+++ b/build/webpack/webpack.extension.node.config.js
@@ -129,6 +129,10 @@ const config = {
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromq/**/*.node' }] }),
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromq/**/*.json' }] }),
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/@aminya/node-gyp-build/**/*' }] }),
+        new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromqold/**/*.js' }] }),
+        new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromqold/**/*.node' }] }),
+        new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromqold/**/*.json' }] }),
+        new copyWebpackPlugin({ patterns: [{ from: './node_modules/node-gyp-build/**/*' }] }),
         new webpack.DefinePlugin({
             IS_PRE_RELEASE_VERSION_OF_JUPYTER_EXTENSION: JSON.stringify(
                 typeof process.env.IS_PRE_RELEASE_VERSION_OF_JUPYTER_EXTENSION === 'string'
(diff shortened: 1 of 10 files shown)


src/kernels/raw/session/rawNotebookSupportedService.node.ts
```diff
@@ -1,15 +1,10 @@
...
 
-import * as fs from 'fs';
-import * as os from 'os';
 import { inject, injectable } from 'inversify';
-import { traceWarning } from '../../../platform/logging';
 import { IConfigurationService } from '../../../platform/common/types';
 import { IRawNotebookSupportedService } from '../types';
-import { Telemetry, sendTelemetryEvent } from '../../../telemetry';
-import { noop } from '../../../platform/common/utils/misc';
-import { DistroInfo, getDistroInfo } from '../../../platform/common/platform/linuxDistro.node';
```
DonJayamanne, line 12: Moved all of this into a separate file

build/webpack/webpack.extension.node.config.js
```diff
@@ -125,10 +126,15 @@ const config = {
         // However we don't webpack to manage this, so it was part of the excluded modules. Delete it from there
         // so at runtime we pick up the original structure.
         new removeFilesWebpackPlugin({ after: { include: ['./out/node_modules/zeromq.js'], log: false } }),
+        new removeFilesWebpackPlugin({ after: { include: ['./out/node_modules/zeromqold.js'], log: false } }),
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromq/**/*.js' }] }),
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromq/**/*.node' }] }),
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromq/**/*.json' }] }),
         new copyWebpackPlugin({ patterns: [{ from: './node_modules/@aminya/node-gyp-build/**/*' }] }),
+        new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromqold/**/*.js' }] }),
+        new copyWebpackPlugin({ patterns: [{ from: './node_modules/zeromqold/**/*.node' }] }),
```
DonJayamanne, line 135: old code that applied to the older node module.
(this is a temporary solution, where we ship the old binary as well, this will help us determine how many users are potentially going to be impacted by the new binary)
I.e. the goal is to not break any existing users, but gather information about who would be broken if we do not ship the fallback.

src/kernels/raw/session/zeromq.node.ts
```diff
@@ -0,0 +1,75 @@
...
+ * version of zmq binary to be loaded.
+ * Source from node_modules/@aminya/node-gyp-build/index.js
+ * (@aminya/node-gyp-build is what is used by zeromq.js)
+ */
+function getPlatformInfo() {
+    try {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const vars: Record<string, any> = (process.config && process.config.variables) || {};
+        const arch = process.env.npm_config_arch || os.arch();
+        const platform = process.env.npm_config_platform || os.platform();
```
DonJayamanne, line 61: This is pretty much existing code. but moved around{ to this new file.

roblourens (approved):

PR #13121, opened 2023-03-21
Fail build of vsix if zmq binaries are missing

DonJayamanne:
We have a more full proof solution here https://github.com/microsoft/vscode-jupyter/pull/13120/files
To ensure we never run into 403


roblourens (approved):

TylerLeonhardt (commented):
if this is running in github actions, does using the env var GITHUB_TOKEN, prevent the 403?

DonJayamanne:
> if this is running in github actions, does using the env var GITHUB_TOKEN, prevent the 403?

Yes

PR #13153, opened 2023-03-23
Refresh python interpreters only once

### LINKED ISSUE #11854: Running Python code in a Jupyter notebook fails to install missing notebook dependencies
Encountered this in a UX study, cc @bamurtaugh for environment details

DonJayamanne:
Fixes #11854


amunger:
is there a reason that `onDidChangeActiveNotebookEditor` is fired so often with no user activity?

DonJayamanne:
> is fired so often with no user activity?

As you can see, interpreter change is detected, and then we trigger the `onDidChangeActiveNotebookEditor`
Howver inside that method the refreshInterpreetrs is called once again, and that causes all of the interpreters to get refreshed all over again

amunger (approved):

DonJayamanne:
@amunger oh goodie, all of this code  can be removed, will fix in debt week for sure

PR #13169, opened 2023-03-27
Improved searching of interpreter kernelSpecs

DonJayamanne:


src/kernels/raw/finder/jupyterPaths.node.ts
```diff
@@ -281,12 +283,27 @@ export class JupyterPaths {
             }
         }
     }
+    private cachedKernelSpecRootPaths?: { promise: Promise<Uri[]>; stopWatch: StopWatch };
     /**
      * This list comes from the docs here:
      * https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs
      */
-    @traceDecoratorVerbose('Get KernelSpec root path')
     public async getKernelSpecRootPaths(cancelToken: CancellationToken): Promise<Uri[]> {
```
DonJayamanne, line 291: Caching the results and logging only when we call without cachine.
Else this results in excessive logging, also with the caching we've improved perf slightly

src/kernels/raw/finder/localPythonAndRelatedNonPythonKernelSpecFinder.node.ts
```diff
@@ -283,41 +282,53 @@ export class LocalPythonAndRelatedNonPythonKernelSpecFinder extends LocalKernelS
...
-        // then list all of the global python kernel specs.
-        if (this.extensionChecker.isPythonExtensionInstalled) {
-            await Promise.all(
-                interpreters.map(async (interpreter) => {
-                    const kernels = await listPythonAndRelatedNonPythonKernelSpecs(
+        await Promise.all(
+            interpreters.map(async (interpreter) => {
+                let finder = this.interpreterKernelSpecs.get(interpreter.id);
+                if (!finder) {
+                    finder = new InterpreterSpecificKernelSpecsFinder(
```
DonJayamanne, line 300: Now we find kernelspecs for each interpreter in a seprate class with its own state.
when extension is loading, pytyohn extension would trigger changes to envs every few ms, and that would result in blowing aways everything even though there were not changes to interepters, but only new interpreters were discovoered.

As a result things should be fractionally faster, with fewer resource usage.

src/kernels/raw/finder/jupyterPaths.node.ts
```diff
@@ -341,11 +361,7 @@ export class JupyterPaths {
      * We need to look at the 'kernels' sub-directory and these paths are supposed to come first in the searching
      * https://jupyter.readthedocs.io/en/latest/projects/jupyter-directories.html#envvar-JUPYTER_PATH
      */
-    @traceDecoratorVerbose('Get Jupyter Sub Paths')
-    private async getJupyterPathSubPaths(
-        @ignoreLogging() cancelToken?: CancellationToken,
-        subDir?: string
-    ): Promise<Uri[]> {
+    private async getJupyterPathSubPaths(cancelToken?: CancellationToken, subDir?: string): Promise<Uri[]> {
```
DonJayamanne, line 364: Reduce excessive logging

src/kernels/raw/finder/interpreterKernelSpecFinderHelper.node.ts
```diff
@@ -34,33 +33,269 @@ import { createPromiseFromCancellation } from '../../../platform/common/cancella
...
+        return Array.from(distinctKernelMetadata.values());
+    }
+}
+
+/**
+ * We can have kernel specs in global locations and they could be python.
+ * Some of these could have the python interpreter information in the kernelspec.json file.
+ * Either in the argv or in the metadata (latter being the case where Jupyter extension may have added interpreter info into kernelSpec metadata).
+ */
+export class GlobalPythonKernelSpecFinder implements IDisposable {
```
DonJayamanne, line 249: Another new class to clearly separate the concerns
previously when searching through each interpreter ahdn when looking for kernlespecs inside that interrpeter, we would also see whether any of the global kernelspecs matched that interpreter.

That was wrong, 
i.e. if we have 10 interpreters, then 10 times we'd try to match a global kernelspec with each interpeter.
So, if we have 5 global python kernelspecs, then we turn a matche 10*5 = 50 times

Now we only try to match 5 times.

aeschli (approved):

PR #13204, opened 2023-03-28
re-added IW smoke test

amunger:
re-enabled one test and deleted the other (disabled) test for checking the kernel matches the IW. That command still doesn't take a parameter AFIACT, and that test could just in the integration suite anyway.


src/test/smoke/datascience.smoke.test.ts
```diff
@@ -42,32 +39,32 @@ suite('Smoke Tests', function () {
...
+            'simple_note_book.py'
+        );
+        const outputFile = path.join(path.dirname(file), 'ds.log');
+        if (await fs.pathExists(outputFile)) {
+            await fs.unlink(outputFile);
+        }
+        const textDocument = await openFile(file);
+
+        // Wait for code lenses to get detected.
+        console.log('Step0');
```
amunger, line 58: test comment

PR #13206, opened 2023-03-28
Create new task if previous task has been disposed

### LINKED ISSUE #13199: `clear_output(wait=True)` does not work inside async call
## Environment data

-   VS Code version: 1.76.2
-   Jupyter Extension version (available under the Extensions sidebar): v2023.2.1200692131
-   Python Extension version (available under the Extensions sidebar): v2023.4.1
-   OS (Windows | Mac | Linux distro) and version: Mac 13.2.1 (22D68)
-   Python and/or Anaconda version: 3.10
-   Jupyter server running: Remote

## Expected behaviour

`clear_output(wait=True)` clears the output even during an async call.

## Actual behaviour

`clear_output(wait=True)` does not clear the output.

## Steps to reproduce:

Run the following code in a cell, notice that the `clear_output(wait=True)` call does not have any effect (both print1 and print2 are displayed).

```python
import asyncio
from IPython.display import display_html, clear_output, display

display({"text/html": "<b>print1</b>"}, raw=True, clear=True, include=["text/html"])

def f():
    #clear_output(wait=False) # uncomment this and it works (but causes flickering)
    clear_output(wait=True)
    display({"text/html": "<b>print2</b>"}, raw=True, clear=True, include=["text/html"])
asyncio.get_event_loop().call_later(1, f)
```

Note that this does work as expected in a Jupyter notebook.

DonJayamanne:
Fixes #13199

### DIFF
--- a/src/kernels/execution/cellExecutionMessageHandler.ts
+++ b/src/kernels/execution/cellExecutionMessageHandler.ts
@@ -503,18 +503,18 @@ export class CellExecutionMessageHandler implements IDisposable {
             return;
         }
         traceCellMessage(this.cell, 'Update output');
-        // Clear if necessary
-        this.clearOutputIfNecessary(this.execution);
-        // Keep track of the display_id against the output item, we might need this to update this later.
-        if (displayId) {
-            CellOutputDisplayIdTracker.trackOutputByDisplayId(this.cell, displayId, cellOutput);
-        }
-
         // Append to the data (we would push here but VS code requires a recreation of the array)
         // Possible execution of cell has completed (the task would have been disposed).
         // This message could have come from a background thread.
         // In such circumstances, create a temporary task & use that to update the output (only cell execution tasks can update cell output).
         let task = new Lazy(() => this.execution || this.createTemporaryTask());
+        // Clear if necessary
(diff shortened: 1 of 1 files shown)


src/kernels/execution/cellExecutionMessageHandler.ts
```diff
@@ -503,18 +503,18 @@ export class CellExecutionMessageHandler implements IDisposable {
...
         }
         traceCellMessage(this.cell, 'Update output');
+        // Append to the data (we would push here but VS code requires a recreation of the array)
+        // Possible execution of cell has completed (the task would have been disposed).
+        // This message could have come from a background thread.
+        // In such circumstances, create a temporary task & use that to update the output (only cell execution tasks can update cell output).
+        let task = new Lazy(() => this.execution || this.createTemporaryTask());
         // Clear if necessary
-        this.clearOutputIfNecessary(this.execution);
+        this.clearOutputIfNecessary(task.getValue());
```
Tyriar, line 512: Doesn't calling `getValue` immediately defeat the purpose of using `Lazy`?

Tyriar (changes requested):

Tyriar (approved):

PR #13219, opened 2023-03-30
Restart instead of interrupt as interrupt is flaky

DonJayamanne:


src/test/datascience/notebook/executionService.vscode.test.ts
```diff
@@ -318,17 +318,7 @@ suite('Kernel Execution @kernelCore', function () {
             5_000,
             'Cell did not get cleared'
         );
-        await kernel.interrupt();
-        if (getOSType() == OSType.Windows) {
-            // Interrupting a cell on Windows is flaky. there isn't much we can do about it.
```
DonJayamanne, line 323: We know interrupting can be flaky on windows (due to the way Python runtime works)
Instead of interrupting the kernels in this this test, just restart the kernel

chrmarti (approved):

PR #13273, opened 2023-04-04
Faster connections to live remote kernels

DonJayamanne:
For #13275 and #13276

### DIFF
--- a/src/kernels/kernel.ts
+++ b/src/kernels/kernel.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import type * as nbformat from '@jupyterlab/nbformat';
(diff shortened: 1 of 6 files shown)


src/kernels/kernel.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import type * as nbformat from '@jupyterlab/nbformat';
+import * as nbformat from '@jupyterlab/nbformat';
```
rebornix, line 4: Any reason we are not just `import type`anymore?

DonJayamanne: thats weird, don't remember chaning that.

rebornix (approved):

PR #13297, opened 2023-04-10
refactor multiple RestartKernel commands into single, under Jupyter category

### LINKED ISSUE #13295: Multiple Restart Kernel commands for different contexts lead to needless clutter/confusion
Both of these commands share identical implementation, just appear in different contexts/menus.

https://github.com/microsoft/vscode-jupyter/blob/main/src/notebooks/notebookCommandListener.ts#L70-L81
https://github.com/microsoft/vscode-jupyter/blob/main/src/notebooks/notebookCommandListener.ts#L89-L95

Can be refactored to just one command, and change `when` and `enablement` conditions

Yoyokrazy:
Fixes #13295

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -42,7 +42,7 @@ interface ICommandNameWithoutArgumentTypeMapping {
     [DSCommands.InterruptKernel]: [{ notebookEditor: { notebookUri: Uri } } | undefined];
     [DSCommands.RestartKernel]: [{ notebookEditor: { notebookUri: Uri } } | undefined];
     [DSCommands.NotebookEditorRemoveAllCells]: [];
-    [DSCommands.NotebookEditorRestartKernel]: [{ notebookEditor: { notebookUri: Uri } } | undefined | Uri];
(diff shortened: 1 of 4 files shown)


src/commands.ts
```diff
@@ -42,7 +42,7 @@ interface ICommandNameWithoutArgumentTypeMapping {
     [DSCommands.InterruptKernel]: [{ notebookEditor: { notebookUri: Uri } } | undefined];
     [DSCommands.RestartKernel]: [{ notebookEditor: { notebookUri: Uri } } | undefined];
     [DSCommands.NotebookEditorRemoveAllCells]: [];
-    [DSCommands.NotebookEditorRestartKernel]: [{ notebookEditor: { notebookUri: Uri } } | undefined | Uri];
+    // [DSCommands.NotebookEditorRestartKernel]: [{ notebookEditor: { notebookUri: Uri } } | undefined | Uri];
```
rebornix, line 45: we can remove it now

rebornix (approved):

rebornix (approved):

PR #13316, opened 2023-04-12
Add `Restart Kernel and Run All Cells` command

### LINKED ISSUE #9242: Add "Restart and run all" command
### Discussed in https://github.com/microsoft/vscode-jupyter/discussions/9217

<div type='discussions-op-text'>

<sup>Originally posted by **benlindsay** March  3, 2022</sup>
In the Jupyter toolbar, there's a Run All button 

image

which looks a lot like the Restart and Run All button in Jupyter Lab

image

My initial expectation was that those buttons would be analogous, but in VSCode, it does not restart the kernel or clear outputs. Having that button fail to clear outputs is harmful because state from the previous run can be reused. For example, say you have a notebook with 2 cells:

```python
print(foo)
```

and

```python
foo = 42
```

If you run just the second cell, then click "Run All", your notebook will not fail even though it should. I can't think of any reason someone should want to rerun a notebook without also restarting the kernel. As for the clearing outputs bit, there is a case to be made for not clearing outputs, because it can be nice to see what was previously there. However, 2 points in favor of clearing outputs are 1) it would match behavior expectations from Jupyter Lab, and 2) when outputs are cleared prior to rerunning, it's much easier to tell where a failure occurred when scrolling through the notebook. Harder to tell when old outputs remain.

I hacked together a solution by adding a multicommand keybinding to keybindings.json, but that really shouldn't be necessary for a critical functionality. My keybinding looks like this:

```json
{
  "key": "alt+r",
  "command": "extension.multiCommand.execute",
  "args": {
    "sequence": [
      "notebook.clearAllCellsOutputs",
      "jupyter.notebookeditor.restartkernel",
      "notebook.execute"
    ]
  },
  "when": "notebookEditable && activeEditor == 'workbench.editor.notebook'"
}
```

Thanks for all of y'all's work on this project!</div>

Yoyokrazy:
Fixes #9242

### DIFF
--- a/package.json
+++ b/package.json
@@ -620,6 +620,13 @@
                 "enablement": "isWorkspaceTrusted && (jupyter.interactive.canRestartNotebookKernel || (notebookKernel =~ /^ms-toolsai.jupyter\\// && jupyter.notebookeditor.canrestartNotebookkernel))"
             },
             {
+                "command": "jupyter.restartkernelandrunallcells",
+                "title": "%jupyter.command.jupyter.restartkernelandrunallcells.title%",
+                "shortTitle": "%jupyter.command.jupyter.restartkernelandrunallcells.shorttitle%",
+                "category": "Jupyter",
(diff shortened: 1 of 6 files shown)


rebornix (approved):

rebornix:
It looks great, thank you!

PR #13342, opened 2023-04-17
Add support for `Restart Kernel + Run up to Selected Cell` command

### LINKED ISSUE #9242: Add "Restart and run all" command

Yoyokrazy:
Fixes #9242 

Only thought is a different enablement, taking into account `notebookEditorFocused` and/or `jupyter.hascodecells`? On the other hand, commands like `jupyter.runcurrentcell` only require `isWorkspaceTrusted` to be used. 

### DIFF
--- a/package.json
+++ b/package.json
@@ -631,7 +631,7 @@
                 "title": "%jupyter.command.jupyter.restartkernelandrunuptoselectedcell.title%",
                 "shortTitle": "%jupyter.command.jupyter.restartkernelandrunuptoselectedcell.title%",
                 "category": "Jupyter",
-                "enablement": "isWorkspaceTrusted && (jupyter.interactive.canRestartNotebookKernel || (notebookKernel =~ /^ms-toolsai.jupyter\\// && jupyter.notebookeditor.canrestartNotebookkernel))"
+                "enablement": "isWorkspaceTrusted && (notebookKernel =~ /^ms-toolsai.jupyter\\// && jupyter.notebookeditor.canrestartNotebookkernel)"
             },
             {
                 "command": "jupyter.notebookeditor.addcellbelow",
--- a/package.json
+++ b/package.json
@@ -627,6 +627,13 @@
                 "enablement": "isWorkspaceTrusted && (jupyter.interactive.canRestartNotebookKernel || (notebookKernel =~ /^ms-toolsai.jupyter\\// && jupyter.notebookeditor.canrestartNotebookkernel))"
             },
             {
+                "command": "jupyter.restartkernelandrunuptoselectedcell",
+                "title": "%jupyter.command.jupyter.restartkernelandrunuptoselectedcell.title%",
+                "shortTitle": "%jupyter.command.jupyter.restartkernelandrunuptoselectedcell.title%",
+                "category": "Jupyter",
+                "enablement": "isWorkspaceTrusted && (jupyter.interactive.canRestartNotebookKernel || (notebookKernel =~ /^ms-toolsai.jupyter\\// && jupyter.notebookeditor.canrestartNotebookkernel))"
+            },
+            {
                 "command": "jupyter.notebookeditor.addcellbelow",
                 "title": "%jupyter.command.jupyter.notebookeditor.addcellbelow.title%",
                 "category": "Notebook",
--- a/src/notebooks/notebookCommandListener.ts
+++ b/src/notebooks/notebookCommandListener.ts
@@ -103,9 +103,9 @@ export class NotebookCommandListener implements IDataScienceCommandListener {
                 Commands.RestartKernelAndRunUpToSelectedCell,
                 (context?: { notebookEditor: { notebookUri: Uri } }) => {
                     if (context && 'notebookEditor' in context) {
-                        this.RestartKernelAndRunUpToSelectedCell(context?.notebookEditor?.notebookUri).catch(noop);
+                        this.restartKernelAndRunUpToSelectedCell(context?.notebookEditor?.notebookUri).catch(noop);
                     } else {
-                        this.RestartKernelAndRunUpToSelectedCell(context).catch(noop);
+                        this.restartKernelAndRunUpToSelectedCell(context).catch(noop);
                     }
                 }
             )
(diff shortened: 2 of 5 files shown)


src/notebooks/notebookCommandListener.ts
```diff
@@ -175,6 +187,20 @@ export class NotebookCommandListener implements IDataScienceCommandListener {
         this.runAllCells();
     }
 
+    private async RestartKernelAndRunUpToSelectedCell(notebookUri: Uri | undefined) {
```
rebornix, line 190: restartKernelAndRunUpToSelectedCell

src/notebooks/notebookCommandListener.ts
```diff
@@ -175,6 +187,20 @@ export class NotebookCommandListener implements IDataScienceCommandListener {
...
+        const uri = notebookUri ?? this.notebookEditorProvider.activeNotebookEditor?.notebook.uri;
+
+        await this.restartKernel(uri);
+        if (this.notebooks.activeNotebookEditor) {
```
rebornix, line 194: Use `uri`.

src/notebooks/notebookCommandListener.ts
```diff
@@ -175,6 +187,20 @@ export class NotebookCommandListener implements IDataScienceCommandListener {
...
+            this.commandManager
```
DonJayamanne, line 195: I think we should capture this information before line 193 (i.e. before we restart), as its possible for focus to change while we're restarting (sure its unlikely, but who knows what `restartKernel` could do in the future).

DonJayamanne:
Out of curiosity is there any reason why we cannot do something like this https://github.com/microsoft/vscode-jupyter/issues/9242#issuecomment-1489370206

Yoyokrazy:
> Out of curiosity is there any reason why we cannot do something like this [#9242 (comment)](https://github.com/microsoft/vscode-jupyter/issues/9242#issuecomment-1489370206)

Primarily for ease of use and discoverability for the user. It seems like using a dual command as mentioned to bind them together would force the user to search for specifically the term `runCommand`. Ideally there would be a way to add a label, then it would probably be a more desirable solution. Talked with @rebornix and think that core commands that jupyter supports can be implemented as in this PR, and more custom sets of commands can be implemented via `keybindings.json`.

rebornix (dismissed):

src/notebooks/notebookCommandListener.ts
```diff
@@ -175,6 +187,20 @@ export class NotebookCommandListener implements IDataScienceCommandListener {
         this.runAllCells();
     }
 
+    private async restartKernelAndRunUpToSelectedCell(notebookUri: Uri | undefined) {
+        const activeNBE = this.notebookEditorProvider.activeNotebookEditor;
```
rebornix, line 191: do we need `this.notebookEditorProvider.activeNotebookEditor` here? It's possible that restart happens on an inactive notebook editor but this check will block the restart from running.

package.json
```diff
@@ -626,6 +626,13 @@
                 "category": "Jupyter",
                 "enablement": "isWorkspaceTrusted && (jupyter.interactive.canRestartNotebookKernel || (notebookKernel =~ /^ms-toolsai.jupyter\\// && jupyter.notebookeditor.canrestartNotebookkernel))"
             },
+            {
+                "command": "jupyter.restartkernelandrunuptoselectedcell",
+                "title": "%jupyter.command.jupyter.restartkernelandrunuptoselectedcell.title%",
+                "shortTitle": "%jupyter.command.jupyter.restartkernelandrunuptoselectedcell.title%",
+                "category": "Jupyter",
+                "enablement": "isWorkspaceTrusted && (jupyter.interactive.canRestartNotebookKernel || (notebookKernel =~ /^ms-toolsai.jupyter\\// && jupyter.notebookeditor.canrestartNotebookkernel))"
```
DonJayamanne, line 634: Wait, why are we adding this new command, i thought this was `Restart and Run all`, this seems to be `Restart and Run upto selected cell`
Why do we need this new command, i didn't realize this new command was a requirement (also i do not think this is something that exists in Jupyter either, hence feels a little odd to add this)

rebornix: image

@DonJayamanne it exists in Jupyter Lab.

src/notebooks/notebookCommandListener.ts
Yoyokrazy: Fair point, then I'll shift to taking the uri as done on L#193 and then it should be covered either case. Just need to change the range on L#197

rebornix (approved):

PR #13351, opened 2023-04-19
Capture zmq load failure message

DonJayamanne:


src/kernels/raw/session/zeromq.node.ts
```diff
@@ -21,18 +21,23 @@ export function getZeroMQ(): typeof import('zeromq') {
             const requireFunc = typeof __webpack_require__ === 'function' ? __non_webpack_require__ : require;
             const zmq = requireFunc(path.join(EXTENSION_ROOT_DIR, 'out', 'node_modules', 'zeromqold'));
             traceInfo('ZMQ loaded via fallback mechanism.');
-            sendZMQTelemetry(false, true).catch(noop);
+            sendZMQTelemetry(false, true, e.message || e.toString()).catch(noop);
             return zmq;
         } catch (e2) {
-            sendZMQTelemetry(true, true).catch(noop);
+            sendZMQTelemetry(true, true, e.message || e.toString(), e2.message || e2.toString()).catch(noop);
```
DonJayamanne, line 27: @lramos15 is this a safe way of passing the error message?

bpasero (approved):

src/kernels/raw/session/zeromq.node.ts
lramos15: This should be fine, it should clean the message. You may want to test to verify. Also feel free to use this caller for errors. This caller allows errors to get through when the level is set to error or lower rather than all or lower.
`sendTelemetryErrorEvent(eventName: string, properties?: TelemetryEventProperties, measurements?: TelemetryEventMeasurements): void;`

PR #13402, opened 2023-04-25
Fix bundling of windows files

DonJayamanne:


build/webpack/webpack.extension.node.config.js
```diff
@@ -171,12 +191,24 @@ const config = {
...
+                        )
+                }
+            ]
+        }),
+        new copyWebpackPlugin({
+            patterns: [
                 // Copy files from fallback zmq package.
                 {
                     from: './node_modules/zeromqold/**/*',
-                    filter: shouldCopyFileFromZmqFolder.bind(this, './node_modules/zeromqold')
```
DonJayamanne, line 179: This was the root cause, on windows the path sep was `\`
However modified all of the code to be platform agnostic for the builds.

connor4312 (approved):

PR #13447, opened 2023-05-03
Don't rely on the vscode-interactive schema for IW resources

amunger:
necessary if we are going to use untitled resources for IW in https://github.com/microsoft/vscode/issues/172974

including @debonte because of the change to `getNotebookUriFromInputBoxUri`, let me know if there's something I should be testing against that change.


DonJayamanne (approved):

debonte:
> let me know if there's something I should be testing against that change.

@amunger, given this executed cell:

```python
existingVar = 1
```

Please verify:
1. When you type `exist` in the input box, you get `existingVar` as a completion suggestion.
2. After accepting that suggestion, ensure that your usage of `existingVar` (in the input box, before execution) does not have a red squiggle indicating that it is unknown / not defined.

I think that should be sufficient to verify that Pylance, Python, and Jupyter are all working together to make the IW inbox box look like a notebook cell in LSP.

PR #13554, opened 2023-05-24
Remove unused telemetry

lramos15:
Removes some old unused telemetry


amunger:
completed in https://github.com/microsoft/vscode-jupyter/pull/13563

PR #13560, opened 2023-05-25
Do not look for duplicates

### LINKED ISSUE #13484: Can't connect to a new remote kernel without restarting vscode

DonJayamanne:
Fixes #13484

### DIFF
--- a/src/standalone/userJupyterServer/userServerUrlProvider.ts
+++ b/src/standalone/userJupyterServer/userServerUrlProvider.ts
@@ -223,19 +223,6 @@ export class UserJupyterServerUrlProvider implements IExtensionSyncActivationSer
                     const uri = input.value;
 
                     try {
-                        for (let server of this._servers) {
-                            if (server.uri === uri) {
-                                // already exist
-                                input.validationMessage = DataScience.UserJupyterServerUrlAlreadyExistError;
-                                return;
-                            }
-                        }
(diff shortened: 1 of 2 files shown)


src/standalone/userJupyterServer/userServerUrlProvider.ts
```diff
@@ -222,19 +222,6 @@ export class UserJupyterServerUrlProvider implements IExtensionSyncActivationSer
                 input.onDidAccept(async () => {
                     const uri = input.value;
 
-                    try {
-                        for (let server of this._servers) {
-                            if (server.uri === uri) {
-                                // already exist
-                                input.validationMessage = DataScience.UserJupyterServerUrlAlreadyExistError;
```
DonJayamanne, line 229: This is not a valid check as this fail incorrectly.

Assume a user has two remote providers, one is built in provider (this class) & azml.
Now user selects Az ML kernels over and every time we somehow end up with a new provider + handle.
Now lets assume we did this 10 times, now we have 10 MRU entries for AzML.
Ie. the user entered URL is not in the MRU.

However this class `userServerUrlProvider.ts` keeps track of that and does not know that it has been removed.
So, its best to not check for duplicates and let them add duplicates if they want (its much safter). If they end up with duplicates they can always delete them.

alexr00 (approved):

PR #13667, opened 2023-06-07
Rename internal IJupyterUriProvider, add new prop

DonJayamanne:
Part of the larger refactoring.

### DIFF
--- a/src/api.d.ts
+++ b/src/api.d.ts
@@ -3,21 +3,16 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { CancellationToken, Event, NotebookDocument, QuickPickItem, Uri } from 'vscode';
+import { CancellationToken, Disposable, Event, NotebookDocument, QuickPickItem, Uri } from 'vscode';
 import type { Kernel } from '@jupyterlab/services/lib/kernel';
 import type { Session } from '@jupyterlab/services';
-
-/**
- * Data represents the message payload received over the WebSocket.
- */
-export type WebSocketData = string | Buffer | ArrayBuffer | Buffer[];
 
 export interface JupyterAPI {
     /**
      * Registers a remote server provider component that's used to pick remote jupyter server URIs
      * @param serverProvider object called back when picking jupyter server URI
      */
-    registerRemoteServerProvider(serverProvider: IJupyterUriProvider): void;
+    registerRemoteServerProvider(serverProvider: IJupyterUriProvider): Disposable;
     /**
      * Adds a remote Jupyter Server to the list of Remote Jupyter servers.
      * This will result in the Jupyter extension listing kernels from this server as items in the kernel picker.
@@ -38,7 +33,7 @@ export interface JupyterAPI {
      */
     openNotebook(uri: Uri, kernelId: string): Promise<NotebookDocument>;
 }
-
+//#region Jupyter Server Providers
 export interface IJupyterServerUri {
     baseUrl: string;
     /**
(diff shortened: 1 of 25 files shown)


src/standalone/api/api.ts
```diff
@@ -133,10 +129,23 @@ export function buildApi(
...
             sendApiUsageTelemetry(extensions, 'registerRemoteServerProvider');
             const container = serviceContainer.get<IJupyterUriProviderRegistration>(IJupyterUriProviderRegistration);
-            container.registerProvider(picker);
+            let disposeHook = noop;
+            const register = async () => {
+                const extensions = serviceContainer.get<IExtensions>(IExtensions);
+                const extensionId = provider.id.startsWith('_builtin')
+                    ? JVSC_EXTENSION_ID
+                    : (await extensions.determineExtensionFromCallStack()).extensionId;
+                container.registerProvider(Object.assign(provider, { extensionId }));
```
DonJayamanne, line 141: `registerProvider` supports sending a disposable
I believe you added this, so that one has a way to remove the server provider.
Hence i changed the return value to now return a `Disposable`, that is a change in the API, but safe,

DonJayamanne, line 141: I'm passing the extensionId from the very top and ensuring it is a property of the provider itself, 
this way, internally in our classes we know how owns a provider, else we need to work around this.
This is required for part of the larger refactoring to make it easier to determine the owners of each providcer

You might recall, i wanted to replace ServerId with `{extensionId; providerId; handle}`

src/kernels/jupyter/types.ts
```diff
@@ -189,11 +189,12 @@ export interface IJupyterServerUri {
 
 export type JupyterServerUriHandle = string;
 
-export interface IJupyterUriProvider {
+export interface IInternalJupyterUriProvider {
```
DonJayamanne, line 192: This type is used internally and the only difference between this and the exported type for API is this has `extensionId`
Similar to VS Code, we have two types, one for exported APIs and one for internal.
This way we can add `extensionId` to the internal type. else it gets too messy,  other alternatives is always welcome

src/api.d.ts
```diff
@@ -0,0 +1,386 @@
+// Copyright (c) Microsoft Corporation.
```
DonJayamanne, line 1: Moved extension.d.ts into api.d.ts

src/kernels/jupyter/types.ts
```diff
@@ -154,86 +155,17 @@ export interface IJupyterServerProvider {
     getOrStartServer(options: GetServerOptions): Promise<IJupyterConnection>;
 }
 
-export interface IJupyterServerUri {
```
DonJayamanne, line 157: Removed all of these duplicate types, we have them here and the extension.d.ts files.
These types are meant to be public for 3rd party extensions.

src/kernels/jupyter/types.ts
```diff
@@ -154,86 +155,17 @@ export interface IJupyterServerProvider {
...
-    getServerUri(handle: JupyterServerUriHandle): Promise<IJupyterServerUri>;
-    /**
-     * Gets a list of all valid Jupyter Server handles that can be passed into the `getServerUri` method.
-     */
-    getHandles?(): Promise<JupyterServerUriHandle[]>;
-    /**
-     * Users request to remove a handle.
-     */
-    removeHandle?(handle: JupyterServerUriHandle): Promise<void>;
+export interface IInternalJupyterUriProvider extends IJupyterUriProvider {
```
DonJayamanne, line 158: The only difference between this internal type and the public type is we need extensionId in the internal type
This is part of the bigger refactoring, making it easier to determine the owner of a Jupyter Provider.

rebornix (approved):
Looks good!

PR #13675, opened 2023-06-08
Reduce Jupyter Session API in favour of JupyterLab API

DonJayamanne:


src/kernels/common/baseJupyterSession.ts
```diff
@@ -243,85 +208,6 @@ export abstract class BaseJupyterSession<T extends 'remoteJupyter' | 'localJupyt
...
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            this.session.kernel.sendInputReply(content);
-        }
-    }
-
-    public registerCommTarget(
-        targetName: string,
-        callback: (comm: Kernel.IComm, msg: KernelMessage.ICommOpenMsg) => void | PromiseLike<void>
-    ) {
-        if (this.session && this.session.kernel) {
```
DonJayamanne, line 297: These are just wrappers, the property `session.kernel` is actually a public property
Hence remove these wrappers in favour of directly using the `session.kernel`, which is a JupyterLab API.
I.e. fewer wrappers.

src/kernels/raw/session/rawJupyterSession.node.ts
```diff
@@ -50,18 +50,10 @@ export class RawJupyterSession extends BaseJupyterSession<'localRaw'> implements
...
         resource: Resource,
         workingDirectory: Uri,
-        interruptTimeout: number,
         kernelConnection: KernelConnectionMetadata,
         private readonly launchTimeout: number
     ) {
-        super('localRaw', resource, kernelConnection, workingDirectory, interruptTimeout);
-    }
-
-    public async waitForIdle(timeout: number, token: CancellationToken): Promise<void> {
```
DonJayamanne, line 60: This method is identical in both sub classes, hence moved into the base class

src/kernels/types.ts
```diff
@@ -571,31 +570,8 @@ export interface IBaseKernelSession<T extends 'remoteJupyter' | 'localJupyter' |
     onSessionStatusChanged: Event<KernelMessage.Status>;
     onDidDispose: Event<void>;
     onDidShutdown: Event<void>;
-    interrupt(): Promise<void>;
     restart(): Promise<void>;
     waitForIdle(timeout: number, token: CancellationToken): Promise<void>;
-    requestExecute(
```
DonJayamanne, line 577: No longer used, as these are wrappers around the `kernel` property

chrmarti (approved):

PR #13712, opened 2023-06-15
Engineering - update pipeline

lszomoru:


DonJayamanne (changes requested):
We AirAsia release from the release branches
This changes how we ship the extension
and blocks the main branch from being released

I don’t think we can go with this

DonJayamanne (approved):

PR #13725, opened 2023-06-18
Unify storage of Remote Uris

### LINKED ISSUE #11645: Unify remote kernel finder with ServerUriStorage
With the introduction of the new remote finder and server picker experience, we might want to revisit how we are registering kernels from remote uris and what we store in `ServerUriStorage`

* `ServerUriStorage` can contain following info
  * url list
    * user provided server urls 
    * third party contributed server urls
    * `local` when users explicitly set the server to `None` from the server picker
  * `currentServerId` referring to the currently picked entry
  * `localOnly` referring to if we should connect to local only (❓)
* Traditional workflow
  * `LocalKernelFinder` contributes all local kernels
  * `RemoteKernelFinder` checks `ServerUriStorage#currentServerId` and decides if/how to detect kernel specs/sessions
    * If the selected entry in `ServerUriStorage` is `local`, then it doesn't attempt to scan kernel specs/sessions.
* New workflow
  * `UniversalRemoteKernelFinderController` creates a `UniRemoteKernelFinder` for all entries
  * Each `UniRemoteKernelFinder` follows the same logic as `RemoteKernelFinder`, if its `uri` is `local`, it contributes empty kernel specs/sessions.

It would be great if we can move `local` out of the `ServerUriStorage` and we can simply the `RemoteKernelFinder` ctor and registrations into 

* `ServerUriStorage` only holds server uri infos
* Traditional remote kernel finder controller
  * Hold the `currentSelectedUri` info
  * Register/dispose remote kernel finder per `currentSelectedUri` update
* New remote kernel finder controller
  * Register/dispose all remote kernel finders per `ServerUriStorage#getUris`

cc @IanMatthewHuff @DonJayamanne

DonJayamanne:
Fixes #11645

### DIFF
--- a/src/kernels/jupyter/types.ts
+++ b/src/kernels/jupyter/types.ts
@@ -159,7 +159,6 @@ export interface IInternalJupyterUriProvider extends IJupyterUriProvider {
     readonly extensionId: string;
 }
 export type JupyterServerProviderHandle = {
-    extensionId: string;
     /**
      * Jupyter Server Provider Id.
      */
--- a/src/kernels/jupyter/types.ts
+++ b/src/kernels/jupyter/types.ts
@@ -223,7 +223,10 @@ export interface IJupyterServerUriStorage {
     remove(serverId: string): Promise<void>;
     clear(): Promise<void>;
     get(serverId: string): Promise<IJupyterServerUriEntry | undefined>;
-    add(jupyterHandle: { id: string; handle: JupyterServerUriHandle }): Promise<void>;
+    add(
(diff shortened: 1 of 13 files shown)


src/kernels/jupyter/types.ts
```diff
@@ -159,7 +159,6 @@ export interface IInternalJupyterUriProvider extends IJupyterUriProvider {
     readonly extensionId: string;
 }
 export type JupyterServerProviderHandle = {
-    extensionId: string;
```
DonJayamanne, line 162: Not required

chrmarti (approved):

PR #13736, opened 2023-06-20
Remove export specific methods form Session

### LINKED ISSUE #13577: Export Notebook for Web: call `JupyterLab#ContentManager` directly to simplify `JupyterKernelSession`
```typescript
export interface IJupyterKernelSession extends IBaseKernelSession<'remoteJupyter' | 'localJupyter'> {
    invokeWithFileSynced(contents: string, handler: (file: IBackupFile) => Promise<void>): Promise<void>;
    createTempfile(ext: string): Promise<string>;
    deleteTempfile(file: string): Promise<void>;
    getContents(file: string, format: Contents.FileFormat): Promise<Contents.IModel>;
}
```

All of the above methods in the interface are only used in one place and one context.
Hence we should try to avoid adding these methods to a very generic interface, 

Better to just expose the `ContentManager` and move all of the code into the Export or other layer rathe than keeping creation/deletion/reading of files in the `IJupyterKernelSession` class/interface.

**Suggestions:**
* Expose the `ContentManager` of `IJupyterSessionManager`, this way we do not need anything `IJupyterKernelSession`. 

After all, all of the above 4 methods on the session has nothing to do with the kernel session, but its about content management.

Also see https://github.com/microsoft/vscode-jupyter/issues/13578

DonJayamanne:
Fixes #13577


src/kernels/jupyter/connection/jupyterConnection.unit.test.ts
```diff
@@ -161,121 +166,79 @@ suite('Jupyter Connection', async () => {
                 verify(sessionManager.getRunningKernels()).once();
                 verify(sessionManager.dispose()).once();
             });
-
-            /* eslint-disable , @typescript-eslint/no-explicit-any */
-            suite('JupyterConnection', () => {
```
DonJayamanne, line 166: Moved to a separate file, these tests didn't belong here

src/kernels/jupyter/connection/jupyterConnection.ts
```diff
@@ -144,4 +185,197 @@ export class JupyterConnection {
...
+            fetch: this.requestCreator.getFetchMethod(),
+            Request: this.requestCreator.getRequestCtor(cookieString, allowUnauthorized, connInfo.getAuthHeader),
+            Headers: this.requestCreator.getHeadersCtor()
+        };
+
+        return this.jupyterlab.ServerConnection.makeSettings(serverSettings);
+    }
+
+    // If connecting on HTTP without a token prompt the user that this connection may not be secure
+    private async insecureServerWarningPrompt(): Promise<boolean> {
```
DonJayamanne, line 326: Moved from JupyterSessionManager
(unfortunately this introduces a cyclic dependency, but the plan is to fix this)

src/kernels/jupyter/session/jupyterSession.ts
```diff
@@ -196,53 +192,6 @@ export class JupyterSession
         return promise;
     }
 
-    async invokeWithFileSynced(contents: string, handler: (file: IBackupFile) => Promise<void>): Promise<void> {
```
DonJayamanne, line 199: Now moved into export class

src/kernels/jupyter/session/jupyterSessionManager.ts
```diff
@@ -339,190 +316,4 @@ export class JupyterSessionManager implements IJupyterSessionManager {
             return [];
         }
     }
-
-    private async getServerConnectSettings(connInfo: IJupyterConnection): Promise<ServerConnection.ISettings> {
```
DonJayamanne, line 343: Moved into JupyterConnection class

Issue #13889, opened 2023-07-10
Interactive Window sys info cell is not updated after connecting to kernel


amunger:
1. run a `# %%` cell from a python file
2. wait for the kernel to connect and cell to run
:bug: info message is still "connecting to..." and doesn't update

PR #13891, opened 2023-07-10
workaround to make cell update

amunger:
https://github.com/microsoft/vscode-jupyter/issues/13889

this is a workaround, as it looks like there is a bug in the `WorkspaceEdit.replace` API call: the edit only takes effect if it's editing text in the last cell of the notebook or if another edit is made afterward, as this PR does.


amunger:
core issue https://github.com/microsoft/vscode/issues/187643

sandy081 (approved):

Issue #13894, opened 2023-07-10
Review API `JupyterServerProvider` and prepare for stable API


rebornix:
This would be our exercise for our first stable API `JupyterServerProvider`
```[tasklist]
### Tasks
- [x] Use `CancellationError` and return `undefined` to distinguish between `cancel` and `back` (see how done in VS code Kernels)
```

DonJayamanne:
@mjbvz Please can you have another look at the API.
https://github.com/microsoft/vscode-jupyter/blob/main/src/api.proposed.d.ts

Note: I used `JupyterServerCommand` instead of  VS Codes built in `Command`.
The reason is, I need a detail that gets displayed in the quick pick, the VS Code command only has label and tooltip.

Here is a screenshot of what it looks like today, and as you can see the detail is pretty important (to let the user know about the specs of the server).
Unfortunately that text cannot be part of the `label` as the `label` is used as the name of the server which gets displayed elsewhere.


![Screenshot 2023-08-11 at 11 38 45](https://github.com/microsoft/vscode-jupyter/assets/1948812/24ee53d2-ea34-4e8e-9d4f-fdeabbe95759)

DonJayamanne:
This is where the discussions are taking place https://github.com/microsoft/vscode-jupyter/pull/14106

DonJayamanne:
API Discussion outcomes
* [ ] Use CancellationError and return undefined to distinguish between cancel and back (see how done in VS code Kernels)
* Commands is tricky
    * We would like to enforce a good UX (ensure 3rd party extension have back button etc)
    * The command API is very much tied to the UI

DonJayamanne:
Done

PR #13936, opened 2023-07-18
Review API JupyterServerProvider and prepare for stable API

DonJayamanne:
For #13894

### DIFF
--- a/src/proposed.api.d.ts
+++ b/src/proposed.api.d.ts
@@ -87,20 +87,17 @@ export interface JupyterServerCollection {
      * A link to a page providing more information to the user about this item.
      */
     documentation?: Uri;
-    onDidRemoveServer: Event<JupyterServer>;
     /**
      * Creates an entry in the list of Jupyter Servers from which the user can pick.
      *
      * @param {string} id
      * @param {string} label
      * @param {GetConnectionInformation} resolveConnectionInformation Gets the connection information for the Jupyter Server.
-     * @param {boolean} [supportsRemoval] Whether the user can remove this entry from the list of Jupyter Servers (allows users to manage the lifetime of the server).
      */
     createServer(
         id: string,
         label: string,
         resolveConnectionInformation: () => Promise<JupyterServerConnectionInformation>,
-        supportsRemoval?: boolean
     ): JupyterServer;
     /**
      * Creates an entry in the list of Jupyter Servers from which a user can pick to create a server.
--- a/src/proposed.api.d.ts
+++ b/src/proposed.api.d.ts
@@ -1,9 +1,10 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { Disposable, Event, Uri } from 'vscode';
+import { IKernelConnection } from '@jupyterlab/services/lib/kernel/kernel';
+import { Disposable, Event, Uri, extensions } from 'vscode';
 
-export interface JupyterServerAuthenticationInformation {
+export interface JupyterServerConnectionInformation {
     readonly baseUrl: Uri;
     /**
      * Jupyter auth Token
@@ -14,6 +15,22 @@ export interface JupyterServerAuthenticationInformation {
      */
     readonly authorizationHeader?: Record<string, string>;
     /**
+     * The local directory that maps to the remote directory of the Jupyter Server.
+     * E.g. assume you start Jupyter Notebook with --notebook-dir=/foo/bar,
+     * and you have a file named /foo/bar/sample.ipynb, /foo/bar/sample2.ipynb and the like.
+     * Then assume the mapped local directory will be /users/xyz/remoteServer and the files sample.ipynb and sample2.ipynb
+     * are in the above local directory.
+     *
+     * Using this setting one can map the local directory to the remote directory.
+     * In this case the value of this property would be /users/xyz/remoteServer.
+     *
+     * Note: A side effect of providing this value is the session names are generated the way they are in Jupyter Notebook/Lab.
+     * I.e. the session names map to the relative path of the notebook file.
+     * As a result when attempting to create a new session for a notebook/file, Jupyter will
+     * first check if a session already exists for the same file and same kernel, and if so, will re-use that session.
+     */
+    readonly mappedRemoteNotebookDir?: string;
+    /**
      * Returns the sub-protocols to be used. See details of `protocols` here https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket
      * Useful if there is a custom authentication scheme that needs to be used for WebSocket connections.
      * Note: The client side npm package @jupyterlab/services uses WebSockets to connect to remote Kernels.
@@ -21,41 +38,16 @@ export interface JupyterServerAuthenticationInformation {
     readonly webSocketProtocols?: string[];
 }
 
-export type GetAuthenticationInformation = () => Promise<JupyterServerAuthenticationInformation>;
-/**
- * The local directory that maps to the remote directory of the Jupyter Server.
- * E.g. assume you start Jupyter Notebook with --notebook-dir=/foo/bar,
- * and you have a file named /foo/bar/sample.ipynb, /foo/bar/sample2.ipynb and the like.
- * Then assume the mapped local directory will be /users/xyz/remoteServer and the files sample.ipynb and sample2.ipynb
- * are in the above local directory.
- *
- * Using this setting one can map the local directory to the remote directory.
- * In this case the value of this property would be /users/xyz/remoteServer.
- *
- * Note: A side effect of providing this value is the session names are generated the way they are in Jupyter Notebook/Lab.
- * I.e. the session names map to the relative path of the notebook file.
- * As a result when attempting to create a new session for a notebook/file, Jupyter will
- * first check if a session already exists for the same file and same kernel, and if so, will re-use that session.
- */
-export type GetRemoteNotebookDirectoryMapping = () => string;
-
 export interface JupyterServer extends Disposable {
-    /**
-     * Triggered when the server is removed by the user.
-     */
-    readonly onDidRemove: Event<void>;
     readonly id: string;
     /**
      * A human-readable string which is rendered prominent.
      */
     label: string;
-    /**
-     * Date time when the user last ran some code against a kernel on this server.
-     */
-    readonly lastActivity?: Date;
+    resolveConnectionInformation: () => Promise<JupyterServerConnectionInformation>;
 }
 
-export class JupyterServerPicker extends Disposable {
+export class JupyterServerCreationItem extends Disposable {
     /**
      * A human-readable string which is rendered prominent. Supports rendering of {@link ThemeIcon theme icons} via
      * the `$(<name>)`-syntax.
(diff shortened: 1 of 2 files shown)


src/proposed.api.d.ts
```diff
@@ -0,0 +1,99 @@
...
+     * Called when a server is to be removed from the list of servers.
+     */
+    removeServer(server: JupyterServer): Promise<void>;
+    /**
+     * Provides full control of the quick pick to the extension, so that it can add additional buttons, etc.
+     * The result of this promise should be the selected server or undefined if no server was selected.
+     *
+     * If nothing is selected, then the quick pick reverts back to the previous View before this method was invoked.
+     */
+    handleQuickPick?(quickPick: QuickPick<QuickPickItem>): ProviderResult<JupyterServer>;
```
DonJayamanne, line 81: Giving full access to the QuickPick is the easiest
* Today we have custom arguments for `back`, 
* custom handling for when something is selected we get a return value of the handle and then from that we get the server info
* We have custom handling for selecting the default quick pick item
* We have requirements to change this for user uri picker (https://github.com/microsoft/vscode-jupyter/issues/13892)

All of the above is very easy if we just provide the quick pick as is,
All we need to do is, once the promise resolves with `undefined`, then we need to restore the quick pick state back to the previous view
& thats easy
* We just update the items, selections, title, buttons, etc (basically the same as running some existing code that builds the quick pick)

DonJayamanne: For reference this is the existing API
* We can remove concept of `handles`, and have a simpler API

https://github.com/microsoft/vscode-jupyter/blob/285988fa81fd56598bffea06dfa86a5e2ab97a2f/src/api.d.ts#L74-L110

rebornix: We discussed offline about one solution to simplify the creation of `JupyterServer` by making `JupyterServer` a managed object by Jupyter extension.

* 3rd party extension calls `JupyterAPI.createServerProvider(id, displayName)` to get a managed `ServerProvider` object
* call `serverProvider.createServer(id, displayName, url, authHandler)` to create a `JupyterServer` for this provider. Once created, the server will be visible to users on the kernel picker source
* `JupyerServer` is an object managed by Jupyter extension. It can expose `JupyterServer.serverConnection` property, which is a JupyterLab api object.
* The `JupyterServer` can be removed by either the Jupyter extension or the 3rd party extension.

3rd party extension will use `ServerProvider.createServer` to indicate there are servers available, they no longer need to use `EventEmitters` (`onDidChangeServers`). The `JupyterAPI.addRemoteJupyterServer` can also be removed in favor of this.

src/proposed.api.d.ts
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Optional flag indicating if this item is picked by default.
+     * If there are no existing servers, and this flag is true, then this item will be picked by default.
+     *
+     * Note: this property is ignored when {@link JupyterServerCollection.createServer createJupyterServer} has been called.
+     */
+    picked?: boolean;
+}
+
+export class JupyterServerCollection extends Disposable {
```
mjbvz, line 75: Should users be able to create an instance of this? If not, mark the constructor private

src/proposed.api.d.ts
```diff
@@ -0,0 +1,131 @@
...
+     */
+    label: string;
+    /**
+     * A link to a page providing more information to the user about this item.
+     */
+    documentation?: Uri;
+    /**
+     * Creates an entry in the list of Jupyter Servers from which the user can pick.
+     *
+     * @param {string} id
```
mjbvz, line 91: [nit] Don't use types in JSDoc if you're writing typescript

mjbvz: You also don't need `memeberof`

src/proposed.api.d.ts
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * Using this setting one can map the local directory to the remote directory.
+     * In this case the value of this property would be /users/xyz/remoteServer.
+     *
+     * Note: A side effect of providing this value is the session names are generated the way they are in Jupyter Notebook/Lab.
+     * I.e. the session names map to the relative path of the notebook file.
+     * As a result when attempting to create a new session for a notebook/file, Jupyter will
+     * first check if a session already exists for the same file and same kernel, and if so, will re-use that session.
+     */
+    readonly mappedRemoteNotebookDir?: string;
```
mjbvz, line 31: Should this be a `Uri`?

src/proposed.api.d.ts
```diff
@@ -0,0 +1,131 @@
...
+export interface JupyterAPI {
+    /**
+     * Provides the ability to register multiple collections of Jupyter Servers.
+     *
+     * @param {string} id Identifier must be globally unique.
+     * @param {string} label
+     * @return {*}  {JupyterServerCollection}
+     * @memberof JupyterAPI
+     */
+    createServerCollection(id: string, label: string): Promise<JupyterServerCollection>;
```
mjbvz, line 130: Are collections scoped to the extension? For instance, what happens if two extensions use the same `id`?

src/proposed.api.d.ts
```diff
@@ -0,0 +1,131 @@
...
+export interface JupyterServer extends Disposable {
+    readonly id: string;
+    /**
+     * A human-readable string which is rendered prominent.
+     */
+    label: string;
+    resolveConnectionInformation: () => Promise<JupyterServerConnectionInformation>;
+}
+
+export class JupyterServerCreationItem extends Disposable {
```
mjbvz, line 49: Does this need to be a managed object? Or could it just be an interface

mjbvz (changes requested):
@DonJayamanne Can you share a code example of how you'd use this API?

My overall impression is that the flow of the API seems a little strange. Typically APIs like this fit into one of two models:

- A pull model. Callers register providers that the host then calls
- A push model. Callers call functions to modify a collection

This seems closer to the push model but I don't currently understand why all the `create*` functions are needed. It would help to know which objects are meant to be manage/live and which ones are simply meant to contain data

I'd also be interested to see if we could use a provider model instead as this gives the host more control

src/proposed.api.d.ts
DonJayamanne: Aah yes

src/proposed.api.d.ts
DonJayamanne: Managed, this allows the extension to remove this item by calling the `dispose`

src/proposed.api.d.ts
DonJayamanne: I don't manually type these, just use an extension that auto populates them.
will remove the types in the final *.d.ts file

src/proposed.api.d.ts
DonJayamanne: Yes they are scoped to the extension,
I discussed with Peng the old code requires it to be unique (i.e. if another extension uses the same id, then that could result in problems), but I guess thats a problem that we need to fix.

So, lets go with unique per extension, which is my preference.
However, the challenge here is we need to know the caller (extension), today we get this information by analyzing the stack trace.

src/proposed.api.d.ts
DonJayamanne: Sure, thing. I think i will turn them into interfaces, that way its clear that users cannot create them directly

DonJayamanne: Turned into interfaces as users cannot create the classes directly

DonJayamanne:
@mjbvz 

* I have turned all claases into interfaces
* Here are some screenshots and link to the sample extension

### This is the standard kernel picker (built by VS Code)
Screenshot 2023-07-20 at 13 27 35

### From previous quick pick user selects the option `Select Another` kernel (this is provided by vscode)
The list of items here is contributed by extensions.
Since other extensions can contribute Jupyter Servers to the Jupyter extension they contribute their `collection` with us, and we add an entry into this VS Code list
So an entry in this item is called a collection, as each item can in turn have multiple servers (see asmples befolow).
Here the second from last `Local Jupyter Server` is contributed to the Jupyter extension via the api `createServerCollection('Local Jupyter Server')`


Screenshot 2023-07-20 at 13 27 41

### Once the user selects the `Local Jupyter Server`, we get to this screen.
At this point jupyter extension will look for servers contributed by the 3rd party extension, if there are we display them.  
However for the first time its unlikely a server would exist, hence the 3rd party extension needs a way to get the user to start/select a server.  
For this they can add items into the quick pick that create the servers.  
This is done by 3d party calling the api as follows
```typescript
collection.createServerCreationItem('', 'Start New Jupyter Lab Server, ()=> {....});
collection.createServerCreationItem('', 'Start New Jupyter Notebook Server, ()=> {....});
```

If there is only one item in the list, the 3rd party extension can mark that has `picked` and we (Jupyter extension) will bypass displaying this list and automatically select that first item.
See below, we have dispalyed the two items contributed by the 3rd party

Screenshot 2023-07-20 at 13 27 47


### Upon selecting on of the above items the Jupyter extension calls the callback
The 3rd party handles the callback, and displays their own UI (AzML, Synapse and other extensions do this today, and display their own quick picks).   
In this sample  I have a custom quick pick with options (this is from 3rd party extension)
The UI can be as complex as the 3rd party needs it to be (AzML has 3 steps in their own quick pick)  
So here once we select something we start a Jupyter server 
The 3rd party extension will then call `collection.createServer(<Display Name>)` and return that server from the callback.

If the Jupyter extension gets a server retrurned from the callback, we know that a server was created and we display that server in the next quick pick and we display the list of the kernels.



Screenshot 2023-07-20 at 13 27 58

### Here the Jupyter extension got the server via the callback and we now dispaly that server and its kernels
Screenshot 2023-07-20 at 13 29 08

### If we hit the back buttoon, then we get to the same screen but this time we have the servers listed along with the creation items

At this point the 3rd party extension can optionally remove the creation items if they want
They can dispose the `Start Local Jupyter Server` as they know one has been started (after all 3rd party extensions have created the server via the api call)
Hence the reason to expose the `dispose` on JupyterServerCreationItem` (this way they can add/remove anytime)

If you were to go back to the first step in the kernel picker and select `Local Jupyter Server` then you get the same list of items 

Screenshot 2023-07-20 at 13 28 22

### Another sample if we have more than one server.
As we can have multiple servers contributed by the 3rd party extension, I chose the name `ServerCollection`.
AzML too has similar concepts

Screenshot 2023-07-20 at 13 29 29

### `supportsRemoval` argument for `createServer`

@rebornix I added this as we missed this item.

If a 3rd party extension supports shutting down a jupyter server they can let Jupyter extension know about this.
If this is supported, then Jupyter extension displays the little `Delete` icon against each server.
See below.   
Note: this icon is only displayed when the flag is true, and if the user clicks this icon, then we support removing such items and we trigger the `onDidRemoveServer` event, and 3rd party extensions can handle that event knowing the fact that the user has removed it from the list.   
At this point 3rd party extension can shutdown/dispose the servers as user is no longer interested in seeing  them.

Screenshot 2023-07-20 at 13 29 29

mjbvz:
Thanks @DonJayamanne! Based on your description of the flow, it seems like a pull model may be a better fit. Have you consider a more "provider like" approach?

DonJayamanne:
@mjbvz thanks, we did have a provider style API earlier (with quick picks), will try to get that again and pass it your way.
thanks again for your feedback.

DonJayamanne:
hi @mjbvz we decided to go with a push model as the UI and requirements are similar to notebook controllers.
In Notebook controllers we had a pull (provider) model and then moved to a push model, going with the same here.

Hence prefer to go with the push, however I have updated this PR by adding another file with the pull (provider) model API. See here https://github.com/microsoft/vscode-jupyter/blob/86da8624cbff9a9cee0844642b848bd4637472bf/src/api.proposed.d.ts

Will bring this to the next API discussion.

DonJayamanne:
Closing this PR as most of the issues ahve been addressed.

Issue #13999, opened 2023-07-28
Bug in Jupyter extension for VS Code


saga-gnassingbe:
### Applies To

- [X] Notebooks (.ipynb files)
- [ ] Interactive Window and\/or Cell Scripts (.py files with \#%% markers)

### What happened?

I have updated the Jupyter extension for VS Code to the latest version (July 2023). VS Code fails to launch the extension. Older versions (e.g. November 2022) of this extension can launched successfully.

Main error message:

[ms-toolsai.jupyter]Invalid regular expression: **/SagaGnassingbe(Sigma/**: Unterminated group
$onExtensionRuntimeError @ mainThreadExtensionService.ts:81
mainThreadExtensionService.ts:82 SyntaxError: Invalid regular expression: **/SagaGnassingbe(Sigma/**: Unterminated group
	at new RegExp (<anonymous>)

### VS Code Version

Version: 1.80.2 (system setup) Commit: 2ccd690cbff1569e4a83d7c43d45101f817401dc Date: 2023-07-27T20:40:28.909Z Electron: 22.3.14 ElectronBuildId: 22695494 Chromium: 108.0.5359.215 Node.js: 16.17.1 V8: 10.8.168.25-electron.0 OS: Windows_NT x64 10.0.22621

### Jupyter Extension Version

v2023.6.1101941928

### Jupyter logs

```shell
workbench.desktop.main.js:sourcemap:627  INFO [perf] Render performance baseline is 53ms
workbench.desktop.main.js:sourcemap:762 Extension activation failed, run the 'Developer: Toggle Developer Tools' command for more information.
c @ workbench.desktop.main.js:sourcemap:762
workbench.desktop.main.js:sourcemap:627   ERR [Extension Host] extension activation failed Error: No matching bindings found for serviceIdentifier: Symbol(IExtensions)
	at c:\Users\SagaGnassingbe_Sigma\.vscode\extensions\ms-toolsai.jupyter-2023.6.1101941928-win32-x64\out\extension.node.js:2:534020
	at b (c:\Users\SagaGnassingbe_Sigma\.vscode\extensions\ms-toolsai.jupyter-2023.6.1101941928-win32-x64\out\extension.node.js:2:534309)
	at w (c:\Users\SagaGnassingbe_Sigma\.vscode\extensions\ms-toolsai.jupyter-2023.6.1101941928-win32-x64\out\extension.node.js:2:534396)
[...]

PR #14001, opened 2023-07-28
more within try block. escape regex special chars

amunger:
Fixes https://github.com/microsoft/vscode-jupyter/issues/13999

### DIFF
--- a/src/extension.node.ts
+++ b/src/extension.node.ts
@@ -243,9 +243,14 @@ function addConsoleLogger() {
     }
 }
 
+function escapeRegExp(text: string) {
+    return text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
+}
+
 function tryGetUsername() {
     try {
-        return userInfo().username;
+        const username = escapeRegExp(userInfo().username);
+        return new RegExp(username, 'ig');
     } catch (e) {
         console.info(
             `jupyter extension failed to get username info with ${e}\n username will not be obfuscated in local logs`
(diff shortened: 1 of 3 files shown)


src/extension.node.ts
```diff
@@ -243,9 +243,14 @@ function addConsoleLogger() {
...
 
+function escapeRegExp(text: string) {
+    return text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
+}
+
 function tryGetUsername() {
     try {
-        return userInfo().username;
+        const username = escapeRegExp(userInfo().username);
+        return new RegExp(username, 'ig');
```
Tyriar, line 253: This seems pretty fragile, can we use indexOf/contains instead?

Tyriar (changes requested):

src/extension.node.ts
amunger: to escape each occurrence of a regex special character? That sounds more fragile

Tyriar: I mean do plain string comparisons without touching regex at all. If you replace one of these special characters isn't this logic going to fall over? Posix paths can have many more characters than Windows paths for example which could be part of `$HOME`.

amunger: I think it would still work after escaping the character, I'm not removing them.
But yeah, I can try to manually find and replace all occurrences, still doubt it will be less fragile than just using regex.

Tyriar: Up to you, I approved if you don't think it's worth it

Tyriar (approved):

src/extension.node.ts
amunger: I'll let @DonJayamanne take a look, since he actually owns the logging area

PR #14037, opened 2023-08-02
New Python Environment Picker without using any of the old Python API types

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/14039

### DIFF
--- a/src/platform/common/providerBasedQuickPick.ts
+++ b/src/platform/common/providerBasedQuickPick.ts
@@ -17,7 +17,7 @@ import {
 } from 'vscode';
 import { InputFlowAction } from './utils/multiStepInput';
 import { Disposables } from './utils';
-import { Common, DataScience } from './utils/localize';
+import { Common } from './utils/localize';
 import { noop } from './utils/misc';
 
 abstract class BaseQuickPickItem implements QuickPickItem {
@@ -46,14 +46,14 @@ export class SelectorQuickPickItem<T extends { id: string }> extends BaseQuickPi
 export interface IQuickPickItemProvider<T extends { id: string }> {
     readonly title: string;
     onDidChange: Event<void>;
-    onDidFail: Event<Error>;
+    // onDidFail: Event<Error>;
     onDidChangeStatus: Event<void>;
-    onDidChangeRecommended: Event<void>;
-    onDidChangeSelected: Event<void>;
+    // onDidChangeRecommended: Event<void>;
+    // onDidChangeSelected: Event<void>;
     readonly items: T[];
     readonly status: 'discovering' | 'idle';
-    readonly recommended?: T;
-    readonly selected?: T;
+    // readonly recommended?: T;
+    // readonly selected?: T;
     refresh: () => Promise<void>;
     toQuickPick(item: T): SelectorQuickPickItem<T>;
     getCategory(item: T): { label: string; sortKey?: string };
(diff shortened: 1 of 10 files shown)


src/platform/interpreter/helpers.ts
```diff
@@ -34,3 +49,26 @@ export function getPythonEnvironmentName(pythonEnv: PythonEnvironment) {
...
+export function getEnvironmentType(env: Environment): EnvironmentType {
+    for (const type of environmentTypes) {
+        if (env.tools.some((tool) => tool.toLowerCase() === type.toLowerCase())) {
+            return type;
+        }
+    }
+    return EnvironmentType.Unknown;
+}
+
+export function isCondaEnvironmentWithoutPython(env: Environment) {
```
DonJayamanne, line 72: Support for the new Python API types

src/platform/interpreter/pythonEnvironmentPicker.node.ts
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT License.
+
+import { CancellationToken } from 'vscode';
+import { Environment } from '../api/pythonApiTypes';
+import { BaseProviderBasedQuickPick } from '../common/providerBasedQuickPick';
+import { ServiceContainer } from '../ioc/container';
+import { PythonEnvironmentQuickPickItemProvider } from './pythonEnvironmentQuickPickProvider.node';
+
+export class PythonEnvironmentPicker extends BaseProviderBasedQuickPick<Environment> {
+    constructor(options: {
```
DonJayamanne, line 11: This class deals with just the new API

src/platform/common/providerBasedQuickPick.ts
```diff
@@ -0,0 +1,407 @@
...
+        quickPick.onDidTriggerButton(
+            async (e) => {
+                if (e === refreshButton) {
+                    const oldButtons = quickPick.buttons;
+                    quickPick.buttons = quickPick.buttons
+                        .filter((btn) => btn !== refreshButton)
+                        .concat(refreshingButton);
+                    await this.provider.refresh().catch(noop);
+                    quickPick.buttons = oldButtons;
+                }
```
alexr00, line 113: Is this replacing the `loading` icon button with the `loading~spin` icon? If so, is this a common pattern that we want to use? I think just `busy = true` is what we suggest all extension use. @TylerLeonhardt in case you want to chime in as the quick pick owner.

alexr00: Pattern aside, it might make more sense to use `map` to preserve button order: 

```suggestion
                if (e === refreshButton) {
                    const oldButtons = quickPick.buttons;
                    quickPick.buttons = quickPick.buttons
                        .map((btn) => btn !== refreshButton ? btn : refreshingButton);
                    await this.provider.refresh().catch(noop);
                    quickPick.buttons = oldButtons;
                }
```

alexr00 (dismissed):

src/platform/common/providerBasedQuickPick.ts
DonJayamanne: Agreed makes sense, will use `busy=true`
Not sure why we did this in the past (this is code that I copied)
Will merge this PR and makde that change in a smaller PR, thanks

DonJayamanne:
@alexr00 updated to remove the funky button in favour of `busy = true`

alexr00 (approved):

PR #14081, opened 2023-08-08
Avoid validation of Jupyter Server providers

DonJayamanne:
For #13894

### DIFF
--- a/src/kernels/jupyter/finder/remoteKernelFinderController.ts
+++ b/src/kernels/jupyter/finder/remoteKernelFinderController.ts
@@ -21,7 +21,7 @@ import { JupyterConnection } from '../connection/jupyterConnection';
 import { IFileSystem } from '../../../platform/common/platform/types';
 import { ContributedKernelFinderKind } from '../../internalTypes';
 import { generateIdFromRemoteProvider } from '../jupyterUtils';
-import { traceWarning } from '../../../platform/logging';
+import { swallowExceptions } from '../../../platform/common/utils/decorators';
 
 @injectable()
 export class RemoteKernelFinderController implements IExtensionSyncActivationService {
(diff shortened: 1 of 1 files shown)


src/kernels/jupyter/connection/serverUriStorage.ts
```diff
@@ -372,16 +370,7 @@ class NewStorage {
                     displayName: item.displayName || uri,
                     provider: item.serverHandle
                 };
-                if (!validate) {
-                    entries.push(server);
-                    return;
-                }
-                try {
-                    await this.jupyterPickerRegistration.getJupyterServerUri(item.serverHandle, true);
```
DonJayamanne, line 380: This is only required in one calling place, removed this and move the validation to the calling code.
Thereby removing a cycling reference and more layers.
This also makes listing of Jupyter Server much faster, as we no longer validate each server, which in turn waits for all related extensions to completely activate.

src/kernels/jupyter/connection/serverUriStorage.ts
```diff
@@ -350,17 +334,17 @@ class NewStorage {
             })
             .catch(noop));
     }
-    public async getAll(validate = true): Promise<IJupyterServerUriEntry[]> {
-        return this.getAllImpl(validate).then((items) => items.sort((a, b) => b.time - a.time));
+    public async getAll(): Promise<IJupyterServerUriEntry[]> {
+        return this.getAllImpl().then((items) => items.sort((a, b) => b.time - a.time));
     }
     public async clear(): Promise<void> {
-        const all = await this.getAllImpl(false);
```
DonJayamanne, line 357: `false` is the only value supported now.

src/kernels/jupyter/connection/serverUriStorage.unit.test.ts
```diff
@@ -219,35 +218,30 @@ suite('Server Uri Storage', async () => {
                 .sort((a, b) => a.time - b.time)
                 .map((a) => {
                     return {
-                        displayName: a.displayName,
```
DonJayamanne, line 222: `displayName` is deprecated

src/kernels/jupyter/types.ts
```diff
@@ -184,6 +184,7 @@ export interface IJupyterServerUriEntry {
     time: number;
     /**
      * An optional display name to show for this server as opposed to just the Uri
+     * @deprecated Used only for migration of display names into the User Provided Server list. Else other providers will have the Display Names.
```
DonJayamanne, line 187: `displayName` is deprecated, only used for migration of data from one storage into another

lramos15 (approved):

PR #14101, opened 2023-08-11
Remove remote Kernel finder provider

DonJayamanne:
For #13894

### DIFF
--- a/src/kernels/internalTypes.ts
+++ b/src/kernels/internalTypes.ts
@@ -11,7 +11,7 @@ export enum ContributedKernelFinderKind {
 }
 
 export interface IContributedKernelFinder<T extends KernelConnectionMetadata = KernelConnectionMetadata> {
-    status: 'discovering' | 'idle';
+    readonly status: 'discovering' | 'idle';
     onDidChangeStatus: Event<void>;
     /**
      * Last error thrown when listing the kernels.
--- a/src/kernels/jupyter/finder/remoteKernelFinder.ts
+++ b/src/kernels/jupyter/finder/remoteKernelFinder.ts
@@ -79,6 +79,8 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IDisposable {
         removed?: { id: string }[];
     }>();
     onDidChangeKernels = this._onDidChangeKernels.event;
+    private readonly _onDidChange = new EventEmitter<void>();
+    onDidChange = this._onDidChange.event;
 
     private readonly disposables: IDisposable[] = [];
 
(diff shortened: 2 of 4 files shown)


src/kernels/internalTypes.ts
```diff
@@ -11,7 +11,7 @@ export enum ContributedKernelFinderKind {
 }
 
 export interface IContributedKernelFinder<T extends KernelConnectionMetadata = KernelConnectionMetadata> {
-    status: 'discovering' | 'idle';
+    readonly status: 'discovering' | 'idle';
     onDidChangeStatus: Event<void>;
```
Tyriar, line 15: readonly too?

src/kernels/jupyter/finder/remoteKernelFinder.ts
```diff
@@ -79,6 +79,8 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IDisposable {
         removed?: { id: string }[];
     }>();
     onDidChangeKernels = this._onDidChangeKernels.event;
+    private readonly _onDidChange = new EventEmitter<void>();
+    onDidChange = this._onDidChange.event;
```
Tyriar, line 83: readonly?

src/kernels/jupyter/finder/remoteKernelFinder.ts
```diff
@@ -79,6 +79,8 @@ export class RemoteKernelFinder implements IRemoteKernelFinder, IDisposable {
         removed?: { id: string }[];
     }>();
     onDidChangeKernels = this._onDidChangeKernels.event;
+    private readonly _onDidChange = new EventEmitter<void>();
```
Tyriar, line 82: You should be registering all events to the disposable, otherwise they could result in leaked memory when disposed.

Tyriar (changes requested):

src/kernels/jupyter/finder/remoteKernelFinder.ts
DonJayamanne: Already done, see Line 126, these are registered in the disposable,

src/kernels/internalTypes.ts
DonJayamanne: Yes, but these are old interfaces/code that is slowly being replaced by another interface, in that all are readonly and this stats is the only overlapping property, and I had to make this reaodnly to ensure those two interfaces are compatible.

Basically not too keen on adding `readonly` to all of the others as its pointless as they will be phased out and removed.
Hope thats ok.

src/kernels/jupyter/finder/remoteKernelFinder.ts
DonJayamanne: Please see previous comment

Tyriar (approved):
> Already done, see Line 126, these are registered in the disposable,

Oh I see, I find it much easier to avoid mistakes when the object extends a disposable and then the emitter initialization line can include its lifecycle management:

```ts
... = this.register(new EventEmitter())
```

That's a preference though, looks like you have my problem covered 👌

DonJayamanne:
> Oh I see, I find it much easier to avoid mistakes when the object extends a disposable and then the emitter initialization line can include its lifecycle management:

Oh yes, thats a neat solution, yes like that, will use that pattern in the future, 
Totally agree about benefit avoiding mistakes.

PR #14106, opened 2023-08-11
Not to be merged, for API review discussion only

DonJayamanne:


src/api.proposed.x.d.ts
```diff
@@ -0,0 +1,167 @@
...
+    export interface JupyterServerCommandProvider {
+        /**
+         * Default command to be used when there are no servers. This can be read and updated by the extension.
+         * If not set, and there are not servers, then the user will be prompted to select a command from a list of commands returned by `getCommands`.
+         */
+        selected?: JupyterServerCommand;
+        /**
+         * Returns a list of commands to be displayed to the user.
+         */
+        getCommands(token: CancellationToken): Promise<JupyterServerCommand[]>;
```
DonJayamanne, line 133: @mjbvz Please can you have another look at the API.

Note: I used `JupyterServerCommand` instead of  VS Codes built in `Command`.
The reason is, I need a detail that gets displayed in the quick pick, the VS Code command only has label and tooltip.

Here is a screenshot of what it looks like today, and as you can see the detail is pretty important (to let the user know about the specs of the server).
Unfortunately that text cannot be part of the `label` as the `label` is used as the name of the server which gets displayed elsewhere.


Screenshot 2023-08-11 at 11 38 45

DonJayamanne: Open to suggestions if using a derivative of `Command` is not right.

src/api.proposed.x.d.ts
```diff
@@ -0,0 +1,167 @@
...
+        /**
+         * Removes this Server Collection.
+         */
+        dispose(): void;
+    }
+    export interface JupyterAPI {
+        /**
+         * Creates a Jupyter Server Collection that can be displayed in the Notebook Kernel Picker.
+         */
+        createJupyterServerCollection(id: string, label: string): Promise<JupyterServerCollection>;
```
mjbvz, line 165: Does this need to return a promise?

src/api.proposed.x.d.ts
```diff
@@ -0,0 +1,167 @@
...
+     * - JupyterServer  : The Jupyter Server object that was created
+     * - 'back'         : Go back to the previous screen
+     * - undefined|void : Do nothing
+     */
+    export interface JupyterServerCommandProvider {
+        /**
+         * Default command to be used when there are no servers. This can be read and updated by the extension.
+         * If not set, and there are not servers, then the user will be prompted to select a command from a list of commands returned by `getCommands`.
+         */
+        selected?: JupyterServerCommand;
```
mjbvz, line 129: Is there a better name than `selected`? I'm finding the name confusing even after reading the docs

Also having both `selected` and `getCommands` is confusing me. Could `getCommands` handle both cases?

src/api.proposed.x.d.ts
```diff
@@ -0,0 +1,167 @@
...
+        readonly id: string;
+        /**
+         * A human-readable string representing the collection of the Servers. This can be read and updated by the extension.
+         */
+        label: string;
+        /**
+         * A link to a resource containing more information. This can be read and updated by the extension.
+         */
+        documentation?: Uri;
+        /**
```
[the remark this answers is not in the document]
DonJayamanne: I went with a read write as thats what we have done with notebook controllers where the details can be changed, will go with a readonly approach and accept these during creation.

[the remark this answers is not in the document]
DonJayamanne: Kept them as is, felt weird to have some properties read/write and other readonly.
Eg. the provider properties are read/write.

src/api.proposed.x.d.ts
```diff
@@ -0,0 +1,171 @@
...
+     * Each command allows the user to perform an action.
+     */
+    export interface JupyterServerCommandProvider {
+        /**
+         * Returns a list of commands to be displayed to the user.
+         * If there are no JupyterServers and one of the returned commands has `picked = true`,
+         * then the `handleCommand` method will be invoked with that command.
+         * @param options Reserved for future use. Extensions can ignore this argument for now.
+         */
+        getCommands(options: unknown, token: CancellationToken): Promise<JupyterServerCommand[]>;
```
DonJayamanne, line 121: @jrieken @mjbvz 
We have a proposed API on top of this.
the Proposed API is only for use by the Jupyter extnsion and another extension that will be similar to Jupyter extension.

The first argument will be `value:string`, which represents the value typed by the user in the quick pick, that
allows the 3rd party extension to change the contributed commands based on the values entered in the quik pick

Basically we're after the ability to dynamically populate quick pick items, just as we do with `Open Remote Repository` quick pick, see screenshots below.

Once again, the proposed API is only for internal use, right now we do not plan on exposing this for others.
However if we did, want to (@mjbvz suggested that it might be useful for other extension), then we have the ability to do so and have a clean api. 
All this to ensure the `token` is the last argument

Screenshot 2023-08-15 at 20 56 22
Screenshot 2023-08-15 at 20 56 28

jrieken: > Basically we're after the ability to dynamically populate quick pick items, just as we do with Open Remote Repository quick pick, see screenshots below.

I believe a difference is that the remote repo picker can not list all remotes, that's why it reacts to what's being typed (so that it can query GH as you type). Do you envision a similar use-case and is that generic to all commands or just one?

src/api.proposed.x.d.ts
```diff
@@ -0,0 +1,171 @@
...
+         * Default command to be used when there are no servers.
+         * If not set, and there are not servers, then the user will be prompted to select a command from a list of commands returned by `getCommands`.
+         */
+        picked?: boolean;
+    }
+    /**
+     * Provider of Jupyter Server Commands.
+     * Each command allows the user to perform an action.
+     */
+    export interface JupyterServerCommandProvider {
```
jrieken, line 114: I have trouble understanding why there is a need for a command provider and why it isn't just `JupyterServerCollection#commands: vscode.Command[]`. These commands can then do whatever is needed to establish a new server connection, be that nested quick picks or something else. 

Do you foresee these command to be changing a lot and be different each time the UI shows?

jrieken (commented):
left some nit about wording for providers and I need some clarification about the commands here. This seems tightly coupled with things I don't understand yet or haven't seen yet

src/api.proposed.x.d.ts
DonJayamanne: No, I don’t see this changing very often,
It can change after users select a command, at least that’show code spaces seems to work


> believe a difference is that the remote repo picker can not list all remotes, that's why it reacts to what's being typed (so that it can query GH as you type). Do you envision a similar use-case and is that generic to all commands or just one?

Right now, I see this being specific for just one command, and it’s specific to Jupyter extension, hence happy to leave it as an internal api/implementation that won’t be used by others. 
Meaning i think leaving commands as a property, as suggested is totally fine

> JupyterServerCollection#commands: vscode.Command[].

one problem with ‘Command’ type is the fact that there’s no way to tell us which one is selected, we need a way for extensions to tell us which item is to be selected by default in some cases

the only use case is, if there’s only one command item and no servers and that command item has it’s ‘picked’ property set to ‘true’ then we automatically select that instead of displaying the quick pick to the user,

this is why I went with another type named JupyterServerCommand

The other reason for this type is to have a strong return type documented
When commands are handled 
we have a very specific return type, the return type when a command is handled and executed has three states, 
- user selected a server (hence server returned)   
- user wishes to go back to the previous ui (hit a back button in their extension ui)
- User hits esc key
- 

Spoke to @mjbvz and his suggestion was to go with tours approach which I like, but happy to change

src/api.proposed.x.d.ts
```diff
@@ -0,0 +1,171 @@
...
+        /**
+         * A human-readable string representing the name of the Server. This can be read and updated by the extension.
+         */
+        label: string;
+    }
+
+    /**
+     * Provider of Jupyter Servers.
+     */
+    export interface JupyterServerProvider {
```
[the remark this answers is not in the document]
DonJayamanne: > nit but also important:

Agreed, should stick to standards. my bad, I should have checked the existing naming convenstions.

src/api.proposed.x.d.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * Represents a Jupyter Server displayed in the list of Servers.
+     */
+    export interface JupyterServer {
+        /**
+         * Unique identifier for this server.
+         */
+        readonly id: string;
+        /**
```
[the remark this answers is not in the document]
DonJayamanne: Got it.

src/api.proposed.x.d.ts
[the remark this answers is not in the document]
DonJayamanne: Created a separate type `ResolvedJupyterServer` to ensure the optional properties are made mandatory.
Got inspiration from the `resolveEnvironment` from Python Extension.

src/api.proposed.x.d.ts
DonJayamanne: Hopefully the following is a better explanation of why we need this.
when a command is selected, then we need to call back into the 3rd party extension and the return value of the called function has the following signature

```typescript
        /**
         * Invoked when a command has been selected.
         * @param command The command selected by the user.
         * @returns
         * - JupyterServer  : The Jupyter Server object that was created
         * - 'back'         : Go back to the previous UI in the workflow
         * - undefined|void : Do nothing
         */
        handleCommand(
            command: JupyterServerCommand,
            token: CancellationToken
        ): Promise<JupyterServer | 'back' | undefined | void>;

```

Here's a sample UI displayed by the 3rd party extension when a command is selected.
As you can see there are 3 things users can do:
* Click back button (at that point, control should come back to the Jupyter extension and we need to re-display the same quick pick)
* User can hit `esc` key
* Or user can go through the custom UI and evntually that will yield a Jupyter Server.
Screenshot 2023-08-18 at 14 40 14

DonJayamanne:
Closing this as most of the issues have been addressed

PR #14107, opened 2023-08-11
Utility method in base disposable class

DonJayamanne:


src/platform/common/utils.ts
```diff
@@ -425,6 +425,10 @@ export class Disposables implements IDisposable {
     constructor(...disposables: IDisposable[]) {
         this.disposables.push(...disposables);
     }
+    protected registerDisposable<T extends IDisposable>(disposable: T): T {
```
Tyriar, line 428: Making this `register` (like VS Code's `Disposable._register`) or `add` (like VS Code's `DisposableStore.add`) will make things a lot more concise

Tyriar (changes requested):

src/platform/common/utils.ts
DonJayamanne: Thats why i couldn't find it, i looked for `register` instead of `_register`

Tyriar (approved):

Issue #11750, opened 2022-10-23
Navigate to file/line when clicking on traceback


RossBencina:
## Environment Data
VS Code version: 1.72.0
Unified Python Extension version: v2022.16.1
Jupyter Extension version: v2022.9.1202862440
Python Version: 3.10.7 64-bit
OS and version: Windows 10 10.0.19043

## Actual behavior

When running a cell in the VSCode IPython Notebook editor, if a traceback is printed, you can't click on an item in order to navigate to the matching file/line.

## Expected behavior
Should navigate when clicking on file/line entries in traceback.

## Steps to reproduce

1. Create a file `buggy.py` containing the following code:
```
 def trigger_bug():
        a = 1
        b = 2
        c = 0
        return (a+b)/c
```
2. Create a new notebook `runbug.ipynb`, enter the following code in a cell, and run the cell:
```
import buggy
buggy.trigger_bug()
```

3. A traceback is printed. Try to click (or ctrl-click) on a file/line indicator inside a traceback. Nothing happens.

## Related

This feature was previously added for clicking on tracebacks when running Python scripts in the terminal window. See: https://github.com/microsoft/vscode-python/issues/608

amunger:
this was previously implemented but the changes in the latest ipykernel versions have broken the functionality and made it more difficult

PR #14109, opened 2023-08-11
loosen regex to find cell/line location when creating ref link

amunger:
Fixes https://github.com/microsoft/vscode-jupyter/issues/11750

### DIFF
--- a/src/notebooks/outputs/tracebackFormatter.ts
+++ b/src/notebooks/outputs/tracebackFormatter.ts
@@ -58,7 +58,7 @@ export class NotebookTracebackFormatter implements ITracebackFormatter {
 
             // Then replace the input line with our uri for this cell: '<cell line: 2>[0;34m()[0m\n'
             const cellAt = DataScience.cellAtFormat(getFilePath(cell.document.uri), cell.index + 1);
-            return afterLineReplace.replace(/.*?\n/, `\u001b[1;32m${cellAt}\u001b[0m in \u001b[0;36m${line}\n`);
+            return afterLineReplace.replace(/.*?\n/, `\u001b[1;32m${cellAt}\u001b[0m line \u001b[0;36m${line}\n`);
         };
 
         const inputMatch = /^Input.*?\[.*32mIn\s+\[(\d+).*?0;36m(.*?)\n.*/.exec(traceFrame);
@@ -66,7 +66,7 @@ export class NotebookTracebackFormatter implements ITracebackFormatter {
             return tracebackLinkify(traceFrame, inputMatch[2]);
         }
 
-        const cellMatch = /Cell.*?\[.*32mIn\s+\[(\d+)\]\,\s+line\s+([(\d+)])(.*?)\n.*/gm.exec(traceFrame);
+        const cellMatch = /Cell.*?\[.*32mIn\s?\[(\d+)\]\,\s+line\s+([(\d+)])(.*?)\n.*/gm.exec(traceFrame);
         if (cellMatch && cellMatch.length > 1) {
             return tracebackLinkify(traceFrame, cellMatch[2]);
         }


src/notebooks/outputs/tracebackFormatter.ts
```diff
@@ -58,15 +58,15 @@ export class NotebookTracebackFormatter implements ITracebackFormatter {
 
             // Then replace the input line with our uri for this cell: '<cell line: 2>[0;34m()[0m\n'
             const cellAt = DataScience.cellAtFormat(getFilePath(cell.document.uri), cell.index + 1);
-            return afterLineReplace.replace(/.*?\n/, `\u001b[1;32m${cellAt}\u001b[0m in \u001b[0;36m${line}\n`);
+            return afterLineReplace.replace(/.*?\n/, `\u001b[1;32m${cellAt}\u001b[0m line \u001b[0;36m${line}\n`);
```
amunger, line 61: I assume in was a typo, line makes much more sense

src/notebooks/outputs/tracebackFormatter.ts
```diff
@@ -58,15 +58,15 @@ export class NotebookTracebackFormatter implements ITracebackFormatter {
...
         };
 
         const inputMatch = /^Input.*?\[.*32mIn\s+\[(\d+).*?0;36m(.*?)\n.*/.exec(traceFrame);
         if (inputMatch && inputMatch.length > 1) {
             return tracebackLinkify(traceFrame, inputMatch[2]);
         }
 
-        const cellMatch = /Cell.*?\[.*32mIn\s+\[(\d+)\]\,\s+line\s+([(\d+)])(.*?)\n.*/gm.exec(traceFrame);
+        const cellMatch = /Cell.*?\[.*32mIn\s?\[(\d+)\]\,\s+line\s+([(\d+)])(.*?)\n.*/gm.exec(traceFrame);
```
joyceerhl, line 69: Should this be 
```suggestion
        const cellMatch = /Cell.*?\[.*32mIn\s*\[(\d+)\]\,\s+line\s+([(\d+)])(.*?)\n.*/gm.exec(traceFrame);
```

The previous regex afaict allows for one or more whitespace characters, but this makes it zero or one

src/notebooks/outputs/tracebackFormatter.unit.test.ts
```diff
@@ -46,7 +46,7 @@ suite(`Notebook trace formatter`, function () {
         const expected = [
             '[0;31m---------------------------------------------------------------------------[0m',
             '[0;31mNameError[0m                                 Traceback (most recent call last)',
-            "[1;32merror.ipynb Cell 1[0m in [0;36m<cell line: 2>[0;34m()[0m\n[1;32m      <a href='vscode-notebook-cell:error.ipynb#W0sZmlsZQ%3D%3D?line=0'>1</a>[0m [39m1[39m\n[0;32m----> <a href='vscode-notebook-cell:error.ipynb#W0sZmlsZQ%3D%3D?line=1'>2</a>[0m x\n",
+            "[1;32merror.ipynb Cell 1[0m line [0;36m<cell line: 2>[0;34m()[0m\n[1;32m      <a href='vscode-notebook-cell:error.ipynb#W0sZmlsZQ%3D%3D?line=0'>1</a>[0m [39m1[39m\n[0;32m----> <a href='vscode-notebook-cell:error.ipynb#W0sZmlsZQ%3D%3D?line=1'>2</a>[0m x\n",
```
joyceerhl, line 49: Should we support both the latest ipykernel format as well as the previous one?

amunger: this is the result to display

src/notebooks/outputs/tracebackFormatter.ts
amunger: that's probably better. I don't know if any version had more than one space, but that might be why there was a `+` there

src/notebooks/outputs/tracebackFormatter.unit.test.ts
joyceerhl: To clarify, does our regex now only match on the results from newer ipykernel versions and not results from older versions as a result of this PR?

amunger: no, it should still match previous versions as well.

joyceerhl (approved):

PR #14205, opened 2023-08-28
Add trailing slashes

DonJayamanne:


src/standalone/userJupyterServer/jupyterPasswordConnect.ts
```diff
@@ -305,7 +305,7 @@ export class JupyterPasswordConnect {
         postParams.append('_xsrf', xsrfCookie);
         postParams.append('password', password);
 
-        const response = await this.makeRequest(new URL('login?', url).toString(), {
+        const response = await this.makeRequest(new URL('login?', addTrailingSlash(url)).toString(), {
```
Tyriar, line 308: This will make the URL `login?/` which seems like it's not what you're after

Tyriar (changes requested):

src/standalone/userJupyterServer/jupyterPasswordConnect.ts
DonJayamanne: Thats not true, this will make it `${url}/login?`
here's the ctor for URL
```js
declare var URL: {
    prototype: URL;
    new(url: string | URL, base?: string | URL): URL;
```

Tyriar (approved):

PR #14233, opened 2023-09-01
Updates to the Juptyer Server API

DonJayamanne:
For #13894

### DIFF
--- a/src/api.proposed.d.ts
+++ b/src/api.proposed.d.ts
@@ -5,7 +5,7 @@
 // To use these types please reach out to the Jupyter Extension team (file an issue on the Jupyter Extension GitHub repo).
 // Or please wait for these to be finalized and released.
 
-import { CancellationToken } from 'vscode';
+import { CancellationToken, ProviderResult } from 'vscode';
 import { Event, Uri } from 'vscode';
 
 declare module './api' {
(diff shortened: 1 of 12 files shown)


src/api.proposed.d.ts
```diff
@@ -97,28 +97,25 @@ declare module './api' {
         /**
          * Returns the list of servers.
          */
-        provideJupyterServers(token: CancellationToken): Promise<JupyterServer[]>;
+        provideJupyterServers(token: CancellationToken): ProviderResult<JupyterServer[]>;
```
rzhao271, line 100: I notice a lot of `doSomething || []` throughout the diffs so I think a lot of these return types could also have `| undefined`

src/kernels/jupyter/connection/jupyterServerProviderRegistry.ts
```diff
@@ -262,9 +274,9 @@ class JupyterUriProviderAdaptor extends Disposables implements IJupyterUriProvid
         if (!this.provider.serverProvider) {
             throw new Error(`No Jupyter Server Provider for ${this.provider.extensionId}#${this.provider.id}`);
         }
-        const servers = await this.provider.serverProvider.provideJupyterServers(token);
+        const servers = await Promise.resolve(this.provider.serverProvider.provideJupyterServers(token));
```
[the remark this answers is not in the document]
DonJayamanne: its possible its a sync return value, it’s a method that’s implemented by 3rd party extensions
 simple trick to avoid checking if return value is a promise or not
We do this in core as well for apis

src/api.proposed.d.ts
DonJayamanne: Yes that’s the signature of ProviderResult
It could be a promise, it could be non undefined or defined

rebornix (approved):

PR #14245, opened 2023-09-02
Update typescript

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/issues/13874

### DIFF
--- a/build/ci/postInstall.js
+++ b/build/ci/postInstall.js
@@ -246,6 +246,77 @@ async function downloadZmqBinaries() {
     await downloadZMQ();
 }
 
+function fixUiFabricCompilationIssues() {
+    const filesAndSourceToFix = [
+        {
+            file: 'node_modules/@uifabric/foundation/lib/createComponent.d.ts',
+            lines: [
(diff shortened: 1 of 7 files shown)


build/ci/postInstall.js
```diff
@@ -246,6 +246,77 @@ async function downloadZmqBinaries() {
     await downloadZMQ();
 }
 
+function fixUiFabricCompilationIssues() {
+    const filesAndSourceToFix = [
+        {
+            file: 'node_modules/@uifabric/foundation/lib/createComponent.d.ts',
```
DonJayamanne, line 252: The `@uifabric` package does not compile in strict mode at all.
The solution (approach) to using `@uifabric` is to turn off `skipLibCheck`,
However we do not want to do that, not best practice and has a lot of drawbacks,
easier to patch the *.d.ts files in this package, which we only use in 3 places for dropdown and textbox.

Hopefully we can remove this entire react package.
With this we can finally upgrate TS, we haven't beeen able to do that for a few months now,

aeschli (approved):

PR #14284, opened 2023-09-07
Remove Error Renderer

DonJayamanne:
@rebornix @amunger 
Based on my knowledge these files are no longer required, removing them to reduce the footprint and speed upbuilds.


rebornix (approved):
Thank you!

PR #14435, opened 2023-10-04
Update alpinex64 Python container to include ipykernel

DonJayamanne:


rzhao271 (approved):
This is fine for personal use, but if it's being used in CI, then we'll need to use an MCR container instead.

Issue #13045, opened 2023-03-09
Review and standardize how we normalize code sent to the interactive window


amunger:
pythonFiles/normalizeSelection.py seems to be a carry-over from the python extension split and is only used in some cases for normalizing code sent to the IW - specifically when executing a selection, but not when executing the contents of a cell.

If we do want to continue using the code here, perhaps we should also use it for cell execution as well. Otherwise, perhaps we can remove the duplication between python and jupyter extensions

amunger:
cc @DonJayamanne

PR #14449, opened 2023-10-05
Simpler normalization for executing selection in IW

amunger:
https://github.com/microsoft/vscode-jupyter/issues/13045

This will also let us enable the command in web.


src/platform/terminals/codeExecution/codeExecutionHelper.unit.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    mockTextEditor.selection = selection;
+    return mockTextEditor;
+}
+
+const inputText = `print(1)
+
+if (true):
+    print(2)
+    print(3)
+`;
```
rebornix, line 26: We can have tests for multi-line string, and also mix use of spaces and tabs.

amunger: this does not support mixed tabs and spaces, neither does the current implementation

rebornix (approved):

PR #14599, opened 2023-10-30
esbuild for main bundle

DonJayamanne:


src/test/standardTest.node.ts
```diff
@@ -59,6 +59,14 @@ async function createTempDir() {
     });
 }
 
+// function updatePackageJson() {
```
amunger, line 62: is this needed for something?

amunger (approved):

src/test/standardTest.node.ts
DonJayamanne: Damn, will remove that

PR #14647, opened 2023-11-01
remove deprecated linting settings and use flake8, fix lint errors

amunger:
python.linting settings are deprecated and we're getting a warning notification about it.

Add flake8 as our recommended linting extension and fix the associated linting errors.


amunger:
looks like black is used as part of our CI process, I should probably see if I can use that extension instead

amunger:
There are a bunch of linting fixes for flake8, unused variables, bare exceptions, imports at the top. But these aren't actually necessary if we're going to just use the much less strict rules from black.

DonJayamanne (approved):

PR #14679, opened 2023-11-07
Proposed Kernel API

DonJayamanne:


src/kernels/api/api.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+    get onDidChangeStatus(): Event<
+        'unknown' | 'starting' | 'idle' | 'busy' | 'terminating' | 'restarting' | 'autorestarting' | 'dead'
+    > {
+        return this.kernel.onStatusChanged;
+    }
+
+    constructor(private readonly kernel: IKernel) {}
+
+    executeCode(code: string, token: CancellationToken): ExecutionResult {
```
amunger, line 27: Do we need to whitelist access to this?

amunger (approved):

src/kernels/api/api.ts
DonJayamanne: That will come later, for now this API is not exposed  (i.e. code is not implemented to return this API from the Jupyter extension)

PR #14731, opened 2023-11-16
attach a basic variable provider

amunger:
#14821

- initial implementation for a variable provider
- only makes one python call per request (all variables at that level)
  - No `inspect` requests to the kernel as those were significantly slower than just calling python code
- values are very simple at this point
  - raw types are displayed as is
  - collections are displayed as `[...]`, objects as `{...}`
- The data pipe is separate from the jupyter variable viewer since the types didn't mesh, and this way they can be removed cleanly when we get around to fully replacing the current view.
- No caching yet, but can be implemented for the execution count
- Also missing logic around paging/limiting number of children returned

### DIFF
--- a/package.json
+++ b/package.json
@@ -2193,7 +2193,8 @@
         "notebookKernelSource",
         "interactiveWindow",
         "quickPickItemTooltip",
-        "notebookExecution"
+        "notebookExecution",
+        "notebookVariableProvider"
     ],
     "scripts": {
         "package": "gulp clean && gulp prePublishBundle && vsce package -o ms-toolsai-jupyter-insiders.vsix",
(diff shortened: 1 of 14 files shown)


amunger:
Still need to fix a DI error in the tests

rebornix (approved):

amunger:
* initial implementation for a variable provider
* only makes one python call per request (all variables at that level)
* No inspect requests to the kernel as those were significantly slower than just calling python code
* values are very simple at this point - raw types are displayed as is, collections are displayed as [...], objects as {...}
* The data pipe is separate from the jupyter variable viewer since the types didn't mesh, and this way they can be removed cleanly when we get around to fully replacing the current view.

PR #14738, opened 2023-11-19
Remove dead code and update settings.json

DonJayamanne:


.vscode/settings.json
```diff
@@ -38,33 +38,34 @@
...
     "mochaExplorer.files": "./out/test/**/*.unit.test.js",
     "mochaExplorer.configFile": "./build/.mocha.unittests.js.json",
     "mochaExplorer.ui": "tdd",
-    "mochaExplorer.nodeArgv": [
-        "--enable-source-maps"
-    ]
-}
\ No newline at end of file
+    "mochaExplorer.nodeArgv": ["--enable-source-maps"],
+    "workbench.colorCustomizations": {
```
rzhao271, line 65: Should these be user settings instead?

rzhao271 (changes requested):
Requesting changes because the workbench.colorCustomizations change looks off to me.

.vscode/settings.json
DonJayamanne: Nope, these can be here.

rzhao271 (approved):

PR #14784, opened 2023-11-27
Move Kernel API into API folder

DonJayamanne:


DonJayamanne:
Nope, its in the right place

PR #14926, opened 2023-12-18
Downgrade test-web to fix security issue

DonJayamanne:
Fixes https://github.com/microsoft/vscode-jupyter/security/dependabot/89

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -144,7 +144,7 @@
                 "@typescript-eslint/parser": "^6.9.0",
                 "@vscode/dts": "^0.4.0",
                 "@vscode/test-electron": "^2.3.8",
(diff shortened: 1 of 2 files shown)


lramos15 (changes requested):
This doesn't seem like a great solution to lower a package version that much. Especially when part of SDL is to maintain current packages. 

We own the test package and there is already a dependabot PR open for it to bump Koa to 5.0.0. Let's work with @aeschli to get it in

https://github.com/microsoft/vscode-test-web

DonJayamanne:
> This doesn't seem like a great solution to lower a package version that much. Especially when part of SDL is to maintain current packages.


Until that is fixed, I think we should downgrade, else that leaves repos like this vulnerable.

PR #14935, opened 2023-12-20
In Kernel API exec return output metadata along with output items

### LINKED ISSUE #14922: Unable to distinguish stack frames using proposed kernels API
In the proposed kernel `executeCode` API, tracebacks in error results are concatenated together into one string (despite the underlying messages from the kernel having them split up by stack frame). This makes it difficult / generally impossible to distinguish each stack frame when using the new API.

For example, executing the code:
```python
def div(a, b):
  return a / b
div(1, 0)
```

The Jupyter `error` message directly from the kernel gives:
```json
{
  "ename": "ZeroDivisionError",
  "evalue": "division by zero",
  "traceback": [
    "---------------------------------------------------------------------------",
    "ZeroDivisionError                         Traceback (most recent call last)",
    "Cell In[1], line 3\n      1 def div(a, b):\n      2   return a / b\n----> 3 div(1, 0)\n",
    "Cell In[1], line 2, in div(a, b)\n      1 def div(a, b):\n----> 2   return a / b\n",
    "ZeroDivisionError: division by zero"
  ]
}
```

Notice that each stack frame in the traceback is contained in its own string in the array. Also note that the stack frames themselves may contain multiple lines.

But executing the same code with the new API and parsing the data from the `application/vnd.code.notebook.error` output gives:
```json
{
  "name": "ZeroDivisionError",
  "message": "division by zero",
  "stack": "---------------------------------------------------------------------------\nZeroDivisionError                         Traceback (most recent call last)\nCell In[2], line 3\n      1 def div(a, b):\n      2   return a / b\n----> 3 div(1, 0)\n\nCell In[2], line 2, in div(a, b)\n      1 def div(a, b):\n----> 2   return a / b\n\nZeroDivisionError: division by zero"
}
```

Here the stack frames have all been joined together, and it is generally infeasible to split them back out. We need to be able to distinguish the stack frames so that we can exclude some frames before displaying to the user.

DonJayamanne:
Fixes #14922

### DIFF
--- a/src/api.proposed.kernelApi.d.ts
+++ b/src/api.proposed.kernelApi.d.ts
@@ -4,6 +4,16 @@
 import type { CancellationToken, Uri } from 'vscode';
 
 declare module './api' {
+    interface Output {
+        /**
+         * The output items of this output.
+         */
+        items: OutputItem[];
+        /**
+         * Arbitrary metadata for this cell output. Can be anything but must be JSON-stringifyable.
+         */
+        metadata?: { [key: string]: any };
+    }
     interface OutputItem {
         /**
          * The mime type of the output.
@@ -35,9 +45,9 @@ declare module './api' {
          *
          * @param code Code to be executed.
          * @param token Triggers the cancellation of the execution.
-         * @returns Async iterable of output items, that completes when the execution is complete.
+         * @returns Async iterable of outputs, that completes when the execution is complete.
          */
-        executeCode(code: string, token: CancellationToken): AsyncIterable<OutputItem[]>;
+        executeCode(code: string, token: CancellationToken): AsyncIterable<Output>;
     }
     export interface Kernels {
         /**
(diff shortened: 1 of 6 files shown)


src/api.proposed.kernelApi.d.ts
```diff
@@ -35,9 +45,9 @@ declare module './api' {
          *
          * @param code Code to be executed.
          * @param token Triggers the cancellation of the execution.
-         * @returns Async iterable of output items, that completes when the execution is complete.
+         * @returns Async iterable of outputs, that completes when the execution is complete.
          */
-        executeCode(code: string, token: CancellationToken): AsyncIterable<OutputItem[]>;
+        executeCode(code: string, token: CancellationToken): AsyncIterable<Output>;
```
DonJayamanne, line 50: @rebornix 
Change in the return value, I'm returning the 'NotebookCellOutput' instead of the 'NotebookCellOutputItems[]'

This is because we need to also return the metadata from outputs.
Today we're only returning raw outputs.
This is the right thing to do and also fixes the issue where DW cannot access the original traceback from the error.

bpasero (approved):

PR #15116, opened 2024-02-02
Bundle jsoncparser with esbuild

DonJayamanne:


build/esbuild/build.ts
```diff
@@ -43,10 +43,6 @@ const commonExternals = [
     'commonjs',
     'node:crypto',
     'vscode-jsonrpc', // Used by a few modules, might as well pull this out, instead of duplicating it in separate bundles.
-    // jsonc-parser doesn't get bundled well with esbuild without any changes.
-    // Its possible the fix is very simple.
```
DonJayamanne, line 47: Fix was very simple indeed.
Just needed to take a break and get back to this.

DonJayamanne: No more webpack for production builds, only used for creation of web bundle for tests... will look into that some time later

connor4312 (approved):

Issue #15121, opened 2024-02-03
Data Viewer Not Working in Debug Mode


MrLuolk:
### Applies To

- [ ] Notebooks (.ipynb files)
- [X] Interactive Window and\/or Cell Scripts (.py files with \#%% markers)

### What happened?

There is no longer a "View Value in Data Viewer" when I am debugging a .py file. 
![image](https://github.com/microsoft/vscode-jupyter/assets/99521412/f53494b7-ef02-4cd5-ae5d-cb7a8f18526c)

### VS Code Version

1.86.0

### Jupyter Extension Version

2024.1.0

### Coding Language and Runtime Version

Python 3.12.1

### Language Extension Version (if applicable)

Python Extension 2024.0.0

### Running Jupyter locally or remotely?

Local

nourelsheikh:
I've faced the same issue, I've done a clean install of VS Code, tried previous versions of Python, Python debugger and Jupyter; however, the problem still persists. This has only become an issue last week (Friday).

NicolasBoissin:
Same issue here.

NeuronautLtd:
Same here. I've twice uninstalled and reinstalled python, vs code, the jupyter extension, and still no change...

jake-nybo-wsu:
same issue here. can't get the data viewer to show up as an option. i've tried uninstall/reinstall, different configurations, etc.

jakenybo:
same issue here. can't get the data viewer to show up as an option. i've tried uninstall/reinstall, different configurations, etc. no data viewer here either

starball5:
Potentially related: #15077, #15128

Potentially related on Stack Overflow:

- [View value in Data Viewer stopped showing - view Dataframes](https://stackoverflow.com/q/77931440/11107541)
- [Unable to See "View Value in Data Viewer" Option in VS Code Debugger](https://stackoverflow.com/q/77946460/11107541)
- ["View in Data Viewer" in VS Code not available (without jupyter)](https://stackoverflow.com/q/77942465/11107541)

askar-fabricnano:
This is not related to this specific version, but for those who just need this to work right now, I managed to get it to work by rolling back versions.
- Uninstall Python debugger Extension
- Roll back python extension (I rolled back to v2023.18.0)
- Roll back jupyter (I did to v2023.9.1102792234)
I dont think it really matters specifically which version you roll back to, but I just rolled back 4 months just to be safe.

amunger:
[...]

PR #15141, opened 2024-02-06
register watcher for new debugger type

amunger:
Fixes https://github.com/microsoft/vscode-jupyter/issues/15121

the newly split out python debugger extension uses the debug type `debugpy`, rather than `python` as it was previously


amunger:
Went through the relevant code for this, and it's all just for tracking the state of the debug session and ensuring that we can provide an up-to-date view of the dataframe. So I think this is safe for a recovery.

Need to disable those widget tests tho

Yoyokrazy (approved):

PR #15273, opened 2024-02-29
Update pre-release helper to fix boolean checks

kycutler:
To enable some experiments only in pre-release versions, the truthiness of `isPreReleaseVersion()` is tested. But the function currently returns a string type, so the value is always truthy.

This PR updates the helper to return a more intuitive boolean type, and places where it is needs to be a string (i.e. telemetry) to determine the string based on the boolean value.

-   [ ] ~Has sufficient logging.~
-   [ ] ~Has telemetry for feature-requests.~
-   [ ] ~Unit tests & system/integration tests are added/updated.~
-   [ ] ~[Test plan](https://github.com/Microsoft/vscode-jupyter/blob/main/.github/test_plan.md) is updated as appropriate.~
-   [ ] ~[`package-lock.json`](https://github.com/Microsoft/vscode-jupyter/blob/main/package-lock.json) has been regenerated by running `npm install` (if dependencies have changed).~

### DIFF
--- a/src/kernels/serviceRegistry.node.ts
+++ b/src/kernels/serviceRegistry.node.ts
@@ -132,7 +132,7 @@ export function registerTypes(serviceManager: IServiceManager, isDevMode: boolea
 
     // Subdirectories
     registerJupyterTypes(serviceManager, isDevMode);
-    setSharedProperty('isInsiderExtension', isPreReleaseVersion());
+    setSharedProperty('isInsiderExtension', isPreReleaseVersion() ? 'true' : 'false');
 
     const isPythonExtensionInstalled = serviceManager.get<IPythonExtensionChecker>(IPythonExtensionChecker);
     setSharedProperty(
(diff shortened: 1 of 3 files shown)


DonJayamanne (changes requested):
I don't see a difference in behaviour at all
The return value is not any arbitrary string but an enum )'true', 'false')

The findings values is used in telemetry which expected a string
All those does is convert aboolean to string , unless im missing something

DonJayamanne (approved):

Tyriar (approved):

Tyriar:
It looked the same to me as well fwiw

kycutler:
The behavioral difference is not in the changed files here, but in the other places that `isPreReleaseVersion()` is invoked (specifically in `ExperimentService`). There it is called as if it returns a boolean, but currently it does not, and so the `"false"` return value would not actually be handled as falsy there.

Tyriar (approved):

PR #15284, opened 2024-03-04
Do not resolve env (sysPrefix) if not required

### LINKED ISSUE #15140: Explore not resolving an environment using Python API unless selected
Currently it looks like Jupyter is resolving all environments discovered in the system leading to execution for those binaries, which could be avoided. Just as it is in the Python extension.

Related: https://github.com/microsoft/vscode-jupyter/issues/12719

Practically speaking, `known` should serve as a good enough alternative for display purposes. It might have partial (best-effort) data at first, but is corrected/populated with full information soon anyways, once discovery finishes.

DonJayamanne:
Fixes #15140


TylerLeonhardt (approved):

khanhkit:
Are you sure it already fixed ?
In my PC, i'm using Jupyter v2024.1.2024012502 (pre-release), and this issues still continue spam process get_output_via_markers.py

Issue #15374, opened 2024-03-15
Improve variable summarization


rebornix:
With @DonJayamanne 's background code execution, we could potentially look into how to provide more detailed and fine tuned summary for kernel variables, for example

* Dataframe
  * `df.info`
  * `df.describe`
  * `df.shape`
* List like data structures
  * similar to data frame, we could also provide the shape info like `len(list)`, especially when it's very large
* String
  * provide shape info when it's large

DonJayamanne:
I think all non critical should be run in background, ie everything related to variables view.

PR #15377, opened 2024-03-17
Improve Variable Summarization

rebornix:
Re #15374.


pythonFiles/vscode_datascience_helpers/getVariableInfo/vscodeGetVariableInfo.py
```diff
@@ -9,9 +9,22 @@ def _VSCODE_getVariable(what_to_get, is_debugging, *args):
...
+                in [
+                    _VSCODE_builtins.str,
+                    _VSCODE_builtins.list,
+                    _VSCODE_builtins.tuple,
+                    _VSCODE_builtins.set,
+                    _VSCODE_builtins.dict,
+                ]
+                else ""
+            )
+            return string[: maxStringLength - 1] + "..." + sizeInfo
```
amunger, line 27: the number of children would probably be more useful for collection types, which is already stored in the returned result. 
If I saw `['value1','value2', ... \n\n length: 100`, I would assume that length is the size of the array, so I assume copilot would too

amunger (approved):

PR #15404, opened 2024-03-20
Support summarization for jupyter variables

rebornix:

### DIFF
--- a/src/kernels/variables/JupyterVariablesProvider.ts
+++ b/src/kernels/variables/JupyterVariablesProvider.ts
@@ -10,13 +10,14 @@ import {
     VariablesResult,
     EventEmitter
 } from 'vscode';
-import { IJupyterVariables, IVariableDescription } from './types';
+import { IJupyterVariables, IRichVariableResult, IVariableDescription } from './types';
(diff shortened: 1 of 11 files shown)


src/kernels/variables/JupyterVariablesProvider.ts
```diff
@@ -45,6 +46,15 @@ export class JupyterVariablesProvider implements NotebookVariableProvider {
         }
     }
 
+    private _getVariableResultCacheKey(notebookUri: string, parent: Variable | undefined, start: number) {
```
amunger, line 49: these could stay in the cache class and be overridden by the implementations

amunger (approved):

src/kernels/variables/JupyterVariablesProvider.ts
rebornix: @amunger I was thinking about that originally but then later on found that we need both `parent` and `start` for generating cache key, but for generating summary, only `parent/variable` is needed, thus I went with moving the cache key generation out of the cache class. Hope this makes sense.

amunger (approved):

PR #15494, opened 2024-04-04
Reduce logging unwanted kernel messages

DonJayamanne:


src/kernels/raw/launcher/kernelProcess.node.ts
```diff
@@ -196,66 +205,45 @@ export class KernelProcess extends ObservableDisposable implements IKernelProces
                     );
                 }
             });
+            let sawKernelConnectionFile = false;
             proc.stdout?.on('data', (data: Buffer | string) => {
-                traceVerbose(`Kernel output: ${(data || '').toString()}`);
-                this.sendToOutput((data || '').toString());
+                let output = (data || '').toString();
```
connor4312, line 210: Fyi, it is not safe to convert a stream of buffers directly to a string individually as it does not handle multibyte characters. Instead you can use `proc.stdout?.setEncoding('utf8').on('data', data =>` and `data` is always a string

connor4312 (approved):

src/kernels/raw/launcher/kernelProcess.node.ts
DonJayamanne: i thought it shuold be fine as we've set the coding earlier on.
will do this in a separate PR, thanks

Issue #15485, opened 2024-04-03
use background thread on kernel to gather variable information


amunger:
example: https://insiders.vscode.dev/github.com/microsoft/vscode-jupyter/blob/main/src/standalone/chat/extesnion.ts#L82-L100

amunger:
@DonJayamanne can you help me with a couple questions:

- Our namespaced built-ins aren't available in the scope of the bg thread, does that mean that normal built-ins are ok to use?
- `"../../standalone/api/kernels/backgroundExecution" imported in restricted zone. Only modules from ./src/platform and ./src/telemetry can be imported into ./src/kernels.`
  - can we move this hleper into platform?

DonJayamanne:
Yes to both,
Though I'm not sure why first isn't working. it should
Perhaps its the way the code is structured,
If i recall we used to send multiple requests, instead of sending multiple requests, we might want to send one single request,
Else a function created in one request might not be available to the other request (after all the scope is different as each run in a function scope in different threads).

PR #15495, opened 2024-04-04
Get variable info with background thread

amunger:
Fixes https://github.com/microsoft/vscode-jupyter/issues/15485

https://github.com/microsoft/vscode-jupyter/pull/15495/commits/5b84fba7f541dffcf168b9b8a176790e91ba3c1d might be going against the intention of code dependency restrictions, but I seem to need to have a dependency from kernels -> standalone Api,


src/standalone/api/kernels/backgroundExecution.ts
```diff
@@ -10,6 +10,19 @@ import { raceCancellation } from '../../../platform/common/cancellation';
...
+import { injectable } from 'inversify';
+
+@injectable()
+export class BackgroundThreadService implements IBackgroundThreadService {
+    execCodeInBackgroundThread<T>(
+        kernel: IKernel,
+        codeWithReturnStatement: string[],
+        token: CancellationToken
+    ): Promise<T | undefined> {
+        return execCodeInBackgroundThread(kernel, codeWithReturnStatement, token);
```
DonJayamanne, line 23: Just noticed this PR, and wanted to provide some quick feedback,
We try to avoid injections where possible, as this slows the extension startup.
In this case you can just keep the class, but remove teh `@injectable decorator
& construct the class when you need it, or just pass it from the calling code
Or remove the class entirely and just use the function, I'm assuming you need the class for mocking/testing purposes
Else having this class is not necessary, as it just wraps a function

For instance today we don't use this class in other places, but use the function directly.
Personally prefer to remove this class and keep the function, else there are two ways of doing the same thing 

@rebornix /cc thoughts?

amunger: I need it to work around import zone restrictions - the function is in `standalone` (which doesn't seem right to me, but it depends on the API code which is also in that zone), and the caller is in kernel. Neither of those zones can depend on the other.

rebornix: +1 for avoiding `injectable` if possible. This is a static helper function that can be optimized by bundlers, making it an `injectable` function doesn't seem to give us any benefit at this moment.

Reading the dependencies of `execCodeInBackgroundThread`, it seems we can move it into `kernel` component. `src\kernels\execution\helpers.ts` might be a good place.

rebornix: I missed that `const api = createKernelApiForExtension(JVSC_EXTENSION_ID, kernel);` which depends on `NotebookKernelExecution#onDidReceiveDisplayUpdate`. It needs some code movement to make it happen. Like @DonJayamanne suggested, probably we leave it to him for the refactoring.

lramos15 (approved):

PR #15503, opened 2024-04-05
only handle reading outputs in one place

amunger:
remove some duplicated code.

From what I can tell, we only need the API listener to determine when we have the final output that we are looking for.
Without this, an error can be thrown in both promises, leaving one rejected promise orphaned which adds another error to the log.


amunger:
giving up here since both output channels are necessary (at least with the current unit tests), I'll open another with the fixes I think are still necessary

PR #15506, opened 2024-04-05
dispose listener, don't end early on unrelated error

amunger:
I'm not sure this was actually leaking a listener, but AFAICT, we should still be able to dispose the store in the finally block when we're done listening to display updates. 

also fix a bug where an error output from one bg execution could potentially cause another one to return early with an empty result, introduced in https://github.com/microsoft/vscode-jupyter/pull/15491.

### DIFF
--- a/src/standalone/api/kernels/backgroundExecution.ts
+++ b/src/standalone/api/kernels/backgroundExecution.ts
@@ -94,31 +94,33 @@ del __jupyter_exec_background__
             );
         })
         // We no longer need to track any more outputs from the kernel that are related to this output.
-    ).finally(() => kernel.session && unTrackDisplayDataForExtension(kernel.session, displayId));
+    ).finally(() => {
+        kernel.session && unTrackDisplayDataForExtension(kernel.session, displayId);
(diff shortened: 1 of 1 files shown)


src/standalone/api/kernels/backgroundExecution.ts
```diff
@@ -94,31 +94,33 @@ del __jupyter_exec_background__
...
+    ).finally(() => {
+        kernel.session && unTrackDisplayDataForExtension(kernel.session, displayId);
+        disposables.dispose();
+    });
 
     for await (const output of api.executeCode(codeToSend, token)) {
         if (token.isCancellationRequested) {
             return;
         }
-        const error = output.items.find((item) => item.mime === mimeErrorResult);
```
DonJayamanne, line 103: This isn't right
if the bg thread throws an error, then we sent `mimErrorResult`. (see line `60`).
Thus in the past when bg thread fails, we get the error and stop waiting for output and return.
But now we will never resolve this promise because we never get an output back agaihn.

DonJayamanne (approved):

Issue #15524, opened 2024-04-09
Jupyter is not activating conda environment properly after latest update


guileryu01:
### Applies To

- [X] Notebooks (.ipynb files)
- [X] Interactive Window and\/or Cell Scripts (.py files with \#%% markers)

### What happened?

Starting from version 2024.3.0, Jupyter isn't activating conda environment properly. In particular, it doesn't update PATH environment variable in a such way to ensure those that would be included by executing "conda activate <env>" are included in PATH.

For example, after doing "conda activate python311" in a command line window, I can see following are included in PATH:

```
~\Miniconda3\envs\python39
~\Miniconda3\envs\python39\Library\mingw-w64\bin
~\Miniconda3\envs\python39\Library\usr\bin
~\Miniconda3\envs\python39\Library\bin
~\Miniconda3\envs\python39\Scripts
~\Miniconda3\envs\python39\bin
```

However, in Jupyter Notebook when selecting the same kernel python311, only the first one is included in PATH.
This has caused loading certain python packages to fail in a conda environment, due to not being able to find the dependent DLL files in the search paths, including the popular numpy package via anaconda. 

### VS Code Version

1.88.0

### Jupyter Extension Version

2024.3.0

### Jupyter logs

```shell
15:45:36.611 [info] Disposing kernel .jvsc74a57bd06533efc5e0a53ed70a4e497e066dc8e6955fc1ec7e7f077c49253f608232da42.~\Miniconda3\envs\python39\python.exe.~\Miniconda3\envs\python39\python.exe.-m#ipykernel_launcher for notebook c:\dev\notebooks\example.ipynb due to selection of another kernel or closing of the notebook
15:45:36.611 [info] Dispose Kernel 'c:\dev\notebooks\example.ipynb' associated with 'c:\dev\notebooks\example.ipynb'
15:45:36.629 [info] Process Execution: c:\windows\System32\taskkill.exe /F /T /PID 3996
15:45:36.643 [info] Starting Kernel startUsingPythonInterpreter, .jvsc74a57bd0daf6e195def237259cfab279133bb3c9755699f4dcc242d7c80150c17f82427f.~\Miniconda3\envs\python311\python.exe.~\Miniconda3\envs\python311\python.exe.-m#ipykernel_launcher  (Python Path: ~\Miniconda3\envs\python311\python.exe, Conda, python311 (Python 3.11.8), 3.11.8) for 'c:\dev\notebooks\example.ipynb' (disableUI=true)
[...]

PR #15527, opened 2024-04-09
Fix activated conda env vars from PythonApi (#15526)

DonJayamanne:
release port for https://github.com/microsoft/vscode-jupyter/issues/15524

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "jupyter",
-    "version": "2024.3.0",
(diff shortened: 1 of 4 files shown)


amunger (approved):

grantsimongrant:
These issues are still present, and these issues seem to be causing slow connections to the jupyter kernel in conda, so I have to temporarily use a connection to a remote jupyter server instead of directly connecting to conda.
image

PR #15601, opened 2024-04-24
Request symbols only if document text in cell

DonJayamanne:
For https://github.com/microsoft/vscode/issues/209787

@rebornix /cc

### DIFF
--- a/src/standalone/executionAnalysis/symbols.ts
+++ b/src/standalone/executionAnalysis/symbols.ts
@@ -259,8 +259,10 @@ export class NotebookDocumentSymbolTracker {
                     });
 
                     e.cellChanges.forEach((change) => {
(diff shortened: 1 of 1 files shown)


src/standalone/executionAnalysis/symbols.ts
```diff
@@ -259,8 +259,10 @@ export class NotebookDocumentSymbolTracker {
                     });
 
                     e.cellChanges.forEach((change) => {
-                        this._requestCellSymbols(change.cell, false).then(noop, noop);
-                        this._updateCellStatus(change.cell, CellExecutionStatus.Stale);
+                        if (change.document) {
```
DonJayamanne, line 262: I think this should be done only when there are changes to the text in the cells

PR #15622, opened 2024-04-28
Remove unwanted methods

DonJayamanne:


src/kernels/raw/launcher/kernelLauncher.node.ts
```diff
@@ -312,3 +225,63 @@ export class KernelLauncher implements IKernelLauncher {
...
+        return;
+    }
+    const service = await pythonExecFactory.createActivatedEnvironment({
+        interpreter,
+        resource
+    });
+    const output = await service.exec(
+        [
+            '-c',
+            'import ipykernel; print(ipykernel.__version__); print("5dc3a68c-e34e-4080-9c3e-2a532b2ccb4d"); print(ipykernel.__file__)'
```
Tyriar, line 255: You could put this guid in a constant. Also if you don't generate a new guid is there a risk of someone injecting their own to break the service?

bpasero (approved):

Issue #15680, opened 2024-05-09
Data viewer contribution should handle multiple entrypoints


pwang347:
Based on offline chat with @rebornix @amunger, it would be good for the Jupyter extension to support handling of other launch points such as the Debug panel and anywhere else the existing viewer is shown. Thanks!

amunger:
currently blocked on Data Wrangler's `jupyterVariableViewers` command contribution needing to handle cases with a debug session and no notebook.

amunger:
verification steps:
1. Install Python, DataWrangler and Jupyter pre-release extensions
2. Create a .py file with the content below and set a breakpoint on the last print line
3. Debug the file (not in the Interactive Window)
4. When the breakpoint is hit, navigate to the variables view in the debug panel, right click df, and open the data viewer
5. Ensure Data Wrangler's data viewer is opened

```
import pandas as pd

# create a dictionary with sample data
data = {
    "name": ["Alice", "Bob", "Charlie", "David"],
    "age": [25, 30, 35, 40],
    "gender": ["F", "M", "M", "M"],
}

df = pd.DataFrame(data)

# print the dataframe
print(df)
```

rzhao271:
Does this count as the viewer opening for step 5?

![Data Wrangler opens to an error screen](https://github.com/microsoft/vscode-jupyter/assets/7199958/b4298343-7672-490d-b635-4ad95cf11a2d)

amunger:
it is being delegated correctly, we can just make that a new issue for DW

PR #15681, opened 2024-05-09
delegate requests from the debug variable view as well

amunger:
Fixes https://github.com/microsoft/vscode-jupyter/issues/15680


amunger:
waiting to see if DW can handle cases where no notebook is available (debugging a python file)

rebornix (approved):

PR #15719, opened 2024-05-26
Fix typo ; Update configMigration.ts

furkanakkurt1335:
Fixes #


amunger (changes requested):
the actual config ID has the typo, so only fixing it here will not help.
Also, just fixing the config ID would break the configuration for everyone that has it set - migrating would be necessary, which is a lot of work to fix a typo

PR #15862, opened 2024-07-22
fix: Kubeflow Authentication

### LINKED ISSUE #15659: Kubeflow Authentication breaks with JupyterLab 4
With JupyterLab 4.x the `/tree` path returns a `404` for some reason. 
This breaks iff authenticating with a Kubeflow instance in `getXSRFToken` ([link](https://github.com/microsoft/vscode-jupyter/blob/main/src/standalone/userJupyterServer/jupyterPasswordConnect.ts#L222)), as the result is expected to be a http 200.

This only applies to Kubeflow instances, for which the `sessionCookie` param is used [see line 205](https://github.com/microsoft/vscode-jupyter/blob/main/src/standalone/userJupyterServer/jupyterPasswordConnect.ts#L205).

I've tried using `tokenUrl = new URL('lab', addTrailingSlash(url)).toString();` instead, which works great for JupyterLab 4 Notebooks hosted on Kubeflow.

I'll happily create a PR if this is an appropriate fix. Thanks!

TobiasGoerke:
Fixes #15659

Use a Jupyter URL that does not return 404 with Jupyter 4.x and thus breaks the authentication.

### DIFF
--- a/src/standalone/userJupyterServer/jupyterPasswordConnect.ts
+++ b/src/standalone/userJupyterServer/jupyterPasswordConnect.ts
@@ -202,7 +202,7 @@ export class JupyterPasswordConnect {
         let tokenUrl = new URL('login?', addTrailingSlash(url)).toString();
 
         if (sessionCookie != '') {
-            tokenUrl = new URL('tree', addTrailingSlash(url)).toString();
+            tokenUrl = new URL('lab', addTrailingSlash(url)).toString();
             headers = {
                 Connection: 'keep-alive',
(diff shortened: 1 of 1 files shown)


TobiasGoerke:
@DonJayamanne FYI

DonJayamanne (changes requested):
Older versions of Jupyter will fail with this change.
We need to be backwards compatible.
I.e. what if users use `Jupyter Notebook` instad of lab?

DonJayamanne:
Closing as stale.

PR #15863, opened 2024-07-22
Removal of builtin data viewer entry point.

rebornix:
Users should now be directed to extension recommendations for data frame viewers instead of using the deprecated builtin one.

### DIFF
--- a/src/webviews/extension-side/dataviewer/dataViewerDelegator.ts
+++ b/src/webviews/extension-side/dataviewer/dataViewerDelegator.ts
@@ -43,7 +43,7 @@ export class DataViewerDelegator {
                 }
                 // show quick pick
                 const quickPick = window.createQuickPick<QuickPickItem & { command: string }>();
-                quickPick.title = 'Select DataFrame Viewer';
+                quickPick.title = localize.DataScience.selectExternalDataViewer;
                 quickPick.items = variableViewers.map((d) => {
                     return {
                         label: d.jupyterVariableViewers.title,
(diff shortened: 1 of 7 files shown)


src/webviews/extension-side/dataviewer/dataViewerDelegator.ts
```diff
@@ -14,46 +13,52 @@ import * as localize from '../../../platform/common/utils/localize';
...
             } else {
-                return commands.executeCommand(Commands.ShowJupyterDataViewer, variable);
+                const thirdPartyViewers = variableViewers.filter((d) => d.extension.id !== JVSC_EXTENSION_ID);
+                if (thirdPartyViewers.length === 1) {
+                    const command = thirdPartyViewers[0].jupyterVariableViewers.command;
+                    return commands.executeCommand(command, variable);
+                }
+                // show quick pick
+                const quickPick = window.createQuickPick<QuickPickItem & { command: string }>();
+                quickPick.title = 'Select DataFrame Viewer';
```
Tyriar, line 46: Needs localization, I see it's not new though

Tyriar (approved):

src/webviews/extension-side/dataviewer/dataViewerDelegator.ts
rebornix: Right, it's not new. With that said, it's better to get it localized.

Tyriar (approved):

PR #15916, opened 2024-08-02
Add logging for kernel context keys

Yoyokrazy:
Re: #200896

Added debug logging to diagnose context key issues causing disappearing global toolbar buttons (restart).


DonJayamanne:
@Yoyokrazy If you only need this for debugging purposes, then you can
* Make this change, create a PR as you have done
* Take the VSIX from the build output
* Share the VSIX with users, ask them to test by installing this VSIX manually
* Capture the logs

Else we end up with a lot of logs when users click around/typing etc (as such actions causes these context keys to change and that would produce an excessive amount of logging for all users, which wouldnt be ideal)

DonJayamanne (changes requested):
Only requesting chanages so as to ensure we do not merge this PR,
Feel free to use the VSIX and share that with the users for capturing logs.

Issue #15713, opened 2024-05-24
Workspace-level `jupyter.runStartupCommands`


mthiboust:
Currently, the `jupyter.runStartupCommands` setting can only be set in the user settings file. It would be useful if it could also be defined in the workspace settings file.

Use case: sharing this setting with collaborators through a `.vscode/settings.json` in a git repository.

DonJayamanne:
Thank you for this feature request, however there haven't been sufficient upvotes on this feature request, hence closing this

PR #16038, opened 2024-09-13
Make `jupyter.runStartupCommands` show in the workspace settings

### LINKED ISSUE #15626: Kernel startup commands should be definable at the workspace level not just user.
Honestly, it seems like a mistake that the Jupyter Kernel Startup Commands apply at the user level only. This seems likely way too broad for common use-cases. Couldn't we have one that works at the workspace level?

I ask because my DS chat participant needs to run code whenever a kernel is started. However, the user may not want to use this particular chat participant with EVERY project they ever work on. The extension itself is smart enough to detect whether the needed code is registered when it starts up. This can kick in with new workspaces just as easy.

dbirks:
Fixes #15626 

Just a small change to make `jupyter.runStartupCommands` not only show in the User settings.

Built the extension and confirmed that it now auto-completes in a `.vscode/settings.json` file (which it wasn't doing before):

image

And that it shows in the Workspace tab in the Settings:

image

Thanks for all the effort in maintaining this!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1740,7 +1740,7 @@
                     "type": "array",
                     "default": "",
                     "description": "%jupyter.configuration.jupyter.runStartupCommands.description%",
-                    "scope": "application"
+                    "scope": "machine-overridable"
                 },
                 "jupyter.debugJustMyCode": {
                     "type": "boolean",


DonJayamanne:
Thank you for the pull request, however we cannot accept this at this stage as we have concerns over making this setting a workspace scoped setting.
@rebornix /cc

dbirks:
Ah ok, I'm not sure what those concerns are, but maybe I can make my best case briefly and we'll see if we're talking past each other. :slightly_smiling_face:

I've found that Azure Synapse and Microsoft Fabric give you a `spark` variable pre-initialized with the SparkSession, and without this `jupyter.runStartupCommands`, I've not found a way to duplicate what the live environment is doing. Without this there's that little bit of friction each time of having to add some cells just to your local notebook to get the same behavior.

Having `jupyter.runStartupCommands` in the User settings is amazing, as I can now mimic the behavior locally, but I can't easily share that setting with the other people on my team, as I can't add it to the `.vscode/settings.json` in the repo.

Quite a big part of why extensions are helpful for me is their ability to be shared, both by defining the extensions themselves in the `extensions.json` as you know, and the config for them in the `settings.json`. Without that I'm limited to just making notes in a readme, and can't source control all of my config directly.

Either way, thanks for all your work on this extension. It's a very helpful thing to have in the toolbox.

DonJayamanne:
@dbirks Please can you file an issue with the details you have mentioned above
Along with some samples and a bit more details of the issue you are running into.

> , I've not found a way to duplicate what the live environment is doing. Without this there's that little bit of friction each time of having to add some cells just to your local notebook to get the same behavior.

This is what I do not understand.
Perhaps some more details of the Environment (Python version, kind of kernel, etc) and issues you are running into.
Please do file a new issue and lets continue the conversation there.

iosifache:
Hi @DonJayamanne,

> however we cannot accept this at this stage as we have concerns over making this setting a workspace-scoped setting.

Was there any reassessment of the risk you were thinking about?

I also consider it will be beneficial to allow VS Code workspace creators to automatically run commands inside the notebooks in their workspaces.

P.S.: Besides this PR created by @dbirksm, this request was also made in #15626 by @greazer and in #15713 by @mthiboust.

Thank you in advance 🙏🏿!

Issue #16026, opened 2024-09-09
Data Wrangler (often) doesn't get called when trying to open a DataFrame from other frames on the call stack


danhje:
This bug was originally reported over at the DataWrangler repo ([link](https://github.com/microsoft/vscode-data-wrangler/issues/279)), but they think the issue lies with the Jupyter extension.

## Environment data

-   VS Code version: 1.92.2
-   Data Wrangler Extension version (available under the Extensions sidebar): v1.8.0
-   Jupyter Extension version (available under the Extensions sidebar): v2024.7.0
-   Python Extension version (available under the Extensions sidebar): v2024.14.0
-   OS (Windows | Mac | Linux distro) and version: Windows 11 with Ubuntu on WSL2
-   Pandas version: 2.2.2
-   Python and/or Anaconda version: 3.12
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): poetry venv

## Expected behaviour

DataFrame opens in data view / Data Wrangler.

## Actual behaviour

Nothing happens.

## Steps to reproduce:

Put a breakpoint on the first `return` line of the following script, and run it in VS Code debugger.

```python
import pandas as pd


def inner():
    inner_var = pd.DataFrame()
    return

def outer():
    outer_var = pd.DataFrame()
    inner()

outer()
```

The (empty) DF `inner_var` can be opened in DW. Now find the "CALL STACK" panel and click "outer". The "VARIABLES" panel now previews `outer_var`. But opening it in DW often fails, though not always.

PR #16057, opened 2024-09-17
try delegating even if we cannot get the full info

amunger:
Fixes https://github.com/microsoft/vscode-jupyter/issues/16026

We add the context menu command only if we decide that the variable can support a Data viewer, so it seems pretty safe to try and delegate the Data viewer command out even if we can't get the type info at that moment.

### DIFF
--- a/src/kernels/variables/helpers.ts
+++ b/src/kernels/variables/helpers.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 import { DebugProtocol } from 'vscode-debugprotocol';
+import { IJupyterVariable } from './types';
 
 export const DataViewableTypes: Set<string> = new Set<string>([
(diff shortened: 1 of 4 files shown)


amunger:
@pwang347 - It will now call the DW command in that case, where it didn't before, but DW fails to open the viewer with `Could not retrieve variable <MyEnum.ONE: 1> from the Jupyter extension. Please file an issue on the Data Wrangler GitHub repository.`

info we are passing:
`{"name":"<MyEnum.ONE: 1>","count":0,"shape":"","size":0,"supportsDataExplorer":false,"value":"...","truncated":true,"frameId":9}`

for some reason, the name from the context does not include the parent variable `d` on that variable, which may be an issue with debugpy

connor4312 (approved):

PR #16116, opened 2024-10-10
fix: tag experimental settings with new system

rzhao271:
I'm retagging some Jupyter extension settings as part of the experimental -> onExP rename where:
- experimental refers to a setting not being stable yet
- onExP refers to a setting that can be part of ExP A/B experiments

### DIFF
--- a/package.json
+++ b/package.json
@@ -1708,7 +1708,7 @@
                     "default": false,
                     "description": "%jupyter.configuration.jupyter.interactiveWindowNotebookRepl.description%",
                     "tags": [
-                        "experimental"
+                        "experimental", "onExP"
                     ]
                 },
                 "jupyter.interactiveWindow.textEditor.cellFolding": {
@@ -1802,7 +1802,10 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Experimental feature to enable execution analysis in notebooks",
-                    "scope": "application"
+                    "scope": "application",
+                    "tags": [
+                        "experimental"
+                    ]
                 },
                 "jupyter.enableKernelCompletions": {
                     "type": "boolean",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1802,15 +1802,18 @@
...
+                    ]
                 },
                 "jupyter.enableKernelCompletions": {
                     "type": "boolean",
                     "default": true,
                     "markdownDescription": "%jupyter.configuration.jupyter.enableKernelCompletions.markdownDescription%",
                     "scope": "application",
                     "tags": [
-                        "experimental"
+                        "onExP"
```
rzhao271, line 1816: This setting looks like it already rolled out to 100% of users. Could we stop the experiment and consider this setting a stable setting?

package.json
```diff
@@ -1802,16 +1803,16 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Experimental feature to enable execution analysis in notebooks",
-                    "scope": "application"
+                    "scope": "application",
```
DonJayamanne, line 1806: Why are the settings `executionAnalysis.enabled` and `interactive.notebookRepl`  using different tags.
One has onExp and other doesn't have it

package.json
DonJayamanne: Yes, thank you.

package.json
rzhao271: onExP means the setting could be assigned to an A/B experiment.
If that's not the intended meaning, I could remove that tag from the setting.

amunger: is there an A/B experiment for (or planned for) `executionAnalysis.enabled`? that's what that tag is for.
Eventually we'll get an experiment for the notebook REPL.

DonJayamanne (approved):

amunger (approved):

PR #16214, opened 2024-11-08
Add public API event for kernel post-initialization

pwang347:


src/standalone/api/kernels/index.ts
```diff
@@ -29,18 +31,6 @@ export function getKernelsApi(extensionId: string): Kernels {
                 });
                 return;
             }
-            const execution = kernelProvider.getKernelExecution(kernel);
-            if (!isRemoteConnection(kernel.kernelConnectionMetadata) && execution.executionCount === 0) {
-                // For local kernels, execution count must be greater than 0,
-                // As we pre-warms kernels (i.e. we start kernels even though the user may not have executed any code).
-                // The only way to determine whether users executed code is to look at the execution count
```
DonJayamanne, line 36: We need some check to ensure we do not trigger these events for kernels that are automatically started
Events should be triggered only for kernels that were explicitly started by user

src/standalone/api/kernels/index.ts
```diff
@@ -29,18 +31,6 @@ export function getKernelsApi(extensionId: string): Kernels {
...
-                sendTelemetryEvent(Telemetry.NewJupyterKernelsApiUsage, undefined, {
-                    extensionId,
-                    pemUsed: 'getKernel',
-                    accessAllowed
-                });
-                return;
```
DonJayamanne, line 42: Not sure why this telemetry point was removed

pwang347: this one was removed since it was in the if block I removed but maybe we'll still need it based on comment above

DonJayamanne: Please can you bring in the logic to check if `public async start(options?: IDisplayOptions):` was invoked with `disableUI = false`
E.g. keep a property on kernel to indicate how it was started, then check that variable here..

E.g. create a property 
```typescript
public get userStartedKernel() {
    return this.startupUI.disableUI;
}
```

This way, we still have a way of not returning kernels until and unless it was explicitly started by users (which is what we hoped for in the earlier condition).

src/standalone/api/kernels/api.vscode.common.test.ts
```diff
@@ -185,6 +195,104 @@ suiteMandatory('Kernel API Tests @typescript', function () {
             'Traceback does not contain original error'
         );
     });
+    testMandatory('Kernel start event is not triggered by silent executions', async function () {
```
DonJayamanne, line 198: Can we have another test for `kernel Start event is triggered when user executes code` or the like.
Or is that not possible?

src/kernels/kernel.ts
```diff
@@ -246,7 +251,16 @@ abstract class BaseKernel implements IBaseKernel {
...
             this.startCancellation = new CancellationTokenSource();
         }
-        return this.startJupyterSession(options);
+        return this.startJupyterSession(options).then((result) => {
+            // If we started and the UI is no longer disabled (ie., a user executed a cell)
+            // then we can signal that the kernel was created and can be used by third-party extensions.
+            // We also only want to fire off a single event here.
+            if (!options?.disableUI && !this._postInitializedOnStart) {
+                this._onPostInitialized.fire();
+                this._postInitializedOnStart = true;
```
DonJayamanne, line 260: Awesome, nice n simple.

src/standalone/api/kernels/index.ts
pwang347: Let me know if the latest iteration based on feedback below addresses this. Thanks!

src/standalone/api/kernels/index.ts
pwang347: added!

src/standalone/api/kernels/api.vscode.common.test.ts
pwang347: should now be covered by "Kernel start event is triggered when user executes code and the event execution runs first"

DonJayamanne (approved):

joyceerhl (approved):

PR #16252, opened 2024-11-26
chore: work around Missing Symbols error

rzhao271:
This PR registers zeromq-prebuilt as a dependent pipeline so that the engineering pipeline template could download an exact artifact from that pipeline for APIScan to use. I tried using symbol publishing but couldn't get it to resolve our internal Missing Symbols error.


Tyriar (approved):

rzhao271:
Verification build: https://dev.azure.com/monacotools/Monaco/_build/results?buildId=307193&view=results

Issue #15987, opened 2024-09-03
Expose a unified way to retrieve Python Env Associated with Jupyter Notebooks


DonJayamanne:
Currently there's a private API that Pylance gets to use in order to determine the Python Env associated with Jupyter notebooks.
Unfortunately there are a few problems with this API:
* Its built specifically for Pylance, other extensions cannot get this same information
* Upon selecting a new Kernel, the Pylance extension won't always know that a new Python Env has been selected for a Jupyter Notebook
* The API is inconsistent, currently there's an API in Python extension to get the Python Env associated with a Uri, but Jupyter Notebooks is not part of this (when Jupyter Notebooks are also just Uris).

**Solution:**
* Keep the Python extension + Jupyter Extension API internal (as today)
* Update the Python extension as follows:
```typescript
export type ActiveEnvironmentPathChangeEvent = EnvironmentPath & {
    /**
     * Workspace folder the environment changed for.
     * Or Resource the environment changed for.
     */
    readonly resource: WorkspaceFolder | Uri | undefined;
};
```

Internally, the call to `getActiveEnvironmentPath(resource?: Resource): EnvironmentPath;` would now handle Uris a little differently. If the `Uri` is a notebook Uri, then get the interpreter from Jupyter extension (using existing code).
The check if the Uri is a notebook can optionally be passed into Jupyter as well, i.e. keep it super simple.

DonJayamanne:
This API is now available in Python extenison
Done here https://github.com/microsoft/vscode-python/pull/24771

PR #16332, opened 2024-12-24
Improved api to get Python env associated with Jupyter Notebook

DonJayamanne:
For https://github.com/microsoft/vscode-jupyter/issues/15987

### DIFF
--- a/src/api.proposed.notebookEnvironment.ts
+++ b/src/api.proposed.notebookEnvironment.ts
@@ -1,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import type { Event, Uri } from 'vscode';
+
+declare module './api' {
+    /**
+     * These types are not required for any other extension, except for the Python extension.
+     * Hence the reason to keep this separate. This way we can keep the API stable for other extensions (which would be the majority case).
+     */
+    export interface Jupyter {
+        /**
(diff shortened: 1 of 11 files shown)


src/api.proposed.notebookEnvironment.ts
```diff
@@ -0,0 +1,37 @@
...
+         */
+        onDidChangePythonEnvironment: Event<Uri>;
+        /**
+         * Returns the EnvironmentPath to the Python environment associated with a Jupyter Notebook or Interactive Window.
+         * If the Uri is not associated with a Jupyter Notebook or Interactive Window, then this method returns undefined.
+         * @param uri
+         */
+        getPythonEnvironment(uri: Uri):
+            | undefined
+            | {
```
karthiknadig, line 24: I would really just return the python executable path. Since we intend to switch to a different style of API. we can either wait here for the same of this API to be consistent with python environments extension, or return path which can be used to get the details from python environment extension.

karthiknadig (changes requested):

src/api.proposed.notebookEnvironment.ts
DonJayamanne: > Since we intend to switch to a different style of API. we can either wait here for the same of this API to be consistent with python environments extension

Lets wait, i'd prefer to return something thats consistent with what we have today.
Else its a mish mash of different things.

karthiknadig (approved):

PR #16377, opened 2025-01-14
change the scope of excludeUserSitePackages to window

alexfanqi:
Hi,

It sometimes would be helpful to choose to include user site packages depending on the env, so also depending on workspace.
For example, for system python interpreter, I would like to include user site packages.
Whereas for conda environment, I prefer not to because user site packages take priority over conda packages, breaking the standalone conda env and mix with package dependencies that are installed for system python interpreter.

This pr just allows user to set excludeUserSitePackages as workspace local variable.


alexfanqi:
Could you perhaps help with this? @DonJayamanne Sorry if it is not polite to ping person directly.

DonJayamanne (approved):

DonJayamanne:
@alexfanqi apologies for missing this. & thank you for this PR

chrmarti (approved):

PR #16592, opened 2025-04-29
guidance on appropriate tool usage, ask for confirmation for installing

amunger:
fix https://github.com/microsoft/vscode-python-environments/issues/344
fix https://github.com/microsoft/vscode-copilot/issues/16727

### DIFF
--- a/package.json
+++ b/package.json
@@ -2053,9 +2053,9 @@
         ],
         "languageModelTools": [
             {
-                "name": "notebook_list_packages_tool",
+                "name": "notebook_list_packages",
                 "displayName": "List Notebook Kernel Packages",
-                "modelDescription": "List the installed packages that are currently available in the selected kernel for a notebook editor.",
+                "modelDescription": "List the installed packages that are currently available in the selected kernel for a notebook editor. This tool should be used when working with a notebook but only if already working in a notebook file.",
                 "toolReferenceName": "listNotebookPackages",
                 "tags": [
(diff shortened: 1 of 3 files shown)


connor4312 (approved):

package.json
```diff
@@ -2053,9 +2053,9 @@
         ],
         "languageModelTools": [
             {
-                "name": "notebook_list_packages_tool",
+                "name": "notebook_list_packages",
                 "displayName": "List Notebook Kernel Packages",
-                "modelDescription": "List the installed packages that are currently available in the selected kernel for a notebook editor.",
+                "modelDescription": "List the installed packages that are currently available in the selected kernel for a notebook editor. This tool should be used when working with a notebook but only if already working in a notebook file.",
```
DonJayamanne, line 2058: @amunger you might want to update the message to indicate this ONLY applies to `Jupyter Notebooks`, we don't want this being used for other notebooks. Just like this got used for non-notebook files like Python files.

amunger: > but only if already working in a notebook file

DonJayamanne: but that doesn't mean that  applies to Jupyter notebooks, 
I.e. i could be working on a kusto notebook file or the like.

DonJayamanne: Similarly, if we have packages missing in nodejs notebooks, then I don't think the model will know not to use this tool

PR #16594, opened 2025-04-29
Fix PDF export file path handling

### LINKED ISSUE #16590: Cannot export PDF just with vscode extension (other methods work)
## Environment Data

-   ~~VS Code version: 1.99.3.0~~(not relevant)
-   Jupyter Extension version: 2025.3.0 **(however, even 2020.2.0 applies, so it doesn't matter)**
-   ~~Python Extension version: 2025.4.0~~(not relevant)
-   OS: Windows 11 24H2
-   ~~Python and/or Anaconda version: 3.8, 3.11 and 3.12~~(not relevant)
-   ~~Type of virtual environment used: conda~~(not relevant)
-   ~~Jupyter server running: Local and Remote~~(not relevant)

## Expected Behavior

Exporting a Jupyter notebook to PDF from the VS Code Jupyter extension bar, and get a pdf.

## Actual Behavior

Failed to export.

## Steps to Reproduce

1.  Open any `.ipynb` file in VS Code.
2.  Right-click and select "Export as PDF".
3.  Observe the export failure, despite a valid PDF existing in the temp directory (for example "C:\Users\...\AppData\Local\Temp\tmp-53764aZammKviOmV0.pdf.pdf").

## Additional Information

1. available options

| Source            | To   | Via               | Success? |
| ----------------- | ---- | ----------------- | -------- |
| Web Server GUI    | PDF  | Download          | ✅        |
| Command Line      | TEX  | Command Line      | ✅        |
| Command Line      | PDF  | Command Line      | ✅        |
| VS Code Extension | HTML | VS Code Extension | ✅        |
| VS Code Extension | PY   | VS Code Extension | ✅        |
| VS Code Extension | PDF  | VS Code Extension | ❌        |

2. This issue is distinct from #13912, #10910, and #12742, which suggest `pandoc` or `nbconvert` problems. These are not the root cause.
3. **root cause**: pdf is actually generated, however, the extension failed to get his name. For example, the true ath is  `...\AppData\Local\Temp\tmp-54084AG2zf2BvEQSm.pdf.pdf`, but the extension accesses `...\AppData\Local\Temp\tmp-54084AG2zf2BvEQSm.pdf` and thus got a failure.
Okay, here's the updated text in English:
4.  The code snippet from `src/notebooks/export/exportBase.node.ts`, `if((await this.fs.stat(Uri.file(tempTarget.filePath))).size > 1) await this.fs.copy(Uri.file(tempTarget.filePath), target);` is the source of the error. `tempTarget.filePath` contains the correct temporary filename (e.g., `...\AppData\Local\Temp\tmp-54084AG2zf2BvEQSm.pdf`), but the extension is looking for a file with an extra `.pdf` extension. A temporary workaround is to modify the code to `if((await this.fs.stat(Uri.file(pdfFilePath + '.pdf'))).size > 1) await this.fs.copy(Uri.file(pdfFilePath + '.pdf'), target);`.

CNDY1390:
Fixes #16590

Fixes the long-standing bug where exporting a Jupyter notebook to PDF via the VS Code Jupyter extension fails, even though the PDF is correctly generated in the temp directory. The extension was looking for the wrong filename (missing or duplicating the `.pdf` extension), resulting in a file-not-found error and failed export.

## Background and Motivation

- As detailed in [issue #16590](https://github.com/microsoft/vscode-jupyter/issues/16590), exporting to PDF from the VS Code extension has been broken for years (at least since 2020).
- The bug is not present in the web interface or command-line tools (`nbconvert`), only in the VS Code extension.
- Many users believe Jupyter's export is convenient, but in practice, exporting to PDF from VS Code is unreliable and cumbersome. But if this commit is merged,  exporting to PDF from VS Code would be much easier, especially for begginners.

## Root Cause

- The extension generates a PDF file in the temp directory (e.g., `tmp-xxxx.pdf`), but then tries to access or copy a file with an incorrect name (e.g., `tmp-xxxx.pdf.pdf` or missing `.pdf`).
- This is due to incorrect filename/path handling in the export logic, specifically in `src/notebooks/export/exportBase.node.ts`:
    ```js
    if ((await this.fs.stat(Uri.file(tempTarget.filePath))).size > 1) {
        await this.fs.copy(Uri.file(tempTarget.filePath), target);
    }
    ```

## What has been changed
- The filename and path handling logic is corrected to ensure the extension always looks for the actual generated PDF file.

## Additional context
- Thanks to the community for detailed analysis and temporary workarounds.
- My implementation might not be robust for non-Windows environment. Please follow the post if there is any problem. 👍 

---

Closes #16590

### DIFF
--- a/src/notebooks/export/exportBase.node.ts
+++ b/src/notebooks/export/exportBase.node.ts
@@ -108,8 +108,12 @@ export class ExportBase implements IExportBase {
             return;
         }
         try {
-            if ((await this.fs.stat(Uri.file(tempTarget.filePath))).size > 1) {
-                await this.fs.copy(Uri.file(tempTarget.filePath), target);
+            let pdfFilePath = tempTarget.filePath;
+            if (!pdfFilePath.endsWith('.pdf') && (await this.fs.exists(Uri.file(pdfFilePath + '.pdf')))) {
+                pdfFilePath = pdfFilePath + '.pdf';
+            }
+            if ((await this.fs.stat(Uri.file(pdfFilePath))).size > 1) {
+                await this.fs.copy(Uri.file(pdfFilePath), target);
             } else {
                 throw new Error('File size is zero during conversion. Outputting error.');
             }


src/notebooks/export/exportBase.node.ts
```diff
@@ -108,8 +108,12 @@ export class ExportBase implements IExportBase {
             return;
         }
         try {
-            if ((await this.fs.stat(Uri.file(tempTarget.filePath))).size > 1) {
-                await this.fs.copy(Uri.file(tempTarget.filePath), target);
+            let pdfFilePath = tempTarget.filePath;
+            if (!pdfFilePath.endsWith('.pdf') && (await this.fs.exists(Uri.file(pdfFilePath + '.pdf')))) {
```
amunger, line 112: did you check why it creates a file that ends in .pdf.pdf?

CNDY1390: Thanks for the feedback. I’ve dug deeper into the issue and found that `nbconvert` automatically appends a `.pdf` suffix to the `--output` parameter, causing the filename error. Instead of modifying the judgment function, a better approach is to pass a filename without the suffix. I’ll close this PR and submit a new one with the improved fix.

CNDY1390:
Hi there, I've opened a new PR with an improved fix for this issue. Please take a look at the updated solution here: #16651. Thank you!

PR #16622, opened 2025-05-09
chore: start locking down workflows

rzhao271:
Applies some [zizmor](https://docs.zizmor.sh/) audit suggestions to the workflow.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -4,8 +4,7 @@
 
 name: Build and Test
 
-permissions:
-  deployments: write
+permissions: {}
 
 on:
   pull_request:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test.yml
```diff
@@ -837,6 +848,6 @@ jobs:
           DISPLAY: 10
           VSC_JUPYTER_FORCE_LOGGING: 1
           VSC_JUPYTER_CI_TEST_VSC_CHANNEL: 'insiders'
-        uses: GabrielBB/xvfb-action@v1.7
+        uses: GabrielBB/xvfb-action@b706e4e27b14669b486812790492dc50ca16b465 # v1.7
```
amunger, line 851: where do these values come from?

amunger:
unrelated test error

.github/workflows/build-test.yml
amunger: or rather, how did you collect them to be inserted here?

rzhao271: I did it manually by looking up the repositories corresponding to the actions and then finding suitable commit SHAs, e.g. https://github.com/GabrielBB/xvfb-action/releases -> https://github.com/GabrielBB/xvfb-action/commit/b706e4e27b14669b486812790492dc50ca16b465.

zizmor does not trust actions outside of `actions/*` by default and encourages pinning to specific commit SHAs instead of version tags.

amunger (approved):

PR #16644, opened 2025-05-16
Prefer kernel last used by Python Tools

DonJayamanne:


src/platform/api/python-envs/pythonEnvsApi.ts
```diff
@@ -9,17 +9,17 @@ export async function getEnvExtApi(): Promise<PythonEnvironmentApi | undefined>
     if (_extApi) {
         return _extApi;
     }
-    const extension = vscode.extensions.getExtension('ms-python.vscode-python-envs');
```
DonJayamanne, line 12: @amunger made some changes while looking at the API code (as I too need to use this)

bpasero (approved):

PR #16648, opened 2025-05-16
Aamunger/package tools

amunger:
Fixes https://github.com/microsoft/vscode-copilot/issues/17448
Fixes https://github.com/microsoft/vscode-jupyter/issues/16637

PR #16655, opened 2025-05-19
Re #16648. Ensure nb editor is open before kernel activation.

rebornix:
The editor should be open before we start kernels. Opening notebook document itself is not sufficient.

### DIFF
--- a/src/standalone/chat/installPackageTool.ts
+++ b/src/standalone/chat/installPackageTool.ts
@@ -43,6 +43,10 @@ export class InstallPackagesTool implements vscode.LanguageModelTool<IInstallPac
             }
         }
 
+        if (!vscode.window.visibleNotebookEditors.some((e) => e.notebook.uri.toString() === notebook.uri.toString())) {
+            await vscode.window.showNotebookDocument(notebook);
+        }
+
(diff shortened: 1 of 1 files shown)


rebornix:
FYI @amunger

anthonykim1:
[Build and Test / Tests (web, python, 3.10, ^[^@]+$|@mandatory, ubuntu-latest) (pull_request)](https://github.com/microsoft/vscode-jupyter/actions/runs/15122313411/job/42507294902?pr=16655) is failing

src/standalone/chat/installPackageTool.ts
```diff
@@ -43,6 +43,10 @@ export class InstallPackagesTool implements vscode.LanguageModelTool<IInstallPac
             }
         }
 
+        if (!vscode.window.visibleNotebookEditors.some((e) => e.notebook.uri.toString() === notebook.uri.toString())) {
```
DonJayamanne, line 46: recommend using `isEqual` function (which we've copied from core)

DonJayamanne (dismissed):

src/standalone/chat/installPackageTool.ts
rebornix: good catch.

PR #16695, opened 2025-06-03
make note that kernel state is empty after restart

amunger:
Fixes https://github.com/microsoft/vscode-copilot/issues/18003

Another option is to push it to use the summary tool, but hopefully this can take care of it without needing that extra tool call.

### DIFF
--- a/src/standalone/chat/restartKernelTool.node.ts
+++ b/src/standalone/chat/restartKernelTool.node.ts
@@ -33,7 +33,7 @@ export class RestartKernelTool implements vscode.LanguageModelTool<RestartKernel
         const uri = vscode.Uri.file(options.input.filePath);
         sendLMToolCallTelemetry(RestartKernelTool.toolName, uri);
         await vscode.commands.executeCommand('jupyter.restartkernel', uri);
-        const finalMessageString = `The kernel for the notebook at ${options.input.filePath} has been restarted.`;
+        const finalMessageString = `The kernel for the notebook at ${options.input.filePath} has been restarted and any state from previous cell executions has been cleared.`;
         return new vscode.LanguageModelToolResult([new vscode.LanguageModelTextPart(finalMessageString)]);
     }
 


src/standalone/chat/restartKernelTool.node.ts
```diff
@@ -33,7 +33,7 @@ export class RestartKernelTool implements vscode.LanguageModelTool<RestartKernel
         const uri = vscode.Uri.file(options.input.filePath);
         sendLMToolCallTelemetry(RestartKernelTool.toolName, uri);
         await vscode.commands.executeCommand('jupyter.restartkernel', uri);
-        const finalMessageString = `The kernel for the notebook at ${options.input.filePath} has been restarted.`;
+        const finalMessageString = `The kernel for the notebook at ${options.input.filePath} has been restarted and any state from previous cell executions has been cleared.`;
```
DonJayamanne, line 36: You might want to nudge the model by mentioning what tool to use in order to get the latest state

amunger: yeah I was hoping that it could get the same information with this message without needing an extra tool call

DonJayamanne: We do this in other parts of copilot, i.e. we explicitly call our the tool to be used.

amunger: doesn't seem necessary since it will always just show that the kernel state is empty. I assume there is some value in reducing the amount of tool calls - a little quicker if nothing else

amunger: from https://github.com/microsoft/vscode-copilot/issues/18051 it sounds like you agree that reducing tool calls is beneficial

DonJayamanne: What I'm getting at is, the model will know what tool to call if it needs the latest state. We're not asking the model to call the tool always to get the state, just telling it how to get the state if it needs it.

DonJayamanne (approved):

PR #16758, opened 2025-06-30
remove prompt reference for tools

amunger:
Fixes https://github.com/microsoft/vscode-copilot/issues/16799

### DIFF
--- a/package.json
+++ b/package.json
@@ -2023,7 +2023,6 @@
                     "notebooks"
                 ],
                 "icon": "$(notebook)",
-                "canBeReferencedInPrompt": true,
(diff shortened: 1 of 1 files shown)


DonJayamanne (approved):

amunger:
@rebornix noted that the setting also allows users to include/exclude these tools in custom chat modes, so we'll just leave it alone for now.

DonJayamanne:
I'm glad @rebornix  reviewed this
I mis read this completely, I thought we were adding `prompt references` to all the notebook tools 🫤 
Which is what we'd want , so users can select/unselect any tool they want.

PR #16865, opened 2025-07-30
Implementation for kernel reconnection + tests

### LINKED ISSUE #3998: Long running kernels
**Investigation Running Server & JupyterLab API for extensibility**
**Goals:**
* Long running kernels
* Extensibility for extension authors
This is a by-product of the long running kernels (i.e. you get this for free - almost)
		
		
	
	
**Planned Prototypes**
* Long running kernels
Solve problems related to kernel/session being lost due to :
    * VS Code Shutdown
    * VS Code Restart
    * Computer sleeping
    * SSH connection issues
    * AML Compute will benefit
* Julia Widgets (i might end up doing this first, might be easier)
* IPyWidget outside notebooks
* Variable viewer using the new api
* Data Frame viewer using the new api

**Technical details**
* Server
Background process  
Manages kernels & sessions  
Expose kernel socket connection over this connection (we already have the code/technology for this) - proxy socket (dummy kernel in UI layer, by creating a dummy socket connection)  
Security - how do we secure this web server (will need to be addressed, but i'm leaving that for later)	
* Expose Jupyter extension extensibility over Jupyter Lab API 
	I wont be exposing a connection, instead will just expose the SessionManager, KernelManager & other class instances from extension API

suprjinx:
Fixes #3998


suprjinx:
@microsoft-github-policy-service agree company="G-Research OSS"

DonJayamanne:
@suprjinx thank you for submitting this PR.
Wanted to leave a comment on this before you spend more time on this.
Unfortunately we will not be able to take this PR, as this relies on spinning up Jupyter server locally.
Not everyone will have Jupyter lab/notebook installed. let alone Python. 
The Jupyter extension works with Kernels such as R, Julia and the like even without installing Python/Jupyter.

Enforcing such a limitation wouldn't work for all
Besides you can already spin up a Jupyter application externally and connect to that using Jupyter extension and the connections are never lose. I.e. even after you shutdown VS Code and restart it, it will re-connect to the same server/kernels.

Again, thank you for submitting this PR and attempting to fix this, however I cannot take this due to above mentioned issues.

suprjinx:
@DonJayamanne thanks for the feedback -- what do you think of simpler implementation that just offers an option to start server in the kernel connection chooser eg: 

image

The idea would be a `Launch Persistent Server` command that runs `jupyter server`, and we then connect using the Existing Server pathway. Implemented here: https://github.com/microsoft/vscode-jupyter/pull/16915 

Re: users who don't have Jupyter installed -- perhaps we could disable this option when there are no python environments available.

PR #16885, opened 2025-08-06
Add support for custom CDN URLs

### LINKED ISSUE #16875: Allow to configure custom CDN for widget script sources
Currently the widget script sources for CDN are limited to two hard-coded options. This prevents using widgets with VSCode in scenarios where internet access to these is blocked.

https://github.com/microsoft/vscode-jupyter/blob/db29c7afa34aa68778f7331273e3786940c6eee8/package.json#L1451-L1468

Could vscode-jupyter allow users to configure a custom CDN?

I am happy to open a PR. To retain the current UX I would propose to wrap the existing `enum` in `oneOf`  section which would also contain an option to input a string with a valid URL.

krassowski:
Fixes #16875

The completion support in JSON editor is preserved:

image

Invalid URLs are validated:

image

Custom jsdelivr style CDNs are supported:

image

Things I am not sure about:
- translations: should I reuse identifiers even if technically these are no longer enum descriptions?
- `requiresExactVersion` vs `supportsVersionRanges`
- specification of `requiresExactVersion` and `requiresExtension` vs something like `"jsdelivrStyle": true`


src/notebooks/controllers/ipywidgets/scriptSourceProvider/cdnWidgetScriptSourceProvider.ts
```diff
@@ -38,26 +42,34 @@ function moduleNameToCDNUrl(cdn: string, moduleName: string, moduleVersion: stri
...
+        cdn = {
+            url: cdnBaseUrl,
+            // Js Delivr doesn't support ^ in the version. It needs an exact version
+            requiresExactVersion: true,
+            // Js Delivr also needs the .js file on the end.
+            requiresExtension: true
+        };
+    }
+    if (cdn instanceof Object) {
+        if (cdn.requiresExactVersion && moduleVersion.startsWith('^')) {
```
DonJayamanne, line 55: thanks for this PR, how about we change the format of the package json to just strings, `jsdelivr.com` and `unpkg.com` and for these we look at the value and hard code the requireExactVersion and requiresExtension.
If its something else, then we leave it as some default.

However this might not work for mirror jsdeliv sites. then again, we haven't had any requests for mirrored sites, hence I'm inclined to avoid having a more complex package.json and go with just strings.

What to you think?
Does this work for the site you're using?

krassowski: Thank you for taking a look!

> Does this work for the site you're using?

No, currently the CDN I target only supports `jsdelivr.com` format - but I am checking if this can be changed. Out of big CDNs `cdnjs.com` also requires an exact version and `.js` suffix (and compared to `jsdelivr.com` does not have redirects at all), although to support `cdnjs` the format should also include `/` instead of `@` for version specifier.

Another option would be giving some kind of a format string, like:

```
https://cdnjs.cloudflare.com/ajax/libs/<library>/<version>/<file>
```

This would keep the `package.json` to accept only strings, while allowing for a degree of customization. What do you think?

DonJayamanne:
@krassowski thank you very much, this is great

DonJayamanne (approved):

dbaeumer (approved):

Issue #16977, opened 2025-09-12
Version Jupyter Extension version: 2025.8.0 cannot select kernel


cplager:
### Applies To

- [x] Notebooks (.ipynb files)
- [ ] Interactive Window and\/or Cell Scripts (.py files with \#%% markers)

### What happened?

With version Jupyter Extension version: 2025.8.0 I am unable to select a kernel (Python environments or Existing kernels never appears).

### VS Code Version

1.102

### Jupyter Extension Version

Jupyter Extension version: 2025.8.0

### Jupyter logs

```shell
Visual Studio Code (1.102.0, undefined, desktop)
Jupyter Extension Version: 2023.8.1002501831.
Python Extension Version: 2025.14.0.
Platform: win32 (x64).
Workspace folder ~\LocalCode\Notebooks, Home = c:\Users\cplager
13:53:43.004 [info] User belongs to experiment group 'NewJupyterSession'
13:53:43.253 [error] Failure during activation. [TypeError: this.vscNotebook.onDidChangeNotebookCellExecutionState is not a function
	at new Rk (~\.vscode\extensions\ms-toolsai.jupyter-2023.8.1002501831\out\extension.node.js:24:501992)
	at f (~\.vscode\extensions\ms-toolsai.jupyter-2023.8.1002501831\out\extension.node.js:2:544315)
	at ~\.vscode\extensions\ms-toolsai.jupyter-2023.8.1002501831\out\extension.node.js:2:546205
	at t.resolveInstance (~\.vscode\extensions\ms-toolsai.jupyter-2023.8.1002501831\out\extension.node.js:2:546232)
	at ~\.vscode\extensions\ms-toolsai.jupyter-2023.8.1002501831\out\extension.node.js:2:549264
	at ~\.vscode\extensions\ms-toolsai.jupyter-2023.8.1002501831\out\extension.node.js:2:549601
	at ~\.vscode\extensions\ms-toolsai.jupyter-2023.8.1002501831\out\extension.node.js:2:548924
	at v (~\.vscode\extensions\ms-toolsai.jupyter-2023.8.1002501831\out\extension.node.js:2:548956)
	at ~\.vscode\extensions\ms-toolsai.jupyter-2023.8.1002501831\out\extension.node.js:2:548818
	at ~\.vscode\extensions\ms-toolsai.jupyter-2023.8.1002501831\out\extension.node.js:2:548752
	at Array.map (<anonymous>)
[...]

Issue #16973, opened 2025-09-11
error jupyter extension vscode


jeo20:
### Applies To

- [x] Notebooks (.ipynb files)
- [ ] Interactive Window and\/or Cell Scripts (.py files with \#%% markers)

### What happened?

With today's update 2023.8.1002501831, It doesn't recognize the interpreter for a virtual environment.
The extension started to fail in VScode:

	warn(message: string | Error, ...args: any[]): void {
		if (this.canLog(LogLevel.Warning)) {
			if (this.useColors) {
				console.warn('%c WARN', 'color: #993', message, ...args);
			} else {
				console.log(message, ...args);
			}
		}
	}

	error(message: string, ...args: any[]): void {
		if (this.canLog(LogLevel.Error)) {
			if (this.useColors) {
				console.error('%c  ERR', 'color: #f33', message, ...args);
			} else {
				console.error(message, ...args);
			}
		}
	}

### VS Code Version

1.103.2

### Jupyter Extension Version

2023.8.1002501831

### Jupyter logs

```shell

```

### Coding Language and Runtime Version

Python 3.13.7

### Running Jupyter locally or remotely?

None

jinwooYoon:
same problem ..

shepherdjerred:
fixed this by reverting to version `2025.7.0` of the extension

Michael-wzl:
Same problem! I cannot choose the conda kernel I have always been using. Judging from the following error output, I guess my version of Jupyter Notebook extension failed to catch up with the latest APIs of VS Code?

```
ERR [Extension Host] extension activation failed TypeError: this.vscNotebook.onDidChangeNotebookCellExecutionState is not a function
    at new Rk 
```

(I am using the extension on a remote server.)

Update: solved by selecting 2025.7.0 version.

ginofft:
Same problem, it does not recognize the kernel for a notebook, running on a remote machine using ssh. Revert to ealier version helps

pianojay:
Had a same problem with everything up to date:

`Extension activation failed, run the 'Developer: Toggle Developer Tools' command for more information. - Jupyter`

And the error logs:

workbench.desktop.main.js:sourcemap:34   ERR Extension 'ms-toolsai.jupyter' wants API proposal 'notebookCellExecutionState' but that proposal DOES NOT EXIST. Likely, the proposal has been finalized (check 'vscode.d.ts') or was abandoned.

[...]

PR #17027, opened 2025-10-08
Downgrade VS Code engine version requirement

shem8:
Since there's a bug that VSCode install a very old Jupyter version, and the fix is on VSCode 1.105 that is not yet fully released, we need to have the latest version to support the katest released of VSCode

related bugs:
https://github.com/microsoft/vscode-jupyter/issues/16977 https://github.com/microsoft/vscode-jupyter/issues/16973 https://github.com/microsoft/vscode/issues/270285

Fixes #

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
         "theme": "light"
     },
     "engines": {
-        "vscode": "^1.105.0"
+        "vscode": "^1.104.0"
     },
     "l10n": "./l10n",
     "extensionKind": [


shem8:
Relevant comment for the specific problem:
https://github.com/microsoft/vscode/issues/270285#issuecomment-3379107255

amunger:
this will be fixed in the next stable core version that will be shipped within the next two days, the same time as the next release of this extension.

Issue #17042, opened 2025-10-10
Unable to run notebook cells with spaces in path


dakenblack:
VSCode does not recognise ipykernel package within virtual environment. the issue is fixed by using an older version of the jupyter extension (2025.8.0).

## Environment data

-   Jupyter Extension version (available under the Extensions sidebar): 2025.9.0 (it works with 2025.8.0)
-   Python Extension version (available under the Extensions sidebar): 2025.16.0
-   Python and/or Anaconda version: 3.13
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): venv
-   Jupyter server running: Local | Remote | N/A

System:
<img width="363" height="252" alt="Image" src="https://github.com/user-attachments/assets/6ae882f5-68ac-4356-b70e-d0df7c837fd0" />

ipykernel==6.30.1 is installed in local venv.

## Expected behaviour

Notebook executes

## Actual behaviour

I get an error message:

<img width="502" height="127" alt="Image" src="https://github.com/user-attachments/assets/67caf30e-f856-4360-a597-b49350082978" />

If I click "Install" I get this error message in the output section of the jupyter cell:
<img width="1719" height="121" alt="Image" src="https://github.com/user-attachments/assets/c1aa8feb-3872-46f9-9b57-c01bd638f733" />

## Steps to reproduce:

Try to execute on empty notebook.

## Logs

<details>

<summary>Output for <code>Jupyter</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Jupyter</code>)
</summary>

<p>

```
14:07:57.089 [info] Starting Kernel (Python Path: ~\Documents\Projects\003 Data Projects\jw-parsers\.venv\Scripts\python.exe, Venv, 3.13.5) for '~\Documents\Projects\003 Data Projects\jw-parsers\experiments\20251010 asx\00 explore.ipynb' (disableUI=false)
14:07:57.092 [info] Process Execution: ~\Documents\Projects\003 Data Projects\jw-parsers\.venv\Scripts\python.exe -c "import ipykernel; print(ipykernel.__version__); print("5dc3a68c-e34e-4080-9c3e-2a532b2ccb4d"); print(ipykernel.__file__)"
[...]

PR #17058, opened 2025-10-10
Fix: Clarify 'no pip' error with hint about spaces in path

### LINKED ISSUE #17057: Kernel failure due to spaces in workspace path is not clearly reported
When a user's project is located in a folder path that contains spaces (Z:\AI for Industrial_ Applications\3rd Semester\Machine Learning) the Jupyter extension often fails to connect to or discover the kernel. The error messages that appear are generic (e.g., There is no Pip installer available in the selected environment) and do not indicate that the spaces in the file path are the root cause. Simply replacing the space with _ made the whole thing work. 

I spent a significant amount of time troubleshooting this, trying to reinstall ipykernel, checking my Python installation, and manually setting interpreter paths, none of which worked. The problem was only resolved by renaming the folders to remove the spaces.

prosckol:
Currently, when a user's project is in a folder path (e.g Z:\AI for Industrial Applications\3rd Semester\Machine Learning)  with spaces, kernel discovery can fail with the error "There is no Pip installer available in the selected environment." This message is misleading as it suggests a problem with the Python environment itself, rather than the file path. This leads to frustrating and time-consuming troubleshooting.

This change directly amends the error string to add a helpful tip. It informs the user that the issue may be caused by spaces or special characters in their workspace path, guiding them toward a common but difficult-to-diagnose solution.

Related Issue 
Fixes #17057

### DIFF
--- a/src/platform/common/utils/localize.ts
+++ b/src/platform/common/utils/localize.ts
@@ -837,7 +837,7 @@ export namespace Installer {
     export const noCondaOrPipInstaller = l10n.t(
         'There is no Conda or Pip installer available in the selected environment.'
     );
-    export const noPipInstaller = l10n.t('There is no Pip installer available in the selected environment.');
+    export const noPipInstaller = l10n.t('There is no Pip installer available in the selected environment.This may be caused by spaces or special characters in the workspace path, or an environment misconfiguration.');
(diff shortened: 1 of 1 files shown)


DonJayamanne:
Duplicate of https://github.com/microsoft/vscode-jupyter/issues/17042

Issue #15813, opened 2024-07-01
Dev process: npm i (or npm ci) freezes / takes too long


xl0:
### Applies To

- [ ] Notebooks (.ipynb files)
- [ ] Interactive Window and\/or Cell Scripts (.py files with \#%% markers)

### What happened?

I want to hack a bit. Following https://github.com/microsoft/vscode-jupyter/blob/main/CONTRIBUTING.md I ran
`npm ci`
and it hangs after a while, not showing any progress for over 15 minutes, and not showing up in top.
I see that in CI you use ` npm ci --ignore-scripts --prefer-offline --no-audit`

In my limited experience with node, I've never faced this issue, so not sure if it should be fixed, or the docs should be updated.

(I installed recommended node version through mamba, and then installed the recommended npm version)

### VS Code Version

recent

### Jupyter Extension Version

master

### Running Jupyter locally or remotely?

Local

xl0:
I also had to run
`npm run postinstall`
To actually get a working environment. But at least it worked. Is this a bug, or just lack of documentation?

DonJayamanne:
Sorry for never getting back to you. Are you still running into this issue.
We used to have similar issues, however that was related to an npm package that has since been removed.

xl0:
```
$ npm i
npm WARN deprecated crypto@1.0.1: This package is no longer supported. It's now a built-in Node module. If you've depended on crypto, you should switch to the one that's built-in.
npm WARN deprecated @types/del@4.0.0: This is a stub types definition. del provides its own type definitions, so you do not need this installed.
npm WARN deprecated postinstall-build@5.0.3: postinstall-build's behavior is now built into npm! You should migrate off of postinstall-build and use the new `prepare` lifecycle script with npm 5.0.0 or greater.
npm WARN deprecated circular-json@0.3.3: CircularJSON is in maintenance only, flatted is its successor.
npm WARN deprecated lolex@6.0.0: lolex has been renamed to @sinonjs/fake-timers. No API changes made. Please use the new package instead
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated vscode-debugprotocol@1.51.0: This package has been renamed to @vscode/debugprotocol, please update to the new name
[...]

PR #17096, opened 2025-10-17
docs: update setup to working npm ci variant

kevineger:
`npm ci` hangs indefinitely. This repros across multiple machines, in a dev container and in a GitHub Codespace.

Mimicking the GitHub Action install step, this change updates the docs to a working _flavour_ of `npm ci` to unblock contributors.

Fixes https://github.com/microsoft/vscode-jupyter/issues/15813 (which has been closed, but is happening again).

### DIFF
--- a/src/kernels/raw/session/rawKernelConnection.node.ts
+++ b/src/kernels/raw/session/rawKernelConnection.node.ts
@@ -570,6 +570,7 @@ async function postStartKernel(
             channel: 'control' as unknown as 'shell',
             username: kernel.username,
             session: kernel.clientId,
+            subshellId: kernel.subshellId,
             content: {}
         });
         kernel
--- a/src/platform/common/configSettings.ts
+++ b/src/platform/common/configSettings.ts
@@ -105,6 +105,8 @@ export class JupyterSettings implements IWatchableJupyterSettings {
             this.disableJupyterAutoStart = true;
         }
     }
+    completionTriggerCharacters?: Record<string, string[]> | undefined;
+    public logKernelMessages: boolean = false;
 
     public static getInstance(
         resource: Uri | undefined,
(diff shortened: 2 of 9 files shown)


DonJayamanne (approved):

kevineger:
@Tyriar I'm using the version of Node that's in the `.nvmrc`, with `nvm use` executed at the repo root ([source](https://github.com/microsoft/vscode-jupyter/blob/4efd36fb61d83bf0c99008648ca633ad688d4ab9/.nvmrc#L1)).

Tyriar:
I mean whatever the difference is. `npm ci` does run `postinstall`, this change is disabling that and running it explicitly.

kevineger:
Isn't there a difference? `npm ci --ignore-scripts && npm run postinstall` will run **no** scripts (that includes dependencies).

No doubt about it, running `npm ci` is simpler and should be expected to work. But somewhere, I suspect there's a dependency hanging things up.

Reading @DonJayamanne's old comment [here](https://github.com/microsoft/vscode-jupyter/issues/15813#issuecomment-2241825451), it looks like this isn't the first time this has happened.

---

Are you unable to repro this hang (following a full clean & removing `node_modules/`)? As mentioned in the description, I hit this across multiple machines and even tried a GH Codespace (to remove any possible network/proxy issues).

The right fix is to identify and fix the problematic dependency, just figured I'd send this PR as a workaround since I'm unable to look into the root issue right now.

Tyriar:
My comment was more @DonJayamanne, if he's fine with it then that's fine. I don't know enough about the project to debate it.

DonJayamanne:
@Tyriar The reason I'm ok with this is because we've seen `npm install` hanging due to the depenency on zeromq.
The plan is to remove this eventually, and thats the only one that has a posinstall script of compiling it.
Hence I'm happy to run `npm install --ignore-scripts`, besides we do that on CI because of this same reason.

DonJayamanne:
But getting this merged seems unlikely as there seems to be a lot of failing CI tests, 
looks like something has changed in some jupyter packages that has broken our CI 😞

Issue #17094, opened 2025-10-17
Jupyter API: A resolved server cannot be used on a notebook which had a server that was removed. NB controller disposed.


kevineger:
## Environment data

-   VS Code version: 1.105.0
-   Jupyter Extension version (available under the Extensions sidebar): 2025.9.1
-   Python Extension version (available under the Extensions sidebar): 2025.16.0
-   OS (Windows | Mac | Linux distro) and version: Debian
-   Python and/or Anaconda version: N/A
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A
-   Jupyter server running: Remote

## Repro Steps

This bug report is in extending the `Jupyter` API. I've implemented a `JupyterServerProvider` which resolves remote servers.

1. Open a new notebook.
2. Create and resolve a new server from the `JupyterServerProvider`.
3. Execute a code cell.
4. Remove the server, firing the `onDidChangeServers` event for the `JupyterServerProvider`. `provideJupyterServers` now returns no servers.
5. Observe that the kernel picker disconnects.
6. Create and resolve a new server from the `JupyterServerProvider`.
7. Execute a code cell.

## Expected behaviour

The cell can be executed.

## Actual behaviour

Nothing happens. The Notebook controller has been disposed. See logs below.

## Investigation notes

- If I don't execute a cell, which creates the session, (step 3 in repro) on the first server that gets removed, the bug doesn't repro.
- Each time I execute a cell on the second server (step 7), `Error: Cannot call start again` is logged to the extension host output.
- The only workaround is to re-load the file (which reconstructs the NB controller).

### Jupyter

I've uploaded the Jupyter logs: [jupyter-output-trace-logs.txt](https://github.com/user-attachments/files/22961317/jupyter-output-trace-logs.txt). To facilitate with debugging, I've separated the logs by the actions I took in the repro steps. The markers are:

- `~~~ Start handleCommand to resolve a server`
- `~~~ Click on one of the kernels, as returned by the kernelspecs`
- `~~~ Successfully execute a cell`
- `~~~ Remove a server, which fires the onDidChangeServers event`
- `~~~ Resolve the second server`
- `~~~ Select a kernel`
- `~~~ Execute a cell`

Ctrl+f for `~~~` should make jumping between the actions easier than perusing the verbose `trace` logs. Hope that helps with debugging!

### Extension host

Each time the cell is executed on the second server (whose notebook controller is disposed), this message is emitted:

```txt
[...]

PR #17097, opened 2025-10-17
fix: dispose a torn down NB controller's kernel

kevineger:
It's my first time looking at any of this code and the lifecycle of everything is a little challenging to discover at a glance. Sorry if this is overlooking anything obvious! This passes unit tests and I verified manually it solves [the issue I reported](https://github.com/microsoft/vscode-jupyter/issues/17094).

## Problem

If a server/kernel is removed, then another was created for the active notebook editor, cell executions against the new kernel would fail. The issue with clear repro steps and the corresponding logs: https://github.com/microsoft/vscode-jupyter/issues/17094.

## Why

The root cause is improper resource cleanup during the disposal of a `VSCodeNotebookController`. When a `JupyterServerProvider` is removed, the extension correctly disposes of the associated controllers. However, the `dispose` method for the controller did not explicitly dispose of the underlying `IKernel` instance that it had created and managed.

This left the kernel in an orphaned or "zombie" state. When a new server and controller were subsequently created for the same notebook, it'd try to use the old, still-lingering kernel, which was already in a started state, leading to the "Cannot call start again" error.

## Fix

When a controller is disposed, clean up any kernels it was responsible for. This is done by iterating through all currently open NB docs and disposing any kernels who's connection metadata IDs match.

---

Fixes https://github.com/microsoft/vscode-jupyter/issues/17094

### DIFF
--- a/.github/workflows/aggregate-test-results.yml
+++ b/.github/workflows/aggregate-test-results.yml
@@ -20,7 +20,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           persist-credentials: false
 
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -64,7 +64,7 @@ jobs:
     if: github.repository == 'microsoft/vscode-jupyter'
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           persist-credentials: false
 
@@ -83,7 +83,7 @@ jobs:
 
       # Caching of npm packages (https://github.com/actions/cache/blob/main/examples.md#node---npm)
       - name: Cache npm on linux/mac
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: ~/.npm
           key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
@@ -107,7 +107,7 @@ jobs:
     if: github.repository == 'microsoft/vscode-jupyter'
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           persist-credentials: false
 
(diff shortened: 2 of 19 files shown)


kevineger:
@microsoft-github-policy-service agree company="Google"

src/notebooks/controllers/vscodeNotebookController.ts
```diff
@@ -334,6 +334,13 @@ export class VSCodeNotebookController implements Disposable, IVSCodeNotebookCont
             } called from ${new Error('').stack}`
         );
         this.isDisposed = true;
+        // Dispose all kernels associated with this controller
+        workspace.notebookDocuments.forEach((doc) => {
```
DonJayamanne: Controller instances are not shared across notebooks.
We cannot dispose all other kernels associated with the same kernel.

DonJayamanne (changes requested):
This isn't the right solution as this will kill all kernels
E.g. assume you have a kernel for Python 3.14 and the id is 314
Now one can have two different notebooks running code against 314.
Both are two different instances and processes running against Python 3.14 (i.e. Kernel Id = 314).

Just because we close the first notebook we souldn't be killing the kernel instance running against the other notebook.

kevineger:
Thanks for taking a look and sorry I was off-base @DonJayamanne. I'm having a bit of a hard time understanding the life-cycle of everything, with all the wrapping and indirection.

Using the same repro as listed in the issue, if I create a 3rd party Jupyter server, execute code, delete it (provider returns `[]`) and then create another:

- `this.controller` [here](https://github.com/microsoft/vscode-jupyter/blob/4efd36fb61d83bf0c99008648ca633ad688d4ab9/src/notebooks/controllers/vscodeNotebookController.ts#L551) when the initial cell execution is kicked off uses the new/updated controller.
- But `kernel.controller.id !== controller.id` and so a second execution is started [here](https://github.com/microsoft/vscode-jupyter/blob/4efd36fb61d83bf0c99008648ca633ad688d4ab9/src/notebooks/controllers/vscodeNotebookController.ts#L578) using the old, disposed controller.

I tried taking out that line but all the `CellExecution` logic still maintains a tie to the disposed controller.

---

I'll try and spend some more time building an understanding of all this, but if you have any pointers or recommendations for a fix that'll save me time :+1:.

src/notebooks/controllers/vscodeNotebookController.ts
kevineger: The latest [commit](https://github.com/microsoft/vscode-jupyter/pull/17097/commits/82a5ecd2d8926105ba7e76457b0638753ec711aa) filters to the `associatedDocuments`. Is that sufficient? If there's a better way, please let me know!

DonJayamanne:
@kevineger I'm sorry for not getting back to you on this.
I've had a look at this issue, and need to make a few tweeks.
Please bear with me just one more week, as you can see there a number of failures in our pipelines (all related to a recent update to IPYKernel, with a number of breaking changes as well that were fixed upstream, I suspect there are more which is whats causing the pipelines to fall over).

Will look into those and this pr next week.
Thank you for your contribution and continued patience

DonJayamanne:
@kevineger I'm sorry this PR took a while. As you can see we have a few issues with our tests.
Once again thank you for your patience and thank you for this PR.

kevineger:
Hey @DonJayamanne - any ETA on when this fix will be released? Sorry to comment here, but https://github.com/microsoft/vscode-jupyter/issues/17094 was locked to collaborators via automation.

I tried pinning to the latest pre-release (`2025.10.2026010601`) and the issue still repros. Is this commit in there?

PR #17098, opened 2025-10-17
fix: first try widget downloads with custom fetch

kevineger:
Initially, I set out to fix a logging bug; if the _HTTP_ attempt (no custom fetch) failed but the custom fetch succeeded, the HTTP error would still be logged. After making that fix, I came to the opinion that it likely makes more sense to first try the custom fetch and then fallback. Otherwise we're issuing a ton of redundant requests.

This change also:
- Removes some duplication by lifting the two private `getWidgetScriptSource*` methods to an abstracted `downloadFile` function.
- Simplifies the logging. Previously the errors included _from ([httpclient|fetch])_ and the relevant `ex`. This duplicated the context (httpclient vs. fetch). Since the inner error contains this info, we can use the same top level error message when we log. 

With minimal context of the existing code, I can't easily determine... but for **reviewers with more background**: 

1. When would using the custom fetch ever fail?
2. Do we even need the fallback, or would it make more sense to strictly rely on the custom fetch?
3. Is the fallback behaviour I'm proposing here desirable or would you still prefer we issue both requests simultaneously and I instead change this PR to only log errors if both fail?

I looked to write or modify a unit test, but the `RemoteIPyWidgetScriptManager` doesn't appear to have existing coverage. If reviewers feel the change necessitates a test, I can look to cover the component. Though IMO that's out of scope given the precedent.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -271,7 +271,7 @@ jobs:
       matrix:
         jupyterConnection: [raw] # valid values include raw = uzing zqm, local = using local jupyter, remote = using remote jupyter, web = using remote jupyter in web mode
         python: [python] # valid values include python,conda,noPython
-        pythonVersion: ['3.10']
+        pythonVersion: ['3.12']
         # Whether we're using stable (<empty>) or pre-release versions of Python packages.
         # When installing pre-release versions, we're only focused on jupyter & related packages.
         # Not pre-release versions of pandas, numpy or other such packages that are not core to Jupyter.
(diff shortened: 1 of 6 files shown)


DonJayamanne:
Will try to get the CI resolved, have had a number of failures caused by some regressions from upstream ipykernel.
I suspect there are more regressions.
Need to go through and figure this out.
Will look into this next week

alexdima (approved):

kevineger:
Thanks!

PR #17296, opened 2026-02-17
feat: add "Find..." browse option to notebook kernel picker

### LINKED ISSUE #17281: Add ability to "browse" for kernels
When choosing the python interpreter for a python `.py` file  we have the "browse" option to choose the interpreter:

Image

But when choosing the Kernel for a notebook we don't

Image

It rely on faulty environments extension that fails to detect all the available environments properly, specially on multi project workspaces with the *.venvs* on subfolders,  so we can not choose them manually for the notebooks.

So having an equivalent "browse" option on the kernel selector would be great as by now getting it to choose the right one implies tricking the environment manager plugin by opening a `.py` file choosing manually the desired one and then go to the notebook to choose it **for each notebook and environment**.

As the feature itself is already present on the `.py` interpreter selector from environments extension it should not imply a complex addition to reuse or replicate it on the notebooks kernel selector.

ericcanton:
I realize that this was proposed and assigned in the issue referenced below. But I needed the solution sooner than later for my setup. I thought I should at least offer what I have in case it's useful.

Adds a file browser option to the Python environment kernel picker, allowing users to manually locate a Python interpreter on the filesystem. This addresses the gap where the Environments extension fails to detect all available environments, particularly for multi-project workspaces with virtual environments in subfolders or outside the workspace.

Fixes microsoft#17281

### DIFF
--- a/src/notebooks/controllers/kernelSource/localPythonKernelSelector.node.ts
+++ b/src/notebooks/controllers/kernelSource/localPythonKernelSelector.node.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { CancellationError, CancellationToken, NotebookDocument, commands, extensions, workspace } from 'vscode';
+import { CancellationError, CancellationToken, NotebookDocument, commands, extensions, window, workspace } from 'vscode';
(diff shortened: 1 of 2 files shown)


ericcanton:

PR #17351, opened 2026-03-25
Add executeSilentlyWithRetry and enable for some idempotent requests 

Deependra-Patel:
Sometimes execute_request message sent as we are connecting to a kernel gets lost due to connection resets etc, in that case the notebook is stuck in "Connecting to kernel ..." indefinitely for remote kernels.
This is because `execute_request` message is not retried.

`determineVersionImpl` runs as part of `initializeAfterStart`

TESTED=Added unit tests. 
All unit tests pass 
```npm run test:unittests

1173 passing (5s)
  232 pending
```
Tested manually where retry got triggered

### DIFF
--- a/.github/workflows/aggregate-test-results.yml
+++ b/.github/workflows/aggregate-test-results.yml
@@ -36,7 +36,7 @@ jobs:
           COLLECTION_DATE: ${{ inputs.collection_date }}
 
       - name: Upload test result files
-        uses: actions/upload-artifact@v4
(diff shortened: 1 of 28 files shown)


Deependra-Patel:
@microsoft-github-policy-service agree company="Google"

DonJayamanne:
@Deependra-Patel thank you for this PR.
Please can you file an issue so that we can track what this PR fixes

PR #17372, opened 2026-04-09
feat: make magic command regex configurable in interactive window

### LINKED ISSUE #16499: Add magic line comments regex pattern as configuration for uncommentMagicCommands
I found this [issue](https://github.com/microsoft/vscode-jupyter/issues/14686) that was closed and could not ask there. I am having the same use case, databricks uses a different magic line format, like:

```
# MAGIC %run ./another_notebook
```

When working with databricks connect locally, if I change that line to 

```
#! %run ./another_notebook
```

It does work locally, but not when running on databricks.

A possible solution would be to expose the pattern of the magic line comments as configuration. I can see it hardcoded in 
[uncommentMagicCommands](https://github.com/microsoft/vscode-jupyter/blob/d64c3b754bec5a54eb253da85ce1cb443bab88c6/src/interactive-window/editor-integration/cellFactory.ts#L18) and it would be in line with the cell marker regular expressions already exposed as configuration

DaveT1991:
Fixes #16499

The magic command regex pattern (`^#\s*!`) used by `uncommentMagicCommands` is currently hardcoded, making it impossible to support alternative magic comment formats such as Databricks-style `# MAGIC %run ./notebook`.

## Changes
- Add `magicCommandsRegex` to `IJupyterSettings` interface and `JupyterSettings` class (default: `^#\s*!`, preserving existing behaviour)
- Register config migration key in `ConfigMigration`
- Add `jupyter.interactiveWindow.textEditor.magicCommandsRegex` setting to `package.json`
- Update `uncommentMagicCommands` to accept optional settings — backward-compatible, existing callers unaffected
- Add unit tests covering default and custom regex behaviour

### DIFF
--- a/package.json
+++ b/package.json
@@ -1699,6 +1699,12 @@
                     "default": false,
                     "description": "%jupyter.configuration.jupyter.magicCommandsAsComments.description%"
                 },
+                "jupyter.interactiveWindow.textEditor.magicCommandsRegex": {
+                    "type": "string",
+                    "default": "^#\\s*!",
+                    "markdownDescription": "Regular expression pattern to identify magic line comments in the interactive window text editor. The matched prefix is stripped when executing code. Default (`^#\\s*!`) handles standard `#!` style. Example for Databricks: `^#\\s*MAGIC\\s*`.",
+                    "scope": "resource"
+                },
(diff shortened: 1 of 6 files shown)


DaveT1991:

Issue #17462, opened 2026-07-14
Notification for Cell Execution


Gravifer:
Rebubbling #15039 since it's closed and locked; no context changed.

Related:

- https://github.com/microsoft/vscode/issues/267459
- https://stackoverflow.com/questions/41947022/ui-notification-of-task-completion
- https://pypi.org/project/jupyternotify/
- various systems notification extensions like [this one](https://marketplace.visualstudio.com/items?itemName=ssv445.vscode-notify-desktop)

Cruder but still helful option: raise a modal dialog when a certain time have passed since user initiated evaluation, and Code is out of focus.

Gravifer:
### Prior art

Some useful precedents:

- [JupyterNotify](https://pypi.org/project/jupyternotify/) provides explicit `%%notify` and automatic notifications after a configurable minimum duration. Selecting the browser notification returns to the notebook tab.
- [JupyterLab Notifications](https://github.com/mwakaba2/jupyterlab-notifications) defaults to a 60-second threshold, distinguishes success and failure, and has a `last_cell_only` option to avoid notification storms during multi-cell execution.
- [Jupyter Cell Notifier](https://marketplace.visualstudio.com/items?itemName=KaimingCui.jupyter-cell-notifier) already provides much of the requested VS Code UX: thresholded cell notifications, success/failure reporting, and click-to-cell behavior. Its native notification implementation relies on platform-specific commands, so it does not solve remote or web execution cleanly.
- [MeerkatIO](https://marketplace.visualstudio.com/items?itemName=MeerkatIO.meerkatio) explicitly documents the same remote-host limitation: because the extension executes on the workspace host, native notification code can target the remote machine rather than the user’s desktop.
- Google Colab has a good UX precedent: notify after background execution finishes, subject to user permission and notification settings.

VS Code itself now has two particularly relevant first-party precedents:

- [Long-running task completion notifications](https://code.visualstudio.com/updates/v1_105#_os-notification-for-longrunning-task-completion), implemented from [microsoft/vscode#267459](https://github.com/microsoft/vscode/issues/267459), use a configurable duration threshold, notify only when the originating window is unfocused, and return focus to that window when selected.
- System notifications ("toasts") of the chat module support `off`, `windowNotFocused`, and `always` policies and route notification selection back to the relevant chat session.

These behaviors are a good model for notebooks.

### Existing building blocks

[...]

Issue #15039, opened 2024-01-19
Notification for Cell Execution


ahmadmustafaanis:
When a Cell is executed in Google Colab, it gives a notification for execution (Successfully or Failed).
![image](https://github.com/microsoft/vscode-jupyter/assets/47111429/1db44fbe-c4cc-4619-af2f-2d05f0eb2633)

Since VS Code is also based on Chrome, I believe it should be possible to add such a notification in VS Code Jupyter notebook for Cell Execution.

DonJayamanne:
Thanks for filing this feature request, will look into this.

ahmadmustafaanis:
Thank you, kindly let me know if I can also contribute in it.  My JS is not very good though, can we do it with `window.alert` or something similar?

DonJayamanne:
Given the fact taht there haven't been any upvotes nor any other similar issues, I'm closing this issue as not planned.

PR #16375, opened 2025-01-13
Add `waitUntil` for `onDidStart` event

pwang347:
For more context, see #16387

Issue #15613, opened 2024-04-26
Jupyter extension doesn't need to notify user of extension running code in their kernel nearly as often as it does.


greazer:
![WhyNotificationNotBusy](https://github.com/microsoft/vscode-jupyter/assets/11685408/8089849d-3e4c-40e7-a3bf-946e4501278f)

When my extension calls `executeCodeInKernel`, the jupyter extension always notifies me. While I understand the thought behind this, it seems like overkill and can make for a bad experience if an extension consistently executes code in the kernel.

Seems like a one-time notification the first time this occurs after a kernel has been started would be sufficient. This is enough to tell the user that an extension is indeed running code in their kernel. 

In addition, just making use of a kernel busy state indicator for additional extension executions, would be called for as well. The user could hover over this indicator and be given the reason that their kernel is busy. 

The stop button on the notebook is appropriate for user-invoked cell executions, but isn't appropriate for a "system-invoked" execution.

DonJayamanne:
This is required to ensure users are aware of the fact that the kernel is busy running some other code even though the user might not be aware of this.
Else users could end up complaining about the fact that Notebook/Kernel execution is very slow, only to find out that some other extension is busy running some code in the kernel.

Today we give a grace period of 500ms, if after 500ms the 3rd party code has not yet completed, we display a progress message.
I'll change the timeout from 500ms to 1s.

DonJayamanne:
> In addition, just making use of a kernel busy state indicator for additional extension executions, would be called for as well.

Unfortunately this isn't supported yet in VS Code API, and no plans for this just yet.

DonJayamanne:
> The stop button on the notebook is appropriate for user-invoked cell executions, but isn't appropriate for a "system-invoked" execution.

Not following what you mean by `system-invoked execution`
Are you suggesting that the execution from a co-pilot extension is deemed as `system-invoked`, i would have thought this is an extension instead of `system` as its not part of VS Code.

DonJayamanne:
> Seems like a one-time notification the first time this occurs after a kernel has been started would be sufficient. This is

Agreed, will display once per session.

> In addition, just making use of a kernel busy state indicator for additional extension executions, would be called for as well.

Yes

PR #17464, opened 2026-07-14
[WIP] Notification for Cell Execution

Gravifer:
Notifications are associated with notebook cell execution requests rather than generic kernel activity, so background code executed by other extensions does not produce cell-completion messages or duplicate the existing third-party execution progress behavior.

**What's done**:
- Long-running, user-requested notebook executions can produce completion notifications.
- Notifications are opt-in, resource-scoped, and configurable by duration and focus state.
- Failures can use a stronger notification mode than successful executions.
- Single-cell and grouped Run All/Run Selected executions produce one result-aware message.
- Results distinguish completed, failed, and stopped executions using internal execution data where VS Code’s summary is ambiguous.
- “Show Cell” returns to the relevant notebook and reveals the most useful cell.
- Works in both desktop and web extension hosts using stable APIs.
- Interactive Windows, non-Jupyter notebooks, stale cells, closed notebooks, and never-started requests are excluded.
- Existing completion bell ring and unrelated execution behavior remain unchanged.

The message bodies as well as setting names could use more thinking.

**Blocking problem**: the issue calls for system notifications, while at the point we are just poping vs code host ones.

Addresses #15039 #17462


Gravifer:
This work builds on an existing distinction in the Jupyter extension between user-requested notebook execution and code executed invisibly by another extension. The history around #15613 is useful prior art, but it addresses the opposite half of the user experience from #17462 .

Behavior at #15613 was opened: 
Introduced third-party kernel execution visibility in 2024 as part of #15139 ([concerned commit](https://github.com/microsoft/vscode-jupyter/commit/c0f8b15)). When another extension executed code through the Jupyter kernel API, the Jupyter extension could show an in-editor progress notification identifying both the kernel and responsible extension:

> Executing code in {kernel} from {extension}

It already incorporated several useful constraints:

- It waited until the request was actually sent for execution.
- It suppressed executions shorter than 500 milliseconds.
- It showed the notification only while the associated notebook was visible.
- It kept the progress notification active until execution completed.
- It attributed the otherwise-hidden activity to the extension that initiated it.
- By #15230, Jupyter's own internal API executions were excluded from this third-party notification path.

This provided important transparency: users could understand why a kernel was busy even when they had not run a cell themselves. However, every sufficiently long third-party execution could produce another popup, which promted #15613. 

The issue proposed a more restrained division of responsibilities: show a one-time notification to establish that an extension is executing code, then represent subsequent activity through a quieter kernel or notebook busy indicator.

#15629 refined the behavior in several ways:

- Grace period increased from .5s to 1s.
- The popup was limited to once per extension, per kernel.
- Third-party work was represented using VS Code’s notebook-level `NotebookExecution` state.
- Concurrent third-party executions shared that state through reference counting, keeping the notebook busy until the last execution finished.
- The user-cell execution path and the system-initiated execution path remained conceptually separate.

This established a two-layer model:

1. A one-time, attributable notification provides transparency.
2. Notebook-level execution progress quietly represents continuing background activity.

-----

Before this PR, the implementation had been moved into a dedicated `kernelProgressIndicator.ts` component as part of #16375 ([commit](https://github.com/microsoft/vscode-jupyter/commit/5fe5bb5)). The extraction also made notebook-execution acquisition tolerant of conflicts—for example, when a user cell already owns the notebook execution state—and made cleanup defensive.

The resulting implementation can be seen in [`kernelProgressIndicator.ts` on main](https://github.com/microsoft/vscode-jupyter/blob/a2d5d27/src/standalone/api/kernels/kernelProgressIndicator.ts), answering:

> Why is another extension currently executing hidden code in my kernel?

But it does not answer a different question raised by #17462:

> Has the long-running cell or batch I requested now finished?

-----

### What this PR adds

This PR addresses that missing completion side of the experience. It associates notifications with notebook-controller cell execution requests rather than generic kernel busy/idle transitions. That boundary is important: background work initiated through the extension API does not produce cell-completion messages or duplicate the existing third-party progress behavior.

The new behavior adds:

- An opt-in completion notification mode.
- A configurable minimum execution duration.
- Focus-sensitive and always-notify modes.
- A separate failure mode that may strengthen, but not weaken, the general mode.
- Single-cell and requested-batch timing and aggregation.
- Completed, failed, and stopped result classification.
- One notification per qualifying execution request.
- Navigation back to the relevant completed, failed, or stopped cell.
- Desktop and web support through stable VS Code in-editor notification APIs.

The default remains `off`, with a 60-second threshold when enabled. This conservative default is consistent with the notification-fatigue lesson from #15613.

The existing and new mechanisms are therefore complementary:

- Third-party kernel progress explains ongoing system-initiated activity.
- Cell-completion notifications report the end of long-running user-requested work.

Neither mechanism relies on generic kernel-idle notifications, which would conflate user execution with extension-injected code. The first version also intentionally uses stable in-editor notifications; native operating-system toasts remain unavailable to ordinary extensions through the stable VS Code API.

Gravifer:
I think it's now in good enough shape that I won't be embarrassed if maintainers took a look at it.

DonJayamanne:
Thank you for creating this PR, however we cannot take this, please see comments here  https://github.com/microsoft/vscode-jupyter/issues/17462#issuecomment-4981615678

